PR #1582, opened 2019-10-01
Add Parameter File Option to `run cloud` CLI command

### LINKED ISSUE #1551: Add Parameter Specification to run CLI command
## Current behavior
Currently the CLI command doesn't allow for the specification of parameters for a flow run

## Proposed behavior
Couple options we could take here:

Multiple --parameter NAME=VALUE options (e.g. each --parameter similar to how docker build does --build-arg)

Single --parameters {name: value} that takes a dictionary

Also we should have a file option where you can provide a json file

joshmeek:

## What does this PR change?
Adds a `--parameters` option to the `prefect run cloud` CLI command for flow run parameter specification. Takes in a json file of parameters to use.

I chose to go the route of providing a parameter file specification and am open to adding other methods if desired. This felt more natural.

Closes #1551 

## Why is this PR important?
Previously we were unable to specify parameters for this function

### DIFF
--- a/src/prefect/cli/run.py
+++ b/src/prefect/cli/run.py
@@ -1,3 +1,4 @@
+import json
 import time
 
 import click
@@ -46,6 +47,12 @@ def run():
 )
 @click.option("--version", "-v", type=int, help="A flow version to run.", hidden=True)
 @click.option(
+    "--parameters",
+    help="A parameters JSON file.",
+    hidden=True,
+    type=click.Path(exists=True),
+)
+@click.option(
     "--watch",
     "-w",
     is_flag=True,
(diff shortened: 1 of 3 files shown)


src/prefect/cli/run.py
```diff
@@ -107,7 +115,11 @@ def cloud(name, project, version, watch, logs):
         click.secho("{} not found".format(name), fg="red")
         return
 
-    flow_run_id = client.create_flow_run(flow_id=flow_id)
+    loaded_params = None
+    if parameters:
+        loaded_params = json.load(open(parameters))
```
cicdw: this looks like it could leave the file open, we should probably use a context manager

joshmeek: Ah yes this should be using a context manager 👍

joshmeek:
@cicdw The other (using click) would be to pass in a string representing a json payload. I'm fine with adding it if wanted but it seemed ugly at the time

joshmeek:
@cicdw I separated out a `--parameters-file` and `--parameters-string` option

src/prefect/cli/run.py
```diff
@@ -107,7 +130,20 @@ def cloud(name, project, version, watch, logs):
...
+        with open(parameters_file) as params_file:
+            file_params = json.load(params_file)
+
+    # Load parameters from string if provided
+    string_params = {}
+    if parameters_string:
+        string_params = json.loads(parameters_string)
+
+    flow_run_id = client.create_flow_run(
+        flow_id=flow_id, parameters={**file_params, **string_params}
```
cicdw: Niiiice thinking

PR #1592, opened 2019-10-03
Fargate Task Environment

### LINKED ISSUE #1549: Fargate Task Environment
## Use Case
Currently when using AWS Fargate the only supported environment is the Remote Environment with a specific set of configuration. We want to have an environment which allows users of the Fargate Agent to deploy their flows using a completely custom Fargate task.

## Solution
Create an environment which takes in a Fargate task definition

## Alternatives
N/A

joshmeek:

## What does this PR change?
Adds the `FargateTaskEnvironment`
Closes #1549 

## Why is this PR important?
Works in a similar way to the `KubernetesJobEnvironment` by allowing users to specify completely custom task definitions for their flow runs

### DIFF
--- a/src/prefect/environments/execution/fargate/fargate_task.py
+++ b/src/prefect/environments/execution/fargate/fargate_task.py
@@ -19,16 +19,18 @@ class FargateTaskEnvironment(Environment):
 
     def __init__(
         self,
-        labels: List[str] = None,
         aws_access_key_id: str = None,
         aws_secret_access_key: str = None,
         region_name: str = None,
+        labels: List[str] = None,
         **kwargs
     ) -> None:
         # Not serialized, only stored on the object
-        self.aws_access_key_id = aws_access_key_id
-        self.aws_secret_access_key = aws_secret_access_key
-        self.region_name = region_name
+        self.aws_access_key_id = aws_access_key_id or os.getenv("AWS_ACCESS_KEY_ID")
+        self.aws_secret_access_key = aws_secret_access_key or os.getenv(
+            "AWS_SECRET_ACCESS_KEY"
+        )
+        self.region_name = region_name or os.getenv("REGION_NAME")
 
         # Parse accepted kwargs for definition and run
         self.task_definition_kwargs, self.task_run_kwargs = self._parse_kwargs(kwargs)
@@ -73,12 +75,12 @@ class FargateTaskEnvironment(Environment):
         task_definition_kwargs = {}
         for key, item in user_kwargs.items():
             if key in definition_kwarg_list:
-                self.task_definition_kwargs.update({key: item})
+                task_definition_kwargs.update({key: item})
 
         task_run_kwargs = {}
         for key, item in user_kwargs.items():
             if key in run_kwarg_list:
-                self.task_run_kwargs.update({key: item})
+                task_run_kwargs.update({key: item})
 
         return task_definition_kwargs, task_run_kwargs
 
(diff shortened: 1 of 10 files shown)


src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -0,0 +1,279 @@
...
+from prefect.environments.storage import Docker
+
+# from boto3 import client as boto3_client
+# from botocore.exceptions import ClientError
+
+
+class FargateTaskEnvironment(Environment):
+    """
+    FargateTaskEnvironment is an environment which deploys your flow (stored in a Docker image)
+    as a Fargate tasl. This environment requires AWS credentials and extra boto3 kwargs which
```
cicdw:
```suggestion
    as a Fargate task. This environment requires AWS credentials and extra boto3 kwargs which
```

src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -0,0 +1,279 @@
...
+
+import cloudpickle
+
+import prefect
+from prefect import config
+from prefect.environments.execution import Environment
+from prefect.environments.storage import Docker
+
+# from boto3 import client as boto3_client
+# from botocore.exceptions import ClientError
```
cicdw: nbd but maybe remove these?

src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -0,0 +1,279 @@
...
+class FargateTaskEnvironment(Environment):
+    """
+    FargateTaskEnvironment is an environment which deploys your flow (stored in a Docker image)
+    as a Fargate tasl. This environment requires AWS credentials and extra boto3 kwargs which
+    are used in the creation and running of the Fargate task.
+
+    When providing a custom container definition spec the first container in the spec must be the
+    container that the flow runner will be executed on.
+
+    These environment variables are required for cloud do not need to be included because
```
cicdw:
```suggestion
    These environment variables are required for cloud but do not need to be included because
```

src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -0,0 +1,279 @@
...
+    https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task
+
+    The secrets and kwargs that are provided at initialization time of this environment
+    are not serialized and will only ever exist on this object.
+
+    Args:
+        - aws_access_key_id (str, optional): AWS access key id for connecting the boto3
+            client. Defaults to the value set in the environment variable
+            `AWS_ACCESS_KEY_ID`.
+        - aws_secret_access_key (str, optional): AWS secret access key for connecting
```
cicdw: Should we also expose a method of pulling these from Prefect Secrets? jw

joshmeek: Like an optional flag to say whether or not to grab these from a Secret?

joshmeek: Would make sense but the initialization will happen locally when building the environment

cicdw: Hmm more like:
```
- aws_credentials_secret (str, optional)
```
like on our AWS tasks which refers to a Secret name that is only pulled at runtime.  We don't have to do that for this PR but something to maybe think about.

joshmeek: Yeah let's put that in a future enhancement, after this is merged I'll open an issue for it. I like the idea

PR #1600, opened 2019-10-07
Allow authentication to Azure Blob Storage using SAS token

frsann:

## What does this PR change?
Allows authentication to Azure Blob Storage using SAS tokens. This concerns `AzureResultHandler` as well as the tasks `BlobStorageDownload`, `BlobStorageUpload`, and `DatastoreRegisterBlobContainer`.

## Why is this PR important?
Allows Blob Storage users to authenticate more securely.

### DIFF
--- a/src/prefect/tasks/azure/blobstorage.py
+++ b/src/prefect/tasks/azure/blobstorage.py
@@ -14,7 +14,7 @@ class BlobStorageDownload(Task):
     Args:
         - azure_credentials_secret (str, optional): the name of the Prefect Secret
             that stores your Azure credentials; this Secret must be a JSON string
-            with two keys: `ACCOUNT_NAME` and `ACCOUNT_KEY`
+            with two keys: `ACCOUNT_NAME` and either `ACCOUNT_KEY` or `SAS_TOKEN`
         - container (str, optional): the name of the Azure Blob Storage to download from
         - **kwargs (dict, optional): additional keyword arguments to pass to the
             Task constructor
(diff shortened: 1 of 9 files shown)


src/prefect/tasks/azure/blobstorage.py
```diff
@@ -57,10 +57,13 @@ def run(
...
         az_account_name = azure_credentials["ACCOUNT_NAME"]
-        az_account_key = azure_credentials["ACCOUNT_KEY"]
+        az_account_key = azure_credentials.get("ACCOUNT_KEY")
+        az_sas_token = azure_credentials.get("SAS_TOKEN")
 
         blob_service = azure.storage.blob.BlockBlobService(
-            account_name=az_account_name, account_key=az_account_key
+            account_name=az_account_name,
+            account_key=az_account_key,
+            sas_token=az_sas_token,
```
joshmeek: What happens if both `account_key` and `sas_token` are provided? Is it rightfully handled or is an error raised?

frsann: Good question. 

For `Datastore.register_azure_blob_container` it basically comes down to a call to `azureml.data.datastore_client._DatastoreClient._get_credential_type(account_key, sas_token)` which gives preference to `account_key`. No error. 

For `BlockBlobService` we initiate its base class `azure.storage.blob.baseblobservice.BaseBlobService`, which follows the same logic; `account_key` is given preference, and no error. 

This functionality should maybe be mentioned in the docstrings?

joshmeek: Yeah! Just add a quick note about preference!

frsann: Updated!

PR #1606, opened 2019-10-08
Improved Docker Tasks' Logging

alexisprince1994:

## What does this PR change?
This PR allows for better debug logging of docker tasks.

## Why is this PR important?
Debug logs help debug errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -842,6 +842,7 @@ Released August 20, 2018
 - Automatic versioning via Versioneer - [#70](https://github.com/prefecthq/prefect/issues/70)
 - `TriggerFail` state - [#67](https://github.com/prefecthq/prefect/issues/67)
 - State classes - [#59](https://github.com/prefecthq/prefect/issues/59)
+- Additional debug logs to `Docker Container` and `Docker Image` tasks - [#920](https://github.com/PrefectHQ/prefect/issues/920)
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Raise more informative error when calling `flow.visualize()` if Graphviz executable not installed - [#1602](https://github.com/PrefectHQ/prefect/pull/1602)
 - Allow authentication to Azure Blob Storage with SAS token - [#1600](https://github.com/PrefectHQ/prefect/pull/1600)
+- Additional debug logs to `Docker Container` and `Docker Image` tasks - [#920](https://github.com/PrefectHQ/prefect/issues/920)
 
 ### Task Library
 
@@ -842,7 +843,6 @@ Released August 20, 2018
 - Automatic versioning via Versioneer - [#70](https://github.com/prefecthq/prefect/issues/70)
 - `TriggerFail` state - [#67](https://github.com/prefecthq/prefect/issues/67)
 - State classes - [#59](https://github.com/prefecthq/prefect/issues/59)
-- Additional debug logs to `Docker Container` and `Docker Image` tasks - [#920](https://github.com/PrefectHQ/prefect/issues/920)
 
 ### Fixes
 
--- a/src/prefect/tasks/docker/images.py
+++ b/src/prefect/tasks/docker/images.py
@@ -1,3 +1,4 @@
+import itertools
 import json
 from typing import Any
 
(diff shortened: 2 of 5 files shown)


src/prefect/tasks/docker/images.py
```diff
@@ -119,8 +120,27 @@ def run(
...
+        self.logger.debug(
+            "Starting docker pull for repository {repo} with tag {tag}...".format(
+                repo=repository, tag=tag
+            )
+        )
+        feed = client.pull(repository=repository, tag=tag, stream=True)
+
+        output = ""
+        return_stream, log_stream = itertools.tee(feed)
+
```
alexisprince1994: Planning on removing the itertools.tee call, as I remember a slack chat that Prefect isn't "not not threadsafe", so just going to append to the output string.

alexisprince1994:
@joshmeek I wanted to get some initial feedback on the implementation. I'll add logging in the same style for the rest, but wanted to make sure I wasn't off base before doing so.

tests/tasks/docker/test_images.py
```diff
@@ -105,6 +107,81 @@ def test_repository_is_replaced(self, monkeypatch):
         task.run(repository="test")
         assert api.return_value.pull.call_args[1]["repository"] == "test"
 
+    def test_logs_docker_stream_values(self, monkeypatch, caplog):
+
+        task = PullImage(repository="original")
+
+        api = MagicMock()
+        docker_output = ["first message", "second message", "third message"]
```
alexisprince1994: Placeholder fake data. Going to grab the real logs from downloading busybox and use here instead.

src/prefect/tasks/docker/images.py
```diff
@@ -119,8 +120,27 @@ def run(
             raise ValueError("A repository to pull the image from must be specified.")
 
         client = docker.APIClient(base_url=docker_server_url, version="auto")
+        self.logger.debug(
+            "Starting docker pull for repository {repo} with tag {tag}...".format(
+                repo=repository, tag=tag
+            )
+        )
+        feed = client.pull(repository=repository, tag=tag, stream=True)
```
alexisprince1994: The output logs from the call when `stream=True`, aren't identical to those being output when `stream=False`. Should we be worried about backwards compatibility issues? If so, we can log the output from `stream=False` and just return that?

joshmeek: Hmm yeah this is an interesting one. I'm not sure if the best way to add logs here would be to actually log the output of `stream=True` or add a log before the `.pull` what says something along the lines of `"Pulling {}:{}".format(repository, tag)`. This is because the pull stream output is quite massive and would add a lot of noise to the logs. I vote for a simple `Pulling` log before the call and changing the return to be the output of the stream.

src/prefect/tasks/docker/images.py
joshmeek: Actually not fully sure about this one but simply appending to the output string wouldn't hurt.

cc @cicdw

tests/tasks/docker/test_images.py
joshmeek: You're probably going to want to keep it mocked with fake data. Unit tests which actually pull images aren't very portable and I know CircleCI throws a fit.

joshmeek:
This looks really great! The reason your tests are failing is because black formatting needs to be applied. Try running `black .` at the root directory of the repo and then committing those changes

src/prefect/tasks/docker/images.py
```diff
@@ -1,3 +1,4 @@
+import itertools
```
joshmeek: Unused import, can be removed

alexisprince1994:
> This looks really great! The reason your tests are failing is because black formatting needs to be applied. Try running `black .` at the root directory of the repo and then committing those changes

Super strange, I have VSCode set up to automatically run Black on save, but it looks like it didn't in that case. Ran it from the terminal and it fixed it up. Thanks!

CHANGELOG.md
```diff
@@ -845,6 +845,7 @@ Released August 20, 2018
 - Automatic versioning via Versioneer - [#70](https://github.com/prefecthq/prefect/issues/70)
 - `TriggerFail` state - [#67](https://github.com/prefecthq/prefect/issues/67)
 - State classes - [#59](https://github.com/prefecthq/prefect/issues/59)
+- Additional debug logs to `Docker Container` and `Docker Image` tasks - [#920](https://github.com/PrefectHQ/prefect/issues/920)
```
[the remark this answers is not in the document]
alexisprince1994: Done, thanks!

src/prefect/tasks/docker/images.py
alexisprince1994: Removed

PR #1607, opened 2019-10-08
Enhancement Fargate Agent - temporary credentials 

mhmcdonald:

## What does this PR change?
This PR adds in functionality to support temporary credentials and IAM role based credentials to the Fargate agent by introducing the aws_session_token per documentation found here: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html

## Why is this PR important?
This is important because using the existing state of the agent does not allow you to use temporary credentials. Also, it should never be required to supply access keys when using a library inside of AWS compute.


joshmeek:
This is great @mhmcdonald! It seems like your commits aren't linked to you as a user (https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user) and show up as the grey Octocat. Would you like to adjust that and reopen this PR?

PR #1609, opened 2019-10-09
Fix issue with merge task and comparisons

cicdw:

## What does this PR change?
This PR updates the `merge` task to do an `isinstance` check instead of a direct comparison; this allows for the merge task to work with objects that don't have a robust `__bool__` method such as pandas dataframes and numpy arrays.

## Why is this PR important?
We should strive to be agnostic to what types of objects tasks return

### DIFF
--- a/tests/tasks/test_control_flow.py
+++ b/tests/tasks/test_control_flow.py
@@ -1,4 +1,3 @@
-import numpy as np
 import pytest
 
 import prefect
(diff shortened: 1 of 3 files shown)


tests/tasks/test_control_flow.py
```diff
@@ -95,6 +95,29 @@ def test_merging_diamond_flow():
         assert isinstance(state.result[merge_task], Success)
 
 
+def test_merging_with_objects_that_cant_be_equality_compared():
+    class SpecialObject:
+        def __eq__(self, other):
+            return self
+
+        def __bool__(self):
+            raise SyntaxError("You can't handle the truth!")
```
joshmeek: Great message

PR #1610, opened 2019-10-09
Introducing: Environment Callbacks

### LINKED ISSUE #1495: Storage + Execution Concept Docs
In our Cloud docs section, we need to add a Storage / Execution concept doc.  It can re-use some of our other wording, but I think deserves a section.

### LINKED ISSUE #1574: Add on_start / on_exit callbacks for all Environments
## Use Case
I have a situation where my Flow run job exits (oftentimes because tasks become zombies), and I'd like to be notified of that exit.  Having the ability to register an `on_exit` or `at_exit` callback on my environment would achieve that purpose.  

Similarly, if I want to broadcast info to my Dask cluster (e.g., configure my dask workers to log to Prefect), I could use the `on_start` callback.

### LINKED ISSUE #1581: PIN 12: Environment Callbacks
# PIN: Environment Callbacks

Date: October 1, 2019
Author: Josh Meek

# Status
Proposed

# Context
Prefect Environments currently support a `setup` and `execute` paradigm when it comes to creation of the environment and using it to execute a flow (outlined in  [PIN-3](https://docs.prefect.io/core/PINs/PIN-03-Agent-Environment.html#process-details)). These two primary functions are key for infrastructure related processes which may arise during the execution of a flow. While they mainly deal with infrastructure creation there is another request which has arisen through use of the system in the form of pre/post-processing of the flow execution itself from the environment level. 

# Proposal
Implement `on_start` and `on_exit` callback options to environments which users can use to provide _optional_ functions which execute before the flow is run and after the run has ended (or the job has exited). These functions will provide a mechanism for users to create custom hooks and notifications for their environment execution. The callbacks are meant more as a tool for execution and are not intended to replace aspects such as state handlers.

The new Environment class signature would look like:
```python
Environment(labels: Iterable[str] = None,
            on_start: Callable = None,
            on_exit: Callable = None)
```

### Callback Examples
`on_start`
- Broadcast info to a Dask cluster (e.g. `register_worker_callbacks`)
- Collect available resources for worker allocation

`on_exit`
- Send custom notification if a task became a zombie
- Collect metrics / logs from resources created by the environment
 
# Consequences

Users will be able to specify `on_start` and `on_exit` callbacks for their environments. One new requirement would be that users of these callbacks have to be aware of where these callbacks occur because it will be different for each environment. While they all do still run a flow in a main process somewhere they differ in where that main process happens.

An example of this is the `RemoteEnvironment`, deployed on Kubernetes, will run the main process on the initial Prefect job while the `KubernetesJobEnvironment` will run the main process on a separate custom job that spawns after the initial Prefect job.

Another consequence that still falls into the awareness category is that of knowing which callbacks are actually useful for the specified method of execution. One example to keep in mind is how a custom Dask `on_start` callback will not be entirely useful if a flow is not using a Dask executor.

# Actions
Implement #1574

Decide whether or not the failure of a callback will be allowed to fail the main process of the Environment. For example, if the `on_start` callback fails does the flow execution still occur or is the failure logged as a failure for that particular flow run?

Add example callbacks which users could use out of the box.

joshmeek:

## What does this PR change?
This PR is an implementation of PIN-12 for environment `on_start` and `on_exit` callbacks. It also adds a short concept doc for Environments and Storage.
Closes #1495 
Closes #1574 
Closes #1581 

## Why is this PR important?
Environment related callbacks are important tools for performing work outside of environment infrastructure and flow operations.


docs/cloud/concepts/execution.md
```diff
@@ -0,0 +1,72 @@
...
+or assign it directly:
+
+```python
+from prefect.environments.storage import Docker
+
+f = Flow("example-storage")
+f.storage = Docker(registry_url="prefecthq/storage-example")
+```
+
+When you deploy your flow to Prefect Cloud the storage object attached to the flow will be built. At this step the flow is serialized to byte code and placed inside of the storage. For added convenience, `flow.deploy` will accept arbitrary keyword arguments which will then be passed to the initialization method of your configured default storage class (which is `Docker` by default). Consequently, the following code will actually create a `Docker` object for you at deploy-time and push that image to your specified registry:
```
cicdw: I feel like it's worth calling out that users can optionally build the storage prior to deploy using `storage.build()` and then `build=False` in the `deploy` command.

src/prefect/environments/execution/dask/k8s.py
```diff
@@ -265,6 +274,10 @@ def run_flow(self) -> None:
             )
             raise exc
 
+        # Call on_exit callback if specified
+        if self.on_exit:
+            self.on_exit()
```
cicdw: I wonder if users will expect that this is called _no matter what_ (meaning, prior to the exception raise above).  What do you think?

joshmeek: I was considering that as well and I meant to add a comment to discuss. I think it should happen regardless. Should I move it into a `finally`?

cicdw: Yup, agreed - a `finally` sounds great.

PR #1612, opened 2019-10-14
Add Optional "Name" to Agents

### LINKED ISSUE #1611: Add Optional "Name" to Agents
- in init
- in CLI `--name`

and the logger name will adjust accordingly.

joshmeek:

## What does this PR change?
Closes #1611 
Adds a `name` init arg to agents that can also be set through the `AGENT_NAME` environment variable

## Why is this PR important?
Being able to name agents is a great tiny feature to have for division of work and clarity.

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import time
 from typing import Union
 
@@ -38,11 +37,11 @@ class Agent:
 
     Args:
         - name (str, optional): An optional name to give this agent. Can also be set through
-            the environment variable `AGENT_NAME`. Defaults to "agent"
+            the environment variable `PREFECT__CLOUD__AGENT__NAME`. Defaults to "agent"
     """
 
     def __init__(self, name: str = None) -> None:
-        self.name = name or os.getenv("AGENT_NAME", "agent")
+        self.name = name or config.cloud.agent.get("name", "agent")
 
         token = config.cloud.agent.get("auth_token")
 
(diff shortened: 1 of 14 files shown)


src/prefect/agent/agent.py
```diff
@@ -34,16 +35,21 @@ class Agent:
 
     In order for this to operate `PREFECT__CLOUD__AGENT__AUTH_TOKEN` must be set as an
     environment variable or in your user configuration file.
+
+    Args:
+        - name (str, optional): An optional name to give this agent. Can also be set through
+            the environment variable `AGENT_NAME`. Defaults to "agent"
```
jlowin: I suggest PREFECT_AGENT_NAME to keep all configs immediately visible via `env | sort`. This won't get put in the config (although it could?) because there's no double underscore.

joshmeek: It _could_ get put in the config but @cicdw and I had said offline that it doesn't feel like a needed config option. However it definitely could get put in if we think it would be a desired option!

src/prefect/cli/agent.py
```diff
@@ -43,27 +43,31 @@ def agent():
...
-@click.argument("name", default="local")
+@click.argument("agent-option", default="local")
 @click.option(
     "--token", "-t", required=False, help="A Prefect Cloud API token.", hidden=True
 )
 @click.option(
     "--verbose", "-v", is_flag=True, help="Enable verbose agent logs.", hidden=True
 )
+@click.option(
+    "--name", "-n", required=False, help="A name to use for the agent", hidden=True
```
cicdw: could you explicitly set `default=None` here? I'm worried about a future in which click passes `""` as the name and it gets used accidentally.

PR #1614, opened 2019-10-14
Fargate Agent configuration options enhancements

mhmcdonald:

- [x ] adds new tests (if appropriate)
- [x ] updates `CHANGELOG.md` (if appropriate)

## What does this PR change?
This PR makes a few changes to the Fargate Agent. The first change is that it allows subnets and security groups to be set based on environment variables. The second change is that it adds two AWS configuration options to the Fargate Agent: `task_role_arn` and `execution_role_arn`.

## Why is this PR important?
This PR is important because Fargate requires the task definition to have execution role ARN to support ECR images. These configurations are necessary for the IAM role of the agent to be properly set.

### DIFF
--- a/src/prefect/agent/fargate/agent.py
+++ b/src/prefect/agent/fargate/agent.py
@@ -51,6 +51,8 @@ class FargateAgent(Agent):
         aws_access_key_id: str = None,
         aws_secret_access_key: str = None,
         aws_session_token: str = None,
+        task_role_arn: str = None,
+        execution_role_arn: str = None,
         region_name: str = None,
         cluster: str = None,
         subnets: list = None,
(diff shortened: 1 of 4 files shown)


joshmeek:
This is really great @mhmcdonald! I'm excited to get these extra options into the Fargate Agent!

mypy is failing for two minor reasons:
```
src/prefect/agent/fargate/agent.py:104: error: Item "None" of "Optional[str]" has no attribute "split"
src/prefect/agent/fargate/agent.py:112: error: Item "None" of "Optional[str]" has no attribute "split"
```

src/prefect/agent/fargate/agent.py
```diff
@@ -73,13 +84,34 @@ def __init__(
...
         # Agent task config
+        self.task_role_arn = task_role_arn or os.getenv("TASK_ROLE_ARN")
+        self.logger.debug("Task role arn {}".format(self.task_role_arn))
+
+        self.execution_role_arn = execution_role_arn or os.getenv("EXECUTION_ROLE_ARN")
+        if not self.execution_role_arn:
+            self.logger.exception(
+                "Fargate requires task definition to have execution role ARN to support ECR images"
+            )
+            raise Exception("Fargate task execution role required")
```
joshmeek: I don't think this Exception needs to be raised however a warning could still be logged. The reason being (for example) in my set up I am using images based in DockerHub and this `execution_role_arn` wasn't required.

mhmcdonald: Sounds good - I'm happy to remove the Exception and just log the warning. The reason we did this was because without the `execution_role_arn`, the iam role for the agent wasn't being set properly and this fixed it. In your aws environment is an iam role being used for the fargate tasks? Is it somehow defaulting to a iam role? (Just curious). I will update shortly.

joshmeek: For mine I'm am actually not using an IAM role so using one may actually warrant the need of the `execution_role_arn` I will test various set ups / configurations and see if that will affect it. Thanks for the feedback 👍

src/prefect/agent/fargate/agent.py
```diff
@@ -23,15 +24,23 @@ class FargateAgent(Agent):
         - aws_session_token (str, optional): AWS session key for connecting the boto3
             client. Defaults to the value set in the environment variable
             `AWS_SESSION_TOKEN`.
+        - task_role_arn (str, optional): AWS Amazon Resource Name (ARN) of the AWS
+            Identity and Access Management (IAM) role that grants containers in the
+            task permission to call AWS APIs on your behalf. Defaults to the value set
+             in the environment variable `TASK_ROLE_ARN`.
+        - execution_role_arn (str, required): AWS Amazon Resource Name (ARN) of the
```
joshmeek:
```suggestion
        - execution_role_arn (str, optional): AWS Amazon Resource Name (ARN) of the
```

src/prefect/agent/fargate/agent.py
```diff
@@ -14,6 +14,7 @@ class FargateAgent(Agent):
     Fargate Agent can be found at https://docs.prefect.io/cloud/agent/fargate.html
 
     Args:
+
```
joshmeek:
```suggestion
```

joshmeek: I think this extra space is what is causing the documentation formatting error :)

Issue #1638, opened 2019-10-17
Optimize our Dockerfiles


cicdw:
We have two "Dockerfiles" that we utilize in Prefect:

- one for our own base image
- one for how we build docker images for Flows

We should optimize the size and build speed for both of these to improve UX.

cicdw:
Closed in #1648

PR #1648, opened 2019-10-21
Reduce docker image size

wagoodman:

## What does this PR change?
- Reduces the docker image size from ~900MB to ~490MB by leveraging the `python:*-slim` images (addressing #1638 ).
- Introduces a prefect docker image for each supported python version (3.5, 3.6, and 3.7) which are built on each tagged master commit.

## Why is this PR important?
This cuts down the time it takes to pull down the prefect image for the first time, thus getting users to their first flow faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,11 @@
-version: 2
+version: 2.1
 
 references:
   workspace_root: &workspace_root /tmp/workspace
   attach_workspace: &attach_workspace
     attach_workspace:
       at: *workspace_root
+
 
 jobs:
   # ----------------------------------
@@ -147,7 +148,7 @@ jobs:
   # Run unit tests in Python 3.5-3.7
   # ----------------------------------
 
-  test-3.5:
+  test-35:
     docker:
       - image: python:3.5
 
@@ -176,7 +177,7 @@ jobs:
           paths:
             - coverage
 
-  test-3.6:
+  test-36:
     docker:
       - image: python:3.6
     steps:
@@ -204,7 +205,7 @@ jobs:
           paths:
             - coverage
 
-  test-3.7:
+  test-37:
     docker:
       - image: python:3.7
     steps:
@@ -266,41 +267,74 @@ jobs:
           name: Upload Coverage
           command: bash <(curl -s https://codecov.io/bash) -cF python -s "/tmp/workspace/coverage/"
 
-  build_image:
+  build-docker-image:
     docker:
       - image: docker
+    parameters:
+      python-version:
+        type: string
+      tag-latest:
+        type: boolean
+        default: false
+    environment:
+      PYTHON_VERSION: << parameters.python-version >>
+      PYTHON_TAG: python<< parameters.python-version >>
     steps:
-      - setup_remote_docker
       - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
       - run:
-          name: Docker Build
-          command: docker build -t prefecthq/prefect .
+          name: Build image
+          command: >-
+            docker build 
+            --build-arg GIT_SHA=$CIRCLE_SHA1 
+            --build-arg BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+            --build-arg PREFECT_VERSION=$CIRCLE_TAG 
+            -t prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG}
+            -t prefecthq/prefect:$PYTHON_TAG
+            .
+      - when:
+          condition: << parameters.tag-latest >>
+          steps:
+            - run:
+                name: Tag latest image
+                command: |
+                  docker tag prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG} prefecthq/prefect:latest
       - run:
           name: Test image
           command: |
             docker run -dit prefecthq/prefect /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
       - run:
-          name: Authenticate with Docker Hub and push
+          name: Push versioned tags
           command: |
             docker login --username  $DOCKER_HUB_USER --password $DOCKER_HUB_PW
-            docker push prefecthq/prefect
+            docker push prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG}
+            docker push prefecthq/prefect:$PYTHON_TAG
+      - when:
+          condition: << parameters.tag-latest >>
+          steps:
+            - run:
+                name: Push latest tag
+                command: |
+                  docker login --username  $DOCKER_HUB_USER --password $DOCKER_HUB_PW
+                  docker push prefecthq/prefect:latest
 
 workflows:
   version: 2
 
   "Run tests":
     jobs:
-      - test-3.5
-      - test-3.6
-      - test-3.7
+      - test-35
+      - test-36
+      - test-37
       - test-lower-prefect
       - test-vanilla-prefect
       - test-py352-import-prefect
       - test-airflow
       - upload-coverage:
           requires:
-            - test-3.5
-            - test-3.6
+            - test-35
+            - test-36
             - test-vanilla-prefect
             - test-airflow
 
@@ -312,7 +346,25 @@ workflows:
 
   "Build docker images":
     jobs:
-      - build_image:
+      - build-docker-image:
+          python-version: "3.5"
           filters:
             branches:
               only: master
+            tags:
+              only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+      - build-docker-image:
+          python-version: "3.6"
+          filters:
+            branches:
+              only: master
+            tags:
+              only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+      - build-docker-image:
+          python-version: "3.7"
+          tag-latest: true
+          filters:
+            branches:
+              only: master
+            tags:
+              only: /^[0-9]+\.[0-9]+\.[0-9]+$/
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -266,41 +267,74 @@ jobs:
...
           name: Test image
           command: |
             docker run -dit prefecthq/prefect /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
       - run:
-          name: Authenticate with Docker Hub and push
+          name: Push versioned tags
           command: |
             docker login --username  $DOCKER_HUB_USER --password $DOCKER_HUB_PW
-            docker push prefecthq/prefect
+            docker push prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG}
```
cicdw: I think we might want to finesse the logic here a little, because I suspect this will end up pushing an image with a messy git tag

Dockerfile
```diff
@@ -1,8 +1,20 @@
...
+ARG BUILD_DATE
+
+LABEL maintainer="help@prefect.io"
+LABEL io.prefect.python-version=${PYTHON_VERSION}
+LABEL org.label-schema.schema-version = "1.0"
+LABEL org.label-schema.name="prefect"
+LABEL org.label-schema.url="https://www.prefect.io/"
+LABEL org.label-schema.version=${PREFECT_VERSION}
+LABEL org.label-schema.vcs-ref=${GIT_SHA}
+LABEL org.label-schema.build-date=${BUILD_DATE}
```
cicdw: Nice!

src/prefect/environments/storage/docker.py
```diff
@@ -79,19 +72,50 @@ def __init__(
...
+
         self.env_vars = env_vars or {}
+        self.env_vars["PREFECT__USER_CONFIG_PATH"] = "/root/.prefect/config.toml"
+
         self.files = files or {}
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
         self.base_url = base_url or default_url
         self.local_image = local_image
+        self.extra_commands = []
```
cicdw: This feels like a really convenient abstraction - I'd love let's expose this as a new keyword argument for `Docker` storage.

wagoodman: Maybe, but the ultimate abstraction to this is the `base_image` option where you can run anything that you'd like. In this way we reduce the risk of breaking anyone downstream when we make any overhauls to the base image or reorder where the commands are run.

.circleci/config.yml
wagoodman: I'm open to a better way to do it. I added [circle-ci git tag filters here](https://github.com/PrefectHQ/prefect/pull/1648/files/3aad2e7a686429f40856e074712a1dd4d3fc9d0e#diff-1d37e48f9ceff6d8030570cd36286a61R354-R355) to help reduce the risk of an odd tag coming through.

cicdw: Hmmm yea but won't this job still run on master?

src/prefect/environments/storage/docker.py
cicdw: That's true, but many of users aren't super familiar with Docker - this is also the reason we exposed `files` and `python_dependencies` as keywords, to help optionally customize the image without building a base image yourself.  But the risk of breakage is a good point, happy to leave this as-is.

src/prefect/environments/storage/docker.py
```diff
@@ -79,19 +72,50 @@ def __init__(
...
             self.prefect_version = "master" if len(version) > 1 else version[0]
         else:
             self.prefect_version = prefect_version
 
+        if base_image is None:
+            python_version = "{}.{}".format(
+                sys.version_info.major, sys.version_info.minor
+            )
+            if re.match("^[0-9]+\.[0-9]+\.[0-9]+$", self.prefect_version) != None:
+                # note: this does not necessarily mean that we have built/pushed all previous prefect versions to dockerhub
```
cicdw: luckily a user will only hit this block of code if they're running 0.6.8+ so we don't have to retroactively build old images with this pattern

src/prefect/environments/storage/docker.py
```diff
@@ -79,19 +72,50 @@ def __init__(
         self.image_name = image_name
         self.image_tag = image_tag
         self.python_dependencies = python_dependencies or []
+        self.python_dependencies.append("wheel")
+
         self.env_vars = env_vars or {}
+        self.env_vars["PREFECT__USER_CONFIG_PATH"] = "/root/.prefect/config.toml"
```
cicdw: Just out of curiosity why was this necessary?  I believe this should be the default.

cicdw: Nvm I now see that you just moved this from where it previously was 👍

tests/environments/storage/test_docker_storage.py
```diff
@@ -47,11 +47,13 @@ def test_empty_docker_storage(monkeypatch, platform, url):
...
     assert not storage.registry_url
-    assert storage.base_image.startswith("python:")
+    assert storage.base_image.startswith("prefecthq/prefect:python")
     assert not storage.image_name
     assert not storage.image_tag
-    assert not storage.python_dependencies
-    assert not storage.env_vars
+    assert storage.python_dependencies == ["wheel"]
+    assert storage.env_vars == {
+        "PREFECT__USER_CONFIG_PATH": "/root/.prefect/config.toml"
```
cicdw: @joshmeek do you remember why we had to include this originally? I suspect it's no longer necessary but I don't want to take any chances

joshmeek: I don't think this is used anymore

TylerWanner: don't remove it as part of this PR, but do remove it as part of another PR after this gets merged pls

.circleci/config.yml
wagoodman: good call, I thought these were `and`ed instead of `or`ed. I'll update...

src/prefect/environments/storage/docker.py
wagoodman: I think I'll leave it as is for now, it doesn't prevent us from exposing this in the future.

src/prefect/environments/storage/docker.py
wagoodman: true, but this won't stop them from specifying an older version [here](https://github.com/PrefectHQ/prefect/blob/7576d7ba0b9c813d9af2004a67c74ef7a458a60d/src/prefect/environments/storage/docker.py#L62) (such as 0.6.6) when using 0.6.8+ locally. I don't see why someone would want to do this, but they could I suppose. 

Though, I take your larger point... I'm willing to let the docker pull fail like we would do for any other user-specified `base_image`. I'll take out the code comment.

.circleci/config.yml
```diff
@@ -266,41 +267,74 @@ jobs:
           name: Upload Coverage
           command: bash <(curl -s https://codecov.io/bash) -cF python -s "/tmp/workspace/coverage/"
 
-  build_image:
+  build-docker-image:
```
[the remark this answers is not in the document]
wagoodman: righto, just swapped over all job names and parameters to underscore nomenclature

.circleci/config.yml
wagoodman: Interestingly enough there isn't a good way to do this, @TylerWanner and I came up with this combo:
- https://github.com/PrefectHQ/prefect/pull/1648/files#diff-1d37e48f9ceff6d8030570cd36286a61R361-R364
- https://github.com/PrefectHQ/prefect/pull/1648/files#diff-1d37e48f9ceff6d8030570cd36286a61R286-R292

Up to a better approach, but this is all I've got until circleci supports and'ed filters.

tests/environments/storage/test_docker_storage.py
```diff
@@ -362,9 +387,20 @@ def test_create_dockerfile_from_everything():
             with open(os.path.join(tempdir, "Dockerfile"), "r") as dockerfile:
                 output = dockerfile.read()
 
+            print(output)
+
```
joshmeek:
```suggestion

```

.circleci/config.yml
```diff
@@ -266,43 +267,85 @@ jobs:
...
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Build image
+          command: >-
+            docker build 
+            --build-arg GIT_SHA=$CIRCLE_SHA1 
+            --build-arg BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+            --build-arg PREFECT_VERSION=$CIRCLE_TAG 
+            -t prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG}
```
cicdw: whenever `CIRCLE_TAG` is not populated (e.g., on a normal merge-to-master), will this be a valid image tag?

.circleci/config.yml
cicdw: I think this is the only place I'm still hesitant; here's what I imagine happening:
- a PR is merged to master
- `build_docker_image` is run
- `CIRCLE_TAG` is empty, because this isn't a tagged commit
- either this step or the build step errors out in some unique way based on this

Everything else looks great though! And to be honest, we could always clean things up if that happens but I want to make sure I understand this piece

.circleci/config.yml
wagoodman: no it will not, I believe it would be an empty or missing value. That is the reason for [the filter](https://github.com/PrefectHQ/prefect/pull/1648/files#diff-1d37e48f9ceff6d8030570cd36286a61R364), such that we are only calling this job upon a tagged commit.

However, I could add a `set -u` before this command such that bash will throw an error upon attempting to access a non-existent or empty variable?

.circleci/config.yml
wagoodman: I don't believe that it is possible to reach this step because of the workflow filter, however, I'm all for adding the precautionary `set -u` mentioned in the other comment below.

PR #1649, opened 2019-10-21
Refactor Fargate Agent to Allow for All boto3 kwargs

### LINKED ISSUE #1626: Use better naming conventions for deployed tasks/jobs
## Current behavior
e.g. the Fargate agent creates tasks called `prefect-task` and the FargateTaskEnvironment created tasks called `prefect-task`. These should be better named

## Proposed behavior
Come up with better names for the intermediate tasks/jobs and those which actually run the flow

## Example
N/A

### LINKED ISSUE #1627: Clarify use of ARN roles in Fargate Agent and Fargate Task Environment
## Current behavior
Having to specify the ARNs twice is a bit confusing 

## Proposed behavior
Write documentation to clarify why there are two places to specify them

## Example
N/A

### LINKED ISSUE #1629: Sort out the AWS keys vs roles vs IAM for Fargate deployments
## Current behavior
We need to iron out how/when you should specify access keys vs roles vs creating IAM etc...

## Proposed behavior
Allow for all configurations to be supported.
Idea: Fargate Task Environment could inherit credentials from the Fargate Agent

## Example
N/A

joshmeek:

## What does this PR change?
Closes #1629 
Closes #1627 
Closes #1626 
This PR updates the Fargate Agent to remove all static arguments and instead opt for a general `**kwargs` which will take in all possible arguments for boto3's `run_task` and `register_task_definition`.

The main difference is that previously we had the supported arguments typed out in snake case and they are now adjusted to boto3's camel case. (e.g. `execution_role_arn` becomes `executionRoleArn`). This is a breaking change.

Also updates the `FargateTaskEnvironment` to remove the `prefect-task-UID-custom` name and allow for user specified names.

## Why is this PR important?
Instead of adjusting for everyone's individual AWS use case we can open the floodgates to full customizability by allowing for all possible boto3 configurations.


src/prefect/agent/fargate/agent.py
```diff
@@ -25,111 +32,35 @@ class FargateAgent(Agent):
...
-        - assign_public_ip (str, optional): Whether the task's elastic network interface
-            receives a public IP address. Defaults to the value set in the environment
-            variable `ASSIGN_PUBLIC_IP` or `ENABLED` otherwise.
-        - task_cpu (str, optional): The number of cpu units reserved for the container.
-            Defaults to the value set in the environment variable `TASK_CPU` or `256`
-            otherwise.
-        - task_memory (str, optional): The hard limit (in MiB) of memory to present to
-            the container. Defaults to the value set in the environment variable
-            `TASK_MEMORY` or `512` otherwise.
+        - **kwargs (dict, optional): additional keyword arguments to pass to boto3 for
```
cicdw: You're certain there are no kwarg overlaps here?

joshmeek: There are but they're handled in the `_parse_kwargs` function

src/prefect/agent/fargate/agent.py
```diff
@@ -144,20 +75,70 @@ def __init__(
...
+            "proxyConfiguration",
+            "inferenceAccelerators",
+        ]
+
+        run_kwarg_list = [
+            "cluster",
+            "count",
+            "startedBy",
+            "group",
+            "placementConstraints",
```
cicdw: Are there any edge cases / bugs that could be introduced with `placementConstraints`, since it equally applies to runs and definitions?  E.g., is it possible a user would want to apply some setting here to _only_ the definition or _only_ the run?

PR #1661, opened 2019-10-26
Add section on Flow versions and archiving

cicdw:
I realized we don't have much documentation here other than the self-documenting GraphQL schema, which isn't very discoverable!


docs/cloud/concepts/flows.md
```diff
@@ -36,3 +36,23 @@ mutation($flow: JSON!) {
     serializedFlow: <the serialized flow JSON>
 }
 ```
+
+## Flow Versions and Archiving <Badge text="GQL"/>
+
+Flows with the same name in the same Project are considered "versions" of each other.  Anytime you deploy a flow to a project which contains a flow of the same name, Prefect Cloud will automatically "archive" the old version in place of the newly deployed flow.  Archiving means that the old version's schedule is set to "Paused", no new flow runs can be created and it is not as visible in the UI.  You can always revisit old versions and unarchive them, if for example you want the same Flow to run on two distinct schedules.  To archive or unarchive a flow, use the following GraphQL mutations:
```
joshmeek: I believe you can see older versions in the UI by clicking the `Versions` button :P

cicdw: Updated!

docs/cloud/concepts/flows.md
```diff
@@ -36,3 +36,23 @@ mutation($flow: JSON!) {
     serializedFlow: <the serialized flow JSON>
 }
 ```
+
+## Flow Versions and Archiving <Badge text="GQL"/>
+
+Flows with the same name in the same Project are considered "versions" of each other.  Anytime you deploy a flow to a project which contains a flow of the same name, Prefect Cloud will automatically "archive" the old version in place of the newly deployed flow.  Archiving means that the old version's schedule is set to "Paused" and no new flow runs can be created.  You can always revisit old versions and unarchive them, if for example you want the same Flow to run on two distinct schedules.  To archive or unarchive a flow, use the following GraphQL mutations:
```
jlowin: It might be worth a small note that this versioning behavior can be customized, and this is just the common sense default.

cicdw: I think I need to expose this in the Python Client before I document it here; while the functionality does exist, someone would have to build their flow, serialize it, and then independently of `flow.deploy` craft the correct GraphQL mutation.  I'll open an issue for exposing this.

cicdw: #1665

jlowin: Fair point

PR #1664, opened 2019-10-27
New secrets API

### LINKED ISSUE #1346: Consider a minor refactor of the Secrets API to be more pluggable
We could introduce a Secret base class that Prefect inherits from for it's secrets, and provide documentation about how to subclass for implementing other sensitive storage options.

As discussed [here](https://github.com/PrefectHQ/prefect/pull/1343/files#r312719069)

### LINKED ISSUE #1587: Have Tasks and/or Flows specify the secrets they require at init
## Current behavior
Currently, the Prefect Secrets API is very explicit:
```python
@task
def my_task():
    api_secret = Secret("api_secret").get()
```
this means that, for example, if I map this task over some large collection of items, I'm going to pull this Secret value for every mapped instance.  This is clearly redundant.  

## Proposed behavior
Instead, we could:
```python
@task(secrets=["api_secret"])
def my_task():
    api_secret = prefect.context['secrets']['api_secret']
```
which would trigger the TaskRunner to auto-populate context with the secret values.  We could also specify this instead at the Flow level, but my bias is to keep this is as "local" as possible (open to push-back here of course).  A third option would be to specify it on the task, but have the FlowRunner call the Secrets API (instead of the TaskRunner).  This way, if the same secret is specified across multiple tasks, the FlowRunner can track that and not make redundant calls.  We'd want to think about how to raise the error if a Secret doesn't exist (this implementation would result in a "global" Flow failure instead of a local task failure).

Either way, this new pattern would allow us to pull the secret _one time_ (prior to mapping or prior to task runner execution) and provide it within context to every mapped child.

Note that open-source users would see absolutely 0 difference, as `Secret('key').get()` is a wrapper to `prefect.context['secrets']['key']` anyway.

**cc:** @TylerWanner @zdhughes @joshmeek

cicdw:

## What does this PR change?
This PR introduce a new collection of Secret Tasks along with Secret Result Handlers.  The basic idea is that the run method of this task will retrieve a potentially sensitive piece of data and return it - to prevent this data from being accidentally stored incorrectly, a new Secret Result Handler is introduced which _always_ re-retrieves the underlying secret using the `run` method of the Secret Task.

## Why is this PR important?
This allows a way of not only clearly specifying the secrets a Task / Flow requires via a direct task dependency, but also makes for a pluggable Secrets API.  All users need to do is subclass the base `Secret` Task and their data will be handled securely.

Closes #1346  
Closes #1587


CHANGELOG.md
```diff
@@ -14,10 +14,11 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 - Adds an "Ask for help" button with a link to the prefect.io support page - [#1637](https://github.com/PrefectHQ/prefect/pull/1637)
 - Reduces the size of the `prefecthq/prefect` Docker image by ~400MB, which is now the base Docker image used in Flows - [#1648](https://github.com/PrefectHQ/prefect/pull/1648)
 - Add a new healthcheck for environment dependencies - [#1653](https://github.com/PrefectHQ/prefect/pull/1653)
+- Add new Secret Tasks for a pluggable and reusable Secrets API - [#1346](https://github.com/PrefectHQ/prefect/issues/1346), [#1587](https://github.com/PrefectHQ/prefect/issues/1587)
```
jlowin: I would promote this to a Feature

cicdw:
@jlowin I added two tests [here](https://github.com/PrefectHQ/prefect/pull/1664/commits/2ffc7d894119e6aa95545d99dfcedd24b7e38ee7); let me know if there is another angle you were thinking of

src/prefect/tasks/secrets/base.py
```diff
@@ -0,0 +1,29 @@
...
+    Args:
+        - name (str): The name of the underlying secret
+        - **kwargs (Any, optional): additional keyword arguments to pass to the Task constructor
+    """
+
+    def __init__(self, name, **kwargs):
+        kwargs["name"] = name
+        kwargs.setdefault("max_retries", 2)
+        kwargs.setdefault("retry_delay", datetime.timedelta(seconds=1))
+        kwargs["result_handler"] = SecretResultHandler(secret_task=self)
```
jlowin: I think if this isn't actually customizable we should raise an error if `result_handler` is provided

tests/engine/cloud/test_cloud_flow_runner.py
```diff
@@ -381,6 +385,49 @@ def is_p_three(p):
...
+    assert not state.result[res].cached_inputs["p"] == exp_res
+    exp_res.store_safe_value()
+    assert state.result[res].cached_inputs["p"] == exp_res
+
+    ## here we set the result of the secret to a saferesult, ensuring
+    ## it will get converted to a "true" result
+    ## moreover, we make the safe result value 3 and we expect
+    ## that the upstream value will actually get recomputed from context
+    ## through the SecretResultHandler
+    safe = SafeResult(3, result_handler=SecretResultHandler(p))
```
jlowin: In reality, wouldn't the SafeResult be `p` since that's the secret name, rather than the secret value?

src/prefect/tasks/secrets/base.py
cicdw: updated

tests/engine/cloud/test_cloud_flow_runner.py
cicdw: Yea, that is true - I was just trying to stack the cards in favor of the value "3" just in case -- I can change if you'd prefer

jlowin: I hate to nitpick on this but since I view this test as "enshrining" the expected high level API, i think it's important to be clear that the value isn't part of the expected payload.

cicdw: No worries at all - so would you like me to mention this in the comment or just convert it to the safe result we'd expect in practice?

jlowin: My preference would be change it to `SafeResult('p'...)`

cicdw: 👍 updated

PR #1685, opened 2019-10-30
Flow save / load

cicdw:

## What does this PR change?
This PR introduces a `save` / `load` interface to Flows for a smoother deployment experience.

## Why is this PR important?
Some users were annoyed with `cloudpickle`'s behavior if you create a Flow in a file called "my_flow.py" and then imported that flow inside "my_deploy.py" like `from my_flow import flow`, deploy health checks would fail.  This is because `cloudpickle` sees your flow's logic as living inside a module called "my_flow".  

Using the new `save / load` interface allows users to save their flow at build time, so that cloudpickle correctly manages the imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Enhancements
 
-- None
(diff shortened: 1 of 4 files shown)


aroder:
Yeah! This is helpful. I find that creating a "build" folder at the top level of my project and excluding it from source control is a convenient place to put the pickled flow.

PR #1694, opened 2019-11-01
Background logging

cicdw:

## What does this PR change?
This PR introduces a background queuing / thread setup so that Flow logs are batch uploading during Cloud runs instead of making an authenticated request with each log.

## Why is this PR important?
Reduces network traffic and should result in some minor performance improvements when running flows in Cloud.


src/prefect/utilities/logging.py
```diff
@@ -30,6 +33,57 @@ def __init__(self) -> None:
...
+        if self.client is not None:
+            self.batch_upload()
+            self._thread.join()
+
+    def batch_upload(self) -> None:
+        logs = []
+        try:
+            while True:
+                log = self.queue.get(False)
+                logs.append(log)
```
jlowin: There's some (small) risk that the arrival rate of logs in `self.queue` is > the processing rate of this function, in which case it might never leave this `while True` loop. Perhaps a `max_iter` counter just in case?

src/prefect/utilities/logging.py
```diff
@@ -44,26 +98,22 @@ def emit(self, record) -> None:  # type: ignore
...
-            self.client.write_run_log(
-                flow_run_id=flow_run_id,
-                task_run_id=task_run_id,
-                timestamp=timestamp,
-                name=name,
-                message=message,
-                level=level,
-                info=record_dict,
-            )
+            log["info"] = record_dict
```
jlowin: As long as we're looking at this, this `record_dict` is the ultimate source of duplication when looking at Cloud logs in raw form. For example, it contains duplicates of all `Id`s and the `message` even though those have also been broken out into the formal `log` object. Should we `pop` them instead of `get`-ting them?

src/prefect/config.toml
```diff
@@ -38,6 +38,8 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 # Send logs to Prefect Cloud
 log_to_cloud = false
```
jlowin: I think maybe both `log_to_cloud` and `heartbeat` should be in a subsection of `logging` called `cloud`, as they only apply in a cloud context.

cicdw: Discussed offline, I'll move `heartbeat` to `logging_heartbeat` inside the `[cloud]` section

cicdw:
@jlowin updated

PR #1696, opened 2019-11-03
Local storage labels

cicdw:

## What does this PR change?
This PR auto-labels flow environments with a slugified flow name whenever `Local` storage is used.

## Why is this PR important?
Using local storage greatly restricts what types of agents can run the work - this PR pre-emptively helps with that.

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1250,6 +1250,7 @@ class Flow:
             self.storage = get_default_storage_class()(**kwargs)
 
         if isinstance(self.storage, prefect.environments.storage.Local):
+            self.environment.labels.add("local")
             self.environment.labels.add(slugify(self.name))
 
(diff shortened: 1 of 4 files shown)


src/prefect/core/flow.py
```diff
@@ -1249,6 +1249,10 @@ def deploy(
         if self.storage is None:
             self.storage = get_default_storage_class()(**kwargs)
 
+        if isinstance(self.storage, prefect.environments.storage.Local):
+            self.environment.labels.add("local")
+            self.environment.labels.add(slugify(self.name))
```
joshmeek: Can you add a note of these labels being automatically added? Maybe to the docstrings for the Local storage

cicdw: Good call

cicdw: Updated

PR #1729, opened 2019-11-12
Proactively fail flow runs if there is an issue with flows storage

### LINKED ISSUE #1395: When a flow has no storage attached, the agent should mark the run as failed
It appears that if the Agent picks up a run with no storage, it will mark it as submitted. This run will then exist in limbo forever, as it can never be run. The agent should put it in a failed state if it can recognize it'll never run.

cicdw:

## What does this PR change?
This PR proactively fails flow runs from the local docker agent if there are issues with the Flow's storage, including an image pull error.

## Why is this PR important?
Previously we simply logged the error and waited for Lazarus to come along and reschedule the flow, which typically just results in the flow run eventually failing after some time.  Users expect an image pull error to immediately translate to a flow run failure.  Closes #1395

### DIFF
--- a/src/prefect/agent/local/agent.py
+++ b/src/prefect/agent/local/agent.py
@@ -5,6 +5,7 @@ import docker
 
 from prefect import config, context
 from prefect.agent import Agent
+from prefect.engine.state import Failed
 from prefect.environments.storage import Docker
 from prefect.serialization.storage import StorageSchema
 from prefect.utilities.graphql import GraphQLResult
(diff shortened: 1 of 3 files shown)


src/prefect/agent/local/agent.py
```diff
@@ -78,9 +79,13 @@ def deploy_flows(self, flow_runs: list) -> None:
 
             storage = StorageSchema().load(flow_run.flow.storage)
             if not isinstance(StorageSchema().load(flow_run.flow.storage), Docker):
-                self.logger.error(
-                    "Storage for flow run {} is not of type Docker.".format(flow_run.id)
+                msg = "Storage for flow run {} is not of type Docker.".format(
```
joshmeek: Could you add something noting that this log is coming from an Agent

src/prefect/agent/local/agent.py
```diff
@@ -97,7 +102,11 @@ def deploy_flows(self, flow_runs: list) -> None:
                         "Successfully pulled image {}...".format(storage.name)
                     )
                 except docker.errors.APIError as exc:
-                    self.logger.error("Issue pulling image {}".format(storage.name))
+                    msg = "Issue pulling image {}".format(storage.name)
```
joshmeek: Same here

src/prefect/agent/local/agent.py
cicdw: When you say "log" do you mean the state message?

joshmeek: Ah yep that is in fact what I meant

cicdw: Updated

src/prefect/agent/local/agent.py
cicdw: updated

PR #1730, opened 2019-11-12
No constants

cicdw:

## What does this PR change?
This PR completely avoids the auto-generation and execution of Python constants which are upstream of a given task.  Instead, these constants are tracked on the Flow object directly, and artificial edges are created at runtime to manage them as upstream dependencies.

## Why is this PR important?
The autogeneration of constants is perhaps one of the more confusing aspects of Prefect's functional API for newcomers.  In addition, Cloud users do not find `Constant[str]` type tasks to be useful or informative, and ultimately these just eat up real estate on all screens.  The current PR completely avoids this situation unless a user explicitly opts into it via wrapping their objects in `prefect.tasks.core.constants.Constant()`

Labeled as WIP as I still need to:
- ~add more tests~
- ~update documentation~
- ~handle Cloud retries / failures~

but I wanted some initial feedback on the implementation here.

### DIFF
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -16,6 +16,8 @@ import pendulum
 import prefect
 from prefect.core import Edge, Flow, Task
 from prefect.engine import signals
+from prefect.engine.result import Result
+from prefect.engine.result_handlers import ResultHandler
 from prefect.engine.runner import ENDRUN, Runner, call_state_handlers
 from prefect.engine.state import (
     Failed,
@@ -415,6 +417,18 @@ class FlowRunner(Runner):
                 for edge in self.flow.edges_to(task):
                     upstream_states[edge] = task_states.get(
                         edge.upstream_task, Pending(message="Task state not available.")
+                    )
+
+                # augment edges with upstream constants
+                for key, val in self.flow.constants[task].items():
+                    edge = Edge(
+                        upstream_task=prefect.tasks.core.constants.Constant(val),
+                        downstream_task=task,
+                        key=key,
+                    )
+                    upstream_states[edge] = Success(
+                        "Auto-generated succcess state",
+                        result=Result(val, result_handler=ResultHandler()),
                     )
 
                 # -- run the task
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -17,7 +17,7 @@ import prefect
 from prefect.core import Edge, Flow, Task
 from prefect.engine import signals
 from prefect.engine.result import Result
-from prefect.engine.result_handlers import ResultHandler
+from prefect.engine.result_handlers import ConstantResultHandler
 from prefect.engine.runner import ENDRUN, Runner, call_state_handlers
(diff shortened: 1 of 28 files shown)


joshmeek:
What's the _key_ noticeable difference that one would see when this is merged? Am I thinking about it right in that we won't see all the `Constant` task run logs anymore unless we declare `Constant`s in our flow?

cicdw:
@joshmeek yup that's one major difference; another is that
```python
with Flow("blah") as flow:
    add(2, 4)
```
will result in a Flow with a single task.  This will be very apparent in flow.visualize() and in the UI views.

cicdw:
@jlowin / @joshmeek this is ready for review: note that I am choosing to _not_ implement a serializer for `ConstantResultHandler` at this time because the planned cadence of our releases could cause issues for people who want to start working off 0.7.2 once it's released later this week.

src/prefect/engine/result_handlers/constant_result_handler.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any
+
+from prefect.engine.result_handlers import ResultHandler
+
+
+class ConstantResultHandler(ResultHandler):
+    """
+    Hook for storing and retrieving constant Python objects. Only intended to be used
+    internally.
```
joshmeek: What do you mean by _internally_?

cicdw: Only the "Prefect engine" should ever deal with this result handler, I don't recommend any users ever use it

src/prefect/engine/flow_runner.py
```diff
@@ -417,6 +419,18 @@ def get_flow_run_state(
...
 
+                # augment edges with upstream constants
+                for key, val in self.flow.constants[task].items():
+                    edge = Edge(
+                        upstream_task=prefect.tasks.core.constants.Constant(val),
+                        downstream_task=task,
+                        key=key,
+                    )
+                    upstream_states[edge] = Success(
+                        "Auto-generated succcess state",
```
jlowin:
```suggestion
                        "Auto-generated constant value",
```

src/prefect/utilities/tasks.py
```diff
@@ -45,13 +45,18 @@ def add(x, y):
...
+def as_task(
+    x: Any, flow: Optional["Flow"] = None, convert_constants: bool = True
+) -> "prefect.Task":
     """
     Wraps a function, collection, or constant with the appropriate Task type.
 
     Args:
         - x (object): any Python object to convert to a prefect Task
         - flow (Flow, optional): Flow to which the prefect Task will be bound
+        - convert_constants (bool, optional): a boolean specifying whether all passed Python
```
jlowin: Your docstring points out this is slightly misnamed, since some "constants" (lists of constants) are never affected by this

cicdw: Yea this is fair; while we used to call "lists" constants, I think our previous terminology was the misnomer as those are more "collections of constants".  However, totally open to changing this if we can come up with a clearer name.

PR #1736, opened 2019-11-13
Respect order of inputs in merge

bnaul:

## What does this PR change?
Currently `merge(a, b)` randomly returns either `a` or `b` if both tasks have results. However, the docs say:

> The merge will return the first real result it encounters, or None.

which to me implies that the result in that case should be `a`. The randomness comes from the fact that task results come from edges, which are stored as a set, which have non-deterministic ordering.

## Why is this PR important?
@cicdw mentioned this is maybe not the original intended use of `merge`, but to me it feels valid and the docs definitely suggest that this should work. Also less randomness in results when re-running some flows seems desirable.

### DIFF
--- a/tests/tasks/test_control_flow.py
+++ b/tests/tasks/test_control_flow.py
@@ -224,9 +224,9 @@ def test_merge_with_list():
 
 def test_merge_order():
     with Flow(name="test") as flow:
-        x = 'x'
-        y = 'y'
-        merge_task = merge('x', 'y')
+        x = "x"
+        y = "y"
+        merge_task = merge("x", "y")
 
     state = flow.run()
     assert state.result[merge_task].result == "x"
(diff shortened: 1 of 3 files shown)


tests/tasks/test_control_flow.py
```diff
@@ -222,6 +222,16 @@ def true_branch():
         assert state.result[merge_task].result == [1, 2]
 
 
+def test_merge_order():
+    with Flow(name="test") as flow:
+        x = "x"
+        y = "y"
+        merge_task = merge("x", "y")
```
cicdw: this is perfectly valid, but given the changes to how prefect manages constants soon it might be more robust to implement this with two full `@task` objects which return `"x"` and `"y"`

cicdw: just to make sure the test continues testing the correct thing

bnaul: makes sense (also that version was a mistake anyway)

PR #1750, opened 2019-11-18
Worker client fix

cicdw:

## What does this PR change?
This PR refactors how we manage background heartbeats from task / flow runs.

## Why is this PR important?
It turns out that the old implementation of heartbeating led to non-deterministic deadlocks when using dask `worker_client`s.  Ultimately this boiled down to some complicated interaction between dask / requests / our Timer implementation.  These deadlock situations would ultimately show themselves as Zombie Tasks in Prefect Cloud.  The new implementation uses a `ThreadPoolExecutor` to manage the thread and its associated state, which empirically has resolved the issue!


src/prefect/utilities/executors.py
```diff
@@ -19,12 +20,29 @@
 StateList = Union["State", List["State"]]
 
 
-class Heartbeat(threading.Timer):
-    def run(self) -> None:
-        self.finished.wait(self.interval)  # type: ignore
-        while not self.finished.is_set():  # type: ignore
-            self.function(*self.args, **self.kwargs)  # type: ignore
-            self.finished.wait(self.interval)  # type: ignore
+class Heartbeat:
```
joshmeek: Could you add a docstring or two to this class?

cicdw: @joshmeek UPDATED

PR #1752, opened 2019-11-18
Update Environment `run_flow` call to use Flow's original Environment

joshmeek:

## What does this PR change?
Updated the commands for a few environments which were calling `run_flow` on a new Environment object

## Why is this PR important?
If a new Environment object is created on each run then some of the information stored on the environment won't persist. This is mainly an issue in specifying things like `min_workers` and `max_workers` on the `DaskKubernetesEnvironment`.

### DIFF
--- a/docs/cloud/execution/fargate_task_environment.md
+++ b/docs/cloud/execution/fargate_task_environment.md
@@ -95,7 +95,7 @@ The container dictionary above will be changed during setup:
 [
     "/bin/sh",
     "-c",
-    "python -c 'import prefect; from prefect.environments.storage import Docker; Docker().get_flow(prefect.context.flow_file_path).environment.run_flow()'",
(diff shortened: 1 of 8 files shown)


docs/cloud/execution/fargate_task_environment.md
```diff
@@ -95,7 +95,7 @@ The container dictionary above will be changed during setup:
 [
     "/bin/sh",
     "-c",
-    "python -c 'from prefect.environments import FargateTaskEnvironment; FargateTaskEnvironment().run_flow()'",
+    "python -c 'import prefect; from prefect.environments.storage import Docker; Docker().get_flow(prefect.context.flow_file_path).environment.run_flow()'",
```
cicdw: So I get what you're doing here, but I think for readability / generality you should use:
```suggestion
    "python -c 'import prefect; from prefect import Flow; Flow.load(prefect.context.flow_file_path).environment.run_flow()'",
```
and same for the other commands like this below

PR #1753, opened 2019-11-18
S3 Storage

joshmeek:

## What does this PR change?
Adds a new storage option for AWS S3

## Why is this PR important?
Blob storage in S3 buckets is a common pattern and the ability to store Flows there allows for better diversity of platforms for Prefect. This is also nice from the standpoint of some new low-dependency deployment methods 👀


src/prefect/environments/storage/s3.py
```diff
@@ -0,0 +1,154 @@
...
+            - ValueError: if a flow with the same name is already contained in this storage
+        """
+        if flow.name in self:
+            raise ValueError(
+                'Name conflict: Flow with the name "{}" is already present in this storage.'.format(
+                    flow.name
+                )
+            )
+
+        # Pickle Flow
```
cicdw: So, purely following our docker pattern, we might want to consider actually performing the upload at `build` time, although I see how that conflicts with how we're storing `self.flows` here - thoughts?

src/prefect/environments/storage/s3.py
```diff
@@ -0,0 +1,154 @@
...
+    `slugified-flow-name/slugified-current-timestamp`.
+
+    Args:
+        - aws_access_key_id (str, optional): AWS access key id for connecting to S3.
+            Defaults to the value set in the environment variable
+            `AWS_ACCESS_KEY_ID` or `None`
+        - aws_secret_access_key (str, optional): AWS secret access key for connecting to S3.
+            Defaults to the value set in the environment variable
+            `AWS_SECRET_ACCESS_KEY` or `None`
+        - aws_session_token (str, optional): AWS session key for connecting to S3
```
cicdw: As we discussed offline, I'm still on the fence about the best auth pattern here.  When building locally, anything can go, but whenever this Storage needs to extract a flow object while running in Cloud, it's much tougher to specify things like env vars in that environment

joshmeek: Yeah I need to think about this some more 🤔

src/prefect/environments/storage/s3.py
joshmeek: Yeah we should definitely move it to `build` time, good call

src/prefect/environments/storage/s3.py
```diff
@@ -0,0 +1,154 @@
...
+    def __init__(
+        self,
+        aws_access_key_id: str = None,
+        aws_secret_access_key: str = None,
+        aws_session_token: str = None,
+        bucket: str = None,
+        key: str = None,
+    ) -> None:
+        self.flows = dict()  # type: Dict[str, str]
+        self.bucket = bucket or ""
```
cicdw: Why do you use an empty string as a default value here?

src/prefect/environments/storage/s3.py
cicdw: Bump on the above comment @joshmeek

src/prefect/environments/storage/s3.py
```diff
@@ -0,0 +1,154 @@
...
+            - Storage: a Storage object that contains information about how and where
+                each flow is stored
+        """
+        return self
+
+    @property
+    def _boto3_client(self):  # type: ignore
+        from boto3 import client as boto3_client
+
+        return boto3_client(
```
cicdw: What happens if all of these are `None`? Does it naturally default to some environment variables and / or some credentials file?

src/prefect/serialization/storage.py
```diff
@@ -61,6 +61,25 @@ def create_object(self, data: dict, **kwargs: Any) -> Docker:
...
 
 
+class S3Schema(ObjectSchema):
+    class Meta:
+        object_class = S3
+
+    aws_access_key_id = fields.String(allow_none=True)
+    aws_secret_access_key = fields.String(allow_none=True)
+    aws_session_token = fields.String(allow_none=True)
+    bucket = fields.String(allow_none=True)
```
cicdw: Won't the user need to specify a bucket name for all of this to work?

src/prefect/environments/storage/s3.py
joshmeek: Had talked about this with @wagoodman. Going to make the bucket a required argument because this won't work without it

src/prefect/environments/storage/s3.py
joshmeek: Yep! boto3 defaults to the values found in the environment https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables

src/prefect/serialization/storage.py
joshmeek: Yeah, see comment above

joshmeek:
@cicdw How do you propose I handle the test:
```
FAILED serialization/test_storage.py::test_serialization_on_all_subclasses[S3] - TypeError: __init__() missing 1 required positional argument: 'bucket'
```
It's a test parameterized by each type of storage but fails on S3 because of the required bucket kwarg

cicdw:
@joshmeek good question; that test is there to prevent us from forgetting to add a serializer for each of our storage subclasses.  Maybe we could replace it with a test like:

```python
def test_all_storage_subclasses_have_schemas():
    "Test that ensures we don't forget to include a Schema for every subclass we implement"

    subclasses = {c.__name__ for c in storage.Storage.__subclasses__()} # a set
    schemas = set(StorageSchema().type_schemas.keys())
    assert subclasses == schemas
```

Something else this + the heartbeat test have made me realize is that on the margin we should try to start adding doc strings to some of our tests about what they're testing and why.

tests/serialization/test_storage.py
```diff
@@ -19,7 +19,7 @@ def test_all_storage_subclasses_have_schemas():
     "Test that ensures we don't forget to include a Schema for every subclass we implement"
 
     subclasses = set(c.__name__ for c in storage.Storage.__subclasses__())
-    subclasses.update(storage.Storage)  # add base storage, not a subclass
+    subclasses.update([storage.Storage.__name__])  # add base storage, not a subclass
```
cicdw: Pure FYI: sets have an `add` method for adding single elements

src/prefect/environments/storage/s3.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Raises:
+            - ValueError: if the flow is not contained in this storage
+        """
+        if not flow_location in self.flows.values():
+            raise ValueError("Flow is not contained in this Storage")
+
+        stream = io.BytesIO()
+
+        logging.info("Downloading {} from {}".format(flow_location, self.bucket))
```
cicdw: Should we elevate this (and the other logs) to a prefect logger stored on the class?  Not a request, just curious if you think that would help with debugging.

tests/environments/storage/test_s3_storage.py
```diff
@@ -0,0 +1,265 @@
...
+    assert storage
+    assert storage.flows == dict()
+    assert storage.aws_access_key_id == "id"
+    assert storage.aws_secret_access_key == "secret"
+    assert storage.aws_session_token == "session"
+    assert storage.bucket == "bucket"
+    assert storage.key == "key"
+
+
+def test_serialize_docker_storage():
```
cicdw:
```suggestion
def test_serialize_s3_storage():
```

cicdw:
Also, looks like you're getting bit by Python 3.5 in one of your tests; you appear to implicitly assume that the order in which you add the flows to your storage object is also the order in which they will be built, which is not true in Python 3.5 :/

src/prefect/environments/storage/s3.py
joshmeek: Yeah I think we should attach a prefect logger to the base storage class. Similar to how execution environments have it

PR #1755, opened 2019-11-19
Serialize result handlers on flows

zdhughes:

## What does this PR change?

This PR serializes result handlers for Flows. 

## Why is this PR important?

This PR is important because it will allow us to install certain safeguards on the Cloud API side based on the presence or absence of result handlers.


src/prefect/core/flow.py
```diff
@@ -163,7 +166,7 @@ def __init__(
         self.result_handler = (
             result_handler or prefect.engine.get_default_result_handler_class()()
         )
-
+        self.has_result_handler = has_result_handler or bool(result_handler)
```
cicdw: So I totally see where you're going here; however, this does introduce a race condition wherein a user could claim their Flow has a result handler when in fact it does not.

src/prefect/serialization/flow.py
```diff
@@ -51,6 +51,7 @@ class Meta:
     )
     environment = fields.Nested(EnvironmentSchema, allow_none=True)
     storage = fields.Nested(StorageSchema, allow_none=True)
+    has_result_handler = fields.Boolean(allow_none=True)
```
cicdw: I think a better route might be to serialize the result handler directly, using the [`ResultHandlerSchema`](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/serialization/result_handlers.py#L91), and `allow_none=True`.  That way we can test for not-None in Cloud

zdhughes:
Thanks @cicdw! Was concerned about making sure our serialization was anonymous, but you rightly pointed out that that's already accounted for. This is way simpler.

zdhughes:
@cicdw Sure thing! I think the tests I added address your concerns, but happy to make any additional changes if necessary.

tests/serialization/test_flows.py
```diff
@@ -134,6 +135,43 @@ def test_serialize_container_environment():
     assert deserialized.storage.registry_url == storage.registry_url
 
 
+def test_serialize_result_handler():
+    serialized = FlowSchema().dump(Flow(name="n", result_handler=JSONResultHandler()))
+    assert serialized["result_handler"]["type"] == "JSONResultHandler"
+    deserialized = FlowSchema().load(serialized)
+    assert deserialized.result_handler
```
jlowin: These asserts don't demonstrate that the deserialization worked and restored the specified result handler

zdhughes:
Managed to mess up my git fu while trying to rebase. Will reopen once this branch's issues are resolved.

PR #1758, opened 2019-11-20
Cancel state

cicdw:

## What does this PR change?
This PR:
- renames `Aborted` state `Cancelled`
- introduces graceful translation of `KeyboardInterrupt`s to `Cancelled` states when running Flows in Core
- introduces heartbeat check in Cloud to look for cancelled flow runs and respond via cancelled states / termination of execution

## Why is this PR important?
Core users will immediately see the benefit of CTRL+C when running `flow.run`, and this will soon enable the ability to cancel flow runs from the UI once the new state serializers have made their way to Prod.


cicdw:
Great question @jlowin - because we will only support cancellation of at the Flow Run level, downstream tasks will never be able to enter a Running state.  This is because once the Flow Run enters a `Cancelled` state, _all_ heartbeats will fail / raise errors in the main thread.   Moreover, once we implement version locking with Flow Run versions as well, downstream tasks won't even be able to change their state as their flow run state version will be out-of-date.

However, there is the potential for some inefficiency when running with Dask - whenever the Flow Run process raises the `KeyboardInterrupt`, Dask will capture that and "cancel" all Futures associated with that client.  I found that in practice all those futures still ran, meaning the very initial task runner pipeline steps will still run for all tasks, but they will fail to actually change the task's state.

jlowin:
Ah, thanks for clearing that up for me.

PR #1770, opened 2019-11-22
Better heartbeat thread management

cicdw:

## What does this PR change?
This PR revisits the logging of heartbeat thread failures and improves our ability to detect exceptions raised in the heartbeat thread.  We `try / except` inside the `_heartbeat` method but there is still the possibility that the thread dies or ceases running, and this PR ensures that this event will get logged.

## Why is this PR important?
Improved debugging power - if a task run is marked as zombie we will have better visibility into whether the thread died or there was a deadlock.

### DIFF
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -1,5 +1,6 @@
 import datetime
 import signal
+import sys
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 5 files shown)


src/prefect/utilities/executors.py
```diff
@@ -57,17 +60,42 @@ def looper() -> None:
                 iters = (iters + 1) % self.interval
                 time.sleep(self.rate)
 
-        self.executor = ThreadPoolExecutor(max_workers=2)
+        def monitor() -> None:
+            while not self._exit:
+                if not self.fut.running():
```
joshmeek: Is there a possibility where this can be called before `self.fut` it initialized?

cicdw: Good question - there is not, because I only submit the `monitor` function for execution _after_ I've created `self.fut` (see the lines below)

joshmeek: Yeah I saw, just wanted to make sure 👍

PR #1779, opened 2019-11-26
Try to make fixtures a little more resilient to prevent nondeterminis…

cicdw:
…tic job failures

We keep seeing some appveyor builds failing, presumably because of Dask worker cleanup not being successful.


cicdw:
Closing as we are working on this in other branches

PR #1799, opened 2019-12-05
Fix cloud task runner heartbeat test

joshmeek:
Minor tweak for non-deterministic test failure


cicdw:
The test failed again - I've managed to recreate the issue on my local machine by repeating the tests 25 times for each executor.  I'll keep digging!

cicdw:
@joshmeek I updated this branch to fix the test ;) 

Ultimately, this test was intended to test that heartbeats were only sent to Cloud during the actual run of the Task.  However, no Cloud logic would break if this were not the case and we have multiple other tests which test that heartbeats actually work during the `CloudTaskRunner` pipeline.

PR #1809, opened 2019-12-09
Add Google Cloud Storage backend

wagoodman:

## What does this PR change?
Adds a new storage option for Google Cloud Storage.

## Why is this PR important?
This allows the user to delegate flow storage to a remote backend.

### DIFF
--- a/src/prefect/environments/storage/gcs.py
+++ b/src/prefect/environments/storage/gcs.py
@@ -63,7 +63,7 @@ class GCS(Storage):
 
         bucket = self._gcs_client.get_bucket(self.bucket)
 
-        # self.logger.info("Downloading {} from {}".format(flow_location, self.bucket))
+        self.logger.info("Downloading {} from {}".format(flow_location, self.bucket))
 
         blob = bucket.get_blob(flow_location)
         if not blob:
@@ -128,9 +128,9 @@ class GCS(Storage):
 
             blob = bucket.blob(blob_name=self.flows[flow_name])
 
-            # self.logger.info(
-            #     "Uploading {} to {}".format(self.flows[flow_name], self.bucket)
-            # )
+            self.logger.info(
+                "Uploading {} to {}".format(self.flows[flow_name], self.bucket)
+            )
 
             blob.upload_from_string(content)
 
--- a/src/prefect/environments/storage/gcs.py
+++ b/src/prefect/environments/storage/gcs.py
@@ -63,7 +63,9 @@ class GCS(Storage):
         blob = bucket.get_blob(flow_location)
         if not blob:
             raise StorageError(
-                f"Flow not found in bucket: flow={flow_location} bucket={self.bucket}"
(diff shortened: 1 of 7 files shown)


src/prefect/environments/storage/gcs.py
```diff
@@ -0,0 +1,143 @@
...
+        - bucket (str, optional): the name of the GCS Bucket to store the Flow
+        - key (str, optional): a unique key to use for uploading this Flow to GCS. This
+            is only useful when storing a single Flow using this storage object.
+        - project (str, optional): the google project where any GCS API requests are billed to
+    """
+
+    def __init__(
+        self, bucket: str = None, key: str = None, project: str = None,
+    ) -> None:
+        if not bucket:
```
joshmeek: Why raise the error over making `bucket` are required kwarg?

src/prefect/environments/storage/gcs.py
```diff
@@ -0,0 +1,143 @@
...
+
+            blob.upload_from_string(content)
+
+        return self
+
+    @property
+    def _gcs_client(self):  # type: ignore
+        from google.cloud import storage
+
+        return storage.Client(project=self.project)
```
joshmeek: How are GCP creds handled with this? Does it require the user to log in outside of running this, creds in a known location, etc.?

wagoodman: indeed, this requires the user to already have authenticated with the `gcloud` cli and make the credentials path available via the environment variable

src/prefect/environments/storage/gcs.py
```diff
@@ -0,0 +1,138 @@
...
+        Returns:
+            - Storage: an GCS object that contains information about how and where
+                each flow is stored
+        """
+        for flow_name, flow in self._flows.items():
+            content = cloudpickle.dumps(flow)
+
+            bucket = self._gcs_client.get_bucket(self.bucket)
+
+            blob = bucket.blob(blob_name=self.flows[flow_name])
```
joshmeek: Do you think it should handle possible `google.cloud.exceptions.GoogleCloudError` here by logging an error and then raising it?

joshmeek: Oh actually not entirely sure it will encounter an error like that https://googleapis.dev/python/storage/latest/blobs.html

src/prefect/environments/storage/gcs.py
```diff
@@ -0,0 +1,138 @@
...
+            raise ValueError("Flow is not contained in this Storage")
+
+        bucket = self._gcs_client.get_bucket(self.bucket)
+
+        self.logger.info("Downloading {} from {}".format(flow_location, self.bucket))
+
+        blob = bucket.get_blob(flow_location)
+        if not blob:
+            raise StorageError(
+                f"Flow not found in bucket: flow={flow_location} bucket={self.bucket}"
```
joshmeek: Can't use f-strings for lower version of python :(

PR #1811, opened 2019-12-09
Handle possible ClientErrors in S3 storage

joshmeek:

## What does this PR change?
Handles client errors that may arise when uploading and downloading flows from S3. Adds a helpful graceful logging error message

## Why is this PR important?
If an S3 storage object tries to download a file which does not exist or upload to a bucket that does not exist a botocore `ClientError` will be raised.

### DIFF
--- a/src/prefect/environments/storage/s3.py
+++ b/src/prefect/environments/storage/s3.py
@@ -1,5 +1,5 @@
 import io
-from typing import TYPE_CHECKING, Any, Dict
+from typing import TYPE_CHECKING, Any, Dict, Union
 
 
 import cloudpickle
@@ -56,7 +56,7 @@ class S3(Storage):
 
         super().__init__()
 
-    def get_flow(self, flow_location: str) -> "Flow":
+    def get_flow(self, flow_location: str) -> Union["Flow", None]:
         """
         Given a flow_location within this Storage object, returns the underlying Flow (if possible).
 
@@ -65,7 +65,7 @@ class S3(Storage):
                 a file path where a Flow has been serialized to
 
         Returns:
-            - Flow: the requested flow
+            - Flow: the requested Flow if found
 
         Raises:
             - ValueError: if the flow is not contained in this storage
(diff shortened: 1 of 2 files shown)


src/prefect/environments/storage/s3.py
```diff
@@ -78,9 +79,15 @@ def get_flow(self, flow_location: str) -> "Flow":
...
-            Bucket=self.bucket, Key=flow_location, Fileobj=stream
-        )
+        from botocore.exceptions import ClientError
+
+        try:
+            self._boto3_client.download_fileobj(
+                Bucket=self.bucket, Key=flow_location, Fileobj=stream
+            )
+        except ClientError as err:
+            self.logger.info("Error downloading Flow from S3: {}".format(err))
```
cicdw: two questions:
- if we log this it should be `error`
- I'm worried that returning `None` just causes a more cryptic error downstream, especially when flows are run in Cloud; wouldn't it be more transparent to let the ClientError surface so that the flow run is marked as failed?

joshmeek: Oh yeah that should be `error`. And okay I'm fine with surfacing the client error 👍

joshmeek:
Ugh @cicdw appveyor worker processes left alive 🤔

cicdw:
Noooooo

src/prefect/environments/storage/s3.py
```diff
@@ -59,16 +59,18 @@ def __init__(
...
         Given a flow_location within this Storage object, returns the underlying Flow (if possible).
+        If the Flow is not found an error will be logged and `None` will be returned.
 
         Args:
             - flow_location (str): the location of a flow within this Storage; in this case,
                 a file path where a Flow has been serialized to
 
         Returns:
-            - Flow: the requested flow
+            - Flow: the requested Flow if found, otherwise `None`
```
cicdw:
```suggestion
            - Flow: the requested Flow
```

PR #1817, opened 2019-12-10
Add auto-labeling of flow environments based on storage type

cicdw:
I went with using a property to do it so I didn't have to have a long `if / else` flow inside the flow code.


setup.py
```diff
@@ -17,7 +17,7 @@
         "azureml-sdk >= 1.0.65, < 1.1",
         "azure-cosmos >= 3.1.1, <3.2",
     ],
-    "dev": dev_requires,
+    "dev": dev_requires + test_requires,
```
wagoodman: Nice, makes it easy

src/prefect/environments/storage/gcs.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, bucket: str, key: str = None, project: str = None) -> None:
 
         super().__init__()
 
+    @property
+    def labels(self) -> List[str]:
+        return ["gcs-storage"]
```
wagoodman: Nit pick, consider "gcs-flow-storage" (and similar in other places) to maximize specificity and prevent potential overlap with future features that we may distinct with similar labels

joshmeek: Yeah I'm on board with `-flow-storage` because `gcs-storage` seems like something that a user could tag their flow with if it interfaces with their GCS storage

cicdw:
Thanks!! I've responded to both of your reviews @wagoodman and @joshmeek

joshmeek:
Looks good, barring appveyor

PR #1819, opened 2019-12-10
Universal Deploy

cicdw:
The official universal deploy PR to master.  Opening to investigate any remaining test failures.  

~Windows support TBD.~


src/prefect/agent/kubernetes/resource_manager.py
```diff
@@ -29,11 +31,10 @@ def __init__(self) -> None:
 
         logger = logging.getLogger("resource-manager")
         logger.setLevel(logging.DEBUG)
-        ch = logging.StreamHandler()
+        ch = logging.StreamHandler(sys.stdout)
         ch.setLevel(logging.DEBUG)
```
wagoodman: we can solve this later after this PR, but it seems that this is hard coded to a debugging log level instead of deriving the level from the config

PR #1829, opened 2019-12-11
Enhance nested / ordered constant treatment

jlowin:

## What does this PR change?
This PR enhances the treatment of collections of constants, specifically nested collections and ordered collections.

Previously, collections of constants were represented as a `Collection` task with each item added as an upstream `Constant` (or `Collection`) dependency. This had two effects: first, it ignored order; second, it created a confusing graph because something like `[1, 2, 3]` would unexpectedly be represented by four tasks.

This PR examines nested objects to determine if they only contain constants. If they do, then the entire object is wrapped as a single `Constant`, preserving order and simplicity.

This PR also simplifies the `convert_constants` flag by removing it. Instead, if the result of `as_task` is a `Constant`, we simply add its `value` attribute to the flow's constants.

## Why is this PR important?

As users are supplying more complex arguments to their tasks, they are discovering this undesirable behavior.


jlowin:
@cicdw updated changelog for 0.8.0

cicdw:
@jlowin sorry for the extremely delayed reviews on this - looks like there's a merge conflict in the changelog

jlowin:
Updated!

PR #1831, opened 2019-12-11
Azure Blob Storage for Flows

joshmeek:

## What does this PR change?
Adds Azure Blob storage to accompany the new storage methods for Flows. It connects to Azure Blob storage using a [connection string](https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string).

Also bumps the `azure.storage.blob` requirement to the much newer version of `12.1.0`

## Why is this PR important?
This covers the third main Cloud provider after GCP and AWS :)


tests/serialization/test_storage.py
```diff
@@ -153,6 +154,60 @@ def test_s3_serialize_with_flows():
...
+    )
+    serialized = AzureSchema().dump(azure)
+
+    assert serialized
+    assert serialized["__version__"] == prefect.__version__
+    assert serialized["container"] == "container"
+    assert serialized["blob_name"] == "name"
+
+
+def test_azure_aws_creds_not_serialized():
```
cicdw:
```suggestion
def test_azure_creds_not_serialized():
```

joshmeek: Lol

PR #1848, opened 2019-12-13
Agent Process Enhancements

joshmeek:

## What does this PR change?
Refactors the main agent process.

Previously:
The agent deployed _before_ submitting. It also deployed and submitted over entire list of found flow runs at once which means if there was an error in deploying a single one of those flow runs then they would all fail to be deployed / submitted.

Now:
The process now iterates over flow runs individually which allows for new functionality! Marking as submitted happens _after_ the deployment takes place. Deployment returns some information about the infra created (e.g. k8s job name, process ID, etc...) which is then passed into the message for the `Submitted` state. If any error is encountered during the deployment process it is caught and the flow run is marked `Failed` with the message being the error that occurred. 

## Why is this PR important?
This surfaces more informative information for flow runs and makes deployment safer by treating each flow run as an isolated event.

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -311,7 +311,9 @@ class Agent:
         won't be picked up by any other processes
 
         Args:
-            - flow_runs (list): A list of GraphQLResult flow run objects
+            - flow_run (GraphQLResult): A GraphQLResult flow run object
+            - deployment_info (str): Identifier information related to the Flow Run
+                deployment
         """
         self.logger.debug(
             "Updating states for flow run {}".format(flow_run.id)  # type: ignore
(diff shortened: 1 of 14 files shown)


src/prefect/agent/agent.py
```diff
@@ -184,8 +184,16 @@ def agent_process(self, tenant_id: str) -> bool:
...
-                self.deploy_flows(flow_runs)
+                # Iterate over flow runs
+                for flow_run in flow_runs:
+
+                    # Deploy flow run and mark failed if any deployment error
+                    try:
+                        deployment_info = self.deploy_flow(flow_run)
+                        self.update_state(flow_run, deployment_info)
+                    except Exception as exc:
+                        self.mark_failed(flow_run=flow_run, exc=exc)
```
cicdw: Nice, I like how this API is shaping up

src/prefect/agent/local/agent.py
```diff
@@ -63,65 +63,61 @@ def heartbeat(self) -> None:
...
+            stderr=STDOUT,
+            env=current_env,
+        )
+
+        self.processes.append(p)
+        self.logger.debug(
+            "Submitted flow run {} to process PID {}".format(flow_run.id, p.pid)
+        )
+
+        return "PID {}".format(p.pid)
```
cicdw:
```suggestion
        return "PID: {}".format(p.pid)
```

Issue #1836, opened 2019-12-11
Jira Ticket Task


dylanbhughes:
## Use Case
*Please provide a use case to help us understand your request in context*
When a critical Flow has a Failed Flow Run, it would be awesome if I could open a Jira Ticket assigned to the correct person so that investigating the failure would automatically drop into their workflow.

joshmeek:
Closed in #1861

PR #1861, opened 2019-12-17
Jira Ticket Task

zhen0:

## What does this PR change?
- Adds a Jira notifier that creates and assigns a jira issue if a task enters a specific state (probably failed but I've left it open and can be decided at time of use)

## Why is this PR important?
- Requested in issue Jira Ticket Task #1836

Creating as a draft as I realize this needs tests, changelog update and possibly some docs


src/prefect/utilities/notifications.py
```diff
@@ -279,5 +292,83 @@ def add(x, y):
...
+        def add(x, y):
+            return x + y
+        ```
+    """
+    username = cast(str, prefect.client.Secret("JIRAUSER").get())
+    password = cast(str, prefect.client.Secret("JIRATOKEN").get())
+    serverURL = cast(str, prefect.client.Secret("JIRASERVER").get())
+
+    if not projectName:
+        projectName = cast(str, prefect.client.Secret("JIRAPROJECT").get())
```
cicdw: Two thoughts on the use of secrets here:
- outside of the scope of this PR, it would be nice to find a way around _requiring_ Prefect Cloud Secrets
- since you require 4 distinct secrets, you might consider refactoring this to one `JIRA_CREDS` secret which is a JSON payload (a dictionary in Python) with these four keys; might make the setting of the secret easier on the user (that is also just my own personal preference - open to push back)

src/prefect/utilities/notifications.py
```diff
@@ -279,5 +292,83 @@ def add(x, y):
...
+    tracked_obj: TrackedObjectType,
+    old_state: "prefect.engine.state.State",
+    new_state: "prefect.engine.state.State",
+    ignore_states: list = None,
+    only_states: list = None,
+    projectName: str = None,
+    assignee: str = -1
+) -> "prefect.engine.state.State":
+    """
+    Slack state change handler; requires having the Prefect slack app installed.
```
cicdw: This whole docstring is still the slack docstring

src/prefect/utilities/notifications.py
```diff
@@ -152,6 +157,13 @@ def slack_message_formatter(
     return data
 
 
+def jira_message_formatter(tracked_obj: TrackedObjectType, state: "prefect.engine.state.State") -> str:
+    time = datetime.now()
```
cicdw: This call will create a timestamp in the local timezone; you might want to consider using `datetime.utcnow()` instead

src/prefect/utilities/notifications.py
```diff
@@ -279,5 +292,83 @@ def add(x, y):
...
     r = requests.post(webhook_url, json=form_data)
     if not r.ok:
-        raise ValueError("Slack notification for {} failed".format(tracked_obj))
+        raise ValueError(
+            "Slack notification for {} failed".format(tracked_obj))
+    return new_state
+
+
+@curry
+def jira_notifier(
```
cicdw: In terms of docs, we'll also want to add this to the `docs/outline.toml` file

src/prefect/utilities/notifications.py
dylanbhughes: I like Chris's suggestion here

aroder:
@zhen0 I like the idea of being able to pass an assignee, and I think that we will need more fields than that. We have some custom Jira fields that we would want to pass in, that allow us to use Jira to configure automatic routing and assignment of newly created tickets. Have you found a way to pass in a collection of custom fields? 

Our use case is Jira Service Desk, but I have not had a chance to review and compare the APIs between Jira Software and Jira Service Desk. They may be the same API.

From past experience, these kinds of automatic ticket creation triggers can result in thousands of tickets being created. This could be handled by the developer if careful thought is put into when a ticket is generated--but often there are emergent cases not thought of during design. The pattern I've seen that handles this is for the state handler to recognize a case when lots of duplicate tickets are about to be created, and do something like create 1 ticket with a note that the event occurred N number of times. This definitely is not a priority for us, just some feedback as you design this.

zhen0:
@cicdw - I think this is ready for another review.  I've incorporated your requested changes (one jira  creds secret, time in utc, added docs and tests.). I've also added a bit more flexibility for tailoring the call to jira which should allow adding different issue types (which means the state handler can work for jira service desk as well as standard jira) and custom fields.

joshmeek:
@zhen0 Do you happen to have an image of what this notification looks like in Jira?

zhen0:
Screen Shot 2020-02-28 at 11 52 06 AM

Here you go @joshmeek

zhen0:
And from Jira Service Desk:

Screen Shot 2020-02-28 at 11 57 29 AM

src/prefect/utilities/jira_notification.py
```diff
@@ -0,0 +1,121 @@
...
+    from jira import JIRA
+except ImportError:
+    raise ImportError(
+        'Using `jira_notifier` requires Prefect to be installed with the "jira" extra.'
+    )
+
+if TYPE_CHECKING:
+    import prefect.engine.state
+    import prefect.client
+    from prefect import Flow, Task
```
cicdw: One of our CI checks is failing because of these unused imports; I recommend removing this line and replacing

```python
TrackedObjectType = Union[prefect.Flow, prefect.Task]
```

below

docs/outline.toml
```diff
@@ -449,6 +449,11 @@ title = "Notifications and Callback Tools"
 module = "prefect.utilities.notifications"
 functions = ["callback_factory", "slack_notifier", "gmail_notifier"]
 
+[pages.utilities.jira_notification]
```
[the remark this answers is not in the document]
zhen0: As jira is not imported as standard, it needs to be imported as an extra.  We thought moving it to its own page would allow us to raise a warning if jira is not already installed for the jira notifier without affecting the other notifiers.

src/prefect/utilities/jira_notification.py
zhen0: A ha!  Thank you!

docs/outline.toml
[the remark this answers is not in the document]
cicdw: Gotcha; so whether it requires an extra or not actually doesn't matter for the auto-generated documentation -- I'm able to patch any missing imports so that I can still extract the relevant docstrings.  If you put it in the same page as the others, you'd need to add a 
```python
try:
    from .jira import jira_notifier # might not be 100%
except ImportError:
    pass
```
into `utilities.notifications.__init__.py`

Up to you though, if you want it on its own documentation page that's OK

src/prefect/utilities/jira_notification.py
```diff
@@ -0,0 +1,120 @@
...
+    old_state: "prefect.engine.state.State",
+    new_state: "prefect.engine.state.State",
+    ignore_states: list = None,
+    only_states: list = None,
+    server_URL: str = None,
+    options: dict = {},
+    assignee: str = "-1",
+) -> "prefect.engine.state.State":
+    """
+    Jira Notifier requires a Jira account and API token.  They API token can be created at: https://id.atlassian.com/manage/api-tokens
```
joshmeek:
```suggestion
    Jira Notifier requires a Jira account and API token.  The API token can be created at: https://id.atlassian.com/manage/api-tokens 
```

src/prefect/utilities/jira_notification.py
```diff
@@ -0,0 +1,123 @@
...
+    The Jira account username ('JIRAUSER'), API token ('JIRATOKEN') should be set as part of a 'JIRASECRETS' object in Prefect Secrets. 
+    An example 'JIRASECRETS' object looks like this:
+    JIRASECRETS = { JIRATOKEN = "XXXXXXXXX", JIRAUSER = "xxxxx@yyy.com", JIRASERVER = "https://???.atlassian.net", JIRAPROJECT = "TEST" }
+    The server URL can be set as part of the 'JIRASECRETS' object ('JIRASERVER') or passed to the jira notifier state handler as the "server_URL" argument.
+    Jira Notifier works as a standalone state handler, or can be called from within a custom
+    state handler.  This function is curried meaning that it can be called multiple times to partially bind any keyword arguments (see example below).
+    Jira Notifier creates a new ticket with the information about the task or flow it is bound to when that task or flow is in a specific state. 
+    (For example it will create a ticket to tell you that the flow you set it on is in a failed state.)  
+    You should use the options dictionary to set the project name and issue type.  
+    You can use the "assignee" argument to assign that ticket to a specific member of your team.
```
joshmeek:
```suggestion
    Jira Notifier requires a Jira account and API token.  The API token can be created at: https://id.atlassian.com/manage/api-tokens 
    The Jira account username ('JIRAUSER'), API token ('JIRATOKEN') should be set as part of a 'JIRASECRETS' object in Prefect Secrets. 

    An example 'JIRASECRETS' object looks like this:
    ```
    JIRASECRETS = { JIRATOKEN = "XXXXXXXXX", JIRAUSER = "xxxxx@yyy.com", JIRASERVER = "https://???.atlassian.net", JIRAPROJECT = "TEST" }
    ```

    The server URL can be set as part of the 'JIRASECRETS' object ('JIRASERVER') or passed to the jira notifier state handler as the "server_URL" argument.
    Jira Notifier works as a standalone state handler, or can be called from within a custom
    state handler.  This function is curried meaning that it can be called multiple times to partially bind any keyword arguments (see example below).
    Jira Notifier creates a new ticket with the information about the task or flow it is bound to when that task or flow is in a specific state. 
    (For example it will create a ticket to tell you that the flow you set it on is in a failed state.)  
    You should use the options dictionary to set the project name and issue type.  
    You can use the "assignee" argument to assign that ticket to a specific member of your team.
```

cicdw:
OK so I got this working so that the new jira notifier can be documented alongside the other notifiers and bypasses import issues.  You can see my changes here: https://github.com/PrefectHQ/prefect/pull/1861/commits/7ff4d8f17476da8802495f7b76d84e728f60cd74

Note that this change allows us to document classes / functions from un-imported submodules alongside other classes / functions

PR #1862, opened 2019-12-17
Allow for task looping to beyond Python's maximum recursion depth

### LINKED ISSUE #1827: LOOPing a Task depends on the recursion limit for Python
## Description
Right now, Looping a task relies on recursion. This can cause the user to experience the following error: `RecursionError: maximum recursion depth exceeded in comparison`. Unless the user updates the system recursion limit for python (which may not be a great idea) the maximum number of times their task can loop is capped.

## Expected Behavior

I expected looping to support more than a limited loop count.

## Reproduction

```python
import prefect
from prefect.engine.signals import LOOP

@task
def example():
    loop_payload = prefect.context.get("task_loop_result", 0)
    if loop_payload < 4000:
        loop_payload += 1
        raise LOOP(result=loop_payload)

    return loop_payload

with Flow("Example") as flow:
    result = example()
```

I run into the following error:
```bash
In [16]: flow.run()
[2019-12-10 23:16:25,571] INFO - prefect.FlowRunner | Beginning Flow run for 'Example'
[2019-12-10 23:16:25,575] INFO - prefect.FlowRunner | Starting flow run.
[2019-12-10 23:16:25,583] INFO - prefect.TaskRunner | Task 'example': Starting task run...
[2019-12-10 23:16:29,646] ERROR - prefect.TaskRunner | Task 'example': unexpected error while running task: RecursionError('maximum recursion depth exceeded in comparison')
  File "/Users/dylanhughes/dev/prefect/src/prefect/engine/task_runner.py", line 229, in run
    with prefect.context(context):
  File "/Users/dylanhughes/miniconda3/envs/product_flows/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/dylanhughes/dev/prefect/src/prefect/utilities/context.py", line 111, in __call__
    previous_context = self.copy()
  File "/Users/dylanhughes/dev/prefect/src/prefect/utilities/collections.py", line 107, in copy
    return type(self)(self.__dict__.copy())
  File "/Users/dylanhughes/dev/prefect/src/prefect/utilities/context.py", line 78, in __init__
    super().__init__(*args, **kwargs)
  File "/Users/dylanhughes/dev/prefect/src/prefect/utilities/collections.py", line 62, in __init__
    super().update(init_dict)
  File "/Users/dylanhughes/miniconda3/envs/product_flows/lib/python3.7/_collections_abc.py", line 839, in update
    if isinstance(other, Mapping):
  File "/Users/dylanhughes/miniconda3/envs/product_flows/lib/python3.7/abc.py", line 139, in __instancecheck__
    return _abc_instancecheck(cls, instance)
RecursionError: maximum recursion depth exceeded in comparison
[2019-12-10 23:16:29,750] INFO - prefect.TaskRunner | Task 'example': finished task run for task with final state: 'Failed'
[2019-12-10 23:16:29,753] INFO - prefect.TaskRunner | Task 'example': finished task run for task with final state: 'Failed'
[2019-12-10 23:16:29,753] INFO - prefect.FlowRunner | Flow run FAILED: some reference tasks failed.
Out[16]: <Failed: "Some reference tasks failed.">
```

## Environment
Standard python environment (python 3.7 docker container)

wagoodman:

## What does this PR change?
Modifies the recursive call for task LOOPing to not be tied to python's recursion depth limitations. Specifically, instead of invoking the `run` function recursively, a `RecursiveCall` exception is raised and handled by the `tail_recursion` decorator. This is not optimized to utilize the same stack frame for each invocation of `run`, however, performance impact has been measured to be negligible.

Closes #1827


cicdw:
This is super clever!! I have a question about the implementation: because you rely on raising an exception here, will Python collect the stacks from each raise?  Meaning, if I raise a `ValueError` on the third recursive call, will the stack trace include all three `RecursiveCall` exception raises?

src/prefect/utilities/executors.py
```diff
@@ -10,7 +10,7 @@
 from concurrent.futures import TimeoutError as FutureTimeout
 from functools import wraps
 from logging import Logger
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Union, Callable, Optional
```
domarp-j: Very minor, but it looks like `Callable` is being imported twice.

wagoodman:
@cicdw Thanks! Regarding the stack trace, no, only the latest call will be in the trace. This is because the exception is caught outside of the run function and ignored, then the loop will invoke the function again. As far as the interpreter is concerned from a stack perspective, the old stack is gone and a new one is created. For example:
```python
import prefect
from prefect.engine.signals import LOOP

@task
def example():
    loop_payload = prefect.context.get("task_loop_result", 0)
    if loop_payload < 40:
        loop_payload += 1
        raise LOOP(result=loop_payload)
    if loop_payload > 5:
        raise ValueError("blerg")

    return loop_payload

with Flow("Example") as flow:
    result = example()
    flow.run()
```
will result in:
```bash
[2019-12-17 18:08:32,583] INFO - prefect.FlowRunner | Beginning Flow run for 'Example'
[2019-12-17 18:08:32,587] INFO - prefect.FlowRunner | Starting flow run.
[2019-12-17 18:08:32,603] INFO - prefect.TaskRunner | Task 'example': Starting task run...
[2019-12-17 18:08:32,990] ERROR - prefect.TaskRunner | Unexpected error: ValueError('blerg')
  File "/Users/wagoodman/code/prefect/src/prefect/engine/runner.py", line 48, in inner
  File "/Users/wagoodman/code/prefect/src/prefect/engine/task_runner.py", line 877, in get_task_run_state
  File "/Users/wagoodman/code/prefect/src/prefect/utilities/executors.py", line 250, in timeout_handler
    return fn(*args, **kwargs)
  File "example.py", line 12, in example
    raise ValueError("blerg")
ValueError: blerg
[2019-12-17 18:08:33,007] INFO - prefect.TaskRunner | Task 'example': finished task run for task with final state: 'Failed'
[2019-12-17 18:08:33,009] INFO - prefect.FlowRunner | Flow run FAILED: some reference tasks failed.
```

src/prefect/utilities/executors.py
wagoodman: good call. Actually, both are dead code... I'll do a last sweep through the imports.

src/prefect/engine/cloud/task_runner.py
```diff
@@ -249,6 +250,7 @@ def check_task_is_cached(self, state: State, inputs: Dict[str, Result]) -> State
 
         return state
 
+    @tail_recursive
```
cicdw: for my own understanding, is this decorator necessary because the standard `TaskRunner` uses `self.run` within its recursive raise?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 - Add flow run name to `prefect.context` - [#1855](https://github.com/PrefectHQ/prefect/pull/1855)
 - Add `--namespace` option for Kubernetes Agent start CLI command - [#1859](https://github.com/PrefectHQ/prefect/pull/1859)
 - Add Prefect job resource configuration for Kubernetes Agent - [#1859](https://github.com/PrefectHQ/prefect/pull/1859)
+- Allow for task looping to beyond Python's maximum recursion depth - [#1862](https://github.com/PrefectHQ/prefect/pull/1862)
```
cicdw: We'll need to move this entry once #1865 is merged

src/prefect/utilities/executors.py
```diff
@@ -286,3 +286,56 @@ def run_with_ctx(*args: Any, _ctx_dict: dict, **kwargs: Any) -> Any:
...
+
+
+class RecursiveCall(Exception):
+    def __init__(self, func: Callable, *args: Any, **kwargs: Any):
+        self.func = func
+        self.args = args
+        self.kwargs = kwargs
+
+
+def tail_recursive(func: Callable) -> Callable:
```
cicdw: So I think we need to use `functools.wraps` here, because it appears that this decorator is hiding the `run` method of the runner classes from our API documentation (e.g., https://github.com/PrefectHQ/prefect/blob/167f8afd660e3f24225be168974d533383ee636f/src/prefect/engine/runner.py#L42)

src/prefect/utilities/executors.py
```diff
@@ -286,3 +286,56 @@ def run_with_ctx(*args: Any, _ctx_dict: dict, **kwargs: Any) -> Any:
...
+
+                # there may be multiple nested recursive calls, we should only respond to calls for the
+                # wrapped function explicitly, otherwise allow the call to continue to propagate
+                if call_func != func:
+                    raise exc
+                args = exc.args
+                kwargs = exc.kwargs
+                continue
+
+    setattr(wrapper, "__wrapped_func__", func)
```
cicdw: out of curiosity, why use `setattr` here instead of setting the attribute directly as you do below?

src/prefect/utilities/executors.py
```diff
@@ -286,3 +286,56 @@ def run_with_ctx(*args: Any, _ctx_dict: dict, **kwargs: Any) -> Any:
...
+    wrapper.__doc__ = func.__doc__
```
cicdw: This shouldn't be necessary if you use `functools.wraps`

tests/utilities/test_executors.py
```diff
@@ -224,3 +230,95 @@ def my_fun(x, **kwargs):
...
+    @tail_recursive
+    def my_func(a=0):
+        if a > 5:
+            return a
+        raise RecursiveCall(my_func, a + 2)
+
+    assert 6 == my_func()
+
+
+def test_recursion_beyond_python_limits():
```
cicdw: Haha very nice

src/prefect/utilities/executors.py
wagoodman: purely to make mypy happy, which claimed there was no `__wrapped_func__` attribute

src/prefect/engine/cloud/task_runner.py
wagoodman: Great question! I found this surprising too when cloud tests started to fail (yea for great tests!), but after taking a closer look at https://github.com/PrefectHQ/prefect/pull/1862/files#diff-198ff4fb33d18887ded1ff424b56dbb0R333 it made sense. When I added the ability for `tail_recursive` to determine which function it should invoke functions for, it allows for recursive calls for other functions to propagate. Strictly speaking, `CloudTaskRunner.run != TaskRunner.run`, and this is a good thing! Otherwise, we may invoke from the CloudTaskRunnner `super().run()` instead of `self.run()` if the TaskRunner.run() `tail_recursive` decorator was blindly catching _any_ RecursiveCall, resulting in different behavior than today. 

So the above line ensures that the CloudTaskRunner.run() is invoked on each `LOOP`, not just TaskRunner.run(). This is loosely what the `test_recursion_multiple` test is testing for: only catch `RecursiveCall`s that belong to exactly this function.

src/prefect/utilities/executors.py
wagoodman: great catch, I'll incorporate

CHANGELOG.md
cicdw: @wagoodman - looks like #1865 is in master!

PR #1863, opened 2019-12-17
Update flow docs for Cloud

zdhughes:

## What does this PR change?

This PR updates broken links in `flows.md` and documents how to set flow settings.

## Why is this PR important?

There is a very real demand for these flow settings, but they're not particularly useful without docs explaining how to use them and what they do.

### DIFF
--- a/docs/cloud/concepts/flows.md
+++ b/docs/cloud/concepts/flows.md
@@ -85,6 +85,10 @@ mutation {
 }
 ```
 
+::: tip Prefect version
+Heartbeats can only be disabled for flows registered using Prefect versions `0.8.1` and up. The other settings listed below are not version-dependent.
+:::
+
 To reenable heartbeats for a Flow, rerun this mutation with `value` set to `False`. Future runs of this Flow will resume standard heartbeat functionality.
 
 ### Disable Lazarus
--- a/docs/cloud/concepts/flows.md
+++ b/docs/cloud/concepts/flows.md
@@ -73,7 +73,7 @@ These safeguards can be disabled on a Flow-by-Flow basis using Flow settings.
 Disabling these safeguards can alter fundamental assumptions about how Flows run in Cloud. Be sure to read the docs and understand how each of these settings alters Flow behavior in Cloud.
 :::
 
-### Disable Heartbeats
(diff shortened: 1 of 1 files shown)


docs/cloud/concepts/flows.md
```diff
@@ -39,22 +39,80 @@ mutation($flow: JSON!) {
...
+- [Lazarus resurrections](lazarus-process.html)
+- and version locking
+
+These safeguards can be disabled on a Flow-by-Flow basis using Flow settings.
+
+::: warning Disabling safeguards
+Disabling these safeguards can alter fundamental assumptions about how Flows run in Cloud. Be sure to read the docs and understand how each of these settings alters Flow behavior in Cloud.
+:::
+
+### Disable Heartbeats
```
cicdw: Actually, could you call out that this mechanism will only work on Flows registered with Prefect Core 0.8.1+

zdhughes: Good shout-- change incoming.

zdhughes: Updated.

docs/cloud/concepts/flows.md
```diff
@@ -85,6 +85,10 @@ mutation {
 }
 ```
 
+::: tip Prefect version
```
cicdw: Question for @joshmeek / @zdhughes: tip or warning?

zdhughes: I had the same question. My gut said `tip`, but I can't exactly quantify why I leaned that way.

cicdw: Obviously this isn't a huge deal, but I was thinking warning because it's saying "Note this won't work in some circumstances!" instead of "Here's how you can improve this"

joshmeek: What if instead of a box you add a badge to the `Disable Heartbeats` title? 
```
### Disable Heartbeats <Badge text="0.8.1+"/>
```

cicdw: That works too and is more visible -- let's go with that!

zdhughes: Works for me-- final change incoming.

PR #1871, opened 2019-12-19
Add task for kicking off a flow run

zdhughes:

## What does this PR change?

This PR adds a task to the library for kicking off a flow run in Cloud.

## Why is this PR important?

We have semi-frequent questions about how to kick off a flow from another flow. This task is a lightweight first step for addressing this.


src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+        flow = self.client.graphql(query).data.flow
+        # verify that an ID was returned
+        if not flow:
+            raise ValueError(
+                "No flow {} found in project {}.".format(flow_name, project_name)
+            )
+        # and that there was only one ID returned
+        if len(flow) > 1:
```
joshmeek: Will this error out every time you have multiple versions of a flow?

zdhughes: V. good point-- let me update that.

joshmeek:
Could this be updated to accept two options:
1. Flow name w/ project name and version
2. A flow id

Also can you return the ID of the flow run that was created?

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError(
+                "More than one flow named {} found in project {}.".format(
+                    flow_name, project_name
+                )
+            )
+        flow_id = flow[0].id
+        self.client.create_flow_run(flow_id=flow_id, parameters=parameters)
```
joshmeek: Can you return the ID of the flow run that was created?

tests/tasks/cloud/test_flow_run.py
```diff
@@ -0,0 +1,76 @@
...
+        create_flow_run=MagicMock(return_value="xyz890"),
+    )
+    monkeypatch.setattr(
+        "prefect.tasks.cloud.flow_run.Client", MagicMock(return_value=cloud_client),
+    )
+    yield cloud_client
+
+
+class TestFlowRunTask:
+    def test_initialization(self):
```
cicdw: Could you throw a few task-generic kwargs into this test? e.g., "name", "tags" -- this will ensure that `super().__init__()` was called correctly

tests/tasks/cloud/test_flow_run.py
```diff
@@ -0,0 +1,76 @@
...
+            flow_name="Test Flow",
+            parameters={"test": "ing"},
+        )
+        # verify that run returns the new flow run ID
+        assert task.run() == "xyz890"
+        # verify the GraphQL query was called with the correct arguments
+        client.graphql.assert_called_once_with(
+            {
+                "query": {
+                    'flow(where: { name: { _eq: "Test Flow" }, project: { name: { _eq: "Test Project" } } }, order_by: { version: desc })': {
```
cicdw: This test is failing in a lower version of python; I'm OK with this test but it also feels like one of those that could be frustrating to maintain if we add new outputs or anything.  Maybe we can relax what we test for?  Something like:
```python
arg = client.graphql.call_args_list[0][0]
assert "Test Flow" in arg
# etc.
```

just a thought, not a requirement

zdhughes: Ah man, think I'm getting bitten by dict order issues. I'm fine with that proposal-- didn't see this issue locally.

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        flow_name: str = None,
+        project_name: str = None,
+        parameters: dict = None,
+        **kwargs: Any
+    ):
+        self.flow_name = flow_name
+        self.project_name = project_name
+        self.parameters = parameters
+        self.client = Client()
```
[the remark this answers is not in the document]
zdhughes: Good point-- will update.

src/prefect/tasks/cloud/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+This Task class is "Cloud" in the sense that it relies on Cloud to do anything meaningful.
```
jlowin: This 
```suggestion
These Task classes rely on Prefect Cloud.
```

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,75 @@
...
+from prefect import Task
+
+from prefect.client import Client
+from prefect.utilities.tasks import defaults_from_attrs
+from prefect.utilities.graphql import with_args, EnumValue
+
+
+class FlowRunTask(Task):
+    """
+    Task used to kick off a Flow Run in Cloud. If multiple versions of the flow are found,
```
jlowin:
```suggestion
    Task used to kick off a Flow Run in Prefect Cloud. If multiple versions of the flow are found,
```

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,75 @@
...
+        Returns:
+            - str: the ID of the newly-scheduled flow run
+        """
+        # verify that flow and project names were passed in some capacity or another
+        if project_name is None:
+            raise ValueError("Must provide a project name.")
+        if flow_name is None:
+            raise ValueError("Must provide a flow name.")
+
+        # find the flow ID to schedule
```
jlowin: This might pick up archived flows - I'm not sure of the expected behavior in that case (I believe the flow run will fail)

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,75 @@
...
+        client = Client()
+        flow = client.graphql(query).data.flow
+        # verify that something's been returned
+        if not flow:
+            raise ValueError(
+                "No flow {} found in project {}.".format(flow_name, project_name)
+            )
+        # grab the ID for the most recent version
+        flow_id = flow[0].id
+        return client.create_flow_run(flow_id=flow_id, parameters=parameters)
```
jlowin: I think this is fantastic that it returns the run id

src/prefect/tasks/cloud/flow_run.py
zdhughes: Yep, it'll fail. Probably better to proactively handle that-- good shout.

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,76 @@
...
+                with_args(
+                    "flow",
+                    {
+                        "where": {
+                            "name": {"_eq": flow_name},
+                            "project": {"name": {"_eq": project_name}},
+                            "archived": {"_eq": False},
+                        },
+                        "order_by": {"version": EnumValue("desc")},
+                    },
```
jlowin: Sorry @zdhughes one more thing -- for efficiency, you could throw a limit here to avoid a long list of flows (note my GitHub suggestion won't pass Black)
```suggestion
"limit": 1                    },
```

zdhughes: Sure thing, updated.

src/prefect/tasks/cloud/flow_run.py
```diff
@@ -0,0 +1,96 @@
...
+                this method will use the parameters provided at initialization
+
+        Returns:
+            - str: the ID of the newly-scheduled flow run
+
+        Raises:
+            - ValueError: if flow or project names were not provided, or if the flow provided cannot be found
+
+        Example:
+            ``` python
```
cicdw:
```suggestion
            ```python
```

PR #1880, opened 2019-12-26
Allow intervals less than 60 seconds

daryllstrauss:
**Thanks for contributing to Prefect!**

Please describe your work and make sure your PR:

- [x] adds new tests (if appropriate)
- [x] updates `CHANGELOG.md` (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

Note that your PR will not be reviewed unless all three boxes are checked.

## What does this PR change?
IntervalClock should allow any interval greater than 0.

## Why is this PR important?
60 seconds was an arbitrary time period. Some applications can use short intervals.

PR #1881, opened 2019-12-27
Validate interval clocks for cloud deployment

cicdw:
Follows up on https://github.com/PrefectHQ/prefect/pull/1880 with some doc / serialization changes to ensure Prefect Cloud requirements are unchanged.

### DIFF
--- a/src/prefect/serialization/schedule.py
+++ b/src/prefect/serialization/schedule.py
@@ -1,7 +1,7 @@
 from typing import TYPE_CHECKING, Any
 
 import marshmallow
-from marshmallow import fields, post_load
+from marshmallow import fields, post_load, post_dump
 
 import prefect
 from prefect.serialization import schedule_compat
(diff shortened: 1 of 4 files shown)


src/prefect/serialization/schedule.py
```diff
@@ -39,6 +39,17 @@ class Meta:
     end_date = DateTimeTZ(allow_none=True)
     interval = fields.TimeDelta(precision="microseconds", required=True)
 
+    @post_dump
+    def _interval_validation(self, data: dict, **kwargs: Any) -> dict:
```
joshmeek: Should this validation happen on the cloud side? What would happen if this was commented out and then a flow was registered with a sub 60 second clock?

cicdw: This is a great callout; I'd rather it happen prior to any cloud communication for efficiency purposes, but your comment does highlight that I should add an additional `post_load` check that users can't edit (because it'll run in Cloud)

cicdw: Updated

src/prefect/serialization/schedule.py
```diff
@@ -39,6 +39,26 @@ class Meta:
...
+        """
+        if data["interval"] / 1e6 < 60:
+            raise ValueError(
+                "Interval can not be less than one minute when deploying to Prefect Cloud."
+            )
+        return data
+
+    @post_load
+    def create_object(self, data: dict, **kwargs: Any):
+        if data["interval"].total_seconds() < 60:
```
cicdw: @jlowin can you sanity check me here? I was surprised to find this was a `timedelta` object in my tests instead of a raw number and want to make sure this won't cause any errors if the deserialization occurs in another order -- should I perform a type check here?

jlowin: I think you're looking good -- `post_load` is called *after* the deserialization takes place, so it should be a full Python object.

PR #1882, opened 2019-12-27
Google Utilities

cicdw:

## What does this PR change?
This PR begins a larger refactor of how we handle authentication with external services, along with DRY'ing out some re-used code.  Ultimately, we want to move away from _requiring_ Prefect Cloud Secrets and instead give users control for how they authenticate with external services (including GCP), while still allowing users to use our off-the-shelf secret mechanisms.

In particular, this PR refactors the handling and credentialing of Google Clients into a common set of files.  It also removes the implicit requirement that users use Prefect Cloud Secrets for GCP auth.

## Why is this PR important?
This PR marks the first in a series which will:
- do the same for AWS
- do the same for other common Google operations
- allow users to use dynamic environment variables on their agents / machine settings to customize their authentication
- remove the requirement that Secrets are used for GCS Result Handlers (which will be a breaking change)


src/prefect/client/secrets.py
```diff
@@ -74,6 +74,23 @@ class Secret:
...
+        Determine if the secret exists.
+
+        Returns:
+            - bool: a boolean specifying whether the Secret is accessible or not
+        """
+        secrets = prefect.context.get("secrets", {})
+        if self.name in secrets:
+            return True
+        elif prefect.context.config.cloud.use_local_secrets is False:
+            client = Client()
```
jlowin: I'm wondering if it would be more efficient to create the `Client` object when the class __init__ is called, rather than creating it on each method invocation?

src/prefect/utilities/google.py
```diff
@@ -0,0 +1,62 @@
...
+        - submodule: a Python submodule with a Client attribute
+        - credentials (dict, optional): a dictionary of Google credentials used to initialize the Client; if
+            not provided, will attempt to load the Client using ambient environment settings
+        - project (str, optional): the Google project to point the Client to; if not provided, Client defaults
+            will be used
+
+    Returns:
+        - Client: an initialized and authenticated Google Client
+    """
+    Client = getattr(submodule, "Client")
```
jlowin: This abstraction is usually true but I'm not confident it's always true. PubSub has two different clients for example (one for pub, one for sub).

cicdw: Yea that's very fair; my vote would be to cross that bridge when we get to it if that works for you

src/prefect/client/secrets.py
cicdw: Can do - will update

cicdw:
@jlowin I deprecated all `google` references in favor of `gcp` references.

src/prefect/client/secrets.py
cicdw: Updated

setup.py
```diff
@@ -19,6 +19,10 @@
     ],
     "dev": dev_requires + test_requires,
     "dropbox": ["dropbox ~= 9.0"],
+    "gcp": [
+        "google-cloud-bigquery >= 1.6.0, < 2.0",
+        "google-cloud-storage >= 1.13, < 2.0",
+    ],
     "google": [
         "google-cloud-bigquery >= 1.6.0, < 2.0",
         "google-cloud-storage >= 1.13, < 2.0",
```
joshmeek: I think `google` can be removed

jlowin: I agree

src/prefect/tasks/gcp/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    from prefect.tasks.gcp.storage import GCSDownload, GCSUpload, GCSCopy
+    from prefect.tasks.gcp.bigquery import (
+        BigQueryTask,
+        BigQueryLoadGoogleCloudStorage,
+        BigQueryStreamingInsert,
+        CreateBigQueryTable,
+    )
+except ImportError:
+    raise ImportError(
+        'Using `prefect.tasks.gcp` requires Prefect to be installed with the "google" extra.'
```
jlowin:
```suggestion
        'Using `prefect.tasks.gcp` requires Prefect to be installed with the "gcp" extra.'
```

setup.py
cicdw: Yea but that would be a breaking change for requirements files; I figured we would remove that when we remove the analogously changed submodule

cicdw: @jlowin / @joshmeek let me know if you agree; I can remove it if y'all feel strongly.

PR #1884, opened 2019-12-28
Agent environment variables

### LINKED ISSUE #1878: All agents should accept additional env vars for flow "jobs"
Right now, agents set a pre-defined list of environment variables on all flow run jobs which configure Core to run a certain way and talk to Prefect Cloud.

We should open this up a little and allow for users to specify a dictionary of environment variables at agent start time via both `__init__` as well as CLI flags.

This would allow for security / sensitive configurations to be set much more seamlessly and securely.  E.g., if I want to start a Docker agent which grants S3 access to all of its flow runs, it can set AWS creds on each container job and that doesn't need to be persisted as either a Secret or in code.

cicdw:

## What does this PR change?
This PR introduces new initialization kwargs on all Agents for easily setting environment variables on all submitted flow runs.

## Why is this PR important?
This allows users to set environment variables on all flow runs submitted through an agent, which allows for easier and more secure authentication in some situations.  Closes #1878


joshmeek:
Not needed in this PR but a future enhancement could be allowing something like https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/run.py#L49 for providing an env file that is read and passed to each run

PR #1886, opened 2019-12-29
AWS Auth changes

cicdw:

## What does this PR change?
This PR changes how all S3 hooks authenticate with AWS; in particular, it removes the requirement that users store their AWS credentials in a Prefect Cloud Secret and allows for the standard boto3 auth chain.  Note that this resulted in a breaking change - S3 tasks and handlers that relied on the default value for `aws_credentials_secret` will need to explicitly provide it.

## Why is this PR important?
Gives users control over how the authenticate with AWS and allows for a more natural authentication strategy, especially when running Flows in AWS.

### DIFF
--- a/src/prefect/tasks/aws/s3.py
+++ b/src/prefect/tasks/aws/s3.py
@@ -13,20 +13,24 @@ class S3Download(Task):
     Task for downloading data from an S3 bucket and returning it as a string.
     Note that all initialization arguments can optionally be provided or overwritten at runtime.
 
+    For authentication, there are two options: you can set a Prefect Secret containing
+    your AWS access keys which will be passed directly to the `boto3` client, or you can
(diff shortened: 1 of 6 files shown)


src/prefect/tasks/aws/s3.py
```diff
@@ -46,7 +47,9 @@ def run(
             - key (str): the name of the Key within this bucket to retrieve
             - aws_credentials_secret (str, optional): the name of the Prefect Secret
                 that stores your AWS credentials; this Secret must be a JSON string
```
jlowin: Reading this closely, is it actually a "JSON string"? Or just JSON

cicdw: it really should be just JSON, but in this particular task we also convert strings to JSON automatically: https://github.com/PrefectHQ/prefect/pull/1886/files#diff-2ce447dec456df067e7517655ea99398R51-R52

PR #1888, opened 2019-12-30
Default Result Handlers

cicdw:

## What does this PR change?
This PR does many things (some breaking):
- refactors result handlers to not rely on Prefect Secrets, instead recommending users auth through environment variables
- adds sensible default result handlers to various storage options so that many users don't even need to think about configuration
- _requires_ result handlers for all flows with keyed edges that are registered with Prefect Cloud

## Why is this PR important?
This PR will vastly simplify our recovery from failure story and hopefully allows many users to use off-the-shelf result handlers more easily.


src/prefect/environments/storage/azure.py
```diff
@@ -39,12 +40,17 @@ def __init__(
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
 
-        self.connection_string = connection_string or os.getenv("CONNECTION_STRING")
+        self.connection_string = connection_string or os.getenv(
```
joshmeek: Can you update the reference to the env var in https://docs.prefect.io/cloud/execution/storage_options.html#azure-blob-storage

src/prefect/environments/storage/local.py
```diff
@@ -45,7 +44,8 @@ def __init__(self, directory: str = None, validate: bool = True) -> None:
             abs_directory = directory
 
         self.directory = abs_directory
-        super().__init__()
+        result_handler = LocalResultHandler(self.directory, validate=validate)
```
joshmeek: This will write results to the `.prefect/flows` directory. Thoughts? I think it should have it's own separate directory for results

joshmeek: It does `os.path.join(prefect.config.home_dir, "results")` but only if there is no directory specified

joshmeek: Oh wait I see that it does `commonpath`, nvm

src/prefect/environments/storage/azure.py
cicdw: I'll actually take this opportunity to update all of those docs

cicdw: @joshmeek updated this file

src/prefect/core/flow.py
```diff
@@ -1363,6 +1364,10 @@ def register(
         if labels:
             self.environment.labels.update(labels)
 
+        # register the flow with a default result handler if one not provided
+        if not self.result_handler:
+            self.result_handler = self.storage.result_handler
```
jlowin: Since this mutates the flow object - why not do this in `__init__()` instead?

cicdw: The most common pattern for registration is `flow.register(project, **storage_init_kwargs)` which was originally designed for simple docker deploys and so anytime a user uses default storage we don't know what their storage option is until registration time.

That being said, I can _also_ do this at init in the case that a user provides storage; update incoming.

cicdw: Updated

PR #1898, opened 2020-01-08
Always Cache Inputs

cicdw:

## What does this PR change?
This PR introduces `cached_inputs` on _all_ Prefect States.  Moreover, these inputs are stored at all points in the pipeline.  This will allow Cloud users to trivially recover from failure at any point in their Flow.

Moreover, this PR sets checkpointing as the default behavior on all tasks (note that Core users will only see a difference if they set their [checkpointing environment variable](https://docs.prefect.io/core/concepts/persistence.html#checkpointing)).

## Why is this PR important?
This PR brings Prefect Core in line with user expectations whenever things go wrong; instead of realizing that result handlers would have been useful after the fact, Core now requires result handlers and uses them aggressively to make sure recovery / restarting from any point is always an option with no caveats.


joshmeek:
Should the follow-up changes be branched off of this or do you plan on merging this and then making the other PRs? @cicdw

cicdw:
Ah that's a great idea @joshmeek  -> I'm happy to make PRs to this branch for review if that's easiest!  It definitely makes sense

src/prefect/engine/state.py
```diff
@@ -612,6 +642,8 @@ class Success(Finished):
         - message (str or Exception, optional): Defaults to `None`. A message about the
             state, which could be an `Exception` (or [`Signal`](signals.html)) that caused it.
         - result (Any, optional): Defaults to `None`. A data payload for the state.
+        - cached_inputs (dict): A dictionary of input keys to fully hydrated `Result`s.
+            Used / set if the Task requires Retries.
```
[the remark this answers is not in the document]
cicdw: @jlowin updated - the state docstrings were the only place `Retries` was used.

src/prefect/client/client.py
```diff
@@ -573,8 +573,8 @@ def register(
                 "Flows with required parameters can not be scheduled automatically."
             )
         if any(e.key for e in flow.edges) and flow.result_handler is None:
-            raise ClientError(
-                "Flows are required to have a result handler for storing inputs and outputs."
+            warnings.warn(
+                "It is strongly recommended you include a result handler on your Flow for certain Cloud features to work properly."
```
cicdw: @jlowin any thoughts on how to improve this message?

wagoodman: It may be worth making the side effects more explicit, such as `No result handler was specified on your Flow. Cloud features such as input caching and task resuming may not work properly.`

joshmeek:
@cicdw I received the warning twice in the output
```
/Users/josh/Desktop/code/prefect/src/prefect/client/client.py:577: UserWarning: It is strongly recommended you include a result handler on your Flow for certain Cloud features to work properly.
  "It is strongly recommended you include a result handler on your Flow for certain Cloud features to work properly."
```

As a note I got another warning only a single time:
```
/Users/josh/Desktop/code/prefect/src/prefect/environments/storage/docker.py:302: UserWarning: This Docker storage object has no `registry_url`, and will not be pushed.
```

src/prefect/client/client.py
cicdw: Yea that's a good call out

PR #1900, opened 2020-01-08
Augment CloudFlowRunner cancellation

wagoodman:

## What does this PR change?
- Adds `executor.shutdown` to mimic [concurrent.future's Executor.shutdown](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown). A notable difference in behavior is subsequent calls to `submit` or `map` after `shutdown` is called will result in raising `ExecutorShutdown` instead of `RuntimeError`.
- Runner will `ENDRUN` for upon catching an `ExcecutorShutdown` exception, with a `Cancelled` state. This is preferred over a `Failed` state since `executor.Shutdown` is only called on a controlled shutdown, implying that the user wishes to cancel any remaining work --it is intentional that the interface for the executor is functionally different than that of the CloudFlowRunner ("shutdown" vs "cancel").
- Executors collect all futures that are returned to callers. In case `shutdown` is called with the `wait=True`, then the shutdown call will block until all previously submitted work has returned.
- Separate the functions for sending heartbeats to Cloud and checking for states of interest and acting accordingly; currently only acting on a `Cancelled` state is supported, but others can easily be added if necessary. Now these two functionalities can be done on independent cadences.
- Refactored the CloudFlowRunner's main thread as a command and control loop, all other concerns are run in child threads (including running the flow).

## Why is this PR important?
- Migrates cancellation control to the FlowRunner. Today cancellation is an artifact of not being able to update the state due to a change in the version --this is a more intentional action.


src/prefect/engine/cloud/flow_runner.py
```diff
@@ -59,35 +66,136 @@ def say_hello():
...
+
+        self._on_exit()
+        return self.state
+
+    def fetch_current_flow_run_state(self, flow_run_id: str) -> Type[State]:
+        query = {
+            "query": {
+                with_args("flow_run_by_pk", {"id": flow_run_id}): {
+                    "state": True,
+                    "flow": {"settings": True},
```
jlowin: I don't think this `flow: settings` payload is used anywhere, and especially since this query is periodically issued we could safely remove it to avoid the join. 

```suggestion

```

cicdw: It should be used to determine whether heartbeats should fire: https://github.com/PrefectHQ/prefect/blob/d7b122b2afb7f13265f8d7f0a6ae6d6afd26b634/src/prefect/engine/cloud/flow_runner.py#L83

src/prefect/engine/executors/dask.py
```diff
@@ -182,13 +193,30 @@ def wait(self, futures: Any) -> Any:
...
+
+    def shutdown(self, wait: bool = True) -> List[Any]:
+        """
+        Signal the executor that it should cancel current pending work, prevent future work from 
+        being submitted, and optionally wait for any currently running futures to finish execution.
+        Args:
+            - wait (bool): wait for any pending work to be completed (defaults to True)
+        """
+        super().shutdown(wait=wait)
+        if self.futures and wait:
```
joshmeek: While the executor shutdown is waiting for current work to finish will future work still be able to occur?

joshmeek: Oh I see that's what the `self.accepting_work` is used for

src/prefect/engine/cloud/flow_runner.py
jlowin: Ah, I missed it. It might be worthwhile to split it into two calls at this time - one for the flow info and one for the flow run info. The join to the flow table is small but unecessary to make on every single poll.

src/prefect/engine/executors/dask.py
wagoodman: The only odd thing is that the base class doesn't do anything with `wait`, so it reads kinda odd-like here in a child class

src/prefect/engine/cloud/flow_runner.py
wagoodman: Looks like I've removed the ability to disable heartbeats, I'll add that back in. There shouldn't be a reason why we need to get the settings info periodically? That is, we really only need it when we start the flow... the fewer calls the better.

joshmeek:
Would love some thoughts on this. I have a flow that I am using to test Cancellation and it's effect combined w/ a trigger. IIRC Cancelled is a substate of Failed but I could be wrong.
```python
@task
def sleep_me():
    time.sleep(300)
    return "300"

@task(trigger=any_failed)
def should_i_run(x):
    print(x)
    print("I was able to run")

with Flow("long-sleep") as flow:
    s = sleep_me()
    should_i_run(s)
```

I observed some interesting behavior. During the run of the `sleep_me` task I set the state to `Cancelled`. I saw three different outputs occur on three different cancellations. **Note** that the Prefect Cloud state always reflected `Cancelled` once I cancelled it. These are all local logs from the process which runs the flow.

1. Cancelled, flow finishes in `Success` state
```
[2020-01-08 18:55:15,290] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'long-sleep'
[2020-01-08 18:55:15,400] INFO - prefect.CloudFlowRunner | Starting flow run.
[2020-01-08 18:55:15,400] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Scheduled to Running
[2020-01-08 18:55:15,657] INFO - prefect.CloudTaskRunner | Task 'sleep_me': Starting task run...
[2020-01-08 18:55:15,657] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Pending to Running
[2020-01-08 18:55:16,033] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Calling task.run() method...
[2020-01-08 18:55:56,693] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Running to Success
[2020-01-08 18:55:56,854] INFO - prefect.CloudTaskRunner | Task 'sleep_me': finished task run for task with final state: 'Success'
[2020-01-08 18:55:56,857] INFO - prefect.CloudFlowRunner | Flow run SUCCESS: all reference tasks succeeded
[2020-01-08 18:55:56,857] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Running to Success
[2020-01-08 18:55:57,004] DEBUG - prefect.CloudFlowRunner | Flowrunner completed

Aborted!
[2020-01-08 18:56:10,641] INFO - agent | Process PID 51413 returned non-zero exit code

```

2. Cancelled, flow finished locally in `Cancelled` state
```
[2020-01-08 18:59:32,051] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'long-sleep'
[2020-01-08 18:59:32,171] INFO - prefect.CloudFlowRunner | Starting flow run.
[2020-01-08 18:59:32,171] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Scheduled to Running
[2020-01-08 18:59:32,416] INFO - prefect.CloudTaskRunner | Task 'sleep_me': Starting task run...
[2020-01-08 18:59:32,416] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Pending to Running
[2020-01-08 18:59:32,799] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Calling task.run() method...
[2020-01-08 19:00:16,039] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Running to Success
[2020-01-08 19:00:16,199] INFO - prefect.CloudTaskRunner | Task 'sleep_me': finished task run for task with final state: 'Success'
[2020-01-08 19:00:16,202] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Running to Cancelled
[2020-01-08 19:00:16,334] DEBUG - prefect.CloudFlowRunner | Flowrunner completed

Aborted!
[2020-01-08 19:00:26,803] INFO - agent | Process PID 52419 returned non-zero exit code
```

3. Cancelled, nothing happens and the flow continues to run for the five minute duration of the `sleep_me` task then finishes in a `Cancelled` state
```
[2020-01-08 18:57:54,150] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'long-sleep'
[2020-01-08 18:57:54,279] INFO - prefect.CloudFlowRunner | Starting flow run.
[2020-01-08 18:57:54,279] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Scheduled to Running
[2020-01-08 18:57:54,555] INFO - prefect.CloudTaskRunner | Task 'sleep_me': Starting task run...
[2020-01-08 18:57:54,556] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Pending to Running
[2020-01-08 18:57:54,932] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Calling task.run() method...
[2020-01-08 19:02:54,933] DEBUG - prefect.CloudTaskRunner | Task 'sleep_me': Handling state change from Running to Success
[2020-01-08 19:02:55,140] INFO - prefect.CloudTaskRunner | Task 'sleep_me': finished task run for task with final state: 'Success'
[2020-01-08 19:02:55,142] DEBUG - prefect.CloudFlowRunner | Flow 'long-sleep': Handling state change from Running to Cancelled
[2020-01-08 19:02:55,287] DEBUG - prefect.CloudFlowRunner | Flowrunner completed

Aborted!
[2020-01-08 19:03:08,789] INFO - agent | Process PID 52780 returned non-zero exit code
```

wagoodman:
@joshmeek Interesting! I would have expected your case 3 to be the result every time. That is, the already running task should complete, there is no extra behavior in this branch that would cancel an already-running task. I'll dig further.

To focus on one thing, in your use case you are cancelling when the last task in the flow is already running... which brings up an interesting point: what is the least surprising state result here?

- result 1: the final state reports `Cancelled` even though all tasks completed.

- result 2: the final state reports the true state of the flow, as if no cancellation took place.

Result 1 is somewhat artificial, since the flow technically "completed" and we have the ability to accurately report what actually happened. That is, really nothing was cancelled, so don't report the final state as `Cancelled`. I think result 2 ultimately makes the most sense here.

joshmeek:
I've been testing this out and I am getting consistent results now 👍

wagoodman:
This will be part of a larger cancellation PIN (partially drafted, see me) when it is prioritized on the Roadmap.

PR #1918, opened 2020-01-15
Add Docker daemon reconnect in the Docker agent

### LINKED ISSUE #1916: Reconnect Docker Agent on daemon restart
## Description
When a Docker Agent is running and connected to a docker daemon it will raise a broken pipe error when the daemon restarts and a new flow run is found.

## Expected Behavior
Docker Agent either gracefully reconnects to the daemon or exits when the pipe is broken. I'm leaning on the side of reconnecting as all it takes is reinstantiating the Docker low level API client

wagoodman:

## What does this PR change?
Modifies the docker agent to ping the docker daemon on heartbeat. After 3 failed reattempts, the agent will exit.

Closes #1916

## Why is this PR important?
This will stabilize the Docker agent further when there is intermittent connections or if the Docker daemon is restarted.


src/prefect/agent/docker/agent.py
```diff
@@ -67,6 +68,23 @@ def __init__(
             )
             raise exc
 
+    def heartbeat(self) -> None:
```
joshmeek: When is this heartbeat called?

joshmeek: Oh nvm I remember. it's called in the base agent class 👍

src/prefect/agent/docker/agent.py
```diff
@@ -67,6 +68,23 @@ def __init__(
...
+            if not self.docker_client.ping():
+                raise RuntimeError("Unexpected Docker ping result")
+            if self.failed_connections > 0:
+                self.logger.info("Reconnected to Docker daemon")
+            self.failed_connections = 0
+        except Exception as exc:
+            self.logger.warning("Failed heartbeat: {}".format(repr(exc)))
+            self.failed_connections += 1
+
+        if self.failed_connections >= 3:
```
joshmeek: I think we should bump this number up. At an iteration of 10s a daemon restart will take probably around a minute to occur. I vote for increasing this to 6 for a full minute of retries

joshmeek: Actually the more I think of it the more inclined I am to say we should exit at all 🤔 I think it should continue to make connection until a daemon is found since there is no external process responsible for restarting this it could go down on a long restart of the daemon. Thoughts?

wagoodman: I feel that the agent should not be pulling work if there is risk that the work will never be done. Making an attempt to reconnect makes sense at the least --we want to be somewhat resilient to disconnects and restarts.

So it ends up that a daemon restart doesn't take that much time. On a linux system where there is not an underlying VM it should be much less than 30 seconds. On Windows/OSX, this restart time includes the time it takes to restart the VM. Initial tests on OSX shows that it squeaks in under 30 seconds... I see no problem in increasing the number of reconnects to allow for more time here, and a minute seems reasonable.

The one issue with the approach on this PR is that it is somewhat abusing the heartbeat mechanism. Agent heartbeating is intended to indicate to external orchestrators or process managers that the agent is still alive. The way that heartbeating is being used in this PR is to measure liveliness of downstream dependences (in this case the Docker daemon) and take action appropriately (logging and exiting, essentially). (in short: I realize that this is overloading the heartbeat for other purposes)

cicdw: Yea, with the new agent management screens I'd rather the agent exit so that users can see "this agent is no longer capable of submitting work"

joshmeek: Okay that all makes sense to me! Let's bump to a minute 👍

PR #1928, opened 2020-01-17
Background agent flow deployments to worker threads

### LINKED ISSUE #1923: Docker Agent should pull images in the background
## Current behavior
Right now, the docker agent pulls flow images synchronously which, depending on slow connections can take up to 5 minutes for large images.

## Proposed behavior
It would be nice if the Docker agent instead pulled these images "in the background" (maybe in a background thread if possible which can be cleaned up with each heartbeat?).  This would increase the speed at which flows can be submitted in parallel and prevent any edge cases with the proposed new "agent registration" hooks.

wagoodman:

## What does this PR change?
Adds a thread pool to all agents where all flow deployments are attempted. The agent stops pulling work upon the thread pool reaching max capacity.

Closes #1923 

## Why is this PR important?
This allows for an agent to act on new available work more quickly, even if the act of deploying the pulled flow runs takes time. This is the case with docker flow run deployments, where pulling an image may take considerable time, which would previously block the agent from deploying more work. With the introduction of backgrounding flow run deployments, docker pulls will not block the foreground thread and work can continue to be pulled from cloud.


src/prefect/agent/agent.py
```diff
@@ -190,22 +246,17 @@ def agent_process(self, tenant_id: str) -> bool:
...
-                        self.update_state(flow_run, deployment_info)
-                    except Exception as exc:
-                        self.mark_failed(flow_run=flow_run, exc=exc)
-
-                self.logger.info(
-                    "Submitted {} flow run(s) for execution.".format(len(flow_runs))
+            for flow_run in flow_runs:
+                fut = executor.submit(self.deploy_and_update_flow_run, flow_run)
+                self.submitting_flow_runs.add(flow_run.id)
+                fut.add_done_callback(
```
cicdw: Clever

src/prefect/agent/agent.py
```diff
@@ -254,8 +305,25 @@ def query_flow_runs(self, tenant_id: str) -> list:
                 }
             },
         )
-        flow_run_ids = result.data.getRunsInQueue.flow_run_ids  # type: ignore
-        self.logger.debug("Found flow runs {}".format(flow_run_ids))
+
+        # we queried all of the available flow runs, however, some may have alreay been pulled
```
cicdw:
```suggestion
        # we queried all of the available flow runs, however, some may have already been pulled
```

src/prefect/agent/agent.py
```diff
@@ -125,20 +129,26 @@ def start(self) -> None:
...
-            while not exit_event.wait(timeout=loop_intervals[index]):
-                self.heartbeat()
 
-                runs = self.agent_process(tenant_id)
-                if runs:
-                    index = 0
-                elif index < max(loop_intervals.keys()):
-                    index += 1
+            # the max workers default has changed in 3.5 and 3.8. For stable results the
+            # default 3.8 beahvior is elected here.
```
cicdw:
```suggestion
            # default 3.8 behavior is elected here.
```

src/prefect/agent/agent.py
```diff
@@ -125,20 +129,26 @@ def start(self) -> None:
...
+            max_workers = min(32, (os.cpu_count() or 1) + 4)
 
-                self.logger.debug(
-                    "Next query for flow runs in {} seconds".format(
-                        loop_intervals[index]
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
```
cicdw: Potentially dumb question: in going with this implementation, is there a risk that a deadlock occurs if the agent finds `max_workers + 1` flow runs to submit?

wagoodman: Not under that specific condition, no, there will always be a [future immediately returned by `submit`](https://github.com/python/cpython/blob/3.8/Lib/concurrent/futures/thread.py#L184-L189). There are two risks of deadlocks:
- during executor shutdown, if there is an underlying race with the [`_shutdown_lock`](https://github.com/python/cpython/blob/3.8/Lib/concurrent/futures/thread.py#L174)
- when putting work into the queue, if the [underlying file descriptors that make up the queue](https://github.com/python/cpython/blob/f84f65be5602e561fef04b66bb487fbc4e560db5/Lib/multiprocessing/connection.py#L527-L529) are in contention.
Though they would prevent `submit` from returning, both are unrelated to the value of max_workers and are very unlikely to occur.

I think the highest risk case is when the agent turns into a "black hole" for work: if the underlying workers never return, leaving new work to never be scheduled. The main thread would continue to pull work, backing up the queue further with work that wont be done. In this case, Lazarus will kick in and it will be scheduled again in which case the same work will not be queued again (due to the `self.submitting_flow_runs set`).

joshmeek:
Yeah I want to run a big test on this to see what happens @cicdw 

@wagoodman is this in a good spot to test?

joshmeek:
After some preliminary testing this appears to be working great! Really speeds up execution. There's also a nice added benefit I noticed in the case of using a flow which has Docker based storage and the Docker agent. If I pull multiple runs of the same flow then the pulls now happen in parallel and subsequently occur much faster since they are all collaborating on the layers. 👍

PR #1929, opened 2020-01-21
Add show_flow_logs to the Docker agent

wagoodman:

## What does this PR change?
Adds the `show_flow_logs` agent option to the `DockerAgent`, forwarding all subprocess flow logs to the agent's standard output.

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -118,27 +118,45 @@ class Agent:
         The main entrypoint to the agent. This function loops and constantly polls for
         new flow runs to deploy
         """
-        with exit_handler(self) as exit_event:
-            tenant_id = self.agent_connect()
+        try:
+            with exit_handler(self) as exit_event:
+                tenant_id = self.agent_connect()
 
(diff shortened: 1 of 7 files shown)


src/prefect/agent/agent.py
```diff
@@ -118,27 +118,45 @@ def start(self) -> None:
...
+                    self.logger.debug(
+                        "Next query for flow runs in {} seconds".format(
+                            loop_intervals[index]
+                        )
                     )
-                )
+        finally:
+            self.on_shutdown()
+
+    def on_shutdown(self) -> None:
```
joshmeek: Question: do you think this shutdown hook is worth of a changelog update?

joshmeek: Discussed offline, no need for an update

wagoodman: completely subjective: I vote no. The hook isn't being used by any other agent right now thus doesn't modify any existing behavior. If this was a notable hook in another way I might be inclined to say otherwise.

PR #1904, opened 2020-01-09
Relax result handler requirement

cicdw:
**NOTE**: PR'ing into #1898 

This PR:
- reverts checkpointing behavior _essentially_ to its defaults, with one change: if checkpointing is turned on (which occurs in Core with an env var setting and in Cloud always), any task with a result handler will be checkpointed
- updates cached inputs to gracefully pass in the absence of a result handler
- removes the strict requirement that _all_ Flows have result handlers in favor of a warning

PR #1933, opened 2020-01-21
Add per-task checkpointing opt-out

### LINKED ISSUE #1910: Opt-out of checkpointing on a single Task
## Current behavior
With 0.9.0 Prefect Flows will begin aggressively checkpointing, taking a hint to enable or disable checkpointing based on the existence of a result handler. It may be desired to have select tasks to not checkpoint by specifying `@prefect.task(checkpointing=False)`, however, this option is now ignored.

## Proposed behavior
Allow for select tasks to disable checkpointing with `@prefect.task(checkpointing=False)`. The Flow should continue to checkpoint all other tasks.

wagoodman:

## What does this PR change?
- Adds the ability for a user to specify `checkpoint=False` directly on a task to opt out of storing the result via the result handler (even if checkpointing is enabled globally)
- Enables checkpointing globally in the configuration for flows and tasks.

Closes #1910 

## Why is this PR important?
Both #1904 and #1898 brought a "batteries included" mentality to using result handlers. However, in select cases a user may want to not store a specific result in a given result handler for a single task (maybe due to sensitive information being present, the volume of the result data is prohibitively large, or storing the result is unnecessary and causes more overhead in cleanup). For these cases, it is important to have an opt-out without forcing the user to opt out of result handlers completely (or refactor their flow).


src/prefect/config.toml
```diff
@@ -52,7 +52,7 @@ eager_edge_validation = false
 # If false, only a single execution will occur (no retries, etc.)
 run_on_schedule = true
 # If true, tasks which set `checkpoint=True` will have their result handlers called
-checkpointing = false
+checkpointing = true
```
cicdw: I think we should keep this `False` for now; this will be a large breaking change for anyone using non-JSON compatible `Parameter` values with Core.

src/prefect/core/task.py
```diff
@@ -125,11 +125,9 @@ def run(self, x, y):
         - cache_key (str, optional): if provided, a `cache_key` serves as a unique identifier for this Task's cache, and can
             be shared across both Tasks _and_ Flows; if not provided, the Task's _name_ will be used if running locally, or the
             Task's database ID if running in Cloud
-        - checkpoint (bool, optional, DEPRECATED): if this Task is successful, whether to
-            store its result using the `result_handler` available during the run; defaults to the value of
-            `tasks.defaults.checkpoint` in your user config.  Note that in the future, all tasks with result handlers
-            will be checkpointed.  Also note that checkpointing will only occur locally if `prefect.config.flows.checkpointing` is
-            set to `True`
+        - checkpoint (bool, optional): if this Task is successful, whether to
```
joshmeek: Should the changelog reflect that this is no longer deprecated?

wagoodman: great point, I'll update

joshmeek:
Pending changelog update I'm good on this. Worked great during testing 👍

CHANGELOG.md
```diff
@@ -12,8 +12,9 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Docker daemon reconnect attempts + exit on heartbeat failure -[#1918](https://github.com/PrefectHQ/prefect/pull/1918)
 - More responsive agent shutdown - [#1921](https://github.com/PrefectHQ/prefect/pull/1921)
-- Add per-task checkpointing opt-out - [#1933](https://github.com/PrefectHQ/prefect/pull/1933)
 - Add show_flow_logs to Docker agent [#1929](https://github.com/PrefectHQ/prefect/issues/1929)
+- Add per-task checkpointing opt-out - [#1933](https://github.com/PrefectHQ/prefect/pull/1933)
+- The Task 'checkpoint' kwarg will no longer be depricated to allow opt-out - [#1933](https://github.com/PrefectHQ/prefect/pull/1933)
```
joshmeek:
```suggestion
- The Task 'checkpoint' kwarg will no longer be deprecated to allow opt-out - [#1933](https://github.com/PrefectHQ/prefect/pull/1933)
```

cicdw:
Test failure appears to be non-deterministic thing unrelated to this PR - this also LGTM after the changelog update!

PR #1939, opened 2020-01-22
Build docker image for 'master' branch on merge

### LINKED ISSUE #1924: Let's build / push a master tagged Docker image with each merge to master
This would be useful for QA / testing.

We could tag the image "master" or "unstable", whichever we prefer.

joshmeek:

## What does this PR change?
Closes #1924 

On each merge to master build a new docker image with the `master` tag.

## Why is this PR important?
Gives us an easy way to get unreleased changes into a docker image for testing / development.


.circleci/config.yml
```diff
@@ -331,6 +331,47 @@ jobs:
...
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Build image
+          command: |
+            set -u
+            docker build \
+              --build-arg GIT_SHA=$CIRCLE_SHA1 \
+              --build-arg BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ') \
+              --build-arg PREFECT_VERSION=$CIRCLE_TAG \
```
wagoodman: just a note: there isn't a guarenteed tag, however I think this will still build just fine (the image label will be blank)

joshmeek: Yeah I don't think that's an issue :)

wagoodman: the only difference is in those cases there is a tag guaranteed

joshmeek: Oh I see. Do you think we should ignore trying to provide this argument?

.circleci/config.yml
```diff
@@ -331,6 +331,47 @@ jobs:
...
+      - image: docker
+    parameters:
+      python_version:
+        type: string
+    environment:
+      PYTHON_VERSION: << parameters.python_version >>
+    steps:
+      - checkout
+      - run:
+          name: Master branch check
```
wagoodman: technically this isn't necessary since there isn't a combined tag/branch filter (also doesn't hurt to leave it)

joshmeek: You're right, I'm going to remove it since it doesn't seem necessary to check the same thing twice considering the circle filter already does it

PR #1942, opened 2020-01-22
Test master build image against recently tagged local image

joshmeek:
The build of a `master` tagged image is successful. This PR updates the _Test image_ step to ensure it uses the recently built `master` image for the test 👍

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -364,7 +364,7 @@ jobs:
       - run:
           name: Test image
           command: |
-            docker run -dit prefecthq/prefect /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
+            docker run -dit prefecthq/prefect:master /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
       - run:
           name: Push master tag
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -364,7 +364,7 @@ jobs:
       - run:
           name: Test image
           command: |
-            docker run -dit prefecthq/prefect /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
+            docker run -dit prefecthq/prefect:master /bin/bash -c 'curl -fL0 https://raw.githubusercontent.com/PrefectHQ/prefect/master/examples/retries_with_mapping.py | python'
```
wagoodman: good catch! This suggests that [this line](https://github.com/PrefectHQ/prefect/pull/1942/files#diff-1d37e48f9ceff6d8030570cd36286a61L318) is missing a similar tag (`${CIRCLE_TAG}-${PYTHON_TAG}`) ?

joshmeek: Which of these two tags do you think is a better fit, since the tag based steps use both:
```
prefecthq/prefect:${CIRCLE_TAG}-${PYTHON_TAG} 
prefecthq/prefect:latest-${PYTHON_TAG}
```

wagoodman: IMHO, `${CIRCLE_TAG}-${PYTHON_TAG}` is more explicit, but is seems that either will do

PR #1958, opened 2020-01-28
Decrease brittleness of timeout tests 

### LINKED ISSUE #1947: Investigate nondeterministic timeout test failure
https://circleci.com/gh/PrefectHQ/prefect/54780?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

```
________________ test_timeout_actually_stops_execution[mthread] ________________

executor = <Executor: DaskExecutor>

    @pytest.mark.skipif(
        sys.platform == "win32", reason="Windows doesn't support any timeout logic"
    )
    @pytest.mark.parametrize("executor", ["local", "sync", "mthread"], indirect=True)
    def test_timeout_actually_stops_execution(executor):
        with tempfile.TemporaryDirectory() as call_dir:
            FILE = os.path.join(call_dir, "test.txt")
    
            @prefect.task(timeout=1)
            def slow_fn():
                "Runs for 1.5 seconds, writes to file 7 times"
                iters = 0
                while iters < 6:
                    time.sleep(0.25)
                    with open(FILE, "a") as f:
                        f.write("called\n")
                    iters += 1
    
            flow = Flow("timeouts", tasks=[slow_fn])
            state = flow.run(executor=executor)
    
            # if it continued running, would run for 1 more second
            time.sleep(0.5)
>           with open(FILE, "r") as g:
E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpzlrv1lk1/test.txt'

tests/core/test_flow.py:2616: FileNotFoundError
```

wagoodman:
Currently the timeout tests fail sporadically in CircleCI. These changes are to reduce the brittleness of the Task timeout tests. 

Closes #1947

### DIFF
--- a/tests/core/test_flow.py
+++ b/tests/core/test_flow.py
@@ -2546,15 +2546,22 @@ def test_timeout_actually_stops_execution(executor):
         def slow_fn():
             with open(FILE, "w") as f:
                 f.write("called!")
-            time.sleep(3)
+            time.sleep(2)
+            with open(FILE, "a") as f:
+                f.write("invalid")
 
         flow = Flow("timeouts", tasks=[slow_fn])
 
         assert not os.path.exists(FILE)
 
+        start_time = time.time()
         state = flow.run(executor=executor)
+        stop_time = time.time()
+        time.sleep(max(0, 3 - (stop_time - start_time)))
 
         assert os.path.exists(FILE)
+        with open(FILE, "r") as f:
+            assert "invalid" not in f.read()
 
     assert state.is_failed()
     assert isinstance(state.result[slow_fn], TimedOut)
--- a/tests/core/test_flow.py
+++ b/tests/core/test_flow.py
@@ -32,6 +32,7 @@ from prefect.engine.state import (
     State,
     Success,
     TriggerFailed,
+    TimedOut,
 )
 from prefect.schedules.clocks import ClockEvent
 from prefect.tasks.core.function import FunctionTask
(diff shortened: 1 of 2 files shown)


tests/core/test_flow.py
```diff
@@ -2524,26 +2525,37 @@ def test_load_accepts_name_and_sluggified_name(self):
...
+
+        assert not os.path.exists(FILE)
+
         state = flow.run(executor=executor)
 
-        # if it continued running, would run for 1 more second
-        time.sleep(0.5)
-        with open(FILE, "r") as g:
-            contents = g.read()
+        assert os.path.exists(FILE)
```
cicdw: So with this refactor, the test is no longer testing what it claims to be testing (you're now testing that a timeout occurred but the sleep could still be occurring in the background).  If you think it's too brittle, we might need to consider removing it altogether.

tests/engine/test_task_runner.py
```diff
@@ -308,28 +308,38 @@ def test_task_runner_accepts_dictionary_of_edges():
...
+
+        assert not os.path.exists(FILE)
 
         state = TaskRunner(slow_fn).run(executor=executor)
 
-        # if it continued running, would run for 1 more second
-        sleep(0.5)
-        with open(FILE, "r") as g:
-            contents = g.read()
+        assert os.path.exists(FILE)
```
cicdw: Same comment as above

tests/core/test_flow.py
wagoodman: I would be OK with removing it, but if timeouts are staying it's a good test to try and keep. I've updated the test to include a footer that should never appear in the file (only added if the task continues).

cicdw: Awesome - good idea!

PR #1960, opened 2020-01-29
Slight refactor of how agents deploy and submit flow runs for execution

### LINKED ISSUE #1956: Agents Should not Fail flow runs due to version mismatches
Suppose I have two agents running, and they both pick up the same flow run.  Whichever one submits _second_ will experience an error like:
```
[{'message': 'State update failed for flow run ID 5073a4ea-61dd-400c-892a-294b93ac3633: 
provided version 1 but current version is 2.', 'locations': [{'line': 4, 'column': 13}], 'path': 
['setFlowRunStates', 'states', 0, 'id'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR'}}]
```
and then [mark the Flow Run as Failed](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/agent/agent.py#L209-L228).  This is incorrect behavior though, because another agent did in fact successfully submit the flow run for work, which will then proceed shortly.  

We should do a check on the exception to match against `"State update failed"` and in that case, _not_ update the state at all.

**cc:** @jlowin @zdhughes

cicdw:

## What does this PR change?
This PR does two things:
- puts Flow Runs in submitted states _prior_ to deploying them; this ensures that jobs / resources aren't created for Flow Runs which have already been submitted by other agents.  Putting them in Submitted _first_ allows Cloud to version lock appropriately
- in the event that placing the Flow Run into a Submitted state fails, we no longer attempt to mark the flow run failed; instead we log the error and continue on our way

## Why is this PR important?
Closes #1956

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -209,16 +209,19 @@ class Agent:
         try:
             self.update_state(flow_run)
             deployment_info = self.deploy_flow(flow_run)
-            self.client.write_run_logs(
-                [
-                    dict(
-                        flowRunId=getattr(flow_run, "id", "UNKNOWN"),  # type: ignore
-                        name=self.name,
-                        message="Submitted for execution: {}".format(deployment_info),
-                        level="INFO",
-                    )
-                ]
-            )
+            if getattr(flow_run, "id", None):
+                self.client.write_run_logs(
+                    [
+                        dict(
+                            flowRunId=getattr(flow_run, "id"),  # type: ignore
+                            name=self.name,
+                            message="Submitted for execution: {}".format(
+                                deployment_info
+                            ),
+                            level="INFO",
+                        )
+                    ]
+                )
         except Exception as exc:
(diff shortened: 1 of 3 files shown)


src/prefect/agent/agent.py
```diff
@@ -207,9 +207,23 @@ def deploy_and_update_flow_run(self, flow_run: "GraphQLResult") -> None:
         """
         # Deploy flow run and mark failed if any deployment error
         try:
+            self.update_state(flow_run)
             deployment_info = self.deploy_flow(flow_run)
-            self.update_state(flow_run, deployment_info)
+            self.client.write_run_logs(
+                [
+                    dict(
+                        flowRunId=getattr(flow_run, "id", "UNKNOWN"),  # type: ignore
```
joshmeek: I believe this will have an error when `UNKNOWN` and then the catch will still try to write a log again w/ `UNKNOWN` and fail.

cicdw: @joshmeek updated

PR #1961, opened 2020-01-29
Attempt to ship logger errors via the CloudHandler

### LINKED ISSUE #1943: Attempt to log possible log writing errors
## Current behavior
Currently if the logger encounters an error it isn't immediately surfaced to Cloud users what the reason is for the logging error.

## Proposed behavior
Something along the lines of (but not limited to):
```python
try:
   write_some_logs()
except Exception as exc:
    sleep(1)
    write_this_error_log(exc)
except Exception as exc2:
    raise exc    # raise exception as it normally would
```

Where if the writing of the logs fails then the exception is passed as an attempted log. Otherwise if the exception logging fails then raise as it normally would. This would give some nice insight into "missing" logs.

## Example
Possible issues such as non serializeable bytes, large payloads, etc. could be surfaced

```
self.logger.info(b"bad bytes string")
CRITICAL - CloudHandler | Failed to write log with error: Object of type bytes is not JSON serializable
```

```
self.logger.info("LARGE STRING...........")
CRITICAL - CloudHandler | Failed to write log with error: 413 Client Error: Request Entity Too Large for url: https://api.prefect.io/graphql/alpha
```

joshmeek:

## What does this PR change?
Adds a step to the CloudHandler to attempt to log possible logger errors to Cloud for visibility. For example if the payload is too large or something non-json serializable is attempted the error will be raised in the flow process but not elevated for visibility.

Example logger issue being elevated as a cloud log:
image 

Closes #1943 

## Why is this PR important?
Improves user visibility into otherwise hidden logger exceptions.

### DIFF
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -142,7 +142,6 @@ class CloudHandler(logging.StreamHandler):
     def _make_error_log(self, message: str) -> dict:
         log = dict()
         log["flowRunId"] = prefect.context.get("flow_run_id", None)
-        log["taskRunId"] = prefect.context.get("task_run_id", None)
         log["timestamp"] = pendulum.from_timestamp(time.time()).isoformat()
         log["name"] = self.logger.name
         log["message"] = message
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -70,7 +70,7 @@ class CloudHandler(logging.StreamHandler):
                 try:
                     assert self.client is not None
                     self.client.write_run_logs([self._make_error_log(message)])
-                except prefect.utilities.exceptions.ClientError as exc:
+                except Exception as exc:
                     self.logger.critical("Unable to write logs to Prefect Cloud")
 
     def _monitor(self) -> None:
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -63,9 +63,15 @@ class CloudHandler(logging.StreamHandler):
                 assert self.client is not None
                 self.client.write_run_logs(logs)
             except Exception as exc:
-                self.logger.critical(
-                    "Failed to write log with error: {}".format(str(exc))
-                )
+                message = "Failed to write log with error: {}".format(str(exc))
+                self.logger.critical(message)
+
+                # Attempt to write batch error log otherwise log invalid cloud communication
+                try:
+                    assert self.client is not None
+                    self.client.write_run_logs([self._make_error_log(message)])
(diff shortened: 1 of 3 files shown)


src/prefect/utilities/logging.py
```diff
@@ -92,7 +98,10 @@ def put(self, log: dict) -> None:
             json.dumps(log)  # make sure the payload is serializable
             self.queue.put(log)
         except TypeError as exc:
-            self.logger.critical("Failed to write log with error: {}".format(str(exc)))
+            message = "Failed to write log with error: {}".format(str(exc))
+            self.logger.critical(message)
+
+            self.queue.put(self._make_error_log(message))
```
cicdw: Could this result in an infinite loop if, heaven forbid, Cloud was completely down for an extended period of time?

src/prefect/utilities/logging.py
```diff
@@ -125,7 +134,21 @@ def emit(self, record) -> None:  # type: ignore
             log["info"] = record_dict
             self.put(log)
         except Exception as exc:
-            self.logger.critical("Failed to write log with error: {}".format(str(exc)))
+            message = "Failed to write log with error: {}".format(str(exc))
+            self.logger.critical(message)
+
+            self.put(self._make_error_log(message))
```
cicdw: Same question as above

src/prefect/utilities/logging.py
joshmeek: Cloud errors like 400s are surfaced in the above handle. Where an extra check happens  and logs `Unable to write logs to Prefect Cloud`

joshmeek: Let me test this again but I believe that's what should happen

joshmeek:
@cicdw Tested this some more and we should be good to go 👍 the only potential _issue_ is the batch upload error logging part:
```python
    def batch_upload(self) -> None:
        logs = []
        try:
            while True:
                log = self.queue.get(False)
                logs.append(log)
        except Empty:
            pass

        if logs:
            try:
                assert self.client is not None
                self.client.write_run_logs(logs)
            except Exception as exc:
                message = "Failed to write log with error: {}".format(str(exc))
                self.logger.critical(message)

                # Attempt to write batch error log otherwise log invalid cloud communication
                try:
                    assert self.client is not None
                    self.client.write_run_logs([self._make_error_log(message)])
                except Exception as exc:
                    self.logger.critical("Unable to write logs to Prefect Cloud")
```

If there's an issue with one of the logs in this part e.g. `Failed to write log with error: 413 Client Error: Request Entity Too Large` then since we are writing all of the logs at once they all fail. Thoughts?

PR #1963, opened 2020-01-30
Add healthchecks to other storage options

### LINKED ISSUE #1934: Run some of the "Docker" healthchecks for all flow registrations
Right now we run some "healthchecks" on Flows at registration time to help users debug errors early on; some of these checks are docker specific (e.g., version mismatches) but some are valid for all storage types, in particular the result handler / retries / checkpointing checks.  We should run these for all Flow registrations to raise errors early.

joshmeek:

## What does this PR change?
Closes #1934
Adds a basic healthcheck pass for Local, GCS, S3, and Azure storage options. 

## Why is this PR important?
Previously the healthchecks were only run on the Docker storage option but some of these could be used for other types.


tests/environments/storage/test_local_storage.py
```diff
@@ -139,3 +139,17 @@ def test_add_flow_with_weird_name_is_cleaned():
...
         assert " " not in loc
         assert "~" not in loc
+
+
+def test_build_healthchecks():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        s = Local(directory=tmpdir)
+        flow = Flow("TestFlow")
+        s.add_flow(flow)
+        assert s.build()
```
cicdw: This test doesn't seem to test anything related to healthchecks?

src/prefect/environments/storage/base.py
```diff
@@ -110,3 +111,13 @@ def serialize(self) -> dict:
...
+
+    def run_basic_healthchecks(self) -> None:
+        """
+        Runs basic healthchecks on the flows contained in this Storage class
+        """
+        if not hasattr(self, "_flows"):
+            return
+
+        _healthcheck.result_handler_check(self._flows.values())  # type: ignore
+        _healthcheck.environment_dependency_check(self._flows.values())  # type: ignore
```
cicdw: So I'm wondering if we should keep this one or not; there are edge cases where this isn't necessarily required:
- register flow w/ S3 storage off laptop
- run flow using agent running on remote machine with all dependencies

Not totally against it, as it's a good reminder for users, but just wanted to call it out

wagoodman: I agree, we'd want to allow folks to register the flow in locations where the dependencies don't necessarily exist. However, it would be nice to at least see if there could be a problem in other cases. We could modify the `environment_dependency_check` to instead output a warning instead of erroring out (add a flag to indicate warn-only).

tests/environments/storage/test_local_storage.py
wagoodman: a little indirect, but it is invoking the new healthcheck method (https://github.com/PrefectHQ/prefect/blob/storage_healthchecks/src/prefect/environments/storage/local.py#L118)

src/prefect/environments/storage/base.py
joshmeek: I think this _is_ necessary because for example if you're registering your flow w/ S3 storage then you need to have the dependencies for S3 storage anyway in order to upload.

tests/environments/storage/test_local_storage.py
joshmeek: Yeah it is indirect. This tests that the build passes w/ no healthcheck errors

src/prefect/environments/storage/base.py
wagoodman: I may be looking in the wrong spots, but I think this would never check for an s3 dependency (or any storage dependency) since `flow.environment.dependencies` are [usually returning hard coded lists](https://github.com/PrefectHQ/prefect/blob/storage_healthchecks/src/prefect/environments/execution/dask/k8s.py#L99). In this example, I would require the `kubernetes` package to be installed, though it's never used during registration.

joshmeek: Oh yep you're right I was confused when I wrote this. I was mixing up environments and storage!

PR #1972, opened 2020-02-03
Attach Task/Flow identity to LogRecords

### LINKED ISSUE #1938: Add Flow and Task identifying information to LogRecords
## Current behavior
- All log entries have the default [LogRecord attributes](https://docs.python.org/3/library/logging.html#logrecord-objects)... 
- There is no universally identifiable flow slug

## Proposed behavior
...It would be useful to attach Core-specific information to log entries:
- `task_name` & `task_slug`: available only on entries emitted from Tasks and TaskRunners
- `flow_name` & `flow_slug`: available only on entries emitted from Flows and FlowRunners

The addition of the `flow_slug` makes is possible to uniquely identify flows other than by name.

## Example
One can now take a single stream of log entries during a local flow run and separate it by entity without parsing the `LogRecord.msg` field (which is unreliable). This makes is possible to build richer feedback interfaces for filtering, organizing, and otherwise selectively manipulating log entries.

Here is an example CLI interface that would be made possible by theses features:

local-progress

(Given the following flow):
```python
from utils import seed, random_data
import prefect

@prefect.task
def extract():
    return seed.get()

@prefect.task
def transform(seed):
    data = []
    for x in range(10):
        c_data = random_data(seed)
        logger.info("debug info: {}".format(c_data))
        data.append(c_data)
    return data

@prefect.task
def load(data):
    # do something with the data....
    pass

with prefect.Flow("ETL") as flow:
    e = extract()
    t = transform(e)
    load(t)
```

wagoodman:

## What does this PR change?
Adds identifying information about the Task/Flow source to LogRecords generated from log events.

## Why is this PR important?
This will allow for selecting log records dynamically, opening up the door for interesting terminal behaviors (see the original [issue](https://github.com/PrefectHQ/prefect/issues/1938) for an example).

Closes #1938


src/prefect/utilities/logging.py
```diff
@@ -128,6 +139,27 @@ def emit(self, record) -> None:  # type: ignore
...
+        - **kwargs: keyword arguments to pass to the original LogRecord Factory
+
+    Returns:
+        - logging.LogRecord: the newly created LogRecord
+    """
+    record = _original_log_record_factory(*args, **kwargs)
+
+    for attr in PREFECT_LOG_RECORD_ATTRIBUTES:
+        value = prefect.context.get(attr, None)
+        if attr:
```
cicdw: Should this be `if value:`?

wagoodman: 😳 yup

PR #1973, opened 2020-02-03
New heartbeats

cicdw:

## What does this PR change?
This PR:
- introduces a CLI interface for sending heartbeats
- uses `subprocess.Popen` to call the CLI for sending heartbeats during runs
- removes cancellation hooks (because all looping no longer occurs within the same process)

## Why is this PR important?
This PR should fix the zombie issues that we have routinely seen with the `DaskKubernetesEnvironment`.  The reasoning is that the old heartbeat thread could not fire if the user's task held onto the GIL, and might have even resulted in threading deadlocks.  The new implementation avoids the GIL entirely and runs in a very lightweight `Popen` call.

I would really like to QA this across our environments / flow sizes to stress test my implementation but I'm confident this will be strictly better.


joshmeek:
@cicdw I can write the CLI tests if you want

cicdw:
@joshmeek yea that would be helpful!

PR #1976, opened 2020-02-03
Pass proper auth token to FargateTaskEnvironment execute tasks

joshmeek:

## What does this PR change?
Bug fix where proper auth token was not being passed to the task ran in the FargateTaskEnvironment's execute step

## Why is this PR important?
Bug fixes are important :P

### DIFF
--- a/src/prefect/environments/execution/fargate/fargate_task.py
+++ b/src/prefect/environments/execution/fargate/fargate_task.py
@@ -254,7 +254,7 @@ class FargateTaskEnvironment(Environment):
                 "environment": [
                     {
                         "name": "PREFECT__CLOUD__AUTH_TOKEN",
-                        "value": config.cloud.agent.auth_token,
+                        "value": config.cloud.auth_token,
                     },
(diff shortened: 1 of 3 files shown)


src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -254,7 +254,7 @@ def execute(  # type: ignore
                 "environment": [
                     {
                         "name": "PREFECT__CLOUD__AUTH_TOKEN",
-                        "value": config.cloud.agent.auth_token,
+                        "value": config.cloud.auth_token,
```
cicdw: to prevent this from causing future issues, could we make this more robust via:
```suggestion
                        "value": config.cloud.agent.auth_token or config.cloud.auth_token,
```
or whichever order you prefer?

joshmeek: Good call

wagoodman: also, we should add a case to ensure that execute responds to `cloud.agent.auth_token` as well

PR #1989, opened 2020-02-06
Point client toward api.prefect.io

zdhughes:

## What does this PR change?

This PR points clients toward the `api.prefect.io`, rather than `api.prefect.io/graphql/alpha`. Functionally, everything remains the same.

## Why is this PR important?

Our API's definitely past the alpha stage, and it's good to note that before release.


cicdw:
What's the idea here - don't we _want_ to be hitting a versioned API from Core just as a safeguard?

zdhughes:
The thought was that pointing users toward an endpoint denoting alpha could potentially be confusing, and that versioned API behavior would probably be more cleanly resolved using the `X-PPREFECT-CORE-VERSION` header than requiring users to manually manage their graphql endpoint.

src/prefect/environments/execution/dask/k8s.py
```diff
@@ -423,7 +423,7 @@ def _populate_scheduler_spec_yaml(
 
         # Required Cloud environment variables
         env_values = [
-            {"name": "PREFECT__CLOUD__GRAPHQL", "value": prefect.config.cloud.graphql},
+            {"name": "PREFECT__CLOUD__GRAPHQL", "value": prefect.config.cloud.api},
```
jlowin: Why still set the env var at all, if nothing points at it? Should this be `PREFECT__CLOUD__API`?

zdhughes: Very good point! There were a lot of values to update, and I got caught up in chugging through them and missed this. Let me confirm there are no additional ones like this.

jlowin: However, -- is this a breaking change? If my agent is running 0.9.2+ but my docker container has 0.9.1-, will it still run without the old value?

zdhughes: Looking at some of our `deploy_flow` methods, I think it could be. Care to opine, @cicdw / @joshmeek ?

cicdw: We should make sure to QA this but I don't think this will be a breaking change.  We have been interpolating the `graphql` route from the `api` config setting [since 0.5.0](https://github.com/PrefectHQ/prefect/commit/c9081d18feee2a352a26a44787c1814832c0766f) and so setting the API config should be sufficient for Flows running old versions as well.

src/prefect/client/client.py
```diff
@@ -287,6 +287,7 @@ def _request(
             raise ValueError("Invalid method: {}".format(method))
 
         # Check if request returned a successful status
+        print(response.status_code)
```
joshmeek:
```suggestion
```

PR #1994, opened 2020-02-07
Always use the same run function for running on schedule and off

### LINKED ISSUE #1430: Unify on- and off- schedule run pipelines
Currently, `flow.run()` has two logic branches:
- if a flow doesn't have a schedule, we create a flowrunner and call it's run() method.
- if a flow does have a schedule, we call the `_run_on_schedule` method which does a number of additional logic.

I would like to combine these into a single branch, because having two different pipelines leads to the potential for different behaviors depending on how you're running a flow (an issue that plagues Airflow). For example in #1429 I realized that tasks results are returned when running on schedule, but not returned when running off schedule.

A solution would be to simply create a "one-off" schedule and call `_run_on_schedule()` no matter what, with the caveat that the schedule would need to be removed so it didn't overwrite the original flow schedule.

jlowin:

## What does this PR change?
This is a minor but possibly important change and closes #1430. 

Previously, if `flow.run(run_on_schedule=False)` was called, then a `FlowRunner` was instantiated and and executed once, visiting each task only once. If `flow.run(run_on_schedule=True)` was called, then a much more rich set of logic was invoked and the flow itself was run to completion -- even if it didn't have a schedule attached. This includes task retries and other features only made possible by calling a `FlowRunner` multiple times. If the flow has a schedule, it also sleeps until the next run. 

This means the following experiences were actually very different:
1. a flow with no schedule that starts running right now: the user calls `flow.run()` and gets the benefits of retries (among other things)
2. a flow with a schedule that starts running right now: the user must call `flow.run(run_on_schedule=False)` and gets no retries

This PR unifies the two routes, so that no matter whether a flow has a schedule or not, it gets the "full" treatment. `run_on_schedule=False` can still be passed to start a run immediately. 

The old behavior (of running each task only once) can be recovered by using a `FlowRunner` directly.

## Why is this PR important?
To avoid the old Airflow issue where backfills and scheduled runs used completely different logic pipelines


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 ### Enhancements
 
 - Improve error handling for unsupported callables - [#1993](https://github.com/PrefectHQ/prefect/pull/1993)
+- Ensure all calls to `flow.run()` use the same execution logic - [#1994](https://github.com/PrefectHQ/prefect/pull/1994)
```
cicdw: Isn't this also a breaking change for `flow.run` now that any retries will actually be respected instead of returned?

wagoodman: Certainly new behavior, but is it breaking?

cicdw: I'd call it breaking - anyone who uses `flow.run()` currently will have things like `Retrying` states immediately returned.  With the proposed changes this call will block until all of the retries have occurred.  I don't think it's a _massive_ deal but given that this is the main entrypoint for open source users I'm a bit nervous about it.

jlowin: Sorry this has sat for a while - would you like me just to move this to the `breaking` section of the changelog, or take a larger step to introduce the behavior?

cicdw: @jlowin yea could you add it to the breaking changes section?

lauralorenz:
@cicdw I changed the base for this PR to be our 0.11.0 release branch. If you can give it another look as discussed in release planning then we can merge it there.

cicdw:
This LGTM - what should we do about the CHANGELOG entry @joshmeek / @lauralorenz ?

joshmeek:
> This LGTM - what should we do about the CHANGELOG entry @joshmeek / @lauralorenz ?

Let's merge it into the release branch then move it up after

PR #2000, opened 2020-02-08
S3 storage: accept additional boto3 kwargs

trapped:

## What does this PR change?

Add a `client_options` optional dict parameter to `prefect.environments.storage.s3.S3` for passing additional parameters when instantiating the `boto3` client; also serialize it as `JSONCompatible` in `prefect.serialization.storage.S3Schema`.

## Why is this PR important?

Passing additional parameters to the `boto3` client (like `endpoint_url` and `use_ssl`) allows storing and fetching Flows from any S3-compatible service, for example a local Minio cluster or alternative services ([list of S3-compatible and competing services on Wikipedia](https://en.wikipedia.org/wiki/Amazon_S3#S3_API_and_competing_services)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 ### Enhancements
 
 - Improve error handling for unsupported callables - [#1993](https://github.com/PrefectHQ/prefect/pull/1993)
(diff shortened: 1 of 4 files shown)


cicdw:
The new serializers are now live in Cloud so anyone who wants to use these new kwargs should be good to go!

PR #2010, opened 2020-02-11
Extend prefect logger and fix literal eval on cpu and memory

braunreyes:

This work has an enhancement and bug fix:

Enhancement:
Adding `extra_loggers` to logging configuration.  This allows user to specify a list of loggers to pass through the Prefect logger configuration so that non-Prefect logs like those for boto3, snowflake.connector or a custom library show up in local stdout and in Prefect Cloud.

BugFix:
the literal eval on the fargate kwargs was breaking when defining cpu and memory because they need to be strings.  I added a list of task definition kwargs to not eval.

- [X] updates `CHANGELOG.md` (if appropriate)

## What does this PR change?

This puts the prefect logger configuration into a new internal function `_create_logger' and has `configure_logging` call this new function with same parameters for backwards compatibility.  Then added `configure_extra_loggers` function that gets called and maps against new logging configuration `extra_loggers` so that non-Prefect logs can go to stdout for local and up to Prefect Cloud via the CloudHandler.

Also fixed the issue with literal_eval on cpu and memory by adding a new kwargs list `definition_kwarg_list_no_eval = ["cpu", "memory"]`.  This list is used to create dictionary called `definition_kwarg_list_eval` which has the kwarg as a key and True/False as value to determine if we should literal_eval or not.  This results in cpu and memory with a value of False.

This dictionary is then used in an if statement to determine if we should literal_eval of not, if False the original kwargs value is added to `task_definition_kwargs`

## Why is this PR important?

This fixes a breaking change in 0.9.3 for the Fargate Agent and allows better logging for non-Prefect logs locally and live which aids in troubleshooting and validation


braunreyes:
I will add to changelog later tonight

docs/core/concepts/logging.md
```diff
@@ -51,4 +51,34 @@ def my_task():
     logger.warning("A warning message.")
 ```
 
+### Extra Loggers
+
+Many libraries like `boto3` and `snowflake.conector` are setup to emit their own internal logs when configured 
+with a logging configuration.  You may even have an internal shared library for your team with the same functionality.
+
+If you are doing this via the standard `logging` library you might do this:
+```
```
joshmeek:
```suggestion
```python
```

joshmeek: Simply for syntax highlighting

test-requirements.txt
```diff
@@ -3,3 +3,4 @@ testfixtures >= 6.10.3, < 7.0
 pytest-cov >= 2.6, < 3.0
 pytest-env >= 0.6.0, < 0.7.0
 pytest-xdist >= 1.23, < 2.0
+pi
```
joshmeek:
```suggestion
```

docs/cloud/agents/fargate.md
```diff
@@ -198,6 +198,35 @@ flow_id = flow.register(project_name="<YOUR_PROJECT>")
...
+            "Resource": "<s3 bucket root>"
+        },
+        {
+            "Sid": "AllowGetPutDeleteWorkObject",
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObjectVersion",
+                "s3:GetObject"
+            ],
+            "Resource": "<s3 bucket kwargs path"
```
joshmeek:
```suggestion
            "Resource": "<s3 bucket kwargs path>"
```

joshmeek:
@braunreyes what exactly was the issue you were running into with CPU and Memory not being parsed (i.e. what and how were you passing it into the Agent)? I though it had a handle possible Value and Syntax errors that may arise from the `literal_eval`.

braunreyes:
@joshmeek 

The issue is with boto3. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.register_task_definition

The cpu and memory keys for the task definition outside of the containerDefinitions list are expected to be strings.  `ast.literal_eval` turns them to ints.

We pass these in as external kwargs...though considering we eval everything it would be an issue for envars and client kwargs as well.

docs/core/concepts/logging.md
braunreyes: change completed

test-requirements.txt
braunreyes: change completed

docs/cloud/agents/fargate.md
braunreyes: change completed

joshmeek:
Tested fargate agent changes as well as the extra logger enhancement ✅ 
image

cc @cicdw One thing I noticed is that setting `extra_loggers` in my config.toml worked fine but providing the `-e` env var to the agent didn't. 🤔 probably unrelated to this but worth pointing out that there may be a bug in that CLI option

PR #2022, opened 2020-02-13
change to fargate agent to fix bug related to agent process threadpool

braunreyes:

- [X] updates `CHANGELOG.md` (if appropriate)

## What does this PR change?

Moving task_definition_name from class attribute to dict value, which is explicitly passed into fargate agent class methods.

## Why is this PR important?

This fixes a new bug with Fargate Agent when task revisions are enabled caused by the new ThreadPool based agent process.  Previously because the loop over flow runs was synchronous we could house the task definition name as an agent class attribute.  Now that the loop as async when there are more than 1 flows runs collected by the agent, the second flow run overrides the task definition name set by the first flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,8 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Ensure `ifelse` casts its condition to `bool` prior to evaluation - [#1991](https://github.com/PrefectHQ/prefect/pull/1991)
 - Do not perform `ast.literal_eval` on cpu and memory task_definition kwargs for Fargate Agent - [#2010](https://github.com/PrefectHQ/prefect/pull/2010)
-- Fix new agent processing with Threadpool causing problem for Fargate Agent with task revisions enabled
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,7 +25,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Ensure `ifelse` casts its condition to `bool` prior to evaluation - [#1991](https://github.com/PrefectHQ/prefect/pull/1991)
 - Do not perform `ast.literal_eval` on cpu and memory task_definition kwargs for Fargate Agent - [#2010](https://github.com/PrefectHQ/prefect/pull/2010)
-
+- Fix new agent processing with Threadpool causing problem for Fargate Agent with task revisions enabled
```
joshmeek:
```suggestion
- Fix new agent processing with Threadpool causing problem for Fargate Agent with task revisions enabled - [#2022](https://github.com/PrefectHQ/prefect/pull/2022)

```

PR #2031, opened 2020-02-14
Update CronClock to respect microseconds and unbound croniter requirement

### LINKED ISSUE #1892: Investigate clock test failure
https://circleci.com/gh/PrefectHQ/prefect/52671

```
================================== FAILURES ===================================
_____________ TestCronClock.test_cron_clock_respects_microseconds _____________
self = <tests.schedules.test_clocks.TestCronClock object at 0x0000008C18CFE9C8>
    def test_cron_clock_respects_microseconds(self):
        every_day = "0 0 * * *"
        start_date = pendulum.datetime(2050, 1, 1, 0, 0, 0, 1)
        c = clocks.CronClock(every_day, start_date=start_date)
>       assert islice(c.events(), 1) == [pendulum.datetime(2050, 1, 2)]
E       AssertionError: assert [DateTime(205...ezone('UTC'))] == [DateTime(205...ezone('UTC'))]
E         At index 0 diff: DateTime(2050, 1, 1, 0, 0, 0, tzinfo=Timezone('UTC')) != DateTime(2050, 1, 2, 0, 0, 0, tzinfo=Timezone('UTC'))
E         Use -v to get the full diff
tests\schedules\test_clocks.py:298: AssertionError
```

joshmeek:

## What does this PR change?
Updates the CronClock to respect microseconds. In previous versions of croniter microseconds were respected by adding a second to the datetime. After a recent release they changed the behavior to round down.

For example (using a daily run):
With start date of `(2050, 1, 1, 0, 0, 0, 1)` in previous versions, the next scheduled time would be `(2050, 1, 2, 0, 0, 0, 0)`. After the recent croniter update it would return `(2050, 1, 1, 0, 0, 0, 0)`. This is fundamentally different and I think we should respect microseconds on the start date.

There is one potential downside with this implementation. In running on older versions of croniter the start date will have a second added to it twice. Once by croniter itself and once with the logic here. I am open for discussion on if this is a bad pattern here but I would rather respect microseconds with an extra second increment on old versions over not respecting microseconds on start date. The _upside_ on this is that cron does not allow sub-minute precision so this extra second increment shouldn't really affect anything.

Closes #1892 

## Why is this PR important?
We can unbound the croniter requirement


src/prefect/schedules/clocks.py
```diff
@@ -264,6 +268,11 @@ def events(self, after: datetime = None) -> Iterable[ClockEvent]:
                 microsecond=after.microsecond,
             )
         )
+
+        # Respect microseconds by rounding up
+        if after_localized.microsecond:
+            after_localized = after_localized + timedelta(seconds=1)
```
cicdw: You could compute the `timedelta` offset by directly inspecting the version of `croniter` running if you want to be targeted / precise here per your comment.

joshmeek: Great idea! I'll add that in

joshmeek: Do you know of a safe way to interpret version from a package that doesn't have `__version__`? croniter doesn't seem to have that 🤔

cicdw: Oh boy.  There are ways that involve metaprogramming but might not be worth it.  Here's a code snippet if you do want to take this route:

```python
import pkg_resources

[p.version for p in pkg_resources.working_set if p.project_name.lower() == "croniter"]
# returns ['0.3.25'] for me
```

Issue #1821, opened 2019-12-10
Can't `map` over `set`


bnaul:
Example:
```
from prefect import task, Flow

@task
def double(x):
    return 2 + x

with Flow('set_map') as flow:
    set_double = double.map({1, 2})  # double.map([1, 2]) works

flow.run()
```
Currently fails on `0.7.3` with:
```
[2019-12-10 07:01:48,973] ERROR - prefect.TaskRunner | Task 'double': unexpected error while running task: TypeError("'set' object is not subscriptable")
Traceback (most recent call last):
  File "/Users/brett/Dropbox/Documents/model/.venv/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 260, in run
    executor=executor,
  File "/Users/brett/Dropbox/Documents/model/.venv/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 720, in run_mapped_task
    upstream_state.result[i],
TypeError: 'set' object is not subscriptable
ERROR:prefect.TaskRunner:Task 'double': unexpected error while running task: TypeError("'set' object is not subscriptable")
Traceback (most recent call last):
  File "/Users/brett/Dropbox/Documents/model/.venv/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 260, in run
    executor=executor,
  File "/Users/brett/Dropbox/Documents/model/.venv/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 720, in run_mapped_task
    upstream_state.result[i],
TypeError: 'set' object is not subscriptable

<Failed: "Some reference tasks failed.">
```
Maybe this is expected behavior but I would have thought it would just work as long as the thing being mapped has a `len` (even if the output wasn't guaranteed to stay a `set`).

cicdw:
Sorry for the delayed response here; the biggest reason that non-ordered collections don't play well with mapping is that in Cloud (and also somewhat in Core) we identify mapped task runs as a pair (task_run_id, map_index) where the map index corresponds to the position of the argument being mapped over.  This becomes more important in situations such as retries when we need to make sure we aren't mismatching state updates. 

There's a decent chance we'll be refactoring mapping in the near future for performance reasons, and when we do we can revisit some of these assumptions and make sure they aren't overly stringent!

bnaul:
Makes sense :) Maybe there's a different error that could be raised that makes it clearer that the behavior is intentional? Regardless feel free to close if you think this won't be addressed except as part of a larger refactor

PR #2032, opened 2020-02-14
Add early check for mapping argument type errors

wagoodman:

## What does this PR change?
Adds a better exception around unsubscriptable mapped arguments, both within a Flow context and at runtime. Until there is a large mapping refactor, this only addresses the message raised by #1821 .

## Why is this PR important?
Brings forth mapping argument Type issues as early as possible.

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -22,6 +22,7 @@ import prefect.engine.cache_validators
 import prefect.engine.signals
 import prefect.triggers
 from prefect.utilities import logging
+from prefect.utilities.tasks import unmapped
 from prefect.utilities.notifications import callback_factory
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -717,6 +717,12 @@ def run_mapped_task(
                         # Therefore, we only try to get a result if EITHER this task's
                         # state is not already mapped OR the upstream result is not None.
                         if not state.is_mapped() or upstream_state._result != NoResult:
+                            if not hasattr(upstream_state.result, "__getitem__"):
+                                raise TypeError(
+                                    "cannot map over an unsubscriptable object ({})".format(
```
cicdw:
```suggestion
                                    "Cannot map over an unsubscriptable object ({})".format(
```

src/prefect/core/task.py
```diff
@@ -475,6 +476,11 @@ def map(
         Returns:
             - Task: a new Task instance
         """
+        for arg in args:
+            if not hasattr(arg, "__getitem__") and not isinstance(arg, unmapped):
+                raise TypeError(
+                    "cannot map over an unsubscriptable object ({})".format(arg)
```
cicdw:
```suggestion
                    "Cannot map over an unsubscriptable object ({})".format(arg)
```

PR #2050, opened 2020-02-19
Bugfix: DeleteNamespacedJob K8s task

romaintha:
There is a bug in the DeleteNamespacedJob K8s task. in the api_client delete_namespaced_job call, api_client is used instead of client.
Fixed it here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Fixes
 
-- None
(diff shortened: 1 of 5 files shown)


romaintha:
Actually, I guess the problem is the same in all K8s delete task ( DeleteNamespacedDeployment, DeleteNamespacedJob, DeleteNamespacedPod , DeleteNamespacedService). I could fix it in this PR if it s ok for you?

joshmeek:
@romaintha Yeah update each task if you want!

romaintha:
@joshmeek Done

joshmeek:
Thanks @romaintha! Do you think you could also add this fix to the changelog and your name to the contributors section?

romaintha:
@joshmeek Done as well

PR #2062, opened 2020-02-21
Core Diagnostics Utility

joshmeek:

## What does this PR change?
This PR adds a `diagnostics` utility to assist in debugging / troubleshooting issues users may be having with using Prefect without sharing any information that is too detailed. There are four main areas of diagnostics that may help in these situations:
- System information (OS, python version, prefect version)
- User configuration overrides
- User set `PREFECT__` environment variables
- Flow attribute presence

This is built in a way where only the presence of such information is displayed and not the actual information itself. For example calling `flow.diagnostics()`on a flow with Docker storage, FargateTaskEnvironment, JSONResultHandler, and a schedule that uses a CronClock would look similar to the following:
```
"flow_information": {
    "environment": {
      "aws_access_key_id": true,
      "aws_secret_access_key": true,
      "aws_session_token": false,
      "labels": false,
      "logger": true,
      "on_exit": false,
      "on_start": false,
      "region_name": true,
      "task_definition_kwargs": {
        "cpu": true,
        "taskRoleArn": true
      },
      "task_run_kwargs": false,
      "type": "FargateTaskEnvironment"
    },
    "result_handler": {
      "type": "JSONResultHandler"
    },
    "schedule": {
      "cron": "* * * * *",
      "end_date": null,
      "parameter_defaults": {},
      "start_date": null,
      "type": "CronClock"
    },
    "storage": {
      "_flows": false,
      "base_image": true,
      "base_url": true,
      "dockerfile": false,
      "env_vars": {
        "PREFECT__USER_CONFIG_PATH": true
      },
      "extra_commands": true,
      "files": false,
      "flows": false,
      "image_name": false,
      "image_tag": false,
      "local_image": false,
      "prefect_version": true,
      "python_dependencies": true,
      "registry_url": false,
      "result_handler": false,
      "type": "Docker"
    },
    "task_count": 1
  },
```

Configuration information takes the following form in order to display the presence of user-set configuration variables / env vars (with an optional `secrets` boolean flag to disable outputting the _name_ of local secrets):
```
{
  "config_overrides": {
    "cloud": {
      "agent": {
        "auth_token": true
      },
      "api": true,
      "use_local_secrets": true
    },
    "context": {
      "secrets": {
        "TEST_SECRET": true
      }
    },
    "flows": {
      "checkpointing": true
    },
    "home_dir": true,
    "logging": {
      "level": true
    }
  },
  "env_vars": [
    "PREFECT__CONTEXT__BIGTEST"
  ],
  "system_information": {
      "platform": "Darwin-16.7.0-x86_64-i386-64bit",
      "prefect_version": "0.9.4+74.g14c23ce9.dirty",
      "python_version": "3.7.3"
    }
}
```

This can be called three ways: `prefect.utilities.diagnostics`, `flow.diagnostics`, or `prefect diagnostics` (from CLI, note: CLI will not give flow-level information).

## Why is this PR important?
First step towards building a more efficient support process


src/prefect/cli/__init__.py
```diff
@@ -48,6 +48,7 @@ def cli():
     Miscellaneous Commands:
         version     Get your current Prefect version
         config      Output your Prefect config
```
jlowin: Not related to your PR but I'm just noticing the word "your" here which should be removed -- "Output the Prefect Config", "Print the Prefect version"

src/prefect/cli/__init__.py
```diff
@@ -79,3 +80,22 @@ def config():
     Output your Prefect config
     """
     click.echo(prefect.config.to_dict())
+
+
+@cli.command(hidden=True)
+@click.option(
```
jlowin: This option defaults to `True`, so I'm not sure how to use it -- doesn't `prefect diagnostics` have the same behavior as `prefect diagnostics --no-secrets`? I agree the default should be to hide secrets, but that suggests to me the flag should maybe be `--include-secret-names` or something making the opt-in explicit.

src/prefect/core/flow.py
```diff
@@ -1225,6 +1226,18 @@ def serialize(self, build: bool = False) -> dict:
...
         return serialized
 
+    # Diagnostics  ----------------------------------------------------------------
+
+    def diagnostics(self, secrets: bool = True) -> str:
+        """
+        Get flow and Prefect diagnostic information
+
+        Args:
+            - secrets (bool, optional): toggle output of Secret names, defaults to True.
```
jlowin: This variable name isn't clear without really parsing the docstring -- I suggest making it more explicit (`include_secret_names`) or something.

tests/utilities/test_diagnostics.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Kwargs presence check
+    assert flow_information["environment"]["executor"] is True
+    assert flow_information["environment"]["executor_kwargs"] is False
+    assert flow_information["environment"]["labels"] is False
+    assert flow_information["environment"]["on_start"] is False
+    assert flow_information["environment"]["on_exit"] is False
+    assert flow_information["environment"]["logger"] is True
+
+    # executor': True, 'executor_kwargs': False, 'labels': False,
```
wagoodman: dead comments I presume

src/prefect/utilities/diagnostics.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Get system information
+
+    Returns:
+        - dict: a dictionary containing some system information
+    """
+    return dict(
+        system_information=dict(
+            platform=platform.platform(),
+            python_version=platform.python_version(),
```
cicdw: Nice, really making sure we get accurate info 👏

PR #2071, opened 2020-02-24
Add PIN 15: Skip as Finished state & Control Flow conditionals

### LINKED ISSUE #2070: Introduce PIN for Skipped state refactor
There have been offline discussions related to #2019 to refactor how `Skipped` states are treated to allow for more nuanced control flows.  Because of the large scope of the change, we should introduce a PIN describing our reasoning.

wagoodman:

## What does this PR change?
Mainly proposes two changes:
- Interpret `Skip` state as fundamentally `Finished` and not `Success`ful
- Add `conditional` Task wrapper as an alternative to `switch` (when there is a single case)

Closes #2070 

## Why is this PR important?
By making `Skip` more semantically correct we are reducing surprising behavior (while enabling more control flow features).


docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
```diff
@@ -0,0 +1,66 @@
...
+5. Add a new `no_failed` trigger, which passes if there are no failed states upstream.
+
+6. Introduce a `conditional` object that acts similar to `switch`, supporting only a single case (the `True` case). This will be a `task.run` wrapper instead of adding an additional CompareValue task. Any task can be a condition, a truthy value will be determined by the result of the given task, and if determined to be “False” then `FAIL` is raised.
+
+   Example usage (within a flow context):
+   `conditional(is_b_enabled, b_task)`
+   instead of
+   `switch(is_b_enabled, {True: b_task})`
+
+7. Add the ability to combine conditions: add `any_condition` and `all_conditions` reducing functions to `prefect.tasks.control_flow.conditional`. This would add an additional task to the flow, with upstreams being the individual conditions, and the new task emitting a “True” result or raising a `FAIL`.
```
jlowin: It's interesting to consider whether this is worth expanding to all trigger functions as a convenience that does exactly what you describe, just with a more dynamic trigger.

`conditional(all_successful(a, b), c)`

`conditional(any_failed(a,b), c)`

jlowin:
Awesome @wagoodman -- this will be a massive boost for complex flows. Could you add the PIN to https://github.com/PrefectHQ/prefect/blob/master/docs/core/PINs/README.md as well?

docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
```diff
@@ -0,0 +1,66 @@
...
+
+Author: Alex Goodman
+
+## Status
+
+Proposed
+
+## Context
+
+The [ControlFlow](https://docs.prefect.io/api/latest/tasks/control_flow.html) Tasks allow for users to branch conditionally within the execution graph, skipping some tasks and allowing execution of others. However, [back-to-back conditions](https://github.com/PrefectHQ/prefect/issues/2017) have proven to be a problem under the existing implementation. Specifically, the problem appears to be with the semantics with the “Skip” Prefect State: it is considered to be fundamentally a “Success” state. That is, it is not easy (if impossible) to determine why a task was skipped, and if ignoring that Skipped state is OK. This means that using `skip_on_upstream_skip` is not a good enough mechanism to implement back-to-back conditions and hints that more functionality is needed.
```
cicdw: Could you add one sentence to this to expand on the ambiguity of the Skip state: right now it represents both tasks that were never run as well as tasks that raise SKIP signals (which is implicit in your comment on "why a task was skipped" but for posterity I want to document that ambiguity in particular)?

In addition, could you add a line / second paragraph maybe about "TriggerFailed" being a subclass of Failed  and why that is not ideal (because failures propagate, `any_failed` triggers run, etc.)

docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
wagoodman: Interesting. As proposed the `any/all_condition` functions take Tasks and replace `Task.run` with a wrapper function that takes care of determining the result and raising `FAIL`. Your addition implies to modify all existing triggers: when given states, behave as today, when given tasks, return the appropriate wrapper function (for the `conditional` to use as originally proposed).

I like this from the perspective that triggers could be used and combined in natural ways and doesn't add the additional `any/all_condition` functions. This would prevent questions about nomenclature ("what do you mean by use `any_condition` with Tasks? is a Condition a Task? How is `conditional` related to the 'condition' in question here?"). 

One downside is that trigger functions become somewhat surprising: they behave in completely different ways when given different information (state parameters vs task parameters).

To sidestep this downside we could introduce a `Trigger` object which `__call__` would evaluate the trigger (taking `States` as today) and another function to create the Task wrapping function (or maybe another Task completely). It would still mean refactoring the existing triggers into these objects.

A third option would be to change the signature of `conditional` to be something like: `conditional(all_successful, [a,b], c)`. I'm not a huge fan of it, but it would reduce the complexity of what I mentioned above (that is, we should be able to use the triggers as we do today).

docs/core/PINs/README.md
```diff
@@ -71,3 +71,27 @@ A proposal for a new way of building schedules that captures a wider variety of
 A proposal for a way to loop over single tasks with arbitrary control logic and all Prefect semantics intact
 
 **Status:** Proposed
+
+## PIN 12: Environment Callbacks](PIN-12-Environment-Callbacks.md)
```
joshmeek:
```suggestion
## [PIN 12: Environment Callbacks](PIN-12-Environment-Callbacks.md)
```

docs/core/PINs/README.md
```diff
@@ -71,3 +71,27 @@ A proposal for a new way of building schedules that captures a wider variety of
...
+
+A proposal to introduce user-provided callbacks for `Environments` which users can use to specify additional behavior before a Flow is run and after a FlowRun has ended.
+
+**Status:** Accepted
+
+## PIN 13: Universal Cloud Deploys](PIN-13-Universal-Deploy.md)
```
joshmeek:
```suggestion
## [PIN 13: Universal Cloud Deploys](PIN-13-Universal-Deploy.md)
```

docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
wagoodman: Good call; added more info in [b3a2e41](https://github.com/PrefectHQ/prefect/pull/2071/commits/b3a2e4170118c4a53296e1bd2e97de1587bf973b) to account for these points.

docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
```diff
@@ -15,7 +15,43 @@ Proposed
...
+@prefect.task(trigger=prefect.triggers.any_failed)
+def c(): pass
+
+with prefect.Flow("My Flow") as flow:
+    flow.chain(a, b, c)
+```
+
+This will result in `B` in a `TriggerFailed` state followed by `C` completing successfully, which is arguably surprising behavior.
+
+Ideally, a trigger should the gatekeeper for Task exection: either attempt to run the current Task (or not) based on upstream states. Or said another way, attempt to run the current Task or Skip it --this does not imply a `Failed` (or `Success`ful) state being determined yet, as that should be the result of the Task itself.
```
cicdw:
```suggestion
Ideally, a trigger should be the gatekeeper for Task exection: either attempt to run the current Task (or not) based on upstream states. Or said another way, attempt to run the current Task or Skip it --this does not imply a `Failed` (or `Success`ful) state being determined yet, as that should be the result of the Task itself.
```

docs/core/PINs/PIN-15-Skip-as-Finished-and-Conditionals.md
```diff
@@ -15,7 +15,43 @@ Proposed
...
+
+Furthermore, indicating when a trigger condition is not met can also be misleading. By raising the TRIGGERFAIL, a `Failed` state propagates through downstream tasks, which could be misconstrude as these downstream tasks failing, when in fact they have semantically been skipped. Take the following flow as an example:
```
cicdw:
```suggestion
Furthermore, indicating when a trigger condition is not met can also be misleading. By raising the TRIGGERFAIL, a `Failed` state propagates through downstream tasks, which could be misconstrued as these downstream tasks failing, when in fact they have semantically been skipped. Take the following flow as an example:
```

cicdw:
The changes LGTM - requested two superficial typo changes

PR #2092, opened 2020-02-26
Add optional stdout logging in tasks

### LINKED ISSUE #2040: Add optional stdout logging for tasks
## Current behavior
Currently, `print` statements or any other stdout is not treated in any special way.  This has two consequences:

- users who use `print` statements locally and can see the output are surprised when the output doesn't appear in the Cloud logs
- print statements from third-party libraries that might be useful are not surfaced in cloud execution

## Proposed behavior
We should introduce a new keyword argument on all Tasks `log_stdout` which defaults to `True`.  When `True`, all standard out is converted to a DEBUG level log.  When toggled to `False`, we will not redirect.

This will capture richer information from task runs, but also allow for incredibly simple logging for our less advanced users:

```python
@task
def example():
    print('hi')
    print('this is a test')
    print('of our emergency broadcast system')
```

will result in 3 debug logs.

## Example
This is actually very simple with Python builtins:

```python
import prefect
from contextlib import redirect_stdout

class RedirectToLog:
    def write(self, s):
        # avoids blank newlines
        if s.strip():
            prefect.utilities.logging.prefect_logger.debug(s)

to_debug_log = RedirectToLog()

with redirect_stdout(to_debug_log):
    print('hi')
    print('this is a test')
    print('of our emergency broadcast system')
```

results in :

```
[2020-02-17 00:02:13,642] 4-- DEBUG - prefect | hi
[2020-02-17 00:02:13,642] 4-- DEBUG - prefect | this is a test
[2020-02-17 00:02:13,643] 4-- DEBUG - prefect | of our emergency broadcast system
```

**NOTE:** In our case, we want want to generate the redirect class on the fly so that it uses the task run specific logger in each case.

joshmeek:

## What does this PR change?
Closes #2040 
By default tasks now have a `log_stdout=True` kwarg which can be toggled based on whether or not to send stdout messages as Prefect logs.

Example:

```python

@task
def me():
    print("@@@")

f = Flow('test', tasks=[me])
```

image

## Why is this PR important?
As we have seen, logging stdout is commonly a new user assumption when it comes to using Prefect / Prefect Cloud. Having to pull the logger from context is not that intuitive to new users:
```python
from prefect import context
logger = context.get('logger')
logger.info('@@@')
```

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -154,6 +154,8 @@ class Task(metaclass=SignatureValidator):
             result of the previous handler.
         - on_failure (Callable, optional): A function with signature `fn(task: Task, state: State) -> None`
             with will be called anytime this Task enters a failure state
+        - log_stdout (bool, optional): Toggle whether or not to send stdout messages to
+            the Prefect logger. Defaults to `True`.
 
     Raises:
         - TypeError: if `tags` is of type `str`
@@ -180,8 +182,8 @@ class Task(metaclass=SignatureValidator):
         result_handler: "ResultHandler" = None,
         state_handlers: List[Callable] = None,
         on_failure: Callable = None,
+        log_stdout: bool = True,
     ):
-
         self.name = name or type(self).__name__
         self.slug = slug or str(uuid.uuid4())
 
@@ -254,6 +256,8 @@ class Task(metaclass=SignatureValidator):
                 callback_factory(on_failure, check=lambda s: s.is_failed())
             )
         self.auto_generated = False
+
+        self.log_stdout = log_stdout
 
     def __repr__(self) -> str:
         return "<Task: {self.name}>".format(self=self)
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -155,7 +155,7 @@ class Task(metaclass=SignatureValidator):
         - on_failure (Callable, optional): A function with signature `fn(task: Task, state: State) -> None`
             with will be called anytime this Task enters a failure state
         - log_stdout (bool, optional): Toggle whether or not to send stdout messages to
-            the Prefect logger. Defaults to `True`.
+            the Prefect logger. Defaults to `False`.
 
     Raises:
         - TypeError: if `tags` is of type `str`
@@ -182,7 +182,7 @@ class Task(metaclass=SignatureValidator):
         result_handler: "ResultHandler" = None,
         state_handlers: List[Callable] = None,
         on_failure: Callable = None,
-        log_stdout: bool = True,
+        log_stdout: bool = False,
     ):
         self.name = name or type(self).__name__
         self.slug = slug or str(uuid.uuid4())
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -876,7 +876,7 @@ class TaskRunner(Runner):
             raw_inputs = {k: r.value for k, r in inputs.items()}
 
             if self.task.log_stdout:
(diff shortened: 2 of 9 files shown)


joshmeek:
Open question: if we keep this enabled by default does it make it a breaking change?

src/prefect/core/task.py
```diff
@@ -180,8 +182,8 @@ def __init__(
         result_handler: "ResultHandler" = None,
         state_handlers: List[Callable] = None,
         on_failure: Callable = None,
+        log_stdout: bool = True,
```
cicdw: As @wagoodman pointed out in the issue, it might be better to use `False` as a default here

src/prefect/engine/task_runner.py
```diff
@@ -873,9 +874,16 @@ def get_task_run_state(
                 timeout_handler or prefect.utilities.executors.timeout_handler
             )
             raw_inputs = {k: r.value for k, r in inputs.items()}
-            result = timeout_handler(
-                self.task.run, timeout=self.task.timeout, **raw_inputs
-            )
+
+            if self.task.log_stdout:
+                with redirect_stdout(prefect.utilities.logging.log_stdout):  # type: ignore
```
cicdw: I think we should log with the Task Runner logger here (`self.logger`) to keep the logs in an intuitive place; maybe you can take your `RedirectToLog` and allow it to be initialized with a logger?  E.g., so this would be
```python
with redirect_stdout(RedirectToLog(self.logger)):
```

src/prefect/utilities/logging.py
```diff
@@ -258,3 +258,37 @@ def get_logger(name: str = None) -> logging.Logger:
...
+
+    def flush(self) -> None:
+        """
+        Implemented flush operation for logger handler
+        """
+        for handler in self.stdout_logger.handlers:
+            handler.flush()
+
+
+log_stdout = RedirectToLog()
```
cicdw:
```suggestion
```

src/prefect/utilities/logging.py
```diff
@@ -258,3 +258,34 @@ def get_logger(name: str = None) -> logging.Logger:
...
+    def write(self, s: str) -> None:
+        """
+        Write message from stdout to a prefect logger.
+        Note: blank newlines will not be logged.
+
+        Args:
+            s (str): the message from stdout to be logged
+        """
+        if s.strip():
+            self.stdout_logger.info(s)
```
cicdw: Apologies for submitting so many small reviews, but are we sure these belong as INFO logs vs. DEBUG?  I'd vote DEBUG but open to other ideas here

joshmeek: I actually think INFO is more appropriate because the way I think about it is people use stdout to opt-in to printing output. Where as DEBUG to me generally seems as something that would be unintentional lower level automated logs

cicdw: @lauralorenz / @wagoodman want to be a tie-breaker?

lauralorenz: tl;dr I vote INFO

I agree with @joshmeek 's rationale. I do like DEBUG when I think of using `print()` only to quickly throw in little stuff early on in development assuming I will configure loggers later, but it seems the intention of this from the get-go is to provide a wholesale replacement for using the logging API even on deployed flows based on known user expectations.

In addition I think users would prioritize their personalized `print()` logs over stuff at debug level; I'm thinking of it in terms of it's better to a) make it easier to set logging levels to INFO to get rid of lower level debug logs *without* getting rid of their personalized `print()` logs, instead of b) making it easier to get rid of their own `print()` logs.

cicdw: It is decided!

PR #2093, opened 2020-02-27
Add extra-requirements.txt file for dependabot version tracking

### LINKED ISSUE #2087: Investigate Dependabot tracking of "extra" dependencies
We have a lot of optional extra dependencies such as kubernetes, boto3, etc. which we currently do not track releases for via Dependabot. Right now it is only looking at our base default dependencies that are present in `requirements.txt`. This could lead to some unexpected behavior especially as Prefect's integrations with other libraries are built out. 

Let's look into if it is possible to track some of our extra dependencies and determine which we want to be responsible for. For example `kubernetes` is an extra dependency but heavily used with things like the KubernetesAgent and DaskKubernetesEnvironment which makes it more of an integral dependency as opposed to something like `airtable-python-wrapper` which is only used in the the Airtable tasks in the task library.

joshmeek:
After looking into the dependabot source (notably [here](https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/file_fetcher.rb#L19)) we can have dependabot track the _extra_ dependencies by putting them into their own `*-requirements.txt` file. This is in an effort to keep extra packages for things like the task library and agents from getting stale. I believe this would mean that when dependabot opens a PR for a dependency in this file then we would have to add an extra commit to that PR to update setup.py (not strongly opposed to that). Unless we found some way to refactor the `extras = {...}` dictionary in the setup.py

**Open for discussion**

Closes #2087

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -1,0 +1,20 @@
+# Extra requirements file for dependabot PRs to check dependency versions
+
+airtable-python-wrapper >= 0.11, < 0.12
+azure-cosmos >= 3.1.1, <3.2
+azure-storage-blob >= 12.1.0, < 13.0
+azureml-sdk >= 1.0.65, < 1.1
+boto3 >= 1.9, < 2.0
+dask-kubernetes >= 0.8.0
+dropbox ~= 9.0
+feedparser >= 5.0.1, < 6.0
+google-cloud-bigquery >= 1.6.0, < 2.0
+google-cloud-storage >= 1.13, < 2.0
+graphviz >= 0.8.3
+jinja2 >= 2.0, < 3.0
+kubernetes >= 9.0.0a1, <= 11.0.0b2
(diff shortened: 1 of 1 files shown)


cicdw:
Hm I generally don't like the idea of two sources of truth for things; this approach makes sense if we can find a way to actually rely on it for specifying our `extras`

joshmeek:
@cicdw Yeah that was my main hesitation. Will look into it 👍

joshmeek:
Closes #2087 

After some investigating dependabot actually looks to parse [everything](https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/requirement_parser.rb) that we need from the extras in our setup.py file. The reason why we haven't seen many PRs opened for the extras is because we have very generous bounds on them. Closing for now because everything appears to be working as planned 😄

PR #2102, opened 2020-02-28
Support modular configuration files

jlowin:

## What does this PR change?
This PR changes the Prefect configuration mechanism to support loading multiple configuration files; previously it accepted two: the default config and a user config. Now it supports arbitrarily many as a single `paths` parameter.

This allows us to easily split the config into multiple files, as appropriate - in this case, a new `cloud-config.toml` file to contain Prefect Cloud-specific settings.


src/prefect/configuration/__init__.py
```diff
@@ -284,45 +290,36 @@ def interpolate_config(config: dict, env_var_prefix: str = None) -> Config:
...
 
     validate_config(config)
     return config
 
 
 # load prefect configuration
-config = load_configuration(
-    path=DEFAULT_CONFIG, user_config_path=USER_CONFIG, env_var_prefix=ENV_VAR_PREFIX
-)
+config_paths = [DEFAULT_CONFIGS]
```
cicdw: Isn't this a list of lists?

src/prefect/configuration/__init__.py
```diff
@@ -284,45 +277,36 @@ def interpolate_config(config: dict, env_var_prefix: str = None) -> Config:
...
 
     Returns:
         - Config
     """
 
-    # load default config
-    default_config = load_toml(path)
-
-    # load user config
-    if user_config_path and os.path.isfile(str(interpolate_env_vars(user_config_path))):
```
joshmeek: @cicdw Do you think this check should be added back for every config option passed in with `paths`?

joshmeek: `paths` could contain default, user, custom configs, etc. so either we could error out if a provided config does not exist or we can skip it

cicdw:
Changes to how our configuration works require a very low risk / reward ratio and I don't think this stands up - closing.

PR #2109, opened 2020-03-02
Output direct link to newly created Cloud flows and runs

### LINKED ISSUE #899: After deploying a flow, display a UI link to see it immediately
UX improvement: after calling `flow.deploy()`, present a message to the user with a link to viewing the flow in the UI.

joshmeek:

## What does this PR change?
Closes #899 

When flows are registered w/ Cloud they will now output a direct link to the UI (can be toggled off)

```bash
# calls flow.register()
$ python etl.py
Result Handler check: OK
Flow: https://cloud.prefect.io/myslug/flow/55b555ce-30c9-48ad-b49d-5555dc20rr5t
```

Cloud runs from the CLI will also output a direct link (can be toggled off)

```
$ prefect run cloud -n ETL -p Demo
Flow Run: https://cloud.prefect.io/myslug/flow-run/5555202e-4a08-40d4-a1f0-d462a1b6rt5w
```

## Why is this PR important?
Makes it quicker to switch from local development to your flows/runs in the UI!


src/prefect/cli/run.py
```diff
@@ -66,8 +69,22 @@ def run():
 @click.option(
     "--logs", "-l", is_flag=True, help="Live logs of the flow run.", hidden=True
 )
+@click.option(
+    "--flow-run-id-only",
```
cicdw: This is very descriptive but also long; not a blocker but what about `--no-url`?  The main benefit would be a common flag / kwarg for our CLI commands which optionally output URLs

src/prefect/cli/run.py
```diff
@@ -147,7 +165,27 @@ def cloud(
...
+        url = (
+            re.sub("api-", "", config.cloud.api)
+            if re.search("api-", config.cloud.api)
+            else re.sub("api", "cloud", config.cloud.api)
+        )
+
+        def join(url: str, items: list) -> str:
+            if len(items) == 1:
+                return urljoin(url, items[0])
+            return join(urljoin(url, items[0]) + "/", items[1:])
```
cicdw: I definitely get this, but since we know that the config URL won't include any terminal pages (e.g., `index.html`) I suspect we can get away with a simpler implementation here, e.g., 
```python
"/".join([url.rstrip("/"), tenant_slug, "flow-run", flow_run_id])
```

src/prefect/client/client.py
```diff
@@ -562,6 +564,8 @@ def register(
                 will be used.
             - compressed (bool, optional): if `True`, the serialized flow will be; defaults to `True`
                 compressed
+            - flow_id_only (bool, optional): if `True`, the stdout from this function will not contain the
```
cicdw: Same suggestion from above about `--no-url`

src/prefect/client/client.py
```diff
@@ -640,8 +644,42 @@ def register(
...
+                if re.search("api-", prefect.config.cloud.api)
+                else re.sub("api", "cloud", prefect.config.cloud.api)
+            )
+
+            def join(url: str, items: list) -> str:
+                if len(items) == 1:
+                    return urljoin(url, items[0])
+                return join(urljoin(url, items[0]) + "/", items[1:])
+
+            flow_url = join(url, [tenant_slug, "flow", flow_id])
```
cicdw: Same suggestion about a (possibly) simpler implementation

src/prefect/client/client.py
```diff
@@ -640,8 +644,42 @@ def register(
...
 
+    def get_default_tenant_slug(self) -> str:
+        """
+        Get the default tenant slug for the currently authenticated user
+
+        Returns:
+            - str: the slug of the current default tenant for this user
+        """
+        res = self.graphql(
+            query={"query": {"user": {"default_membership": {"tenant": "slug"}}}}
```
cicdw: Question: why are you using the default value?  E.g., I could imagine someone registering a flow off a token that doesn't correspond to their default membership.

src/prefect/cli/run.py
joshmeek: Hahaha I like that shorter option much better

src/prefect/client/client.py
joshmeek: I was working with the idea that whichever your current tenant is in conjunction with the token used to authenticate the client would be the ideal tenant to use. I could add an option to also provide a tenant slug if you think it would be useful.

PR #2110, opened 2020-03-02
Closes #2077 update context documentation

### LINKED ISSUE #2077: Update Context documentation
We need to update:
- https://docs.prefect.io/api/latest/utilities/context.html 
- https://docs.prefect.io/core/concepts/execution.html#context 

with more information as to what context can be used for, how you can override values and why you might want to use custom values.

lauralorenz:

## What does this PR change?
Closes #2077. Update context documentation to make it more clear how users can affect context and what is useful from the prefect supplied context, particularly time data.

The original issue also calls out the docstring in `context.py`, but I didn't see anything particularly wrong there, and just added a link to the concept doc. Let me know if you want more there.

## Why is this PR important?
How exactly to modify context was buried a bit in a comprehensive example instead of step-by-step, and the utility of the existing Prefect context was not particularly called out. In addition the warning box was not very nuanced about when it is appropriate to modify context, even though we suggest it in support to people.


docs/core/concepts/execution.md
```diff
@@ -100,6 +126,34 @@ with prefect.context(key='abc'):
...
+[2020-03-02 22:15:58,789] INFO - prefect.FlowRunner | Flow run SUCCESS: all reference tasks succeeded
+```
+Using this context can be useful to write time-aware tasks, such as tasks that trigger future work respective to its start time using `prefect.context.tomorrow` or processing only the prior day's data by using `prefect.context.yesterday`.
+
+::: tip What else is in context?
+For an exhaustive list of values that you can find in context, see the corresponding [API documentation](../../api/latest/utilities/context.html).
+:::
+
+::: warning Avoid modifying Prefect-supplied context
+Since Prefect uses context internally to track metadata during the flow and task run logic, modifying Prefect-supplied context keys can have unintended consequences. It is recommended to avoid overriding the key names described in the API documentation.
```
cicdw: It might be useful to add some additional nuance here: in particular, all of the timestamp related context keys can be overwritten by users and that is a supported use of context (mainly for the purposes of "backfilling").

docs/core/concepts/execution.md
```diff
@@ -81,7 +81,33 @@ Another common use of Prefect signals is when the task in question will be neste
...
+Adding context globally is possible via your `config.toml` in a section called `[context]`. For any keys specified here, as long as Prefect Core does not override this key internally, it will be accessible globally from `prefect.context`, even outside of a flow run.
+```
+# config.toml
+[context]
+a = 1
+```
+
+```python
+>>> import prefect
+>>> prefect.context.to_dict()['a']
```
joshmeek: I believe we can remove the need to call `to_dict` since context acts in a similar manner to a dictionary already!

```python
In [1]: from prefect import context

In [2]: context.test
Out[2]: 'here'

In [3]: context.get('test')
Out[3]: 'here'

In [4]: context['test']
Out[4]: 'here'
```

docs/core/concepts/execution.md
lauralorenz: Updated in https://github.com/PrefectHQ/prefect/pull/2110/commits/42fb15c3114cce60b6db336f0c2edcf89a4fad26

docs/core/concepts/execution.md
lauralorenz: Duh u right! Updated in https://github.com/PrefectHQ/prefect/pull/2110/commits/69e697b31274215c92e0a1affd8de6691a592f45 with some extra examples off of your comment here

PR #2115, opened 2020-03-04
A Pandas DataFrame Result

AndrewRook:

## What does this PR change?
This PR implements a new result handler that saves and loads Pandas DataFrames using Panda's built-in tools. It's based on the same handler in my [`prefect_ds`](https://github.com/AndrewRook/prefect_ds) library. Discussion in [this Slack thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1582684214035000) may also be useful as a reference for why this PR is coming into Prefect looking like it does. 

**This PR, in its current form, is not something I would consider mergeable into Prefect,** but rather is intended to start the conversation about what modifications would need to be made in order to make it mergeable. There are three main reasons for this:

1. `PandasResultHandler` requires the full filepath, including the file name, to be specified. This is in contrast (I believe) to the rest of the file handlers, which have the user specify the directory/bucket/container/etc and then generate the actual filename themselves. My interpretation of this design choice is to make using the result handlers with `map` easier, because with the full filepath specified a mapped task would just write to the same file for every iteration of the map.
2. `PandasResultHandler` has hooks for formatting the filepath based on the inputs to the task, to handle the aforementioned mapping problem. But obviously Prefect does not recognize these hooks. (For reference, to get around this in `prefect_ds` I subclassed the `TaskRunner` and created a new state handler. This was a bit of a kludge, and could probably be implemented more elegantly in Prefect proper if that was a direction folks wanted to go.)
3. `PandasResultHandler` would add an optional Pandas dependency, and currently that's not included in any of the docs, installation tools, or CI (although I did try to account for it in the importing of the handler itself and in the tests). That's not a particularly challenging thing to resolve, but I wanted guidance on how y'all would like that to be implemented.

## Why is this PR important?
A lot of data workloads, especially data analysis/modeling workflows, revolve around Pandas DataFrames. Furthermore, when analyzing data it's important to be able to manually inspect and validate what you're dealing with, which is hard to do with the `LocalResultHandler` writing files to randomly named pickles. 

@cicdw (and/or anyone else) take a look and let me know what you think (no rush through 😄).


docs/outline.toml
```diff
@@ -156,7 +156,7 @@ classes = ["Result", "SafeResult", "NoResultType"]
 [pages.engine.result_handlers]
 title = "Result Handlers"
 module = "prefect.engine.result_handlers"
-classes = ["JSONResultHandler", "GCSResultHandler", "LocalResultHandler", "S3ResultHandler", "AzureResultHandler", "SecretResultHandler"]
+classes = ["JSONResultHandler", "GCSResultHandler", "LocalResultHandler", "PandsResultHandler", ""S3ResultHandler", "AzureResultHandler", "SecretResultHandler"]
```
cicdw:
```suggestion
classes = ["JSONResultHandler", "GCSResultHandler", "LocalResultHandler", "PandsResultHandler", "S3ResultHandler", "AzureResultHandler", "SecretResultHandler"]
```

src/prefect/engine/result_handlers/pandas_result_handler.py
```diff
@@ -0,0 +1,148 @@
...
+    can be fully specified at task instantiation.
+
+    Args:
+        - path (str): Filepath to be read from or written to, including file name and extension.
+        - file_type (str): The type of file to write to, e.g. "csv" or "parquet". Must match the name
+            of the appropriate ``to_[FILETYPE]``/``read_[FILETYPE]`` method.
+        - read_kwargs (dict, optional): If present, passed as **kwargs to the ``read_[FILETYPE]`` method.
+        - write_kwargs (dict, optional): If present, passed as **kwargs to the ``to_[FILETYPE]`` method.
+
+    .. note::
```
cicdw: We actually support markdown in our docstrings, so you might change this to:
```suggestion
    **Note:**
```

src/prefect/engine/result_handlers/pandas_result_handler.py
```diff
@@ -0,0 +1,148 @@
...
+        Because the filepath is fully specified, when using this handler in a ``map``
+        can lead to the same file being read to or written from by every iteration of
+        the map. To deal with this case this handler allows for Python's ``str.format``
+        to be used with the names of arguments to the task instance. For instance, if
+        your task has an argument named ``sample_name`` that you intend to map over,
```
cicdw: This is a great idea! We _might_ consider expanding this as a general rule to also allow for formatting from keys in `prefect.context`

docs/outline.toml
AndrewRook: Good catch, thanks!

src/prefect/engine/result_handlers/pandas_result_handler.py
AndrewRook: Oh right, my bad. I must be brainwashed by Sphinx 😛

AndrewRook:
I made those minor edits, thanks for flagging them. Let me know how I can help with the broader conversation about the `read` method and how to incorporate `PandasResultHandler`.

cicdw:
Hey @AndrewRook - sorry for the period of silence; this PR fits very nicely into a new effort we have underway to refactor how `Result`s and `ResultHandler`s behave (essentially we want to unify their interface into a collection of `Result` subclasses).  You can read our PIN describing the idea here: https://docs.prefect.io/core/PINs/PIN-16-Results-and-Targets.html

Now that https://github.com/PrefectHQ/prefect/pull/2147 has been merged, the result interface is in place and in theory your result handler could be refactored to become one of the first `Result` subclasses.  The only way (I think) your interface would need to change is:

- use the new `filename_template` and `format` it using `prefect.context` 
- refactor the `read` method to optionally accept an absolute path, and otherwise use `self.filename_template.format(**prefect.context)`

To be clear, that's not a requirement here and we can move forward with the pure result handler interface, but I wanted to throw this out there in case it seems more appealing to you!

AndrewRook:
Oh yeah, that PIN seems like it implements exactly (or almost exactly) what I was trying to do here, with not only this result handlers but a lot of the other stuff in prefect_ds. Sweet! It makes way more sense to me to migrate to that new interface rather than persist with the `result_handler` approach. 

I'll take a crack at refactoring it when I get a chance, although I'll be honest I have no idea what my free time will be like for the next several weeks.

cicdw:
Hey @AndrewRook - long time no talk; the new results / targets API is finally live!  I think your PandasResultHandler is a perfect fit for the new API.  Some resources to help you re-orient:
- [Blog](https://medium.com/the-prefect-blog/prefect-0-11-0-improved-apis-and-targets-79fef9d90aec)
- [Idiom on targets](https://docs.prefect.io/core/idioms/targets.html)
- [API docs](https://docs.prefect.io/api/latest/engine/results.html)

Let me know if you want any help refactoring this!

AndrewRook:
Cool, thanks @cicdw! With everything going on right now my time is pretty scattered and limited, but I'll see what I can do :)

One quick question — looking at the pages you linked I can see how the templating can be formatted from the `context` object, but looking at the docs for that object I didn't see any way to grab the values of task arguments. It's not a blocker for this PR if that functionality doesn't exist, but it would make the `Result` class less useful than the one in my library (especially for times when you're mapping over objects).

cicdw:
@AndrewRook at this moment it's not possible but that is the first enhancement request we received: https://github.com/PrefectHQ/prefect/issues/2577 so we'll definitely be looking into it!

AndrewRook:
Ok @cicdw I _think_ I properly re-implemented this as a regular `Result` and cleaned up all of the old result handler code. Take a look!

(Note I have not updated the changelog: since stuff moves so fast I figured I'd wait until it was clear this was almost ready to go in.)

AndrewRook:
Ok, I threw in a pandas extra, let me know if there's anything else (besides updating the changelog).

jcrist:
Apologies for stepping into this conversation a bit late. I have a small concern with this change, IMO it feels like the wrong abstraction-level to special-case Pandas at.

Currently our `Result` classes deal with places to put bytes. We have `LocalResult`, `GCSResult`, `S3Result`, etc... Each of these supports writing a result to their respective filesystems - users choose a `Result` class based on *where* they want to store results, not on *what* they expect to store. This would be the first instance of a `Result` class that cares about what it's storing.

This also breaks the abstraction of choosing *where* to store things. What if I want to store a pandas dataframe to azure? Do I use `AzureResult` or `PandasResult`? There's no clear way to mix them.

This also opens up the door to a multiple of other specific output-type classes (e.g. `NumpyResult`, ...), and doesn't deal well with tasks that might produce more than one dataframe (say a 2-tuple of dataframes).

---

It seems to me that we have two separate concepts here that we want to provide some user control over:

- Where to store the result (GCS, S3, locally, etc...)
- How to convert a task result into bytes

Currently the `Result` class has different implementations for the former, and hardcodes the latter to use `cloudpickle`.

One option would be to allow users to configure `serialization`/`deserialization` functions for individual `Result` instances. These would default to `cloudpickle`, but users might also provide ones that read/write to parquet, csv, etc... This is flexible, but requires user configuration to choose which serializer/deserializer for a specific `Result` instance. We might provide a set of common ones as part of prefect if needed.

```python
# Separate kwargs? dumps/loads are common in the python space
result = GCSResult(dumps=to_csv, loads=from_csv, ...)
# Maybe a single kwarg with an object, since choosing one dumps likely binds you to a matching loads?
result = GCSResult(serialize_as=prefect.serializers.csv)
```

Just a thought.

cicdw:
@jcrist yea I see what you're saying, that's a good call out.  I strongly believe we should support more nuanced forms of storing data (such as using `pandas` native `to_` methods) but I hadn't considered the explosion of `Result` instances that it might cause if we proceed in this direction. 

> One option would be to allow users to configure serialization/deserialization functions for individual Result instances. These would default to cloudpickle, but users might also provide ones that read/write to parquet, csv, etc...

Interesting idea - that makes sense to me.

jlowin:
I think @jcrist hit the nail on the head - the `Result` interface takes bytes and puts it *somewhere*. If we allowed users to supply serialize / deserialize callables (including instantiated but cloud pickleable classes, almost exactly like what @AndrewRook has here), we could get the best of both worlds: `Results` handling storage locations and `serializers` handling native encoding.

AndrewRook:
Yeah, I too think @jcrist is right on this one, he laid it out pretty clearly.

I'm happy to refactor, but this seems like the sort of thing where we'd want to put some additional thought into exactly what this `serializer` interface would look like, both to make it easy to integrate with the `Results` and easy for end users to understand how to use. Is this the sort of thing that a PIN would be good for? 

Oh, also, my $0.02: I would prefer using a single object for both serialization and deserialization, since users will almost always want matching methods. But I'm sure there's some edge case where someone will want to use ones that don't match, or some new technique that isn't already supported, so it ought not to be too difficult to roll your own.

cicdw:
I don't think we need a PIN for it, although after we agree on the interface we might want to update PIN 16 to reflect this new serializer interface as well.

We'll probably want to open an issue to discuss - I think the main complication I foresee is that serializers and backends are somewhat coupled.  For example, it's not immediately clear to me how to make two independent objects that could efficiently use both `pandas`' `to_csv` / `read_csv` methods and write to Azure Blob storage.  

We might have to establish some contract like "serializers should always return strings and deserializers should always assume string input" (or maybe we can assume JSON-serializable inputs and outputs), and then we could write `to_csv` to a StringIO buffer.

jcrist:
>  For example, it's not immediately clear to me how to make two independent objects that could efficiently use both pandas' to_csv / read_csv methods and write to Azure Blob storage.

> We might have to establish some contract like "serializers should always return strings and deserializers should always assume string input" (or maybe we can assume JSON-serializable inputs and outputs), and then we could write to_csv to a StringIO buffer.

I was thinking that serializers should deal with bytes only (write to bytes, read from bytes). This has worked quite well for dask - all of our io routines (csv, parquet, etc...) deal with raw bytes rather than files, the file -> bytes interface is handled by the filesystem abstraction.

cicdw:
Awesome @jcrist , I'm on board with that

cicdw:
Hey @AndrewRook - FYI this PR is ready for phase 3!  Customizable serializers are now a part of the Result interface 👍  Let us know if you have any questions or need any guidance

AndrewRook:
Oh cool! I had not noticed that, so thanks for flagging. I'll take a look when I get a chance.

jcrist:
Since we added serializers to the results classes, we'll want to handle this a different way than is done here. Closing this for now, feel free to reopen when ready.

PR #2122, opened 2020-03-04
add examples to Interactive API Cloud UI Docs

zhen0:

## What does this PR change?

- Adds example GQL Queries and explanations

## Why is this PR important?

- Part of Issue 1235 from the Cloud-web-ui development project: "We saw in user feedback sessions that using the Interactive API could be difficult for users who were unfamiliar with GraphQL. "
- Docs change - no tests


zhen0:
@znicholasbrown and @domarp-j  - Not sure if you can approve PRs on this repo but it would be good to have your comments here!

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
 # Interactive API
 
 
-The Interactive API is an embedded GraphQL client that allows you to access all of your data with complete flexibility. Thanks to features like schema introspection and automatic authentication, the interactive API makes it simple to access Cloud. You can also see the GraphQL API documentation directly inline.
+The Interactive API is an embedded GraphQL client that allows you to access all of your data with complete flexibility. Thanks to features like schema introspection and automatic authentication, the interactive API makes it simple to access Cloud. You can also see the GraphQL API documentation directly inline.  
+
+![](/cloud/ui/interactive-api.png)
+
+
+### Example Query
```
znicholasbrown:
```suggestion
### Basic Query
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+
+To find out information about your flows, you could use a flow query.  The basics for that query would be:
```
znicholasbrown: We should add some sort of transition here, something like:

```suggestion
Let's start with an example of how you might find basic information about your flows.
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+
+```graphql
+query {
+    flow {
+        name
+        id
+    }
+}
+```
+You can also use the autocomplete to show what other fields you could query:
```
znicholasbrown:
```suggestion
Notice that when we want to get the `flow` object, we must wrap it in a `query` block. This tells GraphQL that we're _retrieving_ something from the schema, as opposed to a `mutation` block which allows us to manipulate our data.

We've told the server to return a list of `flow` objects, each containing their respective `name` and `id` fields. As you interact with the schema in the editor window, you'll see an autocomplete window that will show hints of other available fields, which you can select to insert into your query. 
```

znicholasbrown: I think we should provide more instruction of what's going on here.

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+
+![](/cloud/ui/dropdown-on-interactive-api.png)
+
+### Nested Query
+
+To get information about the tasks connected to your flows, you could use a nested query:
```
znicholasbrown:
```suggestion
Let's build on our earlier query to get information about the tasks associated with our flows. One way to accomplish this would be to use a nested query:
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+  flow {
+    name 
+    id
+    tasks {
+      name
+    }
+  }
+}
+```
+
```
znicholasbrown:
```suggestion
We've nested `tasks` within the `flow` object, which tells the server to retrieve tasks only within the context of each flow that it returns.
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+### Limits and Offset
+
+To limit the number of items that are returned, you can use the Limit selector at the top of the Interactive API page. The default limit is 10 and the maximum is 100. Inline limit arguments are overriden by the value set in the Limit selector. 
+
+To the left of the Limit selector there is also an Offset selector. The Offset selector lets you say where your queries should start. So for example, if your query returned 5 flows and you set the limit to 2 and the offset to 0, you would see the first 2 items returned from your query.  If you set the limit to 2 and the offset to 2, you would see the next 2 items.
```
znicholasbrown:
```suggestion
To the left of the Limit selector there is also an Offset selector. The Offset selector tells the server at which index your queries should start. For example, if your unlimited query would return 5 flows and you set the limit to 2, an offset of 0 would would return the first two items in the set. To get the next two items with no overlap, you would set the offset to 2.
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+
+### Limits and Offset
+
+To limit the number of items that are returned, you can use the Limit selector at the top of the Interactive API page. The default limit is 10 and the maximum is 100. Inline limit arguments are overriden by the value set in the Limit selector. 
+
+To the left of the Limit selector there is also an Offset selector. The Offset selector lets you say where your queries should start. So for example, if your query returned 5 flows and you set the limit to 2 and the offset to 0, you would see the first 2 items returned from your query.  If you set the limit to 2 and the offset to 2, you would see the next 2 items. 
+
+### Query Filters
+
+Another way to limit the results returned would be to filter your search results.  For the flow query example, you could filter the results by looking only for flows with a certain name using the 'where' argument:
```
znicholasbrown:
```suggestion
Another way to limit the results returned would be to filter your search results. Building on our earlier flow query, we add a `where` argument to look for flows with a certain name:
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+    name 
+    id
+    tasks {
+      name
+    }
+  }
+}
+```
+
+You could also filter the results by looking only for flows with tasks with a certain name:
```
znicholasbrown:
```suggestion
We can also filter the results by looking only for flows with tasks with a certain name:
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,113 @@
...
+      name
+      id
+      tasks (where: {name: {_eq: "x"}}) {
+        name
+      }
+    }
+}
+```
+
+### Docs
```
znicholasbrown: I wonder if we might move the links to the Hasura and GraphQL docs into the contexts above, like a `Read more about query filters in the [Hasura Docs](https://hasura.io/docs/1.0/graphql/manual/queries/query-filters.html)` in each section.

zhen0: Good suggestion - I've updated.

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,107 @@
...
+### Schema
+
+In the docs for the Interactive API (included on the Interactive API page itself) you can find the schema - information about the queries and mutations you can run and more information about what fields you can request (for queries) or change (for mutations).
+
+![](/cloud/ui/interactive-api-inline-docs.png)  
+
+Finally, as you read further through the Prefect Cloud docs, look out for the GraphQL header:
+### GraphQL <Badge text="GQL"/>
+
+This header shows that we are giving an example of a query or mutation you can run using the Interactive API.
```
znicholasbrown: Could we add a heading to this section, something like `More Examples` maybe? Also the Y-margin around `### GraphQL` breaks up the section pretty oddly, maybe this could be a screenshot instead?

zhen0: I tried out a screenshot but it didn't section much better.  Switched to referencing the badge instead.

zhen0: @znicholasbrown - ready for another review!

zhen0:
@cicdw @joshmeek @wagoodman - as code owners - this is ready for your review.

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,116 @@
...
+    name 
+    id
+    tasks {
+      name
+    }
+  }
+}
+```
+We've nested `tasks` within the `flow` object, which tells the server to retrieve tasks only within the context of each flow that it returns.
+### Limits and Offset
```
joshmeek:
```suggestion
We've nested `tasks` within the `flow` object, which tells the server to retrieve tasks only within the context of each flow that it returns.

### Limits and Offset
```

docs/cloud/ui/interactive-api.md
```diff
@@ -1,6 +1,116 @@
...
+  flow {
+    name 
+    id
+    tasks {
+      name
+    }
+  }
+}
+```
+We've nested `tasks` within the `flow` object, which tells the server to retrieve tasks only within the context of each flow that it returns.
```
joshmeek:
```suggestion
```

We've nested `tasks` within the `flow` object, which tells the server to retrieve tasks only within the context of each flow that it returns.
```

docs/cloud/ui/interactive-api.md
joshmeek: This actually does not affect rendering

docs/cloud/ui/interactive-api.md
joshmeek: This actually does not affect rendering

PR #2130, opened 2020-03-09
(Open to comment) PIN: Obfuscating Data Passed to Cloud 

lauralorenz:

## What does this PR change?
This PR adds a new PIN focused on how to address obfuscating data passed to Cloud.
This is currently a DRAFT seeking comment from the community. Note that the prose is expanded into individual bullets to enable per-line Github comments for debate.
For more context, see a prior conversation on a POC PR, #2048 

## Why is this PR important?
@nathan5280 recently reinvigorated conversation about privacy in Cloud for compliance-focused users, which I understand is at least anecdotally known to be of interest to more Cloud users.
This PR seeks to consolidate the conversation on this into one place and open up the concept for discussion and debate to the broader Core community.

cc @nathan5280, @wagoodman

### DIFF
--- a/docs/core/PINs/PIN-16-Obfuscating-Data-Passed-to-Cloud.md
+++ b/docs/core/PINs/PIN-16-Obfuscating-Data-Passed-to-Cloud.md
@@ -1,0 +1,161 @@
+---
+title: 'PIN-16: Obfuscating Data Passed to Cloud'
+sidebarDepth: 0
+---
+
+# PIN-16: Obfuscating Data Passed to Cloud
+
+Date: March 6, 2020
+
+Authors: Laura Lorenz, Nathan Atkins, Alex Goodman
+
+## Status
+Draft
+
+## Context
+
+Businesses that deal with customer, medical or otherwise sensitive data are concerned about sending that data outside of their controlled environment. This includes exception messages and stack traces, partial or entire log messages that contain sensitive information, and task-produced data such as Result values being sent to Prefect Cloud. 
+
+Despite these concerns, there are some users who wish to still use features from Prefect Cloud if they can feel more confident that the data Cloud stores to provide those features is non-sensitive or obfuscated from the cloud database if so.
+
+Though the hybrid model provides protection from most of the canonical surface area of this risk by isolating user code from Cloud persistence systems, some specific types of leaks are still possible and even enforced by the library today. Currently, the Prefect logger can be bypassed if users always log to their own logger instance, but this is not immediately obvious and does not prevent exceptions from being sent through the Prefect logger to Cloud since the exception is ultimately raised during the Prefect TaskRunner. In addition, any accidental usage of JSONResultHandler (including today’s hardcoded case for Parameter type tasks) emits the return value of a task to the cloud database.
+
+This PIN seeks to argue that more configuration options, including a set-it-and-forget-it global configuration, that protects the data boundary to Cloud is a valid “safe default” scenario that should be provided alongside the current safe default scenario that prioritizes convenience.
+
+In addition, we also considered the need for dynamic obfuscation of logs by Prefect administrators that want peace of mind that data they can infer to be sensitive, such as logs containing certain terms or strings that look like keys, can be interrupted during any accidental transmission to Cloud.
+
+Here are the obfuscation / redaction scenarios that we considered:
+
+1. **Full Redaction**: total obfuscation of information in a LogRecord besides log level. The obfuscated LogRecord is still delivered to the logging destination.
+```
+   logger.info("this is sensitive", private=True)
+   |
+   |
+   Logged: “INFO: ***************”
+```
+2. **Partial Redaction**: search and replace partial strings from individual LogRecords. 
+```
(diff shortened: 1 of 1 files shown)


docs/core/PINs/PIN-16-Obfuscating-Data-Passed-to-Cloud.md
```diff
@@ -0,0 +1,161 @@
...
+Authors: Laura Lorenz, Nathan Atkins, Alex Goodman
+
+## Status
+Draft
+
+## Context
+
+Businesses that deal with customer, medical or otherwise sensitive data are concerned about sending that data outside of their controlled environment. This includes exception messages and stack traces, partial or entire log messages that contain sensitive information, and task-produced data such as Result values being sent to Prefect Cloud. 
+
+Despite these concerns, there are some users who wish to still use features from Prefect Cloud if they can feel more confident that the data Cloud stores to provide those features is non-sensitive or obfuscated from the cloud database if so.
```
jlowin:
```suggestion
Despite these concerns, there are some users who wish to still use features from Prefect Cloud if they can feel more confident that the data Cloud stores to provide those features is non-sensitive or obfuscated from the Cloud database if so.
```

docs/core/PINs/PIN-16-Obfuscating-Data-Passed-to-Cloud.md
```diff
@@ -0,0 +1,161 @@
...
+
+Though the hybrid model provides protection from most of the canonical surface area of this risk by isolating user code from Cloud persistence systems, some specific types of leaks are still possible and even enforced by the library today. Currently, the Prefect logger can be bypassed if users always log to their own logger instance, but this is not immediately obvious and does not prevent exceptions from being sent through the Prefect logger to Cloud since the exception is ultimately raised during the Prefect TaskRunner. In addition, any accidental usage of JSONResultHandler (including today’s hardcoded case for Parameter type tasks) emits the return value of a task to the cloud database.
+
+This PIN seeks to argue that more configuration options, including a set-it-and-forget-it global configuration, that protects the data boundary to Cloud is a valid “safe default” scenario that should be provided alongside the current safe default scenario that prioritizes convenience.
```
jlowin: This sentence is a little confusing - I think I get the gist but it might benefit from being split into two sentences.

docs/core/PINs/PIN-16-Obfuscating-Data-Passed-to-Cloud.md
```diff
@@ -0,0 +1,161 @@
...
+### Design goals
+- Obfuscation behavior driven by prefect configuration: A global setting can obfuscate all task logs and exception traces from Cloud.
+- Obfuscation behavior customizable at many levels: Redaction patterns, obfuscation functions, and setting privacy at log levels or per handler allow the user to customize obfuscation behavior.
+- Warn users as early as possible if obfuscation mechanisms have been elected but there could be a known potential leak of information.
+- Utilize as much of the Python standard logging subsystem as possible
+- Don’t break the existing Cloud UI: The error log tile functionality in Cloud UI is not broken; it can still report number of errors and link to obfuscated error logs in Cloud UI like it does today
+
+### Primary proposals:
+#### Obfuscate Logs sent to Prefect Cloud
+- User can elect all logs to be obfuscated in entirety, for example by setting `prefect.config.logger.set_all_private`. This would still send a LogRecord to Cloud indicating the log level, however, all content from a record’s `msg` attribute would be removed. This is useful for the Cloud UI to determine if there is a log entry indicating a failure within a Flow Run.
```
jlowin: Could we consider using a variable with a fixed number of settings rather than a boolean? I'm imagining `prefect.config.logger.redaction_level = {'NONE', 'PRIVATE', 'ALL', 'PARTIAL', 'SOME_OTHER_LEVEL'}` etc. in order to allow this one config to represent possibly multiple default levels.

jlowin:
This is a really cool feature and I love how it's evolving.

As I understand it, the filtering mechanism would live inside the method that sends logs to Cloud, so the default setting would be applied to non-Prefect loggers as well - is that correct? If so I can also imagine a use case in which users set the default (for all loggers, globally) to be private but want to opt-in selectively to sending logs. In the current design I'm not sure if that's possible -- it seems to be either opt-out selectively by setting `private=True` or opt-out globally via config.

wagoodman:
@jlowin filtering has not been included as part of this PIN (though it was considered):

> two scenarios were specifically dropped from consideration. (6) and (7) protect information by filtering LogRecords entirely from a handler’s emit method

Instead only redaction-based methods are proposed for now. Each LogRecord going to Cloud specifically would have these redaction methods / patterns applied. If the user elects to use other loggers and handlers configuring the prefect logger via `prefect.config.logger.*` should not affect other loggers (from a redaction perspective). 

One of the stated goals is to use as much of the standard lib logging facilities as possible, and depending on how the implementation shakes out, it should be feasible for others to use any standard logging components that we make (Formatters, Filters, Handlers, etc) to redact LogRecords to other logging destinations (the users would need to use them explicitly though).

PR #2134, opened 2020-03-09
Closes #2131 Add PIN for Result update for targets

### LINKED ISSUE #2131: Publish PIN 16: Results and Artifacts
Once PR'ed, we'll break it down into concrete issues.

lauralorenz:

- [n/a] updates `CHANGELOG.md` (if appropriate)

## What does this PR change?
Closes #2131. Adds a PIN describing the goals of an internal roadmap project sophisticating the Result interface.

## Why is this PR important?
We want to be able to support several use cases related to Prefect knowing about where file targets exist, including tracking task artifacts, file dependencies to determine state, and persisting the cache for core only users outside of the in memory cache. We believe this PINs suggested API allow for all of those situations to be achievable through a common interface.

### DIFF
--- a/docs/core/PINs/PIN-16-Results-and-Targets.md
+++ b/docs/core/PINs/PIN-16-Results-and-Targets.md
@@ -1,0 +1,49 @@
+---
+title: 'PIN-16: Results and Targets'
+sidebarDepth: 0
+---
+# PIN 16: Results and Targets
+
+Date: March 4, 2020
+
+Authors: Chris White & Laura Lorenz
+
+## Status
+Accepted
+## Context
+Currently, Prefect tracks all data explicitly returned by tasks for downstream consumption.  In addition, Prefect allows for caching the success of a task based on a certain condition (specified by a `cache_validator`) over a certain time frame (specified by `cache_for` on the task).  Prefect also currently allows for the data returned by individual tasks to be stored in diverse, configurable locations via the `ResultHandler` interface.  Despite the configurable nature of the result handler and caching APIs, there are still a few use cases that are not sufficiently covered within Prefect's API:
+- "Makefile"-like semantics which allow the existence of data in a certain location to prevent the rerun of a task
+- Prefect has no way of tracking data produced as side-effects of tasks
+- Prefect exposes no first-class hooks for validating data (whether returned from a task or produced by a task during the run)
+- Prefect output caches exist only in-memory for Core-only users, so cached computations are not recoverable between Python processes
+
+
+While most of this can technically still be achieved within Prefect, the popularity of these patterns along with the lack of an obvious "correct" way to achieve them in Prefect leaves room for improving Prefect's API to accommodate such patterns in a first-class way.
+
+## Proposal
+The current proposal seeks to address all use cases listed above with a common underlying implementation.  In particular, we will extend the current [`Result` class API](https://docs.prefect.io/core/PINs/PIN-04-Result-Objects.html) with the following new methods and settings:
+- we will introduce new `Result` types which correspond to different storage backends for data (e.g., `LocalFileResult`, `S3Result`, etc.)
+- new `read` / `write` methods which mimic the current [`ResultHandler` implementations](https://docs.prefect.io/core/PINs/PIN-02-Result-Handlers.html) and will supercede the need for result handlers altogether
+- the initialization of `Result` objects will allow for setting a configurable and templatable location (e.g., a templatable file name), along with a collection of validation functions which will optionally be called on the underlying object
+- a new `exists` method for determining whether data exists in a given location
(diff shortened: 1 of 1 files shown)


wagoodman:
Love where this is going, nice work! I have a few questions:

- The current read interface is `ResultHandler.read(location: str)`. Will we be keeping the `location` as a parameter to `read()`? It appears the answer is "no"  and that the location templating is taking care of this. If this is the case, then I feel that we should have more discussion on who owns "where" the result is called. That it sounds like we're moving towards result locations being "caller-influenced" instead of "caller-controlled".

- Is `Result.validate` intended to be a generic validator of a result contents at the end of a Task run, regardless of the backend storage? Or is it intended to check the integrity of data fetched from a storage backend? It seems like there may be a specific intended use cases for `Result.validate` that would be worth explicitly calling out.

- It looks like constructor of Result types are where users are able to provide additional behavioral changes (..."the initialization of Result objects will allow for setting a configurable and templatable location", and "provided validators along with those set at initialization will then be called on the underlying data to validate it in customizable ways"). Are we planning to expose the constructor of `Result` types to an arbitrarily wide set of arguments? Currently it seems that we will only be forwarding select items to `Result` constructors. (This would impede folks that would want to bring their own `Result` implementation that varies based on constructor args)

- Can you speak more to what should be provided to `target`? Is it a Result instance or a Result class type (or neither)? I see your comment on `target: Result` (which implies a Result instance, not a class?), which would override the Result type instantiated by the `TaskRunner`, however, later there is a comment in Actions about a new `Task` constructor argument "`result` or `result_type`" . Does `target` take a `Result` instance or a `Result` class (to later instantiate)?  How does this vary from `Task.__init__(result...)` / `Task.__init__(result_type...)` ? Is the default value `None` (in which case `exists`/`read` is not called... thus we do not depend on cached results)?

- if there is a `target` specified, we will call `Result.exists` --if a previous result exists, we will use it as a cached result and set the Task to `Success` (but not run the Task). Is there a specific reason to use a `Success` state over a `Cached` state?  I realize that there is no use of `cache_for`, however, it is somewhat surprising to a new user that we would have a distinction between `Success` and `Cached` states in Core and decide to use `Success` in a case where we are leveraging cached data.

jlowin:
I'm such a fan of this proposal, it ties up so many use cases in an elegant way that have terrible implementations in other systems.

- Luigi's conflation of data and state via avoided
- Airflow's inability to cache is solved
- Makefile's inability to express complex dataflow is solved by the rest of Core
- Prefect's own local cache will extend across sessions
- Progress reporting becomes possible (unaware of any system that supports this)

One thing that occurred to me while reading the proposal is that this also provides us a natural way to type outputs, which has escaped us thus far (though we have a weak mechanism for it). For example, a task's `Result` could indicate it returns an `int`, which could enable some basic default validation.

cicdw:
@wagoodman some of your questions are more implementation specific and as such don't need to be addressed in the PIN; however, here's a quick walkthrough:

- because `Result`s will be initialized with a _templatable_ location, we will most likely need to preserve the signature of `read` to allow for a specific location
- `Result.validate` is currently expected to be a method that walks through all `validators` and calls them on the underlying data; whether those validators also interact with the backend is open to future discussion
- we could allow for `**kwargs` on `Result`s to cover compatibility with custom implementations
- `target` will need to take an instance, as it will need to have a templatable location specified along with any validators
- I agree that we should use a `Cached` state whenever a `target` is used so that the state carries richer information to users about when / where the cache gets used

docs/core/PINs/PIN-16-Results-and-Targets.md
```diff
@@ -0,0 +1,49 @@
...
+- `cache_for` and `cache_validators` can be set on the Result, allowing for output cache behaviors to be exposed using this same interface
+
+With this new collection of result "types", we can now introduce the following new features / settings on Tasks and Flows:
+- Flows can be provided with a default Result type for its tasks
+- individual tasks can optionally override the default Flow type (identical to the current way result handlers work)
+- tasks can receive an optional set of validation functions under the `validators` kwarg which will be appended to the validators attached to the task's result type
+- when `validate=True` on the `Result` type, the `validate` method described above will be called as an additional pipeline step in the `TaskRunner`; if validation fails, the task run will enter a new `ValidationError` failed state
+- a new `target: Result` kwarg on tasks which serves two purposes: a) the output result type can be overriden through this kwarg and b) the new `exists` method will be called as a pre-run pipeline check; if the result exists, it will be `read` and optionally validated, and the task run will then enter a `Success` state with the data that was read
```
cicdw:
```suggestion
- a new `target: Result` kwarg on tasks which serves two purposes: a) the output result type can be overriden through this kwarg and b) the new `exists` method will be called as a pre-run pipeline check; if the result exists, it will be `read` and optionally validated, and the task run will then enter a `Cached` state with the data that was read
```

PR #2136, opened 2020-03-09
Closes #2108. New boto session for each thread

### LINKED ISSUE #2108: Flow S3ResultHandler Fails for Dask Worker with nthreads > 1
## Description
Specifying S3ResultHandler for a Flow running on Dask worker(s) with nthreads > 1 fails with: `KeyError: 'credential_provider'`, likely due to a race condition in using the global boto3 session (boto3.client) between threads. 

## Expected Behavior
In a multithreaded environment, boto3 recommends creating a session per thread rather than sharing the default boto3 session, i.e. boto3.client. See boto3 documentation at: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing

This thread in Prefect's Community Slack describes using this session-per-thread approach to successfully fix a similar issue when using boto3 in Prefect tasks: https://prefect-community.slack.com/archives/CM28LL405/p1581434710167100

## Reproduction
A Flow with tasks that can run in parallel (e.g. mapped tasks or different Flow branches) and where the Flow-level result_handler is set to S3ResultHandler should reproduce this behavior.

Full stack trace:
```
February 29th 2020 at 8:09:43am | prefect.CloudTaskRunner
ERROR 
Failed to set task state with error: KeyError('credential_provider')
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/cloud/task_runner.py", line 117, in call_runner_target_handlers
    cloud_state = prepare_state_for_cloud(new_state)
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/cloud/utilities.py", line 21, in prepare_state_for_cloud
    res.store_safe_value()
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/result.py", line 93, in store_safe_value
    value = self.result_handler.write(self.value)
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/result_handlers/s3_result_handler.py", line 103, in write
    self.client.upload_fileobj(stream, Bucket=self.bucket, Key=uri)
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/result_handlers/s3_result_handler.py", line 67, in client
    self.initialize_client()
  File "/opt/conda/lib/python3.7/site-packages/prefect/engine/result_handlers/s3_result_handler.py", line 60, in initialize_client
    aws_secret_access_key=aws_secret_access_key,
  File "/opt/conda/lib/python3.7/site-packages/boto3/__init__.py", line 91, in client
    return _get_default_session().client(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/opt/conda/lib/python3.7/site-packages/botocore/session.py", line 823, in create_client
    credentials = self.get_credentials()
  File "/opt/conda/lib/python3.7/site-packages/botocore/session.py", line 428, in get_credentials
    'credential_provider').load_credentials()
  File "/opt/conda/lib/python3.7/site-packages/botocore/session.py", line 923, in get_component
    del self._deferred[name]
KeyError: 'credential_provider'
```

## Environment
We create a long-running Dask cluster where our Dask workers are started with --nprocs 1 --nthreads 3.

(Thanks to @JLouSRM for identifying this issue and capturing log evidence!)

lauralorenz:

## What does this PR change?
Instantiates a new boto session (and thus boto client) if we think we are in a new thread. Closes #2108 which has a lot of context for this. In the end (see commit history for the history) I went with memoizing the boto client not on the result handler instance as before, but in `prefect.context`. Anecdotally, initializing at every call to `S3ResultHandler.client` a la 0852c2c made the [dask example flow](https://docs.prefect.io/core/advanced_tutorials/dask-cluster.html#an-example-flow) twice as slow with two local dask workers using `--nprocs 1 --nthreads 3`.

## Why is this PR important?
Tasks on multithreaded executors can now use the S3ResultHandler (before tasks would fail when initializing the boto client).

❓ **NOTE** ❓ : Should I be attempting to add an integration test testing this edge between Dask executor, mapped flows and S3?

### DIFF
--- a/src/prefect/engine/result_handlers/s3_result_handler.py
+++ b/src/prefect/engine/result_handlers/s3_result_handler.py
@@ -65,9 +65,9 @@ class S3ResultHandler(ResultHandler):
 
     @property
     def client(self) -> "boto3.client":
-        if not prefect.context.caches.get("boto3client"):
+        if not prefect.context.get("boto3client"):
             self.initialize_client()
-            prefect.context.caches["boto3client"] = self._client
+            prefect.context["boto3client"] = self._client
         return self._client
 
     @client.setter
--- a/src/prefect/engine/result_handlers/s3_result_handler.py
+++ b/src/prefect/engine/result_handlers/s3_result_handler.py
@@ -54,7 +54,8 @@ class S3ResultHandler(ResultHandler):
             aws_access_key = aws_credentials["ACCESS_KEY"]
             aws_secret_access_key = aws_credentials["SECRET_ACCESS_KEY"]
 
-        s3_client = boto3.client(
+        session = boto3.session.Session()
+        s3_client = session.client(
             "s3",
             aws_access_key_id=aws_access_key,
             aws_secret_access_key=aws_secret_access_key,
@@ -63,8 +64,8 @@ class S3ResultHandler(ResultHandler):
 
     @property
     def client(self) -> "boto3.client":
-        if not hasattr(self, "_client"):
-            self.initialize_client()
+        # reinitialize every time it is requested in case we are in different threads
+        self.initialize_client()
         return self._client
 
     @client.setter
--- a/src/prefect/engine/result_handlers/s3_result_handler.py
+++ b/src/prefect/engine/result_handlers/s3_result_handler.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any
 import cloudpickle
 import pendulum
 
+import prefect
 from prefect.client import Secret
 from prefect.engine.result_handlers import ResultHandler
 
@@ -64,8 +65,9 @@ class S3ResultHandler(ResultHandler):
 
     @property
     def client(self) -> "boto3.client":
-        # reinitialize every time it is requested in case we are in different threads
-        self.initialize_client()
+        if not prefect.context.caches.get("boto3client"):
+            self.initialize_client()
+            prefect.context.caches["boto3client"] = self._client
         return self._client
 
     @client.setter
--- a/src/prefect/engine/result_handlers/s3_result_handler.py
+++ b/src/prefect/engine/result_handlers/s3_result_handler.py
@@ -55,6 +55,8 @@ class S3ResultHandler(ResultHandler):
             aws_access_key = aws_credentials["ACCESS_KEY"]
             aws_secret_access_key = aws_credentials["SECRET_ACCESS_KEY"]
 
+        # use a new boto session when initializing in case we are in a new thread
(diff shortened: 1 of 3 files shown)


src/prefect/engine/result_handlers/s3_result_handler.py
```diff
@@ -63,8 +65,9 @@ def initialize_client(self) -> None:
 
     @property
     def client(self) -> "boto3.client":
-        if not hasattr(self, "_client"):
+        if not prefect.context.get("boto3client"):
             self.initialize_client()
+            prefect.context["boto3client"] = self._client
```
cicdw: Could you elaborate on why you are putting this in context instead of referencing the attribute?  I can see why this works, but in general modifying context outside of a runner pipeline could lead to unexpected behavior or in this case redundant computation (e.g., if this gets added to context deep in a `TaskRunner` pipeline, it won't be available to the next task because `context` is generally read-only).

lauralorenz: I thought the point was to reinstantiate the session AND client anytime we were in a new thread, which could be per task run (and thus in my mind, per Prefect context). Since you’re asking I’m assuming I’m misunderstanding what “global boto session” means in the issue OP and/or how exactly resources are shared between tasks on dask

cicdw: Yea, that definitely makes sense; my knee-jerk reaction is that modifying context outside of a pipeline is a smell, but in this case because the context is thread-local I agree it's the cleanest way to detect when running in a new thread so I'm 👍

lauralorenz: @cicdw Gotcha yeah I had similar misgivings but couldn't figure out another way from the perspective of result handler to determine if it was likely we were in another thread without adding a dependency (and myself possibly needing to know more) about executor submission.

I will update the changelog, but re: testing do you mind commenting on whether
> Should I be attempting to add an integration test testing this edge between Dask executor, mapped flows and S3?

cicdw: Hmmm at this moment I'd say "no" - the overhead of mocking out the s3 calls across the threads / processes feels too heavy to me, unless you see a clear way to do it.

lauralorenz: Gotcha thanks -- I don't see a clear way commensurate with the PR size, but I can add it to some discussion points for us tomorrow at blue sky. Thanks 🙏

Issue #2124, opened 2020-03-05
MySQL tasks


lauralorenz:
## Use Case
Users who are working with data in MySQL databases may want to add steps in their flows that execute queries or return results from a MySQL connection.

## Solution
A task or set of tasks that can configure a python client to MySQL and execute (and/or fetch) queries. This should be similar to the implementation of https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/postgres/postgres.py but using 
[mysqlclient](https://pypi.org/project/mysqlclient/) or [mysql-connector-python](https://pypi.org/project/mysql-connector-python/) (see [a debate about that on stackoverflow here](https://stackoverflow.com/questions/43102442/whats-the-difference-between-mysqldb-mysqlclient-and-mysql-connector-python)). The client should only be installed as an [extra, like psycopg is](https://github.com/PrefectHQ/prefect/blob/b49db66c23132dea9471663026ada21e80b1b2dc/setup.py#L51).

## Alternatives
At some point as the collection of DB access tasks grows, we may consider abstracting them given their clients should all be following [PEP 249](https://www.python.org/dev/peps/pep-0249/).

manesioz:
Hey, I opened up a WIP PR for this :) I'll try to wrap it up as soon as possible and ask for a review

lauralorenz:
@manesioz awesome, thanks! Will be on the lookout :)

manesioz:
Should this be closed now? :)

joshmeek:
Not Laura but 👍

lauralorenz:
Thanks @manesioz and @joshmeek ! Yay! :)

PR #2142, opened 2020-03-10
#2124: MySQL task 

manesioz:

WIP PR for #2124 

## What does this PR change?
Adds new task that enables users to execute and fetch queries from a MySQL DB. 

## Why is this PR important?
Increase useability due to integration with popular DB.


lauralorenz:
@manesioz Just a heads up that master has a fix in it to some tests that were failing due to an upgrade in an underlying dependency, you might want to merge/rebase master in before you get too detailed on the test failures to avoid any CircleCI noise!

manesioz:
hey @lauralorenz I seem to be failing the circle-ci pipeline due to the following error: 
```sh
_______________ ERROR collecting tests/tasks/mysql/test_mysql.py _______________
ImportError while importing test module '/root/project/tests/tasks/mysql/test_mysql.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src/prefect/tasks/mysql/__init__.py:7: in <module>
    from prefect.tasks.mysql.mysql import MySQLExecute, MySQLFetch
src/prefect/tasks/mysql/mysql.py:1: in <module>
    import pymysql.cursors
E   ModuleNotFoundError: No module named 'pymysql'
```
Despite including `mypysql` in the `extras` of the `setup.py` file, and the tests passing locally. 

Before I continue further testing I was wondering if you had any idea why the environment doesn't seem to be picking up the dependency? Sorry, I don't have much experience with circle-ci.

joshmeek:
@manesioz for the `tests/tasks/mysql/__init__.py` add a pytest `importorskip` similar to https://github.com/PrefectHQ/prefect/blob/master/tests/tasks/snowflake/__init__.py 😄

manesioz:
Ah of course, thanks @joshmeek, sorry for overlooking that!

manesioz:
Hey @lauralorenz @joshmeek I think I have the first iteration done, seems to work fine locally so if you want to review whenever you have time feel free :)

lauralorenz:
@manesioz Hi! FYI I did start kicking the tires on this, I haven't finished yet but here is my setup:

* start a docker container of `mysql:latest`
* run this basic flow:

```python
import prefect
from prefect.tasks.mysql import MySQLExecute, MySQLFetch
from prefect import Flow

flow = Flow('my flow')
settings = {
'host': 'localhost',
'port': 3307,
'db_name': 'mysql',
'password': 'secret',
'user': 'root'
}
settings.update({'query': "SELECT * FROM user"})
s = MySQLFetch(**settings)
settings.update({'query': "SHOW databases;"})
r = MySQLExecute(**settings)

flow.add_task(s)
flow.add_task(r)
```

I got a couple errors, starting off with missing `cryptography`, and then charset not being passed to MySQLFetch. So there are a few little bits missing that might benefit from actually running this use case.

In addition I think we might be able to come up with a few more test cases, and potentially some integration tests for this if we want to go that way. For task library stuff I think that adding in more integration tests against docker containers would really increase confidence in the task library pieces. The postgres inspiration task doesn't really have these, but we could start layering that in here. I think we could add in some more test dependencies for this, like something that integrates docker containers with pytest, to set the stage for all task library tests if we want.

src/prefect/tasks/mysql/mysql.py
```diff
@@ -0,0 +1,192 @@
...
+            Task constructor
+    """
+
+    def __init__(
+        self,
+        db_name: str,
+        user: str,
+        password: str,
+        host: str,
+        port: int = 5432,
```
lauralorenz: FYI the default in these spots should be 3307, as that is the default mysql port

src/prefect/tasks/mysql/mysql.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    Task for executing a query against a MySQL database. 
+    Args:
+        - db_name (str): name of MySQL database
+        - user (str): user name used to authenticate
+        - password (str): password used to authenticate
+        - host (str): database host address
+        - port (int, optional): port used to connect to MySQL database, defaults to 5432 if not provided
+        - query (str, optional): query to execute against database
+        - data (tuple, optional): values to use in query, must be specified using placeholder in query string
```
lauralorenz, line 17:
```suggestion
        - data (tuple, optional, UNUSED): values to use in query, must be specified using placeholder in query string
```
Right now the `data` parameter is not being used by the implementation here, which I think we should either
a) fix in this PR
b) fix in a later PR, but remove the kwargs from here
c) fix in a later PR, and annotate the kwargs from this PR that they are unused

This github suggestion goes for (c) because it's the easiest to make as a one-line suggestion, but any of the 3 work for me depending what you want to tackle!

src/prefect/tasks/mysql/mysql.py
```diff
@@ -0,0 +1,204 @@
...
+
+        try:
+            with conn:
+                with conn.cursor() as cursor:
+                    executed = cursor.execute(query)
+                    if commit:
+                        conm.commit()
+
+            conn.close()
+            print("Execute Results: ", executed)
```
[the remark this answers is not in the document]
manesioz: You're absolutely right, I used it for debugging and forgot to port it over to proper logging. Thanks!

src/prefect/tasks/mysql/mysql.py
manesioz: Good catch, sorry about that. I am going to remove it for this PR as it is not necessary, and was used previously for templating strings (but Python has evolved better ways of templating, such as `.format()` and f-strings)

manesioz:
Hey @lauralorenz , thanks for the review! Sorry for not being as responsive, things have been hectic lately. I hope I addressed your comments sufficiently, and regarding the integration testing I was wondering if I could be pointed towards some examples of what you mean? 
I don't have much experience with database integration tests in Python so I am not quite sure what the best practices are.

bratao:
@manesioz @lauralorenz Is there a chance of MySQL task becoming official?

lauralorenz:
Hi, yes @bratao!

In retrospect I did this kind of overly complicated but I made some changes including an edit I just made adding the changelog in a PR https://github.com/manesioz/prefect/pull/1 to your remote @manesioz. I did that to keep them in queue for comment while I was waiting for the changelog updates but now I realized that I can't merge that PR, though I can make the commit changes to the branch directly 😅 So if I don't hear otherwise shortly I can just cherry pick them in.

manesioz:
Thanks @lauralorenz ! I appreciate your guidance and help with this.

Issue #2144, opened 2020-03-10
PIN 16: Implement `Result.validate`


lauralorenz:
## Use Case
~Blocked by #2143 and #2139~. See [PIN 16](https://github.com/PrefectHQ/prefect/pull/2134) for background. 

## Solution
Implement `Result.validate` such that
- if `Result.run_validators=True`
    - all `Result.validators` are called on the result value
    - if any fail, raise ValidationFailed signal to trigger ValidationFailed state (*see alternative)
    - if all succeed, return True (*see alternative)

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*
- *I may be misrepresenting the intended method signature; if this method will be called by a TaskRunner method, we may want to receive `False` instead of a raise a ValidationFailed signal to allow the TaskRunner method to construct the ValidationFailed state and return it as part of the pipeline a la https://github.com/PrefectHQ/prefect/blob/66703e5f0a2d0d035d4d48af013b87d24a7a2a56/src/prefect/engine/task_runner.py#L957-L964

PR #2147, opened 2020-03-11
Pin 16 result interface extension

### LINKED ISSUE #2139: PIN 16: Extend `Result` interface
## Use Case
See PIN 16 (PR #2134) for background.
This issue includes only extensions to the `Result` interface from the PIN that are independently deliverable from other issues.

## Solution
Update the `Result` base class with the following init kwargs:
- `validators: Iterable[Callable]`
- `should_validate_result: bool`
- `cache_for: datetime.timedelta`
- `cache_validator: Callable`
- `filename_template: str`

and the following new methods:
- `exists -> bool`, not implemented in the base class
- ~`validate -> bool`, don't implement it yet~ (this will be added in #2144 to await ValidationFailed state in case we need it at that time)
- `read`, not implemented in the base class
- `write`, not implemented in the base class

alexisprince1994:

NOTE: It didn't appear that anything needed to be changed in `docs/outline.toml`.

Closes #2139 

## What does this PR change?
This PR implements the methods `exists`, `read`, and `write` on the `Result` object in the core engine. Each of these methods are abstract and should be overriden by the specific target's `Result` object, which have not yet been implemented.

## Why is this PR important?
This PR is the starting point for implementing a more unified `Result` interface, allowing users of the engine to both control caching and storage of a result in one interface. It is also the starting point for the inclusion of validating results.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 ### Enhancements
 
 - - Add examples to Interactive API Docs [#2122](https://github.com/PrefectHQ/prefect/pull/2122)
+- - Add exists, read, and write interfaces to Result [#2139]
```
lauralorenz:
```suggestion
- - Add exists, read, and write interfaces to Result [#2139](https://github.com/PrefectHQ/prefect/issues/2139)
```

lauralorenz:
Woo thank you @alexisprince1994! I agree on your note in your PR that `docs/outline.toml` doesn't need an update; I do think we need to update the module and class docstring on `Result` by the time we get to a release and/or PIN 16 is done but I can handle tracking that. 

I'm going to strikeout stubbing a NonImplemented `Result.validate` from the issue OP #2139 since it can be handled entirely in the blocked issue #2144.

@cicdw last call to reject my attribute naming decisions :) `Result.should_validate_result` (we used to call this `Result.validate` but that collides with the method name) and `Result.filename_template` in particular

wagoodman:
- I'd vote `Result.should_validate` (we don't need to specify what we are validating since it is on a result object)
- Though this is an abstract interface and parameters don't matter yet, will we want to add a `location` parameter to `read()`? I understand that a `filename_template` will be driving this, however, are we intentionally changing the interface for `read()`? (The PIN specifies to mimic the ResultHandler implementation)

alexisprince1994:
@wagoodman That's a good question. I'd think it comes down to how does the `filename_template` portion fit into all of this? If we would have `location` as optional for `read`, my assumption as a user of the library would be that it would behave like most `Task` objects do, in that the value provided to the constructor would be used unless a value is passed directly to `read`, like the below, where `make_location` is some arbitrary function to turn the template into a `location`:

```
class Result:
    def __init__(filename_template: Optional[str] = None):
        self.filename_template = filename_template

    def read(self, location: Optional[str] = None):
        if location is None:
            location = make_location(self.filename_template)
```

In your comment, you noted that `filename_template` will be driving this, so I'm not too sure whether you mean what I described in the code chunk above, whether `filename_template` will always be used and we'd be adding `location` to `read` for backwards compatibility and migration purposes, or some third option I'm not thinking of yet.

cicdw:
- For the `validate` kwarg, I'm generally not a fan of "helping" verbs like "should" -- how about `run_validators`?
- I agree with @wagoodman we should keep the `location` kwarg to the `Reslult.read` method - this allows us to hydrate the `Result` object without knowing the templated filename.

src/prefect/engine/result.py
```diff
@@ -71,12 +73,39 @@ class Result(ResultInterface):
         - value (Any): the value of the result
         - result_handler (ResultHandler, optional): the result handler to use
             when storing / serializing this result's value; required if you intend on persisting this result in some way
+        - validators (Iterable[Callable], optional): Iterable of validation functions to apply to
+            the result to ensure it is `valid`.
+        - should_validate_result (bool): Whether the result value should be validated.
```
lauralorenz:
```suggestion
        - run_validators (bool): Whether the result value should be validated.
```

src/prefect/engine/result.py
```diff
@@ -71,12 +73,39 @@ class Result(ResultInterface):
...
+            result to the destination.
     """
 
-    def __init__(self, value: Any, result_handler: ResultHandler = None):
+    def __init__(
+        self,
+        value: Any,
+        result_handler: ResultHandler = None,
+        validators: Iterable[Callable] = None,
+        should_validate_result: bool = True,
```
lauralorenz:
```suggestion
        run_validators: bool = True,
```

src/prefect/engine/result.py
```diff
@@ -71,12 +73,39 @@ class Result(ResultInterface):
...
+        cache_for: Optional[datetime.timedelta] = None,
+        cache_validator: Optional[Callable] = None,
+        filename_template: Optional[str] = None,
+    ):
         self.value = value
         self.safe_value = NoResult  # type: SafeResult
         self.result_handler = result_handler  # type: ignore
+        self.validators = validators
+        self.should_validate_result = should_validate_result
```
lauralorenz:
```suggestion
        self.run_validators = run_validators
```

tests/engine/test_result.py
```diff
@@ -25,11 +27,19 @@ def test_result_inits_with_value(self):
         assert r.value == 3
         assert r.safe_value is NoResult
         assert r.result_handler is None
+        assert r.validators is None
+        assert r.cache_for is None
+        assert r.cache_validator is None
+        assert r.filename_template is None
```
lauralorenz:
```suggestion
        assert r.filename_template is None
        assert r.run_validators is True
```

tests/engine/test_result.py
```diff
@@ -25,11 +27,19 @@ def test_result_inits_with_value(self):
...
 
         s = Result(value=5)
         assert s.value == 5
         assert s.safe_value is NoResult
         assert s.result_handler is None
+        assert s.validators is None
+        assert s.cache_for is None
+        assert s.cache_validator is None
+        assert s.filename_template is None
```
lauralorenz:
```suggestion
        assert s.filename_template is None
        assert r.run_validators is True
```

alexisprince1994:
@lauralorenz Added the rename suggestions.

@wagoodman / @cicdw for the parameter to be given to the `Result` object's read method, what should that look like? From what I can tell, the base `ResultHandler` has a required parameter, `loc`. Then, each of the subclassers re-implement the `read` method, all with one positional parameter, none of which are named `loc`.

Since this is the case, do we care about actually renaming the parameter to `location` from `loc`? Since the new implementation has an alternative to get the target location from `filename_template`, should this param `location` be optional?

wagoodman:
Good call, I should have listed `loc` not `location` (though "location" is inherently better). The point of my original question was to make certain we were either not changing this interface or very intentionally changing it.  Your code example was exactly what I was thinking --`loc` would override any already provided `filename_template`.

alexisprince1994:
@wagoodman @cicdw Names changed and optional `loc` parameter added to `Result.read`.

Let me know if anything else needs changing!

PR #2151, opened 2020-03-12
Add deprecation warnings for cache_* and result_handler options on Tasks and Flows

### LINKED ISSUE #2140: PIN 16: deprecate Task kwargs that are moving to `Result`
## Use Case
See PIN 16 (PR #2134) for background.

## Solution
Raise DeprecationWarnings upon use of cache_for, cache_validator, cache_key and result_handler kwargs on task initialization. (These are either directly or effectively being moved to the `Result` class during this PIN.)

## Alternatives

wagoodman:

## What does this PR change?
To support PIN 16, add deprecation warnings for cache_* and result_handler options.

Closes #2140


src/prefect/core/flow.py
```diff
@@ -159,6 +159,12 @@ def __init__(
         self.storage = storage
         self.result_handler = result_handler
 
+        if result_handler:
+            warnings.warn(
+                "DEPRECATED: the result_handler Flow option is deprecated, this is being replaced by the prefect.engine.Result object.",
```
cicdw: Maybe now is a good time to decide on the new keyword argument name for these objects so we can say that in this deprecation message?  Options:

- `result_type`
- `output_type`
- `result`
- `output_result`
- others?

**cc:** @lauralorenz

lauralorenz: Hmmm I don't love anything with `output`. Colloquially I feel it's too attached to stuff in stdout. I like `result` or `result_type`, the former if it is a result instance and the latter only if it is a reference to the class. I think we are pretty committed to the former.

Some other brainstorms... `persistence` or `persistence_type`; `return_type` (in the sense its kinda like type annotations for tasks, though this seems too narrow), `file` (feels basic but kinda true)

wagoodman: I'd vote `result` for an instance and `result_cls` for the class to be consistent with the existing nomenclature (e.g. `runner_cls`, `executor_cls`, etc).

cicdw: I'm game for `result`!  In that case, can we do the following:

- update the warning message to include a mention of this new keyword
- actually introduce the keyword on both Tasks and Flows, but don't use it for anything yet (we can specify in the docstring "currently unused" until it is actually used by something)

src/prefect/core/flow.py
```diff
@@ -145,7 +145,8 @@ def __init__(
         state_handlers: List[Callable] = None,
         on_failure: Callable = None,
         validate: bool = None,
-        result_handler: ResultHandler = None,
+        result_handler: Optional[ResultHandler] = None,
+        result: Optional[Result] = None,
```
joshmeek: Is this `result` being used anywhere?

lauralorenz: I'm for changing the deprecation warning to specify that `result_handler` will be deprecated in 0.11.0 in favor of `result` and no longer a valid kwarg in 0.12.0 (we can discuss tomorrow if those are really the true release dates, but its the current proposal). In addition we could switch it out to this stronger language in 0.11.0.

wagoodman: Agreed, we'll finalize this copy after release planning

src/prefect/core/flow.py
```diff
@@ -159,6 +160,12 @@ def __init__(
         self.storage = storage
         self.result_handler = result_handler
 
+        if result_handler:
+            warnings.warn(
+                "DEPRECATED: the result_handler Flow option is deprecated, please provide a 'prefect.engine.result.Result' object to the 'result' option instead.",
```
lauralorenz:
```suggestion
                "DEPRECATED: the result_handler Flow option will be deprecated in 0.11.0, and removed in 0.12.0, in favor of the `result` option instead.",
```

@joshmeek @wagoodman Here's my take on the copy. Then I think for 0.11.0 we can add in the original second clause to "please provide...." since according to our plans it will be the recommended arg by then.

src/prefect/core/task.py
```diff
@@ -248,6 +252,18 @@ def __init__(
...
 
+        if cache_for or cache_key or cache_validator:
+            warnings.warn(
+                "DEPRECATED: all cache_* Task options are deprecated, these functionalities have been moved to the prefect.engine.Result object.",
+                UserWarning,
+            )
+
+        if result_handler:
+            warnings.warn(
+                "DEPRECATED: the result_handler Task option is deprecated, please provide a 'prefect.engine.result.Result' object to the 'result' option instead.",
```
lauralorenz:
```suggestion
                "DEPRECATED: the result_handler Task option will be deprecated in 0.11.0, and removed in 0.12.0, in favor of the `result` option instead.",
```

src/prefect/core/task.py
```diff
@@ -248,6 +252,18 @@ def __init__(
         self.checkpoint = checkpoint
         self.result_handler = result_handler
 
+        if cache_for or cache_key or cache_validator:
+            warnings.warn(
+                "DEPRECATED: all cache_* Task options are deprecated, these functionalities have been moved to the prefect.engine.Result object.",
```
lauralorenz:
```suggestion
                "DEPRECATED: all cache_* options on a Task will be deprecated in 0.11.0, and removed in 0.12.0; the options will be moved to a Task's prefect.engine.Result object.",
```

src/prefect/core/flow.py
joshmeek: I like it 👍

PR #2152, opened 2020-03-12
Pin 16 constant result

### LINKED ISSUE #2145: PIN 16: Implement `Result` subtype mimicking `ConstantResultHandler`
## Use Case
See PIN 16 (PR #2134) for background. This is a spin-off from #2141 for only one (special) handler, `ConstantResultHandler`, which has internal implications.

## Solution
Implement a `Result` subclass for `ConstantResultHandler` that knows how to `.read`, `write`, and check if it `.exists` and operates properly for the same internal purposes as the current `ConstantResultHandler` does.

## Alternatives

alexisprince1994:

Closes #2145

## What does this PR change?
This PR allows for the option of directly using the `ConstantResult` subclass with PIN-16 in mind. Method implementations for `read` and `write` were copied from `ConstantResultHandler`, and `exists` hard-coded as returning True. If users can / do specify `NoResultType` instances as values for constants, we'd probably like to change the `exists` method to take that into account. However, if there's a rework of the `NoResultType` to inherit from `Result` and always return False, maybe we create a separate issue?

Also, slightly related, this PR adds tests for how the current `ConstantResultHandler` behaves to ensure the `ConstantResult` subclass behaves the same way.

## Why is this PR important?
This PR is important for migrating to use of a unified result interface internally for use with the flow-runner.


src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+    def read(self, arg: str) -> Any:
+        """
+        Returns the underlying value regardless of the argument passed.
+
+        Args:
+            - arg (str): an unused argument
+        """
+        return self.value
+
+    def write(self, result: Any) -> str:
```
lauralorenz:
```suggestion
    def write(self) -> str:
```

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Returns the repr of the underlying value, purely for convenience.
+
+        Args:
+            - result (Any): the result to represent
```
lauralorenz:
```suggestion
```

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+        Args:
+            - arg (str): an unused argument
+        """
+        return self.value
+
+    def write(self, result: Any) -> str:
+        """
+        Returns the repr of the underlying value, purely for convenience.
+
+        Args:
```
lauralorenz:
```suggestion
```

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Args:
+        - value (Any): the underlying value that we wish to "handle"
+    """
+
+    def __init__(self, value: Any) -> None:
+        self.value = value
+        super().__init__(value=value)
+
+    def read(self, arg: str) -> Any:
```
lauralorenz:
```suggestion
    def read(self) -> Any:
```

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Returns the underlying value regardless of the argument passed.
+
+        Args:
+            - arg (str): an unused argument
```
lauralorenz:
```suggestion
```

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, value: Any) -> None:
+        self.value = value
+        super().__init__(value=value)
+
+    def read(self, arg: str) -> Any:
+        """
+        Returns the underlying value regardless of the argument passed.
+
+        Args:
```
lauralorenz:
```suggestion
```

alexisprince1994:
Just wanting to make sure I understand the changes before making the changes.

The goal is to match the signature of the base (new) `Result` class, correct? If that's the case, then I originally misunderstood and matched the signature of the `ConstantResultHandler`, which is why those methods took arguments that they promptly ignored. In addition, wouldn't we want to keep the param for `read` since both the parent and other subclassers will be using this signature as well (the param can be ignore for the ConstantResult since it's internal use only).

If this is the case, I can definitely approve the changes on the `write` interface, since that was grabbed directly from `ConstantResultHandler` and was included out of misunderstanding.

lauralorenz:
@alexisprince1994 yes, the intent of the issue was to replicate ConstantResultHandler behavior in a subclass of the new Result interface that met the new interface's contract. I think my language in the original issue was too brief, but the idea is that this is one of several building blocks of different Result subclasses that meet the behavioral criteria of result handlers but put under the new interface. Once they're all ready, we refactor the pipeline logic to utilize `Result.read` instead of things like `Result._result.to_result`, etc,  for caching and passing data dependencies (and more features, eventually).

For more context, it is broken up like this to meet some release criteria, namely: we wanted all the new Result subclasses first and the deprecation warnings on result_handlers in a patch release, then the work on the pipeline refactor as a new minor release, and then take out result handlers altogether in another minor release (since we're still in 0.x).

alexisprince1994:
@lauralorenz Gotcha! In that case then, I'll add a new commit on top of this making the `read` parameter optional to be consistent with the interface, and accept the other change suggestions for the `write` method!

src/prefect/engine/results/constant_result.py
alexisprince1994: Updated the interface from when the change was suggested. Code was previously:
`arg: str`

In order to maintain the consistent interface with the parent `Result`, the interface has changed to make this optional.

src/prefect/engine/results/constant_result.py
alexisprince1994: See above comment

src/prefect/engine/results/constant_result.py
alexisprince1994: Changes integrated, github didn't recognize it.

src/prefect/engine/results/constant_result.py
alexisprince1994: Changes integrated, github didn't recognize it.

src/prefect/engine/results/constant_result.py
alexisprince1994: Outdated, discussed on more updated thread.

src/prefect/engine/results/constant_result.py
alexisprince1994: Outdated, discussed on more updated thread.

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,48 @@
...
+    Hook for storing and retrieving constant Python objects. Only intended to be used
+    internally.
+
+    Args:
+        - value (Any): the underlying value that we wish to "handle"
+    """
+
+    def __init__(self, value: Any) -> None:
+        self.value = value
+        super().__init__(value=value)
```
lauralorenz: We have some decisions happening in-flight over in another result handler migration PR, https://github.com/PrefectHQ/prefect/pull/2161/files. One confusing thing about merging this now is that in that other PR we are updating the Result interface to make `value` optional (I should have included/considered it in the interface ticket you worked on, but we realized afterwards about it since we expect users to instantiate Result subclasses as configuration before the value is set, unlike now).

I'm not sure if I just want to wait for #2161 to land in master before updating this, to avoid duplicating changes in the Result base class signature in both PRs, but if I had to anticipate the changes, I'd suggest:

```suggestion
    def __init__(self, value: Any = None, **kwargs) -> None:
        self.value = value
        super().__init__(value=value, **kwargs)
```
This is inspired from https://github.com/PrefectHQ/prefect/pull/2161/files#diff-c94b1139de5da7d4108afdf61930892cR31-R41.

I don't actually see either `value` not being sent or meaningful `**kwargs` to ever actually be used by ConstantResult based on how it is used in the pipeline now, but the former will be part of the base class signature eventually (at least after #2126) and the `**kwargs` thing just feels nice while we're in here, for a use case I may not be considering if a user ever did go rogue and use ConstantResult themselves.

src/prefect/engine/results/constant_result.py
```diff
@@ -0,0 +1,48 @@
...
+from prefect.engine.result import Result
+
+
+class ConstantResult(Result):
+    """
+    Hook for storing and retrieving constant Python objects. Only intended to be used
+    internally.
+
+    Args:
+        - value (Any): the underlying value that we wish to "handle"
```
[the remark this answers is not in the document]
alexisprince1994: Sure, this sounds good to me! Thanks for watching!

src/prefect/engine/results/constant_result.py
alexisprince1994: Definitely okay with adding the **kwargs! Nothing wrong with future proofing!

Something of this nature sounds fine with me, but I have a couple questions that might come out of not fully understanding its use. Is the constant result handler (and now constant result) _reserved_ for internal use? Or just a concept that happens to be used internally?

The only reason I ask is that the `Constant` versions of the results completely ignores `read` method for updating its internal value, so the only option we have to ever update a value on an instantiated `Constant` is to directly update the `value` like so:

```
constant_result = ConstantResult("my text')
constant_result.value = "my other text"
```

Do you know if there's a need to be able to update a `Constant`'s value after its already been instantiated? Instead of just creating new instances?

lauralorenz: For posterity from IRL conversation: ConstantResult should not be used externally, it kind of got backed into the Result API from the result handlers API from when we took it out of the Task API. Directly to the question, no we do not want to support updating a `Constant`s value after its already been instantiated, a new instance would need to be made, and we expect the pipeline to handle that.

lauralorenz: Hey @alexisprince1994, the PR we were waiting for is in master, and we are back on PIN 16 internally so I merged in the latest here. I think the tests will now complain about the result interface for us :) I can add in the changes to the signature later today unless you're already hanging around!

PR #2154, opened 2020-03-13
Add some clarification on two time filters docs

cicdw:
Noticed there was room for clarity in two of the schedule filters docstrings.

### DIFF
--- a/src/prefect/schedules/filters.py
+++ b/src/prefect/schedules/filters.py
@@ -121,7 +121,7 @@ def between_times(start: time, end: time) -> Callable[[datetime], bool]:
     Filter that allows events between a start time and end time
 
     For example, `between_times(start=datetime.time(14), end=datetime.time(16))` would only
-    allow runs between the hours of 2 PM and 5 PM in the given timezone.
+    allow runs between the hours of 2 PM and 4 PM in the given timezone.
 
     Args:
         - start (time): the start time
--- a/src/prefect/schedules/filters.py
+++ b/src/prefect/schedules/filters.py
@@ -100,6 +100,9 @@ def at_time(t: time) -> Callable[[datetime], bool]:
     """
     Filter that allows events that match a specific time.
 
+    For example, `at_time(datetime.time(4))` would only allow runs during the hour of 4 AM
+    in the given timezone.
+
     Args:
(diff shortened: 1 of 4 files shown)


src/prefect/schedules/filters.py
```diff
@@ -117,6 +120,9 @@ def between_times(start: time, end: time) -> Callable[[datetime], bool]:
     """
     Filter that allows events between a start time and end time
 
+    For example, `between_times(start=datetime.time(14), end=datetime.time(16))` would only
+    allow runs between the hours of 2 PM and 5 PM in the given timezone.
```
lauralorenz:
```suggestion
    allow runs between the hours of 2 PM and 4 PM in the given timezone.
```

tests/engine/test_flow_runner.py
```diff
@@ -1540,7 +1540,7 @@ def log_stuff():
     assert flow_state.result[log_stuff].result == (42, "original")
 
     time.sleep(0.75)
-    assert len(calls) == 1
+    assert len(calls) >= 1
     assert len(calls[0][0]) == 6  # actual number of logs
```
lauralorenz: This part also is nondeterministic and is failing on circle rn, sounds like we expect all 6 to be spread out amongst all of the calls first arg... so maybe safer to sum([call[0] for call in calls]) or something to get the 6 logs

src/prefect/schedules/filters.py
```diff
@@ -100,6 +100,9 @@ def at_time(t: time) -> Callable[[datetime], bool]:
     """
     Filter that allows events that match a specific time.
 
+    For example, `at_time(datetime.time(4))` would only allow runs during the hour of 4 AM
```
jlowin: I'm not sure this example is strictly correct - I think `at_time(datetime.time(4))` would only allow runs at exactly 4am, but not 4:01am.

tests/engine/test_flow_runner.py
cicdw: Updated

src/prefect/schedules/filters.py
cicdw: Ah you're correct - updated

PR #2155, opened 2020-03-13
Closes #2143. Add ValidationFailed state and signal

### LINKED ISSUE #2143: PIN 16: Implement `ValidationFailed` State
## Use Case
See [PIN 16](https://github.com/PrefectHQ/prefect/pull/2134) for background. In particular, this blocks full implementation of `Result.validate`.

## Solution
Implement a new State, `ValidationFailed`, similar to [other Prefect states](https://github.com/PrefectHQ/prefect/blob/66703e5f0a2d0d035d4d48af013b87d24a7a2a56/src/prefect/engine/state.py), that can be utilized during Result validation.
- [x] add new state
- [x] add new signal
- [x] update state docs and diagram

## Alternatives

lauralorenz:

## What does this PR change?
Closes #2143. Adds the ValidationFailed state and signal, and the updates the state diagram and tests that refer to that. Inspo from 

By my eye there are more tests to update in flow runner/task runner tests when this state/signal is actually checked/raised, but I consider that to be added when the actual pipeline logic is being worked on.

## Why is this PR important?
Sets up infrastructure so we can implement more of the behavior for validation on `Result` and in the TaskRunner (in particular unblocks implementation of `Result.validate`)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Features
 
-- None
+- Add ValidationFailed state and signal in anticipation of validating task outputs [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
```
lauralorenz: should I have put this in enhancements? 🤔

src/prefect/engine/state.py
```diff
@@ -810,6 +810,23 @@ class TriggerFailed(Failed):
...
+        - message (str or Exception, optional): Defaults to `None`. A message about the
+            state, which could be an `Exception` (or [`Signal`](signals.html)) that caused it.
+        - result (Any, optional): Defaults to `None`. A data payload for the state.
+        - cached_inputs (dict): A dictionary of input keys to fully hydrated `Result`s.
+            Used / set if the Task requires retries.
+        - context (dict, optional): A dictionary of execution context information; values
+            should be JSON compatible
+    """
+
+    color = "#ff5131"
```
lauralorenz: this is the same hex as TriggerFailed, I can't see the difference between TriggerFailed and TimedOut so not sure why they are different hex codes or if that is just by accident. But it seemed that the idea was that these known-cause failed states should be a soft orange

src/prefect/engine/state.py
```diff
@@ -810,6 +810,23 @@ class TriggerFailed(Failed):
     color = "#ff5131"
 
 
+class ValidationFailed(Failed):
```
lauralorenz: I just realized somewhere along the way I started calling this `ValidationFailed` though the PIN specifies `ValidationError`. Semantically I think they are similar but between the two I am for `ValidationFailed` since it matches the patterns of `TriggerFailed`/`ClientFailed`

CHANGELOG.md
cicdw: Yea I classify this as an enhancement; I typically think of "Features" as headline things worth socializing

src/prefect/engine/state.py
cicdw: 👍 I'm cool with it

src/prefect/engine/state.py
cicdw: Yea I think that's fine, especially since `TriggerFailed` will eventually be deprecated for `Skipped` anyway

CHANGELOG.md
lauralorenz: Gotcha, moved it in 39629e5

Issue #2039, opened 2020-02-16
Local Exception Storage


nathan5280:
## Use Case
Businesses that deal with customer, medical or otherwise sensitive data are concerned about sending that data outside of their controlled environment. This includes exception messages and stack traces being sent to Prefect Cloud.

## Solution
- Modify runner.call_state_handlers() to store the stack trace in local storage using a new LocalExceptionHandler. 
- Log the location of the exception stack trace to Prefect Cloud so that it can be displayed in the log.
- If the exception is raised, replace the exception with a new LocallyStoredError that includes the location of the locally stored exception.  The stack trace for this exception should not include the stack trace for the original exception and ideally only includes file information from the Prefect package.

```python
def call_state_handlers(method: Callable[..., State]) -> Callable[..., State]:
    @functools.wraps(method)
    def inner(self: "Runner", state: State, *args: Any, **kwargs: Any) -> State:
        ...
        except Exception as exc:
            formatted = "Unexpected error: {}".format(repr(exc))
            self.logger.exception(formatted)
            if raise_on_exception:
                raise exc
            new_state = Failed(formatted, result=exc)
            ...
```

## Alternatives
- Additional decorator around @task that catches, logs locally and replaces the exception.  It is a bit clunky and I'm not sure it will work in all cases.   I need to dig in a bit on @curry to make sure I have this decorator working correctly.

nathan5280:
I made a stab at implementing this.   https://github.com/PrefectHQ/prefect/pull/2048

Issue #2038, opened 2020-02-16
Local Log Storage


nathan5280:
## Use Case
Businesses that deal with customer, medical or otherwise sensitive data are concerned about sending that data outside of their controlled environment.   This includes log messages being sent to Prefect Cloud.

## Solution
Similar to pluggable ResultHandlers it would be helpful to have pluggable LogHandlers that worked in conjunction with the CloudHandler to store the private parts of the log message in the tenants preferred local storage and replaced the log message sent to Prefect Cloud with a link to the local data.

## Alternatives

- on_start(): I've used on_start() with the LocalExecutor to modify the logging configuration to add a local handler that saves the log message to a local file and adds the URL to the message sent to Prefect Cloud.  This has a number of problems:
    - on_start() behaves differently with different executors.
    - on_start() doesn't have the flow/run context to initialize the datastore (create directory).  As a result every log message needs to check for the existence of the log directory and create it if it doesn't exist.  This would be the same for databases and other locations that need to be initialized.  If a new hook is added it would be nice to have it somewhere after flow.run() is called so the full context is available.
    - DaskExecutor reconfigures logging for each task driving the need to reconfigure logging for each task.   (Not positive this is the case as I had many changes in flight at once.)

*Latest Thinking 2020/02/16 AM*
- Store local only data in extra argument to logger.info().  
    ```python
    logger = prefect.context.get("logger")
    logger.info("safe information message.", extra={"local_only", "message that needs to stay local."})
    ````
- Through configuration or new flow_start() hook:
    - Replace formatter on StreamHandler on prefect logger.   This formatter which writes to stdout would check to see if the local_only attribute exists on log.record.   If it exists,  append the "local_only" content to the message formatted using the default formatter. 
    - Modify the CloudHandler to call a formatter.   This formatter would check to see if the local_only attribute exists on the log.record.   If it exists, append the URL to the message or to a special key to be used by prefect cloud when it displays the log message.  *The local_only attribute may need to be deleted from the log.record.   I haven't looked closely enough to see how the log.record is being passed to Prefect Cloud.*
    - Add a LocalLogHandler to the prefect logger.   This handler would check to see if the local_only attribute exists on log.record.  If it exists, format the message with the same formatter used by the StreamHandler above.  Log the message to the appropriate local storage (file system, GCS, S3, Postgres)
- Additional Considerations:
    - Enable LocalLogHandler with queuing and buffering capabilities that CloudHandler has.
[...]

PR #2160, opened 2020-03-15
[WIP][PoC] Standalone Redactor

nathan5280:
Solid progress on #2038 #2039 (Local storage of log messages and exceptions.)

Per our discussions last week and getting more comfortable poking around in Prefect-Core I got a working prototype for redacting messages and exceptions locally and only sending a URI reference to the cloud.  Here is what the log looks link in Cloud.   The first reference is a redacted log message and the second is the redacted exception.

redacted-cloud

The tasks and local logfiles are:

**Log Social Security Number**
```python
@task
def redacted_log():
    logger.info("SSN: 123-45-6789")
```

```text
[2020-03-15 16:44:06,512] INFO - prefect.Task: redacted_log | SSN: 123-45-6789
```

**Raise Exception**
```python
@task
def redacted_exception():
    1 / 0
```

```text
[2020-03-15 16:44:07,617] ERROR - prefect.CloudTaskRunner | Unexpected error: ZeroDivisionError('division by zero')
  File "/home/xxx/projects/prefect/src/prefect/engine/runner.py", line 48, in inner
  File "/home/xxx/projects/prefect/src/prefect/engine/task_runner.py", line 884, in get_task_run_state
  File "/home/xxx/projects/prefect/src/prefect/utilities/executors.py", line 184, in timeout_handler
    return fn(*args, **kwargs)
  File "/home/xxx/projects/prefect/redactor-poc/example/redactor.py", line 13, in redacted_exception
    1 / 0
ZeroDivisionError: division by zero
```

I can see scenarios where people might want to redact the task and flow name or use regex to mask certain character strings.   It also seems like we can get from the above log to some simple tool/process to make the links clickable to review the locally stored logs.  Here locally means inside of the uses controlled environment.

## Implementation
In the end I was able to keep all the changes in [logging.py](https://github.com/nathan5280/prefect/blob/standalone-redact/src/prefect/utilities/logging.py) and the additions in (local_file_redactor.py)[https://github.com/nathan5280/prefect/blob/standalone-redact/src/prefect/utilities/local_file_redactor.py].

### Configuration
```yaml
[logging]
# The logging level: NOTSET, DEBUG, INFO, WARNING, ERROR, or CRITICAL
level = "INFO"

    [logging.redactor]
    redactor_class = "prefect.utilities.local_file_redactor.LocalFileRedactor"
    redactor_format = "%(asctime)s - %(name)s - %(levelname)s - Redacted"
    redactor_root_dir = "/tmp"
```
The RedactorFilter, RedactorHandler classes and redactor_mapper() function are all wrapped into the LocalFileRedactor class to make it easier to load them dynamically.   Nothing special about this other than it keeps everything together with the helper functions to build the paths.

### Strategy
1. All the loggers and handlers roll up to the "prefect" logger so that is where things get attached.  Logging configuration is modified in logging._add_redactor() which is called at the end of logging._create_logger.  I haven't confirmed if this works for anything other than the local agent.

1. Add a RedactorHandler to the prefect logger.
    1.  Select log messages from 'prefect Task: <task-name>' logger and write the message to the local file system.
    1. Select log messages from 'prefect CloudTaskRunner' logger that have exc_info associated with them and write the message to the local file.
1. Replace the formatter on the CloudHandler with the RedactorFormatter.  This formatter replaces messages for records that meet the same criteria as the Handler above with the local file URI.
1. Replace the default logic in CloudHandler that maps the log record to the log dictionary that is sent to the Cloud.


src/prefect/utilities/logging.py
```diff
@@ -31,11 +33,15 @@
...
     "task_slug",
     "task_run_id",
+    # Add start time to help the LocalFileRedactor build logical sorted directories of messages.
+    "scheduled_start_time",
 )
 
 
 class CloudHandler(logging.StreamHandler):
     def __init__(self) -> None:
+        with Path("/tmp/logging-trace.txt").open("at") as fh:
```
lauralorenz: Considering deleting this

src/prefect/utilities/logging.py
```diff
@@ -194,6 +251,8 @@ def _create_logger(name: str) -> logging.Logger:
     Returns:
         - logging.Logger: a configured logging object
     """
+    with Path("/tmp/logging-trace.txt").open("at") as fh:
```
lauralorenz: Consider removing this

src/prefect/utilities/logging.py
```diff
@@ -207,6 +266,9 @@ def _create_logger(name: str) -> logging.Logger:
     cloud_handler = CloudHandler()
     cloud_handler.setLevel("DEBUG")
     logger.addHandler(cloud_handler)
+
+    _add_redactor()
```
lauralorenz: We are considering whether the check for whether this will run anything should be inline here (easier to reason that it will be skipped without jumping to the function for new readers) or if it is cleaner the way that it is. I will double check for precedent elsewhere in the codebase, but fundamentally I think this is a matter of style.

Issue #2141, opened 2020-03-10
PIN 16: Add `Result` subclasses corresponding to all current result handlers


lauralorenz:
## Use Case
Blocked by #2139. See PIN 16 (PR #2134) for background.

## Solution
Implement a `Result` subclass for each of the current result handlers that knows how to `.read`, `write`, and check if it `.exists`:

- [x] AzureResultHandler
- [x] GCSResultHandler
- [x] LocalResultHandler
- [x] S3ResultHandler
- [x] SecretResultHandler
- [x] JSONResultHandler

Please do not implement ConstantResultHandler as part of this ticket as it will be handled separately in #2145.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

wagoodman:
Suggested refactor of the directory layout:
```
src/prefect/engine/
├── (result.py) # <--- delete
├── result/
│   ├── __init__.py # <-- import and expose base for backwards compatibility 
│   ├── base.py # <-- today's result.py
│   ├── constant_result.py
│   ├── gcs_result.py
│   ├── json_result.py
│   ├── local_result.py
│   ├── s3_result.py
│   └── secret_result.py
├── result_handlers/
└── ...
```
 There seem to be a couple of different patterns in the repo for this case, happy to hear thoughts. @lauralorenz I could start working on some of this assuming that #2139 is mostly "complete enough"?

lauralorenz:
@wagoodman I see about the directory structure: mimics `engine/executors`, `environments/storage` and `engine/result_handlers` (today), differs from `environments/execution` or `agent`, the latter group having provider subpackages. I think the division might be historically arbitrary but one side effect of the provider subpackages is you have more room to store auxillary files (like `agent/kubernetes/rbac.yaml`). I don't short/medium-term anticipate needing to do that for Results and we can always change it later, so I think what you have here is best organized.

Definitely pick it up, and if you can stand to derive from a non-master branch for now it does seem to me #2139 is getting pretty stable ;) (*shudders in git rebase -i*)

lauralorenz:
Looks like #2139 is officially closed so we are in the clear, and I assigned this to you @wagoodman for realsies if you are still in

wagoodman:
There's two paths to go here depending on the timeline for `ResultHandlers`:

[...]

PR #2161, opened 2020-03-16
Add GCSResult

wagoodman:

## What does this PR change?
- reorg result.py to a directory of result implementations
- Adds `GCSResult` for writing/reading results to/from GCS bucket.

Addresses the GCS component for #2141

Since this is a longer-term PIN, we should update the documentation closer to release in a separate PR.


src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,123 @@
...
+    service account key.
+
+    Args:
+        - bucket (str): the name of the bucket to write to / read from
+        - credentials_secret (str, optional): the name of the Prefect Secret
+            which stores a JSON representation of your Google Cloud credentials.
+    """
+
+    def __init__(
+        self, bucket: str = None, credentials_secret: str = None, **kwargs: Any
```
wagoodman: we should have value be a positional arg, not kwargs

src/prefect/engine/result/base.py
```diff
@@ -123,6 +124,11 @@ def store_safe_value(self) -> None:
                 value=value, result_handler=self.result_handler
             )
 
+    def render_destination(self) -> str:
```
wagoodman: change to `render_filename` (can't really be generic)

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,123 @@
...
+        Writes the result value to a location in GCS and returns the resulting URI.
+
+        Returns:
+            - str: the GCS URI
+        """
+
+        uri = self.render_destination()
+
+        self.logger.debug("Starting to upload result to {}...".format(uri))
+        binary_data = base64.b64encode(cloudpickle.dumps(self.value)).decode()
```
wagoodman: pull this into the result base class

src/prefect/engine/result/base.py
```diff
@@ -123,6 +126,18 @@ def store_safe_value(self) -> None:
...
+        return base64.b64encode(cloudpickle.dumps(self.value)).decode()
+
+    @classmethod
+    def deserialize(cls, serialized_value: str) -> Any:
+        return cloudpickle.loads(base64.b64decode(serialized_value))
+
+    def render_filepath(self) -> str:
+        if not self.filepath_template:
+            raise ValueError("No filepath_template provided")
+        return self.filepath_template.format(**prefect.context)
```
lauralorenz: GREETINGS talked with @cicdw more on what was being discussed on slack at https://prefecthq.slack.com/archives/CH84XHRB4/p1584460170003700, we want to make this more like

```
def format(self, **kwargs: Any) -> Result:
     # something like...
     self._templated_filename = self.filename_template.format(**kwargs)
     return self
```

At this stage, this is basically so that Result object itself isn't so implicitly dependent on prefect.context, which could vary depending when the rendering method is called; it's more explicit that we get to decide in the pipeline logic when we send it context which is also when we know what is in context better.

For example, later on the pipeline is going to do something like this, so its really clear its just moving the **prefect.context expansion to where the pipeline has control:

```
hydrated_result = self.task.result.format(**prefect.context)
hydrated_result.write(val)
state = Success(result=hydrated_result)
return state
```

In addition, to Chris' concerns on the slack thread it makes it easier for us to send the rendered filename to Cloud since it's in an attribute of Result. 
On the one hand this is 'nice to have' so that Cloud UI can render the filename and users can easily inspect that from the UI.
On another hand Chris was explaining that Cloud will need to know this because of the way it sets up flow recovery. In particular the edge case he is thinking of that resonated the most with me is:
  - Cloud user sets up a task with `Result.filename_template="results/{task_name}.txt"`
      - In the suggested situation, when the task writes its result, it also sets its Result.templated_filename = "/Users/laura/.prefect/results/my_super_task.json"
   -  task runs successfully and submits its state to Cloud, including that templated_filename
   -  drama: right when the task finished running my agent and/or infra dies, but there are still downstream dependencies in this flow run
    -  when I wake up my agent/infra, because I am a Cloud user my flow is going to try and start itself again from Cloud's database which has all the task states
        - Cloud grabs the task states from its database to initialize a FlowRunner to recover the flow run we were on, including deserializing upstream results to pass to the next unrun task
        - if upstream results have Result.templated_filename set already, it can call Result.read to grab that result from that location and give it to the next task
       - if the upstream results only have Result.render_filepath and try to do so with the current `prefect.context`, they are actually in FlowRunner context instead of TaskRunner context and might not be able to figure it out if the template requires TaskRunner context like {task_name}.

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,137 @@
...
+    runtime environment has the proper credentials available
+    (see https://cloud.google.com/docs/authentication/production for all the authentication options).
+
+    You can also optionally provide the name of a Prefect Secret containing your
+    service account key.
+
+    Args:
+        - value (Any, optional): the value of the result
+        - bucket (str): the name of the bucket to write to / read from
+        - credentials_secret (str, optional): the name of the Prefect Secret
```
joshmeek: Have you tested this with both authentication methods? Would be nice to have an example Prefect Secret in the class docstring

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @gcs_bucket.setter
+    def gcs_bucket(self, val: Any) -> None:
+        self._gcs_bucket = val
+
+    def __getstate__(self) -> dict:
+        state = self.__dict__.copy()
+        if "_gcs_bucket" in state:
+            del state["_gcs_bucket"]
+        return state
```
joshmeek: Why remove the `_gcs_bucket`? jw for my own knowledge

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,137 @@
...
+    def write(self) -> str:
+        """
+        Writes the result value to a location in GCS and returns the resulting URI.
+
+        Returns:
+            - str: the GCS URI
+        """
+
+        if not self._rendered_filepath:
+            raise ValueError("must call Result.render_filepath() first")
```
joshmeek:
```suggestion
            raise ValueError("Must call `Result.render_filepath()` first")
```

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,137 @@
...
+        Args:
+            - loc (str, optional): the GCS URI
+
+        Returns:
+            - Any: the read result
+        """
+        uri = loc or self._rendered_filepath
+
+        if not uri:
+            raise ValueError("must call Result.render_filepath() first")
```
joshmeek:
```suggestion
            raise ValueError("Must call `Result.render_filepath()` first")
```

src/prefect/engine/result/gcs_result.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Checks whether the target result exists.
+
+        Does not validate whether the result is `valid`, only that it is present.
+
+        Returns:
+            - bool: whether or not the target result exists.
+        """
+        if not self._rendered_filepath:
+            raise ValueError("must call Result.render_filepath() first")
```
joshmeek:
```suggestion
            raise ValueError("Must call `Result.render_filepath()` first")
```

src/prefect/engine/result/gcs_result.py
wagoodman: The `gcs_client.bucket` is not serializable, this removes the offending object from state that is used for pickling.

src/prefect/engine/result/gcs_result.py
wagoodman: indeed, with the default credentials from an already auth'd box and with a Prefect secret containing a service account JSON key --both worked.  I'm not certain by what you mean for an example secret, as it would only be a name of a secret that is already stored in a Prefect Cloud Secret.

joshmeek: By example I mean the format of the Secret, either written out like:
```
{
    'secret_thing1': '...',
    'secret_thing2': '...',
    etc.
}
```
or a link to the format of the credentials secret from the GCP docs

src/prefect/engine/result/base.py
```diff
@@ -123,6 +126,44 @@ def store_safe_value(self) -> None:
...
+        Args:
+            - **kwargs (Any): string format arguments for result.filepath_template
+
+        Returns:
+            - Any: the current result instance
+        """
+        if not self.filepath_template:
+            raise ValueError("No filepath_template provided")
+
+        self._rendered_filepath = self.filepath_template.format(**kwargs)
```
cicdw: We shouldn't mutate `self` here, but instead create a copy of `self` that we return.  Here's an example for the reasoning:

```python
@task(result=Result("my-results/{scheduled_start_time}.csv")
def my_task():
    # do stuff

...
flow.run() # runs on schedule
```

once the first run occurs, the filepath will be set in stone instead of dynamic.

src/prefect/engine/result/base.py
```diff
@@ -123,6 +126,44 @@ def store_safe_value(self) -> None:
...
+
+        Args:
+            - serialized_value (str): The serialized result value
+
+        Returns:
+            - Any: the deserialized result value
+        """
+        return cloudpickle.loads(base64.b64decode(serialized_value))
+
+    def render_filepath(self, **kwargs: Any) -> "Result":
```
cicdw: While this is certainly accurate, I think we benefit from having human-friendly method names in most places; any chance we could change this to `format` to mirror the python string method?

wagoodman: Calling `format` on a string makes semantic sense, calling `format` on a result does not make semantic sense --this is the reason for the extra verbiage. However, I can rename this to `format`.

src/prefect/engine/result/base.py
wagoodman: I can change this to a copy of the result, however, without seeing how this object is used within the TaskRunner pipeline, it is difficult to verify this either way. 

That is, if on each run we are invoking `format`, this should not be a problem. If we are checking for a rendered filepath attribute before calling `format`, then we would indeed incorrectly using only a single value for the filepath for each run.

PR #2171, opened 2020-03-19
Closes #2144. Implemented validate() method on Result base class

### LINKED ISSUE #2144: PIN 16: Implement `Result.validate`

lauralorenz:

## What does this PR change?
Closes #2144. Adds a validate() method onto the Result baseclass that will call all validators for a Result. 

Note that this method isn't hooked up to be called by anything yet, but should be called by the TaskRunner eventually.

## Why is this PR important?
Overall, this is the building block to support arbitrary functions of the signature `validation_fn(r: Result) -> bool` to be produced by users and passed to Results to be run automatically for them. They will not actually be called until we hook this into the TaskRunner logic.


src/prefect/engine/result.py
```diff
@@ -123,6 +123,23 @@ def store_safe_value(self) -> None:
...
             )
 
+    def validate(self) -> bool:
+        """
+        Run any validator functions associated with this result and return whether the result is valid or not.
+
+        Returns:
+            - bool: whether or not the Result passed all validation functions
+        """
+        if self.run_validators and self.validators:
```
lauralorenz: I think we expect to check self.run_validators in the pipeline, but I threw it in here too. 

Calling out that implemented here is an assumption to this question: do we expect a Result with `self.run_validators=False`, if pressed, to consider itself "valid" -- as in, if someone calls Result.validate() on it anyways, should it return True?

src/prefect/engine/result.py
```diff
@@ -123,6 +123,23 @@ def store_safe_value(self) -> None:
...
+            for validation_fn in self.validators:
+                is_valid = validation_fn(self)
+                if not is_valid:
+                    # once a validator is found to be false, this result is invalid
```
lauralorenz: Again with the assumptions, but what I've got here thinks ANY invalid validators means that the result is not valid.

src/prefect/engine/result.py
```diff
@@ -123,6 +123,23 @@ def store_safe_value(self) -> None:
...
+    def validate(self) -> bool:
+        """
+        Run any validator functions associated with this result and return whether the result is valid or not.
+
+        Returns:
+            - bool: whether or not the Result passed all validation functions
+        """
+        if self.run_validators and self.validators:
+            for validation_fn in self.validators:
+                is_valid = validation_fn(self)
```
lauralorenz: Considered taking *args and **kwargs all up in here, in case we pass in context for them (or users want to call validate themselves, passing in more random stuff to their functions) but I think we expect the validation functions to be parameterizable only by the actual data they need to validate; thoughts?

In fact I'm passing the whole `Result`, though really I think we expect people's functions to be mostly dependent on Result.value. I'm considering unpacking Result.value on their behalf for convenience, like `validation_fn(self.value, self)`; thoughts?

tests/engine/test_result.py
```diff
@@ -290,3 +292,54 @@ def test_results_are_pickleable_with_their_safe_values():
...
+
+    def test_result_validate_returns_true_on_none_invalid(self):
+        no_false_validators_fns = [lambda r: True, lambda r: True]
+
+        r = Result(value=None, validators=no_false_validators_fns)
+        is_valid = r.validate()
+
+        assert is_valid is True
+
+    @pytest.mark.parametrize(
```
lauralorenz: I actually never seriously used pytest before, but I'm really into it. LMK if I'm overzealous here or using it weirdly.

src/prefect/engine/result.py
joshmeek: I would consider the case in which `run_validators` is `False` to always consider itself valid. Maybe it could raise a warning if someone calls `Result.validate()` with `self.run_validators=False` to say something along the lines of it not actually validating anything.

src/prefect/engine/result.py
joshmeek: I'm on board with this, definitely seems like an all or nothing kind of validation step to me (i.e. you opt-in to providing X number of validators therefore you are saying they all have to be valid)

src/prefect/engine/result.py
joshmeek: Yeah I believe they will only need to be parameterizable by the data needed for validation.

They will _most likely_ be dependent on the `Result.value` but I do like having the whole result object in case needed. Throwing in my 2𝇍 but I don't feel strongly for it

tests/engine/test_result.py
joshmeek: No this looks great!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 ### Enhancements
 
 - Add ValidationFailed state and signal in anticipation of validating task outputs [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
+- Add Result.validate method [#2144](https://github.com/PrefectHQ/prefect/issues/2144)
```
joshmeek: Could we expand this entry a couple words to say either what the validate method is or what it does? Could be as simple as `Add Result.validate method to check validator functions` or something like that

src/prefect/engine/result.py
wagoodman: `run_validators` is really a hint for the validators to be run or not run from the viewpoint of the TaskRunner pipeline (as you mentioned @lauralorenz ). I would expect `validate` to alway run any user provided validators, and if no validators provided, then it is valid.

src/prefect/engine/result.py
wagoodman: I agree @joshmeek , if they are providing their own functions at Result creation, they can throw most variable data onto that result as they need it (also makes it reproducible). It isn't clear where a user would provide extra variable info during the TaskRunner pipeline.

@lauralorenz  I like the `validation_fn(self)` approach.

src/prefect/engine/result.py
wagoodman: Agreed, any invalid validator result makes the most sense. Besides, if the user wants another kind of boolean operator here (instead of AND) then they can do the same thing by combining multiple validators into a single validator that does the operator they want (like OR).

src/prefect/engine/result.py
lauralorenz: Roger that team, leaving it as is then

CHANGELOG.md
lauralorenz: Fixed in 0bcfab0

src/prefect/engine/result.py
lauralorenz: Ok so I went with a combo,
- validate() will run your validators if you call that method no matter what, you said you wanted it validated and you're getting it - even if run_validators is False. We expect to not call .validate if run_validators is false ourselves in the pipeline. Users who call .validate get what they asked for. This is the @wagoodman piece
- BUT if user dared do so, they get a warning log (softer version of what @joshmeek was suggesting).

I updated related code and tests in 919db95

CHANGELOG.md
lauralorenz: Fixed in 0bcfab0

src/prefect/engine/result/base.py
```diff
@@ -110,6 +111,7 @@ def __init__(
         self.cache_validator = cache_validator
         self.filepath_template = filepath_template
         self._rendered_filepath = None  # type: Optional[str]
+        self.logger = logging.get_logger(type(self).__name__)
```
lauralorenz: Heeyyyy sneaking this in, I copied the style from result handlers.

src/prefect/engine/result/base.py
```diff
@@ -127,6 +129,33 @@ def store_safe_value(self) -> None:
...
+        Emits a warning log if run_validators isn't true, and proceeds to run validation functions anyway.
+
+
+        Returns:
+            - bool: whether or not the Result passed all validation functions
+        """
+        if not self.run_validators:
+            self.logger.warning(
+                "A Result's validate method has been called, but its run_validators attribute is False. "
+                "Prefect will not honor the validators without run_validators=True, so please change it "
```
joshmeek: Isn't this technically false because the validators are still being run? Or does this mean that in the standard pipeline it won't be respected?

lauralorenz: The latter; this is assuming we allow a possible valid case of where a user really wants to call Result(run_validators=False).validate() themselves (like in a task), but where any call to Result.validate in the pipeline by Prefect itself must check if run_validators=True first

PR #2175, opened 2020-03-19
Add note about using Prefect Docker images to installation doc

### LINKED ISSUE #2085: Document Our Docker Images
We should add a page to the docs describing the Prefect docker images we publish on dockerhub and a description of some of the most commonly used ones.

For example: just looking at https://hub.docker.com/r/prefecthq/prefect/tags I wasn't sure of the difference between the `latest` image and the `latest-python3.7` images (there is no difference, they're the same image).

joshmeek:
Adds a note to the Installation doc about Prefect Docker images. The breakdown of Docker image tags is present on the DockerHub README https://hub.docker.com/r/prefecthq/prefect

Closes #2085

### DIFF
--- a/docs/core/getting_started/installation.md
+++ b/docs/core/getting_started/installation.md
@@ -58,3 +58,22 @@ For developing Prefect, see the [development guide](../development/overview.md).
 ## Docker
 
 Prefect provides Docker images for master builds and versioned releases [here](https://hub.docker.com/r/prefecthq/prefect).
+
+To run the latest Prefect Docker image:
+
+```bash
+docker run -it prefecthq/prefect:latest
+```
+
+Image tag breakdown:
+
+| Tag              | Prefect Version          | Python Version |
+|------------------|:------------------------:|---------------:|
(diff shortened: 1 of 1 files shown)


docs/core/getting_started/installation.md
```diff
@@ -54,3 +54,7 @@ The extra packages include:
 ## Development
 
 For developing Prefect, see the [development guide](../development/overview.md).
+
+## Docker
+
+Prefect provides Docker images for master builds and versioned releases [here](https://hub.docker.com/r/prefecthq/prefect).
```
cicdw: I think we should describe our tagging policy in at least some detail (e.g., what types of tags, what do they mean, etc.)

joshmeek: They're on the readme of the Dockerhub page. Do you think it should also go here?

cicdw: Oh nice -- but yea I do actually think we should copy / paste that exact table here.  I didn't even know that table was there because it's at the bottom of the standard README

PR #2177, opened 2020-03-19
Add example Dask Executor use to the executor concept doc

### LINKED ISSUE #2086: Docs: More Detail to the Executors Page.
Hello!

I've found an issue here:`The DaskExecutor is a completely asynchronous engine that can run functions in a distributed Dask cluster. This is the recommended engine for production.`

Page reference: https://docs.prefect.io/core/concepts/engine.html#executors

What's the difference between the `LocalDaskExecutor` and the `DaskExecutor`? Which should I use if I want to run a local Dask cluster on the machine I deploy with Docker/Kubernetes? We could expand this section a bit more and add an example `RemoteEnvironment` with a an executor.

joshmeek:
Adds an example to the Executor section of the Engine concept doc. This is a brief addition to show the use of an executor early for users.

Closes #2086


docs/core/concepts/engine.md
```diff
@@ -34,8 +34,34 @@ The executor classes are responsible for actually running tasks. For example, th
...
+
+executor = DaskExecutor(address="tcp://localhost:8786")
+flow.run(executor=executor)
+```
+
+This `DaskExecutor` will connect to a Dask scheduler over the address `tcp://localhost:8786` and begin submitting work to be executed on Dask workers.
+
+:::tip Dynamic Scheduler
+If no scheduler `address` is specified for the `DaskExecutor` than an in-process scheduler will be created and torn down upon completion. See the [DaskExecutor API Documentation](/api/latest/engine/executors.html#daskexecutor) for more information.
+:::
```
cicdw: Can we add a `::: warning` box documenting the difference between the `LocalDaskExecutor` and the `DaskExecutor`?  Let me know if you need some guidance on what the main differences are 👍

cicdw:
Actually, I think we should highlight the distinction in the top-level `engine/executors.py` API document as well.

PR #2180, opened 2020-03-20
Add maximum polling option to agents

### LINKED ISSUE #2037: Introduce max-polls flag on all Agents
## Current behavior
Currently calling `prefect agent start ...` will create a long-running process that will only stop after a kill signal is sent to the process. This is great for 99% of use cases, especially in production environments.  

However, there are two situations in which I'd like to specify a max number of polls that an agent will make:

- **local development / short-lived agents**: sometimes when quickly iterating users spin up a local agent with many labels just to test a single run of a Flow.  Doing something like `prefect agent start --max-polls 5` would allow users to kick off a single run and know the agent won't stay live
- **mixed schedulers / serverless setups**: there are some situations in which we don't want to have a 24/7 process for scheduling work and instead rely on an internal-infra scheduler to check whether there's anything to do.  For example, users might want to run a flow once a day within a Heroku instance, and use the Heroku scheduler to schedule a short-lived Prefect agent to run right around the time a Flow run is scheduled in Prefect Cloud.  The Heroku scheduler would spin up an agent with e.g., `--max-polls 20` and their work could still be tracked via Cloud.  

For the latter, I fully recognize that this introduces "gotchas" but until we offer a fully serverless agent it at least exposes the pattern to users.

## Proposed behavior
`prefect agent start $AGENT_TYPE --max-polls N` would only poll Prefect Cloud `N` times before gracefully exiting.  For some agents, immediate exit is possible (e.g., k8s) and for others we would need to wait for submitted work to reach completion (e.g., the `LocalAgent`).

Open for discussion to see what others think.

joshmeek:

## What does this PR change?
Closes #2037 
Adds a `--max-polls` flag to the `prefect agent start` CLI command and a `max_polls` kwarg to all agents. This option allows users to specify the amount of times they want the agent to poll Cloud looking for flow runs. This is great for one-off and testing scenarios because the agents no longer exist as infinitely running loops.

## Why is this PR important?
Opens the door for short-lived agents that is beneficial for areas such as testing and serverless platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Enhancements
 
-- Add ValidationFailed state and signal in anticipation of validating task outputs [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
+- Add ValidationFailed state and signal in anticipation of validating task outputs - [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
+- Add max polling option to all agents - [#2037](https://github.com/PrefectHQ/prefect/issues/2037)
 
 ### Task Library
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Add ValidationFailed state and signal in anticipation of validating task outputs - [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
 - Add max polling option to all agents - [#2037](https://github.com/PrefectHQ/prefect/issues/2037)
-- Add ValidationFailed state and signal in anticipation of validating task outputs [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Enhancements
 
+- Add ValidationFailed state and signal in anticipation of validating task outputs - [#2143](https://github.com/PrefectHQ/prefect/issues/2143)
```
cicdw: This appears to be a duplicated changelog entry

src/prefect/agent/agent.py
```diff
@@ -69,16 +70,23 @@ class Agent:
...
+            defaults to infinite
     """
 
     def __init__(
-        self, name: str = None, labels: Iterable[str] = None, env_vars: dict = None
+        self,
+        name: str = None,
+        labels: Iterable[str] = None,
+        env_vars: dict = None,
+        max_polls: int = None,
```
cicdw: Not a big deal either way, but because `math.inf` is a float you could safely use it as the default option here instead of `None`

joshmeek: The reason why I opted to use None was for the `int` typing

PR #2196, opened 2020-03-25
Handle localhost in API endpoint by using docker internal host address

joshmeek:
PR opened against the config update branch. Allows containers launched by the docker agent to reach out to the host machine.


src/prefect/agent/docker/agent.py
```diff
@@ -381,8 +381,13 @@ def populate_env_vars(self, flow_run: GraphQLResult) -> dict:
         Returns:
             - dict: a dictionary representing the populated environment variables
         """
+        if "localhost" in config.cloud.api:
```
cicdw: Would it be more robust to instead check
```python
if config.backend == "server":
```
?

joshmeek: This is in case the IP set for the server backend is not localhost

cicdw: Yea, at some point a power user might bind it to a different IP / name

cicdw: It's also more explicit

PR #2198, opened 2020-03-26
Finish converting to snake case API calls

zdhughes:

## What does this PR change?

This PR updates docs and remaining API calls in the codebase to use snake_casing. 

## Why is this PR important?

Consistency is key.

### DIFF
--- a/docs/api/0.7.3/client/secrets.md
+++ b/docs/api/0.7.3/client/secrets.md
@@ -2,8 +2,11 @@
 sidebarDepth: 2
 editLink: false
 ---
+
 # Secrets
+
 ---
+
 ::: warning Secret Tasks are preferred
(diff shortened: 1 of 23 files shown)


docs/api/0.7.3/client/secrets.md
```diff
@@ -18,7 +21,7 @@ To set a Secret in Prefect Cloud, you can use `prefect.Client.set_secret`, or se
 
 ```graphql
 mutation {
-  setSecret(input: { name: "KEY", value: "VALUE" }) {
+  set_secret(input: { name: "KEY", value: "VALUE" }) {
```
joshmeek: I don't feel _too strongly_ about it but I don't think we should update older versioned API documentation

zdhughes: Ah, good catch-- got caught up in finding + replacing, didn't notice this was an older file. I'll go ahead and revert.

zdhughes: There we go-- should be good to go now.

PR #2216, opened 2020-03-30
Add docker compose to manifest

cicdw:


cicdw:
Don't approve yet, going to add a more informative error message to the subprocess call in server start.

PR #2218, opened 2020-03-30
Open source backend, API and UI

cicdw:
This is the big one


pombredanne:
Is this really an open source license?

cicdw:
We are using the term “open source” in the way it is most commonly used: that the code is freely available to download, modify and propose changes to.  Strictly speaking, Prefect’s Community License is “source available” but not OSI approved “open source” because you cannot sell a SaaS offering of Prefect’s server code.

pombredanne:
@cicdw you wrote:
> We are using the term “open source” in the way it is most commonly used: that the code is freely available to download, modify and propose changes to.

I do not think this is the way `open source` is most commonly used for software. Rather that is  software licensed using an opensource.org-approved license as you pointed out, or a license that complies with the open source definition https://opensource.org/osd

> Strictly speaking, Prefect’s Community License is “source available” 

So whether strictly or loosely speaking, this license cannot be qualified as open source. It would be nice to avoid the possible confusion and not call it open source. 

Note I have no issue with the "Prefect’s Community License", my concern is the confusion I was faced with when reading comments and posts about the license and how the license terms did not match the common acceptance of "open source".

PR #2228, opened 2020-03-30
Update `prefect server start` command to specify and disable port mappings

### LINKED ISSUE #2225: Allow setting port for postgres instance in `prefect server start`
## Use Case

It's likely that there's already a postgres instance running on the default port.
`prefect server start` should include an option for setting the port of its Postgres instance.

## Solution

`prefect server start --postgres-port=5433`

joshmeek:

## What does this PR change?
Open PR up for discussion if this is how we want optional port allocations to work when running `prefect server start`. This allows ports from the server services to be binded to different ports on the host machine. For this to work I had to add `__HOST_PORT` options to some of the services in `config.toml`. I'm guessing this is due to some hardcoded ports that these services use inside their containers. (e.g. overwriting `config.server.hasura.port` broke communication inside the containers when they ran)

This implementation allows us to set ports through CLI `--postgres-port` or environment variables `PREFECT__SERVER__DATABASE__HOST_PORT` however we can definitely go a different route.

This PR also adds the options to disable port mappings on a per-service basis. (e.g. `--no-postgres-port`)

Closes #2225 

## Why is this PR important?
Port customization for the start command is a nice convenience to have.

### DIFF
--- a/src/prefect/cli/docker-compose.yml
+++ b/src/prefect/cli/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   postgres:
     image: "postgres:11"
     ports:
-      - "5432:5432"
+      - "5432:${POSTGRES_HOST_PORT}"
     environment:
       POSTGRES_USER: ${POSTGRES_USER}
       POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
@@ -20,7 +20,7 @@ services:
   hasura:
     image: "hasura/graphql-engine:v1.1.0"
     ports:
-      - "3000:3000"
+      - "3000:${HASURA_HOST_PORT}"
     command: "graphql-engine serve"
     environment:
       HASURA_GRAPHQL_DATABASE_URL: ${DB_CONNECTION_URL}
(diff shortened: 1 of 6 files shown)


src/prefect/cli/server.py
```diff
@@ -96,13 +101,63 @@ def server():
...
 
-    env = make_env()
+    # Temporary config set for port allocation
+    with set_temporary_config(
+        {
+            "server.database.host_port": postgres_port,
+            "server.hasura.host_port": hasura_port,
+            "server.graphql.host_port": graphql_port,
+            "server.ui.port": ui_port,
+            "server.port": server_port,
```
cicdw: Any reason not to make them all `host_port`?

src/prefect/cli/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
   postgres:
     image: "postgres:11"
     ports:
-      - "5432:5432"
+      - "5432:${POSTGRES_HOST_PORT}"
```
cicdw: Not a big deal but could we put defaults here as well? e.g.,
```
"5432:${POSTGRES_HOST_PORT:5432}"
```

src/prefect/cli/server.py
joshmeek: I can make them all host port but then it might get confusing for the Apollo API endpoint. Because if you run it exposing the endpoint at a different port then `server.port` then you will also need to update your `server.port` in order for prefect to communicate with it

joshmeek: But I see the point in making it host port for all because that the purpose it serves here

joshmeek:
@cicdw I just tested it by spinning it up using the no-port options which creates a tempdir to house the docker compose file. Then while it was running I manually went to the tempdir and deleted the file. Everything continued to run and when I shut it down it all shut down properly 👍

PR #2229, opened 2020-03-30
add docker publish jobs for server images

TylerWanner:

## What does this PR change?
adds a circle orb to docker publish prefect server images

## Why is this PR important?
versioned images published to docker hub for open use on each merge to master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -445,7 +445,7 @@ workflows:
             tags:
               only: /^[0-9]+\.[0-9]+\.[0-9]+$/
 
-  'Build Earth Images':
+  'Build Server Images':
     jobs:
       - docker/publish:
           docker-password: DOCKER_HUB_PW
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -455,7 +455,7 @@ workflows:
           tag: $CIRCLE_SHA1,latest,master
           filters:
             branches:
-              only: ci-server-images
+              only: master
 
       - docker/publish:
           docker-password: DOCKER_HUB_PW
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -441,3 +444,35 @@ workflows:
               ignore: /.*/
             tags:
               only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+
+  'Build Earth Images':
```
cicdw:
```suggestion
  'Build Server Images':
```
😛

.circleci/config.yml
```diff
@@ -441,3 +444,35 @@ workflows:
...
+    jobs:
+      - docker/publish:
+          docker-password: DOCKER_HUB_PW
+          docker-username: DOCKER_HUB_USER
+          image: 'prefecthq/apollo'
+          path: 'server/services/apollo'
+          tag: $CIRCLE_SHA1,latest,master
+          filters:
+            branches:
+              only: master
```
cicdw: Can we change this to filter based on tag, similar to the `release_to_pypi` job?  I don't think we need to rebuild these image with every PR, only releases

.circleci/config.yml
```diff
@@ -441,3 +444,35 @@ workflows:
               ignore: /.*/
             tags:
               only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+
+  'Build Earth Images':
+    jobs:
+      - docker/publish:
+          docker-password: DOCKER_HUB_PW
+          docker-username: DOCKER_HUB_USER
```
cicdw: Do these need `$` to be read from the environment?

TylerWanner: @cicdw no actually it's the name of the variable that we're passing to the orb

.circleci/config.yml
TylerWanner: @cicdw can we make it master so we can see it run then we can update it to only on releases?

cicdw: Yup for sure, that makes sense

PR #2273, opened 2020-04-06
(WIP) Closes #2168. Newstyle results pipeline logic

### LINKED ISSUE #2168: PIN 16: 0.11.0: Pipeline logic updates between new Result methods and result handlers in Result
## Use Case

### Extra context

## Solution

## Alternatives

lauralorenz:

## What does this PR change?
Adds pipeline logic changes needed to utilize new `Result` subclasses instead of Result Handlers.
This PR is still a work in progress. 

#### Status 4/6
Right now this
- has a basic conversion function during task instantiation to create a Result on the user's behalf if they sent us a result handler. So there is always a `task.result`
- has plugged in using `task.result` to store values in places that we used to create old style Result classes

Still needs to
- use `task.result` to hydrate results in places we are potentially reading from file
- need to revisiting what is special about upstream tasks? mapped tasks? looped tasks? when it comes to peeling a value out of its result (potentially in persistence)
- fix all the tests I have surely broken
- pretty sure my rebase was weird bc I thought we squashed netlify stuff

Overarching questions during implementation are:
1. what if this had just come from cloud?
2. what if we had already called the read method for this task's Result?

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -582,19 +582,13 @@ class TaskRunner(Runner):
         for edge, upstream_state in upstream_states.items():
             # construct task inputs
             if edge.key is not None:
-                handlers[edge.key] = handler = getattr(
-                    edge.upstream_task, "result_handler", None
-                )
-                task_inputs[  # type: ignore
-                    edge.key
-                ] = upstream_state._result.to_result(  # type: ignore
-                    handler
-                )  # type: ignore
+                # todo: 'if came from cloud'; do I have a handle on the upstream state's result instances?
+                task_inputs[edge.key] = upstream_state._result.to_result()
 
         if state.is_pending() and state.cached_inputs:
             task_inputs.update(
                 {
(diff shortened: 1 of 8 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -982,6 +977,7 @@ def check_for_retry(self, state: State, inputs: Dict[str, Result]) -> State:
             run_count = prefect.context.get("task_run_count", 1)
             if prefect.context.get("task_loop_count") is not None:
                 loop_context = {
+                    # todo: use task's converted result object?
                     "_loop_count": Result(
```
lauralorenz, line 981: These need to be freshly instantiated JSONResults, not dependent on the user's Result subclass that they configured

lauralorenz:
Though this has some good notes/todos in it for myself, most of this stab at it has been split out between ongoing discussions on #2280 and isolating the conversion logic into its own PR in #2300, so closing this PR.

PR #2278, opened 2020-04-06
Discussion for #2168. Annotates task runner pipeline where result handlers/results are used

lauralorenz:
This PR is for discussion on all the diff lines where I've annotated spots in the pipeline where

1. we instantiate Results on the user's behalf
2. we call a Result's `store_safe_value` or `to_result` methods
3. where we pass a result handler (sometimes overlaps with the annotations for 1 or 2 but I went more verbose here)

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -84,11 +84,11 @@ class TaskRunner(Runner):
         self,
         task: Task,
         state_handlers: Iterable[Callable] = None,
-        result_handler: "ResultHandler" = None,
+        result_handler: "ResultHandler" = None, # todo: here
     ):
         self.context = prefect.context.to_dict()
         self.task = task
-        self.result_handler = task.result_handler or result_handler
+        self.result_handler = task.result_handler or result_handler # todo: here
         super().__init__(state_handlers=state_handlers)
 
     def __repr__(self) -> str:
@@ -152,12 +152,12 @@ class TaskRunner(Runner):
                 "task_loop_count": state.cached_inputs.pop(  # type: ignore
                     "_loop_count"
                 )  # type: ignore
-                .to_result()
+                .to_result() # todo: here
                 .value,
                 "task_loop_result": state.cached_inputs.pop(  # type: ignore
                     "_loop_result"
                 )  # type: ignore
-                .to_result()
+                .to_result() # todo: here
                 .value,
             }
             context.update(loop_context)
@@ -582,20 +582,20 @@ class TaskRunner(Runner):
             # construct task inputs
             if edge.key is not None:
                 handlers[edge.key] = handler = getattr(
-                    edge.upstream_task, "result_handler", None
+                    edge.upstream_task, "result_handler", None # todo: here
                 )
                 task_inputs[  # type: ignore
                     edge.key
-                ] = upstream_state._result.to_result(  # type: ignore
+                ] = upstream_state._result.to_result(  # type: ignore # todo: here
                     handler
                 )  # type: ignore
 
         if state.is_pending() and state.cached_inputs:
             task_inputs.update(
                 {
-                    k: r.to_result(handlers.get(k))  # type: ignore
+                    k: r.to_result(handlers.get(k))  # type: ignore # todo: here
                     for k, r in state.cached_inputs.items()
-                    if task_inputs.get(k, NoResult) == NoResult
+                    if task_inputs.get(k, NoResult) == NoResult # todo: here
                 }
             )
         return task_inputs
@@ -622,7 +622,7 @@ class TaskRunner(Runner):
             if self.task.cache_validator(
                 state, sanitized_inputs, prefect.context.get("parameters")
             ):
-                state._result = state._result.to_result(self.task.result_handler)
+                state._result = state._result.to_result(self.task.result_handler) # todo: here
                 return state
             else:
                 state = Pending("Cache was invalid; ready to run.")
@@ -637,7 +637,7 @@ class TaskRunner(Runner):
                     candidate, sanitized_inputs, prefect.context.get("parameters")
                 ):
                     candidate._result = candidate._result.to_result(
-                        self.task.result_handler
+                        self.task.result_handler # todo: here
                     )
                     return candidate
 
@@ -721,9 +721,9 @@ class TaskRunner(Runner):
                                         preview=repr(upstream_state.result)[:10],
                                     )
                                 )
-                            upstream_result = Result(
+                            upstream_result = Result( # todo: here
                                 upstream_state.result[i],
-                                result_handler=upstream_state._result.result_handler,  # type: ignore
+                                result_handler=upstream_state._result.result_handler,  # type: ignore # todo: here
                             )
                             states[edge].result = upstream_result
                         elif state.is_mapped():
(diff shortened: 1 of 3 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -84,11 +84,11 @@ def __init__(
         self,
         task: Task,
         state_handlers: Iterable[Callable] = None,
-        result_handler: "ResultHandler" = None,
+        result_handler: "ResultHandler" = None, # todo: here
     ):
         self.context = prefect.context.to_dict()
         self.task = task
-        self.result_handler = task.result_handler or result_handler
+        self.result_handler = task.result_handler or result_handler # todo: here
```
lauralorenz, line 91: Around here is where I was converting self.result_handler into a self.result on behalf of the user IF they sent me a result handler and no result.

src/prefect/engine/task_runner.py
```diff
@@ -901,28 +901,28 @@ def get_task_run_state(
         except signals.LOOP as exc:
             new_state = exc.state
             assert isinstance(new_state, Looped)
-            new_state.result = Result(
-                value=new_state.result, result_handler=self.result_handler
+            new_state.result = Result( # todo: here
```
lauralorenz, line 904: Sounded from the call today that these need to be instantiated as `JSONResult` objects here as an internal implementation detail for looping.

src/prefect/engine/task_runner.py
```diff
@@ -901,28 +901,28 @@ def get_task_run_state(
...
+                value=new_state.result, result_handler=self.result_handler # todo: here
             )
             new_state.cached_inputs = inputs
             new_state.message = exc.state.message or "Task is looping ({})".format(
                 new_state.loop_count
             )
             return new_state
 
-        result = Result(value=result, result_handler=self.result_handler)
+        result = Result(value=result, result_handler=self.result_handler) # todo: here
```
lauralorenz, line 913: This is where I mutated the task's result a la `self.task.result.value=result` instead of creating a new Result object for them.

src/prefect/engine/task_runner.py
```diff
@@ -84,11 +84,11 @@ def __init__(
         self,
         task: Task,
         state_handlers: Iterable[Callable] = None,
-        result_handler: "ResultHandler" = None,
+        result_handler: "ResultHandler" = None, # todo: here
```
cicdw, line 87: I think we should remove this kwarg from runner classes, it doesn't serve much purpose AFAIK

src/prefect/engine/task_runner.py
```diff
@@ -152,12 +152,12 @@ def initialize_run(  # type: ignore
...
                     "_loop_count"
                 )  # type: ignore
-                .to_result()
+                .to_result() # todo: here
                 .value,
                 "task_loop_result": state.cached_inputs.pop(  # type: ignore
                     "_loop_result"
                 )  # type: ignore
-                .to_result()
+                .to_result() # todo: here
```
cicdw, line 160: Context here:

- this data needs to be stored on the `state` object for the task so that it can be persisted and retrieved
- because of this, this data needs a formal `Result` wrapper (otherwise they will be serialized to `None` / `NoResult`).  `JSONResult` is appropriate here because they are simple objects

These `to_result` calls are then basic "write" calls with the new interface, no filename templating needed.  I would expect the `write` call for `JSONResult` to store the output of `json.dumps` as an attribute of the class.  Or conversely, we can "write" the results within the serialization interface and not even need to explicitly call write here.

src/prefect/engine/task_runner.py
```diff
@@ -582,20 +582,20 @@ def get_task_inputs(
             # construct task inputs
             if edge.key is not None:
                 handlers[edge.key] = handler = getattr(
-                    edge.upstream_task, "result_handler", None
+                    edge.upstream_task, "result_handler", None # todo: here
```
cicdw, line 585: These are stored in a `handlers` dictionary so that we can easily access upstream result handlers down below, outside of this iteration

src/prefect/engine/task_runner.py
```diff
@@ -582,20 +582,20 @@ def get_task_inputs(
...
                 )
                 task_inputs[  # type: ignore
                     edge.key
-                ] = upstream_state._result.to_result(  # type: ignore
+                ] = upstream_state._result.to_result(  # type: ignore # todo: here
```
cicdw, line 589: This step does _two_ things:

- reads the underlying result
- ensures that we use the user's result handler when performing the read (this is the "hydration" step)

src/prefect/engine/task_runner.py
```diff
@@ -582,20 +582,20 @@ def get_task_inputs(
...
                     handler
                 )  # type: ignore
 
         if state.is_pending() and state.cached_inputs:
             task_inputs.update(
                 {
-                    k: r.to_result(handlers.get(k))  # type: ignore
+                    k: r.to_result(handlers.get(k))  # type: ignore # todo: here
```
cicdw, line 596: Same as above -- this call is performing to steps, both a read and a hydration.

src/prefect/engine/task_runner.py
cicdw: This data payload actually represents the output of the task, which we don't want to store as `JSON`.  We need to use the users `self.task.result` for this piece of data.

src/prefect/engine/task_runner.py
cicdw: So this sounds like we might want to use:

```python
result = self.task.result.format(**prefect.context)
result = result.write(value)
```

Which I think we should simplify to a single interface:
```python
result = result.write(value, **prefect.context)
```

src/prefect/engine/task_runner.py
```diff
@@ -901,28 +901,28 @@ def get_task_run_state(
...
         ## checkpoint tasks if a result_handler is present, except for when the user has opted out by disabling checkpointing
         if (
             state.is_successful()
             and prefect.context.get("checkpointing") is True
             and self.task.checkpoint is not False
-            and self.result_handler is not None
+            and self.result_handler is not None # todo: here
         ):
-            state._result.store_safe_value()
+            state._result.store_safe_value() # todo: here
```
cicdw, line 925: This is also a write call; now that we are presumably greedily writing results anyway, this might not be necessary 🤔

src/prefect/engine/task_runner.py
```diff
@@ -982,13 +982,13 @@ def check_for_retry(self, state: State, inputs: Dict[str, Result]) -> State:
...
+                    "_loop_count": Result( # todo: here
                         value=prefect.context["task_loop_count"],
-                        result_handler=JSONResultHandler(),
+                        result_handler=JSONResultHandler(), # todo: here
                     ),
-                    "_loop_result": Result(
+                    "_loop_result": Result( # todo: here
                         value=prefect.context.get("task_loop_result"),
-                        result_handler=self.result_handler,
+                        result_handler=self.result_handler, # todo: here
```
cicdw, line 991: the loop result is user data and therefore we want to use the user's result implementation; the `loop_count` is internal and we can use `JSONResult` for it

lauralorenz:
FYI I added some more references so we have a base for discussion as necessary in b67cd5b re: cloud' runner's additions to the task runner pipeline cc @cicdw. I'm effectively dealing with some of this now with the serializers but if you have some holistic comments seeing all of the references I could spot LMK

cicdw:
Closing as this was a reference PR

PR #2285, opened 2020-04-07
Hardening the Result interface

cicdw:

## What does this PR change?
This PR hardens the new PIN-16 `Result` interface to better match our intended way of using it.  Additionally, this PR introduces a new `JSONResult` type for interacting with data which can safely be stored directly in the database.

There are a few ways this PR could be improved:

- [ ] better tests
- [ ] better docstrings describing and explaining what each of the `Result` methods is intended to be used for

I wanted to open the PR for any comments before addressing the two points above.

### DIFF
--- a/src/prefect/engine/results/constant_result.py
+++ b/src/prefect/engine/results/constant_result.py
@@ -16,12 +16,12 @@ class ConstantResult(Result):
         self.value = value
         super().__init__(value=value, **kwargs)
 
-    def read(self, loc: str) -> Result:
+    def read(self, filepath: str) -> Result:
         """
         Returns the underlying value regardless of the argument passed.
 
         Args:
-            - loc (str): an unused argument
+            - filepath (str): an unused argument
         """
         return self
 
@@ -40,12 +40,12 @@ class ConstantResult(Result):
         self.filepath = repr(self.value)
         return self
 
-    def exists(self, loc: str) -> bool:
+    def exists(self, filepath: str) -> bool:
         """
         As all Python objects are valid constants, always returns `True`.
 
         Args:
-             - loc (str): for interface compatibility
+             - filepath (str): for interface compatibility
 
         Returns:
             - bool: True, confirming the constant exists.
--- a/src/prefect/engine/results/constant_result.py
+++ b/src/prefect/engine/results/constant_result.py
@@ -1,5 +1,4 @@
 from typing import Any
-
 from prefect.engine.result import Result
 
 
@@ -30,13 +29,18 @@ class ConstantResult(Result):
         Returns the repr of the underlying value, purely for convenience.
 
         Args:
-            - value (Any): the new value to store on the class
+            - value (Any): the value to store on the class; must be the same as `self.value`
+                and is exposed purely for interface compatibility
             - **kwargs (optional): unused, for compatibility with the interface
 
         Returns:
             - Result: returns self
+
+        Raises:
+            ValueError: if the provided result is distinct from `self.value`
         """
-        self.value = value
+        if self.value != value:
+            raise ValueError("Cannot write new values to `ConstantResult` types.")
         self.filepath = repr(self.value)
         return self
 
--- a/src/prefect/engine/results/constant_result.py
+++ b/src/prefect/engine/results/constant_result.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional
+from typing import Any
 
 from prefect.engine.result import Result
 
@@ -35,6 +35,7 @@ class ConstantResult(Result):
         Returns:
             - Result: returns self
         """
+        self.filepath = repr(self.value)
         return self
 
     def exists(self, loc: str = None) -> bool:
@@ -49,5 +50,4 @@ class ConstantResult(Result):
         Returns:
             - bool: True, confirming the constant exists.
         """
-
         return True
--- a/src/prefect/engine/results/constant_result.py
+++ b/src/prefect/engine/results/constant_result.py
@@ -16,31 +16,36 @@ class ConstantResult(Result):
         self.value = value
         super().__init__(value=value, **kwargs)
 
-    def read(self, arg: Optional[str] = None) -> Any:
+    def read(self, loc: str = None) -> Any:
         """
         Returns the underlying value regardless of the argument passed.
 
         Args:
-            - arg (str): an unused argument
+            - loc (str): an unused argument
         """
         return self.value
 
-    def write(self) -> str:
+    def write(self, **kwargs: Any) -> Result:
         """
         Returns the repr of the underlying value, purely for convenience.
 
-        Returns:
-            - str: the repr of the result
-        """
-        return repr(self.value)
+        Args:
+            - **kwargs (optional): unused, for compatibility with the interface
 
-    def exists(self) -> bool:
+        Returns:
+            - Result: returns self
+        """
+        return self
+
+    def exists(self, loc: str = None) -> bool:
         """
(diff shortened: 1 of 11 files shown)


src/prefect/engine/results/json_result.py
```diff
@@ -0,0 +1,57 @@
...
+        Args:
+            - **kwargs (optional): unused, for compatibility with the interface
+
+        Returns:
+            - Result: returns self
+        """
+        self.filepath = json.dumps(self.value)
+        return self
+
+    def exists(self, loc: str = None) -> bool:
```
lauralorenz, line 45: Is this the same as Constant result on purpose? Unlike Constants which we expect to have a value at instantiation, it seems to me a JSONResult shouldn't consider itself to `exist` until the task has run aka its value has been 'written' to its self.filepath.

Kind of some steps down the way still but are we going to have a problem distinguishing between JSONResults that have a None type value/filepath because their task hasn't been run, and JSONResults that have a None type value/filepath because that is the return value of their task? I think this is what NoResult type is about, and is probably hard to discuss (for me at least) without hooking in the exists pipeline checks more, but I started thinking about it while reading this.

src/prefect/engine/results/json_result.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def read(self, loc: str = None) -> Result:
+        """
+        Returns the underlying value regardless of the argument passed.
+
+        Args:
+            - loc (str): an unused argument
+        """
+        new = self.copy()
+        new.value = json.loads(loc or new.filepath)
```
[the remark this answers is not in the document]
cicdw: > in anticipation of exclusively serializing only Result.filepath to Cloud, so that Cloud can have a copy of the value to use for restart

Yes that's exactly correct.

Re: the naming idea, that's a great idea!  Let's definitely change the name.  Maybe `PrefectResult` to keep it short?

src/prefect/engine/results/json_result.py
cicdw: Hmmm that's a good call out; in practice the value is available for use if you're interacting with the class, so I'm tempted to keep it as-is (and I also wouldn't recommend anyone use this type for their `target`, which is where this method will be relied on)

src/prefect/engine/results/json_result.py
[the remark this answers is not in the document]
lauralorenz: :+1: Yes `PrefectResult` is short and sweet!

src/prefect/engine/results/json_result.py
```diff
@@ -0,0 +1,57 @@
...
+
+        The value stored within a Constant is logically always present,
+        so `True` is returned.
+
+        Args:
+             - loc (optional): for interface compatibility
+        Returns:
+            - bool: True, confirming the constant exists.
+        """
+        return True
```
lauralorenz, line 57: Change this and its docstring to check on self.filepath instead

cicdw:
@joshmeek / @lauralorenz this is ready for re-review; note:
> Merging #2285 into master will increase coverage by 0.01%.

As you review, here is the way I am thinking about the new `Result` abstraction: Results in Prefect will _primarily_ be used as interfaces to various backends (the Prefect DB, GCS, S3, Azure, etc. etc.).  As such, the main public interface exposes common interactions and looks like this:

```python
class Result:
    def read(self, filepath) -> Result:
    "Reads from the provided filepath and returns a new Result object with the underlying data"

    def write(self, value, **kwargs) -> Result:
    "Formats self.filepath w/ kwargs and writes the data to the formatted filepath, returning a *new* Result instance"

    def exists(self, filepath) -> bool:
    "Checks whether the filepath exists / is compatible with the result interface"
```

Under the hood, Prefect itself will _also_ use this class as an ephemeral container for carrying around actual data (on the `value` attribute) but users don't necessarily need to know this (unless they explicitly call `read` within their tasks and need to access the read data).

src/prefect/engine/results/prefect_result.py
```diff
@@ -0,0 +1,67 @@
...
+        new.value = value
+        new.filepath = json.dumps(new.value)
+        return new
+
+    def exists(self, filepath: str) -> bool:
+        """
+        Confirms that the provided value is JSON deserializable.
+
+        Args:
+             - filepath (str): the value to test
```
joshmeek, line 58:
```suggestion
            - filepath (str): the value to test
```

joshmeek: 😆

src/prefect/engine/results/s3_result.py
```diff
@@ -105,89 +105,88 @@ def __getstate__(self) -> dict:
...
         stream = io.BytesIO(binary_data)
 
         ## upload
-        self.client.upload_fileobj(
-            stream, Bucket=self.bucket, Key=self._rendered_filepath
-        )
-        self.logger.debug(
-            "Finished uploading result to {}.".format(self._rendered_filepath)
-        )
+        self.client.upload_fileobj(stream, Bucket=self.bucket, Key=new.filepath)
```
joshmeek, line 130: Should the upload also handle exceptions similar to https://github.com/PrefectHQ/prefect/blob/3bc254d759b4a86d215b320ce14085898085e1f8/src/prefect/environments/storage/s3.py#L170-L178 where we log the error then raise?

cicdw: Yea I think we should actually - I updated the exception handling logic within both the read and write methods.

src/prefect/engine/results/constant_result.py
```diff
@@ -16,37 +16,37 @@ def __init__(self, value: Any = None, **kwargs: Any) -> None:
...
         """
         return self
 
-    def write(self, **kwargs: Any) -> Result:
+    def write(self, value: Any, **kwargs: Any) -> Result:
         """
         Returns the repr of the underlying value, purely for convenience.
 
         Args:
+            - value (Any): the new value to store on the class
```
lauralorenz, line 33: Is this what we want? I thought constant results values were not supposed to be writable, only provided at instantiation

cicdw: I totally agree it's weird; I struggled a little with this.  However, if I pedantically view `Result`s as interfaces to backends, the `ConstantResult` has the class instance as its backend (at least that's what I reasoned).

Luckily because of how constants are unpacked within the runners this method will never actually be called (and I also don't expect users to interact with Constant Result types either).  

Open to change it if it's just too weird.

lauralorenz: I think it needs to change one of two ways -- work like it used to and stay named `ConstantResult`, or work like this and get a new name. (I don't like calling it ConstantResult if we're not making any guarantees on it never changing, even though we don't expect a user to use it.) A new name makes sense if it's not actually important that its maintains its constant value, but more that it a `PythonResult` (going with our storage theme for naming). 

That being said it seems it just so happens to stay constant in pipeline logic... you say 'Luckily' for that one which gives me ookies: I feel like that means we will have to enforce that better going forward. Arguably that is easier by making the ConstantResult truly constant. Alternatively it seems would could add (unless they exist?) tests proving constants don't overwrite their original values against whatever task runner method does the constant unpacking.

cicdw: @lauralorenz How about I raise an error if the provided `value` is not equal to `self.value`?  That way it's enforced to be constant.

lauralorenz: Yep, that is good with me!

cicdw: @lauralorenz updated 👍

PR #2297, opened 2020-04-07
Fix flow parameter registration

### LINKED ISSUE #2296: Can't pass parameters to a Clock unless they have default values
## Description
I would like to create a flow on a schedule, passing in default parameters. Currently this fails unless all parameters have default values.

## Expected Behavior
Passing parameters should be allowed even if they do not have defaults - we should be able to resolve the following example

## Reproduction
```from prefect import task, Parameter, Flow

from prefect.schedules import Schedule
from prefect.schedules.clocks import CronClock


@task
def add(a, b):
    return a + b


schedule = Schedule(clocks=[
    CronClock(
        cron=f"* * * * *",
        parameter_defaults={
            'a': 1,
        }
    )
])

with Flow('flow', schedule=schedule) as flow:
    a = Parameter('a')
    b = Parameter('b', default=1)
    result = add(a, b)

flow.register()
```

## Environment

cicdw:

## What does this PR change?
This PR fixes an issue server-side with registering flows that have required parameters which are scheduled.  This fix was previously implemented Client-side.

Closes #2296

### DIFF
--- a/server/src/prefect_server/api/flows.py
+++ b/server/src/prefect_server/api/flows.py
@@ -88,7 +88,7 @@ async def create_flow(
         required_names = {p.name for p in required_parameters}
         if not all(
             [
-                required_names == set(c.parameter_defaults.keys())
+                required_names <= set(c.parameter_defaults.keys())
                 for c in f.schedule.clocks
             ]
         ):
--- a/server/src/prefect_server/api/flows.py
+++ b/server/src/prefect_server/api/flows.py
@@ -83,8 +83,17 @@ async def create_flow(
         f = prefect.serialization.flow.FlowSchema().load(copy.deepcopy(serialized_flow))
     except Exception as exc:
         raise ValueError(f"Invalid flow: {exc}")
-    if f.schedule and any([p.required for p in f.parameters()]):
-        raise ValueError("Can not schedule a flow that has required parameters.")
+    required_parameters = [p for p in f.parameters() if p.required]
+    if f.schedule is not None and required_parameters:
+        required_names = {p.name for p in required_parameters}
+        if not all(
+            [
+                required_names == set(c.parameter_defaults.keys())
+                for c in f.schedule.clocks
+            ]
+        ):
+            raise ValueError("Can not schedule a flow that has required parameters.")
+
     # set up task detail info
     reference_tasks = f.reference_tasks()
     root_tasks = f.root_tasks()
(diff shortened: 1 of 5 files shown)


server/src/prefect_server/api/flows.py
```diff
@@ -83,8 +83,17 @@
...
     except Exception as exc:
         raise ValueError(f"Invalid flow: {exc}")
-    if f.schedule and any([p.required for p in f.parameters()]):
-        raise ValueError("Can not schedule a flow that has required parameters.")
+    required_parameters = [p for p in f.parameters() if p.required]
+    if f.schedule is not None and required_parameters:
+        required_names = {p.name for p in required_parameters}
+        if not all(
+            [
+                required_names == set(c.parameter_defaults.keys())
```
jlowin, line 91: I think this might need some set intersection to be complete.

For example, if I have two clocks - one that provides a default for a, and one that provides defaults for a and b, then I think `required_names == {'a'}` but `set(c.parameter_defaults.keys()) == {'a', 'b'}`

server/tests/api/test_flows.py
```diff
@@ -222,6 +222,18 @@ class TestCreateFlow:
         flow = await m.Flow.where(id=flow_id).first("description")
         assert flow.description == description
 
+    async def test_create_flow_intelligently_handles_scheduled_param_defaults(self):
+        a, b = prefect.Parameter("a"), prefect.Parameter("b", default=1)
+        clock = prefect.schedules.clocks.CronClock(
+            cron=f"* * * * *", parameter_defaults={"a": 1,}
```
jlowin, line 228: To align with my suggestion above, I think a second clock with defaults `{"a": 1, "b": 2}` would be interesting

server/src/prefect_server/api/flows.py
cicdw: That wouldn't work though -- each unique scheduled clock event needs values for all required parameters.

jlowin: Right, but in addition to supplying a default for each required parameter, a clock could include defaults for non-required parameters as well. Therefore as long as the clock supplies *as least* the required keys, it's good to go. The current check says it must supply *only* the required keys.

cicdw: Ahhhh sorry I understand

cicdw: @jlowin updated

PR #2299, opened 2020-04-08
Fix some interpolation of config in Prefect server dev commands

lauralorenz:
- Convert ui_path when read from environment variable into a Path object.
- Use config file's host and port for hasura and apollo when starting services with `dev services`.

## What does this PR change?
Fixes some config interpolation that wasn't happening when running services independently/not with docker.

## Why is this PR important?
Needed to dev on Apollo from Prefect Core; Apollo services could not build Apollo schema or healthcheck Apollo when running from dev services without it. Also needed to be able to build the UI from the env variable path instead of the default.


jlowin:
@lauralorenz makes total sense for dev work but I think these env vars set the docker-compose environment as well, and I think that `localhost` won't properly be recognized inside the docker compose environment; it'll need to resolve the service name ("hasura", "graphql", etc.)

lauralorenz:
@jlowin 

I have that the `prefect server start` command aka the docker compose based setup is controlled by config set in this file: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/server.py#L26-L39

and that the changes in this PR are to a different file, which is only used by the `prefect-server dev services` and `prefect-server infrastructure` commands.

jlowin:
Ah, I apologize, I didn't realize it was two different files! 👍

PR #2300, opened 2020-04-08
Closes #2224. Converter for result handlers to newstyle Results, including custom result handlers

### LINKED ISSUE #2224: PIN 16: Add `CustomResult` that will be used to hydrate user Results from cloud states
## Use Case
See [PIN 16](https://docs.prefect.io/core/PINs/PIN-16-Results-and-Targets.html) for more motivation on new style Result classes. 

Today, users may provide their own custom `ResultHandler` subclasses that do not serialize to Cloud; we expect to support the same for custom new style `Result` subclasses. 

## Solution
Add a `CustomResult` that wraps any custom result handler implementation or Result subclass from the user that we can serialize to Cloud and instantiate in `State._result` for a user when hydrating States from Cloud.

The important method to implement is:
`CustomResult.hydrate(result: Result) -> Result`
This should return an instance of the result arg, since the pipeline will call this a la `state._result.hydrate(result=self.task.result)`.

Other methods common to Results like store_safe_value, validate, read, write etc should not be implemented. (Decide if it should raise its own kind of warning or not especially if we miss an edge case and then get the NotImplementedError from tbhe base class.)

## Alternatives

lauralorenz:

## What does this PR change?
Closes #2224. Adds in a new Result class that holds logic to convert result handlers into Result subclasses (for backwards compatability with flows configured with result handlers, including flows configured with custom result handlers.)

## Why is this PR important?
Preserves backwards compatibility of the PIN 16 changes with result handlers


cicdw:
@joshmeek / @lauralorenz this is ready for review.  You'll note that I avoided the `AzureResult` conversion for now because there are still unanswered questions about what default authentication with Azure looks like.

Issue #2298, opened 2020-04-08
Refactor trigger signature


cicdw:
In place of PIN 15, we should refactor the signature of Prefect trigger functions.  In particular, instead of receiving an unordered set of upstream states triggers should instead receive a dictionary of `{Edge: State}`.  This will allow for many new patterns:

- more nuanced skip behavior based on whether an upstream dependency is a data dependency or a state dependency (based on checking `edge.key`)
- more nuanced mapping behavior (based on `edge.mapped`)
- the ability to write custom triggers which respond specifically to certain upstream task states (via `edge.upstream_task`)

Note that this is a breaking change for any users with custom triggers, and as such belongs in the 0.11.0 release branch

PR #2301, opened 2020-04-08
Update trigger signatures to accept a dictionary of [Edge, State]

joshmeek:

## What does this PR change?
Addresses #2298 by refactoring trigger signatures from a set of states to a dictionary of edges and states.

**Breaking change** so being merged into the 0.11.0 release branch

## Why is this PR important?
This will allow for more customized triggers in the future

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -430,15 +430,15 @@ class TaskRunner(Runner):
             - ENDRUN: if the trigger raises an error
         """
 
-        all_states = set()  # type: Set[State]
-        for upstream_state in upstream_states.values():
-            if isinstance(upstream_state, Mapped):
-                all_states.update(upstream_state.map_states)
-            else:
-                all_states.add(upstream_state)
+        # all_states = set()  # type: Set[State]
+        # for upstream_state in upstream_states.values():
+        #     if isinstance(upstream_state, Mapped):
+        #         all_states.update(upstream_state.map_states)
+        #     else:
+        #         all_states.add(upstream_state)
 
         try:
-            if not self.task.trigger(all_states):
+            if not self.task.trigger(upstream_states):
                 raise signals.TRIGGERFAIL(message="Trigger failed")
 
         except signals.PrefectStateSignal as exc:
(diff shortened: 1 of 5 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -430,12 +430,12 @@ def check_task_trigger(
...
-        for upstream_state in upstream_states.values():
+        # Deep copy upstream states and expand mapped states
+        all_states = copy.deepcopy(upstream_states)
+        for edge, upstream_state in upstream_states.items():
             if isinstance(upstream_state, Mapped):
-                all_states.update(upstream_state.map_states)
-            else:
-                all_states.add(upstream_state)
+                for mapped_state in upstream_state.map_states:
+                    all_states[copy.deepcopy(edge)] = mapped_state
```
joshmeek, line 438: Is there a better way to do this? 🤔 I was thinking that we make copies of the edges to put in the dictionary but now I can forsee how these edges will be hard to reason with (e.g. multiple copies of the "same" edge used as keys)

Maybe the values should be a list of States

cicdw: Note: this implementation doesn't work.  You can't have the same key with multiple values -- only the last set value will have any effect here.

cicdw: I'm tempted to say you can let the trigger handle this --> within the trigger it can detect the presence of Mapped upstreams (via the Edge) and do whatever it wants to do (such as unpacking `map_states`).

cicdw: In fact, I think you could take the logic from within the task runner and put it inside each new trigger function and be good.

src/prefect/triggers.py
```diff
@@ -46,20 +46,40 @@ def task_c():
...
 from prefect import context
 from prefect.engine import signals
+from prefect.engine.state import Mapped
 
 if TYPE_CHECKING:
     from prefect.engine import state  # pylint: disable=W0611
     from prefect import core  # pylint: disable=W0611
 
 
+def _get_all_states_as_set(upstream_states: Dict["core.Edge", "state.State"]) -> set:
```
cicdw, line 56: nice

PR #2302, opened 2020-04-09
Minor secrets refactor

cicdw:

## What does this PR change?
This PR does a few things:
- renames our secret tasks for clarity
- allows for new secret names to be passed at runtime
- removes `env_var` as a distinct option from `EnvVarSecret`

## Why is this PR important?
This PR was originally motivated by our work on Results (additional PR for `SecretResult` incoming), and I went ahead and changed the names as well as this has been a common source of confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,11 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Enhancements
 
-- None
+- Allow for dynamically changing secret names at runtime - [#2302](https://github.com/PrefectHQ/prefect/pull/2302)
 
 ### Task Library
 
-- None
+- Rename the base secret tasks for clarity - [#2302](https://github.com/PrefectHQ/prefect/pull/2302)
 
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -29,7 +29,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Breaking Changes
 
-- None
+- Remove `env_var` initialization from `EnvVarSecret` in favor of `name` - [#2302](https://github.com/PrefectHQ/prefect/pull/2302)
```
cicdw, line 32: Note that I don't believe anyone is actually using this task, as it's more of a nice example than a truly useful thing.

src/prefect/tasks/secrets/base.py
```diff
@@ -3,16 +3,37 @@
 from prefect.client.secrets import Secret as _Secret
 from prefect.core.task import Task
 from prefect.engine.result_handlers import SecretResultHandler
+from prefect.utilities.tasks import defaults_from_attrs
 
 
-class Secret(Task):
```
joshmeek, line 8: To ensure the non-breaking of this we could keep the `Secret` name w/ a deprecation warning and formally drop in 0.11.0

cicdw: FYI I do keep the `Secret` name at the bottom of the file; I need to think about how I can raise a deprecation warning if it's used though 🧐

cicdw: ah I know, will update

cicdw: @joshmeek updated

PR #2303, opened 2020-04-09
Relax bool cast on operators

### LINKED ISSUE #2271: Operator task behavior is unintuitive/misleading
The docstrings for the various logical `Operator` tasks suggest that they just call the corresponding Python operator. So I would expect `Or(x, y)` to return `x or y` instead of `bool(x or y)`: https://github.com/PrefectHQ/prefect/blob/07cb6c20342c4053a3eaaaf885670ae5dfd6e74d/src/prefect/tasks/core/operators.py#L233-L254.

If there's a specific reason why these have to return boolean results, then at least the docstrings should probably be updated to reflect that.

jcrist:

## What does this PR change?

Previously conditional operators also cast their result to `bool`. This
relaxes that constraint. Note that `Not` still returns bool, since
this is required in Python by `__bool__` (which `not x` calls).

Fixes #2271.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Enhancements
 
-- None
+- Task logical operators (e.g. `And`, `Or`, ...) no longer implicitly cast to `bool` - [#2303](https://github.com/PrefectHQ/prefect/pull/2303)
 
 ### Task Library
 
--- a/src/prefect/tasks/core/operators.py
+++ b/src/prefect/tasks/core/operators.py
@@ -218,16 +218,16 @@ class And(Task):
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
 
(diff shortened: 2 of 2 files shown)


jcrist:
Note that I didn't:

- Add tests, as I'm not sure what the testing policy for this repo is. Testing that the code doesn't do something (cast to bool) seems like a bit noisy of a test for me, but I'm also happy to add them.
- Update the changelog. I'm not sure what deserves a changelog entry.

cicdw:
@jcrist awesome thanks!

- for tests, we're usually biased to add tests partially for documentation of intent and partially for bug finding; in this particular case I could go either way so it's fine not to include them
- similarly for the changelog, generally we try to add an entry for any behavior change / signature change even if it's small; in this case since it's a user-opened issue and also represents different behavior in some cases (tasks which return `set`s come to mind) I'd vote to include a changelog entry, probably under "Enhancements" in this instance

jcrist:
Thanks Chris. I added a changelog entry, but decided that tests don't really make sense here. Should be good-to-go.

PR #2305, opened 2020-04-10
Add azure result

cicdw:
Adds AzureResult.  Needs:

- [x] tests
- [x] a working `exists` method

If anyone is listening, feel free to commit directly to my branch ;)

### DIFF
--- a/src/prefect/engine/results/azure_result.py
+++ b/src/prefect/engine/results/azure_result.py
@@ -1,0 +1,166 @@
+import json
+import os
+import uuid
+from typing import TYPE_CHECKING, Any
+
+import cloudpickle
+import pendulum
+
+from prefect.client import Secret
+from prefect.engine.result import Result
+
+if TYPE_CHECKING:
+    import azure.storage.blob
+
+
+class AzureResult(Result):
+    """
+    Result for writing to and reading from an Azure Blob storage.
+
+    Note that your flow's runtime environment must be able to authenticate with
+    Azure; there are currently two supported options: provide a connection string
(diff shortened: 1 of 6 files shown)


src/prefect/engine/results/azure_result.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # initialize client and download
+        client = self.service.get_blob_client(container=self.container, blob=location)
+
+        # Catch exception because Azure python bindings do not yet have an exists method
+        # https://github.com/Azure/azure-sdk-for-python/issues/9507
+        try:
+            client.get_blob_properties()
+            return True
+        except HttpResponseError:
```
lauralorenz, line 167: From the issue in the comment linked above, it recommended catching a `ResourceNotFoundError` specifically for the exists problem (which according to the docs at least is for 404s, not sure it actually gets raised or not which maybe is why you picked this one: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/exceptions.py#L308), are we preferring `HttpResponseError` on purpose?

src/prefect/engine/results/gcs_result.py
```diff
@@ -93,9 +91,10 @@ def read(self, location: str) -> Result:
             - location (str): the GCS URI to read from
 
         Returns:
-            - Any: the read result
+            - Result: the read result
         """
         new = self.copy()
+        new.location = location
```
lauralorenz, line 97: lol THIS BUG was messing with me today and reading this is triggering

PR #2307, opened 2020-04-10
Closes #2168 Update taskrunner pipelines to use newstyle Results

### LINKED ISSUE #2168: PIN 16: 0.11.0: Pipeline logic updates between new Result methods and result handlers in Result

lauralorenz:

## What does this PR change?
Updates the taskrunner pipelines (normal `TaskRunner` and `CloudTaskRunner`) to use newstyle Results and their methods everywhere, and also add Result subclass serializers so they can be serialized to Cloud/Server.

## Why is this PR important?
This sets newstyle Results to be the default used by the pipeline. (Notably, backwards compatibility with result handlers is separated in #2300.)

### DIFF
--- a/src/prefect/engine/cloud/task_runner.py
+++ b/src/prefect/engine/cloud/task_runner.py
@@ -215,15 +215,15 @@ class CloudTaskRunner(TaskRunner):
             for candidate_state in cached_states:
                 assert isinstance(candidate_state, Cached)  # mypy assert
                 candidate_state.cached_inputs = {
-                    key: res.to_result(inputs[key].result_handler)  # type: ignore
+                    key: res.populate_result(inputs[key])  # type: ignore
                     for key, res in (candidate_state.cached_inputs or {}).items()
                 }
                 sanitized_inputs = {key: res.value for key, res in inputs.items()}
                 if self.task.cache_validator(
                     candidate_state, sanitized_inputs, prefect.context.get("parameters")
                 ):
-                    candidate_state._result = candidate_state._result.to_result(
-                        self.task.result_handler
+                    candidate_state._result = candidate_state._result.populate_result(
+                        self.task.result
                     )
                     return candidate_state
 
--- a/src/prefect/engine/cloud/task_runner.py
+++ b/src/prefect/engine/cloud/task_runner.py
@@ -46,16 +46,9 @@ class CloudTaskRunner(TaskRunner):
             retrieving and storing state results during execution (if the Task doesn't already have one)
     """
 
-    def __init__(
-        self,
-        task: Task,
-        state_handlers: Iterable[Callable] = None,
-        result_handler: ResultHandler = None,
-    ) -> None:
+    def __init__(self, task: Task, state_handlers: Iterable[Callable] = None) -> None:
         self.client = Client()
-        super().__init__(
-            task=task, state_handlers=state_handlers, result_handler=result_handler
-        )
+        super().__init__(task=task, state_handlers=state_handlers)
 
     def _heartbeat(self) -> bool:
         try:
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -566,11 +566,9 @@ class FlowRunner(Runner):
 
         """
         with prefect.context(self.context):
-            default_handler = task.result_handler or self.flow.result_handler
             task_runner = self.task_runner_cls(
                 task=task,
                 state_handlers=task_runner_state_handlers,
-                result_handler=default_handler,
             )
 
             # if this task reduces over a mapped state, make sure its children have finished
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -567,8 +567,7 @@ class FlowRunner(Runner):
         """
         with prefect.context(self.context):
             task_runner = self.task_runner_cls(
-                task=task,
-                state_handlers=task_runner_state_handlers,
+                task=task, state_handlers=task_runner_state_handlers,
             )
 
             # if this task reduces over a mapped state, make sure its children have finished
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -118,13 +118,7 @@ class Result(ResultInterface):
         if self.value is None:
             return
         if self.safe_value == NoResult:
-            assert isinstance(
-                self.result_handler, ResultHandler
-            ), "Result has no ResultHandler"  # mypy assert
-            value = self.result_handler.write(self.value)
-            self.safe_value = SafeResult(
-                value=value, result_handler=self.result_handler
-            )
+            self.safe_value = SafeResult(value=self.filepath)
 
     def populate_result(self, result: "Result") -> "Result":
         """
@@ -267,7 +261,7 @@ class SafeResult(ResultInterface):
         - result_handler (ResultHandler): the result handler to use when reading this result's value
     """
 
-    def __init__(self, value: Any, result_handler: ResultHandler):
+    def __init__(self, value: Any, result_handler: ResultHandler = None):
         self.value = value
         self.result_handler = result_handler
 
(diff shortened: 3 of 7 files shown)


src/prefect/engine/cloud/task_runner.py
```diff
@@ -46,16 +46,9 @@ class CloudTaskRunner(TaskRunner):
...
 
-    def __init__(
-        self,
-        task: Task,
-        state_handlers: Iterable[Callable] = None,
-        result_handler: ResultHandler = None,
-    ) -> None:
+    def __init__(self, task: Task, state_handlers: Iterable[Callable] = None) -> None:
         self.client = Client()
-        super().__init__(
```
lauralorenz, line 56: We ended up deciding that these actually need to stay because this is how global configuration of result handlers is set by people. So I think I still need to add this back in (and also added back in the places below where I had taken passing it in out from any instantiation of a task runner)

src/prefect/engine/result/base.py
```diff
@@ -120,13 +118,7 @@ def store_safe_value(self) -> None:
...
             return
         if self.safe_value == NoResult:
-            assert isinstance(
-                self.result_handler, ResultHandler
-            ), "Result has no ResultHandler"  # mypy assert
-            value = self.result_handler.write(self.value)
-            self.safe_value = SafeResult(
-                value=value, result_handler=self.result_handler
-            )
+            self.safe_value = SafeResult(value=self.filepath)
```
lauralorenz, line 121: So far I've left the SafeResult stuff still in here, though I've unhooked what uses it (the serializer's `validation_fn=get_safe` I delete below), so need to consider if any calls of `store_safe_value` should still exist.

src/prefect/engine/task_runner.py
```diff
@@ -713,9 +709,9 @@ def run_mapped_task(
                                         preview=repr(upstream_state.result)[:10],
                                     )
                                 )
-                            upstream_result = Result(
+                            upstream_result = Result(  # todo: here
```
lauralorenz, line 712: Not super confident what to be doing here (if it's more like the task inputs/cached inputs checks, or more like the loops) so would appreciate some more discussion on this.

src/prefect/engine/task_runner.py
```diff
@@ -893,28 +889,27 @@ def get_task_run_state(
         except signals.LOOP as exc:
             new_state = exc.state
             assert isinstance(new_state, Looped)
-            new_state.result = Result(
-                value=new_state.result, result_handler=self.result_handler
+            new_state.result = self.task.result.write(
+                new_state.result, **prefect.context
```
lauralorenz, line 893: I think this is taking advantage of the fake that `new_state.result`, as opposed to `new_state._result`, peels out `new_state._result.value` here, but it was pretty weird reading so want to double check that. It does make this call "look" different from all the other ones.

src/prefect/engine/task_runner.py
```diff
@@ -893,28 +889,27 @@ def get_task_run_state(
...
-        ## checkpoint tasks if a result_handler is present, except for when the user has opted out by disabling checkpointing
+        # checkpoint tasks if a result_handler is present, except for when the user has opted out by disabling checkpointing
         if (
             state.is_successful()
             and prefect.context.get("checkpointing") is True
             and self.task.checkpoint is not False
-            and self.result_handler is not None
         ):
-            state._result.store_safe_value()
+            state._result.store_safe_value()  # todo: here
```
lauralorenz, line 912: In the discussion PR @cicdw mentions maybe this whole idea goes away since we are greedily writing results already and the serializers are also already going to take care of sending the filepath (https://github.com/PrefectHQ/prefect/pull/2278#discussion_r404316357); at one point we had an IRL discussion about maybe checking in or during `.write()` whether checkpointing is off or not if we still want to turn this off. Thoughts? How does it play with cloud's checkpoint-all-the-time feature?

src/prefect/serialization/result.py
```diff
@@ -1,16 +1,50 @@
...
+    bucket = fields.String(allow_none=False)
+    credentials_secret = fields.String(allow_none=True)
+
+
+class ConstantResultSchema(ObjectSchema):
+    class Meta:
+        object_class = results.ConstantResult
+
+
+class ResultSchema(ObjectSchema):
```
lauralorenz, line 38: I've been slowly changing most every reference to instantiating a base class `Result` in the pipeline to instantiating a `PrefectResult` in the pipeline. Do we really need serializers for base class Results? I empirically observed and discussed on a call with @cicdw right now this is necessary for as long as `Result` is the default for States:  https://github.com/PrefectHQ/prefect/blob/d593e0e7a0a02d25f441b3d0e942614471ee604a/src/prefect/engine/state.py#L85-L90
Do we think we should change that default to `PrefectResult` now?

src/prefect/serialization/state.py
```diff
@@ -36,10 +36,10 @@ class Meta:
 
     context = fields.Dict(key=fields.Str(), values=JSONCompatible(), allow_none=True)
     message = fields.String(allow_none=True)
-    _result = Nested(StateResultSchema, allow_none=False, value_selection_fn=get_safe)
+    _result = Nested(StateResultSchema, allow_none=False, value_selection_fn=None)
```
lauralorenz, line 39: On a pairing call with @cicdw there was an implication that somewhere around here (or maybe it was just that his memory about it was triggered around here) we need to be worried about maintaining deserialization of old style `State._result` in pipeline logic, so calling that out here in case we want to have the discussion at this location.

src/prefect/engine/task_runner.py
```diff
@@ -152,12 +148,12 @@ def initialize_run(  # type: ignore
                 "task_loop_count": state.cached_inputs.pop(  # type: ignore
                     "_loop_count"
                 )  # type: ignore
-                .to_result()
+                .populate_result(self.task.result)  # todo: here
```
lauralorenz, line 151: I'm a little confused how/why this case is different from a similar case below (https://github.com/PrefectHQ/prefect/pull/2307/files#diff-7e1afdbbda2b3755cc338f22cc84d9dcR972). On the discussion PR there was a suggestion that the serializers handle this instead somehow (https://github.com/PrefectHQ/prefect/pull/2278#discussion_r404311132). Not feeling confident about what exactly this is doing so would like to get some more info on that.

src/prefect/engine/cloud/task_runner.py
cicdw: Yea this is the place where the _Flow_ can pass it's `Result` type down in case the `task` doesn't have one set.  So I think we should:
- expose `result` here and pass it down from the flow
- `self.result = self.task.result or result` at init
- replace `self.task.result` everywhere in the pipeline with `self.result`

src/prefect/engine/flow_runner.py
```diff
@@ -566,11 +566,8 @@ def run_task(
 
         """
         with prefect.context(self.context):
-            default_handler = task.result_handler or self.flow.result_handler
             task_runner = self.task_runner_cls(
-                task=task,
-                state_handlers=task_runner_state_handlers,
-                result_handler=default_handler,
```
cicdw, line 573: Yea I think we should re-introduce this init logic for results

src/prefect/engine/result/base.py
cicdw: `store_safe_value` should be equivalent to `write` so I think we we can remove all those calls.  As discussed offline, we do need to keep the `SafeResult` class itself for backwards compatibility of deserialization logic.

src/prefect/engine/flow_runner.py
lauralorenz: Reintroduced result_handler forwarding (for backwards compatability/conversion) in  af59d67 and added in result forwarding in e1f8e2f

src/prefect/engine/cloud/task_runner.py
lauralorenz: Reintroduced result_handler forwarding (for backwards compatability/conversion) in af59d67 and added in result forwarding/`self.result` instead of `self.task.result` stuff in e1f8e2f

src/prefect/engine/result/base.py
lauralorenz: Ok ripped most everything out re: store_safe_result in 2125c24

joshmeek:
Can this be closed in favor of #2391?

PR #2310, opened 2020-04-10
Control flow tasks return executed branch

### LINKED ISSUE #1735: `ifelse` should behave like a ternary expression
Currently, our conditional `ifelse` operator just sets dependencies in a particular way, but always returns `None`.  It would feel much more natural if it returned a task object whose value at runtime behaved like a standard `x = condition ? if_true : if_false` expression

jcrist:
This modifies `ifelse` and `switch` to return a task for the branch that
was run. Previously users had to explicitly use `merge` to achieve the
same behavior.

Fixes #1735.


jcrist:
Hmm, test only fails on Python 3.6, debugging now.

jcrist:
It was a bug in the test, fixed and pushed.

PR #2326, opened 2020-04-14
Better secrets

cicdw:

## What does this PR change?
This PR is a large refactor of how built-in Prefect hooks authenticate with third party services; in particular:
- all credentials secret kwargs are removed (they were previously deprecated) 
- a new feature for auto provisioning creds in context through storage is provided
- a new document describing the new opinionated model (see "Third Party Authentication" in the netlify preview)

I've done both AWS and GCP in this PR; I vote that we discuss / review the current implementation and then let the remaining task hooks / result hooks be additional PRs after this one:

- Azure
- DB tasks (e.g., Postgres and Snowflake)

## Why is this PR important?
Clarifies and cleans up a lot of our secrets / auth cruft


cicdw:
@joshmeek / @lauralorenz I pointed this at the release/0.11.0 branch now

docs/orchestration/recipes/third_party_auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+In some instances you may need to override the global options presented above on a per-task basis, or you may choose to explicitly provide credentials to every task that needs them.  Either way, all Prefect tasks from the Task library which require credentials to authenticate offer optional runtime arguments for providing this information.
+
+
+### List of default Secret names
+
+The following is a list of the default names and contents of Prefect Secrets that, if set and declared, can be used to automatically authenticate your flow with the listed service:
+
+- `GCP_CREDENTIALS`: a dictionary containing a valid [Service Account Key](https://cloud.google.com/docs/authentication/getting-started)
+- `AWS_CREDENTIALS`: a dictionary containing two keys: `ACCESS_KEY` and `SECRET_ACCESS_KEY` which are passed directly to the `boto3` client
```
joshmeek, line 86: We should link to https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html and also expand the default `AWS_CREDENTIALS` to accept a session token

docs/orchestration/recipes/third_party_auth.md
```diff
@@ -0,0 +1,86 @@
+# Third Party Authentication
```
joshmeek, line 1: This doc is really well written! If we keep it in this location do you think we should expand the other "recipes" in this section to become more like this? The others feel to me more like bits of code while this doc reads like a concept doc.

src/prefect/cli/execute.py
```diff
@@ -62,13 +63,19 @@ def cloud_flow():
...
+        ## populate global secrets
+        secrets = prefect.context.get("secrets", {})
+        for secret in storage.secrets:
+            secrets[secret] = PrefectSecret(name=secret).run()
 
-        environment.setup(storage=storage)
-        environment.execute(
-            storage=storage, flow_location=storage.flows[flow_data.name]
-        )
+        with prefect.context(secrets=secrets):
```
joshmeek, line 71: We may need to propagate secrets down during execution. I don't think this will work for flows using environments like the `DaskKubernetesEnvironment` because the `execute` step creates a "scheduler" job that calls the [run_flow](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/environments/execution/dask/k8s.py#L261) function from the environment

src/prefect/engine/results/gcs_result.py
```diff
@@ -15,21 +15,22 @@ class GCSResult(Result):
...
-        - credentials_secret (str, optional): the name of the Prefect Secret
-            which stores a JSON representation of your Google Cloud credentials.
         - **kwargs (Any, optional): any additional `Result` initialization options
     """
 
     def __init__(
         self, bucket: str = None, credentials_secret: str = None, **kwargs: Any
     ) -> None:
         self.bucket = bucket
+        self.logger = logging.get_logger(type(self).__name__)
```
joshmeek, line 33: The base result class [already has a logger](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/engine/result/base.py#L113)

src/prefect/engine/results/s3_result.py
```diff
@@ -21,25 +19,15 @@ class S3Result(Result):
 
     Args:
```
joshmeek, line 20: Docstring should also be updated like the GCSResult to mention the default secret that it can use

src/prefect/cli/execute.py
joshmeek: Oh nevermind it looks like you already do this!

src/prefect/utilities/aws.py
```diff
@@ -0,0 +1,53 @@
+"""
+Utility functions for interacting with Google Cloud.
```
joshmeek, line 2:
```suggestion
Utility functions for interacting with AWS.
```

src/prefect/utilities/aws.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_boto_client(
+    resource: str, credentials: dict = None, use_session: bool = False, **kwargs: Any
+) -> "boto3.client":
+    """
+    Utility function for loading boto3 client objects from a given set of credentials.
+
+    Args:
+        - resource (str): the name of the resource to retrieve a client for
+        - credentials (dict, optional): a dictionary of Google credentials used to initialize the Client; if
```
joshmeek, line 18:
```suggestion
        - credentials (dict, optional): a dictionary of AWS credentials used to initialize the Client; if
```

docs/orchestration/recipes/third_party_auth.md
cicdw: Hmmm I don't think we need to change them immediately but I think the more guidance we can give in the other docs the better

docs/orchestration/recipes/third_party_auth.md
cicdw: @joshmeek updated per all of your requests

PR #2327, opened 2020-04-14
Initial Zombie Killer implementation

alexisprince1994:
- Created new Zombie Killer service
- Update CLI to use new service
- Update Docker Compose file to include new service

This PR is an initial stab at the Zombie Killer as described in the Prefect Cloud docs. If this is in any way in violation of the license, please let me know and I can take down the PR.

Also, accidentally forgot to remove the changes to `setup.cfg` and `pyproject.toml` as I realized I named my virtualenv as `venv`. Had to add that there or black's check takes 2 minutes :/.

## Outstanding Questions
* Is the docker-compose file in `prefect_server` the source of truth for the services that live as a part of the deployment? I noticed there was both one that existed in `prefect_server` as well as `prefect/src/cli/`.
* As I read the docs, the Zombie Killer only rescues zombie tasks, correct?
* Distressed flows should be handled by the Lazarus process, which is not in scope for this.

## What does this PR change?
This PR adds an initial implementation of the Zombie Killer into the open source orchestration offering.

## Why is this PR important?
This PR is important because rescuing failed tasks is a core feature of workflow orchestration.

### DIFF
--- a/server/src/prefect_server/services/zombie_killer/zombie_killer.py
+++ b/server/src/prefect_server/services/zombie_killer/zombie_killer.py
@@ -1,0 +1,140 @@
+# Licensed under the Prefect Community License, available at
+# https://www.prefect.io/legal/prefect-community-license
+
+
+import asyncio
+
+import pendulum
+import prefect
+from prefect.engine.state import Failed
+from prefect.utilities.graphql import EnumValue
+
+import prefect_server
+from prefect_server import api
+from prefect_server.database import models
+from prefect_server.services.loop_service import LoopService
+
+
+class ZombieKiller(LoopService):
+    """
+    The ZombieKiller is a service that monitors existing task runs for "zombies".
+
+    Zombies are tasks that should be checkpointing with the Orchestration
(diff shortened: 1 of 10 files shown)


server/src/prefect_server/services/zombie_killer/zombie_killer.py
```diff
@@ -0,0 +1,140 @@
...
+                            # the zombie killer would kill it immediately.
+                            {
+                                "_and": [
+                                    {"heartbeat": {"_is_null": True}},
+                                    {"updated": {"_lte": kill_threshold.isoformat()}},
+                                ]
+                            },
+                        ],
+                    },
+                    {  # For backwards compatability, we check whether the
```
jlowin, line 79: This is 💯 but elsewhere in Server we've made the decision not to support versions of Prefect < 0.10.0 (the release that introduced Server) because the burden is too high. I think there's no harm in keeping it in, but just calling it out

jlowin:
This is looking great @alexisprince1994, once you and @lauralorenz hammer out the contrib protocol, let's do it.

cicdw:

Issue #2237, opened 2020-04-01
Remove hardcoded 'localhost' references in Prefect Server UI Vue Apollo JavaScript


szelenka:
## Current behavior
When deploying the Docker image for the `ui` component, it appears to be a simple nginx container serving static webpage assets (JavaScript, CSS, etc.).

The challenge comes when we want to customize the location of the Apollo service inside this Docker container. It seems at this location, it will attempt to read in an environment variable called `VUE_APP_GRAPHQL_HTTP`, but when that's not present, it'll simply default to `http://localhost:4200/graphql/`.
https://github.com/PrefectHQ/prefect/blob/master/server/services/ui/src/vue-apollo.js#L9

This prevents us from modifying the location of Apollo when running the `ui` Docker container.

To contrast this with running on the CLI, you can get around this by calling `npm run serve` and passing in the desired value for `VUE_APP_GRAPHQL_HTTP` (although it seems to also be hardcoded for `localhost`).
https://github.com/PrefectHQ/prefect/blob/master/server/src/prefect_server/cli/services.py#L73

## Proposed behavior
It would be ideal to have the ability to pass in a dynamic location for `VUE_APP_GRAPHQL_HTTP` at runtime for the `ui` Docker container.

This way you can deploy the Prefect Server on a distributed compute platform to gain all the benefits of running on multiple servers, and you're not gated on a single point of failure.

## Example
I hacked around the solution by modifying the entrypoint for the `ui` container, and executing this script to perform a string sub for the desired location of Apollo:
https://github.com/szelenka/prefect-ui/blob/gh-pages/prefect-ui/files/configmap/website_entrypoint.sh

This allows someone to specify the desired location via an environment variable, to dynamically adjust the location of this resource at runtime. Although I'm not sure if that's a valid solution.

adilkhash:
+1, when you run Docker inside virtual machine (Windows 10 Home) you will not be able to access services using localhost hostname.

szelenka:
It's a tad more complicated than just changing the static `localhost` in the javascript files. It'd be possible (but a deviation from Cloud) to have some localstorage component to store the API url in the browser case, but that's kind of hacky.

To really provide a smooth experience, we'd also need to adjust the `prefect backend server` command to update the `backend.toml` with the specified addresses the user wants to use, other than `localhost`.

In the short term, it looks like `prefect server` is intended to run on a dedicated docker host.

jloehel:
:+1: 

[...]

PR #2338, opened 2020-04-15
Ui graphql endpoint

znicholasbrown:

## What does this PR change?
Allows the GraphQL endpoint in the UI to be configured via `config.toml`

## Why is this PR important?
It allows remote deployment of Prefect Server to include a working UI. 

Resolves: #2237

### DIFF
--- a/server/services/ui/shell_scripts/start_server.sh
+++ b/server/services/ui/shell_scripts/start_server.sh
@@ -6,7 +6,7 @@ if [ $PREFECT_SERVER__GRAPHQL_URL != 'http://localhost:4200/graphql' ]
 then
     echo "Second echo: $PREFECT_SERVER__GRAPHQL_URL"
     echo "Replacing graphql references with: $PREFECT_SERVER__GRAPHQL_URL"
-    for i in "/var/www/js/*.js"; do
+    for i in /var/www/js/*.js; do
         echo $i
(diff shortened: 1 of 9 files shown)


server/services/ui/shell_scripts/start_server.sh
```diff
@@ -1,5 +1,15 @@
 #!/usr/bin/env bash
 
+
+if [ $PREFECT_SERVER__GRAPHQL_URL != 'http://localhost:4200/graphql' ]
+then
+    echo "Replacing graphql references with: $PREFECT_SERVER__GRAPHQL_URL"
+    for i in /var/www/js/*.js; do
```
cicdw, line 7: Just out of curiosity how is this different from the expression we were looking at last night?

znicholasbrown: I think the only difference from last night is that I removed the quotes around the file path, since those aren't needed.

server/src/prefect_server/config.toml
```diff
@@ -55,6 +55,9 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
     debug = false
     path = "/graphql/"
 
+    [services.ui]
```
joshmeek, line 58: Why the server/config.toml over the root config.toml?

joshmeek: Oh I see both were updated, nvm

joshmeek:
Tested, worked!

PR #2341, opened 2020-04-16
Update getting started instructions

jlowin:

## What does this PR change?
As Prefect's complexity has grown, especially with the local server, some of our community users are getting a little lost with getting started. For example, a couple common pitfalls I've noticed are:
- not knowing to run `prefect backend server` prior to starting the server
- not knowing to run `prefect agent start` 

This information is available in our docs but as our docs have grown piecemeal, I think the cohesive structure was lost a little bit so it's hard to tell what to look for. As a minor example, the instruction for `prefect backend server` is given in a tip box after `prefect server start`, which might be why some users miss it (even though it's a nice way to offset it).

## Why is this PR important?

I've tried to refactor the "getting started" docs -- mainly the repo README and the "installation" doc -- to organize the information in a way that I think will be more in line with the expectations we're hearing.


docs/core/development/docker.md
```diff
@@ -0,0 +1,22 @@
+# Docker
```
lauralorenz, line 1: @jlowin What's the reasoning for moving these out of installation? I don't get the sense that the docker builds are specifically for development.

jlowin: To be honest I may have overpinned to a conversation I had in the community Slack this morning with someone who was trying to "install" Prefect via Docker and was running into some difficulty; I'm going to move it back. Thanks for pointing that out.

PR #2342, opened 2020-04-16
Document Prefect Cloud roles

### LINKED ISSUE #2331: Document Cloud User Roles
## Current behavior

We currently document token scopes, but not precisely what they mean. We do not, to the best of my knowledge, document the permissions that accompany user roles.

## Proposed behavior

We should document each of Cloud's user roles (user, tenant admin, restricted) and the permissions that come with each of them. 

## Example

It's useful to know exactly what permissions are being assigned with any IAM situation. This issue in particular was prompted by this good question from our public Slack channel: 

https://prefect-community.slack.com/archives/CL09KU1K7/p1586875809246000

zdhughes:

## What does this PR change?

This PR adds documentation about roles in Prefect Cloud and closes #2331.

## Why is this PR important?

We've had folks ask about roles, so it makes sense that we'd want to document them!


docs/orchestration/concepts/roles.md
```diff
@@ -0,0 +1,15 @@
...
+
+User roles are used to denote the permissions a user has as part of a team. There are currently three roles: users, restricted users, and administrators.
+
+### Users
+
+Users have permission to perform all actions required for daily use of Prefect Cloud; this includes registering new flows, kicking off flow runs, etc. They do not have permission to perform actions that impact other members of the team, meaning they cannot change the roles of their teammates, invite others to their tenant, etc.
+
+### Restricted Users
+
+Restricted users are able to view all pages as a normal user would, but are unable to take any actions in Prefect Cloud. For example, they cannot cannot create projects, kick off flow runs, etc. In essence, restricted users are read-only members of a team.
```
joshmeek, line 11:
```suggestion
Restricted users are able to view all pages as a normal user would, but are unable to take any actions in Prefect Cloud. For example, the cannot create projects, kick off flow runs, etc. In essence, restricted users are read-only members of a team.
```

PR #2347, opened 2020-04-16
Auto-remove completed docker containers

### LINKED ISSUE #2344: Cleanup stopped containers in docker agent
Currently containers that finish aren't cleaned up, and will build up over time. We should make use of the `AutoRemove` option on `HostConfig` when starting a container to automatically delete containers after they've completed. This requires docker engine api version >= 1.25, which came out in Jan 2017 - given that it's been around for a few years I think we can apply this universally.

jcrist:
Previously these would accumulate as the agent ran things. We now set
them to autoremove on completion.

Fixes #2344.


znicholasbrown:
Thanks @jcrist !!

PR #2366, opened 2020-04-20
Push bullet notification task

zhen0:

## What does this PR change?
- Adds pushbullet mobile phone notifications to the task library
- Still needs import adding to extras and tests


src/prefect/tasks/notifications/pushbullet.py
```diff
@@ -0,0 +1,47 @@
...
+        super().__init__(**kwargs)
+
+    @defaults_from_attrs("msg")
+    def run(self, msg: str = None) -> None:
+        """
+        Run method for this Task. Invoked by calling this Task after initialization within a Flow context,
+        or by using `Task.bind`.
+
+        Args:
+            - msg (str): The message you want sent to your phone; defaults to the one provided
```
cicdw, line 28: Since this is a Task, I suggest raising an error if `msg` was not provided; something like:
```python
if msg is None:
    raise ValueError("A message must be provided")
```

cicdw: If this were a state handler we could probably craft a default message to send, but as a Task it's harder to anticipate how users might use this

zhen0: Thanks Chris!

zhen0:
Thanks @joshmeek - Updated docs/outline.toml

src/prefect/tasks/notifications/pushbullet_task.py
```diff
@@ -0,0 +1,51 @@
...
+                at initialization
+
+        Raises:
+            - None
+
+        Returns:
+            - None
+        """
+
+        pbtoken = Secret("PUSHBULLET_TOKEN").get()
```
joshmeek, line 43: @cicdw Should we use the newstyle Secrets here?

cicdw: Hmm yea I think we want to expose `token` as an optional keyword argument to this `run` method, and if it's `None` we can fallback on this secret call.  We should call out in the docs that this should be the output of a Secret Task.

PR #2379, opened 2020-04-21
Revert "Control flow tasks return executed branch"

jcrist:
Reverts PrefectHQ/prefect#2310


cicdw:
@jcrist actually I just realized that #2376 adjusted some type annotations

jcrist:
Ah, good catch. Fixed.

PR #2382, opened 2020-04-22
Flow Run Concurrency

alexisprince1994:

- [] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

# Notes / Ongoing Discussions
* There needs to be a larger UX discussion about how users would see a "resource pool". The name "resource pool" itself is also up for debate. It was initially chosen as opposed to "tags" (like on the Tasks) because having a task and flow tagged with the same tags and not having them both share the same resource seems confusing.
* This PR specifically references flow runs for concurrency limiting. If we want to move forward and standardize on "tags", we could implement a shared concurrency between flows and tasks, or keep them separate assuming we could explain it well enough.
* Based on the point above, these "resources" are currently expected to be present within the `settings` of the Flow. This PR _does not_ give any way to add these resources into the `settings`, as I figured our discussions on the first two will likely lead to a change of where they'd be stored.
* Since so much is up for discussion, I did not write the documentation / add to the changelog yet as that would be dependent on the outcome of the discussions. Also, some docstrings aren't 100% accurate as they depend on these discussions as well.
* The descriptions below are based on how the code works before any discussion took place for reviewers to get a better understanding of the initial implementation.
* UX discussion on what "deleting" a resource pool means. Do all flows tagged with this resource get it removed automatically? Do we leave orphaned resource tags? Etc.
* This change only touches on the resources needed to submit new flow runs. This does not handle any changes that would be required for retrying, etc.

EDIT:
(More notes / ongoing discussions)
* As implemented, it isn't a strict guarantee, more of best effort. There are race conditions in play preventing a more strict guarantee about resources being used:

    1. Flow runs in the queue pass resource checks but aren't checked again during flow run submission

    2. Flow runs aren't checked when state is changing to `Running` in the core code (or in the state change mutation)
    3. There isn't a version locking mechanism with resource allocations

* Since this is a first POC of the concept, there isn't any way to prioritize flows with regards to resource capacity. For example, having a differing priority on which flows should be using resources when there are multiple options (between Flows) to run and there aren't enough resources to share between them.

END EDIT.

#### Overview
This PR adds resource pools, which are a means of limiting flow run concurrency.

#### Resource Pool Objects
Resource pools can be created via the `prefect_server.api.resource_pools` (not yet hooked into the API because of ongoing discussions) with a certain number of "slots". Each resource pool is unique by name in the database, and Flows can be associated with resource pools by their name. Each flow run in a running state occupies one slot of the resource, and flow runs will stop being scheduled when the available slots reach 0 until a new slot opens up.

#### Flows
This PR adds the concept of Resource Pools, which are borrowed from Airflow. Like task tags, the resource pools can be used to limit flow run concurrency. Each flow can optionally be tagged with a list of resources (see notes above for open discussion on where that's stored). If flows are not tagged with resource usage, they will continue to operate as they previously have. If a flow is tagged with multiple resources, all resources must have at least one slot available in order for a flow run to be scheduled.

## Why is this PR important?
This PR is important because there is no current way to limit flow level concurrency. There is also no current way to limit concurrency of groups of flows, which may, at the flow level, depend on an external resource.


alexisprince1994:
To get external discussions written down, the following was decided during a synchronous discussion:

* Rename from `ResourcePool` to `FlowConcurrencyLimit` to avoid any confusion with expected behavior from other projects.
* Instead of using a new "concurrency limit" concept, use Execution Environment labels as this feature has been requested (being able to limit by environment like DEV, STAGE, PROD, etc.).
* The concurrency limit should be moved from the `get_runs_in_queue` API to the `set_flow_state` API.
  * We're going to leave the logic in the `get_runs_in_queue` location to avoid a scenario where an Agent will be querying the same X number of flow runs that don't have concurrency slots available, thus blocking any flow runs that may be able to run. However, the ultimate decider of whether a flow can be put into a `Running` state will live in the `set_flow_state` API.

### Things outstanding prior to merge
* Since this contribution will live in `plugins` for a while, a process needs to be finalized for how to alter the "core" code for specific use by the plugins. This logic needs to be implemented both during the flow state setting process and the getting queued flow runs process.
* This likely will require the configuration utilities to be able to load a subset of plugins configuration based on what is enabled. I will put up another PR to merge into this one that has one implementation. Any other discussions on this should be on that PR.

alexisprince1994:
Update on the status of this PR:
As it stands now, the code is pretty "complete". On the action items above, the following decisions were made:
* ResourcePool was renamed to FlowConcurrencyLimit
* FlowConcurrencyLimit processing checks the Execution Environment labels, and does the concurrency computations based on those
* FlowConcurrencyLimits are completely opt in. Untagged execution environments will not be limited. Environments with labels that don't have concurrency limits on them will also not be limited.
* Renamed "slots" to "limit" on the concurrency object for clarity.
* Added check into the flow state setting pipeline to prevent breaking of the concurrency limit

Excluding any functionality or code changes, all that's left to do prior to merge are the following, almost all related to plugin behavior.
* Actually move all possible code from "core" to "contrib"
* Move all tests that use this functionality to "contrib" from their respective locations
* Feature flag code changes that _must_ live in "core" by updating `config.toml`.
* Figure out database migration pattern for plugins (and how to keep Hasura metadata updated)
* Define the process of moving from a plugin from "contrib" to "core"
* Document the code changes once we figure out where the distinction should live between code sections in "contrib" vs "core"
* Update changelog

Things that are left to do that will be outside of this PR are:
* Creating GQL mutations for actually interacting with the FlowConcurrencyLimit plugin

alexisprince1994:
Outstanding action items:
* Figure out database and hasura metadata migrations for plugins
* Ensure docs look okay

server/services/postgres/alembic/versions/2020-04-20T102810_add_flow_concurrency_limits.py
```diff
@@ -0,0 +1,55 @@
...
+    op.execute(
+        """
+        -- Table Definition ----------------------------------------------
+
+        CREATE TABLE flow_concurrency_limit (
+            id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
+            created timestamp with time zone NOT NULL DEFAULT now(),
+            updated timestamp with time zone NOT NULL DEFAULT now(),
+            name character varying NOT NULL UNIQUE,
+            description text,
```
zdhughes, line 30: There's a description column here, but I don't think the ability to modify it is exposed to the user. I'm inclined to say we should nix it for the sake of parity with Cloud, but I'm interested to hear your thoughts.

alexisprince1994: Correct, it isn't able to modified currently. The end goal was to allow users to put arbitrary text explaining either why the limit is in place, what sets of infrastructure is limited by it (E.g. EKS cluster 1, EKS cluster 2), or anything similar. The end idea was to allow users to edit this in the UI, and I'm personally still a fan of surfacing this kind of functionality, but definitely okay with removing it from the table in the short term in favor of having a separate PR that adds specifically that functionality!

cicdw:

Lyalpha:
@cicdw would you mind pointing explicitly to the implementation - I can find no mention of how to limit flow concurrency in prefect anywhere (except the cloud-only task limiting).

alexisprince1994:
@Lyalpha Interesting, I don't see it either! Assuming the functionality isn't hiding somewhere, I can port this PR over to the server repo by early next week! @cicdw Can you confirm it doesn't exist?

cicdw:
Hi @Lyalpha and @alexisprince1994 - flow concurrency limits are not currently in Server; we'd definitely accept a PR for Flow Concurrency as long as the API is identical to the Cloud API for flow concurrency.  I realize that could be a challenge when you can't see the Cloud codebase, so if you'd like to open an issue in https://github.com/PrefectHQ/server requesting this we can look into it at some point hopefully soon!

alexisprince1994:
@cicdw Sounds good to me! Who doesn't appreciate a good challenge?

cicdw:
Haha awesome! if you want to open an issue for it where we can discuss, I can share some of the routes / DB tables you'll need 👍

PR #2386, opened 2020-04-22
Add serializer for RemoteDaskEnvironment and add server start image version logic

### LINKED ISSUE #2369: Add serializer for RemoteDaskEnvironment
The `RemoteDaskEnvironment` from #2367 needs a serializer because currently it defaults to using the `CustomEnvironmentSchema`. I think the only relevant serialized information is the `address` kwarg.

e.g. in UI:
image

### LINKED ISSUE #2375: Iron out steps for resolving serializer changes with Core's server
I am currently working on #2369 and in turn adding a new serializer effectively breaks its use with Core's server due to the version of prefect that the server has being behind (using the `latest` build).

```bash
In [14]: f = Flow('a', environment=RemoteDaskEnvironment(address="yes"))

In [15]: f.register()
Result Handler check: OK
ClientError                               Traceback (most recent call last)
<ipython-input-15-11702a1d3250> in <module>
----> 1 f.register()

~/Desktop/code/prefect/src/prefect/core/flow.py in register(self, project_name, build, labels, set_schedule_active, version_group_id, no_url, **kwargs)
   1371             set_schedule_active=set_schedule_active,
   1372             version_group_id=version_group_id,
-> 1373             no_url=no_url,
   1374         )
   1375         return registered_flow

~/Desktop/code/prefect/src/prefect/client/client.py in register(self, flow, project_name, build, set_schedule_active, version_group_id, compressed, no_url)
    624                     serialized_flow=serialized_flow,
    625                     set_schedule_active=set_schedule_active,
--> 626                     version_group_id=version_group_id,
    627                 )
    628             ),

~/Desktop/code/prefect/src/prefect/client/client.py in graphql(self, query, raise_on_error, headers, variables, token)
    223             elif "Malformed Authorization header" in str(result["errors"]):
    224                 raise AuthorizationError(result["errors"])
--> 225             raise ClientError(result["errors"])
    226         else:
    227             return GraphQLResult(result)  # type: ignore

ClientError: [{'message': "Invalid flow: {'environment': {'type': ['Unsupported value: RemoteDaskEnvironment']}}", 'locations': [{'line': 2, 'column': 5}], 'path': ['create_flow_from_compressed_string'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'errors': [{'message': "Invalid flow: {'environment': {'type': ['Unsupported value: RemoteDaskEnvironment']}}", 'locations': [], 'path': ['create_flow_from_compressed_string']}]}}}]
```

This means that once #2369 (or any future serializers) are merged then when running the Core server images you would need to specify a `--version master` in order for it to work. I'm opening this issue as a discussion point on if this is the functionality we want or if there is a better way.

joshmeek:

## What does this PR change?
Closes #2369 
Closes #2375 

## Why is this PR important?
Now the RemoteDaskEnvironment won't show up as a CustomEnvironment and the image tag selection based on Prefect version is good for local development.

### DIFF
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -91,7 +91,6 @@ def server():
     "--version",
     "-v",
     help="The server image versions to use (for example, '0.10.0' or 'master')",
-    default="master" if len(prefect.__version__.split("+")) > 1 else "latest",
     hidden=True,
 )
 @click.option(
@@ -198,7 +197,7 @@ def start(
     \b
     Options:
         --version, -v   TEXT    The server image versions to use (for example, '0.10.0' or 'master')
-                                Defaults to 'latest'
+                                Defaults to the current installed Prefect version.
         --skip-pull             Flag to skip pulling new images (if available)
         --no-upgrade, -n        Flag to avoid running a database upgrade when the database spins up
         --no-ui, -u             Flag to avoid starting the UI
@@ -277,7 +276,11 @@ def start(
         env = make_env()
 
     if "PREFECT_SERVER_TAG" not in env:
-        env.update(PREFECT_SERVER_TAG=version)
+        env.update(
+            PREFECT_SERVER_TAG=version or "master"
+            if len(prefect.__version__.split("+")) > 1
+            else prefect.__version__
+        )
     if "PREFECT_SERVER_DB_CMD" not in env:
         cmd = (
             "prefect-server database upgrade -y"
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import click
 import yaml
 
+import prefect
 from prefect import config
 from prefect.utilities.configuration import set_temporary_config
 from prefect.utilities.docker_util import platform_is_linux, get_docker_ip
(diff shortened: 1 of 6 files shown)


src/prefect/cli/server.py
```diff
@@ -90,8 +91,7 @@ def server():
     "--version",
     "-v",
     help="The server image versions to use (for example, '0.10.0' or 'master')",
-    # TODO: update this default to use prefect.__version__ logic
-    default="latest",
+    default="master" if len(prefect.__version__.split("+")) > 1 else "latest",
```
cicdw, line 94: Just looking to the future, I think we might want to default to `prefect.__version__` instead of `"latest"`

tests/cli/test_server.py
```diff
@@ -88,6 +89,29 @@ def test_server_start(monkeypatch, macos_platform):
...
 
 
+def test_server_start_image_versions(monkeypatch, macos_platform):
+    check_call = MagicMock()
+    popen = MagicMock(side_effect=KeyboardInterrupt())
+    check_output = MagicMock()
+    monkeypatch.setattr("subprocess.Popen", popen)
+    monkeypatch.setattr("subprocess.check_call", check_call)
+    monkeypatch.setattr("subprocess.check_output", check_output)
+    version = "master" if len(prefect.__version__.split("+")) > 1 else "latest"
```
cicdw, line 99: I think we should parametrize this test with a few mocked values for `prefect.__version__` similar to how we test the docker storage logic

src/prefect/cli/server.py
joshmeek: Ah yeah good call

PR #2389, opened 2020-04-22
Adjust CLI tests for get commands to account for projects

### LINKED ISSUE #2226: Add tests for CLI project / no project logic
As of `0.10.0` some of the CLI commands now make the project name specification optional due to the API differences between Core's server and Cloud. Tests should be added for this behavior

```
# when using Cloud
prefect get flows -n my-flow -p my-project

# when using the server
prefect get flows -n my-flow
```

joshmeek:
Mostly a test update to handle the Core server vs Cloud differences in projects when using the `get` CLI commands. Closes #2226

### DIFF
--- a/src/prefect/cli/get.py
+++ b/src/prefect/cli/get.py
@@ -81,7 +81,7 @@ def flows(name, version, project, limit, all_versions):
 
     headers = ["NAME", "VERSION", "AGE"]
 
-    if config.backend == "cloud" or project:
+    if config.backend == "cloud":
         where_clause["_and"]["project"] = {"name": {"_eq": project}}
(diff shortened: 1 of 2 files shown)


src/prefect/cli/get.py
```diff
@@ -80,7 +81,7 @@ def flows(name, version, project, limit, all_versions):
 
     headers = ["NAME", "VERSION", "AGE"]
 
-    if project:
+    if config.backend == "cloud" or project:
```
cicdw, line 84: What's the idea behind keeping the "or project" piece?

joshmeek: The idea is that there is no idea

PR #2390, opened 2020-04-23
Closes #2332. Use absolute path on windows during docker storage temp directory creation

### LINKED ISSUE #2332: Docker storage cannot locate Dockerfile with relative path on windows OS
## Description
When calling `flow.register()` using a Docker storage configuration that uses a Dockerfile (and not a base Docker image) on windows OS, there is an APIError

```
Exception has occurred: APIError
500 Server Error: Internal Server Error ("Cannot locate specified Dockerfile: .\tmpp4mionwm\Dockerfile")
  File "C:\tmp\prefect-bug\demo.py", line 24, in <module>
    flow.register()
```

After some digging, I observed that `prefect.environments.storage.docker._build_image()` creates a temporary directory that it uses for the Docker build context. However, Docker's `build()` function is unable to find this path. Pausing execution and updating the path to an absolute path resolves the issue. 

image

## Expected Behavior
I expected the flow's storage to be built without error, and for the flow to be registered with the server.

## Reproduction
```
import prefect

DOCKERFILE = """
FROM python:3.7
RUN echo "hello"
"""

with open('Dockerfile', 'w') as file:
    file.write(DOCKERFILE)

@task
def hello():
    print('hello')

with Flow('demo') as flow:
    task()

flow.storage = Docker(
    dockerfile='Dockerfile'
)

flow.register()
```

## Environment
Windows 10 operating system
Prefect 10.1

{
  "config_overrides": {
    "cloud": {
      "use_local_secrets": true
    },
    "context": {
      "secrets": false
    }
  },
  "env_vars": [
    "PREFECT__CLOUD__USE_LOCAL_SECRETS"
  ],
  "system_information": {
    "platform": "Windows-10-10.0.17758-SP0",
    "prefect_version": "0.10.2",
    "python_version": "3.7.4"
  }
}

lauralorenz:

    - though getting a windows test environment up would be cool in the future for things like this

## What does this PR change?
Closes #2332. Adds a check during docker storage whether we are on windows, and if so uses an absolute path for the tempdir created to hold files during docker storage building because of an observed issue with relative paths on Docker for Windows. This may be related to an issue in the `docker-py` repo (https://github.com/docker/docker-py/issues/2127), though that issue is quite old and fwiw I went down quite a rabbit hole to figure out *why* this was an issue, but couldn't figure it out -- I could reproduce the error referred to in #2332, though separately checking the file system while paused at `docker-py`s request to the docker server for the build (aka the point of error) showed that the file did exist so 👻 

On top of that weirdness, when adding this fix in to use the absolute path for docker storage when on MacOS, it causes a similar error at a different place, thus I went for the conditional behind the OS type. The build step that fails when using the abspath on Macos:

```
Step 7/10 : COPY /Users/laura/Desktop/tmpmgz7jdvf/demo.flow /root/.prefect/flows/demo.prefect
COPY failed: stat /var/lib/docker/tmp/docker-builder912549735/Users/laura/Desktop/tmpmgz7jdvf/demo.flow: no such file or directory
```

## Why is this PR important?
So Docker storage can be built by users registering flows from windows machines.


cicdw:
Classic long-debug-session-resulting-in-2-lines-of-code situation; changelog + black and this LGTM

src/prefect/environments/storage/docker.py
```diff
@@ -307,6 +307,10 @@ def _build_image(self, push: bool = True) -> tuple:
             dir="." if self.dockerfile else None
         ) as tempdir:
 
+            if True:  # sys.platform == 'win32':
```
joshmeek, line 310:
```suggestion
            if sys.platform == 'win32':
```

CHANGELOG.md
```diff
@@ -20,7 +20,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Fixes
 
-- None
+- Fix Docker storage path issue when registering flows on Windows machines 0 [#2332](https://github.com/PrefectHQ/prefect/issues/2332)
```
joshmeek, line 23:
```suggestion
- Fix Docker storage path issue when registering flows on Windows machines - [#2332](https://github.com/PrefectHQ/prefect/issues/2332)
```

PR #2391, opened 2020-04-23
Results wiring

### LINKED ISSUE #2168: PIN 16: 0.11.0: Pipeline logic updates between new Result methods and result handlers in Result

cicdw:
This PR wires up the new results classes into our runner pipelines.  There are two groups of failing tests that still need to be resolved:

- [ ] tests involving cloud runs
- [ ] tests involving control flow tasks (this one is still a mystery to me)

I also took the opportunity to isolate _most_ result-handling weirdness to the Cloud runners (so you'll notice that `write` and `populate_result` are called extremely minimally / not at all outside of Cloud-based runs).  I hope this will make the code easier to reason about.

I also need to add some new tests involving:

- [ ] backwards compatibility of serializers
- [ ] ensure that states which are deserialized from old payloads can be used within the pipeline (this is a super rare edge case situation but I think it's worth considering within a timebox)

Despite the tests, this is ready to see some QA action / review.  Closes #2168 

### QA todos
- S3Result
    - [x] core only
    - [ ] server
- GCSResult
    - [ ] core only
    - [ ] server
- LocalResult
    - [x] core only
    - [ ] server
- AzureResult
    - [x] core only
    - [ ] server
- SecretResult
    - [x] core only
    - [ ] server
- PrefectResult
    - [x] core only
    - [ ] server
- CustomResult
    - [x] core only
    - [ ] server
- [x] locally in core with no result handlers/results configured
  - [ ] the same with server
- [ ] docker storage with no results configuration in server
- [ ] double check result handler deprecation warnings went back in


src/prefect/engine/task_runner.py
```diff
@@ -575,30 +579,71 @@ def get_task_inputs(
...
-                    k: r.to_result(handlers.get(k))  # type: ignore
+                    k: r
                     for k, r in state.cached_inputs.items()
                     if task_inputs.get(k, NoResult) == NoResult
                 }
             )
+
+        #        if any(NoResult == val for val in task_inputs.values()):
+        #            state = Failed(message="Some upstream inputs are missing.")
+        #            raise ENDRUN(state)
```
joshmeek, line 600: Should this be removed?

cicdw: yup, will do

src/prefect/engine/cloud/task_runner.py
```diff
@@ -243,6 +242,32 @@ def check_task_is_cached(self, state: State, inputs: Dict[str, Result]) -> State
 
         return state
 
+    def load_results(
```
lauralorenz, line 245: :+1:

src/prefect/engine/task_runner.py
```diff
@@ -575,30 +579,71 @@ def get_task_inputs(
...
+
+        Returns:
+            - State: the state of the task after running the check
+        """
+        result = self.result
+        target = self.task.target
+
+        if result and target:
+            if result.exists(target, **prefect.context):
+                cached_state = Cached(
```
lauralorenz, line 641: 😍

src/prefect/engine/task_runner.py
```diff
@@ -879,7 +919,7 @@ def get_task_run_state(
                         self.task.run, timeout=self.task.timeout, **raw_inputs
                     )
             else:
-                result = timeout_handler(
+                value = timeout_handler(
```
lauralorenz, line 927: ❤️

src/prefect/engine/task_runner.py
```diff
@@ -900,29 +940,28 @@ def get_task_run_state(
...
-        state = Success(
-            result=result, message="Task run succeeded.", cached_inputs=inputs
-        )
-
-        ## checkpoint tasks if a result_handler is present, except for when the user has opted out by disabling checkpointing
+        ## checkpoint tasks if a result interface is present, except for when the user has opted out by disabling checkpointing
         if (
-            state.is_successful()
-            and prefect.context.get("checkpointing") is True
+            prefect.context.get("checkpointing") is True
```
lauralorenz, line 957: 👀

PR #2398, opened 2020-04-23
Add agent registration step

### LINKED ISSUE #2312: Prepare Agents for "registering" themselves with Cloud
Whenever a user spins up an agent against the `"cloud"` backend, we will soon be registering the agent as a first step.  At registration time:

- we will call an agent registration route (name TBD)
- send a payload of information about this agent (type, name, labels, etc.)
- receives back an ID which is stored on the agent instance
- sends this ID within a header for every future query

We should hook up the lifecycle of this into our base agent class so we're ready once the backend is ready.

joshmeek:

## What does this PR change?
Closes #2312 

## Why is this PR important?
Agents can now register themselves for a better management experience.


src/prefect/client/client.py
```diff
@@ -315,6 +319,12 @@ def _request(
 
         return response
 
+    def attach_headers(self, headers: dict) -> None:
+        """
+        Set headers to be attached to this Client
+        """
+        self._attached_headers = headers  # type: ignore
```
cicdw, line 326: In case we want to use this more in the future we might want to store all attached headers as an attribute and update them with each call to this method (so that multiple calls don't wipe old headers but rather add to them / update them)

src/prefect/agent/agent.py
```diff
@@ -126,6 +129,14 @@ def _verify_token(self, token: str) -> None:
         ):
             raise AuthorizationError("Provided token does not have a RUNNER scope.")
 
+    def _register_agent(self) -> str:
+        """
+        Register this agent with Prefect Cloud and store registration ID on agent
+        """
+        # registration_id = self.client.register_agent(type=type(self), name=self.name, labels=self.labels)
+        registration_id = "ID"
```
cicdw, line 137: something looks fishy here

joshmeek:
@zdhughes How is uniqueness enforced? I tested this with LocalAgents using combinations of names/labels and I kept getting the same ID every time. Also there was no way to delete the Agent once registered.

joshmeek:
Update: all good from the Cloud side of things 👍

PR #2406, opened 2020-04-24
Add optional HTTP health check to all agents

### LINKED ISSUE #2313: Local agent health check
## Use Case

The local agent parent process should support external healthchecks

We are trying to use prefect as our task manager, where each compute resource has a docker-compose-based service with a prefect local agent in it, and do standard  docker / docker-compose administration over it. While we can monitor the health of the docker process itself, there is no clear way for docker to ask the prefect local agent if it is ok, which prevents 2000-era monitoring, auto-restart, etc.

## Solution

* the prefect local agent exposes an HTTP status check route
* the status check route does whatever consistency checks make sense, even if just a quick "hello!", and returns a 200 status if OK. (Unclear what those are, we're ok with "i'm up" for starters.)
* fancy: config vars to control this further

That is enough that `supervisord`, `docker`, etc. healthcheck frameworks can hook in

## Alternatives

* Prefect handles all observability & reliability operations internally, including executor monitoring & restart stack, in some unexposed + non-standard way
* Prefect does not support health checks

jcrist:
This adds an optional HTTP server to all agents (default is no server).
Currently this server only serves one route - `/api/health` - which can
be used by orchestration systems to do health checks. If the agent is
running it's assumed to be healthy, which should be a good default for
all systems (e.g. if the k8s api server goes down briefly for
maintenance, the agent pod shouldn't be marked as unhealthy because of
it). Those kind of health checks should be handled by communication with
cloud/server, and possibly reducing/stopping pulling new flows while the
backing infrastructure is unavailable.

Fixes #2313.


jcrist:
Before I keep pushing on this, I want to make sure the behavior I'm intending is what's desired.

- By default no HTTP server is started
- If the user specifies an address, the server is run at that address. Since this should only be used for health checks (for now), https configuration should be unnecessary (should only run at localhost, only serves a simple route). No auth should be needed either.
- I named the config option `api_address`, since we could add other routes to this server at some point in the future (maybe metrics if local administrators wanted to track things locally?). `api_address` is a bit overloaded though, since Prefect's cloud/server also has an api address the user can configure. Happy to change this to a different term.

src/prefect/agent/agent.py
```diff
@@ -80,6 +91,7 @@ def __init__(
         labels: Iterable[str] = None,
         env_vars: dict = None,
         max_polls: int = None,
+        api_address: str = None,
```
joshmeek, line 94: Yeah this could be a bit confusing and possible to be mistaken with the actual cloud/server API address. Maybe more along the lines of `agent_address` or something

src/prefect/agent/agent.py
```diff
@@ -187,7 +205,44 @@ def start(self) -> None:
                             )
                         )
         finally:
-            self.on_shutdown()
+            self.cleanup()
+
+    def setup(self) -> None:
+        self.agent_connect()
+
+        if self.api_address:
```
joshmeek, line 213: Jw, is there a reason _not_ to have the API enabled by default?

jcrist: I think we want users to opt into this, since it's only useful for deployments. It also forces them to choose a port, which is good since any default port might be used by something else already.

jcrist:
Thanks @joshmeek, this should be ready for review.

jcrist:
I don't get mypy. I think in this case it may be the lack of `tornado` in the mypy test environment, so the type annotations aren't being picked up. None of the prefect dependencies are installed for the `mypy` tests though, so I'm not sure how the tests were passing before? Advice here would be appreciated.

cicdw:
Hmmm these type failures don't occur in mypy version `0.740` but do occur in `0.770` which is surprising.  I managed to get them to pass locally by doing the following:

- [update disallow_untyped_calls to `False`](https://github.com/PrefectHQ/prefect/blob/master/setup.cfg#L19)
- change `type: Optional[web.HTTPServer]` to `type: ignore`
- add `assert self._api_server_loop is not None` to your `run()` function as a mypy type assert

None of which were interesting / useful changes to the code.

Every now and then `mypy` catches interesting static bugs but I constantly wonder if we should stop enforcing type checks as a CI job.  This instance + my experience on PIN 16 have me thinking maybe we should move away from it...

jcrist:
The `mypy` tests here actually run twice:

- Once in the `Check code style and docs` build. This is the test that fails.
- Once as part of a test in the `Run tests` build. This test passes fine.

Note that the `Check code style and docs` build doesn't install prefect or its dependencies, so type annotations from dependent libraries may not be available? Running `mypy src` locally with version 0.770 passes fine for me.

> Every now and then mypy catches interesting static bugs but I constantly wonder if we should stop enforcing type checks as a CI job. This instance + my experience on PIN 16 have me thinking maybe we should move away from it...

Personally I don't like Python's type annotations. But I recognize that along with docs, they're a way others may want to help them use the library. Since we're currently shipping with type annotations for the public api, we may want to continue doing so, in which case we'll want to test somewhere that the annotations match the implementation. Is there a way to test only the public-facing method signatures are accurate (I'm not familiar enough with mypy to know).

jcrist:
> update mypy todisallow_untyped_calls = False in setup.cfg (I think this will get CI passing and start relaxing our typing requirements)

Done, but the tests still failed. I don't get mypy.

cicdw:
The original reason was just to have a clear CI job that signaled "static analysis failed" vs. it being buried in a unit test name.  I have no clue why tornado specifically is causing that job to fail, when we have other dependencies that `mypy` seems happy to gloss over?

I'd rather get this PR in than spend time on the typing checks, so if you want to remove [the circle mypy job](https://github.com/PrefectHQ/prefect/blob/master/.circleci/config.yml#L83-L85) I'm cool with it (since it's a duplicate test anyway).

jcrist:
> The original reason was just to have a clear CI job that signaled "static analysis failed" vs. it being buried in a unit test name. I have no clue why tornado specifically is causing that job to fail, when we have other dependencies that mypy seems happy to gloss over?

My guess is that most of the code interacting with dependencies is in files ignored in `setup.cfg` (for example, all of `prefect.tasks` is ignored)?

jcrist:
Uninstalled tornado locally and got the tests to pass. Felt a bit silly. I'll open an issue about mypy to handle separately.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 - Add `botocore_config` option to Fargate agent for setting botocore configuration when interacting with boto3 client - [#2170](https://github.com/PrefectHQ/prefect/issues/2170)
 - Don't create a `None` task for a null condition when using `ifelse` - [#2449](https://github.com/PrefectHQ/prefect/pull/2449)
+- Agents now support an optional HTTP health check, for use by their backing orchestration layer (e.g. k8s, docker, supervisord, ...) - [#2406](https://github.com/PrefectHQ/prefect/pull/2406)
```
joshmeek, line 15: New section of the changelog to move this to after the 0.10.6 release

jcrist: Done

docs/orchestration/agents/overview.md
```diff
@@ -92,3 +92,30 @@ $ export PREFECT__CLOUD__AGENT__LABELS='["dev", "staging"]'
 :::tip Environment Variable
 Setting labels through the `PREFECT__CLOUD__AGENT__LABELS` environment variable will make those labels the default unless overridden through initialization of an Agent class or through the CLI's `agent start` command.
 :::
+
+### Health Checks
+
+Agents can optionally run a private HTTP server for use as a health check.
+Health checks can be used by common orchestration services (e.g.
+``supervisord``, ``docker``, ``kubernetes``, ...) to check that the agent is
+running properly and take actions (suche as restarting the agent) if it's not.
```
joshmeek, line 101:
```suggestion
running properly and take actions (such as restarting the agent) if it's not.
```

lmeyerov:
Thanks @joshmeek  @jcrist ! I think we're installing via pip, so as that propagates, will notify our team! We're running via docker-compose so assuming we get `healthcheck: ...` working, will share

PR #2421, opened 2020-04-27
Allow EC2 Launch Type in addition to FARGATE

joeschmid:
This is a small, low-risk change that adds support for the `EC2` launch type in AWS ECS in addition to the existing `FARGATE` launch type. The default launch type of `FARGATE` remains unchanged. This applies to both the Fargate Agent and `FargateTaskEnvironment`.

## What does this PR change?
Previously, launch type for AWS ECS tasks was hardcoded to `FARGATE`. This is a simple change
that leaves the default set to `FARGATE`, but allows for launching tasks as `EC2`. This applies to both the Fargate Agent and `FargateTaskEnvironment`.

## Why is this PR important?
While AWS Fargate is a gaining popularity as a serverless way to run containers, AWS ECS allows for running containers on EC2 instances. The EC2 version of ECS has some advantages over Fargate, including a significantly faster task start time through the use of image caching.


joeschmid:
@braunreyes if you have a chance to take a quick look at this simple PR that'd be great. (Should be low risk, but since it's Fargate-related I wanted to give you a heads up.)

joshmeek:
@joeschmid Is there something in using this Client that auto provisions/un-provisions EC2 resources? i.e. if I use the EC2 type for these Tasks does it un-provision resources upon completion?

joeschmid:
> Is there something in using this launch type that auto provisions/un-provisions EC2 resources?

Hi @joshmeek  No, the lifecycle for ECS with EC2 instances is:

1. Provision EC2 instance(s) in whatever way you want (Terraform, console, api, etc.) independent of ECS, but register that EC2 instance to join a certain ECS cluster.
2. Once the instance(s) are up and registered with the ECS cluster, now you can request execution of containers through the regular ECS API with task definitions, run task, etc.

By using ECS with EC2, you're essentially opting in to doing your own lifecycle management of the EC2 instances, i.e. that model is NOT serverless. :) Hope that helps explain, but let me know if I can provide more info.

joshmeek:
@joeschmid Ah makes a lot of sense, thanks for clarifying!

joshmeek:
@joeschmid Do you think you could add a test to `test_fargate_agent` for this kwarg? Doesn't have to be anything simple but some simple unit tests w/ FARGATE/EC2/None would be great!

joeschmid:
> @joeschmid Do you think you could add a test to `test_fargate_agent` for this kwarg?

@joshmeek Took a quick crack at it, mostly taking a subset of `FargateTaskEnvironment` tests. See what you think.

PR #2360, opened 2020-04-20
Initial implementation of DaskCloudProviderEnvironment

joeschmid:
The [Dask Cloud Provider](https://cloudprovider.dask.org/) project aims to make it easy to bring up Dask clusters using services from cloud providers, e.g. AWS Fargate. This Prefect environment uses Dask Cloud Provider to dynamically create a Dask cluster, run the Flow on it, and tear down the cluster after use. This environment is meant to provide an easy path to Dask scalability for users of cloud platforms, like AWS.

- [x] adds new tests (if appropriate)
- [x] updates `CHANGELOG.md` (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

## What does this PR change?
This adds a new execution environment, `DaskCloudProviderEnvironment`.

## Why is this PR important?
Prefect offers excellent scalability by running Flows on Dask clusters. However, configuring a truly distributed Dask cluster with multiple workers on separate compute resources can be complicated. Kubernetes can be used for this, but Prefect users who aren't familiar it (and Dask) may find it complex to configure. Even for experienced users currently using Prefect and Dask on k8s, cloud services like Fargate may make integration with CI/CD pipelines and automated deployment of Flows easier, e.g. Docker image versions can be updated in Fargate task definitions after an automated build completes, etc.

PR #2425, opened 2020-04-27
Fix RemoteDaskEnvironment API documentation Security kwarg snippet

joshmeek:
cc @joeschmid in reference to what I said in a review comment on #2360

**Old:**
image

**New:**
image


joeschmid:
@joshmeek thanks for fixing this! Also, I should have moved the link on Dask TLS docs ("For more on using TLS with Dask see https://distributed.dask.org/en/latest/tls.html") out of the docstring for the security argument and into text above so that it's actually a link!

src/prefect/environments/execution/dask/remote.py
```diff
@@ -21,19 +21,23 @@ class RemoteDaskEnvironment(RemoteEnvironment):
     f = Flow("dummy flow", environment=env)
     ```
 
+    If using the Security object then it should be filled out like this:'
```
cicdw, line 24:
```suggestion
    If using the Security object then it should be filled out like this:
```

PR #2429, opened 2020-04-27
Resize logos in README

joshmeek:
GitHub does some weird resizing. Cutting size in half to fit


joshmeek:
No, they look fine now.

PR #2443, opened 2020-04-28
New conditional api

### LINKED ISSUE #2381: New conditional api
## Background

The current conditional api (`switch`, `ifelse`, and `merge`) work, but have a few problems. Using `ifelse` as an example:

```python
ifelse(cond : Task, true_task : Task, false_task : Task) -> None
```

- If `cond` evaluates to `True`, `true_task` is run and `false_task` is skipped. However, tasks upstream to `false_task` aren't skipped. For example:

  ```python
  ifelse(
      some_condition(),
      load(transform(extract())),
      some_other_task(),
  )
  ```

  if `some_condition` evaluates to `False`, only the `load` task is skipped, `transform` and `extract` still execute.

- If the result of a task is conditional based on upstream tasks, you'll need to call `merge` explicitly. We tried making `ifelse`/`switch` return the merge implicitly, but had to revert due to issues with implicitly creating terminal tasks in the graph (#2310, #2379).

  ```python
  on_true = do_something()
  on_false = do_something_else()
  cond = condition()
  ifelse(cond, on_true, on_false)
  result = merge(on_true, on_false)
  ```

  This approach works, but can be tricky to read.

## Proposal

I propose we add a contextmanager api for expressing conditional flows. The existing `switch`/`ifelse`/`merge` api will still exist (and may be used internally by the contextmanager api).

The api consists of two new functions:

- `case`: used to express branches in logic
- `var`: used to express merging outputs (this could equally be a class `Variable`)

A few examples to illustrate use:

### Example 1

Here we use `case` to express an `ifelse` branch:

```python
with Flow("example-1") as flow:
    cond = some_condition()
    
    with case(cond, True):
        a = do_a()
        b = do_b()
        do_c(a, b)

    with case(cond, False):
        do_c(1, 2)
```

If `some_condition` is `True`, `do_a`, `do_b`, and `do_c` are run. If it's false, `do_c` is run with some constant inputs. For each `case` block, a `Condition` task is created. If a task is created inside the `case` block *and* has no upstream dependencies that were also created inside that `case` block, then the condition is set as an upstream dependency. In this case, `a` and `b` would have a `Condition` for `True` set as an upstream dependency, but `do_c` would only depend on `a` and `b`. We could equally have all tasks created inside the block have the condition as an immediate upstream task, but that makes the graph visually complicated.

### Example 2

Here we illustrate merging tasks with `var`. Lets say we wanted to capture the output of `do_c` and use it later on. In normal Python code we might write this as:

```python
def example_2():
    if some_condition():
        a = do_a()
        b = do_b()
        c = do_c(a, b)
    else:
        c = do_c(1, 2)

    do_d(c)
```

Using this proposal, an equivalent flow would be:

```python
with Flow("example-2") as flow:
    cond = some_condition()
    
    c = var("c")

    with case(cond, True):
        a = do_a()
        b = do_b()
        c.set(do_c(a, b))

    with case(cond, False):
        c.set(do_c(1, 2))

    do_d(c)
```

Each call to `c.set` binds the value of `c` inside its corresponding `case` block. Behind the scenes this builds up a merge task branch-by-branch (note: we may not use merge directly, but the general structure will be the same). Since `c` is a `Task` itself, it can then be fed to downstream tasks.

### Example 3

These context managers should work fine even with nested conditional logic. For example, the following function:

```python
def example_3():
    if some_condition():
        if other_condition():
            a = do_a()
        else:
            a = 1
        b = do_b()
        do_c(a, b)
    else:
        d = do_d()
        do_e(d)
```

corresponds to an equivalent flow of:

```python
with Flow() as flow:
    cond = some_condition()

    with case(cond, True):

        cond2 = other_condition()

        a = var("a")
        with case(cond2, True):
            a.set(do_a())
        with case(cond2, False):
            a.set(1)

        b = do_b()
        do_c(a, b)

    with case(cond, False):
        d = do_d()
        do_e(d)
```

Note that in this case, the `Condition` task created by calling `case` on `cond2` will have the `Condition` created for `cond1` as a direct upstream task (following the logic described above, since it has no upstream tasks created inside this `case` context, the corresponding condition is set as a direct upstream task).

---

I think implementing the api as described above is fairly doable. It *might* require #2298 though to properly handle complex branching/merging logic (not sure yet).

jcrist:
This adds two new control flow constructs:

- `case`: for creating conditional branches in a flow
- `Variable`: for merging conditional branches back together

These do similar things to `switch` and `merge`, but should hopefully be
easier to read.

Still needs tests and docs, just getting a WIP up for comments.

Fixes #2381


jcrist:
And an example:

```python
from prefect.tasks.control_flow.case import case, Variable

@task
def some_condition():
    return True


@task
def other_condition():
    return True


@task
def do_a():
    print("Doing a")
    return 1


@task
def do_b():
    print("Doing b")
    return 2


@task
def do_c(a, b):
    print("Doing c")
    return a + b


@task
def do_d():
    print("Doing d")
    return 1


@task
def do_e(d):
    print("Doing e")
    return d + 1


with Flow("example-3") as flow:
    cond = some_condition()

    with case(cond, True):

        cond2 = other_condition()

        a = Variable(name="a")

        with case(cond2, True):
            a.set(do_a())

        with case(cond2, False):
            a.set(do_c(1, 2))

        b = do_b()
        do_c(a, b)

    with case(cond, False):
        d = do_d()
        do_e(d)
```

Results in the following graph:

test

jcrist:
See #2381 for the description of the semantics.

**A few naming questions:**

- The user should never interact with a `case` object directly, but `case` is a class for ease of implementation (it doesn't need to be). We could pull the class into an internal type and make `case` a `@contextmanager` wrapped function just as easily.
- Is `Variable` the right name? Other options with the same connotation: `Var`, `var`, `variable`.
- Should `Variable`'s require a name? Should they be named `Variable("{name"}` instead of just `"{name}"`?
- Is `Variable.set` the right name for this method?
- Currently we return the *last* non-`None` result we get. This breaks down if the upstream task meant to return `None`. This is also the opposite of `merge`, which returns the first. I think using the last makes more sense, since it will use the last successful write of a variable.

**And one semantic question:**

Mixing `Variable.set` calls and uses can be a bit confusing. For example:

```python
@task
def inc(x):
    return x + 1

v = Variable(name="a")
a.set(1)
b = inc(a)
a.set(2)
```

In this case `inc` gets the value `2` instead of `1`, even though `a` was set to `2` *after* the call to `inc`. I see a few options here:

- Leave this footgun, but document it.
- With a small bit of special casing instead `Task` we can transform the above to do the right thing automatically (on each use of a `Variable` a copy is made before adding it to the flow, so future `.set` calls don't mutate the previous call).
- Drop `Variable`, rely on `merge` as before. I can't decide if I find this easier or harder to read.

```python
with Flow("example-3") as flow:
    cond = some_condition()

    with case(cond, True):

        cond2 = other_condition()

        with case(cond2, True):
            a1 = do_a()

        with case(cond2, False):
            a2 = do_c(1, 2)

        a = merge(a1, a2)

        b = do_b()
        do_c(a, b)

    with case(cond, False):
        d = do_d()
        do_e(d)
```

jcrist:
> Mixing Variable.set calls and uses can be a bit confusing. For example

Thinking about this a bit more, I think I'm leaning slightly towards dropping `Variable` entirely in favor of `merge`. In the simple idiomatic cases we can simulate python variable assignment, but I'm not convinced it's more readable and worth the bother.

jlowin:
@jcrist i really love this syntax - I think it removes significant complexity, especially for users who just want to gate a small number of conditional tasks. 

My opinion is to drop the Variable entirely; on the one hand I’m super impressed you solved the problem of working with a deferred multi-assigned variable, but I worry the nuance of applying variables correctly will be overshadowed by the additional burden of explaining them, why they differ from parameters, etc., especially because they add a lot of machinery to replace a relatively small merge op. 

Therefore I think a great example in our documentation of how to merge two conditional branches will provide a good solution with lower maintenance overhead.

jcrist:
Ok, I've updated to remove `Variable`, and added some tests and docs. A few api decisions left:

- Currently `case` is a class, but syntactically I like the lowercase `case` better than the uppercase `Case`. People may be confused by this, idk.
- Currently there is no return value from `__enter__`, meaning that in:

  ```python
  with case(task, value) as temp:
      pass
  ```

  ``temp`` is ``None``. I think this makes sense, I don't expect users to ever need to explicitly reference a `case` object. A few alternatives:

  - Return the `case` object
  - Return the conditional check task (the `CompareValue`) task that's created by the `case` call

- Currently we reuse the `CompareValue` tasks used by `ifelse` and `switch`. Alternatively we could create a new `Case` class (basically identical to `CompareValue`, except in name), which (IMO) may make the task structure clearer (or we could rename `CompareValue` to `Case` - I find the name `Case` a bit clearer for what it's doing). At the same time, it may be confusing to users to have both `case` (the context manager) and `Case` (the conditional task).

joshmeek:
@jcrist here's my 2¢ on your Qs:

* Could you do something where the class is `Case` and then there's an importable `case` which in the original file is something like `case = Case`. Not sure if that's bad practice or not
* I can't think of a time where `case` or `CompareValue` would need to be referenced again BUT I'm not opposed to having `__enter__` return the original case
* Personally I'm used to `CompareValue` so it makes sense to keep it. Having two cases for the context manager and conditional task is a bit confusing

cicdw:
Awesome this is great; to your points:

- I think keeping `case` a lowercased class is fine
- I don't have a strong opinion on the behavior of `case.__enter__` but _if_ you wanted to return something other than `None` my vote would be for the `CompareValue` task created under the hood, because otherwise that task is harder to find and access
- My bias is to keep `CompareValue` and not introduce new control flow tasks / renames

jcrist:
Ok, given the above comments I, I think I'm going to keep things as they are. I've added a changelog entry, I believe this is now ready for merge.

PR #2449, opened 2020-04-29
Dont spuriously create None task when using ifelse

cicdw:

## What does this PR change?
This PR prevents the creation of a task whenever the `false_task` is `None`.  This is a minor optimization that allows for a cleaner DAG when the intent is just using a basic "if".

## Why is this PR important?
Minor UX improvement.

**cc:** @jcrist / @znicholasbrown

### DIFF
--- a/src/prefect/tasks/control_flow/conditional.py
+++ b/src/prefect/tasks/control_flow/conditional.py
@@ -114,7 +114,10 @@ def ifelse(condition: Task, true_task: Task, false_task: Task) -> None:
     def as_bool(x):
         return bool(x)
 
-    switch(condition=as_bool(condition), cases={True: true_task, False: false_task})
+    cases = {True: true_task}
+    if false_task is not None:
+        cases.update({False: false_task})
+    switch(condition=as_bool(condition), cases=cases)
 
 
 def merge(*tasks: Task) -> Task:
--- a/src/prefect/tasks/control_flow/conditional.py
+++ b/src/prefect/tasks/control_flow/conditional.py
@@ -114,10 +114,9 @@ def ifelse(condition: Task, true_task: Task, false_task: Task) -> None:
     def as_bool(x):
         return bool(x)
 
-    cases = {True: true_task}
-    if false_task is not None:
-        cases.update({False: false_task})
-    switch(condition=as_bool(condition), cases=cases)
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/control_flow/conditional.py
```diff
@@ -114,7 +114,10 @@ def ifelse(condition: Task, true_task: Task, false_task: Task) -> None:
     def as_bool(x):
         return bool(x)
 
-    switch(condition=as_bool(condition), cases={True: true_task, False: false_task})
+    cases = {True: true_task}
+    if false_task is not None:
+        cases.update({False: false_task})
+    switch(condition=as_bool(condition), cases=cases)
```
jcrist, line 120: I could see the opposite use case (only a `False` task provided) as well. Perhaps:

```suggestion
    cases = {c: t for c, t in [(True, true_task), (False, false_task)] if t is not None}
    if cases:
        switch(condition=as_bool(condition), cases=cases)
```

jcrist: With this change, should `true_task` and `false_task` be made optional parameters to `ifelse`, defaulting to `None`?

cicdw: Hmm that's a good question; I don't have a strong opinion on the defaults -- I think there's a minor benefit to making users pass the args each time but it's not a big deal either way

jcrist: I think it'd be a bit nicer to have them as optional parameters, otherwise it's not necessarily clear how to avoid creating the branch. If we do go with keyword arguments, I'm not sure I like the existing arg names. Perhaps `on_true`/`on_false` or `if_true`/`if_false` would be nicer?

```python
ifelse(
    cond(),
    if_true=some_task(),
    if_false=other_task(),
)
```

No strong thoughts though, feel free to ignore any signature changes here. The existing PR looks good to me.

znicholasbrown: I really like the kwarg idea @jcrist proposes, but would suggest slightly different kwargs, `on_true` and `on_false`... no strong feelings though and this could be just the Javascript dev in me (those look a lot like vanilla event handlers :^))

PR #2450, opened 2020-04-30
fix start_server.sh when PREFECT_SERVER__GRAPHQL_URL undefined

gregorg:
A small fix that handle undefined PREFECT_SERVER__GRAPHQL_URL env var.


joshmeek:
Thanks for the fix @gregorg! Would you mind adding this to the _Fixes_ section of Prefect's main `CHANGELOG.md` and add yourself as a contributor?

PR #2452, opened 2020-04-30
Ensure that ordered collections maintain numerical order for > 10 items

jlowin:

## What does this PR change?
This PR closes https://github.com/PrefectHQ/prefect/issues/2451, which identified a bug in which ordered Prefect collections implicitly relied on string (lexical) ordering rather than numerical ordering. Since collections maintain a dictionary keyed by `arg_{n}`, where n is the sort order, lexical ordering would result in `arg_0, arg_1, arg_10, arg_11, arg_2, arg_3, ...` for a list of 12 items. This PR adds an explicit key function to the sort of `collections.List` and `collections.Tuple` which extracts the integer part of the key name for numerical sorting.


jlowin:
The situation in which this issue arises is when a Prefect Dict task relies on keys that are a Prefect Constant and values that are a Prefect Task. Due to the recent Constant refactor, the keys signature effectively went from `List[Constant]` to `Constant[List]`, meaning it retains the original sort order; the values are using lexical sort as described here. This mismatch of sorting causes the bug.

PR #2455, opened 2020-04-30
Prefect Idiom documentation section

joshmeek:
First cut at a new section of the docs and welcome any and all feedback. @cicdw these are the first few we agreed upon and then I figured from here on out we could add them incrementally, a little at a time. I'm not sure if these have too much or too little detail. I also made it so that each flow/task snippet has both functional and imperative examples.


docs/core/faq/questions.md
```diff
@@ -0,0 +1,9 @@
+# FAQ (Name TBD)
+
+- [How do I make branches in my flow?](branching.html)
+- [How do I map over specific arguments?](mapping.html)
+- [How do I run my tasks in parallel?](parallel.html)
+- [How do I handle sensitive information in my flows?](secrets.html)
+- [How do I configure and use the Prefect logger?](logging.html)
+- [How do I inspect the results from my tasks?](task-results.html)
+- [How do I test my flows?](testing-flows.html)
```
joshmeek, line 9: Needs some better formatting / organization here

joeschmid:
This is awesome! Will be super helpful for so many Prefect users.

docs/core/faq/branching.md
```diff
@@ -0,0 +1,174 @@
...
+output_2.set_upstream(d, key="value", flow=flow)
+```
+:::
+::::
+
+### Branching based on conditional logic
+
+Prefect ships with various [tasks for control logic](/core/task_library/control_flow.html) that you can use off-the-shelf to control the branching structure of your flow. This is especially useful in cases where under certain conditions you want to control which branches of your flow run based on upstream events.
+
+Let's say you want to have a flow where some conditional task is evaluated and then the downstream branches are run depending on the output of that conditional task—true or false. An example of this might look like the following:
```
cicdw, line 89:
```suggestion
Let's say you want to have a flow where some conditional task is evaluated and then the downstream branches are run depending on the output of that conditional task. An example of this might look like the following:
```

docs/core/faq/branching.md
```diff
@@ -0,0 +1,174 @@
...
+
+![Flow with Conditional Branching](/faq/conditional_branch.png)
+
+If you choose to use some of Prefect's default control flow tasks such as `if_else` and `merge` then the flow will take on a DAG with some additional checks added:
```
cicdw, line 93:
```suggestion
If you choose to use some of Prefect's default control flow tasks such as `if_else` and `merge` then Prefect will automatically add some additional tasks to your Flow:
```

docs/core/faq/logging.md
```diff
@@ -0,0 +1,82 @@
...
+```python
+class MyTask(prefect.Task):
+    def run(self):
+        self.logger.info("An info message.")
+        self.logger.warning("A warning message.")
+```
+
+### Logging stdout
+
+Tasks also provide an optional argument to toggle the logging of stdout—`logstdout`. This means that, if enabled, anytime you send output to stdout (such as `print()`) it will be logged using the Prefect logger. This option is disabled by default.
```
cicdw, line 53:
```suggestion
Tasks also provide an optional argument to toggle the logging of stdout—`log_stdout`. This means that, if enabled, anytime you send output to stdout (such as `print()`) it will be logged using the Prefect logger. This option is disabled by default.
```

cicdw: We might want to mention _why_ it's disabled by default; in particular this option is not suitable for certain types of tasks (ref: https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout)

docs/core/faq/logging.md
```diff
@@ -0,0 +1,82 @@
+# How do I configure and use the Prefect logger?
```
cicdw, line 1: I think we should add a section here about logging with a backend - for example, we hardcode the log level to DEBUG because we want to provide the maximal amount of information if something goes wrong, and DEBUG level logs can always be filtered out in the UI / API

docs/core/idioms/branching.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Branching based on conditional logic
+
+Prefect ships with various [tasks for control logic](/core/task_library/control_flow.html) that you can use off-the-shelf to control the branching structure of your flow. This is especially useful in cases where under certain conditions you want to control which branches of your flow run based on upstream events. Branches that you don't want to run should enter a `Skipped` state.
+
+Let's say you want to have a flow where some conditional task is evaluated and then the downstream branches are run depending on the output of that conditional task. An example of this might look like the following:
+
+![Flow with Conditional Branching](/faq/conditional_branch.png)
+
+If you choose to use some of Prefect's default control flow tasks such as `if_else` and `merge` then Prefect will automatically add some additional tasks to your flow:
```
cicdw, line 93: This function is actually called `ifelse` (no underscore)

docs/core/idioms/mapping.md
```diff
@@ -0,0 +1,62 @@
+# Use task mapping to map over a specific set of arguments
+
+[Mapping](/core/concepts/mapping.html) in Prefect is a great way for dynamic task creation and parallel execution. Sometimes you may want to pass extra arguments to functions that you're using to map over inputs but you don't want to map over those extra arguments. This is where Prefect's [`unmapped`](/api/latest/utilities/tasks.html#unmapped) operator can come into play.
+
+In this example let's take a flow where we want to pass in a [Parameter](/core/concepts/parameters.html) `multiple` to be multiplied against a random list of numbers also specified with a `total` Parameter. This random list of numbers will be generated at runtime. Keep in mind that Prefect supports complete DAG dynamic generation with the option of tasks being able to be generated at runtime instead at the time of flow definition.
```
cicdw, line 5: The "Keep in mind" sentence is a little confusing to parse

docs/core/idioms/mapping.md
```diff
@@ -0,0 +1,62 @@
...
+flow = Flow("unmapped-values")
+
+total = Parameter("total")
+multiple = Parameter("multiple")
+
+numbers = GetNumbers()
+numbers.set_upstream(total, key="total", flow=flow)
+
+output_value = OutputValue()
+output_value.bind(mapped=True, n=numbers, multiple=unmapped(multiple), flow=flow)
```
cicdw, line 59: Open to push back here, but I feel like this is more idiomatic / clear:
```python
output_value.set_upstream(numbers, key="n", mapped=True, flow=flow)
output_value.set_upstream(multiple, key="multiple", mapped=False, flow=flow)
```

docs/core/idioms/parallel.md
```diff
@@ -0,0 +1,95 @@
+# Parallelism in Prefect
```
cicdw, line 1: I think we want to expand this idiom into two subsections: "Parallelism within a Flow" and "Parallel Flow execution".  Your write up covers "Parallelism within a Flow" but doesn't address the agent model that allows for parallel flow execution

joshmeek: For now let's set this to cover parallelism within a flow. And expand on the sections that have to do with orchestration separately

docs/core/idioms/secrets.md
```diff
@@ -0,0 +1,102 @@
+# Authentication and Secrets in Prefect
```
cicdw, line 1: Let's not expose this one yet; we can keep the text in the repo but whenever 0.11.0 goes out this will change slightly

docs/core/idioms/logging.md
```diff
@@ -0,0 +1,86 @@
+# Logging configuration and usage
+
+Prefect natively ships with a default logger configured to handle the management of logs. More information on logging in Prefect can be found in the [logging concept document](/core/concepts/logging.html).
+
+### Configure the logger
+
+The default Prefect logger can be configured by adjusting the settings found in [Prefect's config](/core/concepts/configuration.html).
+
+```toml
+[logging]
```
cicdw, line 10: We just introduced a new config for logging `log_attributes` that we should mention here; it can be used to include context info in your log formats

docs/core/idioms/logging.md
```diff
@@ -0,0 +1,86 @@
...
+
+my_task = MyTask(log_stdout=True)
+flow.add_task(my_task)
+```
+:::
+::::
+
+### Logging with a backend
+
+If you are deploying your flows with the use of a backend such as Prefect Core's server or Prefect Cloud then there are a couple subtle configuration changes that you might happening on each run. First, the configuration option `log_to_cloud` will be set to `True`. This means that the logs from your flow will be sent to the backend where they can be digested and managed. Second, the logging level will always be set to `DEBUG`. This is to ensure that the maximal amount of information is provided in case something goes wrong. The debug logs can always be filtered out after via the UI or API.
```
cicdw, line 86: We should add a blurb about how to turn off backend logging (by using the `--no-cloud-logs` CLI flag)

docs/core/idioms/testing-flows.md
```diff
@@ -0,0 +1,96 @@
...
+assert Edge(upstream_task=e, downstream_task=t, key="x") in flow.edges
+assert Edge(upstream_task=t, downstream_task=l, key="x") in flow.edges
+```
+
+#### Testing state
+
+```python
+state = flow.run()
+
+assert state.is_successful
```
cicdw, line 82: These should all be method calls `()`

docs/.vuepress/config.js
```diff
@@ -268,13 +268,7 @@ module.exports = {
...
           ]
         },
-        {
-          title: 'Prefect Idioms',
-          collapsable: true,
-          children: [
-            'idioms/idioms',
-          ]
-        },
+        '/core/idioms/idioms',
```
cicdw, line 271: Sorry for all the small reviews, but I think the location of this doesn't look very good.  Could we move this to right above "FAQ" for now?

PR #2458, opened 2020-04-30
Add task target kwarg and add check to task runner pipeline

joshmeek:

## Why is this PR important?

**Note**:
Being merged into the `results-wiring` branch. Not fully done with tests but want some feedback on this check. (the couple target tests so far pass but there may be others that will fail due to the base branch)
- is the check in the correct place of the runner pipeline
- how should we handle a target being specified but no result (right now it just returns the original state)
- if we're putting it in a cached state how should we handle the cache duration aspect

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -267,6 +267,13 @@ class Task(metaclass=SignatureValidator):
 
         self.target = target
 
+        if result and target:
+            warnings.warn(
+                "Both `result` and `target` set on task. Task result will use target as location."
+            )
+            self.result = result.copy()
+            self.result.location = target
+
         if state_handlers and not isinstance(state_handlers, collections.Sequence):
             raise TypeError("state_handlers should be iterable.")
         self.state_handlers = state_handlers or []
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -149,6 +149,8 @@ class Task(metaclass=SignatureValidator):
             retrieving and storing state results during execution; if not provided, will default to the
             one attached to the Flow
         - result (Result, optional): the result instance used to retrieve and store task results during execution
+        - target (str, optional): location to check for task Result. If a result exists at that location then
+            the task run will enter a cached state.
         - state_handlers (Iterable[Callable], optional): A list of state change handlers
             that will be called whenever the task changes state, providing an
             opportunity to inspect or modify the new state. The handler
@@ -189,6 +191,7 @@ class Task(metaclass=SignatureValidator):
         on_failure: Callable = None,
         log_stdout: bool = False,
         result: "Result" = None,
+        target: str = None,
     ):
         self.name = name or type(self).__name__
         self.slug = slug or str(uuid.uuid4())
@@ -261,6 +264,8 @@ class Task(metaclass=SignatureValidator):
             )  # type: Optional[Result]
         else:
             self.result = result
+
+        self.target = target
 
         if state_handlers and not isinstance(state_handlers, collections.Sequence):
             raise TypeError("state_handlers should be iterable.")
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -269,7 +269,7 @@ class Task(metaclass=SignatureValidator):
 
         if result.location and target:
             warnings.warn(
-                "Both `result` and `target` set on task. Task result will use target as location."
+                "Both `result.location` and `target` set on task. Task result will use target as location."
             )
             self.result = result.copy()
             self.result.location = target
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -267,7 +267,7 @@ class Task(metaclass=SignatureValidator):
 
         self.target = target
 
-        if hasattr(result, "location") and target:
+        if getattr(result, "location", None) and target:
             warnings.warn(
                 "Both `result.location` and `target` set on task. Task result will use target as location."
             )
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -267,7 +267,7 @@ class Task(metaclass=SignatureValidator):
 
         self.target = target
 
-        if result and target:
+        if result.location and target:
             warnings.warn(
                 "Both `result` and `target` set on task. Task result will use target as location."
             )
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -267,11 +267,11 @@ class Task(metaclass=SignatureValidator):
 
         self.target = target
 
-        if result.location and target:
+        if hasattr(result, "location") and target:
             warnings.warn(
                 "Both `result.location` and `target` set on task. Task result will use target as location."
             )
-            self.result = result.copy()
+            self.result = result.copy()  # type: ignore
             self.result.location = target
 
         if state_handlers and not isinstance(state_handlers, collections.Sequence):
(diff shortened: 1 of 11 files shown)


joshmeek:
Just a bump on this @lauralorenz @cicdw now that the merge has fixed tests 🙌

src/prefect/core/task.py
```diff
@@ -262,6 +265,8 @@ def __init__(
         else:
             self.result = result
 
+        self.target = target
```
cicdw, line 268: We should consider how we intend this to interact with the `result` kwarg and in paritcular `result.location`:
- should we require only one be provided?
- should `target` instead accept a `Result` instance, and then we raise an error if both `result` and `target` are provided?
- will `target` become the `location` of `task.result` ?
- should `target` be a boolean and then we use `result.location` as the filepath?

src/prefect/engine/task_runner.py
```diff
@@ -613,6 +616,39 @@ def load_results(
...
+                to the task's `run()` arguments.
+
+        Returns:
+            - State: the state of the task after running the check
+        """
+        result = self.task.result
+        target = self.task.target
+
+        if result and target:
+            if result.exists(target):
```
cicdw, line 635: I think we need to update the signature of `result.exists` to accept `**kwargs` (just like `result.format`) which are used to optionally template the `target` location.  That way targets like `"data/{task_name}/{today}"` work

src/prefect/engine/task_runner.py
```diff
@@ -613,6 +616,39 @@ def load_results(
...
+                # expiration = pendulum.now("utc") + self.task.cache_for
+                # TODO: How to handle cached durations?
```
cicdw, line 637: Should we introduce the ability to allow for infinite horizon caches?

joshmeek: Yeah we should for this because it doesn't necessarily have a _duration_ but the duration is until the target does not exist

src/prefect/engine/task_runner.py
joshmeek: For a target of `"data/{task_name}/{today}"` are `task_name` and `today` default values?

cicdw: Yup both of those are values we place in context

src/prefect/core/task.py
joshmeek: - I personally think that if a target is provided then result is also required
- No I like keeping target a simple string and separate
- It definitely should
- Interesting idea 🤔 if we did that then why even have target and not a kwarg like `cache_if_result_exists`?

src/prefect/engine/task_runner.py
cicdw: @joshmeek Whenever you create the `Cached` state, if you set `cached_result_expiration=None` then the `duration_only` validator will always return `True`, meaning the cache will effectively last for forever (until your pipeline step says otherwise at least).

src/prefect/core/task.py
cicdw: Yea I think there should be only one obvious way of configuring this, so the fact that `Result` exposes a `location` kwarg makes me think `target` should either be a boolean or a `Result` object.  I think having `target` be a full `Result` object is clearest (with an error if `result` is also provided) but that's just me.  **cc:** @lauralorenz for input here

lauralorenz: Soo I think that target should be a string and result should be separately required to configure everything but location. I like this independently of anything else because it’s basically syntactic sugar and brings us closest to the Luigi/Make simplicity. Everything else kind of falls out of that for me. 

Tbh thinking about it out of nowhere don’t hate it if a Result wrote to its location AND it’s target if both were configured, and the latter is used for the exists pipeline, if we were feeling greedy about writes. But our Results aren’t designed to hold 2 paths. Then I this this becomes about how we have to decide which to choose if both are configured. We could raise an error but that is probably not convenient until the flow is running, so then instead I end up voting for overwriting location with target. Which I think in conclusion means I agree with @joshmeek 🤓

I imagine it will be possible in this set up that someone reuses the same Result object for each of their tasks and vary the target Leary as an easy syntax to split out location for multiple tasks too, especially if we put cache time semantics on targets like our original plan.

cicdw: @lauralorenz yea the pattern you mention in your last paragraph could easily be achieved by doing `with Flow("blah", result=S3Result(bucket="default")):` and then only providing a target on a per-task basis.  

OK cool so the only unknown is how to handle both a provided location and a provided target.

lauralorenz: Yeah, I have a feeling it will annoy people if we throw exceptions but it will confuse people if we simply overwrite. The latter with a warning log is probably the safest in our philosophy of convenient defaults.

cicdw: OK cool: @joshmeek could you add a warning if `result.location is not None and target is not None` ?  In that case we should do something like 
```python
task_result = result.copy()
task_result.location = target
self.result = task_result
```
(store target as the new location but only for this task instance, we don't want to accidentally alter an object shared by other tasks)

joshmeek: @cicdw Would this happen in task.py when the task is initialized?

src/prefect/core/task.py
```diff
@@ -262,6 +265,15 @@ def __init__(
...
+
+        if result and target:
```
cicdw, line 270: we only need to warn if `result.location` is actually set, otherwise it's all good
```suggestion
        if result.location and target:
```

src/prefect/core/task.py
```diff
@@ -262,6 +265,15 @@ def __init__(
...
+            warnings.warn(
+                "Both `result` and `target` set on task. Task result will use target as location."
```
cicdw, line 272:
```suggestion
                "Both `result.location` and `target` set on task. Task result will use target as location."
```

src/prefect/engine/task_runner.py
```diff
@@ -613,6 +616,34 @@ def load_results(
...
+
+        Args:
+            - state (State): the current state of this task
+            - inputs (Dict[str, Result]): a dictionary of inputs whose keys correspond
+                to the task's `run()` arguments.
+
+        Returns:
+            - State: the state of the task after running the check
+        """
+        result = self.task.result
```
cicdw, line 631:
```suggestion
        result = self.result
```
The task runner stores a copy of the result object; this is important in situations where a user provides a result to the _flow_ object but not to individual tasks

src/prefect/engine/task_runner.py
```diff
@@ -613,6 +616,34 @@ def load_results(
...
+        target = self.task.target
+
+        if result and target:
+            if result.exists(target):
```
cicdw, line 635:
```suggestion
            if result.exists(target, **prefect.context):
```

src/prefect/engine/task_runner.py
```diff
@@ -613,6 +616,34 @@ def load_results(
...
+                cached_state = Cached(
+                    result=state._result,
+                    cached_inputs=inputs,
+                    cached_result_expiration=None,
+                    cached_parameters=prefect.context.get("parameters"),
+                    message=f"Result found at task target {target}",
```
cicdw, line 641: We can worry about this later, but because `target` can be a templated string we might want to adjust this message

src/prefect/core/task.py
```diff
@@ -267,11 +267,11 @@ def __init__(
 
         self.target = target
 
-        if result.location and target:
+        if hasattr(result, "location") and target:
```
cicdw, line 270: I think you mean `getattr(result, "location", None)`

joshmeek: Yup I did

PR #2463, opened 2020-05-01
Issue 2423 add bigquery load from file

dkapitan:

- [x] adds new tests for `gcp.bigquery.BigQueryLoadFile`

## What does this PR change?
Add new BigQueryLoadFile task

## Why is this PR important?
Better interface to GBQ.


joshmeek:
Thanks for the PR @dkapitan! Looks like you have some unused imports which are failing tests:

```
Potentially unused imports found
--- original/src/prefect/tasks/gcp/bigquery.py
+++ fixed/src/prefect/tasks/gcp/bigquery.py
@@ -1,4 +1,3 @@
-import os
 from pathlib import Path
 import warnings
 from typing import List, Union
```

and one for the new bigquery tests:

```
self = <tests.tasks.gcp.test_bigquery.TestBigQueryLoadFileInitialization object at 0x7f57240ff910>

    def test_initializes_with_nothing_and_sets_defaults(self):
        task = BigQueryLoadFile()
        assert task.project is None
        assert task.location == "US"
        assert task.credentials_secret is None
        assert task.dataset_id is None
        assert task.table is None
>       assert task.file_obj is None
E       AttributeError: 'BigQueryLoadFile' object has no attribute 'file_obj'
```

dkapitan:
Thanks for the feedback @joshmeek. I was wondering what was causing it but now I understand. Halfway through I switched from using a file object to path-like object. Will fix that.

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -403,6 +407,150 @@ def run(
...
+            containing a JSON representation of your Google Application credentials
+        - **kwargs (optional): additional kwargs to pass to the `Task` constructor
+    """
+
+    def __init__(
+        self,
+        file: Union[str, Path] = None,
+        rewind: bool = False,
+        size: int = None,
+        num_retries=_DEFAULT_NUM_RETRIES,
```
joshmeek, line 435: I think this should become a kwarg and here it can default to 6:
```
num_retries: int = 6,
```

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -403,6 +407,150 @@ def run(
...
+        "project",
+        "location",
+        "credentials_secret",
+    )
+    def run(
+        self,
+        file: Union[str, Path] = None,
+        rewind: bool = False,
+        size=None,
+        num_retries=_DEFAULT_NUM_RETRIES,
```
joshmeek, line 472: Same comment as above

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -403,6 +407,150 @@ def run(
...
+                records to; if not provided here, will default to the one provided at initialization
+            - table (str, optional): the name of a destination table to write the
+                records to; if not provided here, will default to the one provided at initialization
+            - project (str, optional): the project to initialize the BigQuery Client with; if not provided,
+                will default to the one inferred from your credentials
+            - schema (List[bigquery.SchemaField], optional): the schema to use when creating the table
+            - location (str, optional): location of the dataset that will be written to; defaults to "US"
+            - credentials (dict, optional): a JSON document containing Google Cloud credentials.
+                You should provide these at runtime with an upstream Secret task.
+            - credentials_secret (str, optional, DEPRECATED): the name of the Prefect Secret
```
joshmeek, line 497: Since this type of secret is being deprecated I think this kwarg should be removed and only `credentials` should be used

src/prefect/tasks/gcp/bigquery.py
dkapitan: done.

src/prefect/tasks/gcp/bigquery.py
dkapitan: done

src/prefect/tasks/gcp/bigquery.py
dkapitan: done

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -403,6 +404,153 @@ def run(
...
+            resumable upload will be used. Otherwise, multipart upload will be used.
+        - dataset_id (str, optional): the id of a destination dataset to write the
+            records to
+        - table (str, optional): the name of a destination table to write the
+            records to
+        - project (str, optional): the project to initialize the BigQuery Client with; if not provided,
+            will default to the one inferred from your credentials
+        - schema (List[bigquery.SchemaField], optional): the schema to use when creating the table
+        - location (str, optional): location of the dataset that will be queried; defaults to "US"
+        - credentials_secret (str, optional, DEPRECATED): the name of the Prefect Secret
```
joshmeek, line 425: Left over credentials secret in this task 😄

dkapitan: @joshmeek : had a look, to fix it we need to remove all `credentials_secret` in `tasks/gcp/bigquery.py` since e.g. `TestBigQueryCredentialsandProjects` also still uses these.

dkapitan: @joshmeek happy to remove all `credentials_secret` in `bigquery.py`, but that would be a breaking change. Just let me know what I can do to help.

joshmeek: Good call! There's a large refactor coming in the `0.11.0` release which deals with stuff like these `credentials_secrets` to fall in line with Prefect's new preferred way of using sensitive information (#2326). I say let's remove the use of all `credentials_secret` options in your new tasks and then once `0.11.0` is released (sometime later this week or early next week) this will be in line with the new authentication patterns and we can merge 😄

dkapitan:
@joshmeek excellent, I will put it on my to do list for this week to do all of this. I may need some help though: how do you plan to test the new secrets task? AFAIK there are no integration/mock-up tests yet?

joshmeek:
@dkapitan The use of secrets tasks actually should make it more straightforward to test! Since secrets tasks are a way of securely passing credentials directly to other tasks then your new tasks can be tested as if the values were being directly provided. Of course you could also set secrets locally in a test and then build a flow of `Secret Task -> Your Task` but that may not be needed

joshmeek:
@dkapitan Now that the `0.11.0` release changes have been merged into master this can continue! I can resolve the conflicts if you'd like, whatever works.

dkapitan:
@joshmeek yes please, if you could resolve the issues. Have been reading up on the whole architecture and deployment options of prefect, to have a better understanding about each component before making all the changes. I still don't completely understand the new Secret task I think, particularly in relation to a local prefect server. Could you tell we where in the current code the tests for Secret tasks is found?

joshmeek:
@dkapitan Sure thing! We're also going to probably be improving some of the documentation around this. Here's an example from where the Secret tasks are tested:

https://github.com/PrefectHQ/prefect/blob/964a98aa1d4b72401fe64c7522534b4646f0ba7d/tests/tasks/secrets/test_base.py#L49-L54

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -1,4 +1,6 @@
-from typing import List
+from pathlib import Path
+import warnings
```
joshmeek, line 2:
```suggestion
```

dkapitan:
@joshmeek please review my latest commit.

Issue #2419, opened 2020-04-26
Mutating list while iterating


mcg1969:
https://github.com/PrefectHQ/prefect/blob/8c886b64df8e264791ec1d0e401cbbd6c4719af5/src/prefect/agent/local/agent.py#L75-L77

It's a no-no to mutate the list while iterating. Indeed, in my tests, a loop like this fails to examine the _next_ item in the list, so it's actually not working as intended. For instance:

```
x = [0,1,2,3,4,5,6,7,8]
for idx, y in enumerate(x):
    print(idx, y)
    if y == 3:
        z = x.pop(idx)
    print(idx, y)
```
yields this output; note that the `4` element in the list is never examined.
```
0 0
0 0
1 1
1 1
2 2
2 2
3 3
3 3
4 5
4 5
5 6
5 6
6 7
6 7
7 8
7 8
```
To be fair this is somewhat harmless, because the process will likely be examined in the next heartbeat call, or in the worst case after all of the processes earlier in the list are completed.

joshmeek:
Nice catch! I'm going to label this as a bug because it _technically_ is and/or will be.

trkohler:
Hi! I'd like to help with this! May I?
Do you have any solution in mind?

cicdw:
Hi @trkohler - your help would definitely be appreciated!  You could solve this by creating a new list to track the completed processes like:

```python
    def heartbeat(self) -> None:
        completed_procs = []
        for idx, process in enumerate(self.processes):
            if process.poll() is not None:
                completed_procs.append(idx)
                if process.returncode:
                    self.logger.info(
                        "Process PID {} returned non-zero exit code".format(process.pid)
                    )
        for idx in completed_procs:
            self.processes.pop(idx)
        super().heartbeat()
```
or something to that effect

mcg1969:
Heck, I'd just stuff the _non_-completed procs into a separate list and swap it into `self.processes` after the loop is done.

joshmeek:
Haha so many different ways you could do it! Here's my suggestion:

```python
for idx, process in enumerate(list(self.processes)): 
     if process.poll() is not None: 
         self.processes.pop(idx) 
```

Make a new `list()` with `self.processes` 😉

trkohler:
Hi there!
I did as @joshmeek suggested because I'm too afraid to broke something and get lost in unknown challenges :)

[...]

PR #2464, opened 2020-05-01
self.processes -> list

trkohler:

- [x] adds new tests (if appropriate) <- I don't think it is, but if I wrong, I'll try to add tests.
- [x] updates `CHANGELOG.md` (if appropriate) <- I'm not sure about this also.
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate) <- all docstrings remain the same

## What does this PR change?

heartbeat method in LocalAgent object. Before this method worked with self.processes. I changed it to work with the list copy of self.processes.

## Why is this PR important?

It should fix bug with the issue #2419 

### Note
Pytest locally brought me bunch of errors. There are a lot of them connected to not configured AWS, Google, etc. I'm not sure if this change passed tests because I receive too many errors in my console.


joshmeek:
Thanks for opening the PR @trkohler! Would you mind also adding this fix to the CHANGELOG.md and adding yourself as a contributor there too?

trkohler:
@joshmeek I think I can't be named as a contributor because it's not even a full line of code (shame on me!), but seems like I did everything you asked 🦄

joshmeek:
@trkohler All fixes are contributions in my eyes 😄

mcg1969:
Nicely done!

src/prefect/agent/local/agent.py
```diff
@@ -72,7 +72,7 @@ def __init__(
         )
 
     def heartbeat(self) -> None:
-        for idx, process in enumerate(self.processes):
+        for idx, process in enumerate(list(self.processes)):
```
jcrist, line 75: This is still incorrect, but in a more subtle and dangerous way. Before this would error since you can't mutate a collection over which you're iterating. This now succeeds, but will pop out incorrect elements, since the length of the list changes with each mutation.

- Loop over list with 3 elements `[a, b, c]`
- `idx, process = 0, a`, pop out idx 0, list is now `[b, c]`
- `idx, process = 1, b`, pop out idx 1, list is now `[b]` - *the wrong element was removed*

I think `.processes` should be a set of processes instead of a list, and use `self.processes.remove(process)` to remove completed processes. This provides O(1) remove for individual processes (more efficient than the `O(n)` call to `pop` here), and doesn't run the risk of removing the wrong process.

mcg1969: Oops, you're right. In my own internal conception I was just putting the remaining processes into a new list and swapping that one in place.

jcrist: PR incoming soon.

joshmeek: Oh damn @trkohler I let you astray 🤦

trkohler: it's nice to learn! especially, when it comes with such brilliant explanations

PR #2471, opened 2020-05-02
Results serializers

cicdw:

## What does this PR change?
This PR breaks some of the work from #2391 into it's own, independent PR for easier review.  This PR introduces new serializers for all the new-style result classes.

## Why is this PR important?
Serializers are the language by which Core and Prefect backends communicate; because the backend will need to support older versions of Core still running with old result handlers we need to be very mindful of how these serializers work.  You'll notice this implementation layers _on top_ of the serializers but doesn't change anything related to Result Handlers or SafeResult objects.

### DIFF
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -18,14 +18,12 @@ whose value is `None`.
 """
 import base64
 import copy
-import datetime
 import pendulum
 import uuid
 from typing import Any, Callable, Iterable, Union
 
 import cloudpickle
 
-from prefect.engine.cache_validators import duration_only
 from prefect.engine.result_handlers import ResultHandler
 from prefect.utilities import logging
 
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -19,6 +19,8 @@ whose value is `None`.
 import base64
 import copy
 import datetime
+import pendulum
+import uuid
 from typing import Any, Callable, Iterable, Union
 
 import cloudpickle
@@ -80,12 +82,6 @@ class Result(ResultInterface):
         - validators (Iterable[Callable], optional): Iterable of validation functions to apply to
             the result to ensure it is `valid`.
         - run_validators (bool): Whether the result value should be validated.
-        - cache_for (timedelta, optional): The amount of time to maintain a cache
-            of this result.  Useful for situations where the containing Flow
-            will be rerun multiple times, but this task doesn't need to be.
-        - cache_validator (Callable, optional): Validator that will determine
-            whether the cache for this result is still valid (only required if `cache_for`
-            is provided; defaults to `prefect.engine.cache_validators.duration_only`)
         - location (str, optional): Possibly templated location to be used for saving the
             result to the destination.
     """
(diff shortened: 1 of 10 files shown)


src/prefect/engine/result/base.py
```diff
@@ -128,19 +116,20 @@ def store_safe_value(self) -> None:
                 value=value, result_handler=self.result_handler
             )
 
-    def populate_result(self, result: "Result") -> "Result":
+    def from_value(self, value: Any) -> "Result":
```
lauralorenz, line 119: @cicdw Is the change here because in the pipeline you are calling `state._result.read()` inline and then `self.task.result.from_value(what_you_read)`? and/or just generally if you can let me know the new thinking behind this name/signature change

cicdw: Yup yup for sure @lauralorenz ; so there are two patterns we need within the pipeline:

- creating result instances on the fly from values; this is what `from_value` method is for.  This way we can share the configuration that comes from initialization across different pieces of data.  We need to do this mainly for mapped tasks and looped tasks.
- "hydrating" result objects: this is necessary whenever we receive a State from Cloud which is only a JSON representation.  We need to use the data from that representation (specifically the `location`) and read from that location using the actual Result object.  This is what `populate_result` was for -- I realized in working through the pipeline that this felt more like an interface that _State_ objects should be responsible for, so in the other PR you'll notice this method has moved there.

tests/engine/result/test_base.py
```diff
@@ -15,6 +15,11 @@
 )
 
 
+## TODO: write test that multiple writes / formats works
+## what if we allowed indexing into a result which makes a copy...
+## TODO: create from_value(value)
```
joshmeek, line 20: Doing these in this PR?

src/prefect/engine/result/base.py
```diff
@@ -80,12 +80,6 @@ class Result(ResultInterface):
         - validators (Iterable[Callable], optional): Iterable of validation functions to apply to
             the result to ensure it is `valid`.
         - run_validators (bool): Whether the result value should be validated.
-        - cache_for (timedelta, optional): The amount of time to maintain a cache
-            of this result.  Useful for situations where the containing Flow
-            will be rerun multiple times, but this task doesn't need to be.
-        - cache_validator (Callable, optional): Validator that will determine
-            whether the cache for this result is still valid (only required if `cache_for`
-            is provided; defaults to `prefect.engine.cache_validators.duration_only`)
```
joshmeek, line 88: Why removed and not deprecated?

lauralorenz:
For posterity: I kicked this around a bit yesterday with a scratch flow with caching, looping, results, mapping, and state handlers in it ([source is here](https://gist.github.com/lauralorenz/c67c753dab3da1e5b6468e8657e03acd)) on a Prefect core server off of this branch (`results-serialiers`) against an agent off of `results-wiring` branch, and also an agent off of `0.10.0` as an 'old agent', to sanity check the serializers. With a little wrangling of `results-wiring` between `NORESULT`/`NoResult` at the time of testing, that went fine. 

The current status is that `0.10.0` works ok for a simple flow, but FYI @cicdw, currently `0.10.0` is acting weird with cached states into mapped states on server for me, and I'm investigating now if that is just me or it's actually a problem.

tests/engine/result/test_base.py
cicdw: Nope I can remove this comment

src/prefect/engine/result/base.py
cicdw: They never did anything so no need to deprecate

joshmeek: Oh oh I thought this was the task.py file haha

lauralorenz:
FYI I think it was just me, I reset my database and it was all fine 🤷 

BTW @cicdw was I mistaken that the point of this PR was to go into master to deliver the serializers only to Cloud so we can test 0.11.0 agents against dev/staging? because it got merged into 0.11.0.

PR #2478, opened 2020-05-04
Add backend toggle to k8s agent install

joshmeek:

## What does this PR change?
Adds a toggle for `config.backend` when installing the k8s agent.

## Why is this PR important?
When using the k8s agent against Core's server would cause an issue where a token was expected because it didn't know it wasn't communicating with cloud.


src/prefect/cli/agent.py
```diff
@@ -413,6 +422,7 @@ def install(
             cpu_request=cpu_request,
             cpu_limit=cpu_limit,
             labels=list(label),
+            backend=backend,
```
cicdw, line 425: I kind of think we should default this to the backend value found in `config`

PR #2482, opened 2020-05-04
take docker daemon url in docker storage from DOCKER_HOST env var

### LINKED ISSUE #2157: set default_url for Docker storage off DOCKER_HOST env var if it exists
# Use Case
Users (or services they use; see context below) may set their docker daemon's host to something besides the default, and if so they are likely using the environment variable `DOCKER_HOST` to point any docker clients to the correct location. 

# Solution
Set Docker storage's `default_url` more by trying to check a user's `DOCKER_HOST` env var before defaulting to the hardcoded defaults listed already. See example below as well.

## Archived from the [Prefect Public Slack Community](https://join.slack.com/t/prefect-public/shared_invite/enQtNzE5OTU3OTQwNzc1LTQ5M2FkZmQzZjI0ODg1ZTBmOTc0ZjVjYWFjMWExZDAyYzBmYjVmMTE1NTQ1Y2IxZTllOTc4MmI3NzYxMDlhYWU)

**jschmid**: Question for folks doing CI/CD automation to build &amp; register Prefect Flows. Building Docker storage for one of our flows from our CI/CD system (<http://Screwdriver.io|Screwdriver.io>) fails with the exception below. Docker appears to be installed and working properly because `docker run hello-world` succeeds. (There is special "docker-in-docker" config we use with Screwdriver and this works well for many of our other projects.) I can also build Docker storage successfully on my laptop from inside the same image that our CI/CD system is using. I'm sure it's some simple config that I'm missing -- anybody run into this?

```
15:25:50 Traceback (most recent call last):
15:25:50   File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
15:25:50     "__main__", mod_spec)
15:25:50   File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
15:25:50     exec(code, run_globals)
15:25:50   File "/usr/local/lib/python3.7/site-packages/srm_datasci-0.1.0-py3.7.egg/srm/datasci/flows/util/test_flow.py", line 35, in &lt;module&gt;
15:25:50     flow.register("Utilities")
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/core/flow.py", line 1412, in register
15:25:50     no_url=no_url,
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/client/client.py", line 616, in register
15:25:50     serialized_flow = flow.serialize(build=build)  # type: Any
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/core/flow.py", line 1221, in serialize
15:25:50     storage = self.storage.build()  # type: Optional[Storage]
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/environments/storage/docker.py", line 263, in build
15:25:50     self._build_image(push=push)
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/environments/storage/docker.py", line 293, in _build_image
15:25:50     self.pull_image()
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/environments/storage/docker.py", line 483, in pull_image
15:25:50     client = self._get_client()
15:25:50   File "/usr/local/lib/python3.7/site-packages/prefect/environments/storage/docker.py", line 471, in _get_client
15:25:50     return docker.APIClient(base_url=self.base_url, version="auto")
15:25:50   File "/usr/local/lib/python3.7/site-packages/docker/api/client.py", line 190, in __init__
15:25:50     self._version = self._retrieve_server_version()
15:25:50   File "/usr/local/lib/python3.7/site-packages/docker/api/client.py", line 215, in _retrieve_server_version
15:25:50     'Error while fetching server API version: {0}'.format(e)
15:25:50 docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

**jschmid**: I suspect it's related to this line in the Prefect Docker storage code: <https://github.com/PrefectHQ/prefect/blob/master/src/prefect/environments/storage/docker.py#L96>)

**jschmid**: Was able to work through this. Updating here in case anyone runs into something similar. Screwdriver configures their "docker-in-docker" setup with DOCKER_HOST set to `<tcp://localhost:2375>`  So for us the change was:

```

storage = Docker(
    registry_url="&lt;redacted&gt;",
    image_name="&lt;redacted&gt;",
    image_tag=IMAGE_TAG,
    python_dependencies=["&lt;redacted&gt;"],
    base_url=environ.get("DOCKER_HOST"),
)
```

It might be good to bubble this change down into Prefect, e.g:

```
    default_url = environ.get("DOCKER_HOST", "<unix://var/run/docker.sock>")
```

Here: <https://github.com/PrefectHQ/prefect/blob/master/src/prefect/environments/storage/docker.py#L96>

**laura**: ah, you beat me (though my only contribution was to note about base_url being an option for you to configure once you knew what your CI’s docker api endpoint really was)

**laura**: I like your idea, reading docker docs briefly it looks like DOCKER_HOST is an env var natively supported by the docker client if not  dockerd itself

**jschmid**: Thanks <@US4FKJEGH>, that all makes sense!

**laura**: I’m going to try out the issue bot :slightly_smiling_face:

**laura**: hmm its not that

**laura**: <@ULVA73B9P> archive “set default_url for Docker storage off DOCKER_HOST env var if it exists”

TylerWanner:

## What does this PR change?
Take in DOCKER_HOST env var with precedence over base_url parameter in docs

## Why is this PR important?
closes #2157 
Allows for docker-in-docker/ remote docker daemon setup at buildtime

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,7 @@ Released on May 5, 2020.
 - Add `flow_id` to context for Flow runs - [#2461](https://github.com/PrefectHQ/prefect/pull/2461)
 - Allow users to inject custom context variables into their logger formats - [#2462](https://github.com/PrefectHQ/prefect/issues/2462)
 - Add option to set backend on `agent install` CLI command - [#2478](https://github.com/PrefectHQ/prefect/pull/2478)
+- Allow for setting docker daemon at build time using DOCKER_HOST env var to override base_url in docker storage - [#2482](https://github.com/PrefectHQ/prefect/pull/2482)
 
 ### Task Library
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ These changes are available in the [release/0.11.0 branch](https://github.com/Pr
 
 ### Breaking Changes
 
+- Allow for setting docker daemon at build time using DOCKER_HOST env var to override base_url in docker storage - [#2482](https://github.com/PrefectHQ/prefect/pull/2482)
 - Triggers now accept a dictionary of upstream edges and states instead of a set of states - [#2289](https://github.com/PrefectHQ/prefect/issues/2298)
 - Ensure all calls to `flow.run()` use the same execution logic - [#1994](https://github.com/PrefectHQ/prefect/pull/1994)
 - Moved `prefect.tasks.cloud` to `prefect.tasks.prefect` - [#2404](https://github.com/PrefectHQ/prefect/pull/2404)
@@ -121,7 +122,6 @@ Released on May 5, 2020.
 - Add `flow_id` to context for Flow runs - [#2461](https://github.com/PrefectHQ/prefect/pull/2461)
 - Allow users to inject custom context variables into their logger formats - [#2462](https://github.com/PrefectHQ/prefect/issues/2462)
 - Add option to set backend on `agent install` CLI command - [#2478](https://github.com/PrefectHQ/prefect/pull/2478)
-- Allow for setting docker daemon at build time using DOCKER_HOST env var to override base_url in docker storage - [#2482](https://github.com/PrefectHQ/prefect/pull/2482)
 
 ### Task Library
 
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -105,9 +105,11 @@ class Docker(Storage):
         self.files = files or {}
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
-        self.base_url = environ.get(
+        self.base_url = os.environ.get(
             "DOCKER_HOST",
-            "unix://var/run/docker.sock"
+            base_url
+            if base_url
+            else "unix://var/run/docker.sock"
             if sys.platform != "win32"
             else "npipe:////./pipe/docker_engine",
         )
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -57,7 +57,7 @@ class Docker(Storage):
         - files (dict, optional): a dictionary of files to copy into the image
             when building
         - base_url: (str, optional): a URL of a Docker daemon to use when for
-            Docker related functionality
+            Docker related functionality.  DOCKER_HOST env var takes precedence
         - prefect_version (str, optional): an optional branch, tag, or commit
             specifying the version of prefect you want installed into the container;
             defaults to the version you are currently using or `"master"` if your
@@ -92,12 +92,6 @@ class Docker(Storage):
         secrets: List[str] = None,
     ) -> None:
         self.registry_url = registry_url
-
-        if sys.platform == "win32":
-            default_url = "npipe:////./pipe/docker_engine"
-        else:
-            default_url = "unix://var/run/docker.sock"
-
         self.image_name = image_name
         self.image_tag = image_tag
         self.python_dependencies = python_dependencies or []
@@ -111,7 +105,12 @@ class Docker(Storage):
         self.files = files or {}
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
-        self.base_url = base_url or default_url
+        self.base_url = environ.get(
+            "DOCKER_HOST",
+            "unix://var/run/docker.sock"
+            if sys.platform != "win32"
+            else "npipe:////./pipe/docker_engine",
+        )
         self.local_image = local_image
         self.extra_commands = []  # type: List[str]
         self.ignore_healthchecks = ignore_healthchecks
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -92,6 +92,10 @@ class Docker(Storage):
         secrets: List[str] = None,
     ) -> None:
         self.registry_url = registry_url
+        if sys.platform == "win32":
+            default_url = "npipe:////./pipe/docker_engine"
+        else:
+            default_url = "unix://var/run/docker.sock"
         self.image_name = image_name
         self.image_tag = image_tag
         self.python_dependencies = python_dependencies or []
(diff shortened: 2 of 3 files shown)


src/prefect/environments/storage/docker.py
```diff
@@ -108,7 +102,12 @@ def __init__(
         self.files = files or {}
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
-        self.base_url = base_url or default_url
+        self.base_url = environ.get(
+            "DOCKER_HOST",
+            "unix://var/run/docker.sock"
+            if sys.platform != "win32"
+            else "npipe:////./pipe/docker_engine",
+        )
```
joshmeek, line 110: This doesn't allow for someone to set a `base_url` I think this should be updated to use something like
```
env var if set
else if base_url
else default_url
```

TylerWanner: ah yeah good point.  oversight

src/prefect/environments/storage/docker.py
```diff
@@ -108,7 +102,14 @@ def __init__(
...
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
-        self.base_url = base_url or default_url
+        self.base_url = os.environ.get(
+            "DOCKER_HOST",
+            base_url
+            if base_url
+            else "unix://var/run/docker.sock"
+            if sys.platform != "win32"
+            else "npipe:////./pipe/docker_engine",
+        )
```
joshmeek, line 112: While this works it is quite confusing to look at lol. Could you keep the original platform check and then update this to do something like 

```
os.environ.get("DOCKER_HOST", base_url if base_url else default_url)
```

tests/environments/storage/test_docker_storage.py
```diff
@@ -13,6 +13,11 @@
 from prefect.environments.storage import Docker
 
 
+@pytest.fixture
+def no_docker_host_var(monkeypatch):
+    monkeypatch.delenv("DOCKER_HOST", raising=False)
```
joshmeek, line 18: Can you add a docstring to this fixture to mention its purpose

TylerWanner: gladly.  does that work?

joshmeek:
I may be overthinking this but this is starting to seem like a breaking change. `DOCKER_HOST` is a pseudo-reserved env var when working with docker runtimes. We know it is common for users to build their flows' Docker storages in a CI type system and I'm thinking that implementing this change would break those pipelines. Mainly because if they have this `DOCKER_HOST` already set by default without realizing it then it could hurt their connection to the proper daemon. We observed this on the tests for this change where the env var had to be removed in a fixture.

TylerWanner:
@joshmeek I fundamentally agree but am struggling to wrap my head around the existing workflows that would be broken by the change.  Probably safe to wait for the next minor version to introduce this

TylerWanner:
I'm thinking that the "breaking change" would only affect workflows that run using a remote docker daemon but rely on using the LOCAL docker daemon for building the storage object.  I'm not sure that's a real workflow.  Am I overlooking anything here?
@joeschmid would you like to chime in?  This PR comes from your issue #2482

joeschmid:
@joshmeek & @TylerWanner the concern about this being a potentially breaking change makes sense. My guess (and it's only a guess) is this isn't likely to be a huge issue in practice for a few reasons:

1. The docker command itself seems to make DOCKER_HOST available: https://docs.docker.com/engine/reference/commandline/cli/#environment-variables
2. CI systems (e.g. Screwdriver, etc.) appear to set DOCKER_HOST appropriately such that this will work correctly
2. If a Prefect user is already deploying from CI, they've likely had to address this in some form already, i.e. the previous default was unlikely to work in a CI environment, as we observed.

I didn't do an exhaustive search, but it looks like DOCKER_HOST will be available in common cases:

https://www.testcontainers.org/supported_docker_environment/
https://github.com/travis-ci/travis-ci/issues/9104#issuecomment-361021342

src/prefect/environments/storage/docker.py
```diff
@@ -108,7 +106,7 @@ def __init__(
         self.files = files or {}
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "prefect.core.flow.Flow"]
-        self.base_url = base_url or default_url
+        self.base_url = os.environ.get("DOCKER_HOST", base_url or default_url)
```
cicdw, line 109: I think that if a user provides a `base_url` it should be respected; so in this case we should change this line to:
```suggestion
        self.base_url = base_url or os.environ.get("DOCKER_HOST", default_url)
```

src/prefect/environments/storage/docker.py
```diff
@@ -57,7 +57,7 @@ class Docker(Storage):
         - files (dict, optional): a dictionary of files to copy into the image
             when building
         - base_url: (str, optional): a URL of a Docker daemon to use when for
-            Docker related functionality
+            Docker related functionality.  DOCKER_HOST env var takes precedence
```
cicdw, line 60: If my suggestion below is accepted we should update this doc string accordingly

joshmeek:
Now that `0.11.0` has been merged to master I'm good to continue with this if @cicdw's comments are accepted

CHANGELOG.md
```diff
@@ -121,6 +121,7 @@ Released on May 5, 2020.
 - Add `flow_id` to context for Flow runs - [#2461](https://github.com/PrefectHQ/prefect/pull/2461)
 - Allow users to inject custom context variables into their logger formats - [#2462](https://github.com/PrefectHQ/prefect/issues/2462)
 - Add option to set backend on `agent install` CLI command - [#2478](https://github.com/PrefectHQ/prefect/pull/2478)
+- Allow for setting docker daemon at build time using DOCKER_HOST env var to override base_url in docker storage - [#2482](https://github.com/PrefectHQ/prefect/pull/2482)
```
joshmeek, line 124: @TylerWanner Can you move this to the top of the newer section of the changelog? I'm good on this after that 😄

joshmeek: If you add `[ci skip]` to your commit message than it won't have to rerun all of the checks

PR #2491, opened 2020-05-05
Give a better error when trying to serialize a non-function

### LINKED ISSUE #2432: Serialization fails with no attribute __qualname__
## Description
Serialization fails when a task uses an instance of a callable class as a cache_validator.

```
AttributeError                            Traceback (most recent call last)
~/.virtualenvs/iguazu-venv/lib/python3.7/site-packages/prefect/utilities/serialization.py in _serialize(self, value, attr, obj, **kwargs)
    379         try:
--> 380             qual_name = to_qualified_name(value)
    381         except:

~/.virtualenvs/iguazu-venv/lib/python3.7/site-packages/prefect/utilities/serialization.py in to_qualified_name(obj)
     36     """
---> 37     return obj.__module__ + "." + obj.__qualname__
     38

AttributeError: 'Validator' object has no attribute '__qualname__'

... 

~/.virtualenvs/iguazu-venv/lib/python3.7/site-packages/prefect/utilities/serialization.py in _serialize(self, value, attr, obj, **kwargs)
    383                 raise ValidationError("Invalid function reference: {}".format(value))
    384             else:
--> 385                 return qual_name
    386
    387         # sort matches such that the longest / most specific match comes first

UnboundLocalError: local variable 'qual_name' referenced before assignment
```

More details:

I found a small bug on the serialization code. In `prefect.utilities.serialization.StatefulFunctionReference`  :
```python
    def _serialize(self, value, attr, obj, **kwargs):  # type: ignore
        if value is None and self.allow_none:
            return None

        try:
            qual_name = to_qualified_name(value)
        except:
            if self.reject_invalid:
                raise ValidationError("Invalid function reference: {}".format(value))
            else:
                return qual_name
        ... more code ...
```

There is a return with an undefined variable (because when this fails, there is no qual_name variable.

I am not sure if this is a problem of that particular function, or is from `to_qualified_name`, which fails with an input instance instead of a class.
The culprit from my side is that I have a task with `cache_validator=SomeClass()`  … where SomeClass is a class with a `__call__`  member function, instead of using a cache_validator that is a function. I am attaching an example on the reproduction section.

## Expected Behavior
Since `cache_validator` in a task is type-annotated to a `Callable`, I expected to be able to use an instance of a callable class without issues when serializing.

## Reproduction
```python
import datetime

from prefect import Flow, Task


class Validator:
    def __init__(self):
        print('I am a validator, ready to say false al the time')
    def __call__(self, state, inputs, parameters):
        return False

my_task = Task(cache_validator=Validator(), cache_for=datetime.timedelta(days=1))
with Flow('example') as flow:
    my_task()

flow.storage = Docker()
flow.register()
```

## Environment
```
$ prefect diagnostics
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Darwin-19.3.0-x86_64-i386-64bit",
    "prefect_version": "0.10.4",
    "python_version": "3.7.5"
  }
}
```

jcrist:
Prefect's serialization scheme relies on callable objects having fully
qualified names, a property that is only true (for the most part) for
true functions in importable modules. We now raise a better error
message when trying to use other callables in these locations.

Fixes #2432

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Fixes
 
-- None
+- Give a better error for non-serializable callables when registering with cloud/server - [#2491](https://github.com/PrefectHQ/prefect/pull/2491)
 
 ### Deprecations
 
--- a/src/prefect/utilities/serialization.py
+++ b/src/prefect/utilities/serialization.py
@@ -379,10 +379,9 @@ class StatefulFunctionReference(fields.Field):
         try:
             qual_name = to_qualified_name(value)
         except:
-            if self.reject_invalid:
-                raise ValidationError("Invalid function reference: {}".format(value))
-            else:
-                return qual_name
+            raise ValidationError(
+                f"Invalid function reference, function required, got {value}"
+            )
 
         # sort matches such that the longest / most specific match comes first
         valid_bases = sorted(
--- a/tests/utilities/test_serialization.py
+++ b/tests/utilities/test_serialization.py
@@ -279,6 +279,14 @@ class TestStatefulFunctionReferenceField:
         )
         assert deserialized["f_allow_invalid"] is None
 
+    def test_serialize_non_function_good_error(self):
+        class Foo(object):
+            def __call__(self, a, b):
+                return a + b
+
+        with pytest.raises(marshmallow.ValidationError, match="function required"):
+            self.Schema().dump(dict(f=Foo()))
+
     def test_serialize_none(self):
         with pytest.raises(marshmallow.ValidationError):
             self.Schema().dump({"f": None})


jcrist:
I don't think the docstring description of `reject_invalid` is correct. Per the spec, it should raise if the value is invalid, otherwise ignore it but not deserialize it. The question then is - what is "invalid"? Is it just checking that the provided object is from a set of valid functions? Or is it also checking that the provided object is a function (what we do here)?  I'd like to update the docstring to better describe the code before we merge this.

joshmeek:
@jcrist Good question! 

ping @jlowin

jlowin:
@jcrist excellent question, and maybe we should tighten that up - I think the code changed in a small way since its original implementation. Originally, we used the `from_qualified_name()` function in this file to transform the string qualname into a callable. However, this raised a concern that we were importing off a string input and we didn't want users to be concerned that malformed strings could create bad outcomes. So originally, I believe this worked by allowing you to pass a `set` of `valid_functions` and you couldn't deserialize (import) a string unless it matched that whitelist. `reject_invalid` was a way to forcibly error if someone attempted to go around the whitelist; otherwise it would silently ignore it.

The idea was that you could create, for example, a custom trigger function for your task and successfully serialize that task for a remote server (like Prefect Cloud), even though Cloud didn't have access to your custom trigger. Cloud, in turn, would be able to deserialize your task even though you referenced a custom trigger and, more importantly, would not attempt to import that user-supplied string.

However in reviewing the code now, we adopted a {more sensible, saner} approach in which the callables are directly provided as `valid_functions` and no import takes place, so `reject_invalid` isn't doing too much.

jcrist:
Thanks @jlowin - that doesn't quite answer the question I asked above though, but provides a bit of clarity.

The main question was, should non-serializable custom functions be allowed (which would need a different fix), or should they error informatively (as added here).

I guess it's not clear to me in prefect when and where deserialization happens (still wrapping my head around the internals). Does user-run code (e.g. in a flow run started by an agent) ever deserialize a function that was previously serialized here? Or is deserialization only ever used by cloud for display in the UI etc..., and state holding the custom functions is stored in a different manner?

jlowin:
Sorry @jcrist, didn't parse your note closely enough.

I think the question of serializing arbitrary functions has NEVER come up in practice and really shouldn't be a design consideration now. Your fix reduces complexity so I like it!

To your second point about when deserialization happens - in general, *serialization* happens anytime Prefect Core exchanges information with a backend (either Server or Cloud). Serialization attempts to preserve as much metadata as possible without allowing the possibility of executable code (which is why I shared the `reject_invalid` motivation, and why we don't use something like cloudpickle). Deserialization is actually more rare, and most frequently used to validate inputs. The reason for this is that most deserialized objects can't *do anything* (because they don't carry code except Prefect builtins), so asserting that they deserialize correctly is as close as we can get to testing their validity. One exception is `State` classes. Because these are the most important part of the Core / backend communication pattern, `States` are frequently serialized and then rehydrated via deserialization, both from Core to Cloud (via setting states) and Cloud to Core (via retrieving initial states).

^ Trying to straddle a line between "informative" and "too much detail for a PR comment", let me know if I was helpful!

cicdw:
Two additional cents: for most things it's OK to allow for custom functions; even more generally, there's no fundamental reason we can't also allow for arbitrary callables in this instance.  The trick is to make sure that Cloud can still deserialize the JSON into something meaningful despite not having import access to the custom function / class. 

> Does user-run code (e.g. in a flow run started by an agent) ever deserialize a function that was previously serialized here? Or is deserialization only ever used by cloud for display in the UI etc..., and state holding the custom functions is stored in a different manner?

No code that is directly attached to tasks / flows is ever deserialized on the user side.  However, all states go through the serialization / deserialization process on the user-side which typically includes Results / Result Handlers, so that is one place where we have to be extra careful to correctly re-hydrate the underlying Result off the Task / Flow instance (this is one of the things that made PIN 16 trickier than it seems on the surface).

jcrist:
> Two additional cents: for most things it's OK to allow for custom functions; even more generally, there's no fundamental reason we can't also allow for arbitrary callables in this instance. The trick is to make sure that Cloud can still deserialize the JSON into something meaningful despite not having import access to the custom function / class.

IIUC, you're saying in this case (and many others) we could support arbitrary callables - serialization is only to provide extra information to the UI/logs in cloud/server? So a builtin function might send:

```python
{"fn": "prefect.engine.cache_validators.all_parameters", "kwargs": {}}
```

and a custom function that we don't want to send anything meaningful about might be:

```python
{"fn": "user-defined-function", "kwargs": {}}  # or some other way of indicating a custom function
```

Would supporting this be a better fix than what we've done here (which is to provide a better error and explicitly (rather than implicitly) disallow arbitrary callables)?

cicdw:
Yup exactly, for `cache_validators` this is purely for UI / API visibility.

> Would supporting this be a better fix than what we've done here (which is to provide a better error and explicitly (rather than implicitly) disallow arbitrary callables)?

Yea, I think that would be a stronger fix in this case.  I'm cool with merging this as a short-term improvement but I think ultimately allowing for a broader set of callables is the way to go.

jcrist:
> Yea, I think that would be a stronger fix in this case. I'm cool with merging this as a short-term improvement but I think ultimately allowing for a broader set of callables is the way to go.

Alright, lets do that then. I think for now the better error message is an improvement - I believe this is good-to-merge.

 I'll open an issue about improving the serializers.

PR #2511, opened 2020-05-07
Change some tests around given the new boto3 utility function in 0.11.0

lauralorenz:

- [n/a] updates `CHANGELOG.md` (if appropriate)

## What does this PR change?
In release 0.11.0 there is a new utility function to create boto3 clients that honors the new rules regarding Secret authentication (`utilities.get_boto_client`). This PR fixes a typo in its docstring, and moves tests that were being repeated unevenly in AWS related tasks that had to do with the utility function to a single version of those tests only on the utility function. In doing so I decreased code coverage since the repetitive version of those tests were falsely inflating the test coverage for AWS related tasks's `run` methods, so I threw in a few basic unittests that test the run methods more directly.

I got in a tussle with getting my mock to surface method calls properly, and ended up falling into the same pattern as https://github.com/PrefectHQ/prefect/blob/f75a7f8ef97e92ef88655994eb468aaf9db6e12e/tests/tasks/azure/test_cosmosdb.py#L104-L105 as a workaround.

## Why is this PR important?
Simplify tests, and increase code coverage since secrets have changed in 0.11.0 (hopefully this actually properly did this instead of decrease in the end ;) )

### DIFF
--- a/src/prefect/utilities/aws.py
+++ b/src/prefect/utilities/aws.py
@@ -22,7 +22,7 @@ def get_boto_client(
         - **kwargs (Any, optional): additional keyword arguments to pass to boto3
 
     Returns:
-        - Client: an initialized and authenticated Google Client
+        - Client: an initialized and authenticated boto3 Client
     """
     aws_access_key = None
     aws_secret_access_key = None
--- a/tests/utilities/test_aws.py
+++ b/tests/utilities/test_aws.py
@@ -2,6 +2,8 @@ from unittest.mock import MagicMock
 
 import pytest
 
+pytest.importorskip("boto3")
+
 import prefect
 from prefect.utilities.aws import get_boto_client
 from prefect.utilities.configuration import set_temporary_config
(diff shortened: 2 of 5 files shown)


src/prefect/utilities/aws.py
```diff
@@ -22,7 +22,7 @@ def get_boto_client(
         - **kwargs (Any, optional): additional keyword arguments to pass to boto3
 
     Returns:
-        - Client: an initialized and authenticated Google Client
+        - Client: an initialized and authenticated boto3 Client
```
cicdw, line 25: good catch

tests/utilities/test_aws.py
```diff
@@ -0,0 +1,74 @@
+from unittest.mock import MagicMock
+
+import pytest
+
+pytest.importorskip("boto3")
```
cicdw, line 5: I added this line so the tests would pass in barebones environments 👍

PR #2512, opened 2020-05-07
Fixes from integration testing

cicdw:

## What does this PR change?
This PR fixes two issues identified in integration testing:
- an incredibly rare situation in which `flow.run()` runs on an infinite loop
- state serializers for `cached_inputs` losing information

There is still one large outstanding issue we need to address: how to handle `cached_inputs` for mapped children.  For standard situations, task outputs are checkpointed and can be safely / directly stored as a cached input to downstream tasks that request caching, retries, etc.  However, with mapping specifically, the actual input to the task is an altered version of the upstream output (in particular it's an indexed item from the upstream list).  The big question is how to store this piece of data and where?

## Why is this PR important?
For a successful release!

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -897,7 +897,11 @@ class Flow:
                     context=flow_run_context,
                     **kwargs
                 )
-                if not isinstance(flow_state.result, dict):
+
+                # if flow_state is still scheduled; this most likely means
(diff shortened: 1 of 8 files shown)


joshmeek:
@cicdw Not sure if I follow on the mapped children cached inputs part 🤔How exactly has that functionality changes from how it worked before?

PR #2521, opened 2020-05-08
Write flows in Docker storage to /opt directory

### LINKED ISSUE #2025: Expose ability to set location and / or  permissions on Flow files within Docker storage
Currently, we place Flow objects in the `/root` directory of their Docker containers.  This can be a problem when root access is somehow restricted.  We should introduce two new APIs:

- introduce a way to configure the directory / location of the flow within the image
- optionally set the correct permissions on the flow file so that non-root users can read / write it

joshmeek:

## What does this PR change?
Closes #2025 

## Why is this PR important?
Allows flows to be written to Docker storage without requiring root permissions.

### DIFF
--- a/docs/orchestration/recipes/deployment.md
+++ b/docs/orchestration/recipes/deployment.md
@@ -17,7 +17,7 @@ Docker provides an excellent industry-standard abstraction for shipping code alo
 Whenever you call `flow.register` or build a [Docker storage object](../../api/latest/environments/storage.html#docker) yourself, Prefect will perform the following actions:
 
 - calls `cloudpickle.dumps(flow)` on your Flow object to convert it to serialized bytes
-- stores these bytes inside the Docker image in the `/root/.prefect/` directory
+- stores these bytes inside the Docker image in the `/opt/.prefect/` directory
 - runs various health checks on your Flow inside the image to try and catch any issues
 
 ::: warning cloudpickle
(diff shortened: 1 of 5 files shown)


docs/orchestration/recipes/deployment.md
```diff
@@ -17,7 +17,7 @@ Docker provides an excellent industry-standard abstraction for shipping code alo
 Whenever you call `flow.register` or build a [Docker storage object](../../api/latest/environments/storage.html#docker) yourself, Prefect will perform the following actions:
 
 - calls `cloudpickle.dumps(flow)` on your Flow object to convert it to serialized bytes
-- stores these bytes inside the Docker image in the `/root/.prefect/` directory
+- stores these bytes inside the Docker image in the `/opt/.prefect/` directory
```
jcrist, line 20: Dot directories in `/opt/` feel a little off to me, perhaps `/opt/prefect`? Either works though.

joshmeek: What are your thoughts on defaulting to something other than `opt`? I do like keeping with `.prefect` since that's what it is locally

jcrist: IMO dot directories only belong under `~/`, anything global shouldn't be hidden and should have an explicit name. Here `/opt/prefect`, or `/var/prefect` make the most sense to me (`/opt/prefect` I think makes more sense).

There is prior convention for this too - many systems load configuration/plugins from a mix of local and global directories (for example, dask looks in `~/.config/dask` then `/etc/dask`).

All of these are only conventions though, it doesn't really matter where we put these things as long as we're consistent.

PR #2525, opened 2020-05-08
Cleanup code quality tests

### LINKED ISSUE #2483: Change how we use mypy in CI
Currently we use `mypy` as part of our CI process. This frequently leads to extra work trying to appease the type checker for code that is definitely correct (see e.g. https://github.com/PrefectHQ/prefect/pull/2406).

We *do* want mypy to check that the type signatures of public methods actually match their implementation, but we *don't* want mypy to error for internal type checking that it thinks is invalid (internal errors can be caught by tests, and frequently the type checker is more of a burden than an aid). Perhaps there's a mypy configuration setting we can change to get this behavior?

We also currently run the mypy tests in two locations:

- As part of the `Check code style and docs` build
- As part of a test in `tests/test_formatting.py`, in the `Run tests` build

These two CI builds have different environments (in particular, prefect (and dependencies) aren't installed in the former but are in the latter). This difference can lead to mypy tests passing in one environment and failing in the other.

jcrist:
- Removes duplicate code quality checks run by `py.test`, in favor of the ones run directly with the tools themselves (`mypy`, `black`). These duplicated tests ran in slightly different environments, making debugging trickier
- Ensures prefect's dependencies are installed when checking with `mypy`, which lets mypy use the type signatures in the dependencies rather than marking them all as `Any`.
- Cleans up the `mypy` configuration a bit.

Fixes #2483.


conftest.py
```diff
@@ -1,32 +0,0 @@
-import sys
-
-import pytest
-
-
-def pytest_addoption(parser):
-    parser.addoption(
-        "--skip-formatting",
```
cicdw, line 8: This formatting flag is also used for our doc tests; I think we should keep it but with a rename (I generally don't run doc tests locally until I'm nearing the end)

.circleci/config.yml
```diff
@@ -238,7 +238,7 @@ jobs:
 
       - run:
           name: Run tests
-          command: pytest -vvrfEsx --skip-formatting --cov=prefect --cov-report=xml:/tmp/workspace/coverage/python36-coverage.xml .
+          command: pytest -vvrfEsx --cov=prefect --cov-report=xml:/tmp/workspace/coverage/python36-coverage.xml .
```
cicdw, line 241: Because the formatting flag was also used on doc tests we probably want to keep it here so "test failure" means an actual unit test and not a doc test

jcrist: By doc tests, you mean the test in the `docs` directory?

Generally I run `py.test src/prefect`, rather than letting pytest find any files in the directory tree that start with `test_`. So running the docs tests would be `py.test docs` and running the code tests is `py.test src`. Seems cleaner than adding a new mark, but up to you.

jcrist: You could also have pytest ignore the docs folder for general `py.test` runs (so you can keep your workflow), and only a `py.test docs` command runs the docs tests.

cicdw: > By doc tests, you mean the test in the docs directory?

Yup exactly.

> So running the docs tests would be py.test docs and running the code tests is py.test src

That makes sense; the original idea is to make doc tests opt _out_ so that outside developers know that we run doc tests (otherwise they're likely to get surprised when CI jobs start failing).

I don't really have a strong opinion about which pattern to use, @joshmeek / @lauralorenz might have a stronger opinion.

jcrist: > That makes sense; the original idea is to make doc tests opt out so that outside developers know that we run doc tests (otherwise they're likely to get surprised when CI jobs start failing).

Because I'm so used to running tests as:

```shell
$ py.test tests/
```

I actually missed that tests were also stored in the `docs/` folder (since py.test won't look outside the directories you pass explicitly). Definitely just a preference/what you're used to thing, could go either way.

joshmeek: No opinion

lauralorenz: FWIW I'm of the folk who just runs `pytest` on a new project and would have been included in the contributors who would see that docs test exist from that. The `skip-formatting` flag though, if it is in practice the same as `pytest tests/`, seems unnecessary to me (in other words `pytest tests/` seems like the more standard/easier opt-out). I think there may be some nuance that `pytest tests/` and `pytest --skip-formatting` isn't exactly the same if the `formatting` attribute isn't on every test in the docs directory but not sure it's worth distinguishing for us.

jcrist: Ok, following that I'm going to leave the config as is, but update the CI to explicitly state the directories. Unqualified `py.test` will run everything, specify `tests` or `docs` if you only want to run one.

cicdw: Cool cool - so sounds like we should remove the`--skip-formatting` flag and adjust the CI jobs (including appveyor) to not use the flag

jcrist:
I believe I've addressed all comments.

PR #2526, opened 2020-05-08
DbtShellTask

mhmcdonald:

## What does this PR change?

Adds a DbtShellTask, an extension of the ShellTask, to the task library. This task is intended for working with [dbt](https://www.getdbt.com/)

## Why is this PR important?

The dbt community is very strong and growing in the cloud data warehousing space. There is a lot of potential overlap between the prefect and dbt communities.


jlowin:
Thanks for the contribution @mhmcdonald! Awesome to see DBT get some first-class representation

tests/tasks/dbt/test_dbt.py
```diff
@@ -0,0 +1,81 @@
...
+    out = f.run()
+    profiles_path = os.path.join(test_path, "profiles.yml")
+    assert out.is_successful()
+    assert os.path.exists(profiles_path)
+    os.remove(profiles_path)
+
+
+def test_shell_uses_dbt_envar(test_path):
+    dbt_project_path = os.path.join(test_path, "dbt_project")
+    os.environ["DBT_PROFILES_DIR"] = dbt_project_path
```
cicdw, line 51: We should use [monkeypatch] to patch the os environ like https://github.com/PrefectHQ/prefect/blob/7eb7a630be8b89ba569f17571c2b6bb090db82a7/tests/test_configuration.py#L59

joshmeek:
Weird test failures 🤔all look to be related to logging in some way. I attempted to reproduce using this branch on a fresh conda env but everything was passing for me locally. cc @cicdw any ideas?

tests/tasks/dbt/test_dbt.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32", reason="DbtShellTask currently not supported on Windows"
+)
+
+
+@pytest.fixture()
+def test_path():
+    return os.path.dirname(os.path.realpath(__file__))
```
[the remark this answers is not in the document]
mhmcdonald: done. thanks for the feedback @cicdw

joshmeek:
@mhmcdonald I think we've narrowed down why logging related tests are failing on this branch. It appears that the `dbt` library modifies global logging settings when imported 

https://github.com/fishtown-analytics/dbt/blob/8686ab9a9ddffb63b64d3ee2861f132bf77f54c1/core/dbt/logger.py#L362-L363

which in the tests is causing all logs to be redirected elsewhere. Since the `dbt` library isn't actually used in these tasks and they shell out to using the CLI instead I think we should remove the dependency and instead add some type of CLI check where the `importorskip` is in order to determine if dbt is installed.

tests/tasks/dbt/__init__.py
```diff
@@ -0,0 +1,5 @@
+from shutil import which
+import pytest
+
+if not which("dbt"):
```
cicdw, line 4: clever

PR #2529, opened 2020-05-08
Avoid merge conflicts in CHANGELOG.md

### LINKED ISSUE #2311: Avoid merge conflicts in CHANGELOG.md
With the current changelog policy (add an entry as part of the PR, all in one file) it's quite easy to run into merge conflicts as features are developed concurrently. This is not a problem unique to Prefect, many projects have run into this issue. There are three common solutions:

- Add a custom merge strategy for the `CHANGELOG.md` file (`merge=union`) to better handle these conflicts. This will usually work, but *may* run into issues if concurrently added features have very similar lines.
- Split the changelog into file-per-feature, and use a tool ([towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/quickstart.html) is popular) to generate the final changelog from this before release.
- Add a bunch of filler entries (e.g. empty `-` list rows), and rely on developers adding entries in different spots to avoid merge conflicts.

jcrist:
Instead of having every PR modify `CHANGELOG.md` themselves (which
easily led to merge conflicts), we now store changes in the `changes/`
directory. Each PR adds a single file to the directory describing the
change. On release, the `update_changelog.py` script can be run, which
compiles the changelog and updates `CHANGELOG.md` (and clears the
changes directory to restart this process). An `EXAMPLE.yaml` changelog
entry will persist between clears to provide some documentation of this
process.

Fixes #2311.

### DIFF
--- a/changes/EXAMPLE.yaml
+++ b/changes/EXAMPLE.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+enhancement:
+  - "Example change description - [#1234](https://github.com/PrefectHQ/prefect/pull/1234)"
+
+contributor:
+  - "[Contributor Name](https://github.com/contributor_github_username)"
--- a/changes/EXAMPLE.yaml
+++ b/changes/EXAMPLE.yaml
@@ -3,6 +3,7 @@
 # 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
 #   - feature
 #   - enhancement
+#   - server
 #   - task
 #   - fix
 #   - deprecation
--- a/update_changelog.py
+++ b/update_changelog.py
@@ -1,0 +1,131 @@
+import argparse
+import datetime
+import glob
+import os
+import sys
+
+import yaml
+
+SKIPLINES = 1
+SECTIONS = [
+    "feature",
+    "enhancement",
+    "task",
+    "fix",
+    "deprecation",
+    "breaking",
+    "contributor",
+]
+DEDUPLICATE_SECTIONS = ["contributor"]
+
+TEMPLATE = """
+## {version} <Badge text="beta" type="success">
+
+Released on {date}.
+
+### Features
+
+{feature}
+
+### Enhancements
+
+{enhancement}
+
+### Task Library
+
+{task}
+
+### Fixes
+
+{fix}
+
+### Deprecations
+
+{deprecation}
+
+### Breaking Changes
+
+{breaking}
+
+### Contributors
+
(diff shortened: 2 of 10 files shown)


changes/EXAMPLE.yaml
```diff
@@ -0,0 +1,23 @@
...
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+enhancement:
+  - "Example change description - [#1234](https://github.com/PrefectHQ/prefect/pull/1234)"
+
+contributor:
+  - "[Contributor Name](https://github.com/contributor_github_username)"
```
jcrist, line 23: We could definitely change this to be more structured if we wanted (for example, take in github issue/pr numbers and autogenerate the links). I went with the simplest thing possible for now.

jcrist: Example with more structure:

```yaml
description: "Example change description"
kind: enhancement
links:
  - issue: 1234
  - pr: 1235
contributor:
  name: Contributor Name
  github: contributor_username
```

Multiple changelog entries in a single file (rare) could be separated by `---`, and loaded a yaml stream.

I'm not convinced this is better, it's just an option. The current approach might be easier for those familiar with markdown (it's also simpler to implement). The more structured approach might be easier for those that aren't, and could let us enforce changelog formatting more.

jcrist:
> My two cents is that (for me personally as a dev) creating and committing new files is on par with resolving a changelog merge conflict in the GitHub UI

I generally agree with this, but think that as we work more in parallel (and with more external contributors), the pain of the current strategy will get worse. I resolved a merge conflict 3 times a few weeks ago before the PR was finally merged, which was a wee bit frustrating. Note that the current changelog file on master is also malformed due to merge issues from the 0.11.0 release branch (there's two changelog headers, and two unreleased sections (which may be intentional?)). If a PR lapses between releases (which happen weekly), the PR has to update the section it is added to the changelog, or when it is merged it will be added to the incorrect release notes (I fixed one of these last week).

cicdw:
Yea, those are good points.  I'm +1

jlowin:
I'm a big fan - I hate changelog conflicts passionately. In addition to updating our development docs I have two suggestions:

1. Take the comment at the top of `EXAMPLE.yaml` and add it to a README in the `changes/` directory, purely to help users discover the instructions (I still really like the `EXAMPLE.yaml` file, I just was pleasantly surprised to find it contained instructions in addition to an example)

2. (perhaps not for this PR) this feels like it could be the beginning of a `prefect dev` CLI group. Something like `prefect dev create-changelog --pr 2402`

lauralorenz:
For posterity, @joshmeek myself and @jcrist discussed IRL yesterday, and we are for it! In response to https://github.com/PrefectHQ/prefect/pull/2529#discussion_r422417887, we agreed that simpler is better and to leave the metadata yaml in its current format. We felt good to merge and use it by hand for the next few releases, and then add it to automation on CircleCI somehow.

update_changelog.py
```diff
@@ -0,0 +1,131 @@
+import argparse
+import datetime
+import glob
+import os
+import sys
+
+import yaml
+
+SKIPLINES = 1
+SECTIONS = [
```
jlowin, line 10: Just a note - I believe "server" was added as a section in this release
```suggestion
SECTIONS = [
    "server",
```

jcrist:
I've updated this with master, updated the developer docs, updated the PR template, and added a readme. To avoid having to convert existing entries over to yaml files, I think we should merge this right after the next release (maybe tuesday?).

PR #2531, opened 2020-05-09
Deterministic Task Slugs

cicdw:

## What does this PR change?
This PR is an experiment with introducing deterministic task slugs.  The proposed logic is that task slugs are tracked by the Flow object and determined based on:
- the name of the task
- the tags on the task
- the order in which the task is added to the flow

## Why is this PR important?
Right now, the fact that slugs are randomly / non-deterministically generated on the Task objects is the source of some complication in the Prefect ecosystem.  Ultimately, Task slugs are intended to:
- be unique within a Flow
- provide a way of identifying a Task Run with a Task (we have to know which python object a task run is associated with somehow)

Because slugs are randomly generated though, it becomes impossible to:
- edit your code to "fix" a flow run (because you have to register the flow and its tasks)
- track the history of a task across multiple versions of the flow

Moreover, many data engineers think of their Flow code as the Flow artifact itself, whereas Prefect takes a stricter software approach: the code _generates_ an artifact (in this case a Python object) that needs to be stored and deployed, and any rerun of that code necessarily produces a new artifact.  

The current PR would allow for a much more flexible paradigm, of course with _many_ caveats about how it should and should not be used.


jlowin:
There's a method on the `Flow` called `validate()` (which hasn't been touched in about two years) but the motivation is that it ensures integrity among flow objects - it might be interesting to add a validation test that `set(self.slug.values()) == self.tasks`

jlowin:
One note: the concept doc for tasks will need a small update, as it describes how slugs are generated.

gryBox:
> the order in which the task is added to the flow
What does this mean?  Does it relate to situations where a `task` with the same name gets updated to a `flow` with a duplicate name?

cicdw:
@gryBox yup exactly; e.g., 
```python
@task 
def my_task(x):
    return 1

with Flow("ex") as flow:
    my_task(1)
    my_task(2)
```

src/prefect/core/task.py
```diff
@@ -375,9 +375,8 @@ def copy(self, **task_args: Any) -> "Task":
 
         new = copy.copy(self)
 
-        # ensure new slug is provided
-        if "slug" not in task_args:
-            task_args["slug"] = str(uuid.uuid4())
+        if new.slug and "slug" not in task_args:
+            task_args["slug"] = new.slug + "-copy"
```
cicdw, line 379: open to other suggestions here - this is super naive

jlowin: I opened #2806 to remove this entirely, so the naive approach isn't terrible now

src/prefect/core/flow.py
```diff
@@ -1284,7 +1294,10 @@ def serialize(self, build: bool = False) -> dict:
 
         self.validate()
         schema = prefect.serialization.flow.FlowSchema
-        serialized = schema(exclude=["storage"]).dump(self)
+        flow_copy = self.copy()
```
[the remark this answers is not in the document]
cicdw: Comments added @jlowin

gryBox:
Hi - tasks that are not parameters have a `slug` equal to `None` when I tested this with version `0.12.0+72.gefbfb9ef7`.   Shouldn't there be some `slug`?  

```
@task
def add_one(a_number: int): 
    return a_number + 1

@task(log_stdout=True)
def print_result(the_result:int): 
    print(the_result)

with Flow("Add") as add_fl:
    a_number = Parameter("a_number", default=1)
    
    the_result = add_one(a_number)
    
    print_result(the_result)
```


```
for task in add_fl.get_tasks():
    print(f"name {task.name}")
    print(f"type {task}")
    print(f"slug {task.slug} \n")

name add_one
type <Task: add_one>
slug None 

name print_result
type <Task: print_result>
slug None 

name a_number
type <Parameter: a_number>
slug a_number
```

cicdw:
Hi @gryBox - that is correct and we are planning on removing slugs from tasks altogether with https://github.com/PrefectHQ/prefect/issues/2806 

Essentially in order for slugs to be stable, they need to be defined based on (task, flow) instead of on task alone, hence the move to storing them on Flows directly instead of on Tasks.

gryBox:
Hi @cicdw - I've gotten the chance to play around with `flow.slugs`.  I am trying to understand, _as a user_, why Flows have tasks mapped to slugs but tasks have a slug equal to `None`  -- When they are bound to a flow?

cicdw:
I wouldn't expect users to really interact with `slug`s honestly - the main purpose of a slug is to have a stable ID that allows the Prefect API to determine which Python object is undergoing state updates.   Slugs are not an inherent property of tasks, they're a property of tasks within flows, hence the move away from storing them on task objects directly.

PR #2540, opened 2020-05-11
Add serializer for custom Results

joshmeek:

## What does this PR change?
Adds a serializer for custom results

## Why is this PR important?
So users can write their own result handlers and orchestrate them with a backend

### DIFF
--- a/tests/serialization/test_results.py
+++ b/tests/serialization/test_results.py
@@ -153,7 +153,7 @@ def test_result_handler_result():
 
 
 def test_custom_result():
-    class CustomResult(Result):
+    class MySuperAwesomeAmazingTopOfTheLineCustomResult(Result):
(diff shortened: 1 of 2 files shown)


tests/serialization/test_results.py
```diff
@@ -150,3 +150,41 @@ class MyHandler(ResultHandler):
     assert isinstance(new_result, results.ResultHandlerResult)
     assert new_result.location == "foo"
     assert new_result.value is None
+
+
+def test_custom_result():
+    class CustomResult(Result):
```
cicdw, line 156: Can you change the name of this custom class for the sake of the test? It's not clear from the test if we expect the class name to be stored in the `"type"` field or not (spoiler alert: it won't be)

joshmeek: Good call

PR #2548, opened 2020-05-12
Add idiom for using Result targets

joshmeek:
Looking for comments on what we could add to this / any way to make the idiom more specific


docs/core/idioms/targets.md
```diff
@@ -0,0 +1,57 @@
+# Using a Result target
```
joshmeek, line 1: Looking for a way to make this more of a specific idiom if necessary

cicdw: Maybe you can name this "Using Result targets for efficient caching" or something like that?

docs/core/idioms/targets.md
```diff
@@ -0,0 +1,63 @@
+# Using Result targets for efficient caching
+
+Targets in Prefect are templatable location strings that are used to check for the existence of a task [Result](/core/concepts/results.html). This is useful in cases where you might want a task to only write data to a location once or merely not rerun if some piece of data is present. If a result exists at that location then the task run will enter a cached state. This behavior is commonly referred to as caching in Prefect and it is generally used when you do not want to recompute the result of a task if it already exists.
```
cicdw, line 3:
```suggestion
Targets in Prefect are templatable location strings that are used to check for the existence of a task [Result](/core/concepts/results.html). This is useful in cases where you might want a task to only write data to a location once or merely not rerun if some piece of data is present. If a result exists at that location then the task run will enter a cached state. This behavior is commonly referred to as caching in Prefect and it is generally used when you do not want to recompute the result of a task if it already exists and can be retrieved easily.
```

docs/core/idioms/targets.md
```diff
@@ -0,0 +1,63 @@
...
+
+The flow below will write the result of the first task to a target specified with a filepath of `task_name-today`. For a list of all available context variables at runtime visit the [context API documentation](/api/latest/utilities/context.html#context). This is a powerful construct because it means that only the first run of this task on any given day will run and write the result. Any other runs up until the next calendar day will use the cached result found at `task_name-today`.
```
cicdw, line 5: I think you should emphasize the templating more for `task_name-today`.  I don't think it's clear that these values aren't hard coded.

docs/core/idioms/targets.md
```diff
@@ -0,0 +1,63 @@
...
+
+::: warning Result Types
+Please review the [documentation for each result type](/api/latest/engine/results.html) before you configure. Some types have specific configuration options for where to write the result.
+
+For example: the `LocalResult` below has an optional `dir` kwargs which accepts an absolute path to a directory for storing results.
```
cicdw, line 10: Can you add a sentence to this local result call out that says "In this case the `target` value will be appended to the directory of choice."

docs/core/idioms/targets.md
```diff
@@ -0,0 +1,63 @@
+# Using Result targets for efficient caching
```
cicdw, line 1: Can you add an 0.110+ badge?

docs/core/idioms/targets.md
```diff
@@ -1,13 +1,13 @@
...
 
-The flow below will write the result of the first task to a target specified with a filepath of `task_name-today`. For a list of all available context variables at runtime visit the [context API documentation](/api/latest/utilities/context.html#context). This is a powerful construct because it means that only the first run of this task on any given day will run and write the result. Any other runs up until the next calendar day will use the cached result found at `task_name-today`.
+The flow below will write the result of the first task to a target specified with a filepath of `{task_name}-{today}`. These values are not hardcoded and instead will be interpreted using [Result templating](/core/concepts/results.html#templating-result-locations). For a list of all available context variables at runtime visit the [context API documentation](/api/latest/utilities/context.html#context). This is a powerful construct because it means that only the first run of this task on any given day will run and write the result. Any other runs up until the next calendar day will use the cached result found at `{task_name}-{today}`.
```
cicdw, line 5:
```suggestion
The flow below will write the result of the first task to a target specified with a filepath of `{task_name}-{today}`. These values are not hardcoded and instead will be interpolated using [Result templating](/core/concepts/results.html#templating-result-locations). For a list of all available context variables at runtime visit the [context API documentation](/api/latest/utilities/context.html#context). This is a powerful construct because it means that only the first run of this task on any given day will run and write the result. Any other runs up until the next calendar day will use the cached result found at `{task_name}-{today}`.
```

PR #2553, opened 2020-05-13
Add unique name option to Kubernetes Job Environment

joshmeek:

## What does this PR change?
Adds option to use a unique name for each job created with the Kubernetes Job Environment by adding on the first eight characters from the flow run id.

## Why is this PR important?
Previously it would not be possible to run two runs of the same flow at the same time using this environment due to k8s naming collisions.

### DIFF
--- a/src/prefect/environments/execution/k8s/job.py
+++ b/src/prefect/environments/execution/k8s/job.py
@@ -39,6 +39,8 @@ class KubernetesJobEnvironment(Environment):
 
     Args:
         - job_spec_file (str, optional): Path to a job spec YAML file
+        - unique_job_name (bool, optional): Use a unique name for each job created with this environment. Defaults
+            to `False`
         - executor_kwargs (dict, optional): a dictionary of kwargs to be passed to
             the executor; defaults to an empty dictionary
         - labels (List[str], optional): a list of labels, which are arbitrary string identifiers used by Prefect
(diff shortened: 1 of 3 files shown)


joshmeek:
^ I don't think that codecov report is correct

src/prefect/environments/execution/k8s/job.py
```diff
@@ -39,6 +39,8 @@ class KubernetesJobEnvironment(Environment):
 
     Args:
         - job_spec_file (str, optional): Path to a job spec YAML file
+        - unique_job_name (bool, optional): Use a unique name for each job created with this environment. Defaults
```
cicdw, line 42:
```suggestion
        - unique_job_name (bool, optional): whether to use a unique name for each job created with this environment. Defaults
```

src/prefect/environments/execution/k8s/job.py
```diff
@@ -207,6 +211,11 @@ def _populate_job_spec_yaml(
         if not yaml_obj.get("metadata"):
             yaml_obj["metadata"] = {}
 
+        if self.unique_job_name:
+            yaml_obj["metadata"][
+                "name"
+            ] = f"{yaml_obj['metadata']['name']}-{flow_run_id[:8]}"
```
cicdw, line 217: Does this introduce any complications in the event that the agent submits the same flow run id (e.g., for a retry) into the same cluster?  Something like "Job with name "BLAH" already exists"?

joshmeek: Ah yeah I'm going to change this to use the more random UUID approach that the k8s agent does for all prefect jobs, great catch

PR #2558, opened 2020-05-13
Try to parse all config values using literal_eval

### LINKED ISSUE #2536: Use literal_eval for parsing configuration values
## Current behavior
Currently, Prefect takes a [hardcoded stab](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/configuration.py#L37) at converting strings to Python objects. We could replace this function with `ast.literal_eval()` and instantly gain a more flexible cast, including the ability to natively handle lists and dicts. 

## Proposed behavior

Currently, setting `PREFECT__X__Y="['a', 'b', 3]"` results in `prefect.config.x.y == "['a', 'b', 3]"`. This means some Prefect objects (thinking of environments, but also including the config itself) must re-parse config options. This is after we already make an attempt to coerce some values to known types. Replacing this with `ast.literal_eval()` would result in `prefect.config.x.y == ['a', 'b', 3]`.

## Example

cicdw:

## What does this PR change?
This PR introduces a better implementation for parsing config values that will handle secrets passed in through env var in a much more natural way.

## Why is this PR important?
Mainly for secrets, but also for configuration consistency.  Closes #2536

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 - Add option to [set secrets on Storage objects](https://docs.prefect.io/orchestration/recipes/third_party_auth.html#declaring-secrets-on-storage) - [#2507](https://github.com/PrefectHQ/prefect/pull/2507)
 - Add reserved [default Secret names](https://docs.prefect.io/orchestration/recipes/third_party_auth.html#list-of-default-secret-names) and formats for working with cloud platforms - [#2507](https://github.com/PrefectHQ/prefect/pull/2507)
 - Add unique naming option to the jobs created by the `KubernetesJobEnvironment` - [#2553](https://github.com/PrefectHQ/prefect/pull/2553)
+- Use `ast.literal_eval` for configuration values - [#2536](https://github.com/PrefectHQ/prefect/issues/2536)
 
 ### Server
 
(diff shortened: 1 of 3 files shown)


src/prefect/configuration.py
```diff
@@ -57,19 +58,10 @@ def string_to_type(val: str) -> Union[bool, int, float, str]:
     elif val.upper() == "FALSE":
         return False
 
-    # int
+    # dicts
```
lauralorenz, line 61:
```suggestion
    # dicts, ints, floats, or any other literal Python syntax
```

the docstring is also out of date, we can probably rip out the whole "Maps" section since it should work with any literal

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 - Add option to [set secrets on Storage objects](https://docs.prefect.io/orchestration/recipes/third_party_auth.html#declaring-secrets-on-storage) - [#2507](https://github.com/PrefectHQ/prefect/pull/2507)
 - Add reserved [default Secret names](https://docs.prefect.io/orchestration/recipes/third_party_auth.html#list-of-default-secret-names) and formats for working with cloud platforms - [#2507](https://github.com/PrefectHQ/prefect/pull/2507)
 - Add unique naming option to the jobs created by the `KubernetesJobEnvironment` - [#2553](https://github.com/PrefectHQ/prefect/pull/2553)
+- Use `ast.literal_eval` for configuration values - [#2536](https://github.com/PrefectHQ/prefect/issues/2536)
```
jlowin, line 20: headline feature IMO but ok

cicdw: 😂

PR #2560, opened 2020-05-14
add volume to pgdb

### LINKED ISSUE #2591: Make docker server fungible
## Use Case
I'd like to be able to keep data saved in the docker postgres database across docker restarts.

At the moment, a persistent docker volume is not created on the host and any data like ETL registrations and previous runs is lost when docker restarts.

## Solution
Either always create a docker volume in the docker-compose.yml, or add an option in the cli to create a docker volume.

## Alternatives
I've considered maybe a snapshot mechanism from within the prefect service but adding the docker volume seems like the easiest approach.

BartekRoszak:

## What does this PR change?
It adds volume to Postgres container to prevent losing flows and runs after server restart.
It adds an option to assign a path to volume in `config.toml`

## Why is this PR important?
It was discussed several times on Slack. People often use a server for learning or development and now they are forced to register flows after server restart. After this change, it won't be necessary.
[slack1](https://prefect-community.slack.com/archives/CL09KU1K7/p1587993111185900?thread_ts=1587993111.185900)
[slack2](https://prefect-community.slack.com/archives/CL09KU1K7/p1589372987372000?thread_ts=1589372987.372000)

Closes #2591


joshmeek:
Thanks for the PR @BartekRoszak! I'll have to dig into this deeper but so far running with this change is throwing errors with my migrations 🤔

BartekRoszak:
Changing the default PGDATA path could be a problem there. I have removed it.

joshmeek:
Looks like there's an issue with grabbing the directory during initdb:
```
postgres_1   | The files belonging to this database system will be owned by user "postgres".
postgres_1   | This user must also own the server process.
postgres_1   |
postgres_1   | The database cluster will be initialized with locale "en_US.utf8".
postgres_1   | The default database encoding has accordingly been set to "UTF8".
postgres_1   | The default text search configuration will be set to "english".
postgres_1   |
postgres_1   | Data page checksums are disabled.
postgres_1   |
postgres_1   | initdb: directory "/var/lib/postgresql/data" exists but is not empty
postgres_1   | If you want to create a new database system, either remove or empty
postgres_1   | the directory "/var/lib/postgresql/data" or run initdb
postgres_1   | with an argument other than "/var/lib/postgresql/data".
```

I thought this was due to me having a version of postgres running prior but that directory didn't exist on my machine. So I tried adjusting how the volumes were defined to something like (truncated):

```yaml
services:
  postgres:
    volumes:
      - db-data:/var/lib/postgresql/data

volumes:
  db-data:
```

And everything worked great. Did you happen to get your volume claim to work with the changes in this PR? Not sure if it's a machine-dependent thing 🤔

BartekRoszak:
According to [pg_issue](https://github.com/docker-library/postgres/issues/263) it looks like there was could be a problem with the default path to which volume was pinned. It should be empty when starting a server for the first time. 
I have changed the default path in config. Could you check if it is ok now?

joshmeek:
Yep this works @BartekRoszak! Do you mind if I make some changes on this branch? I want to put it behind a CLI flag / add some tests.

BartekRoszak:
@joshmeek no problem 👍

changes/pr2560.yaml
```diff
@@ -0,0 +1,5 @@
+enhancement:
```
cicdw, line 1: I believe this should be of type "server"

PR #2571, opened 2020-05-15
Fix type for Dask Security in RemoteDaskEnvironment

joeschmid:
This is a minor change to fix the type of the optional Dask Security parameter in RemoteDaskEnvironment for those using TLS.

## What does this PR change?
Change the type of the security parameter from `Type[Security]` to `Security`.

## Why is this PR important?
Without this change, those using TLS with Dask will get an annoying and incorrect type error.

### DIFF
--- a/src/prefect/environments/execution/dask/remote.py
+++ b/src/prefect/environments/execution/dask/remote.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Type
+from typing import Callable, List
 
 from distributed.security import Security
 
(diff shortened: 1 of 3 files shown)


src/prefect/environments/execution/dask/remote.py
```diff
@@ -49,7 +49,7 @@ class RemoteDaskEnvironment(RemoteEnvironment):
     def __init__(
         self,
         address: str,
-        security: Type[Security] = None,
+        security: Security = None,
```
cicdw, line 52: As you no longer use `typing.Type` the static analysis check is failing with an usused import

PR #2573, opened 2020-05-15
cut built size of server/ container

jameslamb:

## What does this PR change?

This PR proposes two small changes to reduce the built size of the `server/` docker image:

* clearing the `apt` cache with `rm -rf /var/lib/apt/lists/*`
* adding testing code, cache files, and documentation to that image's `.dockerignore`

## Why is this PR important?

If I understand correctly, the [server Dockerfile](https://github.com/PrefectHQ/prefect/blob/master/server/Dockerfile) is used to build an image that others might pull to run the Prefect Core server themselves. Cutting the size of that container reduces the amount of data that has to be transmitted to pull that image, which should make those pulls faster and less expensive.

The changes in this PR reduce the size of the built image by 2MB, tested with:

```shell
docker build \
    -t prefect-server \
    --no-cache \
    --build-arg VERSION=123 \
    -f Dockerfile \
    .
```

### DIFF
--- a/changes/pr2573.yaml
+++ b/changes/pr2573.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Clear cache, remove kubernetes dependencies to cut size of server container - [#2573](https://github.com/PrefectHQ/prefect/pull/2573)"
+
+contributor:
+  - "[James Lamb](https://github.com/jameslamb)"
--- a/server/.dockerignore
+++ b/server/.dockerignore
@@ -32,10 +32,13 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+tests/
 
 # cache
 .mypy_cache/
 .pytest_cache/
+__pycache__/
+*.pyc
 
 # Flask stuff:
 instance/
@@ -54,3 +57,6 @@ instance/
 env/
 venv/
 ENV/
+
+# documentation
+*.md
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -5,24 +5,26 @@
 
 FROM python:3.7
 ARG VERSION
-ENV SERVER_VERSION=${VERSION}
-
-RUN apt-get -qq -y update && apt-get -qq -y install --no-install-recommends --no-install-suggests --allow-unauthenticated \
-    curl \
-    git \
-    sudo && \
-    rm -rf /var/lib/apt/lists/*
-
 ARG PREFECT_VERSION=master
 ENV PREFECT_VERSION=$PREFECT_VERSION
-
-RUN pip install git+https://github.com/PrefectHQ/prefect.git@${PREFECT_VERSION}#egg=prefect
+ENV SERVER_VERSION=${VERSION}
 
 COPY . /prefect-server
 
-RUN \
-    cd /prefect-server \
-    && pip install -e .
+RUN apt-get -qq -y update && \
+    apt-get -qq -y install \
+        --no-install-recommends \
+        --no-install-suggests \
+        --allow-unauthenticated \
+            curl \
+            git \
+            sudo && \
+    pip install \
+        git+https://github.com/PrefectHQ/prefect.git@${PREFECT_VERSION}#egg=prefect && \
+    cd /prefect-server && \
+    pip install -e . && \
+    sudo apt-get remove -y --auto-remove --purge git && \
+    rm -rf /var/lib/apt/lists/*
 
 WORKDIR /prefect-server
 CMD python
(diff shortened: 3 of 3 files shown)


jameslamb:
Another 35ish MB could be cut by running `apt-get remote -y git`, and I think that should be done if `git` is only needed temporarily for this line

```dockerfile
RUN pip install git+https://github.com/PrefectHQ/prefect.git@${PREFECT_VERSION}#egg=prefect[kubernetes]
```

But I don't know enough about the uses of this container to know if removing `git` would have any unintended consequences, so I decided not to include that change.

jameslamb:
I'll need some help understanding the CI failure. I'm unsure how this failure is related to the changes in this PR:

image

cicdw:
Those logging tests are known to be brittle -- I've gone ahead and rewrote them in https://github.com/PrefectHQ/prefect/pull/2586

joshmeek:
> Another 35ish MB could be cut by running `apt-get remove -y git`, and I think that should be done if `git` is only needed temporarily for this line
> 
> ```dockerfile
> RUN pip install git+https://github.com/PrefectHQ/prefect.git@${PREFECT_VERSION}#egg=prefect[kubernetes]
> ```
> 
> But I don't know enough about the uses of this container to know if removing `git` would have any unintended consequences, so I decided not to include that change.

Not entirely sure on this but I know we have also considered this for the base `prefecthq/prefect` image. Is there some versioneer git related stuff @cicdw?

One thing I notice is that we definitely don't need the `kubernetes` dependency in these images so I think that could be removed as well.

jameslamb:
> One thing I notice is that we definitely don't need Prefect's optional kubernetes dependency in these images so I think that could be removed as well.

I can remove that in this PR

jameslamb:
Removing kubernetes dependencies (https://github.com/PrefectHQ/prefect/pull/2573/commits/94b1295a31926466d000510d156b588c943129d8) cut another ~60 MB out of the image

jcrist:
> Is there some versioneer git related stuff @cicdw? I think it's worth it to do the remove after the install and see if everything still functions!

Yeah, git is needed for versioneer, which is only used on the install. No git apis are used elsewhere, so I believe this should be fine.

joshmeek:
@jcrist Awesome, let's remove it as well

jameslamb:
> @jcrist Awesome, let's remove it as well. Maybe this can kick off us also trimming down the base prefecthq/prefect image in a follow up PR

I can remove `git` from this image right now in this PR

jameslamb:
I just removed `git` in https://github.com/PrefectHQ/prefect/pull/2573/commits/9b2d257a72a1cfd83fbf6ddbe5be186d94f970ff.

Installing `server` with `pip install -e .` requires `git` to still be there, because `uvicorn` is installed from GitHub.

So this became a pretty invasive change where I had to re-order the steps. I don't think it would work to remove `git` in a later `RUN` statement, because `git` would still be there in the earlier layer of the image.

If this change is too big, please let me know. I'd be happy to revert it and let you remove `git` in a separate PR.

joshmeek:
Ah I see what you mean. I believe the steps were ordered the previous way to allow for docker's build caching mechanics to take place. I can't say for certain but I think having every command in a single `RUN` after the `COPY` will require all of those installs to run every time the new files from `COPY` land. If git is only 4MB and removing it causes build times to slow down then I would say it's better to keep it in

jameslamb:
> Ah I see what you mean. I believe the steps were ordered the previous way to allow for docker's build caching mechanics to take place. I can't say for certain but I think having every command in a single `RUN` after the `COPY` will require all of those installs to run every time the new files from `COPY` land. If git is only 4MB and removing it causes build times to slow down then I would say it's better to keep it in

You're right that having on `RUN` that includes dynamic stuff like `pip install -e .` means that layer can never be cached. I wasn't sure whether build time was a concern. For example, if only prefect developers build / push this image and users are expected to pull it from a registry like dockerhub, a few extra minutes of build time would be worth the size cut.

I just reverted the `git` change in https://github.com/PrefectHQ/prefect/pull/2573/commits/0f00b9e00b8449691317d69cf3a62042951e772c, so this PR is back to the state as of https://github.com/PrefectHQ/prefect/pull/2573#issuecomment-630209421

joshmeek:
Thanks @jameslamb do you think you could also add a change for this to the https://github.com/PrefectHQ/prefect/tree/master/changes directory under `server`?

jameslamb:
Sure! Added to `changes/` in https://github.com/PrefectHQ/prefect/pull/2573/commits/5ae84f167716204de987cba250a80dfc1b99136a

PR #2574, opened 2020-05-15
prefer sets to lists for membership tests

jameslamb:

## What does this PR change?

For some memberships tests (`in` and `not in`) in Python code, this PR proposes replacing lists and tuples with set objects. 

## Why is this PR important?

For the cases touched in this PR, I think using sets will offer a very tiny speedup relative to the lists and tuples currently used.

From the [Python wiki](https://wiki.python.org/moin/PythonSpeed#Use_the_best_algorithms_and_fastest_tools):

> Membership testing with sets and dictionaries is much faster, O(1), than searching sequences, O(n). When testing "a in b", b should be a set or dictionary instead of a list or tuple.

### DIFF
--- a/docs/generate_docs.py
+++ b/docs/generate_docs.py
@@ -216,7 +216,7 @@ def get_call_signature(obj):
     if args == []:
         standalone, kwargs = [], []
     else:
-        if args[0] in ["cls", "self"]:
+        if args[0] in {"cls", "self"}:
             args = args[1:]  # remove cls or self from displayed signature
 
         standalone = args[: -len(defaults)] if defaults else args  # true args
--- a/examples/tutorial/aircraftlib/analysis.py
+++ b/examples/tutorial/aircraftlib/analysis.py
@@ -21,7 +21,7 @@ FIELS_OF_INTEREST = (
 def clean_vector(raw_vector: List[Any]):
     clean = dict(zip(AIRCRAFT_VECTOR_FIELDS, raw_vector[:]))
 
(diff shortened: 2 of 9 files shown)


cicdw:
Hey @jameslamb ! I super appreciate the diligence here, but because we are checking for containment in such small ad-hoc collections the overhead of building the set actually overshadows the speed up:

```python
%timeit 1 in [1, 2]
# 26.3 ns ± 1.12 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

%timeit 1 in {1, 2}
# 34 ns ± 1.15 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Interestingly, because each of these checks has to 1.) build the set and 2.) then check containment only once, there's no size of collection that would speed this up:

```python
%timeit 1 in list(range(10000))                                                                                                            
# 211 µs ± 26.1 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit 1 in set(range(10000))                                                                                                            
# 375 µs ± 80.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Closing for these reasons.

jameslamb:
oh interesting!  Ok that makes a lot of sense. I guess you'd see the pay-off more in repeated membership tests (where you pay the construction cost one time).

Thanks for the thorough review and sorry for the distraction.

cicdw:
Yup exactly -- and no worries at all!

PR #2570, opened 2020-05-15
Handle possible malformed evals in agent labels parsing

joshmeek:
Now that #2558 allows for providing a more rich set of env vars via config we were attempting to double parse `config.cloud.agent.labels`. 

Closes #2569

PR #2575, opened 2020-05-15
Isolated ast.literal_eval for consistency and clarity

cicdw:

## What does this PR change?
This PR clarifies and finalizes the work of #2558 and #2570 with one well-placed `literal_eval`.  In particular, it ensures that any string values within a user's Prefect config are parsed / converted to the corresponding Python objects.  This is particularly useful for nested structures such as lists of strings (`extra_loggers`, secrets, etc.)

## Why is this PR important?
For easier debugging of our config values and consistency of typing.

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -1,4 +1,3 @@
-import ast
 import functools
 import logging
 import math
@@ -99,10 +98,7 @@ class Agent:
     ) -> None:
         self.name = name or config.cloud.agent.get("name", "agent")
 
-        self.labels = labels or config.cloud.agent.get("labels", [])
-        # quick hack in case config has not been evaluated to a list yet
-        if isinstance(self.labels, str):
-            self.labels = ast.literal_eval(self.labels)
+        self.labels = labels or list(config.cloud.agent.get("labels", []))
         self.env_vars = env_vars or config.cloud.agent.get("env_vars", dict())
         self.max_polls = max_polls
         self.log_to_cloud = False if no_cloud_logs else True
--- a/src/prefect/configuration.py
+++ b/src/prefect/configuration.py
@@ -43,7 +43,7 @@ def string_to_type(val: str) -> Union[bool, int, float, str]:
         - "true" (any capitalization) to `True`
         - "false" (any capitalization) to `False`
         - any other valid literal Python syntax interpretable by ast.literal_eval
-        
+
     Arguments:
         - val (str): the string value of an environment variable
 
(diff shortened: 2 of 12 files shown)


src/prefect/agent/agent.py
```diff
@@ -99,10 +98,7 @@ def __init__(
     ) -> None:
         self.name = name or config.cloud.agent.get("name", "agent")
 
-        self.labels = labels or config.cloud.agent.get("labels", [])
-        # quick hack in case config has not been evaluated to a list yet
-        if isinstance(self.labels, str):
-            self.labels = ast.literal_eval(self.labels)
+        self.labels = labels or list(config.cloud.agent.get("labels", []))
```
cicdw, line 101: I found some failing tests that were being caused by this manipulating the _same_ list with each test run, hence the 'list()` wrapper.

src/prefect/configuration.py
```diff
@@ -229,7 +229,10 @@ def interpolate_config(config: dict, env_var_prefix: str = None) -> Config:
 
     # interpolate any env vars referenced
     for k, v in list(flat_config.items()):
-        flat_config[k] = interpolate_env_vars(v)
+        val = interpolate_env_vars(v)
+        if isinstance(val, str):
+            val = string_to_type(val)
```
cicdw, line 234: This was the key --> if `interpolate_env_vars` returns a string, there's a good chance it's a nested type

jlowin: I was surprised `interpolate_env_vars` wasn't already calling this on the result, and found [this comment](https://github.com/PrefectHQ/prefect/blob/consistent-eval/src/prefect/configuration.py#L85) (which appears to have a bug - we call this on the third pass with `counter > 1`, but the comment says we call it on the second). 

I think now that we're using `literal_eval` this fear is no longer a concern and we could always call `string_to_val(interpolated)` before returning from `interpolate_env_var`, which would guarantee that `interpolate_env_var` always had this expected behavior.

cicdw: Hm ok, should that be called early on (e.g., [right after this line](https://github.com/PrefectHQ/prefect/blob/consistent-eval/src/prefect/configuration.py#L83)) or at the end (e.g., right before [the function returns](https://github.com/PrefectHQ/prefect/blob/consistent-eval/src/prefect/configuration.py#L90)) ?

jlowin: I think right before the function returns, because the interpolation loop will only work on strings. The only real case I can think of where it matters is if you provided a literal list of variables (`'["$A", "$B"]'`), we would want to keep it as an interpretable string for as long as possible. If the first call turned it into `List[str]`, then the substrings would not be expanded further.

jlowin: Discussed IRL; going to clean up the details later and let's go with this implementation!

PR #2576, opened 2020-05-15
Closes #2395. Adds API healthcheck

### LINKED ISSUE #2395: Add UI alert for lack of API communication
## Use Case
It's possible to spin up the UI without access to the API. In these cases, we want to make it clear that the UI isn't empty because there's no data, it's empty because it can't communicate with the API at the given endpoint.

## Solution
An alert on 500 errors. 

## Alternatives
A series of "Getting Started" screens. This doesn't make a lot of sense because if the API is deployed on a remote server, setting up the UI requires restarting the UI docker image after writing to the config the endpoint of the remote API.

lauralorenz:

## What does this PR change?
Closes #2395. Adds a new tile that showcases the status of the `hello` query to Apollo to determine if we are connected to the API layer, and what the current API endpoint is configured as.

## Why is this PR important?
So people will be less confused about whether their UI is connected correctly :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 
 ### Server
 
-- None
+- Add API healthcheck tile to the UI - [#2395](https://github.com/PrefectHQ/prefect/issues/2395)
 
 ### Task Library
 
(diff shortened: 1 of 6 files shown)


lauralorenz:
Woo @znicholasbrown I think the tests are all good now if you want to give a looksee

PR #2581, opened 2020-05-16
Reuse prefect.context for flow context

jlowin:

## What does this PR change?

The `Flow` context manager works by explicitly placing and removing itself into context on enter and exit, in order to make sure it works across various nesting levels. However, this replicates (better-tested) functionality in the `prefect.context` class itself, so it is preferable to reuse `prefect.context` as a context manager rather than as a manually-manipulated object. 

Specifically, while entering a `with Flow:` block modifies the active `prefect.context`, it is NOT equivalent to entering a new `with prefect.context():` block. This can lead to surprising behaviors. For example, modifying context from inside a `with Flow` block will persist once the `Flow` block is exited, even though the flow itself will be removed from context.

This PR tweaks the `Flow` class so it uses a `contextlib.contextmanger`-style generator to reuse Prefect's existing `context` implementation. This reduces complexity because both `__enter__` and `__exit__` are now boilerplate and all application logic is in `Flow._flow_context`.

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -7,6 +7,7 @@ import os
 import tempfile
 import time
 import warnings
+from contextlib import contextmanager
 from pathlib import Path
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


src/prefect/core/flow.py
```diff
@@ -215,6 +217,12 @@ def __repr__(self) -> str:
     def __iter__(self) -> Iterable[Task]:
         yield from self.sorted_tasks()
 
+    def __getstate__(self) -> Dict[str, Any]:
```
cicdw, line 220: This is the only thing that I'm waffling over about this PR --> I 100% understand why you did it / what it solves for, but it feels "heavy" to me for some reason.  Gonna keep thinking on this.

jlowin: Because `flow.__exit__` deletes the generator, this is really just a redundant guard against an unlikely case (pickling when within a flow context). I'll pull it out.

cicdw:
@jlowin just in case you missed it - could you add a test that context changes that occur within a Flow context don't leak out / persist?

jlowin:
Sorry missed it -- I added a test that shows that setting an arbitrary context value inside a flow context is reset when leaving the context.

PR #2584, opened 2020-05-16
Add warning if tasks are created in flow context but not added to the flow

jlowin:

## What does this PR change?
A common stumbling block for new users is that they instantiate a `Task` class without realizing it hasn't been added to their flow. This is especially easy with `Parameters`:

```python
with Flow('example') as flow:
    x = Parameter('x')

flow.run(x=5) # error, unexpected parameter
```

This PR adds a warning when users create tasks that don't get added to the flow:

```
UserWarning: Tasks were created but not added to the flow: 
{<Parameter: x>}. This can occur when `Task` classes, 
including `Parameters`, are instantiated inside a `with flow:` 
block but not added to the flow either explicitly or as the 
input to another task. For more information, see 
https://docs.prefect.io/core/advanced_tutorials/task-guide.html#adding-tasks-to-flows.
```

(This PR is the ulterior motive for #2581, because I needed to use context and guarantee that context semantics were maintained in flow contexts)

## Why is this PR important?
UX!

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -331,8 +331,23 @@ class Flow:
 
     @contextmanager
     def _flow_context(self) -> Iterator["Flow"]:
-        with prefect.context(flow=self):
+        """
+        When entering a flow context, the Prefect context is modified to include:
+            - `flow`: the flow itself
+            - `_new_task_tracker`: a set of all tasks created while the context is 
+                open, in order to provide user friendly warnings if they aren't added
+                to the flow itself. This is purely for user experience.
+        """
+        new_task_tracker = set()
+
+        with prefect.context(flow=self, _new_task_tracker=new_task_tracker):
             yield self
+
+        if new_task_tracker.difference(self.tasks):
+            warnings.warn(
+                "Tasks were created but not added to the flow: "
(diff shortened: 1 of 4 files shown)


src/prefect/core/flow.py
```diff
@@ -325,9 +325,24 @@ def replace(self, old: Task, new: Task, validate: bool = True) -> None:
...
+                to the flow itself. This is purely for user experience.
+        """
+        new_task_tracker = set()  # type: Set[Task]
+
+        with prefect.context(flow=self, _new_task_tracker=new_task_tracker):
             yield self
 
+        if new_task_tracker.difference(self.tasks):
+            warnings.warn(
+                "Tasks were created but not added to the flow: "
```
cicdw, line 342: This is excellent - only question I have is whether we think this message is sufficient, or should we say more?  e.g., link to https://docs.prefect.io/core/advanced_tutorials/task-guide.html#adding-tasks-to-flows and / or saying "This commonly occurs with Parameter tasks that are unused"

jlowin:
If #2590 is accepted, this warning should arguably move into `initialize_task()` so that it tracks copied tasks as well.

PR #2590, opened 2020-05-17
Ensure target locations are updated when tasks are copied

### LINKED ISSUE #2588: Result location not set when passing target with task_args
## Description
When `target` is passed to @task it is set on the `result` in the task.__init__() method.  If the target is passed through `task_args={"target"="a.data"}`. It is not set and the `result` falls back to the default_location.

## Expected Behavior
The location that the `result` uses should be the same regardless of the method used to set the `target`.

## Reproduction
```python
"""
Set in environment.

PREFECT__FLOWS__CHECKPOINTING=true
"""

@task(
    target="decorator.data", result=LocalResult(dir="."),
)
def gen_decorator():
    return "gen-decorator"


@task
def gen_task_args():
    return "gen-task-args"


with Flow("Target") as flow:
    gen_decorator()
    gen_task_args(
        task_args={"target": "task-args.data", "result": LocalResult(dir="."),}
    )

    state = flow.run()
```

## Environment
Prefect 0.11.1

## Discussion
In `Task.__init__()`:

```python
        if target and result:
            self.result = result.copy()  # type: ignore
            self.result.location = target

            if getattr(result, "location", None):
                warnings.warn(
                    "Both `result.location` and `target` set on task. Task result will use target as location."
                )
```

This works fine if target is available through @task.  If it isn't set yet then `location` isn't set.

In `Task.__call__()` the task is copied.

```python
new = self.copy(**(task_args or {}))
```

In `Task.copy(self, **task_args)`. The `target` and `result` are added to the copy of the task, but location is not set on the result.

```python
        # check task_args
        for attr, val in task_args.items():
            if not hasattr(new, attr):
                raise AttributeError(
                    "{0} does not have {1} as an attribute".format(self, attr)
                )
            else:
                setattr(new, attr, val)
```

jlowin:

## What does this PR change?
When tasks are copied, which happens often in the functional API via `Task.bind()`, their attributes are carried over to the new copy. In addition, users have the opportunity to set new attributes via the `task_args` kwarg. However, any initialization logic that takes place in `Task.__init__()` based on the values of those attributes is unavailable to the copy method. This leads to inconsistencies like the one described in #2588. In the issue described in #2588, the task `Result` is modified based on the presence of a `target` attribute. When copying the task, that step doesn't take place.

Closes #2588

## Why is this PR important?
We want to ensure consistent behavior across both the functional and imperative APIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ These changes are available in the [master branch](https://github.com/PrefectHQ/
 ### Enhancements
 
 - Reuse `prefect.context` for opening `Flow` contexts - [#2581](https://github.com/PrefectHQ/prefect/pull/2581)
-- Create `initialize_task()` method for abstracting initialization logic across new task instances and copies - [#2590](https://github.com/PrefectHQ/prefect/pull/2590)
 
 ### Server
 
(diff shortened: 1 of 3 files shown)


jlowin:
@joshmeek hold off on this for one sec please, trying to come up with a cleaner approach

PR #2598, opened 2020-05-18
add unified ORM interface for record CRUD operations

alexisprince1994:

## What does this PR change?
This PR adds CRUD utilities to the ORM for performing single instance operations. The PR adds opinionated convenience class methods aimed at improving code readability with the naming convention of `<operation>_by_id`, in addition to a generic `create` class method. 

## Why is this PR important?
While this PR wouldn't result in _that_ many lines of code being able to be removed, it will improve code readability. When reading through the codebase, there are two specific cases where I could see this PR's changes could be applicable. 

First is aligning the ORM's ability to issue "sql-adjacent" operations. Hasura allows incredibly rich, specific queries against the Postgres database, however not all queries need to leverage this flexibility. Some queries involve very direct, simple queries based on an item's primary key. Adding this API for CRUD operations allows a direct translation of someone's thought of `DELETE FROM <table> WHERE id = ?` to `<Table>.delete_by_id(?)`.

Second is by improving code intention by being able to remove the calls to a model's `where` class method to then either update or delete it. Until you're very familiar with the ORM and the patterns used in Server, the query looks to be selecting data from the database then issuing the update / delete. While this isn't actually the case, it is an extra logical step that requires differentiating the class methods from the instance methods that someone needs to know prior to interacting with the ORM.

An example of both of these in one can be found in several of the `API` files, but a simple example is in the flows API:
```python
async def delete_flow(flow_id: str) -> bool:
    """
    Deletes a flow.

    Args:
        - flow_id (str): the flow id

    Returns:
        - bool: if the delete succeeded

    Raises:
        - ValueError: if a flow ID is not provided
    """
    if not flow_id:
        raise ValueError("Must provide flow ID.")

    # delete the flow
    result = await models.Flow.where(id=flow_id).delete()
    return bool(result.affected_rows)
```

Since this is a straightforward case of needing to update a record in the database, the code can be rewritten to show clearer intention:
```python
async def delete_flow(flow_id: str) -> bool:
    """
    Deletes a flow.

    Args:
        - flow_id (str): the flow id

    Returns:
        - bool: if the delete succeeded

    Raises:
        - ValueError: if a flow ID is not provided
    """
    if not flow_id:
        raise ValueError("Must provide flow ID.")

    # delete the flow
    return await models.Flow.delete_by_id(flow_id)
```


cicdw:

PR #2602, opened 2020-05-18
Dynamic `__doc__`/`__signature__` on tasks

### LINKED ISSUE #2551: Keep original docstring from @task decorated function
## Current behavior
When decorating a function with `@task`, the original docstring of the function is lost.

## Proposed behavior
Original docstring is kept. Looking at the wrapper [`prefect.utilities.tasks.task`](https://github.com/PrefectHQ/prefect/blob/c1686091731144bbc0fab367e78b93221fccf28b/src/prefect/utilities/tasks.py#L152-L202), I thinks this should be do-able with [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps), although I am not exactly sure what change needs to be made.

jcrist:
This is possibly too much magic.

It does two things:

- On tasks created with `@task`, the docstring is now the original
  function docstring. The `__call__` and `__class__` docstrings are
  still the same.
- All tasks dynamically generate a `__signature__` attribute, which
  provides nicer tab completion and signature inspection when calling a
  task directly (e.g. `mytask(param, other_param=1)`).

Mostly putting this up for discussion, the implementation can be cleaned
up for sure.

Fixes #2551.

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -35,6 +35,23 @@ if TYPE_CHECKING:
     from prefect.core import Edge  # pylint: disable=W0611
 
 VAR_KEYWORD = inspect.Parameter.VAR_KEYWORD
+EXTRA_CALL_PARAMETERS = [
+    inspect.Parameter(
+        "mapped", inspect.Parameter.KEYWORD_ONLY, default=False, annotation=bool
+    ),
+    inspect.Parameter(
+        "task_args", inspect.Parameter.KEYWORD_ONLY, default=None, annotation=dict
+    ),
+    inspect.Parameter(
+        "upstream_tasks",
+        inspect.Parameter.KEYWORD_ONLY,
+        default=None,
+        annotation=Iterable[Any],
+    ),
+    inspect.Parameter(
+        "flow", inspect.Parameter.KEYWORD_ONLY, default=None, annotation="Flow",
+    ),
+]
 
 
 def _validate_run_signature(run: Callable) -> None:
@@ -372,6 +389,19 @@ class Task(metaclass=SignatureValidator):
         new.tags.update(tags)
 
         return new
+
+    @property
+    def __signature__(self):
+        """Dynamically generate the signature, replacing ``*args``/``**kwargs``
+        with parameters from ``run``"""
+        if not hasattr(self, "_cached_signature"):
+            sig = inspect.Signature.from_callable(self.run)
+            parameters = list(sig.parameters.values())
+            parameters.extend(EXTRA_CALL_PARAMETERS)
+            self._cached_signature = inspect.Signature(
+                parameters=parameters, return_annotation="Task"
+            )
+        return self._cached_signature
 
     def __call__(
         self,
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -32,23 +32,6 @@ if TYPE_CHECKING:
     from prefect.core import Edge  # pylint: disable=W0611
 
 VAR_KEYWORD = inspect.Parameter.VAR_KEYWORD
-EXTRA_CALL_PARAMETERS = [
-    inspect.Parameter(
-        "mapped", inspect.Parameter.KEYWORD_ONLY, default=False, annotation=bool
-    ),
-    inspect.Parameter(
-        "task_args", inspect.Parameter.KEYWORD_ONLY, default=None, annotation=dict
-    ),
-    inspect.Parameter(
-        "upstream_tasks",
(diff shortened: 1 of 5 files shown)


jcrist:
Demo:

```python
In [13]: @task
    ...: def demo(a, b=1):
    ...:     """A demonstration of possibly too much magic
    ...:
    ...:     Parameters
    ...:     ----------
    ...:     a, b : int
    ...:
    ...:     Returns
    ...:     -------
    ...:     int
    ...:     """
    ...:     return a + b
```

With ipython, this displays:

```text
In [17]: demo?
Signature:
demo(
    a,
    b=1,
    *,
    mapped: bool = False,
    task_args: dict = None,
    upstream_tasks: Iterable[Any] = None,
    flow: 'Flow' = None,
) -> 'Task'
Call signature:
demo(
    *args: Any,
    mapped: bool = False,
    task_args: dict = None,
    upstream_tasks: Iterable[Any] = None,
    flow: 'Flow' = None,
    **kwargs: Any,
) -> 'Task'
Type:            FunctionTask
String form:     <Task: demo>
File:            ~/Code/prefect/src/prefect/tasks/core/function.py
Docstring:
A demonstration of possibly too much magic

Parameters
----------
a, b : int

Returns
-------
int
Class docstring:
A convenience Task for functionally creating Task instances with
...
```

- I think dynamically generating the `__doc__` attribute for `@task` decorated tasks is fine and makes sense. We might insert a prefix on the generated docstring stating that this is a task or something, but otherwise I think this is something worth doing.

- I'm +0 on the `__signature__` bit. `inspect.Signature.from_callable` provides more information now, but since ipython pulls the signature from both the instance and the class, the output is kind of busy in ipython (they might accept a PR fixing this). It also doesn't work for sphinx, which only checks the class for the signature. It *does* work for ipython's tab completion (i.e. you can tab-complete `demo(...)` calls) though, which is nice, but I'm not sure how useful that is for users.

src/prefect/tasks/core/function.py
```diff
@@ -10,9 +10,20 @@
...
+        if obj is None:
+            return self._source
+        else:
+            return obj.run.__doc__
+
+
 class FunctionTask(prefect.Task):
-    """
-    A convenience Task for functionally creating Task instances with
+    __doc__ = _DocProxy(
```
cicdw, line 25: Could you explain this `DocProxy` object?  I haven't seen this before and don't fully get the idea

jcrist: Sure. This is a descriptor object ([this](https://docs.python.org/3/howto/descriptor.html) is a pretty good ref) used to make the `__doc__` attribute different between the class `FunctionTask` and for an instance of `FunctionTask`.

- `FunctionTask.__doc__` is the docstring here
- `FunctionTask(...).__doc__` is the docstring of the wrapped function

We could also assign `self.__doc__` in the constructor, but that will duplicate the docstring in serialization (in a future PR I also want to cleanup how `FunctionTask`s serialize, as currently they aren't picklable for *reasons*). That said, this is mostly leftover from playing around with different locations for docstrings and how ipython and sphinx load those (I tried overriding the `__call__` docstring first, which required a descriptor to handle, the class docstring doesn't).

cicdw: Ah very cool - TIL!

jcrist:
Apologies for letting this sit so long, should be ready for review/merge.

Here's a demo with bpython:

Screen Shot 2020-06-01 at 3 12 17 PM

src/prefect/core/task.py
```diff
@@ -415,6 +415,19 @@ def copy(self, **task_args: Any) -> "Task":
...
+    @property
+    def __signature__(self) -> inspect.Signature:
+        """Dynamically generate the signature, replacing ``*args``/``**kwargs``
+        with parameters from ``run``"""
+        if not hasattr(self, "_cached_signature"):
+            sig = inspect.Signature.from_callable(self.run)
+            parameters = list(sig.parameters.values())
+            parameters.extend(EXTRA_CALL_PARAMETERS)
+            self._cached_signature = inspect.Signature(
+                parameters=parameters, return_annotation="Task"
```
cicdw, line 427: The return annotation confused me for a second - we _might_ want to include a small comment in the docstring for `__signature__` that this is the signature of `__call__` not `run`, although now that I write it out it should be obvious -- you can choose to ignore this comment if you want

jcrist: Yeah, since the `__signature__` is attached to the object, it's the signature for calling that object, so it has to be for `__call__` not `run`.

PR #2608, opened 2020-05-19
Fix parameter first usage with case

### LINKED ISSUE #2604: Parameter must be bound before case context
## Current behavior 0.11.0
`case.__enter__` mimics python if blocks, yet assumes that parameters are bound to the flow outside its context. That can cause some head-scratching for users.
```python
from prefect import task, Parameter, Flow
from prefect.tasks.control_flow import merge, case

with Flow("test maybe first param use") as f:
    x = Parameter("x")
    p = Parameter("p")
    with case(p > 10, True):
        y = x-p
    y = merge(y, p)

state = f.run(parameters=dict(x=0, p=11))
assert state.result[y].result == -11
state = f.run(parameters=dict(x=0, p=9))
assert state.result[y].result == 9
```
**Errors** with:
```
  File "pref_case_bind.py", line 8, in <module>
    y = x-p
  File "~/miniconda3/lib/python3.6/site-packages/prefect/tasks/control_flow/case.py", line 100, in __exit__
    child.set_upstream(cond, flow=self._flow)
  File "~/miniconda3/lib/python3.6/site-packages/prefect/core/task.py", line 591, in set_upstream
    self.set_dependencies(flow=flow, upstream_tasks=[task], mapped=mapped)
  File "~/miniconda3/lib/python3.6/site-packages/prefect/core/task.py", line 566, in set_dependencies
    mapped=mapped,
  File "~/miniconda3/lib/python3.6/site-packages/prefect/core/flow.py", line 811, in set_dependencies
    mapped=is_mapped,
  File "~/miniconda3/lib/python3.6/site-packages/prefect/core/flow.py", line 482, in add_edge
    "Parameters must be root tasks and can not have upstream dependencies."
ValueError: Parameters must be root tasks and can not have upstream dependencies.
```
**Passes** if: `x = Parameter("x")()`

## Proposed behavior
If `child` is a `Parameter` we would `.bind(self_flow)` between L#99 and L#100. I cannot think of any unintended side-effects, because use outside of a case context also binds yet unbound parameters (please confirm though).

## Example
See above

jcrist:
Previously adding a parameter to a flow under a `case` statement would
error if the parameter hadn't been already used (since all new tasks
under the case statment had the case set as an upstream task, and
parameters can't have upstream tasks). This fixes that.

Fixes #2604.

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -469,9 +469,12 @@ class Flow:
             self.tasks.add(task)
             self._cache.clear()
 
-            case = prefect.context.get("case", None)
-            if case is not None:
-                case.add_task(task, self)
+            # Parameters must be root tasks
+            # All other new tasks should be added to the current case (if any)
+            if not isinstance(task, Parameter):
+                case = prefect.context.get("case", None)
+                if case is not None:
+                    case.add_task(task, self)
 
         return task
 
--- a/changes/issue2604.yaml
+++ b/changes/issue2604.yaml
@@ -1,0 +1,2 @@
+fix:
+  - Fix error when adding `Parameter` to flow under `case` statement - [#2608](https://github.com/PrefectHQ/prefect/pull/2608)
--- a/tests/tasks/test_control_flow.py
+++ b/tests/tasks/test_control_flow.py
@@ -1,7 +1,7 @@
 import pytest
 
 import prefect
-from prefect import Flow, Task, task
+from prefect import Flow, Task, task, Parameter
 from prefect.engine.result import NoResult
 from prefect.engine.state import Skipped, Success
 from prefect.tasks.control_flow import FilterTask, ifelse, merge, switch, case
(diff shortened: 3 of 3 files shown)


src/prefect/core/flow.py
```diff
@@ -469,9 +469,12 @@ def add_task(self, task: Task) -> Task:
             self.tasks.add(task)
             self._cache.clear()
 
-            case = prefect.context.get("case", None)
-            if case is not None:
-                case.add_task(task, self)
+            # Parameters must be root tasks
+            # All other new tasks should be added to the current case (if any)
+            if not isinstance(task, Parameter):
```
jcrist, line 474: I put this check here rather than in the `case` class since the check preventing `Parameter` upstream tasks is also handled in the `Flow` class. Made sense to keep things together.

jcrist:
Test failures are unrelated.

PR #2614, opened 2020-05-20
Gsheets

mattalhonte:

## What does this PR change?
Added some Tasks for interacting with Google Sheets via the gspread package.

## Why is this PR important?
The more connectors, the better!


joshmeek:
Thanks for the PR @mattalhonte! Looks like you need to run black formatting over your branch with `black .` at the root directory 🙂

src/prefect/tasks/gsheets/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    from prefect.tasks.gsheets.gsheets import (
+        GsheetUpdates,
+        GsheetResponse,
+        AuthenticateGsheets,
+        WriteGsheetRow,
+        ReadGsheetRow,
+    )
+except ImportError:
+    raise ImportError(
+        'Using `prefect.tasks.gsheets` requires Prefect to be installed with the "gsheets" extra.'
```
cicdw, line 15: Could you add this extra to `setup.py`?

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,160 @@
...
+    Account: https://gspread.readthedocs.io/en/latest/oauth2.html#for-bots-using-service-account
+    Args:
+        - credentials_filename (Union[str, pathlib.Path]): Location of credentials file
+    """
+    def __init__(self, credentials_filename: Union[str, pathlib.Path], **kwargs: Any):
+        self.credentials_filename = credentials_filename
+        super().__init__(**kwargs)
+
+    def run(self) -> gspread.client.Client:
+        return gspread.service_account(filename=self.credentials_filename)
```
cicdw, line 38: Hmmm this won't work in a lot of situations because clients are not serializable objects.  If you want to separate out this logic into a utility function that's perfectly valid but it shouldn't be structured as a task

src/prefect/tasks/gsheets/__init__.py
mattalhonte: Fixed!

src/prefect/tasks/gsheets/gsheets.py
mattalhonte: K, I replaced it with direct calls to the Client Creator.  Hopefully this won't be cached or anything?  Someone told me a Secret Task was the best way to make sure that didn't happen.

cicdw: So there are a few things here: the only objects that are at risk of being cached are explicit task inputs and outputs.  In the special case that one of your task's upstream inputs is actually a `Secret` task, Prefect will ensure that that particular input is not cached / stored and is instead always recomputed.  

In the original code you had, you weren't using the Auth task as a `Task`, but were instead calling its `run` method directly within another task which is equivalent to the new code that you have here but less explicit.

Hope that explanation helps!

src/prefect/tasks/gsheets/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    from prefect.tasks.gsheets.gsheets import (
+        GsheetUpdates,
+        GsheetResponse,
+        WriteGsheetRow,
+        ReadGsheetRow,
+        gsheet_helper,
+    )
+except ImportError:
+    raise ImportError(
+        'Using `prefect.tasks.gsheets` requires Prefect to be installed with the "gsheets" extra.'
```
cicdw, line 14: You named the extra `"gspread"` so we should update this error message accordingly

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,152 @@
+from prefect.utilities.tasks import defaults_from_attrs
+import gspread
+from typing import Dict, List, Tuple, Union, Any, TypedDict, Callable
+import prefect
+from prefect import task, Flow, Parameter, Task
+from prefect.tasks.secrets.base import SecretBase
+from prefect.tasks.secrets import PrefectSecret
+import pathlib
+from pathlib import Path
```
cicdw, line 9: Looks like there are some unused imports in here that our static analysis job is failing on; it's suggesting:
```
-from typing import Dict, List, Tuple, Union, Any, TypedDict, Callable

-import prefect

-from prefect import task, Flow, Parameter, Task

-from prefect.tasks.secrets.base import SecretBase

-from prefect.tasks.secrets import PrefectSecret

+from typing import Any, Callable, List, TypedDict, Union

+from prefect import Task, task
```

src/prefect/tasks/gsheets/__init__.py
mattalhonte: Fixed!  Changed the extra's name to `gsheets`

src/prefect/tasks/gsheets/gsheets.py
mattalhonte: Removed!

src/prefect/tasks/gsheets/gsheets.py
mattalhonte: Awesome, thanks!  Cool learning Prefect's internals :)

src/prefect/tasks/gsheets/__init__.py
mattalhonte: Not sure what to do with the changelog, though?  I don't see a `changes/` directory.

mattalhonte:
Not sure what to do with the changelog, though? I don't see a changes/ directory.

jcrist:
It's likely that the version of master you branched off of is older than the addition of the `changes/` directory (we recently changed how we did changelogs). If you update this branch with master you should see it.

mattalhonte:
Aha!  Thanks!

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,148 @@
...
+    spreadsheetId: str
+    tableRange: str
+    updates: GsheetUpdates
+
+
+class WriteGsheetRow(Task):
+    """
+    A task for writing a row to a Google Sheet.
+    Note that _all_ initialization settings can be provided / overwritten at runtime.
+    Args:
```
jcrist, line 30: Could you add a blank line before every `Args:` and `Returns:` in the docstrings (here and all the ones below)? The docstring format should be:

```
Description

Args:
    - ...

Returns:
    - ...
```

mattalhonte: Added!

mattalhonte:
Think I made the needed changes!

joshmeek:
@mattalhonte Could you also add an entry for this task to the API doc outline like: 

https://github.com/PrefectHQ/prefect/blob/629a7e0631b73f93eb9258a61b727dc89bf7a5fb/docs/outline.toml#L289-L300

Would look something like:
[pages.tasks.gcp]
title = "Google Sheets Tasks"
module = "prefect.tasks.gsheets"
classes = [
        "GsheetUpdates",
        "GsheetResponse",
        "WriteGsheetRow",
        "ReadGsheetRow",
        ]

mattalhonte:
Done!  :)

joshmeek:
@mattalhonte It looks like you didn't push the commit 😅

mattalhonte:
oh whoops lol.  Pushed!

mattalhonte:
Not sure why it's failing?  btw someone in the Slack mentioned that someone else would handle the Tests

joshmeek:
Thanks @mattalhonte will take a look. Weird error:
```
_________________ ERROR collecting docs/test_generate_docs.py __________________
 
docs/test_generate_docs.py:374: in <module>
 
    for page in OUTLINE
 
docs/test_generate_docs.py:376: in <listcomp>
 
    for fn in get_class_methods(obj)
 
docs/generate_docs.py:299: in get_class_methods
 
    obj, predicate=lambda x: inspect.isroutine(x) and obj.__name__ in x.__qualname__
 
/usr/local/lib/python3.7/inspect.py:354: in getmembers
 
    if not predicate or predicate(value):
 
docs/generate_docs.py:299: in <lambda>
 
    obj, predicate=lambda x: inspect.isroutine(x) and obj.__name__ in x.__qualname__
 
/usr/local/lib/python3.7/unittest/mock.py:600: in __getattr__
 
    raise AttributeError(name)
 
E   AttributeError: __name__
 
```

joshmeek:
Looks like the issue resolves around attempting to generate docs for the `GsheetUpdates` and `GsheetResponse`. @mattalhonte are these only needed for typing purposes and could they be replaces by the output types of `worksheet.append_row` and `worksheet.row_values`?

mattalhonte:
Yeah, they're only for Type Annotations - I could just replace them with
regular Dictionaries if that works better.

mattalhonte:
Just removed the TypedDicts

joshmeek:
Lgtm @cicdw @jcrist pending your approval since you have changes requested that are blocking merge

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,138 @@
...
+def gsheet_helper(fn: Callable):
+    """
+    A 'factory' to make tasks out of various utility methods in Google Sheets that that `gspread`
+    package has. Docs here: https://gspread.readthedocs.io/en/latest/user-guide.html
+
+    Args:
+        - credentials_filename (Union[str, pathlib.Path]): Location of credentials file
+        - sheet_key (str): The key corresponding to the Google Sheet
+        - worksheet_name (str): The worksheet to target
+        - **kwargs (optional): additional kwargs to pass to the `Task` constructor
```
jcrist, line 117: The `Args` field in the docstring here should be documenting `gsheet_helper`, which only takes `fn` (a callable). The `args` here could move to the docstring for `inner` (looks like you just have the contents swapped).

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,138 @@
...
+        Returns:
+            - a list of values from the row 
+        """
+        client = gspread.service_account(filename=credentials_filename)
+        google_sheet = client.open_by_key(sheet_key)
+        worksheet = google_sheet.worksheet(worksheet_name)
+        return worksheet.row_values(row)
+
+
+def gsheet_helper(fn: Callable):
```
jcrist, line 108: Is this helper necessary? It creates some helper tasks, but instills certain signature decisions on them that others may not want. Since the content of `inner` is rather short, I wonder if leaving these up to users would be better? No strong thoughts.

joshmeek: The `ghsheet_helper` was in the outline.toml but removed because it was breaking the doc generation https://app.circleci.com/pipelines/github/PrefectHQ/prefect/4487/workflows/9f7287d9-2140-4551-8521-a96028da3868/jobs/83420/steps

src/prefect/tasks/gsheets/gsheets.py
mattalhonte: Whoops!  Fixed.  And removed the `kwargs` argument that I accidentally left there after copying it from a different Task that I used as a template.

src/prefect/tasks/gsheets/gsheets.py
mattalhonte: Right on!  I figured it'd be useful to to have a "wrapper" that let you turn anything you can do with the `gsheets` package into a Prefect Task (sorta like having a generic "SQL Query" task), but I'm not sure if they way I did it interacts well with the rest of the ecosystem.

jcrist: My gut says we should leave this out - users can write their own `@task` decorated tasks for handling this kind of thing, and may want more flexibility in task format than this provides. For example, they might want to manage clients differently, rather than creating a new one per task.

I also don't find:

```python
find_dough = gsheet_helper(lambda x: x.find("Dough"))

res = find_dough(credentials_filename, sheet_key, worksheet_name)
```

much more readable than

```python
@task
def find_dough(credentials_filename, sheet_key, worksheet_name):
    client = gspread.service_account(filename=credentials_filename)
    google_sheet = client.open_by_key(sheet_key)
    worksheet = google_sheet.worksheet(worksheet_name)
    return worksheet.find("Dough")

res = find_dough(credentials_filename, sheet_key, worksheet_name)
```

just my opinion though. cc @cicdw, @joshmeek for their thoughts.

jcrist: Besides this thread here, the rest of this PR looks good to me. Thanks for all your work here @mattalhonte!

cicdw: Yea I vote remove it - in general I think a "task factory" like this shouldn't ever be necessary because Prefect Tasks can be configured through both init kwargs + runtime data

mattalhonte: Roger!

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,105 @@
...
+            - sheet_key (str): The key corresponding to the Google Sheet
+            - worksheet_name (str): The worksheet to target
+
+        Returns:
+            - a list of values from the row 
+        """
+        client = gspread.service_account(filename=credentials_filename)
+        google_sheet = client.open_by_key(sheet_key)
+        worksheet = google_sheet.worksheet(worksheet_name)
+        return worksheet.row_values(row)
```
jcrist, line 105: You're missing a trailing newline (your text editor probably has a setting to prevent this from happening).

mattalhonte: Reran Black, not sure if that fixed the formatting issue?

src/prefect/tasks/gsheets/gsheets.py
```diff
@@ -0,0 +1,105 @@
+from prefect.utilities.tasks import defaults_from_attrs
+import gspread
+from typing import Any, Callable, List, Union
+from prefect import task, Task
```
jcrist, line 4: Per the [static analysis tests](https://app.circleci.com/pipelines/github/PrefectHQ/prefect/4688/workflows/9d20cd79-99a1-46bd-98c9-66d37163edc1/jobs/85669/steps) , you have a few extraneous imports.

```suggestion
from typing import Any, List, Union
from prefect import Task
```

src/prefect/tasks/gsheets/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+A collection of tasks for interacting with Google Sheets.
+"""
+try:
+    from prefect.tasks.gsheets.gsheets import (
+        WriteGsheetRow,
+        ReadGsheetRow,
+        gsheet_helper,
```
jcrist, line 8: `gsheet_helper` no longer exists

mattalhonte:
Thanks for all your help getting this over the fence!

PR #2567, opened 2020-05-14
[ci] fix masked tests

jameslamb:
**Thanks for contributing to Prefect!**

Please describe your work and make sure your PR:

- [x] adds new tests (if appropriate)
- [x] updates `CHANGELOG.md` (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

## What does this PR change?

Thanks for this awesome project! I've been getting familiar with `prefect` this week, and wanted to contribute back where I could. I noticed you aren't running `flake8` in this project, so I ran that over the codebase to see if there were any issues that could be fixed.

I saw one pattern that I think it would be helpful to address...duplicated tests.

If you run this command from the root of the repo

```shell
flake8 | grep -E "F811 .*unused .*test_"
```

you'll see the following:

```text
./tests/engine/test_flow_runner.py:1187:1: F811 redefinition of unused 'test_parameters_can_be_set_in_context_if_none_passed' from line 1160
./tests/engine/test_flow_runner.py:1194:1: F811 redefinition of unused 'test_parameters_overwrite_context' from line 1167
./tests/engine/test_flow_runner.py:1203:1: F811 redefinition of unused 'test_parameters_overwrite_context_only_if_key_matches' from line 1176
./tests/cli/test_auth.py:133:1: F811 redefinition of unused 'test_switch_tenants' from line 123
./tests/cli/test_execute.py:57:1: F811 redefinition of unused 'test_execute_cloud_flow_fails' from line 24
./tests/agent/test_fargate_agent.py:569:1: F811 redefinition of unused 'test_deploy_flow_raises' from line 545
./tests/tasks/dropbox/test_dropbox.py:39:5: F811 redefinition of unused 'test_creds_are_pulled_from_secret_at_runtime' from line 28
./tests/tasks/azureml/test_datastore.py:191:5: F811 redefinition of unused 'test_missing_datastore_raises_error' from line 184
./tests/client/test_client_auth.py:121:5: F811 redefinition of unused 'test_load_local_api_token_is_called_when_the_client_is_initialized_without_token' from line 110
./tests/core/test_flow.py:1906:5: F811 redefinition of unused 'test_flow_dot_run_handles_cached_states_across_runs' from line 1815
./server/tests/database/test_hasura.py:837:5: F811 redefinition of unused 'test_execute_respects_as_box' from line 830
```

In this PR, I tried to address those warnings by fixing those test files. I saw three flavors of problem:

[...]

PR #2624, opened 2020-05-21
remove unused imports

jameslamb:

## What does this PR change?

Thanks for this great project! I've been getting familiar with `prefect` and really liking it, so I wanted to contribute back. Similar to #2567 , this PR proposes fixing another class of problem I saw running `flake8` on this repo: unused imports.

The "unused imports" detection in `flake8` can lead to a lot of false positives, for example in `__init__.py` files, so I tried to limit the changes I made here to things I'm fairly confident can be removed. You can generate the list of things I changed with this command:

```shell
flake8 \
    --ignore=E501,E231,E503,E266,W503 \
    | grep "F401" \
    | grep -v "__init" \
    | grep -v "'prefect" \
    | grep -v "'pytest'" \
    | grep -v "'asyncio" \
    | grep -v "sqlalchemy" \
    | grep -v "alembic" \
    | grep -v "'docker" \
    | grep -v "'azure\.stoorage" \
    | grep -v "'google\.cloud" \
    | grep -v "'graphviz" \
    | grep -v "\.fixtures\.database_fixtures" \
    | grep -v "'blig" \
    > flake.txt
```

Most of those ignored things are from these common false positive cases:

* things in `__init__.py` files
* things that need to be defined because they're mocked or patched
* imports used to test the availability of a library

I also avoided `sqlalchemy` and `alembic` just because I don't understand the full magic of `sqlalchemy` and didn't know if some of the things `flake8` turned about were actually unused.

## Why is this PR important?

Imports add to the time it takes to load modules.

Unused imports can also make it seem like some dependencies are more widely used in the project than they really are, which can prevent maintainers from noticing limited-use dependencies and trying to factor them out.

To be clear...this PR removes _imports_, but it doesn't change the dependencies for `prefect` at all. All of the things I'm proposing removing come from libraries that are imported elsewhere in the project.

### DIFF
--- a/docs/generate_docs.py
+++ b/docs/generate_docs.py
@@ -17,9 +17,7 @@ import inspect
 import os
 import re
 import shutil
-import subprocess
 import textwrap
-import warnings
 from contextlib import contextmanager
 from functools import partial
 from unittest.mock import MagicMock
--- a/docs/test_generate_docs.py
+++ b/docs/test_generate_docs.py
@@ -1,4 +1,3 @@
-import inspect
 import re
 import sys
 import textwrap
@@ -22,7 +21,6 @@ try:
         format_subheader,
         get_call_signature,
         get_class_methods,
-        get_source,
         patch_imports,
     )
 
--- a/docs/tokenizer.py
+++ b/docs/tokenizer.py
@@ -1,4 +1,3 @@
-from collections import namedtuple
 
 from pygments.lexers import Python3Lexer
 from pygments.token import Comment, Keyword, Name, Number, Operator, Punctuation, String
--- a/docs/tokenizer.py
+++ b/docs/tokenizer.py
@@ -1,4 +1,3 @@
-
 from pygments.lexers import Python3Lexer
 from pygments.token import Comment, Keyword, Name, Number, Operator, Punctuation, String
 
--- a/server/setup.py
+++ b/server/setup.py
@@ -1,7 +1,6 @@
 # Licensed under the Prefect Community License, available at
 # https://www.prefect.io/legal/prefect-community-license
 
-import sys
 
 from setuptools import find_packages, setup
 
--- a/server/src/prefect_server/api/flows.py
+++ b/server/src/prefect_server/api/flows.py
@@ -6,8 +6,6 @@ import asyncio
 import copy
 import uuid
 
-import pendulum
-
 import prefect
 from prefect.engine.state import Finished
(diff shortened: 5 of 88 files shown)


joshmeek:
Looks like it may have removed too much 😅 
```
_________________ ERROR collecting docs/test_generate_docs.py __________________
docs/test_generate_docs.py:9: in <module>
    from prefect import Task, task
src/prefect/__init__.py:8: in <module>
    import prefect.triggers
src/prefect/triggers.py:48: in <module>
    from prefect.engine import signals
src/prefect/engine/__init__.py:9: in <module>
    from prefect.engine.task_runner import TaskRunner
src/prefect/engine/task_runner.py:55: in <module>
    class TaskRunner(Runner):
src/prefect/engine/task_runner.py:615: in TaskRunner
    ) -> Tuple[State, Dict[Edge, State]]:
E   NameError: name 'Tuple' is not defined
```

jameslamb:
haha ah! That's what I get for not running the tests one more time before pushing.

I just fixed that in https://github.com/PrefectHQ/prefect/pull/2624/commits/f5b5b7a2b131ea25412ce03b7048858c7a5e8bc4. Ran the tests locally and all but two passed. From the errors at first glance it seemed unrelated to these changes, and I'm hoping it's just one of those "my laptop looks different from the CI environment" things.

Let's see if I'm right 🤞

jameslamb:
> Let's see if I'm right

I was not haha. Ok sorry, you can ignore this PR until I get CI working. Clearly there are some differences between the way I was running the tests locally and the way they're run in CI.

cicdw:
We do run an [unused import check circle job](https://github.com/PrefectHQ/prefect/blob/master/.circleci/config.yml#L91-L98) so I'll be surprised if you find anything new

jcrist:
> We do run an unused import check circle job so I'll be surprised if you find anything new

This is only run on the `src/prefect` codebase, and ignores both server and the tests.

jameslamb:
> > We do run an unused import check circle job so I'll be surprised if you find anything new
> 
> This is only run on the `src/prefect` codebase, and ignores both server and the tests.

that makes sense! The only unused imports I picked up with `flake8` in `prefect/` were type hints, so it's possible the code you're using in that test doesn't know how to reason about type hints.

jameslamb:
Ok this is passing CI and ready for a review! Thanks for your time.

jcrist:
Thanks @jameslamb!

PR #2625, opened 2020-05-21
Closes #2489. GreatExpectations Task

### LINKED ISSUE #2489: Great Expectations `WarningAndFailureExpectationSuitesValidationOperator` task library task
## Use Case
Some Prefect users use Great Expectations to validate the results of their ETLs in an ad-hoc manner, and we would like to include the ability to run those same validations as a task library task for inclusion in a flow. 

## Solution
The Great Expectations docs [give some general instructions on integrating GE into a pipeline](https://docs.greatexpectations.io/en/latest/getting_started/typical_workflow.html#embedding-automated-testing-into-a-data-pipeline) which centers around using their validation operator objects. 
 This issue is specifically to wrap the [great_expectations.validation_operators.WarningAndFailureExpectationSuitesValidationOperator](https://docs.greatexpectations.io/en/latest/module_docs/validation_operators_module.html#great_expectations.validation_operators.WarningAndFailureExpectationSuitesValidationOperator) as a Prefect task.

Some things to consider based on my first read through
- expectations for validation operators are normally organized into 'suites', which are JSON files on disk, which is a little weird for us in Prefect since we support distributed execution. Probably need to find a way to get users to describe that configuration in Prefect config, or grab it from a cloud store
- most expectation methods only work on tabular data, and in particular pandas dataframes; a few work on raw files when the expectation is general enough to be assertable without knowing the format of the data in the file. See [backend support per expectation in their docs here](https://docs.greatexpectations.io/en/latest/reference/implemented_expectations.html#implemented-expectations)). I think we can expect users who want the GE task to have to send their data in these formats, but just calling that out
- validation operators also provide a way for GE to load the data itself, but we would probably expect the data to already be available in-memory as a Prefect task input, so we either need to bypass this or find a way for this configuration to play nice with what Prefect knows

## Alternatives
Alternatively, we could bypass the validation operators abstraction, since they are a bit heavier in terms of being built around local configuration files and also assuming that the 
GE Validation Operator code itself will be responsible for running "actions" (whereas we would probably want Prefect state handlers to abstract that). I think this would mean we would expose a kwarg on the Prefect task for all the expectations methods to accept, and call those on the user's behalf in the task's run method on any task inputs. Though simpler, this is not their recommended approach in their docs so I think pursuing the above first is worth it, since it takes advantage of existing improvements in GE regarding data sourcing and expectation suite segregation, but if it ends up too heavy we may still consider this alternative.

lauralorenz:

## What does this PR change?
Closes #2489. Creates a task library task that can run a great expectations checkpoints.

Originally I was trying to do something more fancy against remote validation operators config and prefect inputs, but after some hardships there and also after working with the great expectations Getting Started more I went with their checkpoints mechanism: "In normal usage, the best way to validate data is with a Checkpoint." (see step 5 of [Great Expectations' getting started docs](https://docs.greatexpectations.io/en/latest/tutorials/getting_started.html)).

You need to do quite a bit of great expectations setup to have a checkpoint to run (basically all the steps in that same [getting started](https://docs.greatexpectations.io/en/latest/tutorials/getting_started.html) linked above), but you can use the test flow in the examples directory added here instead of doing so on your own -- that example directory has all the local config I would expect a Great Expectations user to have for their checkpoints. It is also used by the automated tests to simulate the local config we expect someone to have in the test environment.

## Why is this PR important?
First flare on an integration with Great Expectations.


src/prefect/tasks/great_expectations/checkpoints.py
```diff
@@ -0,0 +1,100 @@
...
+            batch_kwargs = batch["batch_kwargs"]
+            for suite_name in batch["expectation_suite_names"]:
+                suite = context.get_expectation_suite(suite_name)
+                batch = context.get_batch(batch_kwargs, suite)
+                batches_to_validate.append(batch)
+
+        results = context.run_validation_operator(
+            checkpoint["validation_operator_name"],
+            assets_to_validate=batches_to_validate,
+            run_id={"run_name": prefect.context.get("task_slug")},
```
lauralorenz, line 94: FYI you may be interested that this affects where validation results are stored (assuming the chosen validation operator does that), which affects the data docs UI like below in the `run_name` column. Open to other defaults than task slug, and with it being configurable -- should we get some context templating a la from targets?:

Screen Shot 2020-06-03 at 6 23 06 PM

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.mark.skipif("sys.version_info <= (3,8)")
+class TestIntegrationWithLocalConfig:
+    """
+    Integration test to check the task library task still integrates with how Great Expectations checkpoints are configured.
+    This only runs on 3.8 since it uses a shutil kwarg (dirs_exist_ok) only available in 3.8+ for the purposes of the test.
+    """
+
+    def test_runs_with_different_context_root_dir(self, example_context_root_dir):
+        t = RunGreatExpectationsCheckpoint(
+            "warning_npi_checkpoint_circle", context_root_dir=example_context_root_dir
```
lauralorenz, line 64: I'm not proud of all this but checkpoints don't support interpolation and after a couple iterations I went with this, duplicating the checkpoint file with the difference in relative pathing observed when running tests... just FYI that was the motivation for all of this.

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
```diff
@@ -0,0 +1,78 @@
...
+            "examples/task_library/great_expectations/great_expectations/"
+        )
+        shutil.copytree(examples_path, tmp, dirs_exist_ok=True)
+        yield tmp
+
+
+@pytest.mark.skipif("sys.version_info <= (3,8)")
+class TestIntegrationWithLocalConfig:
+    """
+    Integration test to check the task library task still integrates with how Great Expectations checkpoints are configured.
```
lauralorenz, line 58: This comment explains my reasoning I think; let me know if you disagree with the fundamental ideal here that the integration test, checking that Great Expectations still acts the way we think it does, can be run on 3.8 only.

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
```diff
@@ -0,0 +1,78 @@
...
+
+        assert result["success"]
+
+    def test_raises_if_validation_fails(self, example_context_root_dir):
+        t = RunGreatExpectationsCheckpoint(
+            "guaranteed_failure_npi_checkpoint_circle",
+            context_root_dir=example_context_root_dir,
+        )
+
+        with pytest.raises(prefect.engine.signals.VALIDATIONFAIL):
```
lauralorenz, line 77: Flexing the validation fail state here, unless we don't want that to pop out yet and want it to be failed

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
jlowin: Makes sense to me! No problem with only running on 3.8

src/prefect/tasks/great_expectations/checkpoints.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+import prefect
+from prefect import Task
+from prefect.engine import signals
+from prefect.utilities.tasks import defaults_from_attrs
+
+try:
+    import great_expectations as ge
+except ImportError:
+    pass
```
jlowin, line 16: There's an opportunity to raise a warning / error that this module requires `pip install "prefect[ge]"`, this is my only real note for you!

jlowin: (because I dont think "ge" is documented anywhere else as the extra name)

lauralorenz: Fixed in 5ab324dc335b216e4815ac84a884901442220319, also switched to the style of some of the other ones which raise the importerror it in `__init__.py` instead

setup.py
```diff
@@ -39,6 +39,7 @@ def run(self):
     "dask_cloudprovider": ["dask_cloudprovider >= 0.2.0, < 1.0"],
     "dev": dev_requires + test_requires,
     "dropbox": ["dropbox ~= 9.0"],
+    "ge": ["great_expectations >= 0.11.0"],
```
Aylr, line 42: GE 0.11.0 had a small regression in it so I'd recommend 0.11.1+
```suggestion
    "ge": ["great_expectations >= 0.11.1"],
```

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
Aylr: To me this seems like it might be a missing feature in GE - would you mind opening an issue on our repo?

examples/task_library/great_expectations/great_expectations/notebooks/pandas/validation_playground.ipynb
```diff
@@ -0,0 +1,239 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
```
Aylr, line 4: These notebooks are useful mostly for development and you probably don't need them to be here in Prefect.

examples/task_library/great_expectations/great_expectations_flow.py
```diff
@@ -0,0 +1,13 @@
+from prefect import Flow, Parameter
+from prefect.tasks.great_expectations import RunGreatExpectationsCheckpoint
+
+ge_task = RunGreatExpectationsCheckpoint()
+
+with Flow("great expectations example flow") as flow:
+    checkpoint_name = Parameter("checkpoint_name")
+    validations = ge_task.map(checkpoint_name)
```
Aylr, line 8: This makes me very excited.

examples/task_library/great_expectations/great_expectations/notebooks/pandas/validation_playground.ipynb
lauralorenz: Ooo ty good point, fixed in https://github.com/PrefectHQ/prefect/pull/2625/commits/81ef8d8af276094190f7871a3ebc63a0439974c8!

tests/tasks/great_expectations/test_great_expectations_checkpoints.py
lauralorenz: :+1: totally, opened at https://github.com/great-expectations/great_expectations/issues/1565

PR #2629, opened 2020-05-21
Add TLS configuration option to Docker storage

### LINKED ISSUE #2626: Expose TLS config on Docker storage

**aseev.arsenii**: Hey all! I'm trying to set up a CircleCI job that builds and pushes flows Docker images to AWS ECR. CircleCI provides a remote docker execution environment, and to connect to it I need to pass on TLS configuration to the docker client. Unfortunately Prefect seems to only expose the `base_image` of `docker.APIClient` (<https://github.com/PrefectHQ/prefect/blob/1447982cc63a1f87c4a19cc7be18f2b2693fb883/src/prefect/environments/storage/docker.py#L508>). Any suggestions on how to do this? Thanks!

**laura**: Hi <@UR914ST7H>, I’m not sure of a workaround per se but it looks like the signature of exactly where you found the client instantiation could be changed to accept TLS config, configured by the user against an attribute of the `Docker` Storage instance like you can for `base_url` now. Let me get the bot to open an issue off this thread.

**laura**: <@ULVA73B9P> open “Expose TLS config on Docker storage”

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1590055135302100?thread_ts=1590055135.302100&cid=CL09KU1K7).

joshmeek:

## What does this PR change?
Closes #2626 

## Why is this PR important?
Now a TLS config can be passed directly to Docker storage (https://docker-py.readthedocs.io/en/stable/tls.html#docker.tls.TLSConfig). This is very useful for CI related processes that build flows

### DIFF
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -9,7 +9,7 @@ import textwrap
 import uuid
 import warnings
 from pathlib import PurePosixPath
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Union
 
 import cloudpickle
 import pendulum
@@ -69,7 +69,7 @@ class Docker(Storage):
             for each flow run.  Used primarily for providing authentication credentials.
         - base_url: (str, optional): a URL of a Docker daemon to use when for
             Docker related functionality.  Defaults to DOCKER_HOST env var if not set
-        - tls_config: (docker.tls.TLSConfig, optional): a TLS configuration to pass to the Docker
+        - tls_config: (Union[bool, docker.tls.TLSConfig], optional): a TLS configuration to pass to the Docker
             client. https://docker-py.readthedocs.io/en/stable/tls.html#docker.tls.TLSConfig
 
     Raises:
@@ -92,7 +92,7 @@ class Docker(Storage):
         ignore_healthchecks: bool = False,
         secrets: List[str] = None,
         base_url: str = None,
-        tls_config: "docker.tls.TLSConfig" = None,
+        tls_config: Union[bool, "docker.tls.TLSConfig"] = False,
     ) -> None:
         self.registry_url = registry_url
         if sys.platform == "win32":
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -56,8 +56,6 @@ class Docker(Storage):
             use when building
         - files (dict, optional): a dictionary of files to copy into the image
             when building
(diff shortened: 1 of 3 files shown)


src/prefect/environments/storage/docker.py
```diff
@@ -85,11 +87,12 @@ def __init__(
         image_tag: str = None,
         env_vars: dict = None,
         files: dict = None,
-        base_url: str = None,
         prefect_version: str = None,
         local_image: bool = False,
         ignore_healthchecks: bool = False,
         secrets: List[str] = None,
+        base_url: str = None,
+        tls_config: "docker.tls.TLSConfig" = None,
```
lauralorenz, line 95: should this default to False to match the underlying API call's type and default?

joshmeek: It's kind of a funny kwarg because it can both be `bool` and `TLSConfig`. In my testing setting `TLSConfig` as None has the same effect as using `False` so I would prefer to not mix the two

joshmeek: Actually disregard. In order to match expectations from the Docker library I am going to match on both types

src/prefect/environments/storage/docker.py
```diff
@@ -85,11 +87,12 @@ def __init__(
         image_tag: str = None,
         env_vars: dict = None,
         files: dict = None,
-        base_url: str = None,
         prefect_version: str = None,
         local_image: bool = False,
         ignore_healthchecks: bool = False,
         secrets: List[str] = None,
+        base_url: str = None,
+        tls_config: Union[bool, "docker.tls.TLSConfig"] = False,
```
cicdw, line 95: Do you know if objects of type `TLSConfig` are pickeable?

joshmeek: Just attempted a roundtrip cloudpickle on a `TLSConfig` object and it appears to work as expected:
```
In [6]: cloudpickle.dumps(t)
Out[6]: b'\x80\x04\x95\x8b\x00\x00\x00\x00\x00\x00\x00\x8c\ndocker.tls\x94\x8c\tTLSConfig\x94\x93\x94)\x81\x94}\x94(\x8c\x0fassert_hostname\x94N\x8c\x12assert_fingerprint\x94N\x8c\x0bssl_version\x94\x8c\x03ssl\x94\x8c\x10PROTOCOL_TLSv1_2\x94\x93\x94\x8c\x06verify\x94N\x8c\x07ca_cert\x94Nub.'

In [7]: a = _

In [8]: cloudpickle.loads(a)
Out[8]: <docker.tls.TLSConfig at 0x109266588>
```

PR #2631, opened 2020-05-21
Use host IP for `hostname` label in cases where `LocalAgent` is in container using host network

### LINKED ISSUE #2618: Unique hostname label is `docker-desktop` for Docker "host" networking type
## Current behavior
The Local Agent (and Flow registration) typically generate a unique string (based on a call to `socket.gethostname()`) that gets used as a label to avoid collisions when multiple people are running Local Agents within an organization. If the Local Agent or Flow registration is running in a Docker container with the networking type set to "host" this unique string ends up as `"docker-desktop"`, negating the benefit of using this as a label that is unique to this particular Local Agent.

## Proposed behavior
If the hostname is detected as `"docker-desktop"`, generate a string of random characters (or some other approach). When docker is run without "host" networking type, docker automatically generates a random hostname. Mimicking the default docker behavior might be a good example to follow. Such a change would avoid engineers being confused when their Flow runs get picked up by someone else's local agent.

## Example
image

joshmeek:

## What does this PR change?
Closes #2618 
Uses the host IP when running LocalAgent inside of a docker container that is using the host network

## Why is this PR important?
Would lead to agents attempting to pull runs from other containers due to this constant hostname

### DIFF
--- a/changes/issue2618.yaml
+++ b/changes/issue2618.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - Use host IP for `hostname` label in cases where `LocalAgent` is in container using host network - [#2618](https://github.com/PrefectHQ/prefect/issues/2618)
--- a/src/prefect/agent/local/agent.py
+++ b/src/prefect/agent/local/agent.py
@@ -74,6 +74,11 @@ class LocalAgent(Agent):
             no_cloud_logs=no_cloud_logs,
         )
         hostname = socket.gethostname()
+
+        # Resolve common Docker hostname by using IP
+        if hostname == "docker-desktop":
+            hostname = socket.gethostbyname(hostname)
+
         if hostname_label and (hostname not in self.labels):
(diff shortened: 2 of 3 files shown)


jcrist:
In general this seems fine, but having a special case for a common hostname seems a bit hacky (you also may still get collisions with the ips). I wonder if we should emphasize the need for users to select unique labels themselves more in the docs?

cicdw:
@jcrist yea we introduced these auto-labels on both Storage types + Agents to help users not get confused when their locally-stored flow was erroring out on Kubernetes (for example).  We might consider allowing users to opt out of the auto-labels which would be one way of solving this particular issue.

jcrist:
I think the auto labels are good to prevent e.g. k8s runs going to a local agent, but I think we should clarify in our docs better that they aren't sufficient, and that users should really be adding their own labels to agent deployments.

PR #2641, opened 2020-05-22
A few cleanups to the k8s agent resource manager

### LINKED ISSUE #1588: Set k8s resource manager to only look for prefect identifier labeled resources
## Current behavior
Currently the resource manager is greedy with what it chooses to clean up, taking every job in a set namespace.

## Proposed behavior
Set the resource manager to only clean up jobs/pods with prefect identifier labels (which we already place on them when created)

jcrist:
A few fixes to the k8s agent resource manager

- No longer delete pods explicitly. Deleting a job deletes any directly associated pods, which is all the resource manager should be concerned with. Previously this also had the fun behavior of deleting any non-running pod in a namespace (including pods that were pending or not associated with prefect).
- Only manage jobs that have a `prefect.io/identifier` label. Previously this managed all jobs in a namespace, including those not associated with prefect.
- Update our k8s labels to use a `prefect.io/` prefix, following kubernetes best practices.

Fixes #1588.

### DIFF
--- a/changes/issue1588.yaml
+++ b/changes/issue1588.yaml
@@ -1,0 +1,2 @@
+fix:
+  - Kubernetes agent resource manager is more strict about what resources it manages - [#2641](https://github.com/PrefectHQ/prefect/pull/2641)
--- a/changes/issue1588.yaml
+++ b/changes/issue1588.yaml
@@ -1,2 +1,5 @@
 fix:
   - Kubernetes agent resource manager is more strict about what resources it manages - [#2641](https://github.com/PrefectHQ/prefect/pull/2641)
+
+breaking:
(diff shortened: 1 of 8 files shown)


joshmeek:
@jcrist We also need to update the labels attached to jobs created with the Kubernetes related environments to use the `prefect.io/` pattern.

jcrist:
Ah, good catch. I always forget about the job-that-starts-a-job thing.

changes/issue1588.yaml
```diff
@@ -0,0 +1,2 @@
+fix:
+  - Kubernetes agent resource manager is more strict about what resources it manages - [#2641](https://github.com/PrefectHQ/prefect/pull/2641)
```
joshmeek, line 2: Could you also add to this that the identifier labels on the environments changed?

jcrist: Should I note that as a breaking change? It likely won't affect most users, but it may affect some.

joshmeek: Yeah I'm cool with that 👍

PR #2642, opened 2020-05-22
Move worker stealing warning in docs

lauralorenz:

- [n/a] add a changelog entry in the `changes/` directory (if appropriate)

## What does this PR change?
Move the warning about dask worker stealing to under the docs on version locking, since it is only relevant in that case.

## Why is this PR important?
People deploying on Dask are probably reading our deployment tutorial first, but are unsure if enabling worker stealing (which they are likely prefer to optimize their Dask cluster later) is the source of any other issues they might have.

### DIFF
--- a/docs/orchestration/concepts/flows.md
+++ b/docs/orchestration/concepts/flows.md
@@ -153,9 +153,9 @@ mutation {
 ```
 
 ::: warning Dask Work Stealing with Version Locking
-**If you are using Dask**, we recommend turning off [Dask work stealing](https://distributed.dask.org/en/latest/work-stealing.html) if you enable version locking, as in rare occasions work stealing can result in tasks attempting to run twice which breaks the idempotency constraint of version locking.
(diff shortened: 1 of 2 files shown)


docs/orchestration/concepts/flows.md
```diff
@@ -152,6 +152,14 @@ mutation {
 }
 ```
 
+::: warning Dask Work Stealing with Version Locking
+**If you are using Dask**, we recommend turning off [Dask work stealing](https://distributed.dask.org/en/latest/work-stealing.html) if you enable version locking, as in rare occasions work stealing can result in tasks attempting to run twice which breaks the idempotency constraint of version locking.
```
cicdw, line 156: We might want to relax the wording here -> version locking will successfully prevent the task from running twice _but_ at this moment it causes a lot of log noise and results in some inefficient re-submissions of the flow run (to be clear, this is something we should improve)

PR #2643, opened 2020-05-23
add example result location that includes file-name

zhen0:

## What does this PR change?

Updates the docs to give an example of a result location that includes {filename} for mapped tasks. 

## Why is this PR important?

Gives users an example to better understand what is needed for targets on mapped tasks. 

Should wait until #2640 is closed and filename is populated in local runs (or I can update and include an example for local runs if needed).


joshmeek:
@zhen0 #2640 has been closed by #2717

PR #2646, opened 2020-05-24
Depth first execution

### LINKED ISSUE #2041: Implement Depth-First Execution with Mapping
Currently each "level" of a mapped pipeline is executed before proceeding to the next level.  This is undesirable especially for pipelines where it's important that each "branch" of the pipeline finish as quickly as possible.

To implement DFE, we'll need to rearrange two things:
- how mapped work gets submitted (it should start being submitted from the Flow Runner not the Task Runner)
- in order to submit work to Dask and let Dask handle the DFE scheduling, we'll want to refactor how we walk the DAG and wait to determine the width of a pipeline before we submit it (because mapping is fully dynamic we can only ascertain this width at runtime)

We'll need to be vigilant about:
- performance
- retries
- result handling

cicdw:

Remaining TODO:
- [x] add comments in the code to help devs reason about it
- [ ] update any relevant documentation
- [ ] leave open to provide time for review and QA

## What does this PR change?
This PR introduces "Depth first execution" (a Dask-only feature).  To explain via example:

```python
import time

@task
def times():
    return [1, 5, 10]

@task
def sleep(x):
    print(f'sleeping for {x}')
    time.sleep(x)
    print(f'done sleeping for {x}')
    return x

with Flow("mapping-test") as flow:
    sleep.map(sleep.map(times))

flow.run(executor=DaskExecutor())
```

If you run the above code on the current `master` branch, you'll find that the task execution order looks like:
```
sleeping for 1
done sleeping for 1
sleeping for 5
done sleeping for 5
sleeping for 10
done sleeping for 10
sleeping for 1
done sleeping for 1
...
```

This means that each "level" of the mapped pipeline must complete prior to the next level beginning.  If you run this same code on this PR / branch, you'll instead find:

```
sleeping for 1
sleeping for 5
sleeping for 10
done sleeping for 1
sleeping for 1
done sleeping for 5
sleeping for 5
done sleeping for 10
...
```
indicating that the moment a task's upstream dependencies are ready, it will proceed regardless of other mapped children's status.

## Why is this PR important?

This PR is important for a number of reasons:
- Closes #2041 
- This feature has been highly requested since day 1
- Removes Prefect's reliance on `dask` worker clients, which makes Prefect's dask integration much cleaner


jlowin:
BRAVO!!!!!

Can confirm it works with a dynamic map of four levels...

src/prefect/engine/cloud/task_runner.py
```diff
@@ -354,8 +354,8 @@ def run(
                 representing the states of any tasks upstream of this one. The keys of the
                 dictionary should correspond to the edges leading to the task.
             - context (dict, optional): prefect Context to use for execution
-            - executor (Executor, optional): executor to use when performing
-                computation; defaults to the executor specified in your prefect configuration
+            - mapped_parent (bool): a boolean indicating whether this task run is the run of a parent
```
jlowin, line 357: Minor point but it may be useful to document the difference between "parent" and "children" mapped tasks (or why the concept of the parent exists at all). Could be outside PR scope, just noting.

src/prefect/engine/executors/base.py
```diff
@@ -28,20 +24,6 @@ def start(self) -> Iterator[None]:
         """
         yield
 
-    def map(self, fn: Callable, *args: Any) -> List[Any]:
```
jlowin, line 31: 👋

src/prefect/engine/flow_runner.py
```diff
@@ -432,17 +456,88 @@ def get_flow_run_state(
...
+                    ## either way, we should now have enough resolved states to restructure
+                    ## the upstream states into a list of upstream state dictionaries to iterate over
+                    list_of_upstream_states = prepare_upstream_states_for_mapping(
+                        task_states[task], upstream_states, mapped_children
+                    )
+
+                    submitted_states = []
+
+                    for idx, states in enumerate(list_of_upstream_states):
+                        ## if we are on a future rerun of a partially complete flow run,
```
jlowin, line 497: Following our debugging session, this is definitely the correct comment. However I think in a few months we won't understand how/why it translates into the if statement below it

src/prefect/engine/task_runner.py
```diff
@@ -270,29 +259,16 @@ def run(
...
-                        )
-                    )
-                    raise ENDRUN(state)
-
                 # retrieve task inputs from upstream and also explicitly passed inputs
                 task_inputs = self.get_task_inputs(
                     state=state, upstream_states=upstream_states
                 )
 
+                if mapped_parent:
```
jlowin, line 267: Some pipeline step methods begin with an `if` statement, and others are guarded by `if` statements in the pipeline itself, like this one. `check_task_reached_scheduled_start_time()` is a [good example](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/engine/task_runner.py#L572) of the former; it is called for ALL states and only applies its logic to scheduled states. Both are valid, but should we standardize on one or the other? Right now, I think `check_task_ready_to_map` will fail if called on a non-mappable upstream state.

src/prefect/engine/task_runner.py
```diff
@@ -438,6 +411,45 @@ def check_upstream_skipped(
...
+            raise ENDRUN(state)
+
+        ## we can't map if there are no success states with iterables upstream
+        if upstream_states and not any(
+            [
+                edge.mapped and state.is_successful()
+                for edge, state in upstream_states.items()
+            ]
+        ):
+            new_state = Failed("No upstream states can be mapped over.")  # type: State
```
jlowin, line 438: I'm surprised this type annotation was required, though it doesn't hurt

src/prefect/engine/flow_runner.py
```diff
@@ -453,31 +480,36 @@ def get_flow_run_state(
...
                     list_of_upstream_states = prepare_upstream_states_for_mapping(
-                        task_states[task], upstream_states,
+                        task_states[task], upstream_states, mapped_children
                     )
 
                     submitted_states = []
 
                     for idx, states in enumerate(list_of_upstream_states):
+                        ## if we are on a future rerun of a partially complete flow run,
+                        ## there might be mapepd children in a retrying state; this check
```
joshmeek, line 498:
```suggestion
                        ## there might be mapped children in a retrying state; this check
```

joshmeek:
@cicdw I ran all of the tests that I ran prior to the most recent set of changes and it all works perfectly now 👍

src/prefect/engine/flow_runner.py
```diff
@@ -469,7 +564,8 @@ def get_flow_run_state(
             all_final_states = final_states.copy()
             for t, s in list(final_states.items()):
                 if s.is_mapped():
-                    s.map_states = executor.wait(s.map_states)
+                    # ensure we wait for any mapped children to complete
+                    s.map_states = executor.wait(mapped_children.get(t) or s.map_states)
```
jcrist, line 568: Why the `or s.map_states`? We don't want to store pending futures on the state, since that may be passed around via the executor, and sending futures between clients is error prone and expensive. For clarity I'd suggest:
```python
if t in mapped_children:
    s.map_states = executor.wait(mapped_children[t])
```

src/prefect/engine/flow_runner.py
```diff
@@ -572,13 +673,16 @@ def run_task(
                 default_result=self.flow.result,
             )
 
+            # TODO: wait elsewhere
```
jcrist, line 676: I believe this TODO is resolved.

src/prefect/engine/flow_runner.py
cicdw: At this step there are no further submissions through the executor, and setting this attribute is for user-facing interactions only (e.g., so users can access the states of the spawned tasks).  We need to reference `s.map_states` because this could be a re-run of a partially completed pipeline where we didn't actually submit any new work, and consequently don't want to lose the already resolved state information that `map_states` contains.

jcrist: > and consequently don't want to lose the already resolved state information that map_states contains.

Isn't that information  already in `s.mapped_states` (the existing logic ends up being `s.mapped_states = s.mapped_states` if `t` isn't in `mapped_children`)?

I'm just trying to clarify in the code that no futures are ever in `s.mapped_states`, so there's no point in calling `executor.wait` on them.

cicdw: Oh I see I see - yea that makes sense; will update.

src/prefect/engine/cloud/task_runner.py
cicdw: I'll leave this for a future PR, I'd like to document somewhere what DFE means so I can do it then.

src/prefect/engine/task_runner.py
cicdw: Interesting; I personally prefer this logic in the pipeline (as it is above) because it makes reasoning about the pipeline more transparent for me (visual inspection of the pipeline tells me that this step _can't_ cause issues for a non-mapped parent task, for example).

jlowin: I'm good either way, I just think we should standardize. I like your justification, maybe a future enhancement is to transform some of the others to this format.

jcrist:
Hi Chris, we'd like to merge this now. There's some merge conflicts though - could you handle those? Otherwise I will in a bit.

cicdw:
@jcrist updated 👍

tests/engine/executors/test_executors.py
```diff
@@ -251,6 +169,16 @@ def record_times():
         assert a_start < b_end
         assert b_start < a_end
 
+    def test_context_tags_are_passed_to_submit(self, monkeypatch):
```
jcrist, line 172: Bad merge conflict fix. This test is no longer needed, and won't pass as written. Could just delete the whole thing.

cicdw: Dang I knew I was going to miss something - thanks for checking!

peterroelants:
I'm trying to understand what is needed to run a flow in DFE.
If I understand it correctly it is activated by default when using a DaskExecutor, is this correct?

Issue #2111, opened 2020-03-03
Expose safe opt-in option for serializable environment information


maikelpenz:
## Current behavior
Currently I cannot identify what memory and cpu configuration has been set to a Flow. 

## Proposed behavior
Under the "Environment" section of the Flow Overview besides the Type we can also see configured memory and cpu information.

## Example
Environment
Type > FargateTaskEnvironment
Memory > 8192
Cpu > 2048

joshmeek:
I like this enhancement but the one thing I want to be cautious on is which information gets sent. There are subsets of information from environments (such as creds) that we do not want. 😄On that note I think that this functionality should be an opt-in addition where you may choose which aspects of the environment you want visible from Cloud. As the current environment landscape progresses it should be straightforward enough to reject specific information from being sent such as credentials.

We could serialize this as a generic `info` JSON blob in the environment serializers by using something like

```python
info = fields.Dict(key=fields.Str(), values=JSONCompatible(), allow_none=True)
```

joshmeek:
After `0.12.0` is released the environment `metadata` field seems like a good place to elevate this information. cc @zhen0 for if this is something the UI would be willing to add in the near future

zhen0:
This looks like info we could include in the flow details tile. Cloud UI ticket created in Cloud-web-ui #1624

joshmeek:
Thanks @zhen0!

joshmeek:
Closing as the `metadata` field on environments handles the core side of this 🙂

PR #2666, opened 2020-05-28
Add support for non-Docker storage on containerized environments

### LINKED ISSUE #2517: Support non-docker Storage for containerized Environments
*This came out of some user conversations, and discussion with @cicdw.*

Currently any `Environment` class that uses a container, e.g.

- `KubernetesJobEnvironment`
- `FargateTaskEnvironment`
- `DaskKubernetesEnvironment`

require the corresponding `Storage` object be a `Docker` instance (note that `DaskCloudProviderEnvironment` doesn't for some reason). The same restriction occurs for agents (e.g. `KubernetesAgent` and `FargateAgent`).

This coupling can be a bit troublesome, as it requires every flow have a new docker image built for it containing the flow run. Sometimes you'd like to use the same image for all flows, and store the flow definition in an external filestore (e.g. S3).

The `Docker` storage is unique in that it includes both the software environment, and the serialized flow. Other `Storage` classes only store the serialized flow, the software environment is part of the `Environment` class.

*The goal here is to be able to mix e.g. `KubernetesJobEnvironment` and `S3` storage, where the storage stores the flow definition, and the environment stores the execution environment (with an image possibly shared between flows).*

A proposal to allow this:

- The `Docker` storage still exists as it works today, storing both a software environment and the serialized `flow`
- Environments that use containers (e.g. `KubernetesJobEnvironment`) are updated to support explicit docker image names (in the k8s case, it would be part of the `Job` spec) in the case that the corresponding storage type isn't `Docker`. If a `Docker` storage is provided, it will be used to template over the image name as is currently done (logic is override if `Docker` storage, otherwise fallback to configured in environment if provided, otherwise error). At `flow.register()` time these are merged together, and the `Environment` becomes the source of truth for what image to use when starting a flow run.
- Environments will need to support *some* deserialization in the agent to extract the image name. We could hardcode in `image_name` as an attribute that's serialized, but I'm leaning more towards a generic `environment.metadata` dict (with conventions between agents and environments on common key names). So the `KubernetesAgent` might have the convention of looking for `flow.environment.metadata["image_name"]` when kicking off the initial job. This is needed since we don't want to deserialize the full user environment inside the agent (since that may be private).
- Agents that use containers (e.g. `KubernetesAgent`) are updated to support pulling the image name from `environment.metadata` (instead of `storage.name`). We'll need to do this in a backwards compatible way so that flows serialized using the old conventions still work.

joshmeek:

## What does this PR change?

Closes #2517 

This will allow for flows like the following:

```python
flow.storage = S3(bucket="my-bucket")
flow.environment = RemoteEnvironment(metadata={"image": "testimage:latest"})
# testimage:latest is an image I built locally that is authenticated with my S3 bucket
```

Then this flow can be run using agents like the Docker agent, all without ever putting the flow into a Docker image:

```
[2020-05-28 15:41:07,437] INFO - agent | Starting DockerAgent with labels ['s3-flow-storage']
[2020-05-28 15:41:07,437] INFO - agent | Agent documentation can be found at https://docs.prefect.io/orchestration/
[2020-05-28 15:41:07,438] INFO - agent | Agent connecting to the Prefect API at http://localhost:4200
[2020-05-28 15:41:07,455] INFO - agent | Waiting for flow runs...
[2020-05-28 15:41:07,775] INFO - agent | Found 1 flow run(s) to submit for execution.
[2020-05-28 15:41:07,815] INFO - agent | Deploying flow run 214e54b8-1386-4821-ae1c-b843f97c3ebe
[2020-05-28 15:41:08] INFO - prefect.S3 | Downloading env-metadata/2020-05-28t15-37-05-675587-00-00 from my-bucket
[2020-05-28 15:41:09] INFO - prefect.S3 | Downloading env-metadata/2020-05-28t15-37-05-675587-00-00 from my-bucket
[2020-05-28 15:41:09] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'env-metadata'
[2020-05-28 15:41:09] INFO - prefect.CloudFlowRunner | Starting flow run.
[2020-05-28 15:41:09] DEBUG - prefect.CloudFlowRunner | Flow 'env-metadata': Handling state change from Scheduled to Running
[2020-05-28 15:41:09] INFO - prefect.CloudTaskRunner | Task 'do': Starting task run...
[2020-05-28 15:41:09] DEBUG - prefect.CloudTaskRunner | Task 'do': Handling state change from Pending to Running
[2020-05-28 15:41:09] DEBUG - prefect.CloudTaskRunner | Task 'do': Calling task.run() method...
here
[2020-05-28 15:41:09] DEBUG - prefect.CloudTaskRunner | Task 'do': Handling state change from Running to Success
[2020-05-28 15:41:09] INFO - prefect.CloudTaskRunner | Task 'do': finished task run for task with final state: 'Success'
[2020-05-28 15:41:09] INFO - prefect.CloudFlowRunner | Flow run SUCCESS: all reference tasks succeeded
[2020-05-28 15:41:09] DEBUG - prefect.CloudFlowRunner | Flow 'env-metadata': Handling state change from Running to Success
```

I believe this is fully non-breaking because it checks for metadata attributes on the environment and if not found it falls back to the default accepted storage type.

Few things to address:
- [x] How should we now handle the auto-labeling? What we currently have is fine as long as it is documented. e.g. I had to add the `s3-flow-storage` label to my DockerAgent in order to get it to pull the flow
- [x] Currently the metadata / default storage resolution happens in the `deploy_flows` function of the containerized agents. Should we pull this out into an auxiliary function to make it cleaner?
- [x] I have this set to use the image in the metadata as `image`. Is this a good keyword to use?
- [x] Needs documentation on use. Especially around authentication. We could utilize the functionality where we allow setting secrets on the storage. I don't think this would require a change to the `execute` function because we load the flow when the secrets are already in context. :eyes:

## Why is this PR important?
Greatly reduces iteration time by not having to build images constantly on new versions of flows. Makes the cloud storage options much more versatile by allowing them to be used in conjunction with containerized environments.


server/docker/docker-compose.yml
```diff
@@ -9,8 +9,6 @@ services:
       POSTGRES_USER: ${POSTGRES_USER}
       POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
       POSTGRES_DB: ${POSTGRES_DB}
-    volumes:
-      - ${POSTGRES_DATA_PATH}:/var/lib/postgresql/data
```
joshmeek, line 13: I'm going to open a separate issue/PR for this. It was breaking dev build of server

joshmeek: #2689

src/prefect/agent/docker/agent.py
```diff
@@ -325,24 +326,41 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
-            self.logger.error(
-                "Storage for flow run {} is not of type Docker.".format(flow_run.id)
-            )
-            raise ValueError("Unsupported Storage type")
+        image = None
+        registry_url = False
+
+        # First check if `image` found in environment metadata, then default to storage
+        environment = EnvironmentSchema().load(flow_run.flow.environment)
+        if hasattr(environment, "metadata") and hasattr(environment.metadata, "image"):
```
cicdw: Curious about this - `image` seems like such a fundamental part of the environment, are we sure we want to keep it one level removed in this `metadata` field?

joshmeek: I don't have a big opinion on it but I would say that having things like this in a `metadata` field is nice from a serialization perspective since it wouldn't require new fields if it were to change in the future (all inside a json block). cc @jcrist 

Not to say that there couldn't be two new fields though `image` and `metadata`. Regardless, I'm a fan of a metadata field for generic information store.

cicdw: Interesting - that's a good point, the forwards / backwards compatibility provided by a JSON  blob is convenient

cicdw:
Note: I think this also closes https://github.com/PrefectHQ/prefect/issues/2111 because it introduces a generic `metadata` payload that we can render in the UI

src/prefect/agent/docker/agent.py
```diff
@@ -325,24 +321,16 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
-                "Storage for flow run {} is not of type Docker.".format(flow_run.id)
-            )
-            raise ValueError("Unsupported Storage type")
-
+        image = get_flow_image(flow_run=flow_run)
         env_vars = self.populate_env_vars(flow_run=flow_run)
 
-        if not self.no_pull and storage.registry_url:
-            self.logger.info("Pulling image {}...".format(storage.name))
+        if not self.no_pull and len(image.split("/")) > 1:
```
cicdw, line 327: What's the reasoning behind this image split?

src/prefect/utilities/agent.py
```diff
@@ -0,0 +1,38 @@
+import prefect
+from prefect.utilities.graphql import GraphQLResult
+
+
+def get_flow_image(flow_run: GraphQLResult) -> str:
```
cicdw, line 5: nice, this API function was a good idea

src/prefect/agent/docker/agent.py
joshmeek: If the full image name doesn't have a repo set (`repo/name:tag`) then don't attempt a pull. It's an extra safeguard which doesn't require to _also_ set `no_pull`

joshmeek:
Any comments @jcrist?

jcrist:
> Any comments @jcrist?

Sorry, I'm just wrapping up for the day. I'll give this a read tomorrow morning.

src/prefect/environments/execution/local.py
```diff
@@ -17,15 +17,19 @@ class LocalEnvironment(Environment):
             Agents when polling for work
         - on_start (Callable, optional): a function callback which will be called before the flow begins to run
         - on_exit (Callable, optional): a function callback which will be called after the flow finishes its run
+        - metadata (dict, optional): extra metadata to be set and serialized on this environment
     """
 
     def __init__(
```
jcrist, line 23: Since this `__init__` only forwards to its base class, the whole `__init__` could be removed.

src/prefect/utilities/__init__.py
```diff
@@ -9,3 +9,4 @@
 import prefect.utilities.notifications
 import prefect.utilities.serialization
 import prefect.utilities.tasks
+import prefect.utilities.agent
```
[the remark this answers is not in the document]
joshmeek: I don't have a preference on this. Worth opening a ticket to evaluate all of the `__init__` files because I think this is done in a lot of places

PR #2675, opened 2020-05-30
Refactor serialization of callables

### LINKED ISSUE #2499: Function serializers should support arbitrary callables
Currently our serializers error if callables (e.g. a user provided `cache_validator`) that aren't true functions are used (say callable objects). We should change this to better support serializing arbitrary callables.

See https://github.com/PrefectHQ/prefect/pull/2491#issuecomment-624356800 for context.

jlowin:

## What does this PR change?
Currently, callables are serialized in a brute-force-ish way. Functions are serialized as string references to an importable function, and functions with closure variables are serialized with references to introspected variables. Furthermore, Prefect has both a `FunctionReference` (for individual functions) and a `StatefulFunctionReference` (that handles both individual functions and factory functions with nonlocal state). This has always felt very messy and error-prone, as well as being difficult to communicate.

This PR refactors both of those references into a single `CallableReference` which works similarly to the old `StatefulFunctionReference`, with a minor modification: it no longer looks for closure variables in a factory function. Instead, it expects to serialize EITHER a direct reference to a function OR a class that conforms to a specific template (namely, having a `kwargs` attribute). 

Therefore, this function allows the serialization of both:
- `prefect.triggers.all_successful`, a function
- `prefect.triggers.some_successful(at_least=2)`, which used to be a factory function and is now a callable class.

For a simple example of this callable class structure, consider the new implementation of the schedule filter for a specific datetime:

```python
class on_datetime:

    def __init__(self, dt: datetime):
        self.kwargs = dict(dt=dt)

    def __call__(self, dt: datetime) -> bool:
        return dt == self.kwargs["dt"]
```
For reference, the old version of this was:
```python
def on_datetime(dt: datetime) -> Callable[[datetime], bool]:

    def _filter_fn(dt_2: datetime) -> bool:
        return dt_2 == dt

    return _filter_fn
```

The `CallableReference` field will properly serialize any class that stores its init kwargs in a `kwargs` field, like `on_datetime(dt=xyz)`. 

The old `FunctionReference` field stored function references as strings; this new `CallableReference` field appropriately loads those for backwards-compatibility.

## Why is this PR important?

I've been thinking about this for a while since @jcrist's comments that led to #2499. This PR does NOT allow arbitrary serialization - it still places a whitelist on these fields (because we call the functions, so we need additional measures to prevent calling non-whitelisted functions), but it does put in to place a formal structure for working with all serialized callables, across Prefect, in a backwards-compatible way. A future enhancement to close #2499 would be to allow the field to decide whether to load non-whitelisted functions, and we would achieve arbitrary serialization.


cicdw:
Closing for being too stale.  Happy to revisit if updated

PR #2681, opened 2020-05-31
add flag to include hostname on local storage

### LINKED ISSUE #2653: Allow users to register local flows without autogenerated labels
## Current behavior

As it exists today, flows registered with local storage are auto-labeled with information related to their local machines, and there's no way to opt out.

## Proposed behavior

Auto-labeling provides a sensible default, but we should also provide a way to opt out of labels for local flows.

## Example

As it stands today, it's cumbersome for users to run local flows registered on one machine on a different machine. While that's a pretty rare use case, we have had some folks opt to use Prefect in that way. Allowing users to toggle auto-generated labels would help remove some of that friction.

alexisprince1994:

closes #2653 

## What does this PR change?
This PR includes a flag on `LocalStorage` which will disable including hostname as the default label. By default, the `add_hostname_label` flag defaults to `True`, maintaining backwards compatibility.

## Why is this PR important?
See issue.

### DIFF
--- a/src/prefect/environments/storage/azure.py
+++ b/src/prefect/environments/storage/azure.py
@@ -31,8 +31,7 @@ class Azure(Storage):
             will be used
         - blob_name (str, optional): a unique key to use for uploading this Flow to Azure. This
             is only useful when storing a single Flow using this storage object.
-        - secrets (List[str], optional): a list of Prefect Secrets which will be used to populate `prefect.context`
-            for each flow run.  Used primarily for providing authentication credentials.
+        - **kwargs (Any, optional): any additional `Storage` initialization options
     """
 
     def __init__(
@@ -40,7 +39,7 @@ class Azure(Storage):
         container: str,
         connection_string: str = None,
         blob_name: str = None,
-        secrets: List[str] = None,
+        **kwargs: Any
     ) -> None:
         self.flows = dict()  # type: Dict[str, str]
         self._flows = dict()  # type: Dict[str, "Flow"]
@@ -55,10 +54,10 @@ class Azure(Storage):
         result = AzureResult(
             connection_string=self.connection_string, container=container
         )
-        super().__init__(result=result, secrets=secrets)
+        super().__init__(result=result, **kwargs)
 
     @property
-    def labels(self) -> List[str]:
+    def default_labels(self) -> List[str]:
         return ["azure-flow-storage"]
 
     def get_flow(self, flow_location: str) -> "Flow":
--- a/src/prefect/environments/storage/base.py
+++ b/src/prefect/environments/storage/base.py
@@ -3,6 +3,7 @@ from abc import ABCMeta, abstractmethod
 from typing import TYPE_CHECKING, Any, List
 
 import prefect
+from prefect import config
 from prefect.engine.result import Result
 from prefect.environments.storage import _healthcheck
 from prefect.utilities import logging as prefect_logging
@@ -22,7 +23,8 @@ class Storage(metaclass=ABCMeta):
             for each flow run.  Used primarily for providing authentication credentials.
         - labels (List[str], optional): a list of labels to associate with this `Storage`.
         - add_default_labels (bool): If `True`, adds the storage specific default label (if applicable)
-            to the storage labels. Defaults to `True`.
+            to the storage labels. Defaults to the value specified in the configuration at
+            `flows.defaults.storage.add_default_labels`.
     """
 
     def __init__(
@@ -30,12 +32,15 @@ class Storage(metaclass=ABCMeta):
         result: Result = None,
         secrets: List[str] = None,
         labels: List[str] = None,
-        add_default_labels: bool = True,
+        add_default_labels: bool = None,
     ) -> None:
         self.result = result
         self.secrets = secrets or []
         self._labels = labels or []
-        self.add_default_labels = add_default_labels
+        if add_default_labels is None:
+            self.add_default_labels = config.flows.defaults.storage.add_default_labels
+        else:
+            self.add_default_labels = add_default_labels
 
     @property
     def labels(self) -> List[str]:
--- a/src/prefect/environments/storage/base.py
+++ b/src/prefect/environments/storage/base.py
@@ -20,14 +20,42 @@ class Storage(metaclass=ABCMeta):
             all flows which utilize this storage class
         - secrets (List[str], optional): a list of Prefect Secrets which will be used to populate `prefect.context`
             for each flow run.  Used primarily for providing authentication credentials.
+        - labels (List[str], optional): a list of labels to associate with this `Storage`.
+        - add_default_labels (bool): If `True`, adds the storage specific default label (if applicable)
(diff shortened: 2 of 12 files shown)


src/prefect/environments/storage/local.py
```diff
@@ -31,14 +31,21 @@ class Local(Storage):
             absolute path and created.  Defaults to `True`
         - secrets (List[str], optional): a list of Prefect Secrets which will be used to populate `prefect.context`
             for each flow run.  Used primarily for providing authentication credentials.
+        - add_hostname_label (bool): if `True`, include the current host's hostname
```
jcrist, line 34: Instead of specifying the presence/absence of a specific label, we might add a general flag for whether to add default labels? This kwarg name could be used in other places in prefect where we might want to disable the automatic labels as well, providing some uniformity. Perhaps `add_default_labels=True` or `no_default_labels=False`?

alexisprince1994: Good w/ me! Haven't taken a look at any other default labels, but I'm liking your idea more. We could create a new property on the base `Storage` class that could look like the following:

```python
class Storage:
     def __init__(self, result: Result = None, secrets: List[str] = None, labels: List[str] = None # new) -> None:
         self.result = result
         self.secrets = secrets or []
         # New
         self._labels = labels or []

    # New (changed)
    @property
    def labels(self) -> List[str]:
         # No longer would be abstract, and move that dependency to the `default_labels` method?
        return set(default_labels).update(set(self._labels))

    @property
    def default_labels(self):
        raise NotImplementedError()
```

We could then handle the union of the 2 list of labels (I'm assuming order of the labels doesn't matter here, but I don't know that for sure). 

Thoughts?

jcrist: Currently storage objects don't support setting labels on them at all (environments do) - the only labels ever on a storage class are autogenerated by the class. We could add a `labels` kwarg to all the `Storage` objects for users to add their own labels as you've done above, but since a flow run is environment + storage, I'm not sure if that's needed? No strong thoughts.

Handling the presence/absence of default labels for storage in the base class sounds fine though.

alexisprince1994: One thing to mention too is that it appears the `Agents` don't separate the concept of a `Storage` label vs an `Environment` label. In the `LocalAgent` code, the label list gets extended with the 3 cloud storage magic strings (`azure-flow-storage`, `gcs-flow-storage`, `s3-flow-storage`) as well when checking for work. Although the actual check of whether the agent can deploy the flow is done by loading the flow's `Storage`, it seems like the only part that _could_ matter there is if people are also labeling their `Environment`'s with any of those magic strings? Probably a slim chance though.

I also don't have any strong opinions on it, but I also don't leverage the full functionality behind Prefect, so I'm not sure what the implications would be of tying these sets of labels together. I can put together the base implementation under the assumption that the users can define their label list twice and we can have a more targeted discussion after seeing that implementation?

jcrist: > I can put together the base implementation under the assumption that the users can define their label list twice and we can have a more targeted discussion after seeing that implementation?

Makes sense to me. Thanks for working on this.

src/prefect/environments/storage/azure.py
```diff
@@ -31,16 +31,15 @@ class Azure(Storage):
...
-            for each flow run.  Used primarily for providing authentication credentials.
+        - **kwargs (Any, optional): any additional `Storage` initialization options
     """
 
     def __init__(
         self,
         container: str,
         connection_string: str = None,
         blob_name: str = None,
-        secrets: List[str] = None,
```
joshmeek, line 43: Why was this removed?

src/prefect/serialization/storage.py
```diff
@@ -49,6 +40,8 @@ class Meta:
     flows = fields.Dict(key=fields.Str(), values=fields.Str())
     prefect_version = fields.String(allow_none=False)
     secrets = fields.List(fields.Str(), allow_none=True)
+    labels = fields.List(fields.Str(), allow_none=True)
```
cicdw, line 43: I think we should avoid touching the storage serializers in this PR; this adds compatibility issues and I think we should wait until we see how this is used.  Additionally we'll want a unified place for the "source of truth" of what labels are present on a flow, and right now that source of truth is the `Environment`.

src/prefect/environments/storage/base.py
```diff
@@ -20,14 +20,42 @@ class Storage(metaclass=ABCMeta):
...
+        self,
+        result: Result = None,
+        secrets: List[str] = None,
+        labels: List[str] = None,
+        add_default_labels: bool = True,
+    ) -> None:
         self.result = result
         self.secrets = secrets or []
+        self._labels = labels or []
+        self.add_default_labels = add_default_labels
```
cicdw, line 38: I'm tempted to think we should expose this as a configuration setting, maybe in [this section](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/config.toml#L117)

alexisprince1994: Good with me. Thoughts on still including as part of the constructor? Or exclusively a configuration option?

src/prefect/serialization/storage.py
alexisprince1994: Fine with me, but depending on whether we let users tag their storages with arbitrary labels, we may run into a scenario where:

```python
storage = Storage(labels=["foo", "bar"])
storage == StorageSchema().load(StorageSchema().dump(storage)) # False
```

src/prefect/environments/storage/base.py
cicdw: I actually like it as a part of the constructor (it's a relatively common pattern for us), so something like:
```python
def __init__(self, add_default_labels: bool = None):
    if add_default_labels is None:
        self.add_default_labels = config.flows.storage.add_default_labels
    else:
        self.add_default_labels = add_default_labels
```

src/prefect/serialization/storage.py
cicdw: Yea, that's actually fine - 99% of our serializers don't satisfy the check you wrote actually.  In this case the `labels` on Storage are purely used at registration time (prior to serialization) and never needed again.  As long as the deserialized Storage class can actually _run_ and retrieve the underlying Flow object as users intended it, we're good

src/prefect/environments/storage/base.py
alexisprince1994: Gotcha! Without the serialization changes and only being needed prior to registering flows, this makes sense to me!

alexisprince1994:
Tests are now failing because of the `Subclass` subclass I added to the `test_base.py` file for testing purposes. Any suggestions / recommendations on how to proceed on this? I wanted to be able to test the functionality without relying on the implementation of any _actual_ subclass, but if that's a fine assumption to have, I'm more than happy to replace the dummy subclass w/ one of the actual subclasses.

cicdw:
@alexisprince1994 if you subclass one of the other storage classes then that test should pass (not sure how you're using your class though so you might need to tweak the other tests)

Issue #2508, opened 2020-05-06
Decouple "Dask" and Environments


jcrist:
Currently prefect cloud/server flow runs are configured by 3 concepts:

- `Environment`
- `Executor`
- `Storage`

while local flow runs are configured with an `Executor` alone.

Subclasses of `Environment` and `Executor` both exist with "dask" in their name, which can make things confusing, especially since the usage differs between local and cloud/server usage.

For example, if you're trying to run a flow on kubernetes using Dask:

- If you're running locally, you'd use `dask-kubernetes` to spin up a dask cluster, then configure a `DaskExecutor` to use that cluster, then pass that executor to `flow.run`.
- If you're running using cloud, you'd configure a `DaskKubernetesEnvironment` and attach it to your flow. Environments also have executors, which makes this confusing, since you have a `Dask*` thing inside a `Dask*` thing.

If you then decide you want to not use a dask executor but still run on cloud, you need to change your environment to a `KubernetesJobEnvironment`. But you still can use a dask executor inside the `KubernetesJobEnvironment`? I find the current mixing of concepts a bit confusing - it feels like environments are encroaching in executors' space.

---

I think a potentially cleaner way of specifying flow execution would be to remove anything "dask" related from environments entirely, and move them to executors alone. The separation of duties would be:

- `Environment`:

  Configures where and how to run the process that starts up a `FlowRunner`. This may be a kubernetes job, a docker container, a local process, etc...

- `Executor`:

  Configures where and how to run *tasks* inside a single flow run. An executor could run things without dask, with dask locally, connect to a remote dask cluster, or start a new temporary dask cluster just for this flow run. This is basically how things work now (minus the cluster creation).

One nice thing about this is that the only difference between running a flow locally and running a flow on cloud is the addition of the `Environment`. If you want to use a dask cluster on kubernetes to run your flow, in both cases you configure an `Executor` to spin up a dask cluster using `dask-kubernetes`. The `Environment` is only responsible for managing the initial process, not the dask processes.

---

If people like this plan, I think a way forward would be:

- Create an `Executor` class that manages a dask cluster lifetime (starting/stopping/scaling) using the standard dask cluster interface (similar to #2503, which I think is superseded by this issue). This might look like:

   ```python
   executor = DaskExecutor(
       cluster_class="dask_cloudprovider.FargateCluster",
       cluster_kwargs={
           "image": "my-fancy-prefect-image",
       },
       ...
   )
   ...
   ```

[...]

PR #2687, opened 2020-06-01
Add the Dask Gateway Environment

ThomasLaPiana:

## What does this PR change?

This PR adds a new execution environment, `DaskGatewayEnvironment`.

## Why is this PR important?

DaskGateway provides a valuable alternative to spinning up and managing multiple clusters in Kubernetes via things like the `DaskKubernetes` execution environment. As teams grow larger it becomes more difficult to manage things like access control and resources within Kubernetes and Dask. Gateway gives admins the ability to create a safe environment for users to play in, giving them fine grain control over what customization options are exposed to the end-users. It also more elegantly handles the management and life-cycles of multi-cluster environments. This PR allows users to leverage their Dask Gateway instances for Prefect execution


jcrist:
Hi Thomas, thanks for this PR. We're currently working through #2508, which will make explicit Dask environments like this unnecessary. If it's ok with you, I'd prefer to close this, as we'd need to deprecate this later once that work is done. Since Prefect supports custom environments, not having this in Prefect core shouldn't prevent you from using it while we finish up #2508. Is that acceptable?

jcrist:
The end goal of #2508 is that instead of configuring an environment with Dask specifically, you'd configure an environment for where you want your flow to start (say a kubernetes job), and attach an executor to that for how you want your tasks to be run. With #2667 this executor could create a DaskGateway cluster. Something like:

```python
from prefect.environments import KubernetesEnvironments

executor = DaskExecutor(
    cluster_class="dask_gateway.GatewayCluster",
    cluster_kwargs={...},
    adapt_kwargs={"minimum": 0, "maximum": 10}
)
    ...,
    executor=executor,
)
```

This separates the configuration of where the flow-runner is created and running from where the tasks run, and lets us avoid creating Prefect classes for each of dask's cluster managers. This work is still ongoing, but I'm hoping to resolve it in the next few weeks.

ThomasLaPiana:
> Hi Thomas, thanks for this PR. We're currently working through #2508, which will make explicit Dask environments like this unnecessary. If it's ok with you, I'd prefer to close this, as we'd need to deprecate this later once that work is done. Since Prefect supports custom environments, not having this in Prefect core shouldn't prevent you from using it while we finish up #2508. Is that acceptable?

correct, i'm already using it in prod internally. 

That seems like a much better way to go, and i totally understand this will just get deprecated in a few weeks so I'm good to go ahead and close it. That being said, i'd love to get my name in the contributor list so is there anywhere else I could help out with the Dask implementation?

Issue #2577, opened 2020-05-16
Pass `inputs` through to Result class


limx0:
I would like to be able to access a tasks inputs when reading/writing results. This allows properly caching based on the different inputs to a function, for example by templating a filename as `experiment;m=5`.  

## Current behavior
*Please describe how the feature works today*
The `prefect.context` is passed to the read/write(exists?) methods of the Results class 

## Proposed behavior
*Please describe your proposed change to the current behavior*
Please pass the inputs also. 

## Example
*Please give an example of how the enhancement would be useful*
This would allow something like the following
```
@task(result=GCSResult(bucket='bucket', location='{task_name};m={inputs["multiplier"].txt'))
def experiment(multiplier=1):
    # do something
    return
```

limx0:
In fact on closer inspection, it might be nicer/more flexible if `target` was a callable, with `format` being the default? When passing a string, `format` could just be a `functools.partial` along the lines of 

```
def default_format(s, **context):
    return s.format(**context)
```

For my use case I actually really need the tasks full argspec, so I can also include any default values, which I'm not sure are passed through as `inputs`. I wouldn't want my file names to be generating differently depending on whether an upstream task passed a value (and it therefore was in the inputs) or if it had a default value that was the same. So perhaps just adding `task` and `task_inputs` to the `**prefect.context` dict and passing one dict through would be ideal. 

I'll leave this here for any other thoughts but maybe I'll also throw together a PR..

jlowin:
We already add "small" info about the task to context but I think we could safely add the task itself as well as its inputs and solve this in a general way.

The only caveat I have in mind is to ensure that the pickling protocol appropriately de-dupes this new information because otherwise it will double the serialization overhead of each task when sent to a remote worker.

bachsh:
in some cases, the output of the tasks will depend on the flow parameters, so it makes sense to include the parameters available to the result templating. It is available in the context, BUT not all tasks depend on all the flow parameters. Each task will typically depend on a subset of parameters (recursively based on its upstream tasks). So perhaps include the dictionary `upstream_parameters` in the context can give this functionality without having to give access to the inputs themselves.
WDYT?

cicdw:
Because Prefect inputs can be arbitrary (ranging from simple numbers / strings to large DataFrames / custom classes) we need to provide a generic interface that will work for all types.  

Here's what I'm thinking: we introduce a new `templator` kwarg on all `Result` classes:
```python
[...]

PR #2698, opened 2020-06-02
Result location updates

cicdw:

## What does this PR change?
This PR does a few things:
- stores the full absolute path to the result location for `LocalResult` (note that this change _should_ have no bearing on implementation, but instead makes sure the full path is visible in the UI)
- makes sure that `prefect.context.parameters` now also includes unprovided parameter defaults (previously this dictionary only included parameters that were explicitly passed)
- passes flow `parameters`, task `inputs` _and_ `context` to `result.write(**kwargs)` call for richer templating possibilities
- allows for users to (optionally) provide a callable location for their `Result` locations which will passed _all_ formatting kwargs.  This callable should return a `str` which will be used as the location.

## Why is this PR important?
Closes https://github.com/PrefectHQ/prefect/issues/2577 and provides more expressive result interactions

### DIFF
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -147,10 +147,12 @@ class FlowRunner(Runner):
         """
 
         # overwrite context parameters one-by-one
-        if parameters:
-            context_params = context.setdefault("parameters", {})
-            for param, value in parameters.items():
-                context_params[param] = value
+        context_params = context.setdefault("parameters", {})
+        for p in self.flow.parameters():
+            if not p.required:
+                context_params.setdefault(p.name, p.default)
+        for param, value in (parameters or {}).items():
+            context_params[param] = value
 
         context.update(flow_name=self.flow.name)
         context.setdefault("scheduled_start_time", pendulum.now("utc"))
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -80,8 +80,10 @@ class Result(ResultInterface):
         - validators (Iterable[Callable], optional): Iterable of validation functions to apply to
             the result to ensure it is `valid`.
         - run_validators (bool): Whether the result value should be validated.
-        - location (str, optional): Possibly templated location to be used for saving the
-            result to the destination.
+        - location (Union[str, Callable], optional): Possibly templated location to be used for saving the
+            result to the destination. If a callable function is provided, it should have signature `callable(**kwargs) -> str`
+            and at write time all formatting kwargs will be passed and a fully formatted location is expected
+            as the return value.  Can be used for string formatting logic that `.format(**kwargs)` doesn't support
     """
 
     def __init__(
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -97,7 +97,12 @@ class Result(ResultInterface):
         self.result_handler = result_handler  # type: ignore
         self.validators = validators
         self.run_validators = run_validators
-        self.location = location
+        if isinstance(location, (str, type(None))):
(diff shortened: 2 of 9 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -983,7 +983,13 @@ def get_task_run_state(
             and value is not None
         ):
             try:
-                result = self.result.write(value, filename="output", **prefect.context)
+                result = self.result.write(
+                    value,
+                    filename="output",
+                    **prefect.context.get("parameters", {}),
```
bachsh, line 989: I think this can cause a runtime error if you have a task input with the same name as a flow parameter, can it?

cicdw: Good catch! Will fix. The ultimate goal is for the key priority to be context > inputs > parameters.

tests/engine/results/test_results.py
```diff
@@ -204,3 +204,17 @@ def test_local_exists_full_path(self, tmp_dir):
...
+
+class TestResultFormatting:
+    def test_result_accepts_callable_for_location(self, tmpdir):
+        result = LocalResult(dir=tmpdir, location=lambda **kwargs: "special_val")
+        assert result.location is None
+        new_result = result.format()
+        assert new_result.location.endswith("special_val")
+
+    def test_result_callable_accepts_kwargs(self, tmpdir):
+        result = LocalResult(dir=tmpdir, location=lambda **kwargs: kwargs["key"])
```
joshmeek, line 217: This is pretty neat 🙂

bachsh: Indeed!
I'd replace the lambda function with `operator.itemgetter("key")`

src/prefect/engine/task_runner.py
```diff
@@ -983,12 +983,15 @@ def get_task_run_state(
             and value is not None
         ):
             try:
+                formatting_kwargs = prefect.context.get("parameters", {}).copy()
+                formatting_kwargs.update(
+                    raw_inputs
+                )  # inputs take precedence over flow-level parameters
+                formatting_kwargs.update(
+                    prefect.context
+                )  # context has ultimate precedence
```
bachsh, line 992: could be cleaner this way:
```
formatting_kwargs = {  # precedence: task context > task inputs > flow parameters
    **prefect.context.get("parameters", {}),
    **raw_inputs,
    ** prefect.context,
}
```

WDYT?

src/prefect/engine/flow_runner.py
```diff
@@ -147,10 +147,12 @@ def initialize_run(  # type: ignore
         """
 
         # overwrite context parameters one-by-one
-        if parameters:
-            context_params = context.setdefault("parameters", {})
-            for param, value in parameters.items():
-                context_params[param] = value
+        context_params = context.setdefault("parameters", {})
```
bachsh, line 150: having the default `parameters` inside even when none where entered is a blessing

src/prefect/engine/result/base.py
```diff
@@ -207,9 +214,11 @@ def format(self, **kwargs: Any) -> "Result":
             - Result: a new result instance with the appropriately formatted location
         """
         new = self.copy()
-        if new.location is not None:
+        if isinstance(new.location, str):
             assert new.location is not None
```
bachsh, line 218: I think this assertion is redundant now that you're checking for the type

cicdw: Yea that assert is purely for `mypy` type checking

src/prefect/engine/task_runner.py
cicdw: Great call out - I'll update!

PR #2707, opened 2020-06-02
Reduce size of tasks submitted via prefect

### LINKED ISSUE #2042: Investigate size of cloudpickle payloads passed to `executor.submit`
Currently, we submit a class method to `executor.submit`.  This ultimately has the effect of serializing large portions of the class that could actually be imported.

I believe that our unusually large submissions here are partially responsible for eating up Dask Scheduler CPU cycles, but we'll need to investigate further to decide if that's true.

Ultimately this ticket is for:
- determining the effect of our payload size on the dask scheduler
- reducing our payload sizes where possible

jcrist:
Currently prefect submits very large objects to dask for every call. Large objects are expensive to serialize and send between workers, which results in slower execution and increased memory usage. The depth-first-execution PR helped some, but there's still more we can do to reduce this.

This PR changes the `run_task` method to be a function, and minimizes the set of arguments submitted to that function. Breaking apart dependencies like this also makes it explicit what actions take place in the flow runner, and what things take place in the workers, which (IMO) makes the code easier to understand.

Fixes #2042.


jcrist:
Things work fine in my brief test scripts, but some existing tests are failing (several of them due to the tests monkeypatching internals that have changed, which is a bit brittle). Still need to update the tests, and fix any bugs.

joshmeek:
@jcrist Any way to check on the size difference here between the previous iteration and this?

src/prefect/engine/flow_runner.py
```diff
@@ -461,26 +460,28 @@ def get_flow_run_state(
...
                     ## wait on upstream states to determine the width of the pipeline
                     ## this is the key to depth-first execution
-                    upstream_states.update(
-                        executor.wait(
-                            {e: state for e, state in upstream_states.items()}
-                        )
+                    upstream_states = executor.wait(
+                        {e: state for e, state in upstream_states.items()}
+                    )
+                    _set_upstream_states_map_states(
```
cicdw, line 466: In all of these `_set_upstream_states_map_states ` calls, how do we know that all futures in `.map_states` have been resolved to actual `State` objects?

src/prefect/engine/flow_runner.py
```diff
@@ -630,66 +635,56 @@ def determine_final_state(
...
+            mapped task
+
+    Returns:
+        - State: `State` representing the final post-run state of the `Flow`.
+    """
+    with prefect.context(context):
+        task_runner = task_runner_cls(
+            task=task,
+            state_handlers=task_runner_state_handlers,
+            flow_result=flow_result,
```
cicdw, line 684: Seems like some of the init calls have `flow_result` as the kwarg and others have `default_result`

src/prefect/engine/flow_runner.py
jcrist: I was mistaken about where things were being awaited, reverted.

jcrist:
We generally recommend not submitting methods to dask, as they're more expensive to serialize, and can bring unexpected things with them. The mapping refactor was the bulk of the benefit in serialization cost, since we're no longer serializing the executor or state, which could be large.

This PR extracts `run_task` out into a function to make it clearer what is being passed to a task run. It also refactors the internals to reduce the number of objects needed to serialize, and cleans up some duplicated normalization steps between `TaskRunner` and `FlowRunner`.

The main win here in size is not serializing the `flow` around. Most everything else that's improved here is small in size. But for maintainability, it's a lot easier to make changes to the `FlowRunner` and know you won't accidentally be serializing along some possibly large state, since everything is explicit.

A silly benchmark:

```python


@task
def task_1(a, b, c):
    return a + b + c


@task
def task_2(a, b, c):
    return a - b + c


@task
def task_3(a, b, c):
    return a + b - c


@task
def task_4(a, b, c):
    return a - b - c


with Flow("test") as flow:
    b = 0
    c = 0
    for a in range(100):
        a = task_1(a, b, c)
        b = task_2(a, b, c)
        c = task_3(a, b, c)

    res = task_4(a, b, c)
```

Running with an instrumented executor:

- **Master:** largest task: 125,800 bytes, total: 37,606,159 bytes
- **This PR:** largest task: 9661 bytes, total: 2,654,010 bytes

jcrist:
Tests pass, added a changelog entry. I believe this is ready for review.

joshmeek:
>Running with an instrumented executor:
>Master: largest task: 125,800 bytes, total: 37,606,159 bytes
>This PR: largest task: 9661 bytes, total: 2,654,010 bytes

Whoa that's a pretty hefty decrease in size!

PR #2710, opened 2020-06-03
This is an update to PostgresExecute. It will rollback the transactio…

### LINKED ISSUE #2658: Commit bool in PostgresExecute can be misleading
## Current behavior
PostgresExecute takes commit as a bool and the documentation says "Set to True to commit a transaction, defaults to False". This default to False only applies if you explicitly Begin a transaction in your query, otherwise the query commits automatically, ignoring the Commit Bool. 

## Proposed behavior

Given there could be multiple configurations that change this behavior, creating a ticket to see if there is a preference on how this is fixed with documentation or implementation. 

## Example

kingsleyb:
…n if commit is set to false in the task.

## What does this PR change?

Currently, if you execute a query against Postgres it will automatically commit, irrespective of the commit flag. This PR honors the commit arg on the PostgresExecute task by rolling back the transaction if commit is set to false. 

This change allows you to dry run flows without committing changes to the database. 

Any existing tasks using PostgresExecute with commit set to false will no longer commit by default with this PR. 

## Why is this PR important?
Closes #2658

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Released on June 2, 2020.
 
 ### Task Library
 
-- Fix PostgresExecute task auto commit when commit is set to `False` - [#2658](https://github.com/PrefectHQ/prefect/issue/2658)
+- None
 
 ### Server
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Released on June 2, 2020.
 
 ### Task Library
 
-- None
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,7 +20,7 @@ Released on June 2, 2020.
 
 ### Task Library
 
-- None
+- Fix PostgresExecute task auto commit when commit is set to `False` - [#2658](https://github.com/PrefectHQ/prefect/issue/2658)
```
joshmeek, line 23: The way we update the changelog has changed. Use the process outlined here https://github.com/PrefectHQ/prefect/tree/master/changes

PR #2713, opened 2020-06-03
Add idiom for notifications

joshmeek:
Small idiom addition for notification types. I could expand on this more however a lot of notification based concepts are already present in the docs. This is more of a small way to tie them together.

### DIFF
--- a/docs/core/idioms/notifications.md
+++ b/docs/core/idioms/notifications.md
@@ -47,10 +47,12 @@ slack_task.set_upstream(value, key="message", flow=flow)
 :::
 ::::
 
+In this example the `SlackTask` notification will execute every time the upstream task is successful. To enable a different behavior where this task runs due to different upstream task states attach a [trigger](/api/latest/triggers.html#triggers) to the task. For example, you could execute this task in certain situations such as upstream failures using an `any_failed` trigger.
+
 ### State handlers
 
-Another common pattern for sending notifications is through [state handlers](/core/concepts/notifications.html#state-handlers). State handlers are a way of reacting to specific state changes on a task and Prefect ships with a [few state handlers](/api/latest/utilities/notifications.html) for working with different services. Outside of those default handlers it is straightforward to make completely custom handlers! For more information on this take a look at the relevant [concept documentation](/core/concepts/notifications.html).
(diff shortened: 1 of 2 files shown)


docs/core/idioms/notifications.md
```diff
@@ -0,0 +1,56 @@
+# Configuring notifications
+
+Sending notifications based on events in Prefect is a common pattern. Types of notifications generally include functionality such as posting state of runs to Slack, sending a text message on failure, etc. There are many places in Prefect to hook in notification systems and this idiom will go over three possibilities.
+
+### Downstream tasks
+
+Adding tasks to your flow that respond to upstream events is one way of sending notifications. These tasks can be built in any way you would like and the Prefect task library even ships with a few [notification tasks](/api/latest/tasks/notifications.html)! Below is an example of using the [`SlackTask`](/api/latest/tasks/notifications.html#slacktask) to send a notification to Slack.
```
cicdw, line 7: I think we should call out using `triggers` as a way of only firing the notification in certain situations. (e.g., only notify me when the task fails via an `any_failed` trigger on the slack task)

docs/core/idioms/notifications.md
```diff
@@ -0,0 +1,56 @@
...
+slack_task = SlackTask()
+
+slack_task.set_upstream(value, key="message", flow=flow)
+```
+:::
+::::
+
+### State handlers
+
+Another common pattern for sending notifications is through [state handlers](/core/concepts/notifications.html#state-handlers). State handlers are a way of reacting to specific state changes on a task and Prefect ships with a [few state handlers](/api/latest/utilities/notifications.html) for working with different services. Outside of those default handlers it is straightforward to make completely custom handlers! For more information on this take a look at the relevant [concept documentation](/core/concepts/notifications.html).
```
cicdw, line 52: We should at least differentiate between Flow-level state handlers and task-level state handlers here.

docs/core/idioms/notifications.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Cloud hooks <Badge text="Cloud"/>
+
+Users of Prefect Cloud have access to something called Cloud Hooks which is a way of configuring notifications based on state changes without having to write any code for the tasks or flow itself. For more information on Cloud Hooks visit the [documentation](/orchestration/concepts/cloud_hooks.html).
```
cicdw, line 56: We should call out in this section that Cloud hooks are the only _guaranteed_ way of getting notifications because it doesn't rely on your code running.  E.g., if an agent fails to deploy a flow run and Cloud marks the Flow run as "Failed", the above two patterns would not result in a notification but a Cloud Hook would still fire.

PR #2727, opened 2020-06-05
Add an idiom for conditional flow control

### LINKED ISSUE #2530: Write More Idioms
We should write some more idioms:

- [x] how to define conditional logic using the [new conditional api](https://github.com/PrefectHQ/prefect/pull/2443) and the "old" way
- [x] how to use `target`s (0.11.0+)
- [x] how to configure notifications (three options: write a downstream task, state handler, cloud hook)

jcrist:
This adds an idiom for conditional flow control to the docs. This covers both `case` and `merge`, which (IMO) are the main flow control constructs a Prefect user should need. Note that since this (and a few other idioms) overlapped with the `branching` idiom, I removed that one in favor of this.

This also adds an optional `flow` kwarg to `merge` to make it easier to use with the imperative api.

Fixes #2530.


jcrist:
The test failure appears to be unrelated.

PR #2736, opened 2020-06-07
Dask resources

cicdw:

## What does this PR change?
A user brought to my attention last night that our dask resource tags were not being respected; it appears that we pulled tags from context but those are not present at `executor.submit` time.  Instead of relying on implicit `context` keys this PR adds an `executor_kwargs` keyword to `submit` that allows the explicit passing of info.

## Why is this PR important?
Fixes bug.

### DIFF
--- a/src/prefect/engine/executors/base.py
+++ b/src/prefect/engine/executors/base.py
@@ -24,13 +24,16 @@ class Executor:
         """
         yield
 
-    def submit(self, fn: Callable, *args: Any, **kwargs: Any) -> Any:
+    def submit(
+        self, fn: Callable, *args: Any, executor_kwargs: dict = None, **kwargs: Any
+    ) -> Any:
         """
         Submit a function to the executor for execution. Returns a future-like object.
 
         Args:
             - fn (Callable): function that is being submitted for execution
             - *args (Any): arguments to be passed to `fn`
+            - executor_kwargs (dict, optional): an optional dictionary of executor-specific keyword arguments
             - **kwargs (Any): keyword arguments to be passed to `fn`
 
         Returns:
--- a/src/prefect/engine/executors/base.py
+++ b/src/prefect/engine/executors/base.py
@@ -25,7 +25,7 @@ class Executor:
         yield
 
     def submit(
-        self, fn: Callable, *args: Any, executor_kwargs: dict = None, **kwargs: Any
+        self, fn: Callable, *args: Any, extra_context: dict = None, **kwargs: Any
     ) -> Any:
         """
         Submit a function to the executor for execution. Returns a future-like object.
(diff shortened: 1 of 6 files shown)


src/prefect/engine/executors/local.py
```diff
@@ -9,13 +9,16 @@ class LocalExecutor(Executor):
...
+    def submit(
+        self, fn: Callable, *args: Any, executor_kwargs: dict = None, **kwargs: Any
+    ) -> Any:
         """
         Submit a function to the executor for execution. Returns the result of the computation.
 
         Args:
             - fn (Callable): function that is being submitted for execution
             - *args (Any): arguments to be passed to `fn`
+            - executor_kwargs (dict, optional): an optional dictionary of executor-specific keyword arguments
```
jcrist, line 21: Since we pass the same kwargs everywhere, they're not really executor specific. Perhaps either rename to `extra_context` (the context for this task is the current context in scope, plus the `extra_context` mapping), or explicitly codify the kwargs forwarded as `task_name` and `task_tags`?

cicdw:
Yea adding them to context was my first instinct but that introduces yet another indentation level in the flow runner which felt bad to me.

jcrist:
Yeah, I can see that. I think either renaming the dict kwarg here, or making `task_name` and `task_tags` explicit parameters makes sense in that case. Having explicit kwargs that can shadow function kwargs in `**kwargs` feels weird, but since the flow runner is the only consumer of `submit` it isn't really that bad.

src/prefect/engine/executors/local.py
cicdw: Yea I'm definitely game to change the name.  The reason I chose `executor_kwargs` is that they represent kwargs that the executor may / may not utilize, but this is a low-risk decision either way so I'll update to `extra_context` 👍

cicdw: Updated

jcrist: Yeah, to me `executor_kwargs` implies kwargs that are directly forwarded to the backing executor in some way, which isn't true here. Update looks good.

src/prefect/engine/executors/base.py
```diff
@@ -24,13 +24,16 @@ def start(self) -> Iterator[None]:
...
+    def submit(
+        self, fn: Callable, *args: Any, extra_context: dict = None, **kwargs: Any
+    ) -> Any:
         """
         Submit a function to the executor for execution. Returns a future-like object.
 
         Args:
             - fn (Callable): function that is being submitted for execution
             - *args (Any): arguments to be passed to `fn`
+            - extra_context (dict, optional): an optional dictionary of executor-specific keyword arguments
```
jcrist, line 36: oop, approved too quickly. Could you update the description for this as well? Something like:

```
an optional dictionary with extra information about the submitted task
```

cicdw: yup you got it

cicdw: Updated

PR #2741, opened 2020-06-08
Allow cloud flows to enter queued state

alexisprince1994:
- Flow runs can enter `Queued` state when attempting
    to transition into a `Running` state at the request of
    orchestration

## What does this PR change?
This PR allows flow runs controlled by the `CloudFlowRunner` to enter a `Queued` state at the request of Cloud when attempting to transition into a `Running` state. Flow runs controlled by the `CloudFlowRunner` will attempt to run every `cloud.queue_interval` (config setting) seconds forever. If the process running the `Queued` Flow run dies, Cloud's Lazarus process will revive them.

## Why is this PR important?
This PR is important in implementing flow run concurrency based on execution environments. Without the `CloudFlowRunner` knowing how to react to the `Queued` state, there is no easy way to keep track of which flow runs are currently running, which are pending, and how to control them.


src/prefect/engine/cloud/flow_runner.py
```diff
@@ -123,10 +126,104 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
+        Returns:
+            - State: `State` representing the final post-run state of the `Flow`.
+        """
+        context = context or {}
+        # Pre-emptively load the flow run id that exists in the global
+        # context into the run specific context, which is needed
+        # specifically for entering a Queued state and waiting forever until
+        # an exit condition occurs. Prefect's context will have the flow_run_id
+        # on the first call to run, and `context` will have it on all subsequent runs
+        if prefect.context.get("flow_run_id") and not context.get("flow_run_id"):
```
alexisprince1994, line 180: This feels hacky to me. Is there any way to persist the `flow_run_id` in the `prefect.context` across multiple `run` calls? After light testing, it appears to reset after each call.

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -154,7 +251,7 @@ def initialize_run(  # type: ignore
         """
 
         # load id from context
-        flow_run_id = prefect.context.get("flow_run_id")
+        flow_run_id = prefect.context.get("flow_run_id") or context.get("flow_run_id")
```
alexisprince1994, line 254: Without this change, all the tests still pass, but there currently isn't _too_ much testing of the `Queued` functionality, so I added the default to be safe as we're populating this value in `CloudFlowRunner.run`.

alexisprince1994:
Outside of comments on the couple of small things I've pointed out, I'd like feedback on the following as well:
0) It's my understanding that Flow Runs can't enter the `Queued` state as it stands today. If that isn't the case, please let me know how it's currently being used and what scenarios it happens under.
1) If the design decision to follow the `CloudTaskRunner`'s implementation of how it handles the `Queued` state is correct, and if there are alternatives I'd like to discuss them.
2) What other test scenarios should be added in? Currently, there is incredibly light testing, and I plan to add more when integrating the server changes that actually put Flow Runs into this state, but I'm also a fan of frontloading the testing, so all ideas are welcome!

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -154,7 +251,9 @@ def initialize_run(  # type: ignore
         """
 
         # load id from context
-        flow_run_id = prefect.context.get("flow_run_id")
+        flow_run_id = prefect.context.get("flow_run_id") or context.get(
```
jlowin, line 254: I recommend reversing these - the passed context should take precedence over the global context

alexisprince1994: Removed the check the passed context entirely. I'm not sure how I originally found the case where the `flow_run_id` wasn't in the global context, but I can't reproduce it.

joshmeek:
@alexisprince1994 and @jlowin I want to QA this a bit before we merge with a goal of getting it in post-0.12.0 which should be out in the next day or two.

alexisprince1994:
Sounds great @joshmeek, thanks for keeping me in the loop. Let me know if there's any way I can help with the QA process or if there are any unit test scenarios you'd like me to add for this functionality.

PR #2745, opened 2020-06-08
Add new CI step for a prefect image with all_extras installed

joshmeek:

## What does this PR change?
Adds an `all_extras` `prefecthq/prefect` image to be built on releases of Prefect.

## Why is this PR important?
When doing something like:

```python
f.storage = S3(bucket="my-flows", secrets=["AWS_CREDENTIALS"])
f.environment = RemoteEnvironment(metadata={"image": "image/name:here"})
```

The `image` needs to have the dependencies present for the selected `storage` (in this case, prefect's `aws` extras). This adds an image with `all_extras` and it acts as a self contained image with even more packages baked in. In this case you can use the entire task library without ever having to build an image for deployment.

(edited)

-----

Original text: 
> When doing something like:

> f.storage = S3(bucket="my-flows", secrets=["AWS_CREDENTIALS"])
> f.environment = RemoteEnvironment(metadata={"image": "image/name:here"})
> The image needs to have the dependencies present for the selected storage (in this case, prefect's aws extras). This is a mini proposal to add an extra image that we build on releases which add the dependencies for the cloud storage classes (aws, azure, gcp). If we are positioning this as a way to deploy flows without building images then we need to have a base image with these dependencies or else the user will still have to build at least one image.
> 
> Another option would be to make an image with all_extras and it would act as a self contained image with even more packages baked in. In that case you could use the entire task library without ever having to build an image for deployment.
> 
> cc @cicdw~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -440,7 +440,7 @@ jobs:
             docker login --username  $DOCKER_HUB_USER --password $DOCKER_HUB_PW
             docker push prefecthq/prefect:master
 
-  build_storage_docker_image:
+  build_all_extras_docker_image:
     docker:
       - image: docker
     parameters:
@@ -467,7 +467,7 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
-          name: Build storage image
+          name: Build all_extras image
           command: |
             set -u
             docker build \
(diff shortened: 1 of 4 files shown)


jcrist:
I might go for one with `all_extras` instead to keep the scope of images we provide smallish. This way we provide a lightweight version, and one with everything. Users that want more customization or space optimizations will have to build an image themselves.

joshmeek:
@lauralorenz Ah that's a great callout. I'm going to leave it as a separate step for now for clarity with the plan of cleaning this up at another time (we would need to template / rework image tags a bit since it follows a different pattern)

PR #2748, opened 2020-06-09
Fix env metadata not having access to name of Docker storage

joshmeek:
If a registry or name are not provided to Docker storage then the value of `storage.name` will raise an error if the storage has not been built yet (we add a random name/tag on build if not set). This moves the place where the `storage.name` is appended to the `environment.metadata` to _after_ the build.

### DIFF
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -250,7 +250,9 @@ class Docker(Storage):
         Full name of the Docker image.
         """
         if None in [self.image_name, self.image_tag]:
-            raise ValueError("Docker storage is missing required fields")
+            raise ValueError(
+                "Docker storage is missing required fields image_name and image_tag"
+            )
 
         return "{}:{}".format(
             PurePosixPath(self.registry_url or "", self.image_name),  # type: ignore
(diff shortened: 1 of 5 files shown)


lauralorenz:
FYI I can confirm that before this PR I was having the issue where my Docker storage, on first build wouldn't raise the ValueError:

```
 prefect git:(master) ✗ /opt/anaconda3/envs/prefect_dev/bin/python /Users/laura/Desktop/RandomFlows/filenameQAflow.py
  File "/Users/laura/Desktop/RandomFlows/filenameQAflow.py", line 44, in <module>
    f.register('testing-agents')
  File "/Users/laura/Development/prefect/src/prefect/core/flow.py", line 1432, in register
    self.environment.metadata["image"] = self.storage.name
  File "/Users/laura/Development/prefect/src/prefect/environments/storage/docker.py", line 253, in name
    raise ValueError("Docker storage is missing required fields")
```

And that it worked fine afterwards. 👯

src/prefect/environments/storage/docker.py
```diff
@@ -250,7 +250,9 @@ def name(self) -> str:
         Full name of the Docker image.
         """
         if None in [self.image_name, self.image_tag]:
-            raise ValueError("Docker storage is missing required fields")
+            raise ValueError(
+                "Docker storage is missing required fields image_name and image_tag"
```
cicdw, line 254: 👍

PR #2758, opened 2020-06-10
Move parameters to standalone module

jlowin:

## What does this PR change?
As I work with Parameters to increase their "first-class-ness", including extensions of #2755, I'm becoming acutely aware of the awkwardness of accessing them from `core/task.py`. This PR moves them to a more appropriate home, `core/parameter.py`. It makes no other changes to behavior or code.

I believe this is not a user-facing change as we have always encouraged accessing them at the top-level (in fact, our own code only referenced `task.Parameter` three times), but I've added a backwards-compatible access nonetheless.

So this remains the recommended access:
```python
from prefect import Parameter
```

But these are also possible:
```python
# new location
from prefect.core.parameter import Parameter
# deprecated location
from prefect.core.task import Parameter
```

(I did *not* move `prefect.serialization.task.ParameterSchema` because serializers are more sensitive, but am also looking at cleaning that up in other PRs)

## Why is this PR important?
Code organization

### DIFF
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -21,7 +21,7 @@ import prefect.agent
 # DEPRECATED - this is to allow backwards-compatible access to Parameters
 # This awkward assignment avoids circular imports in core.task
 # https://github.com/PrefectHQ/prefect/pull/2758
-prefect.core.task.Parameter = Parameter # type: ignore
+prefect.core.task.Parameter = Parameter  # type: ignore
 
 # ---------------------------------------------
 # Versioneer
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -18,15 +18,6 @@ import prefect.serialization
 
 import prefect.agent
 
-# DEPRECATED - this is to allow backwards-compatible access to Parameters
-# This awkward assignment avoids circular imports in core.task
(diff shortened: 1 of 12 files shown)


src/prefect/__init__.py
```diff
@@ -18,6 +18,15 @@
 
 import prefect.agent
 
+# DEPRECATED - this is to allow backwards-compatible access to Parameters
+# This awkward assignment avoids circular imports in core.task
+# https://github.com/PrefectHQ/prefect/pull/2758
+prefect.core.task.Parameter = Parameter  # type: ignore
```
jcrist, line 24: If we wanted to actually deprecate access of `Parameter` in the `prefect.core.task` namespace with a warning, I have a hack I've used in other projects to do so. Not sure what the deprecation policy for this project is.

jlowin: I'd love to hear it - in the past I've used `zope.deprecation` but it's way too heavy to bring in just for this. Let me know and I'll update.

jcrist: Mind if I push to this branch?

jcrist: Done. Nothing tricky was needed here, just a subclass with a warning. In the past I've wanted to raise a warning when users imported from a spot without subclassing which was trickier - in this case a subclass should be fine.

PR #2763, opened 2020-06-10
Hashed inputs

cicdw:

## What does this PR change?
This PR introduces a new attribute, currently only on `Cached` states, of `hashed_inputs`.  This fixes a bug introduced into `master` in which caching doesn't work for mapped tasks.  If this is accepted, I propose we move away from `cached_inputs` entirely - now that we checkpoint all task outputs, storing cached inputs is unnecessary (except in this case of caching, in which case the hash is sufficient).

I need to write some more tests but wanted to open this up for review / discussion.


joshmeek:
No qualms from me and I am +1 on moving away from cached inputs but could you clarify what you mean by

> except in this case of caching, in which case the hash is sufficient

In which scenarios using caching are the inputs needed as opposed to only using the upstream outputs? The only one that comes to mind is validation based on inputs

cicdw:
Yea @joshmeek the only time we actually need to reference the inputs from previous runs is to validate whether a `Cached` state is valid or not, and in that case we don't actually need the data itself only the hash.  I don't think there's any other situation (anymore) where we need to reference the _inputs_ of a previous run.

joshmeek:
@cicdw Awesome thanks that makes perfect sense 👍

PR #2767, opened 2020-06-11
Raise error on downstream mapped tasks with trigger other than all_successful

joshmeek:

## What does this PR change?
Raise a `ValueError` when a mapped task's trigger is anything other than `all_successful`. Attempting to map over errors/exceptions can lead to unintentional behavior.

## Why is this PR important?
Situations like this:
```python
@task
def values():
    return [1, 2, 3]

@task
def calculate(x):
    if x == 2:
        raise ValueError()
    return x

@task(trigger=any_failed)
def sumit(x):
    print(sum(x))

with Flow("test") as f:
    vals = values()
    x = calculate.map(vals)
    sumit.map(x)
```

Will lead to unintentional runtime errors. We want to catch these potential errors earlier:

```
  File "/Users/josh/Desktop/code/prefect/src/prefect/engine/runner.py", line 48, in inner
  File "/Users/josh/Desktop/code/prefect/src/prefect/engine/task_runner.py", line 806, in get_task_run_state
  File "/Users/josh/Desktop/code/prefect/src/prefect/utilities/executors.py", line 186, in timeout_handler
    return fn(*args, **kwargs)
  File "maptrigger.py", line 16, in sumit
    print(sum(x))
TypeError: 'ValueError' object is not iterable
```

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -515,10 +515,15 @@ class Flow:
         # flow, so we need to perform this check here and not earlier.
         if validate and key and key in {e.key for e in self.edges_to(downstream_task)}:
             raise ValueError(
-                'Argument "{a}" for task {t} has already been assigned in '
+                "Argument '{a}' for task {t} has already been assigned in "
                 "this flow. If you are trying to call the task again with "
                 "new arguments, call Task.copy() before adding the result "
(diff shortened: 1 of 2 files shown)


joshmeek:
Rejecting this PR, there are some subtle implications here that feel too risky to move forward with. Keeping branch in case we want to revisit

PR #2769, opened 2020-06-11
Allow for task targets to be callables

joshmeek:

## What does this PR change?
This PR allows for task targets to be both strings and callables. This is similar to how result locations can also be both. e.g.:
```python
@task(target=lambda **kwargs: "{task_name}-88")
def get_data():
    """test"""
    return "data"

@task
def print_data(data):
    print(data)


with Flow("using-targets", result=LocalResult(), ) as flow:
    data = get_data()
    print_data(data)

flow.run()
```

```
# First run
[2020-06-11 21:50:17] DEBUG - prefect.LocalResult | Starting to upload result to get_data-88...
[2020-06-11 21:50:17] DEBUG - prefect.LocalResult | Finished uploading result to /Users/josh/.prefect/results/get_data-88...

...

# Second run
[2020-06-11 21:50:19] DEBUG - prefect.LocalResult | Starting to read result from get_data-88...
[2020-06-11 21:50:19] DEBUG - prefect.LocalResult | Finished reading result from get_data-88...
[2020-06-11 21:50:19] DEBUG - prefect.TaskRunner | Task 'get_data': Handling state change from Pending to Cached
[2020-06-11 21:50:19] DEBUG - prefect.TaskRunner | Task 'get_data': can't set state to Running because it isn't Pending; ending run.
```

## Why is this PR important?
More functionality!

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -672,18 +672,18 @@ class TaskRunner(Runner):
         target = self.task.target
 
         if result and target:
+            raw_inputs = {k: r.value for k, r in inputs.items()}
+            formatting_kwargs = {
+                **prefect.context.get("parameters", {}).copy(),
+                **raw_inputs,
+                **prefect.context,
+            }
+
             if not isinstance(target, str):
-                # Pass all necessary kwargs to target callable
-                raw_inputs = {k: r.value for k, r in inputs.items()}
-                formatting_kwargs = {
-                    **prefect.context.get("parameters", {}).copy(),
-                    **raw_inputs,
-                    **prefect.context,
-                }
                 target = target(**formatting_kwargs)
 
-            if result.exists(target, **prefect.context):
-                new_res = result.read(target.format(**prefect.context))
+            if result.exists(target, **formatting_kwargs):
+                new_res = result.read(target.format(**formatting_kwargs))
                 cached_state = Cached(
                     result=new_res,
                     hashed_inputs={
@@ -691,7 +691,7 @@ class TaskRunner(Runner):
                     },
                     cached_inputs=inputs,
                     cached_result_expiration=None,
-                    cached_parameters=prefect.context.get("parameters"),
+                    cached_parameters=formatting_kwargs.get("parameters"),
                     message=f"Result found at task target {target}",
                 )
                 return cached_state
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -184,7 +184,8 @@ class TaskRunner(Runner):
         # If provided, use task's target as result location
         if self.task.target:
             if not isinstance(self.task.target, str):
-                self.result.location = self.task.target(**context)
+                self.result._formatter = self.task.target
+                self.result.location = None
             else:
                 self.result.location = self.task.target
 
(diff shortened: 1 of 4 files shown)


src/prefect/engine/task_runner.py
```diff
@@ -93,7 +93,10 @@ def __init__(
         else:
             self.result = Result().copy() if result is None else result.copy()
             if self.task.target:
-                self.result.location = self.task.target
+                if not isinstance(self.task.target, str):
+                    self.result.location = self.task.target(**prefect.context)
```
cicdw, line 97: So I think this logic needs to be moved somewhere else deeper in the pipeline; e.g., if I used `lambda **kwargs: kwargs['task_run_count']` that wouldn't work here because `task_run_count` isn't populated until we're within the pipeline.

src/prefect/engine/task_runner.py
```diff
@@ -670,6 +673,9 @@ def check_target(self, state: State, inputs: Dict[str, Result]) -> State:
         target = self.task.target
 
         if result and target:
+            if not isinstance(target, str):
+                target = target(**prefect.context)
```
cicdw, line 677: piggy backing off my comment above, there are more values in context here than when the task runner is initialized -- this call works 👍

alexifm:
Hey, I wanted to try to get this in as quick as possible because I have a suggestion for this PR.  Right now, the results locations support Callables and can be templated with the task inputs as well as the context.  Couldn't the target be set as location even if it's a callable, which would allow it to be rendered on run?
```       
        # If provided, use task's target as result location
        if self.task.target:
            if not isinstance(self.task.target, str):
                self.result.location = self.task.target(**context)
            else:
                self.result.location = self.task.target
```
This would also allow the target to support a Callable that takes the task inputs when rendering in `check_target`.

joshmeek:
@alexifm I'm don't think I follow what you're asking 🤔 

> Couldn't the target be set as location even if it's a callable, which would allow it to be rendered on run?

After this PR both targets and result locations will now support callables that are rendered on run

alexifm:
Okay, what follows is my understanding of things as I interpret the code.  

After this PR, in `TaskRunner.initialize_run`, the result location would get overwritten by a rendered version of the target, but only when it's a callable.  If I supply a string, the templated string is provided as the location instead.  That happens here:
https://github.com/PrefectHQ/prefect/blob/target_callables/src/prefect/engine/task_runner.py#L184-L189

When the task runs, the value for the result is written to location as expected.  But if a Callable target has been passed, the location been rendered already, as noted above.  So the formatting args don't matter unless the Callable for target returns a formatting string.  This is not specified in the documentation.

My suggestion is this:
+ Starting at line 184 in `task_runner.py`, use this code block instead
```
        # If provided, use task's target as result location or formatter
        if self.task.target:
            if not isinstance(self.task.target, str):
                self.result._formatter = self.task.target
                self.result.location = None
            else:
                self.result.location = self.task.target
```
+ Replace line 674 in `task_runner.py` with this code block that mimics how location is rendered at `task_runner.py#L848`:
```
            if not isinstance(target, str):
                raw_inputs = {k: r.value for k, r in inputs.items()}
                formatting_kwargs = {
                    **prefect.context.get("parameters", {}).copy(),
                    **raw_inputs,
                    **prefect.context,
                }
                target = target(**formatting_kwargs)
```

joshmeek:
@alexifm I see exactly what you're saying now and I have made the change. You have a good eye, thanks for pointing this out!

src/prefect/engine/task_runner.py
```diff
@@ -666,6 +672,16 @@ def check_target(self, state: State, inputs: Dict[str, Result]) -> State:
...
+            if not isinstance(target, str):
+                # Pass all necessary kwargs to target callable
+                raw_inputs = {k: r.value for k, r in inputs.items()}
+                formatting_kwargs = {
+                    **prefect.context.get("parameters", {}).copy(),
+                    **raw_inputs,
+                    **prefect.context,
+                }
+                target = target(**formatting_kwargs)
+
```
cicdw, line 684: I think you should create these `formatting_kwargs` in line 675 above and use them here + in lines 685 and 686 for consistency.

joshmeek:
@cicdw Updated

alexifm:
Thanks for putting this together!

PR #2776, opened 2020-06-12
[docs] make it clearer that project is required with Prefect Cloud

jameslamb:
- [x] adds new tests (if appropriate)
- [x] add a changelog entry in the `changes/` directory (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

## What does this PR change?

This PR changes the language in `orchestration/` docs to make it clear that `project_name` is required when registering a flow with Prefect Cloud.

## Why is this PR important?

Thanks for the AWESOME documentation! It's made it really easy to work with Prefect. Following the example from https://docs.prefect.io/orchestration/tutorial/first.html#write-flow, I tried to register the example flow code there with Prefect Cloud (I already have Cloud setup and an agent configured).

```python
import prefect
from prefect import task, Flow

@task
def hello_task():
    logger = prefect.context.get("logger")
    logger.info("Hello, Cloud!")

flow = Flow("hello-flow", tasks=[hello_task])

flow.register()
```

On `flow.register()`, I got this error:

> Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/prefect/core/flow.py", line 1442, in register
    no_url=no_url,
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/prefect/client/client.py", line 598, in register
    "'project_name' is a required field when registering a flow with Cloud. "
TypeError: 'project_name' is a required field when registering a flow with Cloud. If you are attempting to register a Flow with a local Prefect server you may need to run `prefect backend server` first.

Since `project_name` is required, in this PR I propose changing the documentation to be consistent with that requirement:

image

Thanks for your time and consideration

PR #2777, opened 2020-06-12
[docs] standardize example project name

jameslamb:

## Overview

This is a follow-up to #2776 . Working through the [First Flow docs](https://docs.prefect.io/orchestration/tutorial/first.html#register-flow-with-the-prefect-api), I came across this language:

> Prefect Cloud allows users to organize flows into projects. In this case we are using the default Hello, World! Project.
> `flow.register(project_name="Hello, World!")>

The use of the word "default" made me think that I could call that code as-is and that it would attach my flow to a project `Hello, World` that is created automatically when you create a Prefect Cloud account.

But that is not what happened:

> Traceback (most recent call last):
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/prefect/core/flow.py", line 1442, in register
    no_url=no_url,
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/prefect/client/client.py", line 615, in register
    project_name, project_name
ValueError: Project Hello, World! not found. Run `client.create_project("Hello, World!")` to create it.

## What does this PR change?

This PR proposes two changes:

1. Linking to the `Create a Project` documentation (https://docs.prefect.io/orchestration/concepts/projects.html#creating-a-project) from the Flows note about projects for Prefect Cloud (https://docs.prefect.io/orchestration/tutorial/first.html#register-flow-with-the-prefect-api).
2. Standardize the example project name throughout the documentation. Right now half of the docs use `Hello, World!` and half use `My Project!`. That means that depending on the route through the docs that you took to get to https://docs.prefect.io/orchestration/tutorial/first.html#register-flow-with-the-prefect-api, you might not be able to just copy and paste that `flow.register()` call and have it work.

## Why is this PR important?

In my opinion, had the docs reflected the changes in this PR when I first started learning how to register a flow with Prefect Cloud, I would have had a smoother experience and would have moved on to more advanced topics sooner. I think that this PR improves the copy-pasteability of the documentation and reduces the searching that new users have to do to get the examples on the [First Flow page](https://docs.prefect.io/orchestration/tutorial/first.html#register-flow-with-the-prefect-api) working.

### DIFF
--- a/src/prefect/cli/create.py
+++ b/src/prefect/cli/create.py
@@ -19,12 +19,12 @@ def create():
 
     \b
     Examples:
-        $ prefect create project "My Project"
-        My Project created
+        $ prefect create project "Hello, World!"
+        Hello, World! created
 
     \b
-        $ prefect create project My-Project --description "My description"
-        My-Project created
+        $ prefect create project "Hello, World!" --description "My description"
+        Hello, World! created
     """
 
 
--- a/docs/orchestration/agents/local.md
+++ b/docs/orchestration/agents/local.md
@@ -26,7 +26,7 @@ def welcome_logger():
 flow = Flow("Welcome Flow", tasks=[welcome_logger])
 
 # Register Flow with the Prefect API
-flow.register("My Project")
+flow.register(project_name="Hello, World!")
 
 # Spawn a local agent and run in process
 flow.run_agent()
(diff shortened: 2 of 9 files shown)


src/prefect/cli/create.py
```diff
@@ -19,12 +19,12 @@ def create():
...
     Examples:
-        $ prefect create project "My Project"
-        My Project created
+        $ prefect create project "Hello, World!"
+        Hello, World! created
 
     \b
-        $ prefect create project My-Project --description "My description"
-        My-Project created
+        $ prefect create project "Hello, World!" --description "My description"
```
jcrist, line 26: To make the presence/absence of quotes for CLI things more transparent, should we default to a name that wouldn't need to be quoted? Perhaps `hello-world` or something similar?

No strong thoughts, happy either way.

jameslamb: I think this is actually a nice little extra documentation of "you can use names with spaces and special characters but have to quote them"

jcrist: Makes sense to me!

PR #2783, opened 2020-06-14
Fix context passing for Cloud

jlowin:

## What does this PR change?
This pr fixes a minor bug that resulted from #2707. `CloudTaskRunner.run()` begins by instantiating a default context as `context = context or {}`, and the passed context has a default of None. Since no `context` was being passed to this method, [this call](https://github.com/PrefectHQ/prefect/blob/d52cc24981a43ea12b7c6d10c6280aa24aa3efc0/src/prefect/engine/cloud/task_runner.py#L325) would fail when attempting an in-process retry since context was empty.

### DIFF
--- a/changes/pr2783.yaml
+++ b/changes/pr2783.yaml
@@ -1,0 +1,21 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - server
+#   - task
+#   - fix
+#   - deprecation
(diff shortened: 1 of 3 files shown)


jcrist:
Nice catch - do you have a test that would have caught this bug?

jlowin:
One of our Cloud integration tests actually caught it, so the good news is those are planned to move more into the Server codebase for faster response -- however (as you can see from these tests) I think I inadvertently broke something with my fix, so I'll be a little longer with a full PR.

Issue #2091, opened 2020-02-26
Implement a `flat_map` operator that's compatible with DFE


jlowin:
(This can be turned into a PIN)

## Use Case
*Please provide a use case to help us understand your request in context*
"Flat maps" refer to iterated mapping steps where each task in the first step produces an iterable, and we want the second step to map over each of those subitems. For example:

```python
@task
def make_list(n):
    return list(range(n+1))

@task
def add_one(x):
    return x + 1

with Flow("test"):
    lists = make_list.map([1,2,3])
    inc = add_one.map(lists)
```

The user might expect `add_one` to be applied to each item of each list, but Prefect will map it over the lists themselves, resulting in a `TypeError`.

We usually recommend solving this like so:

```python
@task
def flatten(lists):
    return [i for l in lists for i in l]

with Flow("test"):
    lists = make_list.map([1,2,3])
    flat_lists = flatten(lists)
    inc = add_one.map(flat_lists)
```

However, once #2041 introduces DFE for mapped chains, this answer will become less appealing because it will force Prefect back into BFE. The lists will all be made, then the reduce / flatten step will take place, then the increment will take place.

It is desirable to introduce some form of flat-mapping such that users with this type of workflow can take advantage of DFE.

## Solution
*Please describe your ideal solution*
Typically, flat maps are explicitly different from maps. It would probably be beneficial for Prefect to follow the same model and introduce a `flat_map` method. 

```python
with Flow("test"):
    lists = make_list.map([1,2,3])
    inc = add_one.flat_map(lists)
```

The key difference between this and normal mapping is not only that it looks "one level down" in the iterable, but that the number of mapped children tasks in the second step is different than the number in the first step. 

(This may violate a current assumption of #2041, but not in an unsolvable way)

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

PR #2784, opened 2020-06-14
Add support for mapped argument in control flows

davidfhaines:

## What does this PR change?
It adds an optional mapped argument to switch, if/else, case and merge control flow functions and passes to downstream bindings where appropriate.

## Why is this PR important?
Provides support for conditional logic in branches of a mapped pipeline.


jcrist:
Hi @davidfhaines, thanks for the PR. It's not immediately clear to me what your goal here is. I *think* the intent of this PR is to support conditional logic in branches of a mapped pipeline? So the condition itself would be a mapped task, affecting downstream mapped tasks?

If so, what should be the behavior of the following?

```python
cond = is_true.map(some_data)  # a mapped conditional

with case(cond, True, mapped=True):
    some_non_mapped_task()
```

Generally when proposing features like this it's good to provide some example context in the PR or documentation showing how this feature would be used, both with the functional and imperative api. As is right now, I'm not sure exactly what the user-experience should be.

davidfhaines:
> Hi @davidfhaines, thanks for the PR. It's not immediately clear to me what your goal here is. I _think_ the intent of this PR is to support conditional logic in branches of a mapped pipeline? So the condition itself would be a mapped task, affecting downstream mapped tasks?

Thanks @jcrist. Spot on with the intent.

> If so, what should be the behavior of the following?
> 
> ```python
> cond = is_true.map(some_data)  # a mapped conditional
> 
> with case(cond, True, mapped=True):
>     some_non_mapped_task()
> ```

I'd expect the non-mapped task to run in each branch of the mapped pipeline where the condition is `True`. How would you expect the following to behave?

```python
cond = is_true.map(some_data)  # a mapped conditional

with case(cond, True):
    some_mapped_task()
```
 
> Generally when proposing features like this it's good to provide some example context in the PR or documentation showing how this feature would be used, both with the functional and imperative api. As is right now, I'm not sure exactly what the user-experience should be.

Without support for conditional logic in mapped branches, we get:

current_behavior

The only way a specific case would match is if the case value represents the list of _all_ mapped results of the condition. Not sure if that's a common use case, but would have thought it intuitive to support something like the following:

branched_conditional

Unless I've missed something, conditional logic within mapped branches appears hamstrung in its current form.

jcrist:
I see the desire for something like this, and think adding `mapped` kwargs to things like `ifelse`, `switch` and `merge` might make sense. I'm less convinced about `case` though. A `case` block is supposed to be similar to an `if` statement in normal python, with everything under the contextmanager executed conditionally. If find the following behavior non-intuitive and not easy to read:

```python
from prefect import Flow, task, case


@task
def gen_data():
    return [1, 2, 3]


@task
def is_one(x):
    return x == 1


@task
def inc(a):
    return a + 1


@task
def dec(x):
    return x - 1


@task
def echo(x):
    print(f"The value is {x}")


with Flow("test") as flow:
    data = gen_data()

    scalar_value = inc(1)

    with case(is_one.map(data), True, mapped=True):
        # This scalar function is conditionally executed once for every element
        # in `data`, even though it doesn't map over `data` itself - all the
        # calls are the same
        b = dec(scalar_value)

        # This call maps over `data`, and is conditionally executed once for
        # every element in `data`.
        echo.map(data)
```

test


I think my confusion lies with the functional mapping api wanting to work with bulk operations, while `case` reads as a scalar `if` statement. The behavior of `ifelse` and `switch` make more sense to me, as they're also functional, and there's some precedence for other functional conditionals (e.g. `np.where`).

---

For a block-wise api like `case`, I'm not sure what would be best. Perhaps a new construct like a `for` block would make things more readable? Not really thought through:

```python
with Flow("test") as flow:

    a = some_scalar_task()

    with foreach(some_list) as x:
        # All tasks added under this block are implicitly mapped
        # over elements in `some_list`. The `x` value can be used
        # to represent a single element in `some_list` as an argument
        with case(x, 1):
            a_mapped_task(x, a)

# Corresponds to the following python code:
a = some_scalar_task()

for x in some_list:
    if x == 1:
        a_mapped_task(x, a)
```

To support mapping over multiple lists we may want to make `foreach` take `*args`, representing `for` + `zip`.

jcrist:
cc @jlowin, @cicdw for additional api thoughts ^^

jlowin:
I'm actually a little surprised this works because we always envision mapped pipelines as parallel sequences of tasks and this brings some nested fan-out into the mapped pipelines, but after looking through I think it's totally valid and if you were constructing the condition yourself, rather than using our `ifelse` helpers, you could absolutely do this. So from a low-level "does it blend" perspective, 👍 .

I hear @jcrist that our control-flow constructs were always designed to be used   at the "flow-level", when composing tasks, and it feels "wrong" to use `with case():` and pass mapped argument -- though I do really like the `foreach` proposal. This would let users apply any Prefect construct inside their mapped pipeline using the same APIs we supply at the top "flow-level", and that makes mapping more accessible to users. (Could this be as simple as defaulting `mapped=True` when inside a foreach block?)

As we begin to explore complexity in mapping - basically pushing it beyond the simple parallel pipeline definition and involving conditionals, nested fanouts, and even [flat_maps](#2091), I'm wondering if this is actually the first "real" motivation (in my opinion) for introducing `subflows`. If I had a task `f` that represented an entire subflow of arbitrary logic, then I could build that flow and map `f` over an input. This would absolve our API from needing to handle every possibility; if you're doing some complicated, bundle it as a `SubFlowTask` and just apply it to its inputs.

I recognize this is a significant departure from @davidfhaines PR and don't want to distract, so we can continue this conversation elsewhere (or here), so I guess I'll sum up my points as:
- I think @davidfhaines proposal is reasonable because one could always reimplement our control flow blocks with `mapped=True` if we chose not to expose it ourselves, and it would work
- I think passing `mapped=True` to control flows has potential to lead to confusion as users work with more complex (and possibly nested) conditional logic
- I think @jcrist's proposal is a nice way to simplify that complexity by giving users an explicit "mapped result" object to work with
- I wonder if we should take another step and just say "do whatever you want in a subflow, including implementing arbitrary nested conditional logic, then map that logic over your tasks"

jcrist:
Given the discussion above, I think I'd be fine with merging this with the following changes:

- Remove support for `mapped` in `case`. I think this is tricky to understand, and we may want to do something else here.
- Add some tests for `ifelse`/`switch`/`merge` w/ `mapped=True` that use the functional api *only* (no `set_upstream` etc... calls).

jlowin:
I'd endorse that, and I agree the tests should use the functional API as it's the most likely way most users will interact with this.

davidfhaines:
Many thanks @jcrist and @jlowin. Insightful thread. Have made the requested changes.

src/prefect/tasks/control_flow/conditional.py
```diff
@@ -116,10 +124,11 @@ def as_bool(x):
 
     cases = {c: t for c, t in [(True, true_task), (False, false_task)] if t is not None}
     if cases:
-        switch(condition=as_bool(condition), cases=cases)
+        bool_condition = as_bool.copy().bind(condition, mapped=mapped)
```
jcrist, line 127: You could just call `as_bool(condition, mapped=True)`.

src/prefect/tasks/control_flow/conditional.py
```diff
@@ -147,11 +156,15 @@ def merge(*tasks: Task, flow=None) -> Task:
             one of them will contain a result and the others will all be skipped.
         - flow (Flow, optional): The flow to use, defaults to the current flow
             in context if no flow is specified
+        - mapped (bool, optional): Whether the results of these tasks should be mapped over
+                with the specified keyword arguments; defaults to `False`.
```
jcrist, line 160: I don't think this docstring makes sense with the context of `merge` (same for `ifelse` and `switch`). Perhaps something more along the lines of:

```
If true, the `merge` operation will be mapped over the arguments instead of applied directly. Defaults to `False`.
```

tests/tasks/test_control_flow.py
```diff
@@ -164,6 +164,107 @@ def test_merge_imperative_flow():
     assert state.result[c].result == 2
 
 
+def test_mapped_switch_and_merge_functional():
+    with Flow("iterated map") as flow:
+        mapped_result = identity.copy().map(["a", "b", "c"])
+
+        a = identity.copy().bind("a")
```
jcrist, line 171: You can just do `a = identity("a")`. Same for those below.

tests/tasks/test_control_flow.py
```diff
@@ -164,6 +164,107 @@ def test_merge_imperative_flow():
...
+        b = identity.copy().bind("b")
+        c = identity.copy().bind("c")
+
+        switch(condition=mapped_result, cases=dict(a=a, b=b, c=c), mapped=True)
+
+        merge_result = merge(a, b, c, flow=flow, mapped=True)
```
jcrist, line 177: No need to pass in the flow, this is inferred from the context. The `flow` kwarg is there to support using `merge` when not under a `Flow` context manager.

tests/tasks/test_control_flow.py
```diff
@@ -164,6 +164,107 @@ def test_merge_imperative_flow():
...
+
+    state = flow.run()
+
+    assert state.result[a].result == ["a", None, None]
+    assert state.result[b].result == [None, "b", None]
+    assert state.result[c].result == [None, None, "c"]
+    assert state.result[merge_result].result == ["a", "b", "c"]
+
+
+def test_mapped_switch_and_merge_imperative():
```
jcrist, line 187: I don't actually think this test (and the `test_mapped_ifelse_and_merge_imperative` test) add any extra coverage - just testing with the standard functional api here seems fine to me. No harm in having them, but if you feel like making the change it would shorten our test suite a bit.

changes/pr2784.yaml
```diff
@@ -0,0 +1,5 @@
+enhancement:
+  - "Add support for mapped argument in control flows - [#2784](https://github.com/PrefectHQ/prefect/pull/2784)"
+
+contributor:
+  - "[David Haines](https://github.com/davidfhaines)"
```
jcrist, line 5: Missing a newline here.

src/prefect/tasks/control_flow/conditional.py
```diff
@@ -81,6 +81,8 @@ def b_branch():
         - cases (Dict[Any, Task]): a dict representing the "case" statements of the switch.
             The value of the `condition` task will be compared to the keys of this dict, and
             the matching task will be executed.
+        - mapped (bool, optional): Whether the results of these tasks should be mapped over
```
jcrist, line 84: Sorry, when I commented on the `merge` docstring, I should have better indicated that request also applied to `switch` and `ifelse` - could you update those docstrings as well?

davidfhaines: On me. All sorted and thanks again.

src/prefect/tasks/control_flow/conditional.py
```diff
@@ -108,6 +114,8 @@ def ifelse(condition: Task, true_task: Task, false_task: Task) -> None:
         - condition (Task): a task whose boolean result forms the condition for the ifelse
         - true_task (Task): a task that will be executed if the condition is True
         - false_task (Task): a task that will be executed if the condition is False
+        - mapped (bool, optional): If true, the `merge` operation will be mapped over the
```
jcrist, line 117: `merge` -> `ifelse`

src/prefect/tasks/control_flow/conditional.py
```diff
@@ -81,6 +81,8 @@ def b_branch():
         - cases (Dict[Any, Task]): a dict representing the "case" statements of the switch.
             The value of the `condition` task will be compared to the keys of this dict, and
             the matching task will be executed.
+        - mapped (bool, optional): If true, the `merge` operation will be mapped over the
```
jcrist, line 84: `merge` -> `switch`

src/prefect/tasks/control_flow/conditional.py
davidfhaines: 👍

src/prefect/tasks/control_flow/conditional.py
davidfhaines: 👍

jcrist:
I've been thinking about this use case since last week, and submitted a proposal for a new api method for composing complex mapped pipelines in #2846. Hoping to make these kinds of flows easier to write and read.

PR #2796, opened 2020-06-16
Add support non-containerized storage to all environments

joshmeek:

## What does this PR change?
Following #2666 the new option to set an image on the environment's metadata was only compatible with environments that didn't have an assumption of Docker storage. This removes that assumption and updates all environments to follow a similar pattern that Agents use when determining how to make use of a flow. It will use the `image` found in `environment.metadata` if set, otherwise it will fall back to the Docker storage found on the flow.

In order to make this pattern work the PR makes a change to the environment interface `setup` and `execute` functions to both take a flow object instead of `storage` and `flow_file_path`.

## Why is this PR important?
Allows all environments to be used with the new storage/env pattern.


src/prefect/environments/execution/dask/k8s.py
```diff
@@ -162,25 +164,17 @@ def setup(self, storage: "Docker") -> None:  # type: ignore
...
             - **kwargs (Any): additional keyword arguments to pass to the runner
-
-        Raises:
-            - TypeError: if the storage is not `Docker`
         """
-        if not isinstance(storage, Docker):
-            raise TypeError("CloudEnvironment requires a Docker storage option")
-
-        self.create_flow_run_job(docker_name=storage.name, flow_file_path=flow_location)
+        self.create_flow_run_job(docker_name=get_flow_image(flow))
```
cicdw, line 177: It seems like having a dedicated `create_flow_run_job` is not really necessary anymore if all `execute` does is immediately call this?  Thoughts on removing that method?

src/prefect/environments/execution/k8s/job.py
```diff
@@ -94,33 +96,23 @@ def __setstate__(self, state: dict) -> None:
...
             - **kwargs (Any): additional keyword arguments to pass to the runner
-
-        Raises:
-            - TypeError: if the storage is not `Docker`
         """
-        if not isinstance(storage, Docker):
-            raise TypeError("CloudEnvironment requires a Docker storage option")
-
-        self.create_flow_run_job(docker_name=storage.name, flow_file_path=flow_location)
+        self.create_flow_run_job(docker_name=get_flow_image(flow),)
```
cicdw, line 108: Same comment / question

src/prefect/utilities/storage.py
```diff
@@ -0,0 +1,35 @@
+from typing import TYPE_CHECKING
+
+import prefect
+
+if TYPE_CHECKING:
+    from prefect.core.flow import Flow  # pylint: disable=W0611
+
+
+def get_flow_image(flow: "Flow") -> str:
```
cicdw, line 9: Isn't this a duplicate of https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/agent.py ?  We should consolidate

joshmeek: Same logic but this one accepts a Flow object and the one that the agent uses doesn't ever have access to the flow object so it uses a GraphQLResult of a FlowRun

src/prefect/environments/execution/dask/k8s.py
joshmeek: Yeah it could be moved into this function thanks for the callout

changes/pr2796.yaml
```diff
@@ -0,0 +1,3 @@
+breaking:
+  - "Environment `setup` and `execute` function signatures now accept Flow objects - [#2796](https://github.com/PrefectHQ/prefect/pull/2796)"
+  - "`create_flow_run_job` logic has been moved into `execute` for `DaskKubernetesEnvironment` and `KubernetesJobEnvironment` - [#2796](https://github.com/PrefectHQ/prefect/pull/2796)"
```
jcrist, line 3: Missing a trailing endline

src/prefect/environments/execution/k8s/job.py
```diff
@@ -163,26 +149,34 @@ def run_flow(self) -> None:
...
+            flow_run = result.data.flow_run[0]
+
+            flow_data = flow_run.flow
+            storage_schema = prefect.serialization.storage.StorageSchema()
+            storage = storage_schema.load(flow_data.storage)
+
+            ## populate global secrets
+            secrets = prefect.context.get("secrets", {})
+            for secret in storage.secrets:
+                secrets[secret] = prefect.tasks.secrets.PrefectSecret(name=secret).run()
```
jcrist, line 173: It looks like this code is duplicated almost verbatim across multiple environments - perhaps this should be a method on the base `Environment` class? Could this `run_flow` implementation be used on all environments? If not, why not?

joshmeek: Ah great callout, this can definitely be abstracted out

Issue #2501, opened 2020-05-06
Can't combine flows that have the same parameter name


jlowin:
## Description
*A clear description of the bug*
If two flows each have a parameter with the same name, they can not be combined via `flow.update()` due to the uniqueness constraint. This is frustrating for users. (Note this is tangentially related to conversations on subflows and other approaches to flow composition)

## Expected Behavior
*What did you expect to happen instead?*

## Reproduction
*A minimal example that exhibits the behavior.*

python
```
import prefect

@prefect.task
def generate1(limit):
    return list(range(10))[:limit]

@prefect.task
def generate2(limit):
    return list(range(100, 1000))[:limit]

@prefect.task
def echo(x):
    print('Got', x)

with prefect.Flow('flow1') as flow1:
    limit1 = prefect.Parameter('limit')
    data1 = generate1(limit=limit1)
    echo.map(x=data1)

with prefect.Flow('flow2') as flow2:
    limit2 = prefect.Parameter('limit')
    data2 = generate2(limit=limit2)
    echo.map(x=data2)

flow3 = prefect.Flow('flow3')

flow3.update(flow1)
flow3.update(flow2)

```

cicdw:
Adding this reference for stackoverflow visibility: https://stackoverflow.com/questions/60679595/how-does-one-update-a-prefect-flow-with-duplicate-parameters/60685605#60685605

Issue #2723, opened 2020-06-05
Visualization not represented correctly when adding task to amother flow


gryBox:
## Description
Hi - I am trying to join flows on a common parameters.  However in the visualization the `parameter` task is represented by an id.  

![image](https://user-images.githubusercontent.com/16763201/83828400-2d1eda00-a6a6-11ea-8a33-97596760a3a9.png)

@cicdw could this issue be similar to #1793 ?  

## Expected Behavior
An edge from the parameter to all the appropriate tasks.

### Flow1
```
@task
def add_one(a_number: int): 
    return a_number + 1

@task(log_stdout=True)
def print_result(the_result:int): 
    print(the_result)

with Flow("Add") as add_fl:
    a_number = Parameter("a_number", default=1)
    
    the_result = add_one(a_number)
    
    print_result(the_result)

add_fl.visualize()
```
![image](https://user-images.githubusercontent.com/16763201/83828557-90107100-a6a6-11ea-9a03-b453e73f5bd3.png)

### Flow2
```
@task
def sub_one(a_number: int):
    return a_number - 1


with Flow("Subtract") as subtract_fl:
    
    a_number = Parameter("a_number", default=1)
    
    the_result = sub_one(a_number)
    
    print_result(the_result)
    
subtract_fl.visualize()
```
![image](https://user-images.githubusercontent.com/16763201/83828872-45dbbf80-a6a7-11ea-92d1-0ce99b42b77a.png)

### Add Tasks and Edges
```
flow1 = add_fl.copy()
flow2 = subtract_fl.copy()

for task in flow2.tasks:
    task_lst = flow1.get_tasks(slug=task.slug)
    
    if len(task_lst)>0:
        print(f"Duped Tasks on Slug: {task_lst}")
    else:
        flow1.add_task(task)
        
for edge in flow2.edges:
    if edge not in flow1.edges:
        flow1.edges.add(edge)

flow1.visualize()
```
![image](https://user-images.githubusercontent.com/16763201/83828400-2d1eda00-a6a6-11ea-8a33-97596760a3a9.png)

## Environment
prefect version `0.11.5+75.g525fb25a6`

cicdw:
Hi @gryBox - actually this is more subtle than that issue.  Here's what's happening:

- you create flow 1 with a Parameter named `a_number`, which becomes the `slug` (Parameter names must be unique within a Prefect Flow since they are identified by name through kwargs / the API)
[...]

PR #2798, opened 2020-06-17
add parmerge kwarg to update

gryBox:

## What does this PR change?
Adds an optional argument to `flow.update` that allows flows to be updated with duplicate parameters. 

#### Example:
`from prefect import Flow, task, Parameter`

#### Flow1
```
@task
def add_one(a_number: int): 
    return a_number + 1

@task(log_stdout=True)
def print_result(the_result:int): 
    print(the_result)

with Flow("Add") as add_fl:
    a_number = Parameter("a_number", default=1)
    
    the_result = add_one(a_number)
    
    print_result(the_result)
```
##### Flow2
```
@task
def sub_one(a_number: int):
    return a_number - 1


with Flow("Subtract") as subtract_fl:
    
    a_number = Parameter("a_number", default=1)
    
    the_result = sub_one(a_number)
    
    print_result(the_result)
```

```
flow_to_merge = copy.deepcopy(flow2)
flow1.update(flow_to_merge, parmerge=True, validate=True)
```
image

## Why is this PR important?
There are times when two non-trivial flows with duplicate parameters must be updated/merged for optimal execution.  Currently this can only be done if:

1. The flows are in the same `.py` file [slack discussion](https://prefect-community.slack.com/archives/CL09KU1K7/p1590761710157000?thread_ts=1590760508.153800&cid=CL09KU1K7) or in some global file.

2. Parameters are written in the imperative style 

Closes: #2501


changes/pr2798.yaml
```diff
@@ -0,0 +1,5 @@
+enhancement:
+  - "flow.update` accepts the optional `kwarg` `parmerge` that allows flows to be updated with common `Parameters` - [#2501](Can't combine flows that have the same parameter name)"
```
cicdw, line 2:
```suggestion
  - "flow.update` accepts the optional `kwarg` `parmerge` that allows flows to be updated with common `Parameters` - [#2501](https://github.com/PrefectHQ/prefect/issues/2501)"
```
FYI these are markdown links

src/prefect/core/flow.py
```diff
@@ -568,30 +568,54 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
             )
         return edges
 
-    def update(self, flow: "Flow", validate: bool = None) -> None:
+    def update(
+        self, flow: "Flow", parmerge: bool = None, validate: bool = None
```
cicdw, line 572: I think we should clarify this kwarg name to `merge_parameters` and default it to `False` for clarity.

src/prefect/core/flow.py
```diff
@@ -568,30 +568,54 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
...
+        Take all tasks and edges in another flow and add it to this flow.  
+        When `parmerge` is set to`True` -- Duplicate parameters in the unput `flow`
+        are replaced with those in the flow being updated. 
+    
         Args:
-            - flow (Flow): A flow which is used to update this flow
-            - validate (bool, optional): Whether or not to check the validity of the flow
-
+            - flow (Flow): A flow which is used to update this flow.
+            - parmerge (bool, optional): Merge two graphs on parameters
```
cicdw, line 581:
```suggestion
            - merge_parameters (bool, optional): Whether to merge parameters that share the same name; defaults to `False`
```

src/prefect/core/flow.py
```diff
@@ -568,30 +568,54 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
...
+            - validate (bool, optional): Whether or not to check the validity of the flow.
+        
         Returns:
             - None
         """
+
         for task in flow.tasks:
-            if task not in self.tasks:
+
+            duped_task = self.get_tasks(slug=task.slug) if parmerge else list()
```
cicdw, line 590: I see what you're getting at here but I think we should make the code more explicit that this is solely for Parameters, so something like:
```python
if merge_parameters and isinstance(task, Parameter):
    duped_params = [p for p in self.parameters() if p.name == task.name]
else:
    duped_params = []
```

src/prefect/core/flow.py
```diff
@@ -568,30 +568,54 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
...
-            if edge not in self.edges:
-                self.add_edge(
-                    upstream_task=edge.upstream_task,
-                    downstream_task=edge.downstream_task,
-                    key=edge.key,
-                    mapped=edge.mapped,
-                    validate=validate,
-                )
+
+            if any(
```
cicdw, line 599: I need to think more about this edge manipulation logic - we'll need to move away from slugs (for reasons unrelated to this PR)

gryBox: @cicdw Perhaps this PR needs to wait until you finish with `slugs`. #2531? Or is the issue with `slugs` larger than that?

cicdw: @gryBox yup just that PR 👍

gryBox: Well #2531 broke that edge logic :) .  I need to investigate further and understand how to use the new slugs.  Its not obvious to me why Parameters have a slug and Tasks do ... but only if they are set by the user.  Except to ensure parameter uniqueness.

gryBox:
Hi @cicdw 

> One other edge case that's on my mind: what should we do if two flows have Parameters with the same name by different defaults / different values for `required`?

That is a difficult question. I am not exactly sure there is a good answer for that,  But the solution I implemented follows [networkx's union](https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.operators.binary.union.html#networkx-algorithms-operators-binary-union) logic -- Reversed

##### Networkx
> If a graph attribute is present in both G and H the value from H is used.

##### Current implementation with `merge parameters`
If  G the flow being updated  and H the flow being added have the same parameter -- Flow H 's parameter is removed altogether. And therefore G takes precedence.

My thinking is to implement this naive approach first.  See if people use it and how it stands the test of time before adding more customization. 

I too originally thought that this question needed to be addressed before `merge_parameters`  could be functional.  However your thinking on disjoint, thankfully, made me realize that handling default values and renaming parameters is unneeded complexity at this point. So would adding the flow's name to tasks on the flow they were originally bound to.  i.e {flow_name}-{task-name} or `slug`.  Still playing around with things.

src/prefect/core/flow.py
```diff
@@ -606,30 +606,55 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
...
         return edges
 
-    def update(self, flow: "Flow", validate: bool = None) -> None:
+    def update(
+        self, flow: "Flow", merge_parameters: bool = False, validate: bool = None
+    ) -> None:
         """
-        Take all tasks and edges in another flow and add it to this flow
+        Take all tasks and edges in another flow and add it to this flow.
+            When `merge_parameters` is set to`True` -- Duplicate parameters in the unput `flow`
```
lauralorenz, line 614:
```suggestion
            When `merge_parameters` is set to`True` -- Duplicate parameters in the input `flow`
```

typo

src/prefect/core/flow.py
```diff
@@ -606,30 +606,55 @@ def chain(self, *tasks: Task, validate: bool = None) -> List[Edge]:
...
-                    upstream_task=edge.upstream_task,
-                    downstream_task=edge.downstream_task,
-                    key=edge.key,
-                    mapped=edge.mapped,
-                    validate=validate,
-                )
+
+            if edge.upstream_task.name in self.slugs.values():
+
+                upstream_task = self.get_tasks(name=edge.upstream_task.name)[0]
```
cicdw, line 646: within Prefect internals we need to be as precise as possible; using `name` to check against `slug` and then `get_tasks` with name could lead to all sorts of weird bugs.

I'm not 100% sure what the intent is here - it seems that this might be the place where the "merge" is happening, but it's not dependent on the value of `merge_parameters`?

gryBox: @cicdw Thank you for taking the time to review.

>  using name to check against slug and then get_tasks with name could lead to all sorts of weird bugs.

Yeah. Not ideal. It rests on two assumptions (a) Prefect does not change the slug naming convention for parameters and (b) Parameters will always be unique. But I think I can find some other way to compare slug vs slug.  Let me know if you have any suggestions.

> I'm not 100% sure what the intent is here - it seems that this might be the place where the "merge" is happening, but it's not dependent on the value of merge_parameters?

Correct.  It does not depend on `parameter_merge`. Because:  In the case of duplicate parameters and `parameter_merge` equal to `False` an error is raised before getting to edges,  If `merge_parameter` is `True`  than there are no parameter slug conflicts anymore .  But the parameter's downstream task needs to be replaced with the parameter of the flow being updated. #2723.  It's similar to `flow,replace` .

cicdw: Thanks for the context @grybox - I’ll get back to you with some thoughts

cicdw: Hey @gryBox - I did a quick pass on your branch (I hope that's OK!) and am going to merge -- thanks for the contribution!

gryBox: @cicdw  Thanks for all the help.  Can't wait to see if anybody else uses it.

cicdw: No problem! Marvin will stop greeting you on your PRs now that you have official "Contributor" status ;) 

I think @bnaul has expressed interest in this functionality before so he might be an early user

PR #2804, opened 2020-06-17
Fix LocalEnvironment execute on flow object

joshmeek:

## What does this PR change?
Fixes the `LocalEnvironment` execute. The environment already has access to the flow object so it doesn't need to attempt this storage `get_flow` call.

## Why is this PR important?
Fixes some use cases when using the `LocalEnvironment`


jcrist:
> we should clarify our environment names / purposes - this environment isn't really supported for Cloud deployments IMO and is kind of a duplicate of "RemoteEnvironment"

Can you clarify this? I was planning on deprecating `RemoteEnvironment` in favor of `LocalEnvironment` since the functionality is almost identical:

- `LocalEnvironment` - run a flow-runner locally
- `KubernetesJobEnvironment` - run a flow-runner as a kubernetes job
- `FargateTaskEnvironment` - run a flow-runner as a fargate task

PR #2819, opened 2020-06-18
Cleanup `LocalDaskExecutor`

jcrist:
- Fixes `LocalDaskExecutor` to only compute once. Previously this would recompute tasks, sometimes many many times (the comment saying "twice" was being optimistic).
- Adds support for setting task names in the `LocalDaskExecutor`.

Also a few misc cleanups.

### DIFF
--- a/src/prefect/engine/executors/dask.py
+++ b/src/prefect/engine/executors/dask.py
@@ -13,6 +13,9 @@ if TYPE_CHECKING:
     from distributed import Future
 
 
+__any__ = ("DaskExecutor", "LocalDaskExecutor")
+
+
 # XXX: remove when deprecation of DaskExecutor kwargs is done
 _valid_client_kwargs = {
     "timeout",
@@ -214,7 +217,7 @@ class DaskExecutor(Executor):
 
         # set a key for the dask scheduler UI
         if task_name:
-            dask_kwargs.update(key=f"{task_name}-{str(uuid.uuid4())}")
+            dask_kwargs.update(key=f"{task_name}-{uuid.uuid4().hex}")
 
         # infer from context if dask resources are being utilized
         dask_resource_tags = [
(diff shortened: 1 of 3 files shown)


src/prefect/engine/executors/dask.py
```diff
@@ -276,31 +279,52 @@ def wait(self, futures: Any) -> Any:
...
+        """Context manager for initializing execution."""
         # import dask here to reduce prefect import times
-        import dask
+        from dask.callbacks import Callback
+
+        class PrefectCallback(Callback):
+            def __init__(self):  # type: ignore
+                self.cache = {}
+
+            def _start(self, dsk):  # type: ignore
```
cicdw, line 315: So is `_start` called at the beginning of computation for every task?

cicdw: I guess I'd be curious to learn about how this works exactly, I don't think I fully understand it.  (whether here or offline)

jcrist: Here's the docs on callbacks: https://docs.dask.org/en/latest/diagnostics-local.html#custom-callbacks, does this help? The callback here lets us modify the graph before computation starts - `_start` is called once on `compute()` setup, `_posttask` is called after every task.

cicdw: I did read those docs but didn't realize that `_start` was called on the first call to `compute` when the graph is fully known (I couldn't quite tell what "state" was referring to, but that might be because my brain is in `distributed` mode and I forgot no work occurs until `compute` is called).

I guess I'm still confused though because I don't see how `self.cache` is populated at this stage?  From reading that doc on `_posttask` as well I assumed it wouldn't be populated until after the task finishes (which necessary occurs _after_ compute has been called).

cicdw: Oh I think I'm being silly - this callback protects us against multiple calls to `compute`, not reruns within a single `compute` call 👍

jcrist: Yah. Only a single `cache` exists during the lifetime of a `Executor.start()` context, and is built up over repeated `compute` calls with (actually `get` calls, which is one level lower). The `_start` callback lets us swap out already computed tasks for their results, preventing them from being recomputed. The `_posttask` callback adds the results of complete tasks to the `cache` to be reused later.

cicdw: Awesome thank you!

PR #2831, opened 2020-06-19
Generate better dask task key names

### LINKED ISSUE #2821: Mapped task names are not grouped on dask / bokeh
## Description
See #2813 for original context.

From the 0.12.0 release task names weren't being inferred / passed properly to the dask scheduler. That was fixed for single tasks after #2814 was merged. However after just testing out current git head it seems like mapped tasks are copied over _with_ their index in the name which results in `bokeh` looking like:

```
mapped_task [0] 1/1
mapped_task [1] 1/1
mapped_task[2] 1/1
...
```

## Expected Behavior

Previous behavior in 0.11.0 where mapped tasks were all named the same thing and so `bokeh` had a single progress bar for them all.

## Reproduction

```python
from distributed import Client

@task
def add_one(x):
    return x + 1

with Flow("test") as flow:
    values = add_one.map(range(10))

cluster = LocalCluster()
client = Client(cluster)

print(f"Dashboard available at: {client.dashboard_link}")

flow.run(executor=DaskExecutor(address=cluster.scheduler_address))
```

## Environment

`prefect`: current git head
`dask[bag]==2.18.1`
`distributed==2.18.0`
`bokeh==2.0.2`

jcrist:
Dask infers if different tasks do similar things with some custom
key-name splitting heuristics (it's *mostly* splitting by `-`, but not
entirely). The previous key names generated by mapped tasks didn't play
nice with this heuristic, resulting in the dask UI not grouping tasks
appropriately. This fixes that.

Fixes #2821

### DIFF
--- a/tests/engine/test_flow_runner.py
+++ b/tests/engine/test_flow_runner.py
@@ -1551,6 +1551,10 @@ def test_constant_tasks_arent_submitted_when_mapped(caplog):
 
 
 def test_flow_runner_provides_consistent_extra_context():
+    """Check that the `extra_context` provided to an executor is consistent.
+
+    The `extra_context` is used by the dask executor to generate good key names
+    and add task resource requirements"""
     extra_contexts = []
 
     class MyExecutor(Executor):
--- a/tests/engine/test_flow_runner.py
+++ b/tests/engine/test_flow_runner.py
@@ -1548,3 +1548,33 @@ def test_constant_tasks_arent_submitted_when_mapped(caplog):
 
     ## to be safe, ensure '5' isn't in the logs
     assert len([log.message for log in caplog.records if "99" in log.message]) == 0
+
+
+def test_flow_runner_provides_consistent_extra_context():
+    extra_contexts = []
+
+    class MyExecutor(Executor):
+        def submit(self, fn, *args, extra_context=None, **kwargs):
+            extra_contexts.append(extra_context)
+            return fn(*args, **kwargs)
+
+        def wait(self, x):
+            return x
+
+    @prefect.task
+    def inc(x):
+        return x + 1
+
+    @prefect.task
+    def do_sum(x):
+        return sum(x)
+
+    with Flow("test") as flow:
+        a = inc(1)
+        b = inc.map(range(3))
+        c = do_sum(b)
+
+    flow.run(executor=MyExecutor())
+
+    for ctx in extra_contexts:
+        assert ctx.keys() == {"task_name", "task_index", "task_tags"}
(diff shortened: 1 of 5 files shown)


jcrist:
> This LGTM but we still don't actually test that the dask keys are set properly that would have caught the minor run_task bug we released with 0.12.0; the tests only test that keys are set assuming we actually pass the right extra_context (spoiler alert: we weren't passing the correct extra context previously). I think it'd be worthwhile to add a real test for this if possible.

You're talking about the bug I fixed in the previous PR, where `FlowRunner` was passing the context expected by the `Executor`?

cicdw:
Yea that's right, where all the keys were all submitted with a `run_task-` prefix - it would be easy to alter `extra_context`, have all tests pass, and still not have the right dask keys.

jcrist:
Done

tests/engine/test_flow_runner.py
```diff
@@ -1548,3 +1548,33 @@ def add(x):
 
     ## to be safe, ensure '5' isn't in the logs
     assert len([log.message for log in caplog.records if "99" in log.message]) == 0
+
+
+def test_flow_runner_provides_consistent_extra_context():
```
cicdw, line 1553: we might want to add a docstring to this test that it's really about the dask executor, because a dev who edits a key from `extra_context` might edit this test to pass without realizing it affects the task keys in dask.  Sorry to be nitpicky, I just think we have a lot of tests that are testing _around_ the intended behavior without testing it directly 
image

jcrist: Since there is a test for the dask executor that will fail if those change, changing one without fixing the other shouldn't be possible. But sure. I feel like that image describes much of the current tests for prefect.

cicdw: Interesting, I find that view contradictory - I think that this key thing has a lot of scattered hyper-specific unit tests that are still easy to make pass without ever knowing whether the real objective is achieved via a test (which we used to have).

jcrist: switched to an integration test

tests/engine/test_flow_runner.py
```diff
@@ -1580,5 +1586,4 @@ def do_sum(x):
 
     flow.run(executor=MyExecutor())
 
-    for ctx in extra_contexts:
-        assert ctx.keys() == {"task_name", "task_index", "task_tags"}
+    assert key_names == {"inc", "do_sum"}
```
cicdw, line 1589: awesome thank you!

PR #2838, opened 2020-06-22
make sure `environment` and serialized flow in client.register() match

jameslamb:

## What does this PR change?

In `client.register()`, the client checks if the flow is using `Docker` storage and, if so, gets the image from that storage and puts it on `flow.environment.metadata["image"]`. This PR proposes ~doing that *before* capturing the serialized flow with `flow.serialize()`.~ updating `serialized_flow` after doing that.

## Why is this PR important?

This will help users avoid a mistake that I just made, which took a bit of time to debug.

I've been using `KubernetesAgent` and `KubernetesJobEnvironment`, and experimenting with different types of storage. From a previous attempt that was using `S3` storage, I had an image set in the environment's `metadata` left over in my `KubernetesJobEnvironmnet` code:

```python
env = KubernetesJobEnvironment(
    job_spec_file="prefect-flow-run.yaml",
    metadata={
       "image": "prefecthq/prefect:0.12.0-python3.7"
   }
)

flow.storage = Docker(
    base_image="prefecthq/prefect:0.12.0-python3.7",
    image_name="prefect-testing",
    image_tag="1.1",
    registry_url="localhost:32000"
)
```

When I registered this flow and ran it, the flow runs failed with an error like this

> Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/prefect/environments/execution/base.py", line 131, in run_flow
  File "/usr/local/lib/python3.8/site-packages/prefect/environments/storage/docker.py", line 244, in get_flow
    with open(flow_location, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/opt/prefect/flows/my-flow.prefect'

After some time spent debugging, I realized that jobs created by `KubernetesAgent` were using the image `prefecthq/prefect:0.12.0-python3.7` (from the `environment`), not the new image built from my storage.

I believe that the changes in this PR will eliminate this error without breaking any expected behavior. The effect of this PR is that `{"metadata": {"image` set in the `environment` is ignored if you use `Docker` storage.

#### Environment info

I observed this behavior with `prefect` 0.12.0, running an agent that used image `prefecthq/prefect:0.12.0-python3.7`.

### DIFF
--- a/changes/pr2838.yaml
+++ b/changes/pr2838.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Make sure image from Docker storage is always used with KubernetesJobEnvironment - [#2838](https://github.com/PrefectHQ/prefect/pull/2838)"
--- a/changes/pr2838.yaml
+++ b/changes/pr2838.yaml
@@ -1,2 +1,5 @@
 fix:
   - "Make sure image from Docker storage is always used with KubernetesJobEnvironment - [#2838](https://github.com/PrefectHQ/prefect/pull/2838)"
+
+contributor:
+ - "[James Lamb](https://github.com/jameslamb)"
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -662,7 +662,7 @@ class Client:
         # Set Docker storage image in environment metadata if provided
         if isinstance(flow.storage, prefect.environments.storage.Docker):
             flow.environment.metadata["image"] = flow.storage.name
-            serialized_flow = flow.serialize(build=False)  # type: Any
+            serialized_flow = flow.serialize(build=False)
 
         # If no image ever set, default metadata to all_extras image on current version
         if not flow.environment.metadata.get("image"):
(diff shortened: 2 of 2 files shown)


jameslamb:
Ah it looks like I misunderstood how this is working. From [the failing tests](https://app.circleci.com/pipelines/github/PrefectHQ/prefect/4735/workflows/b01a0a94-048b-4da3-9cf0-90cf0b607661/jobs/86217/steps), it looks like `storage.image_name` gets updated when you call `Docker.build()`, and that the client is relying on that update.

https://github.com/PrefectHQ/prefect/blob/fc65bc160c7eae04dfb3668e6ed5a41dd9a315c4/src/prefect/environments/storage/docker.py#L287-L290

Maybe something like https://github.com/PrefectHQ/prefect/pull/2838/commits/1491d1529441f5b45977cfee3e33b5b3563c4539 would be better

joshmeek:
Nice fix @jameslamb! Could you also add a `fix` for this to the `changes/` directory according to this https://github.com/PrefectHQ/prefect/blob/master/changes/README.md?

jameslamb:
> Nice fix @jameslamb! Could you also add a `fix` for this to the `changes/` directory according to this https://github.com/PrefectHQ/prefect/blob/master/changes/README.md?

added in https://github.com/PrefectHQ/prefect/commit/00997417bb7278ee4e68c08e0c116a540137faf2

Issue #2281, opened 2020-04-06
Implement `prefect.tasks.kubernetes.RunNamespacedJob` that capstones normal job usage


lauralorenz:
## Use Case
*Please provide a use case to help us understand your request in context*
The Kubernetes Job tasks in our task library mimic the Kubernetes API, but an expected 'normal' use case of them is composed of several steps, namely creating a namespaced job, polling for it to complete, and deleting the job at the end. Right now no task in the task library knows how to poll for job status, and also no task strings the create/poll/delete steps together. (This is a spin-off/replacement for #1724)

## Solution
*Please describe your ideal solution*
Create a new task library task in `prefect.tasks.kubernetes` called `RunNamespacedJob` that creates, polls, and deletes a job. It may be composed of the run methods of the other related kubernetes tasks.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*
We previously considered exposing a kwarg on `CreateNamespacedJob` to poll for completion, but discussed among the Core team that a custom task that wrapped creating, polling, AND cleaning up separately a) fit the common use case better and b) kept the granular tasks isolated from config bloat unrelated to their granular use.

pcieslinski:
I'm going to take care of creating this new `RunNamespacedJob` task. I will keep you updated on progress. :)

pcieslinski:
I missed that this `api_client`, which is used in Kubernetes tasks, is mocked in other tests for these tasks. I will correct it and everything should be fine. The logic of operation is the same, only the form of the code changed. ;)

PR #2845, opened 2020-06-23
Create utility function for getting Kubernetes client

pcieslinski:

## What does this PR change?
While working on resolving this issue: https://github.com/PrefectHQ/prefect/issues/2281, I discovered a lot of duplicate code and tests. I managed to extract into a separate utility function the logic responsible for creating and authorizing the client for Kubernetes, which is used in the task library.

`get_kubernetes_client` was created in a similar way to `get_boto_client`. Thanks to this, I wanted to be consistent with the rest of the codebase.

Example of use:
```python
from prefect.utilities.kubernetes import get_kubernetes_client

api_client = get_kubernetes_client(resource="pod", kubernetes_api_key_secret="KUBERNETES_API_KEY")
```

When the logic of creating and authorizing a client for Kubernetes was extracted into a separate function, some of the tests from the tasks were removed. Now all testing is done at the level of function `get_kubernetes_client`.

## Why is this PR important?
Generally, this PR strives to stick to good software development practices and DRY principle. Thanks to this refactor we will have gathered in one place the logic to create and authorize the client for Kubernetes. We can use this part of the code throughout the codebase.

By removing duplicate tests, the project code will be easier to maintain.


pcieslinski:
I missed that this `api_client`, which is used in Kubernetes tasks, is mocked in other tests for these tasks. I will correct it and everything should be fine. The logic of operation is the same, only the form of the code changed. ;)

joshmeek:
Thanks for the big PR @pcieslinski! Yeah the client is mocked heavily for these tests and with this new utility could definitely be cleaned up / possibly separated into a test fixture.

pcieslinski:
@joshmeek I know that this PR has unfortunately grown quite a lot and can be difficult to check, but it is mainly because the same code was used in many places. I have improved the tests for all Kubernetes tasks and put the mocked Kubernetes client in the fixture. :)

Issue #2792, opened 2020-06-15
Add cloud integration tests to CI


jcrist:
Currently much of our prefect-cloud support test suite is unit tests, with mocking-out the cloud api. While nice for smoke tests, this still leaves room for bugs to slip through, forcing us to do some manual QA before each release.

To improve our test coverage, it would be good to add integration tests against cloud. One way to do this would be to add a build in our test matrix that operates against cloud. We can use circle-ci secrets to hold secret things (e.g. a tenant token) and test cloud operations against a tenant using a temporary local agent.

To support local testing, we'll want a way to run these tests locally as well. The simplest thing I can think of is having the integration tests run against the current configuration (i.e. if I was running them locally, they'd run in my current tenant (if configured), otherwise they'd skip), and our CI test run would then configure things to use the appropriate tenant before running the test suite.

PR #2849, opened 2020-06-24
respect executor from environment

jameslamb:
*created from [this conversation in Prefect Community Slack](https://prefect-community.slack.com/archives/CL09KU1K7/p1592925843192100)*

## What does this PR change?

This PR proposes changing the behavior of `Environment.run_flow()`. Today, if you call this method on an instance of `Environment` or any child class of it, `run_flow()` will use the executor in `self.executor` if that attribute has been set.

https://github.com/PrefectHQ/prefect/blob/191f429cdb1c48cde1b2636e2fa04e24f94dfb79/src/prefect/environments/execution/base.py#L133-L135

That strategy means that despite the changes from #2805 , it is not currently possible to use `KubernetesJobEnvironment` and `DaskExecutor` together. That is because flows that use `KubernetesJobEnvironment` end up being run with `KubernetesJobEnvironment().run_flow()`

https://github.com/PrefectHQ/prefect/blob/191f429cdb1c48cde1b2636e2fa04e24f94dfb79/src/prefect/environments/execution/k8s/job.py#L247

`KubernetesJobEnvironment`'s constructor sets `self.executor` to the default executor from `prefect`'s config if one isn't passed.

https://github.com/PrefectHQ/prefect/blob/191f429cdb1c48cde1b2636e2fa04e24f94dfb79/src/prefect/environments/execution/k8s/job.py#L71-L74

Since `KubernetesJobEnvironment` inherits `run_flow()` from `Environment`, that means that the environment's `self.executor` will always be set and equal to the value from the config.

So if you use something like I tried today:

```python
KubernetesJobEnvironment(
    executor=DaskExecutor(
        cluster_class="prefect.engine.executors.DaskExecutor"
    ),
    job_spec_file="prefect-flow-run.yaml",
)
```

Then you'll never get that `DaskExecutor`. Flows will run with the default, which is often `LocalExecutor`.

This PR proposes changing that behavior, so that if the `environment` in the flow has an `executor`, `Environment.run_flow()` uses that.

## Why is this PR important?

Without this PR, the `executor` you pass to `KubernetesJobEnvironment` or `FargateTaskEnvironment` is ignored when flows are run.

If I'm right, that means that without this PR, the only way to use one of these environments with a non-default executor like `DaskExecutor` is to maybe set the default executor in a [user config](https://docs.prefect.io/core/concepts/configuration.html#user-configuration) (I have not tried this).

I think this is extra important because the use of a different executor is silent...in my case, the job environment happened to be sized appropriately to run my flow, so even though I wanted `DaskExecutor`, when I got `LocalExecutor` the flow happily ran and completed. If my job environment had been poorly sized for the work my flow does, I would have probably hit an out of memory error and it wouldn't have been immediately obvious that this PR's root cause was to blame.

## Questions for Reviewers

* I've tested this manually in my local setup, but I think it should get Prefect unit tests so the bug isn't re-introduced. Could you point me to relevant tests in the repo that I could use as a reference to write unit tests for this change?
* Would it be better to remove the "if no executor given, set `self.executor` to the default" logic in `FargateTaskEnvironment` and `KubernetesJobEnvironment`? I'm not sure what the implications of that would be, but it seems like it would be a cleaner separation of responsibilities.

### DIFF
--- a/changes/pr2849.yaml
+++ b/changes/pr2849.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Change Environment.run_flow() to prefer executor from flow's environment - [#2849](https://github.com/PrefectHQ/prefect/pull/2849)"
--- a/changes/pr2849.yaml
+++ b/changes/pr2849.yaml
@@ -1,2 +1,5 @@
 fix:
   - "Change Environment.run_flow() to prefer executor from flow's environment - [#2849](https://github.com/PrefectHQ/prefect/pull/2849)"
+
+contributor:
+ - "[James Lamb](https://github.com/jameslamb)"
--- a/docs/orchestration/execution/fargate_task_environment.md
+++ b/docs/orchestration/execution/fargate_task_environment.md
@@ -95,7 +95,7 @@ The container dictionary above will be changed during setup:
 [
     "/bin/sh",
     "-c",
-    "python -c 'import prefect; prefect.Flow.load(prefect.context.flow_file_path).environment.run_flow()'",
+    "python -c 'import prefect; prefect.environments.execution.load_and_run_flow()'",
 ]
 ```
 
--- a/docs/orchestration/execution/k8s_job_environment.md
+++ b/docs/orchestration/execution/k8s_job_environment.md
@@ -57,7 +57,7 @@ In the above YAML block, `flow-container` will be changed during execution:
 - `command` and `args` will take the form of:
 
 ```bash
-/bin/sh -c "python -c 'import prefect; prefect.Flow.load(prefect.context.flow_file_path).environment.run_flow()'"
+/bin/sh -c "python -c 'import prefect; prefect.environments.execution.load_and_run_flow()'",
 ```
 
 - `env` will have some extra variables automatically appended to it for Cloud-based Flow runs:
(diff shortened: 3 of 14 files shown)


jcrist:
Ah, good catch @jameslamb! I didn't notice that this was being run on a new instance of the environment. This should really be a `classmethod` instead to avoid issues like this. @joshmeek, will anything break if we switch `run_flow` to a classmethod (I think the answer is no)?

joshmeek:
@jcrist I think we're good to make the switch 👍

jcrist:
Great. @jameslamb we're hoping to get a release out in the next day or two - mind if I push to this PR to finish it up?

*Aside: this is another great argument for adding some integration tests (see #2792), since this was missed by testing.*

jameslamb:
@jcrist not a problem, push away!

For the record, I like the classmethod idea 😀

jcrist:
Updated. It made more sense to pull `run_flow()` out into its own function. The code here could be cleaner (with respect to the sometimes-present `run()` method on environments), but to keep backwards compatibility and get this release out I've used a mixin class. Once the old environments have been deprecated this can definitely be cleaned up and streamlined.

jameslamb:
Looks awesome, thank you! This became bigger than my original PR, so you can definitely remove my name from the `changes/` entry.

jcrist:
You found the bug and instigated the change, so I'm happy to leave your contribution in.

jcrist:
This also fixes #2850, which was due to the same issue.

jameslamb:
ok fair enough, thanks!

PR #2859, opened 2020-06-24
Raise a more informative TypeError when calling execute CLI command

joshmeek:
In the cases where a flow is pickled in one python version (3.7) and then attempted to run in another (3.8) a TypeError will be raised. This adds an error message to do a better job of informing users.

### DIFF
--- a/src/prefect/cli/execute.py
+++ b/src/prefect/cli/execute.py
@@ -72,9 +72,9 @@ def cloud_flow():
             environment.setup(flow)
             environment.execute(flow)
     except Exception as exc:
-        msg = """Flow failed to deserialize — please ensure that all packages / imports
-        that the flow relies on are available and that the flow is being deserialized in
-        an environment that has the same Python version that it was built in.
+        msg = """Flow failed to deserialize — please ensure all packages / imports the
+        flow relies on are available and the flow is being deserialized in an environment
+        that has the same Python version it was built in.
         {}""".format(
             repr(exc)
         )
--- a/src/prefect/cli/execute.py
+++ b/src/prefect/cli/execute.py
@@ -66,20 +66,18 @@ def cloud_flow():
             secrets[secret] = PrefectSecret(name=secret).run()
 
         with prefect.context(secrets=secrets, loading_flow=True):
-            try:
-                flow = storage.get_flow(storage.flows[flow_data.name])
-            except TypeError:
-                raise TypeError(
-                    """Flow failed to deserialize — please check that the python version
-                    / imports / etc. are correct. This is often due to a flow being
-                    serialized in one version of Python and deserialized in another."""
-                )
+            flow = storage.get_flow(storage.flows[flow_data.name])
             environment = flow.environment
 
             environment.setup(flow)
             environment.execute(flow)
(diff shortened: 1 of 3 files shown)


src/prefect/cli/execute.py
```diff
@@ -66,7 +66,14 @@ def cloud_flow():
             secrets[secret] = PrefectSecret(name=secret).run()
 
         with prefect.context(secrets=secrets, loading_flow=True):
-            flow = storage.get_flow(storage.flows[flow_data.name])
+            try:
+                flow = storage.get_flow(storage.flows[flow_data.name])
+            except TypeError:
```
cicdw, line 71: I vote to catch all exceptions types here, not just `TypeError`

joshmeek: We do catch all exceptions below it. In this case of a TypeError we specifically catch and raise a different message

joshmeek: Another option is checking the type of exception that was raised and then handling it in this special case where we want to raise a more informative error

src/prefect/cli/execute.py
```diff
@@ -72,7 +72,12 @@ def cloud_flow():
             environment.setup(flow)
             environment.execute(flow)
     except Exception as exc:
-        msg = "Failed to load and execute Flow's environment: {}".format(repr(exc))
+        msg = """Flow failed to deserialize — please ensure all packages / imports the
+        flow relies on are available and the flow is being deserialized in an environment
+        that has the same Python version it was built in.
+        {}""".format(
+            repr(exc)
+        )
```
jcrist, line 80: When building messages like this, the indent on the newline will persist, resulting in an oddly formatted exception message. I like to use python's implicit multiline strings:

```
message = (
    "These strings "
    "are concatenated together "
    "at compile time, "
    "so there is no "
    "runtime cost."
)
```

but you could also wrap everything in `textwrap.dedent`.

src/prefect/cli/execute.py
joshmeek: Catching all errors here and handling in the same way will be confusing because other exceptions could be raised here on flow retrieval such as downloading from an S3 bucket. If we want to be explicit about catching all pickle exceptions then this should be moved inside of each storage class's individual `get_flow` function

jcrist: That type error isn't the only error you could get from a mismatched python version. For example:

```python
In [4]: value_serialized_with_pickle_5 = b'\x80\x05\x95\t\x00\x00\x00\x00\x00\x00\x00\x8c\x05hello\x94.'

In [5]: pickle.loads(value_serialized_with_pickle_5)  # load in python 3.7, only pickle <= 4 supported
ValueError                                Traceback (most recent call last)
<ipython-input-5-218e273bf89d> in <module>
----> 1 pickle.loads(value_serialized_with_pickle_5)

ValueError: unsupported pickle protocol: 5
```

I think we either want to provide a generic error, or move the handling of this closer to the `pickle.loads` call. Catching a `TypeError` here feels wrong.

joshmeek:
Closing PR to move pickle error handling deeper into the stack 👍

PR #2865, opened 2020-06-25
make storage directory in Docker storage configurable (fixes #2861)

### LINKED ISSUE #2861: Make prefect directory inside Docker storage configurable
## Current behavior

Today when you build a `Docker` storage, `prefect` creates a directory for itself called `/opt/prefect` inside the image.

https://github.com/PrefectHQ/prefect/blob/467bc5b1dcd83716bd896eff549f6bceb59da8cf/src/prefect/environments/storage/docker.py#L469

That directory is used to store configuration files and flows.

## Proposed behavior

I'd like to propose making `prefect`'s directory inside `Docker` storage configurable. Ideally I'd like to be able to do this or something similar:

```python
flow.storage = Docker(
    prefect_directory="/home/my-custom-user/prefect"
)
```

If this is picked up, I believe it would mean the following:

* add a new keyword argument to `Docker` (https://github.com/PrefectHQ/prefect/blob/467bc5b1dcd83716bd896eff549f6bceb59da8cf/src/prefect/environments/storage/docker.py#L81), which defaults to `/opt/prefect` for backwards compatibility
* replace all hard-coded `/opt/prefect` in https://github.com/PrefectHQ/prefect/blob/467bc5b1dcd83716bd896eff549f6bceb59da8cf/src/prefect/environments/storage/docker.py with a reference to that attribute, `self.prefect_directory`

## Example

The current state can lead to issues if the `base_image` you're using runs as a user that does not have permission to write to `/opt`. I just encountered this situation using a base_image where the default user cannot write to `/opt`

> Step 5/9 : RUN mkdir -p /opt/prefect/
 ---> Running in eb30b890ad2b
 mkdir: cannot create directory ‘/opt/prefect/’: Permission denied

Making this configurable could remove an existing limitation that might currently lead people to run the image build process as `root`.

## Notes for Reviewers

If the maintainers here are interested in this change, I'd be happy to contribute it.

jameslamb:

## What does this PR change?

This PR implements my proposal from #2861 to allow users to configure the location inside `Docker` storage where flows and other files are stored.

This PR sets that location to `/opt/prefect` (the previous hard-coded value) by default, so it should not break any existing code.

## Why is this PR important?

See the discussion in #2861 . This argument gives users finer control over the images produced by `Docker` storage, and is useful when working with images that have a non-root `USER` set which doesn't have permission to write to `/opt`.


src/prefect/environments/storage/docker.py
```diff
@@ -71,6 +71,8 @@ class Docker(Storage):
             client. [Documentation](https://docker-py.readthedocs.io/en/stable/tls.html#docker.tls.TLSConfig)
         - build_kwargs (dict, optional): Additional keyword arguments to pass to Docker's build step.
             [Documentation](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.build.BuildApiMixin.build)
+        - prefect_directory (str, optional): A path inside the Docker image that `prefect` uses to store files like
+            configurations and flows. If this is not given, `/opt/prefect` will be used.
```
jcrist, line 75:
```suggestion
        - prefect_directory (str, optional): Path to the directory where prefect configuration/flows
             should be stored inside the Docker image. Defaults to `/opt/prefect`.
```

tests/environments/storage/test_docker_storage.py
```diff
@@ -515,6 +527,32 @@ def test_create_dockerfile_with_weird_flow_name():
             )
 
 
+def test_create_dockerfile_with_weird_flow_name_custom_prefect_dir():
+    with tempfile.TemporaryDirectory() as tempdir_outside:
```
jcrist, line 531: Instead of using `tempfile.TemporaryDirectory`, I recommend using pytest's `tmpdir` fixture: https://docs.pytest.org/en/stable/tmpdir.html

*(I know `tempfile.TemporaryDirectory()` is used elsewhere already, we're slowly migrating towards making better use of fixtures as tests are updated)*

jameslamb: oh sure, I can do that. I just copied this from another test and added in the `prefect_directory` kwarg.

jameslamb: Ok I just changed the two tests in this PR that were using `tempfile.TemporaryDirectory()` to `tmpdir`.

Thanks for showing me that! I think I'll start using it in other projects.

PR #2867, opened 2020-06-25
Version Lock Improvements

### LINKED ISSUE #2730: Only supply versions when setting `SUBMITTED` or `RUNNING` states
## Current behavior
In order to streamline the version-locking process, version-locks will only be respected when a version number is passed. Therefore, enforcing a version lock becomes a client-side decision. 

By default, Core should only pass versions when setting `SUBMITTED` states (in the agent) or `RUNNING` states.

Today, it passes an incremented version number on EVERY state change.

See https://coda.io/d/dATbvpGIUqU/Design-Document-Version-Locking-v2_suh0Q#_luJXb for more detail. 

## Proposed behavior

## Example

### LINKED ISSUE #2731: Gracefully recover from version lock errors
## Current behavior

When setting a state results in a version-lock error, the attempt is aborted and progress stops.

## Proposed behavior

Related to #2730, whenever a version-lock error is encountered, the Client should instead query for the latest state and return it so process can continue. This will enable graceful fallbacks from the latest known state. The only exception is if the latest state is one that *requires* a version lock, like RUNNING or SUBMITTED, because version-locked states indicate run-once semantics.

## Example

joshmeek:

Closes #2730 by only supplying versions when setting `Running` and `Submitted` states.

Closes #2731 by gracefully handling version lock errors. Now it queries for the current state of task/flow runs and uses that state to proceed. In the case of a `Running` state being found it ends the run because it means the flow is running elsewhere.


src/prefect/engine/cloud/flow_runner.py
```diff
@@ -118,8 +119,22 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
-                flow_run_id=flow_run_id, version=version, state=cloud_state
+                flow_run_id=flow_run_id,
+                version=version if cloud_state.is_running() else None,
+                state=cloud_state,
             )
+        except VersionLockError as exc:
+            state = self.client.get_flow_run_state(flow_run_id=flow_run_id)
+
+            if state.is_running():
+                raise ENDRUN(state=state)
```
cicdw, line 130: we should probably issue a log right before raising "Flow run already in a running state" or something to that effect

src/prefect/engine/cloud/task_runner.py
```diff
@@ -111,10 +112,22 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
-                version=version,
+                version=version if cloud_state.is_running() else None,
                 state=cloud_state,
                 cache_for=self.task.cache_for,
             )
+        except VersionLockError as exc:
+            state = self.client.get_task_run_state(task_run_id=task_run_id)
+
+            if state.is_running():
+                raise ENDRUN(state=state)
```
cicdw, line 123: same - log would be useful

src/prefect/engine/cloud/task_runner.py
```diff
@@ -111,10 +112,22 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
+
+            self.logger.debug(
+                "Version lock encountered for task {}, proceeding with state {}...".format(
+                    self.task.name, type(state).__name__
+                )
+            )
+            new_state = state
```
cicdw, line 130: we probably should populate the result here as well (it's generally assumed that results are populated at the beginning of a run and this could occur anywhere).  I think something like:
```python
new_state = state.load_result(self.task.result)
```
should work (you'll want to confirm my syntax).

Not overcomplicate, but this might need to live inside it's own `try / except` as well, and if an error occurs we probably need to raise `ENDRUN`

joshmeek: Yeah good call 👍

src/prefect/client/client.py
```diff
@@ -974,16 +980,41 @@ def update_task_run_heartbeat(self, task_run_id: str) -> None:
...
+        version: int = None,
     ) -> "prefect.engine.state.State":
         """
         Sets new state for a flow run in the database.
 
         Args:
             - flow_run_id (str): the id of the flow run to set state for
-            - version (int): the current version of the flow run state
             - state (State): the new state for this flow run
+            - version (int, optional): the current version of the flow run state
```
jlowin, line 1017: Could you note that the version is optional, but can be supplied to enforce version-locking

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -130,7 +145,8 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
             state.state = old_state  # type: ignore
             raise ENDRUN(state=state)
 
-        prefect.context.update(flow_run_version=version + 1)
+        if version is not None:
```
jlowin, line 148: This is actually not quite right - the version will still increment in the server, so Core should continue incrementing the version even if it's not providing it. The reason for this is that if the version isn't always incremented, then clients that DO want to enforce locking can't actually tell if they have the latest version or not.

src/prefect/engine/cloud/task_runner.py
joshmeek: Instead of `self.task.result` I'm going to do the `self.result` since at initialization of the task runner it is set up to use the Flow's result if none was set on the task:
https://github.com/PrefectHQ/prefect/blob/71d40d7a4d3f97d0ca562b80b9033646e1a2c9ae/src/prefect/engine/task_runner.py#L83-L86

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -145,8 +148,7 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
             state.state = old_state  # type: ignore
             raise ENDRUN(state=state)
 
-        if version is not None:
-            prefect.context.update(flow_run_version=version + 1)
+        prefect.context.update(flow_run_version=(version or 0) + 1)
```
jlowin, line 151: It would be really unusual but do you want to add the same logic [here](https://github.com/PrefectHQ/prefect/blob/version_locking/src/prefect/engine/cloud/task_runner.py#L156)?

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -118,8 +119,25 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
+                    "Version lock encountered and flow is already in a running state."
+                )
+                raise ENDRUN(state=state)
+
+            self.logger.debug(
+                "Version lock encountered, proceeding with state {}...".format(
+                    type(state).__name__
+                )
+            )
+            new_state = state
```
jlowin, line 140: Should this `ENDRUN` as well, to abort further processing of this task? For example, let's say it tries to enter a `RUNNING` state, but learns that the "true" state is Failed. I don't think we want to enter the `RETRYING` part of this task runner's pipeline, which we might if we just return the upstream state. Therefore I'm wondering if anytime we learn that the true state is different from what we thought, we just ENDRUN with it and let the next task proceed. 

Not 100% sure if I'm thinking through everything here - what do you think?

src/prefect/engine/cloud/task_runner.py
```diff
@@ -110,10 +111,37 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
+                raise ENDRUN(state=state)
+
+            self.logger.debug(
+                "Version lock encountered for task {}, proceeding with state {}...".format(
+                    self.task.name, type(state).__name__
+                )
+            )
+
+            try:
+                new_state = state.load_result(self.result)
```
jlowin, line 136: I'm curious - should this ENDRUN rather than return the new state, to abort further processing of this task? For example, let's say it tries to enter a RUNNING state, but learns that the "true" state is Failed. I don't think we want to enter the RETRYING part of this task runner's pipeline, which we might if we just return the upstream state. Therefore I'm wondering if anytime we learn that the true state is different from what we thought, we just ENDRUN with it and let the next task proceed.

Not 100% sure if I'm thinking through everything here - what do you think?

PR #2868, opened 2020-06-25
More context keys

### LINKED ISSUE #2662: Key error when using task_run_id in a target template
## Description
Similar to one of the issues in #2640  where `filename` throws a key error when used in a template, so does `task_run_id`.

```
(py37moc) :~/prefect_guide$ /miniconda3/envs/py37moc/bin/python /home/ilivni/prefect_guide/tst_map.py
0.11.3
[2020-05-27 18:56:17] INFO - prefect.FlowRunner | Beginning Flow run for 'blah'
[2020-05-27 18:56:17] INFO - prefect.FlowRunner | Starting flow run.
[2020-05-27 18:56:17] INFO - prefect.TaskRunner | Task 'return_list': Starting task run...
[2020-05-27 18:56:17] ERROR - prefect.TaskRunner | Unexpected error: KeyError('task_run_id')
  File "/miniconda3/envs/py37moc/lib/python3.7/site-packages/prefect/engine/runner.py", line 48, in inner
  File "/miniconda3/envs/py37moc/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 660, in check_target
    if result.exists(target, **prefect.context):
  File "//miniconda3/envs/py37moc/lib/python3.7/site-packages/prefect/engine/results/local_result.py", line 123, in exists
    return os.path.exists(os.path.join(self.dir, location.format(**kwargs)))
KeyError: 'task_run_id'
```

## Reproduction
```
import prefect 

print(prefect.__version__)

lcl_res = LocalResult(dir="~/prefect_guide/results/{flow_name}")

@task(target="{task_name}-{task_run_id}",
)
def return_list():
    return [1, 2, 3]
@task(target="{task_name}/{map_index}.prefect")
def mapped_task(x):
    return x + 1
with Flow("blah", result=lcl_res) as flow:
    mapped_task.map(return_list)


st = flow.run()
flow.visualize(flow_state=st)

```

## Environment

cicdw:

## What does this PR change?
This PR introduces more context keys into flow runs that are run locally (using `flow.run`) so that things like `flow_run_id` still have local meaning and can be used for templating when testing interactively.

## Why is this PR important?
Closes #2662 and ensures a compatible experience when running locally vs. against an API backend.


src/prefect/engine/task_runner.py
```diff
@@ -167,9 +167,9 @@ def initialize_run(  # type: ignore
             task_run_count=run_count,
             task_name=self.task.name,
             task_tags=self.task.tags,
-            task_slug=self.task.slug,
         )
         context.setdefault("checkpointing", config.flows.checkpointing)
+        context.setdefault("task_slug", self.task.slug)
```
jcrist, line 172: Under what circumstances would `task_slug` not be set above in the flow runner? Alternatively, why should `task_slug` be set in the flow runner instead of always in the task runner?

src/prefect/core/flow.py
```diff
@@ -943,12 +951,17 @@ def _run(
             ## begin a single flow run
             while not flow_state.is_finished():
                 runner = runner_cls(flow=self)
+                task_ctxts = kwargs.pop("task_contexts", {})
+                for t in self.tasks:
+                    task_ctxts.setdefault(t, dict())
+                    task_ctxts[t].update(task_run_id=str(uuid.uuid4()))
```
jcrist, line 957: IIUC this mutates the external dicts passed in as `task_contexts`, which might lead to unexpected behavior:

```python
task_contexts = {task: {...}}
flow.run(task_contexts=task_contexts)
# task_contexts is now mutated from inside the `run` call
```

Perhaps:

```suggestion
                task_ctxts = kwargs.pop("task_contexts", {})
                for t in self.tasks:
                    ctx = task_ctxts.setdefault(t, {}).copy()
                    ctx.update(task_run_id=str(uuid.uuid4()))
```

src/prefect/engine/task_runner.py
cicdw: Your thoroughness is impressive.  With the new slugs mechanism, slugs are _only_ reliable if the Flow provides them so I actually should remove this and adjust the tests accordingly.

cicdw:
> Perhaps a method that both use to populate a standard set of keys for flow/task runs

Yea I could imagine a common function / method that does this; there are two keys that don't make much sense in local runs and that I didn't include in this PR: 
- `task_run_version`
- `flow_run_version`

We _could_ add them but for them to be meaningful we'd also have to increment them with each state change which is fine but also adding more code to maintain for not much benefit (I would be extremely surprised if people used those values for templating).

gryBox:
Hi @cicdw - Do you think `slug` can be added?  It would be convenient to template `location` when a flow has the same task more than once with different data bindings.

```
location= "{parameters[node_input_dict][term_id]}/{slug}"

 make_node_fl.result = LocalResult(
            dir=f"~/{artifact_results_dir}", location=location
        )
```
This would insure unique task result storage.

cicdw:
Hey @gryBox - `task_slug` is something you can already template with 👍

gryBox:
Hi @cicdw 
>  `task_slug` is something you can already template with

 I tried that.  The result filename is `None`.
```
import prefect
print(prefect.__version__)

from prefect import task, Parameter, Flow

@task
def get_subset_kvs(input_dict: dict, keys_to_get: list):
    
    sub_dict = dict((k, input_dict[k]) for k in keys_to_get if k in input_dict)
    
    return sub_dict

with Flow("Make-Node") as flow:
    
    ##### Inputs
    input_dict = Parameter("input_dict", default={"name":"a", "shape":"square"})
    field1 = Parameter("field1", default=["name"])
    field2 = Parameter("field2", default=["shape"])

    #### Script
    id_kvs = get_subset_kvs(
        input_dict=input_dict,
        keys_to_get=field1
        )

    attr_kvs = get_subset_kvs(
        input_dict=input_dict,
        keys_to_get=field2
    )

location= "{parameters[input_dict][name]}/{task_slug}"

flow.result = LocalResult(
    dir=f"~/test_results", location=location
)

flow.run()
```
image

cicdw:
That will be resolved by this PR @gryBox

gryBox:
Works as expected! 
image

PR #2891, opened 2020-06-30
[WIP] - Add `Resource` tasks

jcrist:
This PR adds a new task type to prefect - `Resource` tasks. Resource tasks are similar to normal tasks, except for three key differences:

- The output of a `Resource` task is *never* persisted anywhere, it is always recreated.
- The output of a `Resource` task is *never* serialized between processes (e.g. dask workers), it is always recreated
- `Resource` tasks may optionally define a cleanup step that is run to cleanup the resource when it is no longer needed. Note that cleanup is best effort (it won't be run if the process segfaults, but usually will otherwise), and that in the case of failure the only reporting is a log - the `resource` task will still show as successful

These features make `resource` tasks useful for creating/managing things like database connections, s3 clients, etc... The easiest way to create one is with the `resource` decorator.

```python
from prefect import resource

@resource
def my_client(username, password):
    return MyClient(username, password)
```

A resource task is used like any other task - you can pass it arguments (here you probably want to use a `PrefectSecret` to pass in the `password`) and use the result in downstream tasks.

If the resource needs a cleanup step, you can `yield` the resource instead of returning it. In that case, the code block after the `yield` statement is run to cleanup the resource when it is no longer needed.

```python
@resource
def my_client(username, password):
    # code before the `yield` statement is run on setup
    client =  MyClient(username, password)

    yield client

    # code after the `yield` statement is run on cleanup
    client.close()
```

Since most connection-like objects in python are contextmanagers, this would more commonly be written as

```python
@resource
def my_client(username, password):
    with MyClient(username, password) as client:
        yield client
```

Much of the complications with this PR result from hacking around the `Result` interface to get things to work the way I wanted them to. There may be a cleaner way to accomplish the same semantics.


jcrist:
This works as intended and has *some* tests now, but I no longer think this is a good idea (at least as written).

Resources technically do two things:
- Let users create objects in a task that aren't pickleable (e.g. boto clients)
- Let users optionally register a cleanup for these tasks that will be run on a best-effort basis (good for things where failure to cleanup isn't detrimental, like boto clients).

The first is really to deal with a failure in the library (there's no fundamental reason why boto clients shouldn't be pickleable, they just aren't). The second could be done using other python constructs (destructors, `weakref.finalize`, ...). Other than that, they're just tasks by another name.

However, if a user doesn't read the documentation closely, they might write code where the creation of the resource has side effects. For example:

```python
@resource
def dask_cluster():
    with KubeCluster() as cluster:
        yield cluster


with Flow("example") as flow:
    cluster = dask_cluster()
    some_task(cluster)
    another_task(cluster)
```

This would run fine if using a `LocalExecutor`, but if using some executor that has multiple processes, the `dask_cluster` function would be run multiple times - resulting in multiple clusters! Also, if the cleanup step fails the only notice will be a log message, not a failed task in the flow run. We could document this, but users may not read the docstring fully before using it. This feels like a footgun - seeing that a `resource` looks fundamentally like a `@contextmanager` decorated function I might assume they have the same semantics when they don't.

---

If we step back and look at the user issue, the main goal is to allow users to return objects from tasks that can't be pickled and should always be recreated. From a user view this might be spelled a few ways:

```python
# A new kwarg to `task`
@task(pickleable=False):
def boto_client(...):
    return ...


# A utility decorator to wrap a task function in
@task
@never_pickle_always_recreate
def boto_client(...):
    return ...
```

I slightly like the first one better, but maybe it should have a better keyword name. I could see a future where we also want to support returned values from tasks that we never pickle *and* never recreate - we ensure all tasks that use that return value run on the same node. Perhaps an enum of `pickle={"ok", "recreate", "not ok"}` (with better names).

Either way, to accomplish the goal here I don't think we need to introduce a new concept, especially not one that may easily be used incorrectly.

cicdw:
I'm very intrigued by the `pickleable=False` idea; my initial thought is "I love it!" and then when I take a step back I realize there's a lot here.  We would need to document that the task effectively re-runs on each dask worker it is needed on; this immediately raises the question of triggers / dependencies / how to handle failure.  This then makes me wonder how much of a "Task" this really is.  Without giving it deep thought, it seems that https://github.com/PrefectHQ/prefect/pull/2913 will potentially subsume this pattern.

jcrist:
> Without giving it deep thought, it seems that #2913 will potentially subsume this pattern.

I think #2913 will solve the issue for resources where the above would be a bad idea (things that are stateful, or whose cleanup is actually important, like a temporary Spark cluster). #2913 won't solve the issue for task results that just can't be pickled, like boto clients. Now I'm wondering if we want to provide a "local cache" concept, and not treat these as tasks at all. Will think on things.

Anyway, this PR should be closed, as I don't think we want to use this idea as written. Closing.

jcrist:
> Will think on things.

I've been thinking about this in the background for a bit, and have some further ideas. The following pattern might be nice, and would be quick to work up - but it may be confusing to users. The synopsis is "don't return unpickleable things from tasks, recreate them inside of tasks instead". Potential example code:

```python
from prefect.utilities.something import cached

@cached # <- the new bit of api, a local cache that is cleared after each flow run
def api_client(foo, bar):
    return ...

@task
def step_one(foo, bar):
    client = api_client(foo, bar)
    client.do_something(...)

@task
def step_two(foo, bar):
    client = api_client(foo, bar)
    client.do_something_else(...)

with Flow("example") as flow:
    step_one(foo, bar)
    step_two(foo, bar)
```

If `step_one` and `step_two` both run on the same machine, only one client will be created. Otherwise two separate clients will be created. Since there's no `api_client` task, all places where an api client could be created are explicitly marked and failure to recreate (if not already cached) will be raised in predictable locations. We deal with un-pickleable results by avoiding ever making them results.

The downside is that the code to get a client (in this example) has to be duplicated in multiple tasks, since it can never be returned from a task as a shared result. If we wanted that kind of behavior, then the work done in this PR above would be closer to what we'd want, noting that it comes with all of the potentially confusing behaviors as documented above.

cicdw:
I like this proposal - one of the things I've been thinking about is how to handle the state of such a "resource", and I reached the same conclusion as you: that failure to load the "resource" (or cached object in your new lingo) should result in failure of the task that relies on the object.

jcrist:
The idea above in https://github.com/PrefectHQ/prefect/pull/2891#issuecomment-661383853 has been implemented in #3139.

PR #2898, opened 2020-07-01
Introduce `flatten` operator that supports mapping

### LINKED ISSUE #2091: Implement a `flat_map` operator that's compatible with DFE

jlowin:

## What does this PR change?

This PR introduces a `flatten` operator that, when applied to a nested sequence, removes one layer of nesting. It is compatible with all possible Prefect upstream types, from constants, to mapped tasks, to reduced tasks.  

Closes #2091 

The most-requested use case is to enable a `flatmap` semantic. This comes up frequently in mapped pipelines where a mapped task returns a sequence for each child. For example, task `A` returns a list with three items (`[1, 2, 3]`). Task `B` maps over `A` and returns `list(range(x))` for each item in `A`. This yields `[[0], [0, 1], [0, 1, 2]]`. We now want to map over each item produced by B, but Prefect's existing map functionality will only apply to the "first" level of B's output - three lists. By using a `flatten` operator, we can "unnest" B's output into a list of length 6, and map over that:

```python
from prefect import task, Flow, flatten

@task
def A():
    return [1, 2, 3]

@task
def B(x):
    return list(range(x))

@task
def C(y):
    return y + 100

with Flow('flat map') as f:
    a = A()
    b = B.map(x=a)
    c = C.map(y=flatten(b))

state = f.run()

state.result[a].result # [1, 2, 3]
state.result[b].result # [[0], [0, 1], [0, 1, 2]]
state.result[c].result # [100, 100, 101, 100, 101, 102]
```

## How does it work?

The insight for this PR is that transformations like `unmapped` and `flatten` (and `mapped`, which this PR trivially adds for symmetry) are properties of an edge, one of Prefect's lowest-level objects, but are usually specified in Prefect's highest-level APIs. Therefore, this PR extends the special-cased `unmapped` class by formalizing a concept of `EdgeAnnotations` of which `unmapped` is an example. 

These `EdgeAnnotations` work exactly as `unmapped` does today, except that they are generalized and support composition - `unmapped(flatten(x))` works, as does `flatten(mapped(x))`. 

`EdgeAnnotations` are not computed immediately. Instead, they are accounted for at runtime, just as mapping is performed today. 

If an edge is marked as `flattened`, the `FlowRunner` takes some special steps. If the upstream task is "vanilla", then it simply flattens the upstream result. If the upstream is a mapped task, then the mapped children are each flattened and recombined into a single sequence. 

Reluctantly, I have concluded that this implementation of `flatten` can not support DFE. This is because our implementation of DFE proceeds as soon as it knows the "width" or cardinality of the mapped layer. With traditional mapping, the width of the map never changes, so we always know that the `ith` child output of level `X` will be passed to the `ith` child of level `X+1`. With this insight, we can generate DFE steps as soon as the upstream cardinality is known. With a flatmap, the width can change at any time

With a flat map, the cardinality can change. The `ith` child of level `X` will not necessarily correspond to the `ith` child of `X+1` because the `i-1th` child could produce more than one output. With flattening semantics, it's possible for a map with a single child to feed into a map of 10 children; or a map of 100 children to feed into a map of 2. Therefore, we must wait to see how many outputs are produced by the map before flattening them and proceeding. This means that DFE can not proceed THROUGH the flatten. 

HOWEVER, the flatten operator is more efficient than a naive reduce-and-flatten that users might have implemented on their own. When paired, for example, with Dask, no data needs to be moved or gathered to the flow runner in order for the flatmap to proceed. 

## Why is this PR important?
🥞


src/prefect/utilities/executors.py
```diff
@@ -377,3 +393,41 @@ def prepare_upstream_states_for_mapping(
...
+    new_state.result = state._result.from_value(  # type: ignore
+        state.result[index]
+    )
+    return new_state
+
+
+def flatten_upstream_state(upstream_state: "State") -> "State":
+    new_state = copy.copy(upstream_state)
+    try:
+        flattened_result = [y for x in upstream_state.result for y in x]
```
bachsh, line 410: this basically duplicates the data. For a heavy mapped task (say, with 1000 tasks), that could be a problem, couldn't it? I'm not sure the rest of the code supports it, but would it work to change this to a generator instead?

src/prefect/core/flow.py
```diff
@@ -824,30 +854,30 @@ def _sorted_tasks(self, root_tasks: Tuple[Task, ...] = None) -> Tuple[Task, ...]
 
     def set_dependencies(
         self,
-        task: object,
-        upstream_tasks: Iterable[object] = None,
-        downstream_tasks: Iterable[object] = None,
-        keyword_tasks: Mapping[str, object] = None,
+        task: Any,
+        upstream_tasks: Iterable[Any] = None,
+        downstream_tasks: Iterable[Any] = None,
```
bachsh, line 859: unrelated to this PR but since you've changed it: shouldn't these be `Task` and `Iterable[Task]`? What am I missing?

jlowin: `as_task()` is called on these inputs, both in this function (for the main "task") and in the Edge constructor, so this function can accept any object.

src/prefect/utilities/executors.py
jlowin: Generators are generally incompatible with serialization between tasks, so this would require a tight coupling between this feature and iteration within the task. I would suggest that we look at that for a future enhancements - perhaps tasks that are reducing over a mapped upstream can opt to process the upstream input as a sequence rather than a full in-memory reduction.

jcrist:
This is great to see!

I hope to give this a deeper review sometime tomorrow, but a surface naming thing feels a bit off to me. `unmapped` is a past-tense verb, while `flat` is an adjective. Not a big deal, but perhaps a better name would be `flattened` (past tense) or a verb like `concat`/`concatenate`/`flatten`? There's some precedence here for similar methods:

- `concat` is used by [`toolz`](https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.concat) and [clojure](https://clojuredocs.org/clojure.core/concat)
- `flatten` is used by [`underscore`](https://underscorejs.org/#flatten)

jlowin:
@jcrist I'm totally on board with you, it does feel funny (though interestingly I view `unmapped` as an adjective (a thing is mapped or unmapped)). `flatten()` gets my vote, as its a more imperative voice: 

```
# FLATTEN y and MAP over it
a_task.map(x=flatten(y))
```

jlowin:
Cases to investigate:
- completely empty sequences (what do we expect)?
- cases where a mapped child errors and all others return sequences (how does flat handle it? does it properly pass the error to a single downstream child?)

jlowin:
Following conversation with Jim, going to adopt `flatten()` for the operator and `Edge.flattened` for the edge attribute

joshmeek:
Also noting, I ran a few tests in relation to empties and errors:

- "empty" flat maps over things like `[]`, `[[]]`, `[[None]]` all worked as expected (I believe) where the `None` in the third one was treated as a thing but the other two empty lists were not.
- errors appeared to be handled properly and similar to regular maps with TriggerFailed errors being used where appropriate

^ can easily see this by editing example with something like:

```python
@task
def B(x):
    l = list(range(x))
    l.append(None)
    return l
```

and it returns:

```
[100, TypeError("unsupported operand type(s) for +: 'NoneType' and 'int'"), 100, 101, TypeError("unsupported operand type(s) for +: 'NoneType' and 'int'"), 100, 101, 102, TypeError("unsupported operand type(s) for +: 'NoneType' and 'int'")]
```

jcrist:
> I know this doesn't necessarily fall under the use of a flat map since it isn't an issue with nested iterables but should we better handle cases where we’re not using “nested” flat mapping? For example if I flat map over

This might be what you're saying, but just to clarify -  I think the `["abc"]` example does the right thing here, and we might want to provide a better error message for the `[123]` case?

joshmeek:
@jcrist Yep that is it 🙂

jlowin:
I cleaned up some of the error-handling and added tests, but there's a tricky situation that comes up. 

Suppose a task maps over two flattened inputs (to make it interesting), and one of those inputs is a nested list (`[[1]]`) and the other is not (`[1]`). The result is that the task DOES perform its map, but returns a length-zero result of `[]`. This is surprising to me, as I thought the un-flattenable input would cause an error.

However, flattening takes place during upstream edge processing, prior to the task running. Therefore, if flattening fails, the way it's expressed is that the upstream state is changed to a Failed state. When a task maps, it proceeds as long as there is at least one Success state. Since there is, mapping continues but produces a length-zero result.

I added a test for this situation [here](https://github.com/PrefectHQ/prefect/pull/2898/files#diff-7fdd625124cf8dae542b38cbfaf273ebR1177). I'm not sure it's "wrong" but it was surprising that the map worked; I was expecting an informative error. In hindsight, I'm not sure there's a natural place for that error to go -- the upstream task didn't fail (since flattening takes place after it succeeds) and mapped tasks ARE compatible with failed upstreams as long as there's at least one successful input. So maybe this is actually the correct behavior...?

joshmeek:
I haven't looked into if I know a place that it would belong but I did some runs and having 
```python
.map(a=flatten([[1]]), b=flatten([[1]]))
```
run but return an empty map felt very unexpected. If we had a place to raise the error then that would feel like a better behavior.

jlowin:
@joshmeek do you have a more complete example you can share? I'm actually confused, I think that should work - you have two nested lists that should both yield `[1]` for mapping over

joshmeek:
@jlowin Oops typed that wrong, it was one nested and one not. Edited

jlowin:
@cicdw gave me a clean path forward here: `flatten()` doesn't fail if you give it something that isn't a nested list; it just returns it. Semantically, `flatten()` is not an ACTION we want taken; it's a DESCRIPTION of how we want the output to be. Therefore when you call `flatten()` on something, no error is raised if it's not a flattenable sequence -- however the task that receives it will raise an error if it's value is incompatible with its expectations.

The major effect of this is that now all of the edge cases we've raised in this thread behave "as expected" including `flatten([123])`

src/prefect/utilities/tasks.py
```diff
@@ -182,6 +184,17 @@ def preprocess(a: Any) -> "prefect.Task":
     return res
 
 
+# DEPRECATED backward-compatible import
```
joshmeek, line 187: @jlowin Could you add this deprecation about the new import path to the changes even though it's small

src/prefect/utilities/edges.py
```diff
@@ -0,0 +1,118 @@
...
+                return [1, 2, 3]
+
+        with Flow("My Flow"):
+            add = AddTask()
+            ll = ListTask()
+            result = add(x=mapped(ll), y=5)
+        ```
+    """
+
+    def __init__(self, task: "prefect.core.task.Task"):
```
jcrist, line 50: This type annotation isn't really correct, since `task` could also be an `EdgeAnnotation` object, which isn't a `Task`.

src/prefect/engine/flow_runner.py
```diff
@@ -438,21 +435,42 @@ def extra_context(task: Task, task_index: int = None) -> dict:
...
                     if not edge.mapped and isinstance(upstream_states[edge], Mapped):
-                        upstream_mapped_states[edge] = mapped_children.get(
-                            edge.upstream_task, []
-                        )
+                        children = mapped_children.get(edge.upstream_task, [])
+
+                        # if the edge is flattened, then we need to wait for the mapped children
+                        # to complete and then flatten them
+                        if edge.flattened:
+                            children = executor.wait(children)
```
jcrist, line 468: IIUC, this will pull retrieve the mapped results locally, defeating the point of computing the lengths remotely. I *think* you can safely remove this `wait` call and everything will still work.

jlowin: Thanks for spotting that! I think this got left behind from an earlier attempt, and I've removed it.

PR #2904, opened 2020-07-01
Feature/docker kwargs

limx0:

## What does this PR change?
Simply adds a container_name arg to `CreateContainer`. This allows starting/stopping/removing using a name rather than container_id, which is often more convenient.  

## Why is this PR important?
I have two separate flows that start and stop containers, this allows me to refer create/delete containers based on some name.

### DIFF
--- a/changes/pr2904.yaml
+++ b/changes/pr2904.yaml
@@ -1,0 +1,5 @@
+task:
+    - "Add `container_name` kwarg to `CreateContainer` Docker task - [#2904](https://github.com/PrefectHQ/prefect/pull/2904)"
+
+contributor:
+    - "[Bradley McElroy](https://github.com/limx0)"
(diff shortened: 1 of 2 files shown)


joshmeek:
Thanks for the PR @limx0! Looks like you need to add the `container_name` to the docstring for your tasks. Could you also add this changelog entry to the `changes/` directory under `pr2904.yaml` with the contents of something like:

```yaml
task:
    - "Add `container_name` kwarg to `CreateContainer` Docker task - [#2904](https://github.com/PrefectHQ/prefect/pull/2904)

contributor:
    - [Bradley McElroy](https://github.com/limx0)
```

(I would add it myself but looks like it won't let me make changes to this branch)

limx0:
Thanks @joshmeek, those few things are fixed now

PR #2905, opened 2020-07-02
WIP: in-repo conda recipe with subpackages

mcg1969:

## What does this PR change?

This adds a conda recipe to the repository, which might be nice in its own right, but the real benefit is to demonstrate the feasibility of using conda subpackages in the recipe to mimic pip options.

## Why is this PR important?

It provides a parallel approach to installing Prefect's optional dependencies to the pip "option" approach implemented in setup.py and documented here:

https://docs.prefect.io/core/getting_started/installation.html#optional-dependencies

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,0 +1,199 @@
+{% set data = load_setup_py_data()  %}
+{% set version = data.get('version').lstrip('v') %}
+
+package:
+  name: "prefect"
+  version: {{ version }}
+
+source:
+  path: ../
+
+build:
+  noarch: python
+  number: 0
+  script: python -m pip install --no-deps -vv .
+  entry_points:
+    - prefect = prefect.cli:cli
+
+requirements:
+  host:
+    - python {{ data.get('python_requires') }}
+    - pip
+  run:
+    - python {{ data.get('python_requires') }}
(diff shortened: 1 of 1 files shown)


mcg1969:
It sure would be nice if we could more easily unify the information inside setup.py with this conda recipe—that is, pull the dependencies for each optional requirement from that file. To be honest, we might:
https://github.com/holoviz/hvplot/blob/master/conda.recipe/meta.yaml

jcrist:
Thanks @mcg1969! If we did this, I *think* we'd want to move this to the conda-forge feedstock. Having two versions of the conda recipe doesn't really make sense IMO. Other than that, this seems fairly straightforward to me.

mcg1969:
I certainly respect your choice here. I personally prefer to include in-repo recipes to better facilitate testing. I leverage conda-build's testing cycle to ensure the sufficiency of test environments. Regardless, feel free to lift this pattern into conda-forge and close the PR altogether.

madkinsz:
There's now an open PR for this in the feedstock repo although it looks like it will need some additional polish before it can be merged. https://github.com/conda-forge/prefect-feedstock/pull/72

jcrist:
Closing in favor of work in the feedstock repo.

PR #2906, opened 2020-07-02
Fix edge case in Interval Clocks

cicdw:

## What does this PR change?
This PR fixes an edge case involving interval clocks with short intervals and DST-observing start dates far in the past.

## Why is this PR important?
Scheduling should be robust!

**cc:** @zdhughes / @jlowin this fixes the issue we were poking at earlier today


zdhughes:
Super interesting edge case-- thanks for the tag!

PR #2911, opened 2020-07-02
bump cloudpickle version ceiling to 2.0

### LINKED ISSUE #2910: remove version ceilings on prefect's dependencies?
## Current behavior

Today, `prefect` requires that you have a `cloudpickle` version older than v 1.5.0:

https://github.com/PrefectHQ/prefect/blob/214b2944c0b49eaac0117e06e7fc1d0dabe408fc/requirements.txt#L2

I noticed this through a CI failure today, because [`cloudpickle` did a 1.5.0 release yesterday](https://pypi.org/project/cloudpickle/1.5.0/)

image

Why is this version ceiling in place? Is there a particular incompatibility that the maintainers here knew to anticipate in 1.5.0?

It looks like this was done automatically by a bot: https://github.com/PrefectHQ/prefect/commit/0ff0ce9f0700142dc2f4d8f875768d85d654fc48

## Proposed behavior

I think it would be a better user experience to only use a ceiling to catch new _major_ versions of *cloudpickle` (`cloudpickle<2.0`). I can't tell from [the repo](https://github.com/cloudpipe/cloudpickle) if `cloudpickle` follows semantic versioning, but if they do then there should only be breaking changes in major versions.

jameslamb:

## What does this PR change?

This PR bumps the version ceiling for `cloudpickle` to 1.6, since version 1.5.0 was released yesterday. See #2910 for more background and discussion.

## Why is this PR important?

This PR makes it possible to install the version of `prefect` that is currently on `master` if you have the latest released version of `cloudpickle` installed, so that users don't hit the situation where their versions of `prefect` and `cloudpickle` are incompatible.

In my opinion this version ceiling should be removed completely, but I think for now at least bumping it to 1.6 would fix the immediate issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 click >= 7.0, < 8.0
-cloudpickle >=0.6.0, <1.6
+cloudpickle >=0.6.0, <2.0
 croniter >= 0.3.24, <1.0
 dask[bag] >=0.19.3, <3.0
 distributed >= 1.26.1, < 3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 click >= 7.0, < 8.0
-cloudpickle >=0.6.0, <1.5
+cloudpickle >=0.6.0, <1.6
 croniter >= 0.3.24, <1.0
 dask[bag] >=0.19.3, <3.0
 distributed >= 1.26.1, < 3.0


requirements.txt
```diff
@@ -1,5 +1,5 @@
 click >= 7.0, < 8.0
-cloudpickle >=0.6.0, <1.5
+cloudpickle >=0.6.0, <1.6
```
jcrist, line 2: IMO we should just remove the upper limit on `cloudpickle`. Breaking changes are unlikely, and if needed users are the ones that should be strongly pinning dependencies, not libraries. cc @cicdw for a 👍 / 👎.

cicdw: 👍

jameslamb: @jcrist I think you should change it to a ceiling on major version, `<2.0`

jameslamb: but otherwise I definitely also favor removing the ceiling. Just thought that my be overstepping for me to do as a non-maintainer

jcrist: Sure, that'd match the existing pattern. I still think we should reconsider setting ceilings in the library at all, but that conversation can happen elsewhere.

jameslamb: What would you like me to do with this PR? Maybe change it to `< 2.0` for this narrow case, and then you folks can use #2910 to discuss removing version ceilings altogether?

jcrist: Sure, that makes sense to me.

jameslamb: Ok cool. I just updated this and changed the title on #2910 , thanks

jcrist:
LGTM, will merge on tests pass.

PR #2914, opened 2020-07-03
Added poke handler to notify agent process of available flow runs

asandeep:

## What does this PR change?
This is following the conversion in [2908](https://github.com/PrefectHQ/prefect/discussions/2908), to reduce the latency in executing the flow runs by agent process due to loop intervals.

The PR implements a poke handler to wake up agent process and start processing the flow runs.

## Why is this PR important?
The current loop interval adds a sort of limitation when prefect needs to be invoked in realtime scenarios. The poke handler implemented in this PR provides a way for users to trigger the flow execution without waiting for loop interval to complete.

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -34,12 +34,14 @@ ascii_name = r"""
 # Event to notify agent process to start looking for available flow runs.
 AGENT_WAKE_EVENT = threading.Event()
 
+
 @contextmanager
 def exit_handler(agent: "Agent") -> Generator:
     exit_event = threading.Event()
 
     def _exit_handler(*args: Any, **kwargs: Any) -> None:
         agent.logger.info("Keyboard Interrupt received: Agent is shutting down.")
+        AGENT_WAKE_EVENT.set()
         exit_event.set()
 
     original = signal.getsignal(signal.SIGINT)
@@ -59,16 +61,18 @@ class HealthHandler(web.RequestHandler):
         # Empty json blob, may add more info later
         self.write({})
 
-class PokeHandler(web.RequestHandler):
-    """Respond to /poke
 
-    The handler is expected to be called by python client when new flow run is
-    created, which needs to be immediately executed.
+class PokeHandler(web.RequestHandler):
+    """Respond to /api/poke
+
+    The handler is expected to be called by user to notify agent of available
+    flow runs waiting for execution.
     """
 
     def get(self) -> None:
         # Wake up agent that might be waiting for interval loop to complete.
         AGENT_WAKE_EVENT.set()
+
 
 class Agent:
     """
@@ -181,10 +185,13 @@ class Agent:
 
         return agent_id
 
-    def start(self) -> None:
+    def start(self, _loop_intervals=None) -> None:
         """
         The main entrypoint to the agent. This function loops and constantly polls for
         new flow runs to deploy
+
+        Args:
+            - _loop_intervals (dict, optional): Default loop_intervals override for unit tests.
         """
         if config.backend == "cloud":
             self._verify_token(self.client.get_auth_token())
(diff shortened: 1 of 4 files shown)


src/prefect/client/client.py
```diff
@@ -875,6 +877,16 @@ def create_flow_run(
...
+
+        import requests
+
+        url = urljoin(self.agent_server, "/poke")
+        self.logger.debug("Poking agent process for new flow runs: %s", url)
+        try:
+            requests.get(url, timeout=0.25)
+        except requests.exceptions.RequestException as rex:
+            self.logger.info("Error poking agent process: %s", str(rex))
+
```
joshmeek, line 889: Hmm I'm not sure about having this in the create_flow_run function. I think it would be better split out into a separate function or perhaps actually moved into some different utility. The Client is generally reserved for interacting with the backend API and the agent API is completely separate

src/prefect/agent/agent.py
```diff
@@ -234,7 +248,10 @@ def setup(self) -> None:
                 raise ValueError("Must specify port in agent address")
             port = cast(int, parsed.port)
             hostname = parsed.hostname or ""
-            app = web.Application([("/api/health", HealthHandler)])
+            app = web.Application([
+                ("/api/health", HealthHandler),
+                ("/poke", PokeHandler)
```
jcrist, line 253:
```suggestion
                ("/api/poke", PokeHandler)
```

src/prefect/client/client.py
jcrist: Yeah, we don't even guarantee that the agent api is running. We might not even have a method in prefect core for doing this, and leave poking the agent up to the users themselves. The agent itself may be behind a firewall, running on external infrastructure, behind a proxy with a path prefix, behind an authenticating proxy,  etc... There's too many complications here with how people may deploy prefect in production to make this worth exposing directly to users IMO.

asandeep: Thanks for your inputs @joshmeek @jcrist. I agree that it doesn't really makes sense to have client poke the agent, and should be left to the user on how they want to do it. 

I will make the necessary adjustments and resend the PR.

src/prefect/agent/agent.py
asandeep: Done.

src/prefect/agent/agent.py
```diff
@@ -39,6 +41,7 @@ def exit_handler(agent: "Agent") -> Generator:
 
     def _exit_handler(*args: Any, **kwargs: Any) -> None:
         agent.logger.info("Keyboard Interrupt received: Agent is shutting down.")
+        AGENT_WAKE_EVENT.set()
         exit_event.set()
```
jcrist, line 45: You probably want to set the `exit_event` first, otherwise there might be a race condition where the agent wakes up, sees it shouldn't exit before the `exit_event` is set. With some logic changes we could get by with only one event, but keeping two is fine for now.

src/prefect/agent/agent.py
```diff
@@ -170,10 +185,13 @@ def _register_agent(self) -> str:
...
         return agent_id
 
-    def start(self) -> None:
+    def start(self, _loop_intervals: dict = None) -> None:
         """
         The main entrypoint to the agent. This function loops and constantly polls for
         new flow runs to deploy
+
+        Args:
+            - _loop_intervals (dict, optional): Default loop_intervals override for unit tests.
```
jcrist, line 194: I might just make the docstring description `Exposed for testing only.`, since we really want to make it clear to users that this isn't a public parameter.

src/prefect/agent/agent.py
asandeep: You are right. Updated as suggested.

src/prefect/agent/agent.py
asandeep: I did mentioned unit tests in earlier doc string, but I think this is more concise. Updated.

asandeep:
@jcrist I have addressed your comments. PTAL.

jcrist:
Thanks!

PR #2916, opened 2020-07-05
Implement RunNamespacedJob task for Kubernetes 

pcieslinski:

## What does this PR change?
In this PR, I created the `RunNamespacedJob` task according to what was given in this issue: https://github.com/PrefectHQ/prefect/issues/2281

Generally, the flow looks like this: first, a task is created on the Kubernetes cluster as specified in the `body` argument. Then, in five-second intervals, the Kubernetes API is asked about the status of a given job, if the job has been successfully completed, then all of its rescuers are removed from the cluster: both job and corresponding pods. If the job on Kubernetes failed, the resources are not cleaned so that the user can find out the cause of the failure.

Design decisions:
* `RunNamespacedJob` maintains the `CreateNamespacedJob` interface so that the user can simply create the job and not worry about whether it will be removed and how. Of course, this can be changed in the future.
* If a job fails on Kubernetes, its resources are not removed from the cluster so that the user can view the logs and verify the cause of the failure. If we want to change this behavior, we can add an additional `cleanup_on_failure` argument in the interface, based on which we will decide whether resources should be removed even in the event of failure.
* A logger has been used throughout the task so that it is transparent to the user at what stage the job is being performed.
* In the event that job fails on Kubernetes, I use `singnals.FAIL` to give the entire task appropriate status.

## Why is this PR important?
Currently, in the task library, we don't have a task that would be able to create a job, check its status, and then delete all resources after its completion. This will be a great help in using Jobs on Kubernetes for our users.

This PR is also the answer to the problem described in: https://github.com/PrefectHQ/prefect/issues/2281

### DIFF
--- a/src/prefect/tasks/kubernetes/job.py
+++ b/src/prefect/tasks/kubernetes/job.py
@@ -723,6 +723,6 @@ class RunNamespacedJob(Task):
             api_client.delete_namespaced_job(
                 name=job_name,
                 namespace=namespace,
-                body=client.V1DeleteOptions(propagation_policy="Background"),
+                body=client.V1DeleteOptions(propagation_policy="Foreground"),
             )
             self.logger.info(f"Job {job_name} has been deleted.")
--- a/src/prefect/tasks/kubernetes/job.py
+++ b/src/prefect/tasks/kubernetes/job.py
@@ -1,8 +1,10 @@
+import time
 from typing import Any, cast
 
 from kubernetes import client
 
 from prefect import Task
+from prefect.engine import signals
 from prefect.utilities.tasks import defaults_from_attrs
 from prefect.utilities.kubernetes import get_kubernetes_client
 
@@ -49,7 +51,7 @@ class CreateNamespacedJob(Task):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        **kwargs: Any
+        **kwargs: Any,
     ):
         self.body = body or {}
         self.namespace = namespace
@@ -139,7 +141,7 @@ class DeleteNamespacedJob(Task):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        **kwargs: Any
+        **kwargs: Any,
     ):
         self.job_name = job_name
         self.namespace = namespace
@@ -192,7 +194,7 @@ class DeleteNamespacedJob(Task):
             name=job_name,
             namespace=namespace,
             body=client.V1DeleteOptions(**delete_option_kwargs),
-            **kube_kwargs
+            **kube_kwargs,
         )
 
 
@@ -235,7 +237,7 @@ class ListNamespacedJob(Task):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        **kwargs: Any
+        **kwargs: Any,
     ):
         self.namespace = namespace
         self.kube_kwargs = kube_kwargs or {}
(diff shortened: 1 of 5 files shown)


pcieslinski:
I thought I would add a quick way to test `RunNamespacedJob` here. :) Personally, I tested its operation with unit tests and a very simple flow using the Kubernetes cluster on a `minikube`.

```python
from prefect import Flow
from prefect.tasks.kubernetes.job import RunNamespacedJob


body = {
    "apiVersion": "batch/v1",
    "kind": "Job",
    "metadata": {
        "name": "pi"
    },
    "spec": {
        "template": {
            "spec": {
                "containers": [
                    {
                        "name": "pi",
                        "image": "perl",
                        "command": [
                            "perl",
                            "-Mbignum=bpi",
                            "-wle",
                            "print bpi(2000)"
                        ]
                    }
                ],
                "restartPolicy": "Never"
            }
        },
        "backoffLimit": 4
    }
}


with Flow("Kubernetes Job") as flow:
    task = RunNamespacedJob(body=body, kubernetes_api_key_secret=None)()

result = flow.run()
```

This Flow uses an example job that has been described in the official [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/job/).

src/prefect/tasks/kubernetes/job.py
```diff
@@ -571,3 +573,134 @@ def run(
...
+            else:
+                if res.status.failed:
+                    raise signals.FAIL(
+                        f"Job {job_name} failed, check Kubernetes pod logs for more information."
+                    )
+                elif res.status.succeeded:
+                    self.logger.info(f"Job {job_name} has been completed.")
+                    completed = True
+
+        api_client.delete_namespaced_job(
```
joshmeek, line 701: I also think the delete should be toggleable on the task, perhaps with a kwarg that's something like `delete_job_after_completion=True`

src/prefect/tasks/kubernetes/job.py
```diff
@@ -571,3 +573,134 @@ def run(
         api_client.replace_namespaced_job(
             name=job_name, namespace=namespace, body=body, **kube_kwargs
         )
+
+
+class RunNamespacedJob(Task):
```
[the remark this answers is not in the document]
pcieslinski: Of course, I totally forgot about it. ;)

src/prefect/tasks/kubernetes/job.py
```diff
@@ -571,3 +573,134 @@ def run(
...
+        - kube_kwargs (dict, optional): Optional extra keyword arguments to pass to the
+            Kubernetes API (e.g. `{"pretty": "...", "dry_run": "..."}`)
+        - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
+            which stored your Kubernetes API Key; this Secret must be a string and in
+            BearerToken format
+        - **kwargs (dict, optional): additional keyword arguments to pass to the Task
+            constructor
+    """
+
+    JOB_STATUS_POOL_INTERVAL = 5
```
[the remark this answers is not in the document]
pcieslinski: Fair point, I thought about it at first. In general, some of the jobs that will be created using this task can be quite long and it makes no sense to ask the API every 5 seconds for status.

src/prefect/tasks/kubernetes/job.py
pcieslinski: Full agreement on this. :)

src/prefect/tasks/kubernetes/job.py
```diff
@@ -571,3 +573,156 @@ def run(
...
+                    )
+                elif res.status.succeeded:
+                    self.logger.info(f"Job {job_name} has been completed.")
+                    completed = True
+
+        if delete_job_after_completion:
+            api_client.delete_namespaced_job(
+                name=job_name,
+                namespace=namespace,
+                body=client.V1DeleteOptions(propagation_policy="Background"),
```
joshmeek, line 726:
```suggestion
                body=client.V1DeleteOptions(propagation_policy="Foreground"),
```

joshmeek: This way we can ensure the job is deleted prior to returning

Issue #2771, opened 2020-06-12
Support cancellation of active flow runs


jcrist:
Currently flow runs can be transitioned into a `Cancelled` state, but that only prevents new tasks in the flow run from starting - it does nothing to stop already running tasks (it's also not exposed nicely in the UI).

In practice this is hard to do effectively in all situations, there is no straightforward way to interrupt an arbitrary python function in all contexts. If running in process, we can interrupt the main thread, but not other threads. If running in a subprocess we can kill the process. If running in a dask cluster we can cancel pending work, but not stop running work - but with #2667 the `DaskExecutor` can now manage the lifetime of a dask cluster, so we could have the executor shutdown the cluster as a hard stop.

Given the above restrictions, it seems we could add more support for cancelling flow runs with the following semantics:

- Cancellation of a flow run starts when flow runs are transitioned into a `Cancelling` state, where `Cancelling` is a subclass of `Running`. This indicates that the flow run is in the process of cancelling, but still is technically `Running`. Transitioning directly into `Cancelled` from `Running` skips a necessary intermediate state and should be deprecated.
- Cancellation is still "best effort":
  - If a task is started, we'll cancel it if possible, otherwise it may run until termination
  - If a task hasn't started, it won't be started

Focussing on how this works with orchestration, the following takes place after a user cancels a running flow:

- The flow run is moved to `Cancelling`
- Once in `Cancelling` , `TaskRunner` interactions with cloud/server (either heartbeats or state updates) will let the task runner know that the task should be cancelled. If possible, a running task will be interrupted, otherwise it will be allowed to finish. Tasks that are allowed to finish will transition to their final state as if cancellation hadn't occurred (e.g. `Success` or `Failed`). This provides a more accurate representation of the cancelled flow run. Tasks that are successfully cancelled are marked as  `Cancelled`.
- Likewise, `FlowRunner` interactions with cloud (either heartbeats or state updates) will let the flow runner know that the flow should be cancelled. Any tasks that haven't been submitted yet will transition directly to `Cancelled`.
[...]

PR #2920, opened 2020-07-06
Executor leaves no background work upon exiting `start` contextmanager

jcrist:
Previously when the `executor.start()` contextmanager would exit early (say on an interrupt or error), there might be background work still happening, since the executor wouldn't wait for or cancel the remaining submitted tasks.

We now attempt to cancel all pending work running on the executor, and wait for tasks that haven't completed in cases where we can't efficiently/robustly cancel them. This ensures that when `executor.start()` exits, no tasks will continue to progress.

- For a `DaskExecutor` with a temporary cluster, we shutdown the cluster, so no lingering work can persist.
- For a `DaskExecutor` with an external cluster (specified with an `address`) or an `inproc` cluster (using local threads only), we stop all pending tasks and wait for all running tasks to finish.
- For a `LocalDaskExecutor`, we terminate the backing pool as quickly as possible, then wait for the pool to close.
  - For a process pool, this terminates the processes
  - For a thread pool, we attempt to interrupt the threads (CPython specific).
- A `LocalExecutor` runs everything in the main thread, so no background tasks can persist.

This is a precursor for the cancellation implementation (#2771).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 click >= 7.0, < 8.0
 cloudpickle >=0.6.0, <2.0
 croniter >= 0.3.24, <1.0
-dask[bag] >=0.19.3, <3.0
-distributed >= 1.26.1, < 3.0
+dask[bag] >= 2.8.0, <3.0
+distributed >= 2.8.0, < 3.0
 docker >=3.4.1, <5.0
 marshmallow >= 3.0.0b19, < 3.6.2
 marshmallow-oneofschema >= 2.0.0b2, < 3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 click >= 7.0, < 8.0
 cloudpickle >=0.6.0, <2.0
 croniter >= 0.3.24, <1.0
-dask[bag] >= 2.8.0, <3.0
+dask >= 2.8.0, <3.0
 distributed >= 2.8.0, < 3.0
 docker >=3.4.1, <5.0
 marshmallow >= 3.0.0b19, < 3.6.2
--- a/src/prefect/engine/executors/dask.py
+++ b/src/prefect/engine/executors/dask.py
@@ -1,6 +1,7 @@
 import logging
 import uuid
 import warnings
+import weakref
 from contextlib import contextmanager
 from typing import Any, Callable, Iterator, TYPE_CHECKING, Union, Optional
 
@@ -10,7 +11,8 @@ from prefect.utilities.importtools import import_object
 
 if TYPE_CHECKING:
     import dask
-    from distributed import Future
+    from distributed import Future, Variable
+    import multiprocessing.pool
 
 
 __any__ = ("DaskExecutor", "LocalDaskExecutor")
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
 click >= 7.0, < 8.0
 cloudpickle >=0.6.0, <2.0
 croniter >= 0.3.24, <1.0
-dask[bag] >=0.19.3, <3.0
-distributed >= 1.26.1, < 3.0
+dask[bag] >= 2.8.0, <3.0
```
cicdw, line 4: I actually don't think we use `dask[bag]` anymore so we should be able to remove `[bag]` here

src/prefect/engine/executors/dask.py
```diff
@@ -200,24 +222,71 @@ def start(self) -> Iterator[None]:
...
+            try:
+                futures = [f for f in list(self._futures) if not f.done()]  # type: ignore
+                if futures:
+                    self.logger.info(
+                        "Stopping executor, waiting for %d active tasks to complete",
+                        len(futures),
+                    )
+                    wait(futures)
+            except Exception:
+                pass
```
cicdw, line 281: any benefit to logging the exceptions at warning level?

src/prefect/engine/executors/dask.py
```diff
@@ -271,7 +340,14 @@ def submit(
...
 
         kwargs.update(self._prep_dask_kwargs(extra_context))
-        return self.client.submit(fn, *args, **kwargs)
+        if self._should_run_var is None:
+            fut = self.client.submit(fn, *args, **kwargs)
+        else:
+            fut = self.client.submit(
+                _maybe_run, self._should_run_var, fn, *args, **kwargs
+            )
+            self._futures.add(fut)
```
cicdw, line 349: When I first saw this I was worried about a situation where this was `None` but after a re-read it appears that `_should_run_var` and `_futures` are coupled attributes - any way to simplify that?  

I'm worried about the case where lines 255-256 are refactored / initialized somewhere else without realizing the dependency.

src/prefect/engine/executors/dask.py
jcrist: It indicates that things may be slower to shutdown than the user might want. Happy to change to a different level if you want.

src/prefect/engine/executors/dask.py
jcrist: I don't see a cleaner way to refactor this - they're coupled attributes, but that's what tests are for? Things will only pass with the current implementation working the way it does. I can add a comment somewhere I suppose.

src/prefect/engine/executors/dask.py
cicdw: Makes sense 👍

jcrist:
I believe all comments have been addressed.

src/prefect/engine/executors/dask.py
cicdw: Coupled hidden attributes make me nervous but I'm definitely not interested in holding the PR up for it - the comment helps 👍

PR #2923, opened 2020-07-07
Add Cancelling state

jcrist:
Adds a new state subclassed from `Running` that indicates that a flow run has been requested to be cancelled, but may still have tasks running. Currently we do nothing with this state, this PR is adding it now so we can get cloud integration done first.

Part of the work on #2771.


jcrist:
Done. Added a script to the repo as well.

PR #2929, opened 2020-07-08
[WIP] Capture warnings

cicdw:

## What does this PR change?
This PR adds `py.warnings` as a default extra logger; if provided, `logging.captureWarnings(True)` is also called.  This will auto-convert warnings into WARNING level logs that Prefect ingests.

Note:
- this behavior can be turned off by removing `py.warnings` from `extra_loggers` config
- this applies a _global_ setting at Prefect import time; I'm guessing this is undesirable as a secret side effect so I wanted to open this PR for discussion on other ways of handling this

## Why is this PR important?
This has been informally requested before, and I think will expose useful info (e.g., Dask worker warnings) without being too noisy.

### DIFF
--- a/src/prefect/config.toml
+++ b/src/prefect/config.toml
@@ -99,7 +99,7 @@ datefmt = "%Y-%m-%d %H:%M:%S"
 log_to_cloud = false
 
 # Extra loggers for Prefect log configuration
-extra_loggers = "[]"
+extra_loggers = "['py.warnings']"
 
 [flows]
 # If true, edges are checked for cycles as soon as they are added to the flow. If false,
--- a/changes/pr2929.yaml
+++ b/changes/pr2929.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Add py.warnings as a default extra logger to capture Warnings as logs - [#2929](https://github.com/PrefectHQ/prefect/pull/2929)"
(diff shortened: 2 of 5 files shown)


jcrist:
I wouldn't feel comfortable turning this on by default. Lots of libraries use the `warnings` mechanism, without knowing what the warnings are this feels like a way secret things could be potentially logged to cloud. It's also a global setting as you said, so `import prefect` would silence warnings from other libraries.

src/prefect/config.toml
```diff
@@ -99,7 +99,7 @@ datefmt = "%Y-%m-%d %H:%M:%S"
 log_to_cloud = false
 
 # Extra loggers for Prefect log configuration
-extra_loggers = "[]"
+extra_loggers = "['py.warnings']"
```
jcrist, line 102: Perhaps a boolean flag would make more sense than adding it to `extra_loggers`? It also makes the option more discoverable if not enabled by default, since the option would still appear in the `config.toml`.

cicdw: Yea that makes sense; I originally wanted to avoid adding another config setting for something somewhat minor like this.

bnaul:
@jcrist one additional bit of context: I had brought something like this up in the same context of the Slack [thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1594156794331300) I posted yesterday re: large objects in the task graph; that log was kind of tricky to track down since it didn't appear in Cloud. My original thinking was that surfacing more verbose logs from the FlowRunner would be enough, but I guess since this one is actually from dask log it's tricky to grab it but not also pull in everything from every package....? No suggestions unfortunately, just an example of something that could be easier w/ the right solution.

cicdw:
Closing for now, I don't think we actually intend on merging this

PR #2934, opened 2020-07-08
Improve output format in prefect describe cli

### LINKED ISSUE #2926: prefect describe CLI does not return JSON
## Description
`prefect describe` calls seem to give back no valid JSON but just a printed Python Dict. To make it more useful in downstream commands to e.g. parse the output via jq it would be great if this is JSON. Maybe that was even intended, not sure if it is a bug or enhancement.

One example:
```
prefect describe flows --name TestRecoGenerationFlow_dev  | jq
parse error: Invalid numeric literal at line 1, column 8
```

Output of this looks like the following:
```
{'name': 'TestRecoGenerationFlow_dev', 'version': 17, 'created': '2020-07-07T12:43:44.554817+00:00', 'description': None, ...
```

With the `None` values instead of `null` it is clear that this is no JSON

## Expected Behavior
Proper returning of JSON so that it can be used in downstream tools or pipes.

## Reproduction
```
prefect describe flows --name <anyname>  | jq
```

## Environment
```
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "macOS-10.15.4-x86_64-i386-64bit",
    "prefect_version": "0.12.2",
    "python_version": "3.8.2"
  }
}
```

jcrist:
Previously the default format for `prefect describe` was to print the
python object. This is similar to json, but not actually json. There's
not a good reason to not use json, so we default to json formatting, and
also add yaml as an option. We also pretty-print the json to make it
easier to read.

Fixes #2926.

### DIFF
--- a/changes/pr2934.yaml
+++ b/changes/pr2934.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Improve output formatting of `prefect describe` CLI - [#2934](https://github.com/PrefectHQ/prefect/pull/2934)"
--- a/src/prefect/cli/describe.py
+++ b/src/prefect/cli/describe.py
@@ -6,10 +6,28 @@ from prefect.client import Client
 from prefect.utilities.graphql import EnumValue, with_args
 
 
+output_option = click.option(
+    "--output", "-o", type=click.Choice(["json", "yaml"]), hidden=True, default="json"
+)
+
+
+def display_output(obj, output="json"):
+    """Display `obj` in a selected format {'json', 'yaml'}"""
+    if output == "json":
+        click.echo(json.dumps(obj, sort_keys=True, indent=2))
+    elif output == "yaml":
+        import yaml
+
+        click.echo(yaml.safe_dump(obj))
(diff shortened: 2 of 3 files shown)


jcrist:
Prettier default output:

```
$ prefect describe flows --name "A flow"
{
  "archived": false,
  "created": "2019-11-13T16:46:40.558258+00:00",
  "description": null,
  "environment": {
    "__version__": "0.7.1+57.g1491de3b",
    "executor": "prefect.engine.executors.LocalExecutor",
    "executor_kwargs": {},
    "labels": [
      "local",
      "a-flow"
    ],
    "type": "RemoteEnvironment"
  },
  "name": "A flow",
  "parameters": [],
  "storage": {
    "__version__": "0.7.1+57.g1491de3b",
    "directory": "/Users/chris/.prefect/flows",
    "flows": {
      "A flow": "/Users/chris/.prefect/flows/a-flow.prefect"
    },
    "type": "Local"
  },
  "version": 4
}
```

PR #2935, opened 2020-07-08
Mirror flow run state within task

cicdw:

## What does this PR change?
This PR adds a new `wait` kwarg to `FlowRunTask` which will wait until the triggered flow run completes, and reflect the corresponding state back in this task.  Note that an idempotency key is used so that retries, etc. work robustly without triggering new flow runs.

## Why is this PR important?
This task can now be used to create flow-to-flow dependencies

### DIFF
--- a/src/prefect/engine/signals.py
+++ b/src/prefect/engine/signals.py
@@ -4,8 +4,36 @@ running. Signals are used in TaskRunners and FlowRunners as a way of communicati
 in states.
 """
 
+from typing import Type
+
 from prefect.engine import state
 from prefect.utilities.exceptions import PrefectError
+
+
+def signal_from_state(state: state.State) -> Type["PrefectStateSignal"]:
+    """
+    Given a state instance, returns the corresponding Signal type.
+
+    Args:
+        - state (State): an instance of a Prefect State
+
+    Returns:
+        - PrefectStateSignal: the Prefect Signal corresponding to
+            the provided state
+
+    Raises:
+        - KeyError: if no signal matches the provided state
+    """
+    unprocessed = set(PrefectStateSignal.__subclasses__())
+    signals = dict()
+    while unprocessed:
+        sig = unprocessed.pop()
+        signals[sig._state_cls.__name__] = sig
+        unprocessed = unprocessed.union(sig.__subclasses__())
+    try:
+        return signals[type(state).__name__]
+    except KeyError:
+        raise KeyError(f"No signal matches the provided state: {state}") from None
 
 
 class ENDRUN(Exception):
--- a/src/prefect/engine/signals.py
+++ b/src/prefect/engine/signals.py
@@ -22,7 +22,7 @@ def signal_from_state(state: state.State) -> Type["PrefectStateSignal"]:
             the provided state
 
     Raises:
-        - KeyError: if no signal matches the provided state
+        - ValueError: if no signal matches the provided state
     """
     unprocessed = set(PrefectStateSignal.__subclasses__())
     signals = dict()
@@ -33,7 +33,7 @@ def signal_from_state(state: state.State) -> Type["PrefectStateSignal"]:
     try:
         return signals[type(state).__name__]
     except KeyError:
-        raise KeyError(f"No signal matches the provided state: {state}") from None
+        raise ValueError(f"No signal matches the provided state: {state}") from None
 
 
 class ENDRUN(Exception):
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -2,7 +2,7 @@ import time
 from typing import Any
 from urllib.parse import urlparse
 
-from prefect import config, Task
+from prefect import config, context, Task
 from prefect.client import Client
 from prefect.engine.signals import PrefectStateSignal
 from prefect.utilities.graphql import EnumValue, with_args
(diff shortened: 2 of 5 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -106,4 +113,27 @@ def run(
...
+            parameters=parameters,
+            idempotency_key=context.get("flow_run_id"),
+        )
+
+        if not self.wait:
+            return flow_run_id
+
+        signals = {
+            sig._state_cls.__name__: sig for sig in PrefectStateSignal.__subclasses__()
+        }
```
jcrist, line 130: This bit feels a bit off to me - perhaps make this a utility function in the `signals` module? Something like:

```python
def from_state(state : State) -> PrefectStateSignal:
    ...
```

cicdw: Sure I can do that - to what end though?

jcrist: Just a cleanliness thing. Feels odd to dig into internal attributes like `__subclasses__` and `_state_cls` from inside the task library.

cicdw: @jcrist - updated; also added a test for the newly created function and caught a minor bug so thanks for the suggestion!

src/prefect/engine/signals.py
```diff
@@ -4,10 +4,38 @@
...
+    unprocessed = set(PrefectStateSignal.__subclasses__())
+    signals = dict()
+    while unprocessed:
+        sig = unprocessed.pop()
+        signals[sig._state_cls.__name__] = sig
+        unprocessed = unprocessed.union(sig.__subclasses__())
+    try:
+        return signals[type(state).__name__]
+    except KeyError:
+        raise KeyError(f"No signal matches the provided state: {state}") from None
```
jcrist, line 36: Minor nit, feel free to ignore - this feels like it should be a `ValueError` instead, since the `state` value isn't a valid value. `KeyError` is generally for user-facing code that is doing an actual indexing operation.

cicdw: 👍  works for me, will update

src/prefect/tasks/prefect/flow_run.py
```diff
@@ -106,4 +113,23 @@ def run(
...
         flow_id = flow[0].id
-        return client.create_flow_run(flow_id=flow_id, parameters=parameters)
+
+        # providing an idempotency key ensures that retries for this task
+        # will not create additional flow runs
+        flow_run_id = client.create_flow_run(
+            flow_id=flow_id,
+            parameters=parameters,
+            idempotency_key=context.get("flow_run_id"),
+        )
```
joshmeek, line 123: Not sure if I'm overthinking, and feel free to ignore, but what if I do want my flow to kick off the run again on retries?

jcrist: Perhaps the `task_run_count` should be incorporated as part of the `idempotency_key`, and failures in the network request handled internal to the task instead of external with the retry mechanism. So retries for this task would be retrying the whole flow, not retrying the request to start the flow.

```suggestion
        # Providing an idempotency key ensures that retries for this network request
        # will not create additional flow runs
        flow_run_id = context["flow_run_id"]
        task_run_count = context.get("task_run_count", 1)
        idempotency_key = f"{flow_run_id}-{task_run_count}"
        flow_run_id = client.create_flow_run(
            flow_id=flow_id,
            parameters=parameters,
            idempotency_key=idempotency_key,
        )
```

cicdw: I have pushback on this - we shouldn't confuse this task raising an error with the Flow Run it created requesting a Flow-level retry, which isn't a well defined concept (the suggested implementation would create a new flow run for every "retry" event which isn't really a retry IMO).  _Any_ runtime error (including the task becoming a zombie) would result in an entirely new flow run.  I just can't imagine a scenario where that's desirable behavior.

jcrist: Sorry, I think we might be talking past each other. There are two types of task failures here:

- The flow run fails to submit. This is an error in this task.
- The flow run completes with a `Failed` state, a `FAILED` signal gets raised, and this task gets marked as failed.

If I'm using this task and set a retry on it, what is the retry for? Currently the retry only works for failures in submission - if the flow run fails the task will retry, but won't resubmit the flow run and I'll loop again in the retry loop.

jcrist: Perhaps this shouldn't set the success of this task as the success of the flow run, and instead should return the state of the flow run? Then a failure is always "failed to submit" rather than the flow run failing.

cicdw: > If I'm using this task and set a retry on it, what is the retry for
The retry is for robustness in the face of zombies - if the process dies + has retries, Cloud will resurrect it so the larger Flow Run can continue and this task run still still reflect the state of the submitted flow run.

cicdw: Also FYI The FAILED signal is a short circuit - it will bypass retry logic.

jcrist: > Also FYI The FAILED signal is a short circuit - it will bypass retry logic.

Ah, didn't realize this. Ok, that avoids the issue of endless retries. There's still not a good way to retry a failed flow run, but it sounds like that's not in-scope for this task. 👍 from me.

PR #2938, opened 2020-07-08
Add doc note about show flow logs agent option for debugging

joshmeek:
Update: doc note

### DIFF
--- a/src/prefect/cli/execute.py
+++ b/src/prefect/cli/execute.py
@@ -47,15 +47,15 @@ def cloud_flow():
         }
     }
 
-    client = Client()
-    result = client.graphql(query)
-    flow_run = result.data.flow_run
-
-    if not flow_run:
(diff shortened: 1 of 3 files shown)


src/prefect/cli/execute.py
```diff
@@ -74,7 +76,10 @@ def cloud_flow():
     except Exception as exc:
         msg = "Failed to load and execute Flow's environment: {}".format(repr(exc))
         state = prefect.engine.state.Failed(message=msg)
-        version = result.data.flow_run[0].version
-        client.set_flow_run_state(flow_run_id=flow_run_id, version=version, state=state)
+        if flow_run:
+            version = result.data.flow_run[0].version
+            client.set_flow_run_state(
```
cicdw, line 81: If the client initialization fails, then `client` won't be defined and this will raise an error, putting you back where you started

PR #2940, opened 2020-07-09
Fix FargateTaskEnvironment token retrieval for runs using core server

joshmeek:
When using this environment with core's server there is no auth token present so this access needs to be updated

### DIFF
--- a/src/prefect/environments/execution/fargate/fargate_task.py
+++ b/src/prefect/environments/execution/fargate/fargate_task.py
@@ -286,8 +286,8 @@ class FargateTaskEnvironment(Environment, _RunMixin):
                 "environment": [
                     {
                         "name": "PREFECT__CLOUD__AUTH_TOKEN",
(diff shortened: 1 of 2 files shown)


src/prefect/environments/execution/fargate/fargate_task.py
```diff
@@ -286,8 +286,8 @@ def execute(self, flow: "Flow") -> None:  # type: ignore
                 "environment": [
                     {
                         "name": "PREFECT__CLOUD__AUTH_TOKEN",
-                        "value": config.cloud.agent.auth_token
-                        or config.cloud.auth_token,
+                        "value": config.cloud.agent.get("auth_token")
+                        or config.cloud.get("auth_token"),
```
cicdw, line 290: dumb question but any issue with `None`s in this template?

joshmeek: Not that I am aware of but it wouldn't hurt to make it empty strings, will update

PR #2942, opened 2020-07-10
Support cancellation of active flow runs

### LINKED ISSUE #2771: Support cancellation of active flow runs

jcrist:
Adds support for cancellation of active flow runs.

During a flow run (with cloud/server) the flow runner periodically polls
the backend to see if the flow run's state has updated to `Cancelling`.
If it has, the flow runner will:
- Attempt to interrupt/kill all active tasks. If it can't interrupt the
  running tasks (only the case when running on a non-temporary dask
  cluster), the flow runner will wait for the remaining tasks to
  complete, then proceed.
- Once there are no more active tasks, the flow runner will update the
  final state of the flow run to `Cancelled`, and set all non `Finished`
  tasks to `Cancelled` as well. This means that tasks that completed
  after cancellation began will be marked with their actual final state,
  and only interrupted/not run tasks will be marked as `Cancelled`.

Fixes #2771.


jcrist:
Still needs tests, and need to properly cancel flow runs in the `Queued` state, but I've tested that this works locally against cloud and things seem to be correct.

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -130,10 +134,104 @@ def call_runner_target_handlers(self, old_state: State, new_state: State) -> Sta
...
+                }
+                variables = {
+                    "input": {
+                        "states": [
+                            {"state": task_state, "task_run_id": run.id}
+                            for run in res.data.flow_run_by_pk.task_runs
+                        ]
+                    }
+                }
+                self.client.graphql(mutation, variables=variables)  # type: ignore
```
cicdw, line 167: My initial reaction to this code block is that setting the individual task runs to `Cancelled` should be the responsibility of the backend not the client - do we know at this moment that no task runs are actively running and might report back a state update?

cicdw:
Also the circle failures are odd; I've seen this before but you might have some insight?
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
/root/project/tests/core/test_flow.py:2424: KeyboardInterrupt
 
(to show a full traceback on KeyboardInterrupt use --full-trace)
 
============ 715 passed, 18 xpassed, 8 warnings in 71.58s (0:01:11) ============
 
/usr/local/lib/python3.6/site-packages/aiohttp/client.py:309: ResourceWarning: Unclosed client session <aiohttp.client.ClientSession object at 0x7fa31e10d8d0>
 
Unclosed client session
 
client_session: <aiohttp.client.ClientSession object at 0x7fa31e10d8d0>
 
 
Exited with code exit status 2 
CircleCI received exit code 2 
```

src/prefect/engine/cloud/flow_runner.py
jcrist: That was my initial reaction as well, but it causes issues with users running past versions of `prefect`.

Currently the "cancellation lite" mechanism lets users set the state of a flow run to `Cancelled`. This prevents new tasks from entering `Running`, but allows existing tasks to continue. Changing the transition to `Cancelled` to also move all non-finished tasks to `Cancelled` on the backend side would break this for users running older versions of prefect, since tasks that are still running would be marked as `Cancelled`.

jcrist:
> OK so I left a question that I might already know the answer to. Can you confirm / correct my understanding here:

Your understanding is correct

> Also the circle failures are odd - maybe related to some tests that need to be removed for KeyboardInterrupt handling?

Yeah, I didn't update any tests yet. The previous `KeyboardInterrupt` handling behavior was troublesome (backends like k8s may send SIGINT to stop errant processes, conflating SIGINT with cancelling a task would cause flow runs to appear cancelled that really should be failed) and was removed, but the tests weren't.

src/prefect/engine/cloud/flow_runner.py
jcrist: New idea: Perhaps on transition to a `Cancelled` state, cloud sets all non-`Running` task runs to `Cancelled`. This does what we want for new versions of core (there will never be `Running` task runs at this point, so all remaining task runs will be cancelled), but doesn't affect older versions of core (running tasks will still be marked as running).

cicdw: > Changing the transition to Cancelled to also move all non-finished tasks to Cancelled on the backend side would break this for users running older versions of prefect, since tasks that are still running would be marked as Cancelled.

So what I envisioned was:
- user hits "cancel" 
- flow is put in "Cancelling" state (if it's running)
- executor attempts to cancel things and eventually sets Flow to `Cancelled`
- Zombie killer service finds task runs / flow runs that haven't heartbeated and does some state-logic to determine whether to set task runs / flow runs to Cancelled

I'd rather not aggressively set task run states to `Cancelled`, only because if they do actually finish successfully it's more informative for their state to reflect that.

cicdw: The zombie killer service enhancement I mentioned here is still compatible with Core setting task runs to `Cancelled` if they actually are cancelled

jcrist: > I'd rather not aggressively set task run states to Cancelled, only because if they do actually finish successfully it's more informative for their state to reflect that.

The logic I proposed above would still accomplish that though, since when a flow run transitions to `Cancelled`, there are no active task runs. I'd rather update this state at the same time, as I believe it provides a nicer user experience. Otherwise users will still see active tasks for a while in the UI before the zombie-killer eventually notices them.

cicdw: AH I understand now; yes I 100% agree - the backend should transition all non-Finished task runs to `Cancelled` whenever the flow run enters a `Cancelled` state instead of having the client do it 👍

jcrist:
This is ready for review.

PR #2944, opened 2020-07-10
Expand file-based storage to all other storage types

### LINKED ISSUE #2887: Expand file-based storage option to other storage types
## Current behavior
`0.12.1` added file-based storage through the use of GitHub storage. This should be expanded to all of the other current storage objects that only store pickles.

## Proposed behavior
Add a kwarg to the storage interface that is something along the lines of `store_as_pickle=False` or `store_as_file=True`. Need to iron out if the storage and registration process would follow a similar pattern to the GitHub storage. 

i.e. would users manually upload the file to S3 or would `.build` be responsible for uploading them (similar to how `build` current uploads pickles)

## Example
This could easily unlock hot reloading flows for local or cloud-based storage types.

joshmeek:

## What does this PR change?
This PR closes #2887 by expanding file-based storage (and hot reloading flows) to all other current storage types. This adds a `stored_as_file: bool` kwarg to the storage types that also support pickle storage.

##### Example Workflow

```python
# flow.py
from prefect.environments.storage import S3

@task
def imi():
    print("I am, I am, I am, I am")

f = Flow("s3-file", tasks=[imi])
f.storage = S3(bucket="my-flows", key="flow.py", stored_as_file=True)
```

Upload to S3:

```bash
aws s3 cp flow.py s3://my-flows/flow.py
```

Register with backend:

```bash
prefect register flow -f flow.py -p "Demo"
```

Opening this PR to get some feedback on the implementation. I plan on adding some more documentation around this pattern and making the storage documentation better. I also need to revamp the healthcheck.py for Docker storage because it is pickle-based right now (current implementation skips over healthcheck if using a file but will re-add it once working 👍)


jcrist:
Woot, this is great to see!

> This adds a `stored_as_file`: bool kwarg to the storage types that also support pickle storage.

Since pickles are also stored as files (it's just the type of the contents of the file that is different), I'm not sure if that's the best kwarg. Perhaps something with `py` or `python` or `script` in the name?

docs/core/idioms/file-based.md
```diff
@@ -75,3 +81,20 @@ The flow is ready to run! Every time you need to change the code inside your flo
...
+flow.storage = Docker(
+    path="my_flow.py",
+    files={"/source/of/my_flow.py": "my_flow.py"},
+    stored_as_file=True
+)
+```
+
+To store flows as files in Docker storage three kwargs needs to be set:
+
+- `path`: pointing to the path that the file is stored in the Docker image
```
jcrist, line 98:
```suggestion
- `path`: the path that the file is stored in the Docker image
```

docs/core/idioms/file-based.md
```diff
@@ -75,3 +81,20 @@ The flow is ready to run! Every time you need to change the code inside your flo
...
+- `files`: a dictionary of local file source to path destination in image
```
jcrist, line 99: Isn't this one only needed if you're using prefect's storage build to store your flow source?

If I use a dockerfile that contains `ADD`/`COPY` lines to add all my flow code, then the `Docker` storage would only need the `path` and the `stored_as_script` kwargs to work correctly?

I could see this being a potentially nice pattern. User writes a dockerfile to add flow source for multiple flows, then uses a storage object that indicates which file in the image to use as the flow source. There's no prefect-side image build step, but the source is still stored in the image.

src/prefect/environments/storage/_healthcheck.py
```diff
@@ -117,8 +122,16 @@ def environment_dependency_check(flows: list):
     flow_file_path, python_version = sys.argv[1:3]
 
     print("Beginning health checks...")
+
+    flow_file_paths = ast.literal_eval(flow_file_path)
+
     system_check(python_version)
-    flows = cloudpickle_deserialization_check(flow_file_path)
+
+    if len(flow_file_paths) == 1 and ".py" in flow_file_paths[0]:
```
jcrist, line 130: Will there always be one flow file for file based storage?

docs/core/idioms/file-based.md
```diff
@@ -75,3 +81,20 @@ The flow is ready to run! Every time you need to change the code inside your flo
...
+- `stored_as_file`: boolean opting into file based storage
```
jcrist, line 100:
```suggestion
- `stored_as_script`: boolean enabling file based storage
```

tests/environments/storage/test_azure_storage.py
```diff
@@ -218,3 +228,34 @@ def test_get_flow_azure_runs(monkeypatch):
 
     state = new_flow.run()
     assert state.is_successful()
+
+
+def test_get_flow_form_file_azure_runs(monkeypatch):
```
jcrist, line 233:
```suggestion
def test_get_flow_from_file_azure_runs(monkeypatch):
```

docs/core/idioms/file-based.md
```diff
@@ -75,3 +81,20 @@ The flow is ready to run! Every time you need to change the code inside your flo
...
 :::
+
+
+### File based Docker storage
+
+```python
+flow.storage = Docker(
+    path="my_flow.py",
+    files={"/source/of/my_flow.py": "my_flow.py"},
+    stored_as_file=True
```
jcrist, line 92:
```suggestion
    stored_as_script=True
```

src/prefect/environments/storage/_healthcheck.py
joshmeek: Thinking about it, it actually doesn't need to be a single flow file. Other storage options for file based only support a single file per storage option and I don't see a reason why docker couldn't still allow for multiple flows like it does right now. I will update this

docs/core/idioms/file-based.md
joshmeek: Ah yes this is only needed if you use prefect's default Docker build. I will update this idiom to mention this fact

Issue #2829, opened 2020-06-19
Allow use of KubernetesJobEnvironment + S3 storage + KubernetesAgent


jameslamb:
## Current behavior

As of #2666 , it's now possible to use non-docker storage with dockerized agents such as `KubernetesAgent`. This is a super exciting feature!

However, if you are using `S3` storage and `KubernetesAgent`, it seems that it's not possible to customize the jobs created for flow runs using `KubernetesJobEnvironment`.

I've observed that no errors are generated today (`prefect` 0.12.0) when you use this combination:

* `S3` storage
* `KubernetesJobEnvironment`
* `KubernetesAgent`

The agent I've created is happily creating jobs in the cluster for flow runs, but the manifests for those jobs are all default values and ignore anything I customize in `KubernetesJobEnvironment`.

After digging for a bit, I found the root cause. From ["Kubernetes Job Environment"](https://docs.prefect.io/orchestration/execution/k8s_job_environment.html#overview):

> The `KubernetesJobEnvironment` accepts an argument `job_spec_file` which is a string representation of a path to a Kubernetes Job YAML file. On initialization that Job spec file is loaded and stored in the Environment. It will never be sent to Prefect Cloud and will only exist **inside your Flow's Docker storage.**

I can also see this in code in `KubernetesJobEnvironment.create_flow_run_job()`, where the environment explicitly expects a docker image containing the `job_spec_file`

https://github.com/PrefectHQ/prefect/blob/26518173cf722c742eefcae970260edb5e385f54/src/prefect/environments/execution/k8s/job.py#L116-L149

## Proposed behavior

I'd like to be able to store the `job_spec_file` from `KubernetesJobEnvironment` in `S3` storage, so that the job details of a flow run using `S3` storage and run by `KubernetesAgent` can be customized.

## Example

The benefits of non-docker Storage are explained in https://docs.prefect.io/orchestration/execution/storage_options.html#non-docker-storage-for-containerized-environments. Adding this proposed behavior would allow flows using `KubernetesAgent` to take advantage of that storage without sacrificing the ability to customize the jobs using `KubernetesJobEnvironment`.

Without this proposed behavior, I think users of `KubernetesAgent` have to choose between non-docker storage and customizing their jobs.

joshmeek:
@jcrist Is this something you should address in your environment refactor?

jameslamb:
[...]

PR #2950, opened 2020-07-12
move some job_spec details for KubernetesJobEnvironment to build time

jameslamb:

## What does this PR change?

This PR is my attempt to implement this suggestion: https://github.com/PrefectHQ/prefect/issues/2829#issuecomment-652544616.

It splits `KubernetesJobEnvironment._populate_job_spec_yaml()` into two methods:

* `._populate_build_time_job_spec_details()`: fill in bits of the job spec that can be known at build time when  `KubernetesJobEnvironment` is initialized
* `_populate_run_time_job_spec_detailis()`: fill in details of the job spec that can only be determined at run time of a flow (like the `flow_run_id` and prefect-specific  environment variables)

## Why is this PR important?

Right now, all prefect customization of the user-provided job spec happens at run time. This means that it isn't possible to change the `args`  and `command` for the container that runs `prefect.environments.execution.load_and_run_flow()`.

This PR makes it clear which changes to the template happen at build time and which happen at run time. The "build time" changes get stored in the pickled flow, in `environment._job_spec` and all the run-time changes except setting `image:`are append-only (i.e. they don't override what the user provided in `job_spec_file`).

This change would allow users to do something like this:

```python
     job_spec_file="job_spec.yaml"
)
environment._job_spec["spec"]["template"]["spec"]["containers"][0]["command"] = [
    "/bin/bash",
    "-c"
]
environment._job_spec["spec"]["template"]["spec"]["containers"][0]["args"] = [
    "pip install some-library; python -c 'import prefect; prefect.environments.execution.load_and_run_flow()'"
]
```

In this case I'm `pip install`-ing a library, but this approach would allow users to run *any* code before running the flow.

I believe that this PR doesn't break anyone's existing code and improves users' ability to customize the flow run job. I also think it makes it more obvious which changes to the template provided in `job_spec_file` can be done at build time and which need to be done at run time.

Thanks for the time and consideration!

### DIFF
--- a/src/prefect/environments/execution/k8s/job.py
+++ b/src/prefect/environments/execution/k8s/job.py
@@ -40,9 +40,9 @@ class KubernetesJobEnvironment(Environment, _RunMixin):
     `$ /bin/sh -c "python -c 'import prefect; prefect.environments.execution.load_and_run_flow()'"`
 
     Args:
-        - job_spec_file (str, optional): Path to a job spec YAML file.
-            When this class is initialized, the contents of `job_spec_file` are read in
-            and stored in `self._job_spec`.
+        - job_spec_file (str, optional): Path to a job spec YAML file. This path is only
+            used when the environment is built, so should refer to a file on the machine
+            used to build the flow.
         - unique_job_name (bool, optional): whether to use a unique name for each job created
             with this environment. Defaults to `False`
         - executor (Executor, optional): the executor to run the flow with. If not provided, the
--- a/src/prefect/environments/execution/k8s/job.py
+++ b/src/prefect/environments/execution/k8s/job.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import uuid
 import warnings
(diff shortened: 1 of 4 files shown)


src/prefect/environments/execution/k8s/job.py
```diff
@@ -39,7 +40,9 @@ class KubernetesJobEnvironment(Environment, _RunMixin):
     `$ /bin/sh -c "python -c 'import prefect; prefect.environments.execution.load_and_run_flow()'"`
 
     Args:
-        - job_spec_file (str, optional): Path to a job spec YAML file
+        - job_spec_file (str, optional): Path to a job spec YAML file.
+            When this class is initialized, the contents of `job_spec_file` are read in
+            and stored in `self._job_spec`.
```
jcrist, line 45: I don't think we need to document the internals of this class - noting that the file should be on the machine that registers the flow, not the machine that runs the flow should be enough.

jameslamb: Ok got it, I can reword it.

jameslamb: I changed the wording in https://github.com/PrefectHQ/prefect/pull/2950/commits/2ea3a06eeb973ec036e034c871c66738d45375e1

jameslamb:
> Thanks for working on this! The idea seems sound - looks like you have a few test failures though.

thanks! I forgot to update the `serialization/` tests as well before I opened this. I just fixed those and looks like the build is :heavy_check_mark: now

jameslamb:
> Looks good to me! @jameslamb could you add a changes entry for this?

ah I always forget, sorry! Yep I'll do that right now.

jameslamb:
added a `changes/` entry in https://github.com/PrefectHQ/prefect/pull/2950/commits/a1f7f64951fd02ae8b57d1b0f792e241a6ef97b6

Issue #2656, opened 2020-05-26
Investigate reports of failed mapped tasks returning None to downstream tasks


lauralorenz:
## Description
I've heard from a contributor that an unstable mapping behavior occurs. The way I heard it was:
- in a mapped pipeline
- a dask worker unexpectedly dies
- the downstream task unexpectedly runs and receives 'None' as input, causing a runtime error because of the weird input

I also found reports of this in our slack history (archived here: https://github.com/PrefectHQ/prefect/issues/2655) that implied a link to specific deployment environments and for high volume mapped pipelines.

Note: is this possibly related to https://github.com/PrefectHQ/prefect/issues/2430?

## Expected Behavior
*What did you expect to happen instead?*
The upstream mapped task is Failed, and the downstream mapped task does not run.

## Reproduction
*A minimal example that exhibits the behavior.*
I have not observed it myself yet, but based on the slack thread it seems a high volume mapping task on an unstable network using `DaskKubernetesEnvironment` is the best way to reproduce.

## Environment
*Any additional information about your environment*

*Optionally run `prefect diagnostics` from the command line and paste the information here*

cicdw:
I think this is possibly solved by #2646 but I'm not sure how to confirm.

jcrist:
Without a reproducible example, I'm not sure how to progress on this, especially since it may have been resolved by the mapping refactor. +0.5 on closing if others are ok with it, since we don't have an immediate action plan or reproducer.

dylanbhughes:
I think this just happened to me: https://cloud.prefect.io/prefect/task-run/2cf36073-939b-47c5-9d51-a94089889e1b?logId=

cicdw:
@dylanbhughes do you have a `Result` configured for either your Flow or for any of your tasks?

dylanbhughes:
Just a GCS ResutHandler

dylanbhughes:
Here's another example: {staging_url}prefect-staging/flow-run/be28eebe-06c5-4cb9-b0b4-516722416ffd?logId=

dylanbhughes:
Another: {staging_url}/prefect-staging/task-run/fe1c7aac-c7ae-48de-9c97-5fba8c1a3f95?logId=

dylanbhughes:
Finally I'll say: looks unresolved by the mapping refactor. What I'm seeing is if there's a failure to communicate to cloud or the run is restarted by the zombie killer, mapped tasks in the same level can return `None` when resurrected

[...]

PR #2952, opened 2020-07-13
Fix zombie maps

cicdw:

## What does this PR change?
This PR fixes a very specific situation, where if a Mapped pipeline doesn't complete due to zombie tasks, _and_ if there is a "reducer" task downstream of the mapped task, on the next run of that pipeline the reducer task will receive incomplete data.  Ultimately this was caused by the task runners not properly re-loading mapped children state results.

## Why is this PR important?
Closes https://github.com/PrefectHQ/prefect/issues/2656

### DIFF
--- a/src/prefect/engine/state.py
+++ b/src/prefect/engine/state.py
@@ -111,6 +111,8 @@ class State:
                 s.load_result(result) if s is not None else None
                 for s in self.map_states  # type: ignore
             ]
+            if self.map_states:
+                self.result = [getattr(s, "result", None) for s in self.map_states]
             return self
 
         result_reader = result or self._result
--- a/src/prefect/engine/state.py
+++ b/src/prefect/engine/state.py
@@ -112,7 +112,9 @@ class State:
                 for s in self.map_states  # type: ignore
             ]
             if self.map_states:
-                self.result = [getattr(s, "result", None) for s in self.map_states]
+                self.result = [
+                    s.result if s is not None else None for s in self.map_states
+                ]
             return self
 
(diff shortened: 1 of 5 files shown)


src/prefect/engine/state.py
```diff
@@ -107,6 +107,12 @@ def load_result(self, result: Result = None) -> "State":
             - State: the current state with a fully hydrated Result attached
         """
         if self.is_mapped():
+            self.map_states = [
+                s.load_result(result) if s is not None else None
+                for s in self.map_states  # type: ignore
+            ]
+            if self.map_states:
+                self.result = [getattr(s, "result", None) for s in self.map_states]
```
jcrist, line 115: If `self.map_states` is an empty list, should `self.result` also be an empty list (rather than `None`)?

Also, is the only time `s` won't have a `result` attribute if `s` is `None`? If so, I'd prefer a check for `None` rather than `getattr`, as it better indicates the logic of what's happening.

cicdw: > If self.map_states is an empty list, should self.result also be an empty list (rather than None)?

No effect either way; went with this approach to avoid changing tests

> is the only time s won't have a result attribute if s is None

yup that's right; so the check would look like:
```python
self.result = [s.result if s is not None else None for s in self.map_states]
```
if you think that's clearer then I can make the change

jcrist: Yeah, I think that's clearer. Otherwise looks good to me!

cicdw: Updated 👍

PR #2961, opened 2020-07-14
Add option to log the output from stderr in the case of non-zero exit codes in ShellTask

joshmeek:

## What does this PR change?
Adds a `log_stderr` boolean kwarg to ShellTask and DbtShellTask for logging the full output in the case of a non-zero exit code.

## Why is this PR important?
User encountered a case where an error in their DbtShellTask didn't output the full traceback

### DIFF
--- a/src/prefect/tasks/shell.py
+++ b/src/prefect/tasks/shell.py
@@ -24,6 +24,9 @@ class ShellTask(prefect.Task):
         - return_all (bool, optional): boolean specifying whether this task
             should return all lines of stdout as a list, or just the last line
             as a string; defaults to `False`
+        - log_stderr (bool, optional): boolean specifying whether this task
+            should log the output from stderr in the case of a non-zero exit code;
+            defaults to `False`
         - **kwargs: additional keyword arguments to pass to the Task constructor
 
     Example:
@@ -48,6 +51,7 @@ class ShellTask(prefect.Task):
         helper_script: str = None,
         shell: str = "bash",
         return_all: bool = False,
+        log_stderr: bool = False,
         **kwargs: Any
     ):
         self.command = command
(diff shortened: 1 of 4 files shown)


src/prefect/tasks/shell.py
```diff
@@ -110,6 +115,10 @@ def run(self, command: str = None, env: dict = None) -> str:
                     sub_process.returncode, line
                 )
                 self.logger.error(msg)
+
+                if self.log_stderr:
+                    self.logger.error(lines)
```
jcrist, line 120: Perhaps join these with `"\n"`? Lines is a list, and will be repr'd in the logs, which looks a bit odd:

```
logger.info(["this", "looks", "weird"])
['this', 'looks', 'weird']
```


I might also merge this with the `Command failed...` log above, and drop the `line` from the format message (it appears `line` is the last line of stdout, which feels a bit odd to selectively choose :shrug:)

src/prefect/tasks/shell.py
```diff
@@ -107,9 +112,11 @@ def run(self, command: str = None, env: dict = None) -> str:
             sub_process.wait()
             if sub_process.returncode:
                 msg = "Command failed with exit code {0}: {1}".format(
-                    sub_process.returncode, line
+                    sub_process.returncode,
+                    "\n".join(lines) if self.log_stderr else line,
```
jcrist, line 116: This is about the previous behavior, but what's the reason for logging the last line from `stdout`? This feels like the wrong selection to me (in the case of multi-line output, this will look strangely cropped), and depending on the CLI may leak something secret if the user wasn't expecting it to be logged to cloud.

Also note that if no stdout is sent `line` will be `None`.

joshmeek: Ah okay good call, will make that change

Issue #2917, opened 2020-07-06
pandas results serializer proposal 


bobcolner:
Pandas dataframes are very common in python data workflows, often saved to disk in efficient binary formats. 

I'm sharing  parquet and feather serializers that I am using and might also benefit others. 

I'm not sure where or even if these pandas specific serializer fits into prefect core. What does are the teams thoughts? If we go down this route we will probably want to add CSV as well.
 

```
from io import BytesIO
import pandas as pd
import pyarrow.feather as pf
from prefect.engine.serializers import Serializer


class FeatherSerializer(Serializer):

    def serialize(self, value: pd.DataFrame) -> bytes:
        # transform a Python object into bytes        
        bytes_buffer = BytesIO()
        pf.write_feather(
            df=value,
            dest=bytes_buffer,
            version=2,
        )
        return bytes_buffer.getvalue()

    def deserialize(self, value:bytes) -> pd.DataFrame:
        # recover a Python object from bytes
        df_bytes_io = BytesIO(value)
        return pd.read_feather(df_bytes_io)


class ParquetSerializer(Serializer):

    def serialize(self, value: pd.DataFrame) -> bytes:
        # transform a Python object into bytes
        bytes_buffer = BytesIO()
        value.to_parquet(
            path=bytes_buffer,
            index=False
        )
        return bytes_buffer.getvalue()

    def deserialize(self, value:bytes) -> pd.DataFrame:
        # recover a Python object from bytes        
        df_bytes_io = BytesIO(value)
        return pd.read_parquet(df_bytes_io)
```

jcrist:
Nice work @bobcolner. My personal thoughts are that we should experiment with pandas specific serializers outside of prefect for a bit, just to see how they feel. Others may feel differently though.

Also, cc'ing @AndrewRook for thoughts, who's been thinking about this for a while (and submitted #2115, which kicked off the serializers in results work).

AndrewRook:
I may be rapidly losing my authority on the topic as I've not had much time to spend on this recently, but my personal perspective is that the best way to handle pandas serializing is to provide users with a single entry point — instead of having a `CSVSerializer`, `ParquetSerializer`, and so on, you provide a `PandasSerializer` that takes in how you want to serialize as an argument. I was thinking the options would be only the things that pandas supports (ie only stuff with a `DataFrame.to_[thing]` method), although I'm not sure where that would leave the feather support since in your example the serializer doesn't use a DataFrame method.

[...]

PR #2963, opened 2020-07-14
feature/add-dataframe-serializer

JacksonMaxfield:

## What does this PR change?

Adds a new `DataFrameSerializer` class that allows for basic dataframe -> CSV serialization or more complex formats with user provided `format` and `serialize_kwargs`.

## Why is this PR important?

Allowing easier checkpoint tracking and viewable files from `Result` objects when tasks are returning `pandas.DataFrames`.


JacksonMaxfield:
Bump

lauralorenz:
Hi @JacksonMaxfield! Just connecting a couple threads here... there are quite a few people interested in serializers for pandas types and there are a couple competing ideas going on from my view.

Your PR makes a single `DataFrameSerializer` that changes what it expects the bytes to be stored as (e.g. parquet, CSV) by the `format` argument
Alternatively, https://github.com/PrefectHQ/prefect/issues/2917 proposes separate serializers, one for each pandas convenience method (which is probably my personal bias)
In the comments in the proposal for https://github.com/PrefectHQ/prefect/issues/2917, the person who precipitated this whole idea  @AndrewRook seems to agree with your approach in general (https://github.com/PrefectHQ/prefect/issues/2917#issuecomment-655819695)

Tagging this here and also going to tag in that other thread, and see if we can get everyone interested in Pandas serializers together. Overall as a more official stance we are curious if people are actually using their custom pandas specific serializers in the wild (per https://github.com/PrefectHQ/prefect/issues/2917#issuecomment-655013705), so let us know how you feel about that if you have an opinion!

gryBox:
My vote is for one serializer.  :+1:

JacksonMaxfield:
Thanks @lauralorenz. I didn't even see the original issue, guess I checked for open PRs and not open issues related to "DataFrames"!

It's interesting you mention the "Single Format Single Serializer" route. I originally had that as this implementation in my own code as `DataFrameCSVSerializer` and `DataFrameParaquetSerializer` but as I was tinkering with them in my own pipelines I ended up getting really annoyed about having to go and import a different one each time so combined them.

So I guess I am saying "I support and wrote the generalized version because I wanted to change one line of code instead of two" :joy:

AndrewRook:
Thanks @lauralorenz, I had no idea this PR existed! I've got several thoughts:

1. I agree with the "one serializer, multiple data formats" approach used here. I think the most logical grouping is by data structure — not only to avoid the annoyance of having to change imports as @JacksonMaxfield describes but also because different data structures have different ways of encoding and decoding the data, and it would be a pain to have to enumerate those for a `CSVserializer`, `ParquetSerializer`, `HDF5Serializer`, etc. 
2. Pandas has way more IO methods than just CSV and Parquet, and I think it would be good to write one serializer for all of them. I was working on this on my own fork and have a working implementation of that using `getattr` [here](https://github.com/AndrewRook/prefect/blob/e9285f688f07f4749d5b4dd3678413800c23b127/src/prefect/engine/serializers.py#L113-L208).
3. (minor) Given that many packages have data structures called `DataFrame`, I'm not sure the serializer should be called a `DataFrameSerializer`. I called mine `PandasSerializer`, but honestly I don't really think that's much better (it's not like it will serialize a Pandas Series!). `PandasDataFrameSerializer` is a mouthful though. 

@lauralorenz to your question about usage, it's something that would be core to my workflow, so much so that I've now implemented this functionality something like 4 times as the different interfaces around results have changed 😄. It's also been independently suggested by at least 3 people now; not sure how that stacks up to other features that get implemented but to me it feels like a lot.

JacksonMaxfield:
>     2. Pandas has way more IO methods than just CSV and Parquet, and I think it would be good to write one serializer for all of them. I was working on this on my own fork and have a working implementation of that using `getattr` [here](https://github.com/AndrewRook/prefect/blob/e9285f688f07f4749d5b4dd3678413800c23b127/src/prefect/engine/serializers.py#L113-L208).

This is such a neat implementation with the `getattr`! I only chose CSV and parquet because I couldn't think of how to generalize it to all the formats so nice work on that :smile: 

>     3. (minor) Given that many packages have data structures called `DataFrame`, I'm not sure the serializer should be called a `DataFrameSerializer`. I called mine `PandasSerializer`, but honestly I don't really think that's much better (it's not like it will serialize a Pandas Series!). `PandasDataFrameSerializer` is a mouthful though.

I thought about this too with basically the same thought process of `Series` objects as well but sided on the `DataFrame` side I guess haha.

And seconding the "core to my workflow" statement as well.

AndrewRook:
> This is such a neat implementation with the getattr!

Feel free to take it if you like it. I'm happy to PR it in later but I'm also happy to skip that if you want to grab it now.

JacksonMaxfield:
@AndrewRook I think I should just close this PR and you open one with your version if the prefect team ends up agreeing with the "One Serializer to Rule Them All" implementation.

lauralorenz:
Cool, ok I think the situation is that the prevailing opinion here is to go for a
- single `PandasSerializer` (or similar name)
- that uses something like Andrew's fork to abstract away grabbing the correct pandas method
- for at least everything that has a convenience function in pandas itself
   - Any opinions now about what to do about things like FeatherSerializer (https://github.com/PrefectHQ/prefect/issues/2917) that are outside the pandas library, or deal with that later?

We discussed this internally yesterday at our weekly core meeting and agree with the majority here on the "one serializer to rule them all", so I think we should go for it. (will cross post on https://github.com/PrefectHQ/prefect/issues/2917 too)

If @AndrewRook agrees with you @JacksonMaxfield, let's focus on his fork!

AndrewRook:
@lauralorenz @JacksonMaxfield that's fine by me. I've still got some unit tests to write but I can make a WIP PR now and then we can iterate from there. 

As far as what to do about formats that don't have `to_`/`read_` methods, I put the functionality to identify those methods in special helper functions partially so that it's possible to inject special logic for cases like that relatively easily.

JacksonMaxfield:
Sounds good! This has been a wonderful little interaction in open source :tada: Thanks everyone, closing in favor of #3020

PR #2967, opened 2020-07-14
Drop upper bounds on dependencies

### LINKED ISSUE #2910: remove version ceilings on prefect's dependencies?

jcrist:
Previously prefect set upper bounds on all dependencies, which had to be bumped periodically. These upper bounds over constrained user environments, and usually weren't needed. Having a loose upper bound only provided a false sense of safety, since new versions may still break functionality (no package is exempt from having bugs), and `pip`'s resolver doesn't currently enforce these restrictions across `pip install` calls, so they may not be in effect anyway. In the case where a new version breaks something in prefect we can issue a release with a temporary restriction that's removed when the issue is resolved, or users can pin to a lower version (anyone deploying Python should be pinning their dependencies anyway).

We keep an upper bound on `croniter` per @cicdw, but unbound the rest.

Fixes #2910.


jcrist:
Looks like python-box just released 5.0, which broke a few things.

jcrist:
Added an issue for the `box` compatibility here: #2968.

PR #2973, opened 2020-07-15
Remove stale env vars on dask k8s and update work stealing default

cicdw:

## What does this PR change?
This PR:
- removes blocked handlers from the dask k8s environment spec (stale code essentially)
- changes the default value of work stealing to `True`

## Why is this PR important?
Turning work stealing off hampers the dask scheduler's ability to distribute work, and was originally turned off to avoid bad UX with version locking.  Might help #2958

### DIFF
--- a/src/prefect/environments/execution/dask/k8s.py
+++ b/src/prefect/environments/execution/dask/k8s.py
@@ -81,7 +81,7 @@ class DaskKubernetesEnvironment(Environment):
         self,
         min_workers: int = 1,
         max_workers: int = 2,
-        work_stealing: bool = False,
+        work_stealing: bool = True,
         scheduler_logs: bool = False,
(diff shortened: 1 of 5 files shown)


src/prefect/environments/execution/dask/k8s.py
```diff
@@ -81,7 +81,7 @@ def __init__(
         self,
         min_workers: int = 1,
         max_workers: int = 2,
-        work_stealing: bool = False,
+        work_stealing: bool = True,
```
jcrist, line 84: I didn't know this was the default, I'm pretty sure this explains #2958. Disabling this would mean that tasks assigned to one worker would never rebalance, and since autoscaling is used, all tasks would be assigned to the original single worker. +100 on changing this.

Issue #2100, opened 2020-02-28
Option to include item string in mapped task name


joeschmid:
## Use Case
Minor, non-urgent feature request: mapped tasks are currently referenced in logs and the Cloud UI using their numeric index, e.g. my_task[0], my_task[1]. These numeric values aren't particularly meaningful and, in some cases, the string value of the mapped item would be more useful during log examination or when viewing Flow runs in the Cloud UI.

## Solution
Having an opt-in way to enable use of the string version of a mapped item in the task name would be helpful, e.g.

```
    @task(mapped_item_in_task_name=True)
    def my_task(item):
        ...

    my_task.map(["apple", "banana", "orange"])
```

Which would yield task names of `my_task[apple]`, `my_task[banana]`, and `my_task[orange]` rather than `my_task[0]`, `my_task[1]`. and `my_task[2]`

For tasks mapping over multiple parameters, this option could simply choose to use the first parameter. Elements with long string representations could be truncated to a reasonable size for display purposes. This should *definitely* be opt-in to avoid logging sensitive data.

A better implementation might be to pass an iterable/list to be used for the task names (as a specific kwarg) which might or might not match the parameter to map over (but would need to have the same length):

```
    fruit = {
        0: "apple",
        1: "banana",
        15: "orange"
    }
    my_task.map(fruit.keys(), mapped_task_names=fruit.values())`
```
This would also yield task names of `my_task[apple]`, `my_task[banana]`, and `my_task[orange]`.

## Alternatives
None available currently.

jlowin:
This is an interesting enhancement that I want to think about; it might help solve some other enhancements (though its inclusion would require a change to Cloud, which currently types the `map_index` strictly as an int)

We also need to carefully review where (if anywhere) we have implicitly depended on ordered, numerical indices.

My second-step proposal would actually be to generalize `map_index` to a more broad `key` or `dynamic_key` that could be a stringified int by default, or taken from the item value (or a hash of the item value). The important thing is that on retries this `key` must be stable, so if the item value can change users shouldn't use it as the input (for example, if the value contained a timestamp it would change on retries). We would use this `key` to track a variety of dynamic tasks, including mapping today and some... other... things in the future ;)

jlowin:
[...]

PR #2974, opened 2020-07-15
Add display_name_callable to task.map

joshmeek:

## What does this PR change?
Addresses #2100 

## Why is this PR important?

-----

Opening this as a draft PR to get some thoughts on the implementation. This PR adds a `display_name_callable` to `task.map` (alternatively we could add this as a task attribute itself) and unlocks functionality like the following:

```python
# inputs as names

@task
def pval(val):
    print(f"My value is {val} and that should be my key ^")

with Flow("mapitems") as flow:
    pval.map(["abc", "def", "ghi"], display_name_callable=lambda **kwargs: kwargs["val"])

flow.run()
```

```
[2020-07-15 19:29:13] INFO - prefect.FlowRunner | Beginning Flow run for 'mapitems'
[2020-07-15 19:29:13] INFO - prefect.FlowRunner | Starting flow run.
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval': Starting task run...
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval': finished task run for task with final state: 'Mapped'
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[abc]': Starting task run...
My value is abc and that should be my key ^
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[abc]': finished task run for task with final state: 'Success'
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[def]': Starting task run...
My value is def and that should be my key ^
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[def]': finished task run for task with final state: 'Success'
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[ghi]': Starting task run...
My value is ghi and that should be my key ^
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[ghi]': finished task run for task with final state: 'Success'
[2020-07-15 19:29:13] INFO - prefect.FlowRunner | Flow run SUCCESS: all reference tasks succeeded
```

```python
# values from context

@task
def pval(val):
    print(f"My value is {val}")

with Flow("mapitems") as flow:
    pval.map(
        ["abc", "def", "ghi"],
        display_name_callable=lambda **kwargs: f"{kwargs['map_index']}-{kwargs['flow_name']}",
    )

flow.run()
```

```
[2020-07-15 19:31:52] INFO - prefect.FlowRunner | Beginning Flow run for 'mapitems'
[2020-07-15 19:31:52] INFO - prefect.FlowRunner | Starting flow run.
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval': Starting task run...
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval': finished task run for task with final state: 'Mapped'
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[0-mapitems]': Starting task run...
My value is abc
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[0-mapitems]': finished task run for task with final state: 'Success'
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[1-mapitems]': Starting task run...
My value is def
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[1-mapitems]': finished task run for task with final state: 'Success'
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[2-mapitems]': Starting task run...
My value is ghi
[2020-07-15 19:31:52] INFO - prefect.TaskRunner | Task 'pval[2-mapitems]': finished task run for task with final state: 'Success'
[2020-07-15 19:31:52] INFO - prefect.FlowRunner | Flow run SUCCESS: all reference tasks succeeded
```

There are a few `TODO`s in the code looking for feedback!

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -540,6 +540,7 @@ class Task(metaclass=SignatureValidator):
         upstream_tasks: Iterable[Any] = None,
         flow: "Flow" = None,
         task_args: dict = None,
+        display_name_callable: Callable = None,
         **kwargs: Any
     ) -> "Task":
         """
@@ -562,6 +563,8 @@ class Task(metaclass=SignatureValidator):
                 flow in context if no flow is specified
             - task_args (dict, optional): a dictionary of task attribute keyword arguments,
                 these attributes will be set on the new copy
+            - display_name_callable (Callable, optional): a function that can be used to format the
+                name of mapped tasks
             - **kwargs: keyword arguments to map over, which will elementwise be bound to the
                 Task's `run` method
 
@@ -579,6 +582,11 @@ class Task(metaclass=SignatureValidator):
                     )
                 )
         new = self.copy(**(task_args or {}))
+
+        # TODO: is setting attributes like this a weird access pattern?
+        if display_name_callable is not None:
+            setattr(new, "display_name_callable", display_name_callable)
+
         return new.bind(
             *args, mapped=True, upstream_tasks=upstream_tasks, flow=flow, **kwargs
         )
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -217,16 +217,25 @@ class TaskRunner(Runner):
         upstream_states = upstream_states or {}
         context = context or prefect.context.to_dict()
         map_index = context.setdefault("map_index", None)
-        context["task_full_name"] = "{name}{index}".format(
-            name=self.task.name,
-            index=("" if map_index is None else "[{}]".format(map_index)),
-        )
 
         task_inputs = {}  # type: Dict[str, Any]
 
         try:
             # initialize the run
             state, context = self.initialize_run(state, context)
+
+            # TODO: this attempt to retrieve raw inputs before the checks below. This could be a bad
+            # place to do this. The only issue with moving it later in the pipeline is the initial
+            # starting task run log below will not have the proper formatted display name meaning it will
+            # show up like this because we have logs prior to the initial input retrieval:
+            # prefect.TaskRunner | Task 'mytask[0]': Starting task run...
+            # prefect.TaskRunner | Task 'mytask[abc]': Handling state change from Pending to Running
+            # ...
+
+            # get the full task name
+            context["task_full_name"] = self.get_task_full_name(
+                state=state, upstream_states=upstream_states, map_index=map_index,
+            )
 
             # run state transformation pipeline
             with prefect.context(context):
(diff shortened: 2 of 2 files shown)


cicdw:
A few initial thoughts (which aren't particularly constructive but I thought were worth sharing):
- this is a lot of code for something that is ultimately organizational (not a "feature" per se)
- initially I thought "maybe we could configure this in the flow runner" but that wouldn't work across restarts / retries
- I worry that the ultimate idea is at odds with how our pipeline works - we usually don't collect the task inputs until deep in the pipeline, both to avoid unnecessarily loading data and because we can't assume data even exists until we've determined the upstream tasks have completed in the way the task expects (e.g., if the task's trigger fails or its upstreams aren't complete there are literally no inputs to reference), but if we want to use inputs to format the logger that means we always have to have inputs available at the very beginning

joshmeek:
> we usually don't collect the task inputs until deep in the pipeline

Yeah this is why I struggled to find a safe spot to put it. Ultimately where it exists now does not feel like the appropriate spot. If we see task inputs as the most desireable option to place in the logger name then having the name updated at the time of run could be the way to go. (where inputs are already properly loaded) Saying this because the biggest use for this feature is checking log output for debugging/errors and task content errors will occur during the run so they will be raised in the scope of the updated logger name

joshmeek:
I've had some more time to think this one through and now I am less sold on this implementation or even the need for such a feature. In it's current location the retrieval of task inputs prior to the start of the task runner pipeline will most definitely lead to unintended side effects.

I was discussing this a bit with @cicdw offline and he made the suggestion of possibly doing this in the `check_task_ready_to_map` function which occurs _after_ the proper task inputs are loaded:

https://github.com/PrefectHQ/prefect/blob/3acab303e70b7b10e2ff302bdb08f814311a0d96/src/prefect/engine/task_runner.py#L414-L451
This function is called on the parent mapped task and what could be done here is to make use of the State object to attach the formatted display callable data to it for all of the child mapped tasks to use. The biggest issue with this is it feels like abuse of the state pattern and has the potential to lead to large/bloated State objects.

WRT the need for such a feature I'm not fully sold on its benefit as opposed to simply writing a log (or print to stdout) inside the task code with the desired identification info. For example:

```
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[abc]': Starting task run...
<< error occurs >>
```

Feels no more helpful than:

```
[2020-07-15 19:29:13] INFO - prefect.TaskRunner | Task 'pval[0]': Starting task run...
[2020-07-15 19:29:14] INFO - my_logger | My inputs are: abc
<< error occurs >>
```

Where the error is still identifiable after the inputs are logged (you could even name the internal logger depending on the input value).

pedromachados:
I'm one of those people who would like this feature. I currently log inputs inside the task and that's how I find out what went wrong. I agree with @joshmeek that you can do your own logging. 

However, what would be really useful is if we could see on the UI something that identifies the task other than the task index. This would be helpful on the Gantt view as well as on the schematic view after drilling down into the mapped task. Would something link this be feasible with the current architecture?

joshmeek:
Superseded by #3420

PR #2980, opened 2020-07-16
Some cleanups to dbtshelltask

lauralorenz:

## What does this PR change?
First off, adds `return` that I think just got missed in the original dbtshelltask. Otherwise the stdout from dbt being run in the shell is not retrieved as the return value of the task, against what is says in its docstrings and the intention of ShellTask itself and presumably its derivatives. This was brought up in this slack thread: https://prefect-community.slack.com/archives/CL09KU1K7/p1594912760250100

While writing tests I realized that the super method was also failing to forward `command` and `env` kwargs which made the `@default_from_attrs` functionality not work properly for those two, so I fixed that too.

And while running tests that actually called dbt I realized that it wasn't being added as an `extras` option in setup.py (and thus it wasn't really possible to meaningfully run the tests on CI) so I've added that as well.

## Why is this PR important?
So that people can use the stdout from the DbtShellTask for anything downstream.

### DIFF
--- a/src/prefect/tasks/dbt/dbt.py
+++ b/src/prefect/tasks/dbt/dbt.py
@@ -86,10 +86,12 @@ class DbtShellTask(ShellTask):
         self.dbt_kwargs = dbt_kwargs
         super().__init__(
             **kwargs,
+            command=command,
+            env=env,
+            helper_script=helper_script,
             shell=shell,
             return_all=return_all,
-            log_stderr=log_stderr,
-            helper_script=helper_script
+            log_stderr=log_stderr
         )
 
     @defaults_from_attrs("command", "env", "dbt_kwargs")
--- a/src/prefect/tasks/dbt/dbt.py
+++ b/src/prefect/tasks/dbt/dbt.py
@@ -149,6 +149,4 @@ class DbtShellTask(ShellTask):
         if self.set_profiles_envar:
             os.environ["DBT_PROFILES_DIR"] = self.profiles_dir
 
-        super(DbtShellTask, self).run(
-            command=command, env=env,
-        )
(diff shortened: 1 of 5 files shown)


tests/tasks/dbt/test_dbt.py
```diff
@@ -12,6 +12,65 @@
...
+            "private_key_passphrase": "password123",
+        },
+        overwrite_profiles=True,
+        profiles_dir=str(dbt_dir),
+    )
+    out = task.run()
+    # when set to `return_all=True`, should return a list
+    assert isinstance(out, list)
+    # check that the result is multiple lines
+    assert len(out) > 1
```
lauralorenz, line 71: FYI the output of `dbt --version` I am depending on looks like this:

```
(prefect_dev) ➜  prefect git:(cleanup-on-dbt-task-return) dbt --version
installed version: 0.17.1
   latest version: 0.17.1

Up to date!

Plugins:
  - bigquery: 0.17.1
  - snowflake: 0.17.1
  - redshift: 0.17.1
  - postgres: 0.17.1
```

src/prefect/tasks/dbt/dbt.py
```diff
@@ -149,6 +151,4 @@ def run(
         if self.set_profiles_envar:
             os.environ["DBT_PROFILES_DIR"] = self.profiles_dir
 
-        super(DbtShellTask, self).run(
-            command=command, env=env,
-        )
+        return super(DbtShellTask, self).run(command=command, env=env,)
```
joshmeek, line 154: We should add this to the `/changes`

lauralorenz: Fixed in 41dbcb4!

PR #2995, opened 2020-07-17
Support passing in explicit `default=None` to Parameter

jcrist:
This allows creating a parameter with a default of `None` without explicitly setting `required=False`. I find this more intuitive to use - any value can be passed to `default` to create an optional `Parameter`, `None` is no longer special-cased.


jcrist:
CI failures are unrelated (possibly/probably due to cloudflare being down).

PR #2999, opened 2020-07-18
Update flows.md

zhen0:

## What does this PR change?

Switches the input for the set_schedule_active and set_schedule_inactive mutation examples in the docs to use schedule id instead of flow id.  

## Why is this PR important?

Flow id is only an input option in Cloud.  Schedule Id is an option in both cloud and server.  This switch makes the docs relevant to users of both cloud and server.


jcrist:
`flow_id` is used several other times in this page, should the other graphql examples also be updated?

cicdw (approved):
LGTM

PR #3007, opened 2020-07-21
add option to send in new context to new flow run from FlowRunTask

### LINKED ISSUE #2941: Expose context arg in FlowTaskRun task

**asandeep.me**: Hi all,
Is there a way to pass-on current context to next flow run triggered using `FlowRunTask` task?
Basically, I am sending some context parameter while creating a flow run using python client's `create_flow_run` method. The flow upon completion triggers other downstream flows.
Now, I want the context that I originally sent, available to downstream flows but I haven't had any success so far.

**laura**: Hi <@U013572S7RU>, looks like no since `FlowRunTask` doesn’t accept `context` right now (would eventually expect it to get passed along here <https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/prefect/flow_run.py#L109>), but I think that is a feature we can add pretty quick! I can make an issue from this ticket but first let me check if that fixes your issue. I think in that situation then your Flow A (kicked off with the python client) would have in it a node of `FlowRunTask` to kick off Flow B. You want to send context from Flow A (lets say `prefect.context.foo=bar`) to that Flow B. I just want to clarify that in practice, Flow A would need to have, or be able to re-derive just the added context (just `foo=bar`) since we probably don’t want to send the entire context object from Flow A straight into the `FlowRunTask` to ship to Flow B (as it has a bunch of other prefect sutff in it too). Would that achieve your use case?

**asandeep.me**: Thanks <@UTHEQ4F9R> for your reply.
Yes exactly this is what my use case is and selectively passing the required context to Flow B will work.

**laura**: <@ULVA73B9P> open “Expose context arg in FlowTaskRun task”

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1594286544388000?thread_ts=1594286544.388000&cid=CL09KU1K7).

lauralorenz:

- [n/a] add a changelog entry in the `changes/` directory (if appropriate)

## What does this PR change?
Adds in another change to FlowRunTask to the mega PR that allows people to send NEW context through to the newly generated flow run (closes #2941). I made this a PR to the mega PR to show what I wanted to add but it can go in separately too.

## Why is this PR important?
At least one person from the community is avoiding FlowRunTask and calling the client directly to be able to expose this, so this way they can use our official task instead!

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -22,6 +22,7 @@ class FlowRunTask(Task):
             running with Prefect Core's server as the backend, this should not be provided.
         - parameters (dict, optional): the parameters to pass to the flow run being scheduled;
             this value may also be provided at run time
+        - new_flow_context (dict, optional): the optional run context for the new flow run
         - wait (bool, optional): whether to wait the triggered flow run's state; if True, this
             task will wait until the flow run is complete, and then reflect the corresponding
             state as the state of this task.  Defaults to `False`.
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -34,23 +35,26 @@ def __init__(
         project_name: str = None,
         parameters: dict = None,
         wait: bool = False,
+        new_flow_context: dict = None,
```
lauralorenz, line 38: I avoided the kwarg `context` here at task initilization or runtime, I don't see it as an actual kwarg that gets passed to the Task in the code but in any case I find this naming less confusing

PR #3017, opened 2020-07-22
Update client's register method to accept headers

ThomasLaPiana:

## What does this PR change?

Allows request headers to be passed through to the flow.register() command.

## Why is this PR important?

I am using auth headers for my Prefect server deployment and want to be able to just authenticate by passing in the headers.

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -576,6 +576,7 @@ class Client:
         version_group_id: str = None,
         compressed: bool = True,
         no_url: bool = False,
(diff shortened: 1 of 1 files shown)


ThomasLaPiana:
@joshmeek do you have any preference on setting "default" headers that the client can inject into every request it makes or would you prefer passing headers on each method call specifically?

ThomasLaPiana:
looks like this is already possible using the following

```
c = Client()
c.attach_headers({my: auth})
c.register(my_flow, ...)
```

PR #3020, opened 2020-07-23
Pandas Serializer

AndrewRook:

## What does this PR change?

This PR adds a serializer to convert Pandas DataFrames to any format supported by the standard `pd.read_` and `df.to_` interface. It has the same functionality of #2963 with somewhat more flexibility. 

## Why is this PR important?
See #2963

@JacksonMaxfield @lauralorenz @bobcolner here's that PR, as promised! I've got some more tests to write and I'm holding off on updating the changelog until it's actually ready, but wanted to get this opened now for further discussion of the implementation.


AndrewRook:
@JacksonMaxfield @bobcolner since you two have contributed ideas to this too, if you're interested in being added to the contributor list let me know and I'd be happy to put you on the list.

JacksonMaxfield:
If it is allowed to have multiple contributors I would love to be added but no worries otherwise!

lauralorenz:
@JacksonMaxfield totally allowed! 😇

AndrewRook:
Ok, I think this is ready for review @lauralorenz. Couple of specific things I think need to be looked at:
1. I overrode the `__eq__` method, since `PandasSerializer` has more instantiation options. Is that the right approach here? I'm not sure where equality checks on serializers are used in the code so I wanted to ask.
2. Judging from the test commands, it seems like the CI/CD doesn't run tests with optional dependencies (I see "test vanilla prefect" as a step but not "test prefect with all the bells and whistles"). Right now this results in all the tests I wrote for `PandasSerializer` getting skipped. Is that the correct behavior?
3. I added a pandas extra in `setup.py`. Is there any other place I should add a pandas install option?

lauralorenz:
Awesome work @AndrewRook! I kicked it around briefly (in fact... I am going to show it off today in today's livestream on the Prefect youtube channel because it is a perfect fit for my topic on "Results and Resources"). Will take a deeper code review dive afterwards but at least for me as a user it definitely works!

A quick answer on 2/3: All the other test suites (3.6/3.7/3.8) run with all the bells and whistles (they install prefect[all_extras]) so your tests should be running for all of those checks. Besides including your dependency within some extras key (in your case a new one named 'pandas') there shouldn't be anywhere else you need to install it to get the tests to notice them configuration wise

AndrewRook:
Hey again, one thing that I just discovered when playing around with the serializer is that currently some of the `read_` methods (at least `read_csv`, I assume some others as well) will not decompress data from a buffer. The relevant parts of the [`read_csv` documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) (emphasis mine):

> compression{‘infer’, ‘gzip’, ‘bz2’, ‘zip’, ‘xz’, None}, default ‘infer’
>     For on-the-fly decompression of on-disk data. If ‘infer’ and **filepath_or_buffer is path-like**, then detect compression from the following extensions: ‘.gz’, ‘.bz2’, ‘.zip’, or ‘.xz’ (otherwise no decompression). If using ‘zip’, the ZIP file must contain only one data file to be read in. Set to None for no decompression.

As far as I can tell putting a workaround in the `PandasSerializer` itself would be clunky and brittle, so my thought would be to put a warning in the docs for the function that currently it can't deserialize some compressed file types (it _looks_ like serialization is fine, weirdly enough). Thoughts on that? Should I look harder for a robust workaround?

More generally, it might be worth thinking about decoupling (de)compression from the serializers — you could imagine something like what y'all did refactoring out the serialization methods from the `Results`, you could have a separate `Compression` class that handles (de)compression before/after passing data to the `Serializer`.

cicdw:
Hey @AndrewRook - I think we can worry about supporting compression schemes in future work; for now if we introduce a warning when compression is provided + add a comment to the docstring of this class we should be good.  90% of the time Prefect won't need to decompress / read the data from within a pipeline anyway.

jcrist:
> or now if we introduce a warning when compression is provided

Note that Pandas will already warn at runtime for this, and not all `to_*` methods don't support compression to a file-like object.

jcrist:
Thanks @AndrewRook, this is in. The user-facing API seems sound, we might change the implementation later to special case certain codepaths (e.g. csv to use `StringIO` instead of `BytesIO`, or warning on unsupported parameters), but that can be done later. Thanks for all the work here!

AndrewRook:
Thanks @jcrist, @lauralorenz, @JacksonMaxfield, and everyone else! Super excited this is in, can't wait to no longer have to run my ETL code off my prefect fork 😄

PR #3026, opened 2020-07-23
Add option to limit Docker Agent to only pull images from some registries.

thomasfrederikhoeck:

## What does this PR change?
Adds an option for Docker Agent to only pull from registries specified in the `reg_allow_list`

## Why is this PR important?
To ensure that if someone where to get access to your Prefect Account they would not be able to run their own images.

### DIFF
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -58,6 +58,8 @@ class DockerAgent(Agent):
         - docker_interface (bool, optional): Toggle whether or not a `docker0` interface is
             present on this machine.  Defaults to `True`. **Note**: This is mostly relevant for
             some Docker-in-Docker setups that users may be running their agent with.
+        - reg_allow_list (List[str], optional): Limits Docker Agent to only pull images
+            from the listed registries.
     """
 
     def __init__(
@@ -74,6 +76,7 @@ class DockerAgent(Agent):
         show_flow_logs: bool = False,
         network: str = None,
         docker_interface: bool = True,
+        reg_allow_list: List[str] = None,
     ) -> None:
         super().__init__(
             name=name,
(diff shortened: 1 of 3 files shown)


src/prefect/agent/docker/agent.py
```diff
@@ -330,11 +338,15 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
             self.logger.info("Pulling image {}...".format(image))
-
-            pull_output = self.docker_client.pull(image, stream=True, decode=True)
-            for line in pull_output:
-                self.logger.debug(line)
-            self.logger.info("Successfully pulled image {}...".format(image))
+            registry = image.split("/")[0]
+            if self.reg_allow_list and registry not in self.reg_allow_list:
+                self.logger.warning("Trying to pull image from a Docker registry '{}' which"
+                                    " is not in the reg_allow_list".format(registry))
```
joshmeek, line 344: Instead of continuing on a non-authorized image I think this should raise a ValueError or something of the like (and also keep the error log). Then the outside call to the `deploy_flows` function will safely trap the error and fail the flow run with this reason

thomasfrederikhoeck: Should it be an error instead of warning then?

joshmeek:
Extra comment: @thomasfrederikhoeck it looks like your local git may not authenticated with your github account (commits are showing up as grey octocat logo):

If you want you could try this:
https://docs.github.com/en/enterprise/2.18/user/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user
https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git

src/prefect/agent/docker/agent.py
joshmeek: Yeah let's update it to an error log

thomasfrederikhoeck:
@joshmeek  I have fixed it know so it associates with my account. Anything else needed?

PR #3030, opened 2020-07-24
Issue 3027 remove storage objects check get flows

berosen:

## What does this PR change?

These changes will allow a user to initialize an S3 Storage object with a bucket name and object key, then call `get_flow()` without any args to retrieve that object from S3. Previously, a call to `add_flow()` would need to have been made to add entries to `self.flows()` and `self._flows()`, or the dictionaries would have to be set by the user. If this was not done `get_flow()` would raise an exception. 

## Why is this PR important?

Simplifies process of downloading a Flow from S3 Storage.

### DIFF
--- a/changes/issue3027.yaml
+++ b/changes/issue3027.yaml
@@ -1,0 +1,5 @@
+
+fix:
+  - "Move initial check in `get_flows()` to be checked when a flow_location generated from `add_flow()` is provided. - [#3027](https://github.com/PrefectHQ/prefect/issues/3027)"
+
+
--- a/changes/issue3027.yaml
+++ b/changes/issue3027.yaml
@@ -1,6 +1,6 @@
 
 fix:
-  - "Move initial check in `get_flows()` to be checked when a flow_location generated from `add_flow()` is provided. - [#3027](https://github.com/PrefectHQ/prefect/issues/3027)"
+  - "Fix check of flow existence in storage object `get_flow` to only occur when provided - [#3027](https://github.com/PrefectHQ/prefect/issues/3027)"
 
 contributor:
-    - "[berosen](https://github.com/berosen)"
+  - "[berosen](https://github.com/berosen)"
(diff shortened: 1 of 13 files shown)


joshmeek:
@berosen it looks like your local git may not be authenticated with your github account (commits are showing up as grey octocat logo):

If you want you could try this:
https://docs.github.com/en/enterprise/2.18/user/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user
https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git

src/prefect/environments/storage/s3.py
```diff
@@ -57,24 +57,32 @@ def __init__(
...
-            raise ValueError("Flow is not contained in this Storage")
+        if flow_location:
+            if flow_location not in self.flows.values():
+                raise ValueError("Flow is not contained in this Storage")
+
+        elif self.key:
+            flow_location = self.key
+
+        else:
+            raise Exception("No Flow location provided")
```
joshmeek, line 85: Let's change this to something like ValueError (more specific as compared to Exception)

changes/issue3027.yaml
```diff
@@ -0,0 +1,5 @@
+
+fix:
+  - "Move initial check in `get_flows()` to be checked when a flow_location generated from `add_flow()` is provided. - [#3027](https://github.com/PrefectHQ/prefect/issues/3027)"
+
+
```
joshmeek, line 5:
```suggestion
contributor:
    - "[berosen](https://github.com/berosen)"
```

src/prefect/environments/storage/s3.py
berosen: Will do

berosen:
@joshmeek oops, sorry about that, I had a different email configured than usual and forgot to add it to my settings. 

> If you would be willing I actually want to make some commits on this branch to extend this pattern to the other storage types as well

Sure thing!

PR #3031, opened 2020-07-25
Update FargateTaskEnvironment to throw if task definition is inconsistent with existing task definition

zpencerq:

## What does this PR change?
This PR adds a validation step in the `FargateTaskEnvironment` setup to verify the given task definition. 

It attempts to give a very helpful `ValueError` to point out the exact inconsistency.

For instance, if you update the image and cpu in your `FargateTaskEnvironment` you'll be greeted with:
```
ValueError: The given taskDefinition does not match the existing taskDefinition {family}.
Detail:
    containerDefinition.0.image -> Given: test/image:newtag, Expected: test/image:tag
    cpu -> Given: 1024, Expected: 512

If the given configuration is desired, deregister the existing
taskDefinition and re-run the flow. Alternatively, you can
change the family/taskDefinition name in the FargateTaskEnvironment
for this flow.
```

## Why is this PR important?
Normal operation of `FargateTaskEnvironment` can lead to unexpectedly stale runtime behavior. Currently, the `setup` method will register the task definition that is used for the flow run itself. These task definitions are named by the `family` attribute and versioned in ECS. Since the environment of a flow is critical to ensuring consistency in the flow run, this change ensures that the existing task definition matches the configuration of the `FargateTaskEnvironment`.

### Scenario
0. You start with no task definitions registered in ECS.
1. You register a flow `my-flow` with a `FargateTaskEnvironment` specifying an image of `datadeps:1` and `family` of `fuzzy`. 
2. You run `my-flow`, which will create a task-definition `fuzzy:1`.
2. You update the `datadeps` image and simply update the `FargateTaskEnvironment` of the flow to use image `datadeps:2` and register `my-flow` again.
3. When this new version of `my-flow` runs, it will use the existing task definition which will use the outdated image `datadeps:1` almost certainly causing unexpected behavior.

An alternative approach could be to simply create a new task definition every time there's a change. However, since the revision numbers are out of the enduser's control, this would impact previous versions of flows; the previous flow versions run newer versions of the task definitions which could cause inconsistencies. 
Maybe there's some approach with tagging and version groups that could work? That's a bit out of my depth 😅 

As an aside, could this be too aggressive of a change? It could break existing deployments, perhaps introducing a warning is enough? Where does `prefect` draw the line in avoiding inconsistent flow execution?

### DIFF
--- a/changes/pr3031.yaml
+++ b/changes/pr3031.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Update FargateTaskEnvironment to throw if task definition is inconsistent with existing task definition - [#3031](https://github.com/PrefectHQ/prefect/pull/3031)"
+
+contributor:
+  - "[Spencer Ellinor](https://github.com/zpencerq)"
--- a/src/prefect/environments/execution/fargate/fargate_task.py
+++ b/src/prefect/environments/execution/fargate/fargate_task.py
@@ -10,6 +10,24 @@ from prefect.utilities.storage import get_flow_image
 if TYPE_CHECKING:
     from prefect.core.flow import Flow  # pylint: disable=W0611
 
+_DEFINITION_KWARG_LIST = [
+    "family",
+    "taskRoleArn",
+    "executionRoleArn",
+    "networkMode",
+    "containerDefinitions",
+    "volumes",
+    "placementConstraints",
+    "requiresCompatibilities",
+    "cpu",
+    "memory",
+    "tags",
+    "pidMode",
+    "ipcMode",
+    "proxyConfiguration",
+    "inferenceAccelerators",
+]
+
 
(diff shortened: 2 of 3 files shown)


joshmeek:
This is a great enhancement for that environment @zpencerq! Would you mind adding a `changes/` entry for this? https://github.com/PrefectHQ/prefect/tree/master/changes

> As an aside, could this be too aggressive of a change? It could break existing deployments, perhaps introducing a warning is enough? Where does prefect draw the line in avoiding inconsistent flow execution?

Not aggressive at all, I would say this is a much more desired behavior for this environment. Using the same `family` across versions of a flow could definitely lead to unexpected behavior. Currently it runs with the assumption here that users would be setting a different one for each flow version (which I'm sure isn't always the case) and this difference taken in this PR is a nice way to "enforce" it.

PR #3042, opened 2020-07-28
prevent assignment of duplicate labels to agents

jameslamb:

## What does this PR change?

This PR attempts to prevent the situation where an agent has duplicate labels.

## Why is this PR important?

While testing #3000 , I noticed something surprising...it's possible for duplicates to exist in an agent's set of labels:

For example:

```shell
prefect agent start \
    -t ${PREFECT_RUNNER_TOKEN} \
    --label 'i-am-duplicated' \
    --label 'i-am-duplicated' \
    --label 'i-am-duplicated'
```

image

image

I'm not sure if this would cause any problems today, but I could see it causing these types of problems:

* breaking code that allows wildcards for matching flows to agents, but throws an error if more than one label is matched
* breaking code that queries for agents by label and cannot handle duplicates
* slightly hurting the experience in the Prefect Cloud UI if duplicate labels are enough to push other labels into the collapsed `+n` box like in the screenshot above

So, I'm not aware of anything that is not working well *today* because of this, but I think this change might be a minor way to prevent some weird cases in the future.

### DIFF
--- a/changes/pr3042.yaml
+++ b/changes/pr3042.yaml
@@ -1,0 +1,5 @@
+fix:
+  - 'Prevent duplicated agent labels - [#3029](https://github.com/PrefectHQ/prefect/pull/3042)'
+
+contributor:
+  - "[James Lamb](https://github.com/jameslamb)"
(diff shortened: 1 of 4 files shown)


zdhughes:
Just confirmed-- label matching for run retrieval is done by way of sets, so the list `["foo", "foo"]` gets collapsed down to the set `{"foo"}` on the API side. It looks like the same is not true for label-based flow run concurrency, but I think that's likely an oversight on the Cloud side. 👍

joshmeek:
@zdhughes Awesome thanks for clarifying! Since this is only agent related it's good to go 👍

PR #3035, opened 2020-07-27
Use relative paths in generated Dockerfile

pravindahal:
**Thanks for contributing to Prefect!**

Please describe your work and make sure your PR:

- [x] adds new tests (if appropriate)
- [x] add a changelog entry in the `changes/` directory (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

## What does this PR change?

Before the fix, in the temporary Dockerfile that is created when using Docker storage for a flow, full paths to files are inserted. This causes docker build to fail as reported here: https://github.com/PrefectHQ/prefect/issues/3023

Issue #3023, opened 2020-07-23
Path for .flow location should be relative when using Docker Storage through SSH


thomasfrederikhoeck:
## Description
When using Docker storage on windows with SHH to a docker daamon the build context is copied over, with regards to the placement of the Dockerfile which is located in `project` but prefect inserts the whole path for the file on the local Windows machine. However this doesn't exists in the build context.
```

Step 16/21 : RUN mkdir -p /opt/prefect/
 ---> Running in 4f18a09c30b0
Removing intermediate container 4f18a09c30b0
 ---> 4d6f5c0b81eb
Step 17/21 : COPY C:\Users\tfh\project\tmp3dwq67c0\etl.flow /opt/prefect/flows/etl.prefect
COPY failed: stat /var/lib/docker/tmp/docker-builder766472190/C:Userstfhprojecttmp3dwq67c0etl.flow: no such file or dir
ectory
Traceback (most recent call last):
  File "build_flows.py", line 31, in <module>
    build_image(all_flows)
  File "build_flows.py", line 23, in build_image
    storage = storage.build(push=push)
  File "C:\Users\tfh\Miniconda3\envs\prefect\lib\site-packages\prefect\environments\storage\docker.py", line 315, in build
    self._build_image(push=push)
  File "C:\Users\tfh\Miniconda3\envs\prefect\lib\site-packages\prefect\environments\storage\docker.py", line 38  File "C:\Users\tfh\Miniconda3\envs\prefect\lib\site-packages\prefect\environments\storage\docker.py", line 31
5, in build
    self._build_image(push=push)
  File "C:\Users\tfh\Miniconda3\envs\prefect\lib\site-packages\prefect\environments\storage\docker.py", line 38
0, in _build_image
    "Your docker image failed to build!  Your flow might have "
ValueError: Your docker image failed to build!  Your flow might have failed one of its deployment health checks
 - please ensure that all necessary files and dependencies have been included.

```

## Expected Behavior
I expect the COPY statement to be relative such that the stamtement is `COPY etl.flow /opt/prefect/flows/etl.prefect`

pravindahal:
Also, the paths for healthcheck.py and custom files if any are affected. I will submit a PR.

thomasfrederikhoeck:
[...]

PR #3044, opened 2020-07-28
Revert back to using relative paths in generated Dockerfile on Windows

pravindahal:

## What does this PR change?
https://github.com/PrefectHQ/prefect/pull/2390 introduced a change so that on Windows, Dockerfile contained absolute paths which prevents docker build from working when using wsl backend or ssh (see https://github.com/PrefectHQ/prefect/issues/3023).

The fix here is to use relative paths in Dockerfile (i.e. revert to behavior before https://github.com/PrefectHQ/prefect/pull/2390) but pass absolute path of Dockerfile to `client.build`.

For previous attempt at addressing this problem, see the discussion here: https://github.com/PrefectHQ/prefect/pull/3035

### DIFF
--- a/changes/pr3044.yaml
+++ b/changes/pr3044.yaml
@@ -1,0 +1,6 @@
+
+fix:
+  - "Fix use of absolute path in Docker storage on Windows - [#3044](https://github.com/PrefectHQ/prefect/pull/3044)"
+
+contributor:
+  - "[Pravin Dahal](https://github.com/pravindahal)"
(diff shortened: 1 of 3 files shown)


joshmeek:
Thanks for the PR @pravindahal! While the code _appears_ fine I don't have a windows environment to give this a test just yet so maybe someone can verify?

lauralorenz:
Hi @pravindahal @joshmeek for what it's worth, I just tried out the test case back from https://github.com/PrefectHQ/prefect/issues/2332 but using this branch on regular Windows 10 AND MacOS and had no issues, so since @pravindahal can confirm it works for wsl/ssh I think we are good.

lauralorenz:
also cc'ing @thomasfrederikhoeck in case they want to check

thomasfrederikhoeck:
@lauralorenz I will check it within a couple of hours

thomasfrederikhoeck:
LGTM 👍

PR #3072, opened 2020-08-01
Ignore calls to flow.register when parsing a flow using file based storage

### LINKED ISSUE #3051: Ignore calls to flow.register when parsing a flow using file based storage
## Current behavior

If a flow script contains a reference to the `.register` method, it'll fail with this error:
```
Failed to load and execute Flow's environment: RuntimeError('Attempting to call `flow.register` during execution of flow file will lead to unexpected results.')
```

## Proposed behavior

Flow runner ignores calls to the `.register` method.

sp1thas:

I created an `ast.NodeTransformer` in order to exclude `flow.register` statement then the user tries to register a flow using file based storage. 

## What does this PR change?

This PR fixes #3051 . 

## Why is this PR important?

`RuntimeError` will not be raised when the user tries to register a flow using the file based storage and `flow.register` statement already exists in the script.


sp1thas:
Hi @cicdw and thank you for your review. When these warning occurs, return statement should be something different or the code flow will remains as is?

cicdw:
I think that after the warning is raised, a `return None` is sufficient to exit from the code without performing any actions

PR #3084, opened 2020-08-04
Add Task to support Databricks job submission

VincentTNR:

## What does this PR change?

- This PR adds a DatabricksRunSubmit Task, which supports job submission through Databricks (see https://docs.databricks.com/dev-tools/api/latest/jobs.html#runs-submit)
- It also adds appropriate tests
- Updates the changelog

## Why is this PR important?
Databricks is commonly used when building data pipelines.


joshmeek:
Thanks for the PR @VincentTNR! 🙂  Haven't taken a deeper look into the implementation yet but for starters could you move this task under the `/contrib/tasks` directory? https://github.com/PrefectHQ/prefect/tree/master/src/prefect/contrib

VincentTNR:
@joshmeek Done ! The task has been moved under the /contrib/tasks directory

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+        - spark_jar_task: Indicates that this job should run a JAR.
+        - notebook_task: Indicates that this job should run a notebook.
+                         This field may not be specified in conjunction with spark_jar_task.
+        - spark_python_task: Indicates that this job should run a Python file.
+        - spark_submit_task: Indicates that this job should run spark submit script
+        - new_cluster: A description of a cluster that will be created for each run.
+        - existing_cluster_id: The ID of an existing cluster that will be used for all runs of this job
+        - existing_cluster_name: The name of the cluster, will trigger an extra API call
+        - libraries: An optional list of libraries to be installed on the cluster
+        - timeout_seconds: An optional timeout applied to each run of this job, default is no timeout
```
d80tb7, line 30: Are the descriptions for `timeout_seconds` and `polling_timeout` mixed up here?  The decsriptions make it sound like the former is the timeout for the entire run and the latter is the timeout for the the run state request, however the former defaults to 10 seconds and the latter to one hour.

Also the  doc says that ```timeout_seconds``` is optional and default to no timeout, but the function signature seems to default it to 10.

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+            - timeout: Timeout
+
+        Returns:
+            - Run response payload as python dict,
+              see https://docs.databricks.com/dev-tools/api/latest/jobs.html#runs-get
+        """
+        query = f"{self._end_point}/jobs/runs/get?run_id={run_id}"
+        start_time = time()
+        while time() - start_time < timeout:
+            response = requests.get(
```
d80tb7, line 177: It looks like this won't retry the request.  This means that if there is a transient failure in the databricks api you'll end up erroring out here and losing track of the job.  Worst case is that the job is still running fine and you end up scheduling another one.

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+                        "Authorization": f"Bearer {self._databricks_token}",
+                        "User-Agent": USER_AGENT,
+                    },
+                    timeout=self._timeout_seconds,
+                )
+
+                response.raise_for_status()
+                json_data = response.json()
+                if "run_id" not in json_data:
+                    raise requests.exceptions.RequestException
```
d80tb7, line 232: Could we give a better error message here?

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+
+            response.raise_for_status()
+            data = response.json()
+            run_page, run_name, lifecycle_state = (
+                data["run_page_url"],
+                data["run_name"],
+                data["state"]["life_cycle_state"],
+            )
+
+            if "result_state" in data["state"]:
```
d80tb7, line 194: I don't think checking the presence of `result_state` for task termination is correct here.   Based on https://docs.databricks.com/dev-tools/api/latest/jobs.html#runresultstate there appear to be various cases where the run can have finished but `result_state` is empty.  Instead I think you need to check whether `life_cycle_state` is a terminal state.

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+
+                if result_status != "SUCCESS":
+                    raise Exception(
+                        f"Task {run_name} failed with status {result_status}... "
+                        f"For more info please visit {run_page}"
+                    )
+                return
+
+            except requests.exceptions.RequestException as error:
+                self.logger.error(error)
```
d80tb7, line 254: we should log how many attempts we have done here and whether or not we are going to  retry.

src/prefect/contrib/tasks/databricks/databricks.py
```diff
@@ -0,0 +1,260 @@
...
+        if cluster_name not in clusters:
+            raise Exception(f"Cluster {cluster_name} does not exist !")
+        return clusters[cluster_name]
+
+    def _wait_for_run_to_complete(
+        self, run_id: str, polling_period: int, timeout: int = 3600
+    ) -> dict:
+        """
+        Fetch Databricks run status via api/../jobs/run/get?run_id=<RUN_ID> until run state
+        is terminated. A run is terminated when matching either SUCCESS, FAILED, TIMEDOUT or CANCELED
```
d80tb7, line 163: I don't believe this is is what the code does.  It doesn't check that the run state is is in the set you give, but rather considers the run terminated when any run state is given.

d80tb7:
Hi @VincentTNR 

Thanks for adding this- imho a Databricks task would be a really good addition too Prefect.

That said I'm not sure that this code catches all the edge cases (which is not surprising as the databricks API is pretty complex).  Have you considered taking a look at the [Databricks Airflow Hook and Operator](https://github.com/apache/airflow/tree/master/airflow/providers/databricks) and porting those to Prefect?  I think the paradigms are so similar that this should be easier than writing from scratch and I think they've solved many of the edge cases through experience!

If you don't want to go this route, one suggestion I do have is to separate out the Databricks API logic from the Prefect task logic.  I think that might make the code a bit cleaner and thus a bit easier to reason about,

Ashton-Sidhu:
@VincentTNR hey if you don't have the time to port the Databricks Airflow Hook & Operator over to Prefect, I can give it a go. Let me know!

joshmeek:
This has been superseded by #3166

PR #3111, opened 2020-08-07
Fix breaking change with 0.13.0 agents not able to run old flows

cicdw:


src/prefect/agent/kubernetes/job_spec.yaml
```diff
@@ -17,7 +17,7 @@ spec:
           image: prefecthq/prefect:latest
           imagePullPolicy: IfNotPresent
           command: ["/bin/sh", "-c"]
-          args: ["prefect execute flow-run"]
+          args: ["prefect", "execute", "cloud-flow"]
```
cicdw, line 20: @joshmeek is this OK?

joshmeek: I believe so

cicdw: Gonna be safe and make it no commas

PR #3046, opened 2020-07-28
allow customization of the yaml template

emcake:
**Thanks for contributing to Prefect!**

Please describe your work and make sure your PR:

- [ ] adds new tests (if appropriate)
- [ ] add a changelog entry in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

Note that your PR will not be reviewed unless all three boxes are checked.

## What does this PR change?

Add an extra configurable `YAML_TEMPLATE` argument to the `KubernetesAgent` to allow overwriting of the template job spec.

## Why is this PR important?

Due to environmental setup of our k8s cluster I found myself needing to alter the podspec run in ways not exposed by current environment variables, and it didn't seem sensible to add new environment variables for every change I'd want to make to the template. This approach allows me to build a custom container with a specialized yaml spec.

PR #3113, opened 2020-08-07
Determine flow run command based on flow version

joshmeek:

## What does this PR change?
Adds an agent utility for determining the flow run execution command based on the version of the flow. For core_version < `0.13.0` it will use `execute cloud-flow` and for versions >= `0.13.0` it will use `execute flow-run`.

## Why is this PR important?
Maintain backwards compatibility while also respecting the deprecation of commands

### DIFF
--- a/src/prefect/agent/kubernetes/job_spec.yaml
+++ b/src/prefect/agent/kubernetes/job_spec.yaml
@@ -17,7 +17,7 @@ spec:
           image: prefecthq/prefect:latest
           imagePullPolicy: IfNotPresent
           command: ["/bin/sh", "-c"]
-          args: ["prefect execute cloud-flow"]
+          args: ["prefect execute flow-run"]
           env:
             - name: PREFECT__CLOUD__API
               value: PREFECT__CLOUD__API
--- a/src/prefect/utilities/agent.py
+++ b/src/prefect/utilities/agent.py
@@ -1,3 +1,5 @@
+from distutils.version import LooseVersion
+
 import prefect
 from prefect.utilities.graphql import GraphQLResult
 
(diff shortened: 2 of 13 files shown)


src/prefect/agent/kubernetes/job_spec.yaml
```diff
@@ -17,7 +17,7 @@ spec:
           image: prefecthq/prefect:latest
           imagePullPolicy: IfNotPresent
           command: ["/bin/sh", "-c"]
-          args: ["prefect execute cloud-flow"]
+          args: ["prefect execute flow-run"]
```
cicdw, line 20: Are there any ways for users to override this initial image with something that might be have older version of Prefect?

src/prefect/utilities/agent.py
```diff
@@ -37,3 +39,22 @@ def get_flow_image(flow_run: GraphQLResult) -> str:
...
+    Determine the flow run command to use based on a flow's version. This is due to a command
+    deprecation in `0.13.0` where `execute cloud-flow` was changes to `execute flow-run`
+
+    Args:
+        - flow_run (GraphQLResult): A GraphQLResult flow run object
+
+    Returns:
+        - str: a prefect CLI command to execute a flow run
+    """
+    core_version = flow_run.flow.core_version
```
cicdw, line 55:
```suggestion
    core_version = flow_run.flow.core_version or "0.0.0"
```
I think there are some old (but still technically supported) versions of Core that didn't set this attribute at registration time, so just to be safe

src/prefect/utilities/agent.py
```diff
@@ -1,3 +1,5 @@
+from distutils.version import LooseVersion
```
cicdw, line 1: Ah good call using a built-in!

src/prefect/agent/kubernetes/job_spec.yaml
joshmeek: We don't expose a setting/kwarg for it but if something like #3046 is adopted then there is a chance it could happen

src/prefect/utilities/agent.py
joshmeek: Yeah I hate welcoming in extra dependencies

src/prefect/agent/kubernetes/job_spec.yaml
cicdw: yea, I guess let's just remember this when we go to refactor agents / environments

PR #3128, opened 2020-08-10
Doc fix

andykawabata:

## What does this PR change?

- changes the stated return value `Dict`  in Flow.edges_from() and Flow.edges_to() methods to match their actual return value `Set`.

## Why is this PR important?

documentation should be accurate


jcrist:
Thanks for the PR! For a docs-only fix we don't need to add a changelog entry, would you mind dropping the changelog file? Happy to merge after that!

PR #3139, opened 2020-08-11
Add decorator for per-flow-run/per-process caching

jcrist:
Some flows create temporary resources (database connections, boto clients, etc...).

*Note: in this case, "resources" are temporary objects that are used during a flow, but whose failure to cleanup **isn't** detrimental. This excludes expensive things like cloud resources, where we want to treat the cleanup step like a task, but includes things like boto clients where failure to close cleanly is only unpleasant not detrimental.*

When running locally (i.e. no Dask) & without checkpointing, these objects can be returned from tasks and efficiently shared without the need for pickling. However, once multiple processes are introduced, it becomes harder to share these resources:

- not all resource objects are pickleable
- some resources that *are* pickleable will recreate the connection on unpickling. Depending on the execution backend this may result in the same resource being recreated multiple times even in the same process.
- Prefect `tasks` (when running with checkpointing) expect to be able to store and retrieve the result from disk, but temporary resources like database connections don't really make sense to checkpoint.

In #2891 there was an attempt to create a new type of "resource" task to handle this pattern. This solved some of these issues, but failed to support enough task semantics (run exactly once, can checkpoint to disk) that the idea was scrapped. It feels like treating things that create temporary resources as tasks themselves is error prone and leads to some rough edges.

---

This PR attempts to solve the same issues, but does so in a way that doesn't conflict with task semantics.

Users requiring temporary resources are encouraged to use a new decorator (currently called `memoize`, though I don't love that name). Functions decorated with the `memoize` decorator have their results cached per-process during a flow run - repeated calls in the same process with the same arguments will only execute once and will return the cached result otherwise. At the end of a flow run, all caches are cleared, releasing all resources.

Wrapped functions should **not** be treated as tasks themselves (don't apply a `task` decorator on top of a `memoize` decorator) as that defeats the purpose. Rather, they should be treated as helper functions to call from within other tasks. Any call to a memoized function *may* execute, but it also *may return a cached result instead. This makes it explicit all the places where a resource may be created, rather than in #2891 where this was implicit making it hard to surface and deal with errors.

Wrapped functions may optionally `yield` their result instead of returning it directly. If this is done, code after the `yield` will be run on flow-run completion as a cleanup step.

An example:

```python
from prefect import task
from prefect.utilities.local_cache import memoize

@memoize
def my_client(some_token):
    with create_client(some_token) as client:
        yield client

@task
def task_1(some_token, another_arg):
    client = my_client(some_token)
    client.do_something(another_arg)

@task
def task_2(some_token, more_args):
    client = my_client(some_token)
    client.do_something_else(more_args)
```

In the example above, if `task_1` and `task_2` run in the same process, `my_client` will only execute once, and the cached client will be shared between tasks. If they run in different processes, the client will be recreated once per process. The duplication at the call-site is a bit unfortunate, but makes the creation point explicit which leads to more predictable behavior.


jcrist:
Status:

Things currently work, and I'm happy(ish) with the implementation. Still needs tests and docs.

I'd also like to potentially find a better name. I started with `cached` but that conflicted with other uses of the word `cache` in prefect. `memoize` is used, since it mirrors the `toolz.memoize` decorator, but not all users will be familiar with that word.

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @functools.wraps(func)
+    def inner(*args: Any, **kwargs: Any) -> Any:
+        assert key is not None  # appease mypy
+        k = key(*args, **kwargs)
+        cache = get_local_cache()
+        try:
+            return cache[func_id][k]
+        except TypeError:
+            raise TypeError("Argumentws to a cached function must be hashable")
```
joshmeek, line 114:
```suggestion
            raise TypeError("Arguments to a cached function must be hashable")
```

cicdw:
Implementation seems good to me - still thinking about the name; some other names that crossed my mind were:
- `process_cache`
- `shared_object`

jlowin:
Riffing off @cicdw...

- `@shared_object`
- `@shared`
- `@share`
- `@reuse`
- `@provider`
- `@resource_provider` (use in conjunction with resource manager?)
- `@sidekick`

jcrist:
Ok, I believe this should be ready for review. I renamed it to `shared_object` (still not 100% enthused with that name, but it seemed the best of the options suggested). Something like `cached_per_process` would be more descriptive, but I'm not sure if we want to use the word `cache` since it already has other meanings in prefect. If people have suggestions for a better name I'd love to hear them. Same with the module name (since it shows up in the docs).

A few todos:
- changelog (waiting on a +1/-1 on the final name for the thing)
- should this be a top-level import? (I'm 50/50 on it, I don't think it'll get much use if it's hidden in utilities, but it'll also clutter the top-level).
- add an idiom on using it

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+__all__ = ("shared_object",)
+
+
+class _SerializableLock(object):
+    _locks = weakref.WeakValueDictionary()  # type: weakref.WeakValueDictionary
+    """A Serializable per-process Lock.
+
+    This wraps a normal ``threading.Lock`` or ``threading.RLock`` object and
```
cicdw, line 23:
```suggestion
    This wraps a normal `threading.Lock` or `threading.RLock` object and
```

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,195 @@
...
+def _default_key(*args: Any, **kwargs: Any) -> Any:
+    return (args, frozenset(kwargs.items()) if kwargs else None)
+
+
+@curry
+def shared_object(func: Callable, *, key: Callable = None) -> Callable:
+    """Marks the result of this helper function as a shared object to be reused
+    by other tasks.
+
+    Functions decorated with ``shared_object`` have their results stored in a
```
cicdw, line 112:
```suggestion
    Functions decorated with `shared_object` have their results stored in a
```

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,195 @@
...
+    temporary cache specific to a flow run & process. If the function is called
+    with the same parameters during the same flow run in the same process, it
+    will not be re-evaluated. Upon completion of the flow run, the cache will
+    be cleared.
+
+    Note that ``shared_object`` functions should never be tasks themselves, but
```
cicdw, line 118:
```suggestion
    Note that `shared_object` functions should never be tasks themselves, but
```

cicdw (changes requested):
Minor RST residuals in the docstrings, otherwise this looks good to me!

jcrist:
Any thoughts on the name/location (import path, and should it be imported top-level)? I'm still not 100% sold on `shared_object` being the best name.

cicdw:
I think the location is probably fine for now - we can always move it to the top level if it becomes popular; I agree the name still doesn't feel quite right to me.  The only other name I can think of that seems like a natural fit for our current vocabulary would be `task_resource`?

anishchhaparwal:
hey any updates on this?

src/prefect/engine/executors/dask.py
```diff
@@ -355,6 +358,14 @@ def _post_start_yield(self) -> None:
                 self._should_run_var.delete()
             except Exception:
                 pass
+
+        # Execute on_cleanup hook if present
+        if on_cleanup is not None:
+            try:
+                self.client.run(on_cleanup)  # type: ignore
+            except Exception:
+                pass
```
madkinsz, line 367: Just out of curiosity, why do we pass so many exceptions here without logging?

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,195 @@
...
+    for creating and reusing resources per-worker-process while avoiding
+    treating these clients as the results of tasks (which must be picklable
+    when running with multiple processes).
+
+    Args:
+        - func (Callable): the function to cache. May either `return` a result,
+            or `yield` one. If `yield` is used, the code block after the
+            `yield` statement will be run as a cleanup step upon completion of
+            the flow run (only one `yield` statement is permitted).
+        - key (Callable, optional): a function to use to generate the cache
```
madkinsz, line 129: Might make more sense as `key_fn`, `keygen`, or `key_generator` ? As named the purpose of this variable and the `_default_key` function was initially confusing.

src/prefect/utilities/local_cache.py
```diff
@@ -0,0 +1,195 @@
...
+        clear_local_cache(flow_run_id)
+
+
+def _init_cache() -> dict:
+    out = defaultdict(dict)  # type: dict
+    out["generators"] = []
+    return out
+
+
+_CACHES = defaultdict(_init_cache)  # type: dict
```
madkinsz, line 76: Whilst I can eventually figure out what is going to be stored inside `_CACHES` it takes looking at the code in a few different places. A brief comment here explaining the structure of this dict would be really helpful!

madkinsz:
This is really cool!

> I think the location is probably fine for now - we can always move it to the top level if it becomes popular; I agree the name still doesn't feel quite right to me. The only other name I can think of that seems like a natural fit for our current vocabulary would be `task_resource`?

I like `task_resource` more than `shared_object`, `flow_resource` could make sense to since it is intended to be used per flow run. I think it'd be nice to include that is is only shared by a single process but I think that's not as important to expose to the user and the key is that we are creating a way to define a resource such that we can perform optimizations for the user and if they care about the implementation of those optimizations they'll read the docstring which is pretty clear.

Maybe?
`shared_flow_resource`
`shared_task_resource`

gryBox:
Super cool.  Very useful. `flow_resource` seems like a self explanatory name.

xuevin:
Bump. Any update on this?

Tostino:
Agreed on `flow_resource` being a good name for this.  Being a flow resource rather than a task resource (since a flow is made up of tasks) conveys that that resource should be available within the flow to the tasks that need it.

Also interested in this feature.

madkinsz:
For reference for others, this feature is not complete as written and is stale now. This is still something we're interested in helping provide, but it'll need to be reimplemented and we're continuing to think about the proper design. Thanks for your comments!

PR #3147, opened 2020-08-12
Add command for creating a tenant with server

joshmeek:

## What does this PR change?
Adds a CLI command `prefect server create-tenant` which allows you to create a tenant on the server

## Why is this PR important?
Some users were expecting this command and having to route around by using the Client function is annoying


mr-tester-william-lo:
In attempting to set up Prefect 0.14.15 for Flow execution across multiple Agents using the Docker Compose file produced by prefect server config, I was surprised to discover that that Prefect's Server requires a Tenant in order for an Agent to connect. I investigated to find that Prefect's docs provide info on the usage syntax of their create-tenant and list-tenants CLI operations. However, I've looked extensively through the linked pages — along with the rest of Prefect's documentation and accompanying tutorials — for any mention of Tenants, and only found the following brief mention:

Once authenticated, user-based API keys can be used to generate short-lived JWT auth tokens for any tenant the user belongs to. These API keys inherit any permissions the user has in that tenant, allowing full API access.

What, exactly, is a Tenant in Prefect? It doesn't appear to be specific to Prefect's Hybrid Execution model. How, then, does it fit into the interactions between the Server, Agents, and Client before and during the execution of a Prefect Flow? What does it mean to say that a user "belongs to" or is "in" a tenant? And what is accomplished by requiring it to register an Agent with the Server?

(Same question would like to ask, question source: https://stackoverflow.com/questions/67062496/prefect-server-what-is-the-role-of-a-tenant)

madkinsz:
Hi @mr-tester-william-lo, I've replied there.

PR #3158, opened 2020-08-13
Kubernetes agent job management

### LINKED ISSUE #2943: Kubernetes agent stuck in processing feeback loop after ImagePullBackOff
## Description
When a prefect Kubernetes agent consumes flow with an invalid docker image url:  
```
flow.storage = S3(bucket="bucket", secrets=["AWS_CREDENTIALS"])
flow.environment = LocalEnvironment(
    metadata={
        "image": "wrong_registry_url/image:tag"
    },
    labels=['s3-flow-storage']
)
```
The k8s job goes into a `ImagePullBackOff` state. This is expected.
image

Manually closing the job via kubectl/k9s puts the prefect agent in a feedback loop where no new flow-runs are being processed.
So far the agent was only freed up after deleting the entire cloud project, indicating a run state invalidation issue.
(alternatively changing the state from the cloud UI should be explored)

image

## Expected Behavior
submit flow to run -> k8s agent consumes -> create job -> `ImagePullBackOff` -> fail run 

## Reproduction
s3_k8s_flow.py 
``` 
import prefect
from prefect.environments import LocalEnvironment
from prefect.environments.storage import S3

@task
def hello_task():
    logger.info("Hello, Docker!")


flow = Flow("k8s_agent_S3_loop", tasks=[hello_task])

flow.storage = S3(bucket="bucket-path", secrets=["AWS_CREDENTIALS"])
flow.environment = LocalEnvironment(
    metadata={
        "image": "<wrong_registry>/prefect:latest"
    },
    labels=['s3-flow-storage']
)

flow.register('Bug')


flow_correct = Flow("k8s_agent_S3_correct", tasks=[hello_task])

flow_correct.storage = S3(bucket="bucket-path", secrets=["AWS_CREDENTIALS"])
flow_correct.environment = LocalEnvironment(
    metadata={
        "image": "prefecthq/prefect:latest"
    },
    labels=['s3-flow-storage']
)

flow_correct.register('Bug')
```

create and schedule:
```
python s3_k8s_flow.py
prefect run cloud --name "k8s_agent_S3_correct" --project "Bug"
prefect run cloud --name "k8s_agent_S3_loop" --project "Bug"

prefect get flow-runs --project Bug
```
image

## Environment
prefect version       => 0.12.3
kubernetes agent:  
```
prefect agent install kubernetes -t TOKEN \
  --resource-manager --rbac \
  --image-pull-secrets \
 | kubectl apply -f -
```

### LINKED ISSUE #3141: Fail job if image can't be pulled (k8 agent)
## Current behavior

If the k8 agent tries to submit a job where the image is not accessible it the job gets "ImagePullBackOff". However the Job keeps being marked as "Submitted" in Prefect cloud.

## Proposed behavior
Agent detects that the Job newer started en reports the error.

## Example
This would enable the user to easier detect that the error is related to image pull.

joshmeek:

## What does this PR change?
This PR introduces a new set of functionality into the Kubernetes agent. Now the agent's heartbeat thread is responsible for looking at the available prefect jobs in the namespace (batched of course) and managing their lifecycle. This means that when things like `ErrImagePull` or `ImagePullBackoff` occurs the agent can fully fail the flow run with information about this error. This also takes care of the case of agent restarts. Another thing this unlocks is merging in functionality from the resource-manager where completed jobs will now be cleaned up automatically. This means that the resource manager will be deprecated and eventually removed.

I have looked into providing some functionality like this to other agents such as Fargate but I need to do more work with boto3 to get it working and this PR is already getting large. (will open an issue to track the addition of this functionality to other agents)

e.g.
image

## Why is this PR important?
This is a step in providing users with a window into the status of flow run deployments inside their infrastructure without having to actually go to the infrastructure themselves.

Closes #2943 
Closes #3141


src/prefect/agent/kubernetes/agent.py
```diff
@@ -112,20 +244,42 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
             "Creating namespaced job {}".format(job_spec["metadata"]["name"])
         )
         job = self.batch_client.create_namespaced_job(
-            namespace=self.namespace or os.getenv("NAMESPACE", "default"), body=job_spec
+            namespace=self.namespace, body=job_spec
         )
 
         self.logger.debug("Job {} created".format(job.metadata.name))
 
+        pods = self.core_client.list_namespaced_pod(
```
jcrist, line 261: I don't believe you can rely on any/all pods being created yet, pods managed by a job may be created later (due to processing time or a `ResourceQuota`).

src/prefect/agent/kubernetes/agent.py
```diff
@@ -93,9 +95,139 @@ def __init__(
...
+                    self.logger.debug("List jobs continue token expired, relisting")
+                    _continue = ""
+                    continue
+
+        self.logger.debug(
+            f"Found {len(jobs_list)} existing prefect jobs in the cluster"
+        )
+        return jobs_list
+
+    def heartbeat(self) -> None:
```
jcrist, line 164: This loop makes quite a few calls to the k8s api, and only does `get` operations based on existing job/pod records stored on the agent. This has a few consequences:

- Additional pods started by a job post submit (race condition noted below) won't be caught or checked
- I don't believe this handles the job-starts-a-job (as done currently with the `KubernetesJobEnvironment`) since the subjob won't be stored in the `jobs` mapping.

A different approach would be to drop all local state and either:
- do a periodic full relist of all jobs in the namespace (with appropriate label selectors)
- watch the namespace for job/pod events (again, with appropriate label selectors). This is common for managing infrastructure (it's what we do for dask-gateway and jupyterhub), which would reduce api calls to the k8s api server at the cost of increased local complexity. If `heartbeat` runs on the minute -> minutes scale this complexity likely isn't needed yet.

Pseudocode with a loop:

```python
def heartbeat(self):
    for job in self.generator_of_jobs_in_namespace_matching_label_selector():
        if self.is_done(job) or self.has_pods_stuck_in_bad_state(job):
            self.delete_job(job)
```

Either way, I think we should try to avoid holding onto local state for what jobs/pods to manage and rely on k8s api patterns to handle this instead.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -93,9 +95,139 @@ def __init__(
...
+                delete_job = True
+
+            if not delete_job and not job_status.failed and not job_status.succeeded:
+                for pod_name in job["pod_names"]:
+                    try:
+                        pod_status = self.core_client.read_namespaced_pod_status(
+                            namespace=self.namespace, name=pod_name,
+                        ).status
+                    except self.k8s_client.rest.ApiException:
+                        delete_job = True
```
jcrist, line 190: There's a few reasons we might get an api exception here (unknown pod id, auth issue, network error, ...), I think we'll want to be more specific when determining if that failure should cause the backing job to be deleted.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -93,9 +95,139 @@ def __init__(
...
+        self.logger.debug(f"Reading statuses for {len(self.jobs)} jobs...")
+        for job in copy.deepcopy(self.jobs):
+            delete_job = False
+
+            try:
+                job_status = self.batch_client.read_namespaced_job_status(
+                    namespace=self.namespace, name=job["job_name"],
+                ).status
+            except self.k8s_client.rest.ApiException:
+                delete_job = True
```
jcrist, line 181: Same here

src/prefect/agent/kubernetes/agent.py
joshmeek: Interesting, so the heartbeat here won't run on the minute -> minutes scale and instead looks to run on the same iteration cycle as the normal querying for flow runs (10s). I haven't used watch events in the k8s python client before so I'll have to look into how to perform it

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,156 @@ def __init__(
...
+        for job in jobs:
+            delete_job = False
+
+            try:
+                job_status = self.batch_client.read_namespaced_job_status(
+                    namespace=self.namespace, name=job["job_name"],
+                ).status
+            except self.k8s_client.rest.ApiException as exc:
+                if exc.status != 401:
+                    delete_job = True
```
TylerWanner, line 190: i would guess that most failures that are not 401 are network noise.  I would think in this case we should continue without deleting, assuming that it will be picked up successfully next time around

joshmeek: Ah yes we should have a `continue` here, good catch

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,153 @@ def __init__(
...
+                    self.logger.debug("Refreshing job listing token...")
+                    _continue = ""
+                    continue
+
+        self.logger.debug(
+            f"Found {len(jobs_list)} existing prefect jobs in the cluster"
+        )
+        return jobs_list
+
+    def _read_prefect_jobs(self, jobs: list) -> None:
```
jcrist, line 173: This method doesn't really "read prefect jobs" (since it might delete finished/failed jobs, and also notifies the cloud api). Could you rename to a more descriptive name?

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +105,177 @@ def __init__(
...
+
+        Args:
+            jobs: A list of job identifiers from Kubernetes
+        """
+        self.logger.debug(f"Reading statuses for {len(jobs)} jobs...")
+        for job in jobs:
+            delete_job = False
+
+            try:
+                job_status = self.batch_client.read_namespaced_job_status(
```
jcrist, line 187: The job status should already be returned as part of the previous `list_namespaced_job` call above, we shouldn't need to make another call to the k8s api here if we can avoid it.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +105,177 @@ def __init__(
...
+            except self.k8s_client.rest.ApiException as exc:
+                self.logger.error(
+                    f"{exc.status} error attempting to read status of job {job['job_name']}"
+                )
+                continue
+
+            if not delete_job and not job_status.failed and not job_status.succeeded:
+                for pod_name in job["pod_names"]:
+                    try:
+                        pod_status = self.core_client.read_namespaced_pod_status(
```
jcrist, line 200: Same here, we should already have the pod status as part of `list_namespaced_pod` above.

src/prefect/agent/kubernetes/agent.py
jcrist: Since one method is listing jobs, and the other is re-iterating through those jobs to respond, perhaps this should really be one method, and jobs are handled individually as they're found in the pagination loop? The separation here feels a bit off.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +105,177 @@ def __init__(
...
+                self.logger.debug(f"Deleting job {job['job_name']}")
+                try:
+                    self.batch_client.delete_namespaced_job(
+                        name=job["job_name"],
+                        namespace=self.namespace,
+                        body=self.k8s_client.V1DeleteOptions(
+                            propagation_policy="Foreground"
+                        ),
+                    )
+                except self.k8s_client.rest.ApiException as exc:
```
jcrist, line 265: Perhaps 404s should be exempt from logging, as they'd indicate the job was already deleted.

src/prefect/agent/kubernetes/agent.py
jcrist: Thinking more (not to tell you how to structure your code). Perhaps this would be cleaner as:

- One method that yields tuples of job + pods, performing the pagination as it goes (a generator method). This would prevent loading too many jobs/pods into memory at once.
- One method for handling a single job + pods tuple, doing whatever actions are required.
- One method for doing the whole loop over jobs/pods + processing

Seems simpler to structure and test, as the generator method could be mocked out, and the actions we take on the retrieved info are the tricky bits.

src/prefect/agent/agent.py
```diff
@@ -299,6 +299,19 @@ def stop_server() -> None:
...
+            while True:
+                self.logger.debug("Running agent heartbeat...")
+                self.heartbeat()
+                self.logger.debug("Sleeping heartbeat for 60 seconds")
+                time.sleep(60)
+
+        self._heartbeat_thread = threading.Thread(
+            name="heartbeat", target=run, daemon=True
+        )
+        self._heartbeat_thread.start()
```
jcrist, line 313: Can you add a call to stop the heartbeat thread in `cleanup` above? The code used for the `_api_server_thread` should work identically.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,99 @@ def __init__(
...
+                                                message="Kubernetes Error: {}".format(
+                                                    container_status.state.waiting.message
+                                                )
+                                            ),
+                                        )
+
+                                        delete_job = True
+                                        break
+
+                    if delete_job or job.status.failed or job.status.succeeded:
```
jcrist, line 171: I find the following logic slightly easier to follow:

```suggestion
                    delete_job = job.status.failed or job.status.succeeded
                    job_name = job.metadata.name
                    flow_run_id = job.metadata.labels.get("prefect.io/flow_run_id")

                    if not delete_job:
                        pods = self.core_client.list_namespaced_pod(
                            namespace=self.namespace,
                            label_selector="prefect.io/identifier={}".format(
                                job.metadata.labels.get("prefect.io/identifier")
                            ),
                        )

                        for pod in pods.items:
                            if pod.status.container_statuses:
                                for container_status in pod.status.container_statuses:
                                    waiting = container_status.state.waiting
                                    if waiting and (
                                        waiting.reason == "ErrImagePull"
                                        or waiting.reason == "ImagePullBackOff"
                                    ):
                                        self.logger.debug(
                                            f"Failing flow run {flow_run_id} due to pod {waiting.reason}"
                                        )
                                        self.client.set_flow_run_state(
                                            flow_run_id=flow_run_id,
                                            state=Failed(
                                                message="Kubernetes Error: {}".format(
                                                    container_status.state.waiting.message
                                                )
                                            ),
                                        )

                                        delete_job = True
                                        break

                    if delete_job:
```

jcrist, line 171: I find the following logic slightly easier to follow:

```suggestion
                    delete_job = job.status.failed or job.status.succeeded
                    job_name = job.metadata.name
                    flow_run_id = job.metadata.labels.get("prefect.io/flow_run_id")

                    if not delete_job:
                        pods = self.core_client.list_namespaced_pod(
                            namespace=self.namespace,
                            label_selector="prefect.io/identifier={}".format(
                                job.metadata.labels.get("prefect.io/identifier")
                            ),
                        )

                        for pod in pods.items:
                            if pod.status.container_statuses:
                                for container_status in pod.status.container_statuses:
                                    waiting = container_status.state.waiting
                                    if waiting and (
                                        waiting.reason == "ErrImagePull"
                                        or waiting.reason == "ImagePullBackOff"
                                    ):
                                        self.logger.debug(
                                            f"Failing flow run {flow_run_id} due to pod {waiting.reason}"
                                        )
                                        self.client.set_flow_run_state(
                                            flow_run_id=flow_run_id,
                                            state=Failed(
                                                message="Kubernetes Error: {}".format(
                                                    container_status.state.waiting.message
                                                )
                                            ),
                                        )

                                        delete_job = True
                                        break

                    if delete_job:
```

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,99 @@ def __init__(
...
+                    namespace=self.namespace,
+                    label_selector="prefect.io/identifier",
+                    limit=20,
+                    _continue=_continue,
+                )
+                _continue = jobs.metadata._continue
+                more = bool(_continue)
+
+                for job in jobs.items:
+                    delete_job = False
```
jcrist, line 136: I found the following logic slightly easier to understand (w/ other suggestions below):

```suggestion
                    delete_job = job.status.failed or job.status.succeeded
```

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,99 @@ def __init__(
...
+                    job_name = job.metadata.name
+                    flow_run_id = job.metadata.labels.get("prefect.io/flow_run_id")
+
+                    if not job.status.failed and not job.status.succeeded:
```
jcrist, line 140:
```suggestion
                    if not delete_job:
```

src/prefect/agent/kubernetes/agent.py
jcrist, line 171:
```suggestion
                    if delete_job:
```

src/prefect/agent/kubernetes/agent.py
```diff
@@ -103,9 +104,99 @@ def __init__(
...
+                                name=job_name,
+                                namespace=self.namespace,
+                                body=self.k8s_client.V1DeleteOptions(
+                                    propagation_policy="Foreground"
+                                ),
+                            )
+                        except self.k8s_client.rest.ApiException as exc:
+                            self.logger.error(
+                                f"{exc.status} error attempting to delete job {job_name}"
+                            )
```
jcrist, line 184: 404 errors indicate the job was already deleted, no need to log those.

```suggestion
                            if exc.status != 404:
                                self.logger.error(
                                    f"{exc.status} error attempting to delete job {job_name}"
                                )
```

jcrist (dismissed):

PR #3161, opened 2020-08-14
Make kubernetes extra logger in DaskKubernetesEnvironment optional

### LINKED ISSUE #2988: Make attachment of kubernetes logger to DaskKubernetes env opt-in
## Current behavior

The DaskKubernetes environment automatically attaches Kubernetes loggers to Prefect Cloud logs.  This is a great feature for observability, but will inevitably wind up accidentally shipping very sensitive information from users to Prefect Cloud logs unintentionally.

## Proposed behavior

Perhaps there is a middle ground where, rather than having to explicitly declare the kubernetes logger, a config flag can be set to add kubernetes loggers, but it should default to False.  There may be some additional notes to add to the docs around this so that users know to turn this on to help with debugging.

joshmeek:

## What does this PR change?
Closes #2988 
Adds a new kwarg to the DaskKubernetesEnvironment `log_k8s_errors` which can be used to toggle the use of the `kubernetes` extra logger.

## Why is this PR important?
Could lead to logging unintended information so we should make this opt-in

### DIFF
--- a/changes/issue2988.yaml
+++ b/changes/issue2988.yaml
@@ -1,2 +1,5 @@
 enhancement:
   - "Make use of `kubernetes` extra logger in the `DaskKubernetesEnvironment` optional - [#2988](https://github.com/PrefectHQ/prefect/issues/2988)"
+
+breaking:
+  - "DaskKubernetesEnvironment no longer logs Kubernetes errors by default - [#2988](https://github.com/PrefectHQ/prefect/issues/2988)"
--- a/changes/issue2988.yaml
+++ b/changes/issue2988.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Make use of `kubernetes` extra logger in the `DaskKubernetesEnvironment` optional - [#2988](https://github.com/PrefectHQ/prefect/issues/2988)"
--- a/src/prefect/environments/execution/dask/k8s.py
+++ b/src/prefect/environments/execution/dask/k8s.py
@@ -328,7 +328,7 @@ class DaskKubernetesEnvironment(Environment):
             "distributed.deploy.adaptive",
         ]
 
-        if self.log_k8s_erros:
+        if self.log_k8s_errors:
             cluster_loggers.append("kubernetes")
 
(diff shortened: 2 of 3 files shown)


src/prefect/environments/execution/dask/k8s.py
```diff
@@ -322,8 +326,11 @@ def _extra_loggers(self) -> str:
         cluster_loggers = [
             "dask_kubernetes.core",
             "distributed.deploy.adaptive",
-            "kubernetes",
         ]
+
+        if self.log_k8s_erros:
```
cicdw, line 331: I'm super impressed that you managed to perfectly replicate this typo in exactly two places so that it's still functional and passes tests
```suggestion
        if self.log_k8s_errors:
```

src/prefect/environments/execution/dask/k8s.py
```diff
@@ -116,6 +119,7 @@ def __init__(
         self._scheduler_spec, self._worker_spec = self._load_specs_from_file()
 
         self._identifier_label = ""
+        self.log_k8s_erros = log_k8s_errors
```
cicdw, line 122:
```suggestion
        self.log_k8s_errors = log_k8s_errors
```

changes/issue2988.yaml
```diff
@@ -0,0 +1,2 @@
+enhancement:
+  - "Make use of `kubernetes` extra logger in the `DaskKubernetesEnvironment` optional - [#2988](https://github.com/PrefectHQ/prefect/issues/2988)"
```
cicdw, line 2: On the fence about whether to include this as both an enhancement _and_ a breaking change in case anyone relied on these logs?

src/prefect/environments/execution/dask/k8s.py
joshmeek: 🤦

changes/issue2988.yaml
joshmeek: Yeah it could go there since it is a slight change in that regard, will add it

PR #3166, opened 2020-08-16
Port over DatabricksSubmitRun Operator & Framework from Airflow

Ashton-Sidhu:

## What does this PR change?

This PR ports over the DatabricksSubmitRun Operator from airflow to run automated jobs on Databricks over to Prefect. It also provides the framework to do other Databricks tasks using the Databricks API.

It also addresses the comments in the main comment of this [PR](https://github.com/PrefectHQ/prefect/pull/3084). It ports over the Databricks Hook and the Operator over and separates the Task and the Databricks execution.

## Why is this PR important?

This PR is important as it adds another critical ETL technology platform to the Prefect stack further increasing its coverage to airflow.


Ashton-Sidhu:
Hey @d80tb7 - I ported over the Databricks Operator from Airflow as you suggested #3084 and was wondering if you have any thoughts on this.

d80tb7:
Hi @Ashton-Sidhu thanks for doing this- to me this looks really good.

One thought I did have was whether we should add the parameters you currently have on `DatabricksSubmitRun.__init__()` to `DatabricksSubmitRun.run()`- that way you could dynamically set, for example, the number of executors based on prefect parameters rather than having to define it statically at run time.

@joshmeek what do you think?

joshmeek:
@d80tb7 Oh yes definitely! There's a helper function that you can put on the task's `run` method to automatically grab the `__init__` kwargs if they aren't provided again. Could use a task like this as a guide:

https://github.com/PrefectHQ/prefect/blob/8e7b0488c7949d87d15884e76439366608306541/src/prefect/tasks/kubernetes/job.py#L63-L65

Also, this isn't something that you necessarily _need_ to do but this task currently takes in the name of a prefect secret and calls:

```python
self.databricks_conn_id = Secret(databricks_conn_secret).get()
```

Instead of having to do that it could be refactored to take in the actual secret value retrieved from a `PrefectSecret` task. This type of task is a secure task which safely retrieves the value and then it can be used to directly pass into other tasks without persisting/logging/etc. The use of it would end up looking something like this:

```python
from prefect.tasks.secrets import PrefectSecret
from prefect.contrib.tasks.databricks import DatabricksSubmitRun

with Flow('my flow') as flow:
    conn = PrefectSecret('DATABRICKS_CONNECTION_STRING')
    DatabricksSubmitRun(databricks_conn_string=conn)()
```

Ashton-Sidhu:
> ```python
> rom prefect.tasks.secrets import PrefectSecret
> from prefect.contrib.tasks.databricks import DatabricksSubmitRun
> 
> with Flow('my flow') as flow:
>     conn = PrefectSecret('DATABRICKS_CONNECTION_STRING')
>     DatabricksSubmitRun(databricks_conn_string=conn)()
> ```

Hey @joshmeek thanks for the feedback, in my initial port from Airflow I had added them to the run function but then removed them for more of the Airflow feel. I added them back to the function and ported over the JSON construction to the run file where , to my understanding, the following occurs:

  - If the user passes the args to the task class and nothing to the run function: it uses the arg values passed to the constructor ( which are initialized in the `__init__` function ) because the initialized values get passed to the `run` method anyways due to the `default_from_attrs` decorator.

  - If the user passed the args to both the task class and the `run` method, any args that are the same would be overwritten by the args passed to the `run` method and then a combination of both args passed to the `run` method and task class are used.

  - If the user passed nothing to the task class, they must supply the necessary args to the `method`.

Also changed `databricks_conn_secret` to `dict` to promote the use of the more secure method of secret handling.

Looking forward to hearing your thoughts

joshmeek:
@Ashton-Sidhu that all sounds good and the PR looks great! Will merge in if everything also looks good to @d80tb7.

d80tb7 (approved):
Thanks for doing this- this looks great to me.

PR #3167, opened 2020-08-16
Remove deprecated client concurrency methods

jlowin:

## What does this PR change?
The Core client refers to `task_tag_...` routes that have been deprecated for some time in Cloud and don't exist in Server. The UI is the preferred way to set these limits.


jlowin:
Ah, I just realized we didn't mark these as deprecated in Core, which was an oversight, although they've been deprecated for close to 6 months. Perhaps I shouldn't remove them wholesale quite yet.

docs/orchestration/concepts/task-concurrency-limiting.md
```diff
@@ -39,7 +39,7 @@ Once you have tagged your various tasks and [registered your Flow(s)](flows.html
 
 ### UI
 
-To set Task tag concurrency limits from the UI, go to Prefect Cloud and navigate to Team Settings -> Task Concurrency Limits.
+To set task concurrency limits from the UI, go to Prefect Cloud and navigate to Team Settings -> Task Concurrency Limits.
```
cicdw, line 42: I don't see the benefit in removing the 'tag' descriptor here and elsewhere - the limits are based on tags

src/prefect/client/client.py
```diff
@@ -1351,82 +1351,6 @@ def set_secret(self, name: str, value: Any) -> None:
         if not result.data.set_secret.success:
             raise ValueError("Setting secret failed.")
 
-    def get_task_tag_limit(self, tag: str) -> Optional[int]:
```
cicdw, line 1354: Why not update the GraphQL API routes that are being called and leave the client methods with the same name / signature?

madkinsz:
Note that this PR is now stale with regards to the docs at https://docs.prefect.io/orchestration/flow-runs/concurrency-limits.html#task-run-limits but the deprecation still needs to be done on the core/client side

PR #3173, opened 2020-08-17
Add support for `host_config` and arbitrary keyword arguments in Docker tasks

sk4la:
## What does this PR change?

A couple of small edits in the `prefect.tasks.docker` package. Basically, I just:

* Added the `extra_docker_kwargs` parameter to every Docker-related `Task` class in order to pass in arbitrary keyword arguments to Docker API calls (as in the current `prefect.tasks.docker.containers.CreateContainer` task). This allows for use of the newer possibilities of the [Docker API client](https://docker-py.readthedocs.io/en/stable/api.html#low-level-api) without having to adapt the corresponding `Task` class.
* Rewrote some of the log messages to make them more accurate and homogeneous towards the rest of the project.
* Fixed an issue in the way the current `extra_docker_kwargs` parameter is handled in `prefect.tasks.docker.containers.CreateContainer`. The current class uses the instance variable (cf. `self.extra_docker_kwargs`) instead of the variable that is passed to the `run` method, therefore it's not possible to override the `extra_docker_kwargs` passed to the original `Task` constructor.
* Removed the implicit `typing.Optional` type hint (found them nowhere in the task library).
* Fixed and issue in the `prefect.tasks.docker.containers.CreateContainer` task class that encouraged passing a `Dict` as the `volumes` parameter even though the [documentation](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_container) says otherwise (my bad, that was my PR).
* Added the `host_config` parameter to the `prefect.tasks.docker.containers.CreateContainer` task class in order to pass in arbitrary keyword arguments to Docker API client's `create_host_configuration` method. Before that, you had to call `create_host_configuration` yourself by connecting to the Docker daemon, which was not always possible (and horribly tedious).

I did not update the tests since these are passthrough parameters.

## Why is this PR important?

I'm using Prefect at my workplace and needed to add some more support for [Docker Volumes](https://docs.docker.com/storage/volumes/). Since this capability might be useful to the community, I thought I'd share!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Released on August 11, 2020.
 ### Task Library
 
 - Return `LoadJob` object in `BigQueryLoad` tasks - [#3086](https://github.com/PrefectHQ/prefect/issues/3086)
+- Add support for [`host_config`](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_host_config) and arbitrary keyword arguments in `Docker` tasks.
 
 ### Fixes
 
@@ -28,6 +29,7 @@ Released on August 11, 2020.
 - [Franklin Winokur](https://github.com/fwinokur)
 - [Fraznist](https://github.com/Fraznist)
 - [Jackson Maxfield Brown](https://github.com/JacksonMaxfield)
+- [Nelson Cornet](https://github.com/sk4la)
 
 ## 0.13.1 <Badge text="beta" type="success" />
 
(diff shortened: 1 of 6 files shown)


sk4la:
Looks like the tests are failing because of some other code sections (unrelated to this PR). Can you give me some insight please?

joshmeek:
@sk4la I can take a look at these tests today, looks like it's due to the logging tests needing to be updated to account for the new debug logging statements.

PR #3179, opened 2020-08-18
Logging updates

cicdw:
## Summary
This PR:
- defaults Cloud logging level to INFO to reduce noise
- updates Cloud logging level to be responsive to user config
- updates some log levels

The ultimate goal is to provide a simpler, more configurable and streamlined logging experience.

Additionally, I added a new env var to `setup.cfg` - I realized when running tests today after working on Server that our tests assume a default backend of Cloud.

## Checklist

This PR:


src/prefect/utilities/logging.py
```diff
@@ -129,6 +129,13 @@ def emit(self, record) -> None:  # type: ignore
             assert isinstance(self.client, Client)  # mypy assert
 
             record_dict = record.__dict__.copy()
+
+            # ensures emitted logs respect configured logging level
+            config_level = getattr(logging, prefect.context.config.logging.level, 20)
```
jcrist, line 134: Could you use the logging enum instead of `20` here?

cicdw: Yup, done!

src/prefect/utilities/logging.py
```diff
@@ -131,7 +131,9 @@ def emit(self, record) -> None:  # type: ignore
             record_dict = record.__dict__.copy()
 
             # ensures emitted logs respect configured logging level
-            config_level = getattr(logging, prefect.context.config.logging.level, 20)
+            config_level = getattr(
+                logging, prefect.context.config.logging.level, "INFO"
```
jcrist, line 135: Sorry, I meant `logging.INFO` instead of `20`, not `"INFO"`. I'm actually surprised this doesn't error below.

```suggestion
                logging, prefect.context.config.logging.level, logging.INFO
```

Issue #3184, opened 2020-08-19
Allow for mounting volume in k8 job (for secrets)


thomasfrederikhoeck:
## Current behavior

Right now it is not possible to mount volumes to the k8 jobs. I think there are many use cases for this. I have outlined what I believe is an important one in the example.

## Proposed behavior

Allow for adding a volume. An easy way to achieve this could be to allow for specifying a custom job_spec.yaml that KubernetesAgent use as template. This could be done as

```
KubernetesAgent(custom_job_spec = "/folder/my_custom_jobs.yaml) 
```
Another way it could be achieved which might be better (but not so straight forward)  would be to allow for specifying a `dict ` that would be merged with the resulting dict in  `KubernetesAgent.replace_job_spec_yaml`

## Example

Right now secrets are either given through environmental variables  (or Prefect Cloud Secrets ). A very common way of getting secrets in to Kubernetes is through mounting (see https://github.com/kubernetes-sigs/secrets-store-csi-driver). This is also the recommend way for linking Azure AKS and Azure Key Vault (https://docs.microsoft.com/en-us/azure/key-vault/general/key-vault-integrate-kubernetes). This also makes it easier to prevent leaking credential to logs as cloud logs can be allowed without writing env vars to logs.

thomasfrederikhoeck:
A resulting yaml would be:

```yaml

apiVersion: batch/v1
kind: Job
metadata:
  name: prefect-job
  labels: {}
spec:
  template:
    metadata:
      labels: {}
    spec:
      containers:
        - name: flow
          image: prefecthq/prefect:latest
          imagePullPolicy: IfNotPresent
          command: ["/bin/sh", "-c"]
          args: ["prefect execute flow-run"]
          env:
            - name: PREFECT__CLOUD__API
              value: PREFECT__CLOUD__API
            - name: PREFECT__CLOUD__AUTH_TOKEN
              value: PREFECT__CLOUD__AUTH_TOKEN
            - name: PREFECT__CONTEXT__FLOW_RUN_ID
              value: PREFECT__CONTEXT__FLOW_RUN_ID
            - name: PREFECT__CONTEXT__FLOW_ID
              value: PREFECT__CONTEXT__FLOW_ID
            - name: PREFECT__CONTEXT__NAMESPACE
              value: PREFECT__CONTEXT__NAMESPACE
            - name: PREFECT__CLOUD__AGENT__LABELS
              value: PREFECT__CLOUD__AGENT__LABELS
[...]

PR #3210, opened 2020-08-24
Added option to mount volumes with KubernetesAgent

thomasfrederikhoeck:
## Summary
This PR adds the option to give let KubernetesAgent mount volumes in the k8 jobs.

## Changes
Adds the two options `volume_mounts `and `volumes` to `KubernetesAgent  `

## Importance
Allows for mount volume which has many use-cases. One is #3184 but this PR allows for much broader use.

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -1,7 +1,7 @@
 import os
 import uuid
 from os import path
-from typing import Iterable
+from typing import Iterable, List
 
 import yaml
 
@@ -57,6 +57,12 @@ class KubernetesAgent(Agent):
             (default).
         - no_cloud_logs (bool, optional): Disable logging to a Prefect backend for this agent
             and all deployed flow runs
+        - volume_mounts (list, optional): A list of volumeMounts to mount when a job is
+            run. The volumeMounts in the list should be specified as dicts
+            i.e `[{"name": "my-vol", "mountPath": "/mnt/my-mount"}]`
+        - volumes (list, optional): A list of volumes to make available to be mounted when a
+            job is run. The volumes in the list should be specified as nested dicts.
+            i.e `[{"name": "my-vol", "csi": {"driver": "secrets-store.csi.k8s.io"}}]`
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


joshmeek:
Thanks for the PR @thomasfrederikhoeck! We should also add this option to the agent `start` and `install` CLI commands https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/agent.py#L213. I can add them if you want but I might not get to it for a bit 🙂

src/prefect/agent/kubernetes/agent.py
```diff
@@ -216,6 +226,12 @@ def replace_job_spec_yaml(self, flow_run: GraphQLResult, image: str) -> dict:
             resources["requests"]["cpu"] = os.getenv("JOB_CPU_REQUEST")
         if os.getenv("JOB_CPU_LIMIT"):
             resources["limits"]["cpu"] = os.getenv("JOB_CPU_LIMIT")
+        if self.volume_mounts:
+            job["spec"]["template"]["spec"]["containers"][0][
+                "volumeMounts"
+            ] = self.volume_mounts
+        if self.volumes:
+            job["spec"]["template"]["spec"]["volumes"] = self.volumes
```
joshmeek, line 234: We should probably add something similar to the imagePullSecrets where we call 
```python
del job["spec"]["template"]["spec"]["imagePullSecrets"]
```
to remove them from the spec if they're empty. Not sure if having empty volume mounts would cause errors on some k8s providers

thomasfrederikhoeck: True, that is probably  the safest.

thomasfrederikhoeck:
With regards to the CLI - would this be necessary for approving the PR, or would opening an issue suffice? And what would be the preferred method of passing it in? As json maybe?

joshmeek:
@thomasfrederikhoeck Not fully necessary, we can open an issue for it to do after and yeah it would probably be a string representation of JSON that could then be parsed

thomasfrederikhoeck:
@joshmeek Cheers for the approve. I will open an issue for the CLI.

PR #3215, opened 2020-08-24
Dockerfile Refactoring and Improvements

gabrielcalderon:
## Summary
* Reduce overall size of final docker image by using layer less layers.
* Added default UTF-8 locale to system.
* Avoid having pip cache by using --no-install-cache
* Added comments to Dockerfile
* Shouldn't affect CI process since build arguments were left the same

## Changes
* Overall image size of latest tag uncompressed was reduced from **425 MB** to **326 MB** a decrease of **23%**
* Added default locale UTF-8
* Remove **git** after installing Prefect.
* Reduce RUN layer to just 1 from 4.

## Importance
* Reduce network usage when downloading image.
* Reduce storage usage when using Prefect Docker Image.

## Checklist

This PR:

- [-] adds new tests (if appropriate)
- [-] adds a change file in the `changes/` directory (if appropriate)
- [-] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,17 @@
 ARG PYTHON_VERSION=${PYTHON_VERSION}
 FROM python:${PYTHON_VERSION}-slim
 
-RUN apt update && apt install -y gcc git && rm -rf /var/lib/apt/lists/*
-
+# Build Arguments
 ARG PREFECT_VERSION
 ARG EXTRAS=kubernetes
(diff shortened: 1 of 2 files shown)


gabrielcalderon:
@joshmeek No problem! I'm also working on a similar PR for prefecthq/server.

PR #3221, opened 2020-08-25
Revert "Color pytest ouptut logs"

joshmeek:
I'm not too familiar with `pytest.ini` but the presence of this file causes a lot of my tests to fail because they are no longer independent of my local environment cc @jcrist @cicdw

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,0 @@
-[pytest]
-addopts = --color=yes


jcrist:
Oh, yeah, the presence of a `pytest.ini` file negates all other config (we have config set in `setup.cfg` as well). You should be able to move the line from this over to the `setup.cfg` file (under the `tool:pytest` section) and everything should work then, but I also am fine just removing this. Color config enabled seems like an odd thing to bake into our static configs.

PR #3230, opened 2020-08-28
Optimizing constants out of graph still add downstream

shaunc:
Although an edge can be omitted from the graph if source task is a constant, the target task should still be included in the graph. As it is a single edge graph from constant to task will contain no tasks if "add_edge()" is used to build it, but it should contain at least the target task.

## Checklist

This PR:

### DIFF
--- a/tests/core/test_flow.py
+++ b/tests/core/test_flow.py
@@ -447,6 +447,11 @@ def test_add_edge_returns_edge():
     assert added_edge in f.edges
     assert edge in f.edges
 
+def test_add_edge_from_contant():
+    f = Flow(name="test")
+    c1 = constants.Constant(1)
+    t1 = Task()
+    edge = f.add_edge(upstream_task=c1, downstream_task=t1, key="foo")
 
 def test_chain():
     f = Flow(name="test")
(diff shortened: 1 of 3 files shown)


shaunc:
> Hi @shaunc - thanks for the PR! Would you mind adding a test to `test_flows.py` for this behavior?

How's that?

tests/core/test_flow.py
```diff
@@ -447,6 +448,14 @@ def test_add_edge_returns_edge():
...
     assert edge in f.edges
 
+def test_add_edge_from_contant():
+    f = Flow(name="test")
+    value = 1
+    c1 = constants.Constant(value)
+    t1 = Task()
+    f.add_edge(upstream_task=c1, downstream_task=t1, key="foo")
+    assert t1 in f.get_tasks()
+    assert f.constants[t1]['foo'] == value
```
cicdw, line 458:
```suggestion
    assert f.constants[t1]["foo"] == value
```

in order to pass our formatting checks, then LGTM 👍

cicdw: Alternatively, you can run `black .` on the whole codebase which will autoformat everything for you

PR #3251, opened 2020-09-02
make identifier optional in KubernetesAgent.replace_job_spec_yaml() (fixes #3250)

jameslamb:
## Summary

See #3250 for a summary.

## Changes

Makes `identifier` optional in `KuebrnetesAgent.replace_job_spec_yaml()` to avoid breaking older code that is using that method directly.

## Importance

This PR fixes a small regression introduced in 0.13.5 which could break older code that directly calls `KubernetesAgent.replace_job_spec_yaml()`.

Instead of removing the `identifier` argument in this PR, I opted to give it a default value. If I just removed it, code written in v 0.13.5 which passes `identifier` would break in the next release.

## Checklist

This PR:

### DIFF
--- a/changes/pr3251.yaml
+++ b/changes/pr3251.yaml
@@ -1,4 +1,4 @@
-feature:
+fix:
   - "Make identifier optional in KubernetesAgent.replace_job_spec_yaml() - [#3251](https://github.com/PrefectHQ/prefect/pull/3251)"
 
 contributor:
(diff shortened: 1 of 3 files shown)


changes/pr3251.yaml
```diff
@@ -0,0 +1,5 @@
+feature:
```
joshmeek, line 1:
```suggestion
fix:
```

joshmeek: Small nit but otherwise lgtm!

jameslamb: ah whoops, bad copy-paste from an older PR, sorry!

PR #3266, opened 2020-09-04
Add logger to global context

kingsleyb:
## Summary

Adds logger to global context to remove friction on running task unit tests

## Checklist

This PR:

### DIFF
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -17,6 +17,7 @@ from prefect.tasks.core.resource_manager import resource_manager
 
 from prefect.utilities.tasks import task, tags, apply_map
 from prefect.utilities.edges import mapped, unmapped, flatten
+from prefect.utilities.logging import get_logger
 
 import prefect.serialization
 import prefect.agent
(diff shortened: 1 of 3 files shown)


src/prefect/__init__.py
```diff
@@ -33,3 +34,5 @@
     _signal.signal(29, _sig_handler)
 except:
     pass
+
+context.logger = get_logger()
```
jcrist, line 38: Would you mind moving this to the `prefect.utilities.logging` module? The global logger is already created there, and the context is imported there as well. Makes more sense IMO to keep the logging stuff grouped together.
Could do it here:

https://github.com/PrefectHQ/prefect/blob/936c0e05f704f1fd6fa7da89d94734471418bd3e/src/prefect/utilities/logging.py#L251

```python
context.logger = prefect_logger = configure_logging()
```

jcrist (commented):
Overall this looks good to me, just one small nit.

src/prefect/__init__.py
kingsleyb: Moved the change as suggested.

jcrist (approved):

PR #3267, opened 2020-09-04
Make a regular flow runner heartbeat log a little less verbose

cicdw:
## Summary

This particular log was being generated every 30 seconds or so for flow runs with DEBUG logging level and was crowding up the logs views.  It _is_ useful when the flow run is no longer in a running state, so I've updated the logic accordingly.

## Checklist

This PR:

### DIFF
--- a/src/prefect/engine/cloud/flow_runner.py
+++ b/src/prefect/engine/cloud/flow_runner.py
@@ -182,9 +182,9 @@ class CloudFlowRunner(FlowRunner):
                             "Error getting flow run info", exc_info=True
                         )
                         continue
-                    else:
-                        self.logger.debug(
-                            "Successfully queried server for flow run state: %r",
+                    if not flow_run_info.state.is_running():
+                        self.logger.warning(
+                            "Flow run is no longer in a running state; the current state is: %r",
                             flow_run_info.state,
                         )
                     if isinstance(flow_run_info.state, Cancelling):


src/prefect/engine/cloud/flow_runner.py
```diff
@@ -182,9 +182,9 @@ def interrupt_if_cancelling() -> None:
                             "Error getting flow run info", exc_info=True
                         )
                         continue
-                    else:
-                        self.logger.debug(
-                            "Successfully queried server for flow run state: %r",
+                    if not flow_run_info.state.is_running():
+                        self.logger.warning(
+                            "Flow run is no longer in a running state; the current state is: %r",
```
jcrist, line 187: Is this useful to see? Task runs should eventually bubble back to shutdown the flow runner, so checking for running here isn't necessary (but may still be useful?).

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -182,9 +182,9 @@ def interrupt_if_cancelling() -> None:
                             "Error getting flow run info", exc_info=True
                         )
                         continue
-                    else:
-                        self.logger.debug(
-                            "Successfully queried server for flow run state: %r",
```
jcrist, line 187: Sorry, I added this when debugging the cancellation implementation. Definitely no longer needed.

jcrist (approved):

src/prefect/engine/cloud/flow_runner.py
cicdw: Good question - I think it is; there are two situations I've seen where this log would be nice:
- users whose task runs error out when they attempt to enter a `Running` state because the flow run state is already finished ("cancellation lite"); for some reason many users have a hard time debugging this, and I think seeing this log might help tell the story
- earlier @zhen0 identified a weird race condition where the flow run was running but the backend thought it was in a `Submitted` state, and this log was one way she identified that

src/prefect/engine/cloud/flow_runner.py
cicdw: Ahhhh haha I'll admit I wondered where it came from - I actually do think that "unexpected" states are worth logging so 👍

PR #3269, opened 2020-09-05
Add a new task to create a github issue comment

harupy:
## Summary

Add a new task to create a comment on the specified github issue.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/github/comments.py
+++ b/src/prefect/tasks/github/comments.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, List
+from typing import Any
 
 from prefect import Task
 from prefect.utilities.tasks import defaults_from_attrs
--- a/src/prefect/tasks/github/comments.py
+++ b/src/prefect/tasks/github/comments.py
@@ -20,7 +20,7 @@ class CreateIssueComment(Task):
             init method
 
     References:
-        - [Create an issue
+        - See [Create an issue
           comment](https://developer.github.com/v3/issues/comments/#create-an-issue-comment)
     """
 
--- a/src/prefect/tasks/github/comments.py
+++ b/src/prefect/tasks/github/comments.py
@@ -20,7 +20,7 @@ class CreateIssueComment(Task):
             init method
 
     References:
-        - https://developer.github.com/v3/issues/comments/#create-an-issue-comment
+        - Create an issue comment: https://developer.github.com/v3/issues/comments/#create-an-issue-comment
     """
 
     def __init__(
--- a/src/prefect/tasks/github/comments.py
+++ b/src/prefect/tasks/github/comments.py
@@ -1,0 +1,86 @@
+import json
+from typing import Any, List
+
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class CreateIssueComment(Task):
+    """
+    Task for creating a comment on the specified GitHub issue using the v3 version of the
+    GitHub REST API.
+
+    Args:
+        - repo (str, optional): the name of the repository to create the comment in; must be
+            provided in the form `organization/repo_name`; can also be provided to the `run`
+            method
+        - issue_number (int, optional): the ID of the issue to create the comment on.
+        - body (str, optional): the body of the comment; can also be provided to the `run` method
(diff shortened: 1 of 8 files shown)


src/prefect/tasks/github/comments.py
```diff
@@ -0,0 +1,86 @@
...
+        - repo (str, optional): the name of the repository to create the comment in; must be
+            provided in the form `organization/repo_name`; can also be provided to the `run`
+            method
+        - issue_number (int, optional): the ID of the issue to create the comment on.
+        - body (str, optional): the body of the comment; can also be provided to the `run` method
+        - **kwargs (Any, optional): additional keyword arguments to pass to the standard Task
+            init method
+
+    References:
+        - https://developer.github.com/v3/issues/comments/#create-an-issue-comment
```
harupy, line 23: What do you think about adding a link like this to clarify which API end point each taks uses?

harupy: I think [create_line](https://github.com/PrefectHQ/prefect/blob/f554d1313638286d4183b136f97b56bfbfae58db/docs/generate_docs.py#L129) needs to be updated to render `References` block properly

https://github.com/PrefectHQ/prefect/blob/f554d1313638286d4183b136f97b56bfbfae58db/docs/generate_docs.py#L129-L136

src/prefect/tasks/github/comments.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def __init__(
+        self,
+        repo: str = None,
+        issue_number: int = None,
+        body: str = None,
+        **kwargs: Any
+    ):
```
harupy, line 32: Should I use `Optional` here?

```suggestion
    def __init__(
        self,
        repo: Optional[str] = None,
        issue_number: Optional[int] = None,
        body: Optiona[str] = None,
        **kwargs: Any
    ):
```

jcrist: I believe mypy does this automatically, I think what you have here is fine.

src/prefect/tasks/github/comments.py
jcrist: I like having a `References` section, if you'd want to make the required changes here to support this I'd be +1 on it. I think your suggestion above should be sufficient. Would you be up to adding links like this to the other github tasks as well?

jcrist (commented):
Thanks @harupy, overall this looks great to me!

src/prefect/tasks/github/comments.py
harupy: @jcrist Gotcha. I'll make the change.

jlowin:
@harupy please make sure to add a changes file with yourself as the contributor!

harupy:
@jlowin Thanks! Done!

src/prefect/tasks/github/comments.py
```diff
@@ -0,0 +1,87 @@
...
+            provided in the form `organization/repo_name`; can also be provided to the `run`
+            method
+        - issue_number (int, optional): the ID of the issue to create the comment on.
+        - body (str, optional): the body of the comment; can also be provided to the `run` method
+        - **kwargs (Any, optional): additional keyword arguments to pass to the standard Task
+            init method
+
+    References:
+        - See: [Create an issue
+            comment](https://developer.github.com/v3/issues/comments/#create-an-issue-comment)
```
harupy, line 24: @jcrist Adding a referecne section is more complicated than I expected. Currently, a list item has to follow the following format (otherwise `format_lists` in `generate_doc.py` would ignore it):

```
- xxx: yyy
```

harupy: To allow a format like:

```
- xxx
```

We'd need to fix `format_lists`.

harupy: The current code is rendered like this:

Screen Shot 2020-09-09 at 22 12 18

https://5f58cff0ed4e1e00076692e2--prefect-docs.netlify.app/api/latest/tasks/github.html#createissuecomment

jcrist: Hmmm, that's unfortunate. The fix you've done here looks good. Thanks!

jcrist (approved):

PR #3275, opened 2020-09-07
add boto_kwargs to init method of aws_tasks

Zaubeerer:
## Summary & Changes
kwargs are added to the run method of the secrets manager task to enable the user to specify e.g. `region_name` in the boto client.

## Importance
It seems like there were previously no other functioning methods to specifying `AWS_DEFAULT_REGION`. This provides a general solution to such problems.

## Checklist

This PR:
I only added the kwargs as Args within the AWS secrets manager

### DIFF
--- a/src/prefect/tasks/aws/lambda_function.py
+++ b/src/prefect/tasks/aws/lambda_function.py
@@ -46,6 +46,7 @@ class LambdaCreate(Task):
             subset of incoming requests with Amazon X-Ray
         - layers (List[str], optional): a list of function layers to add to
             the function's execution environment, specify each layer by its ARN
+        - boto_kwargs (dict, optional): additional keyword arguments to forward to the boto client.
         - **kwargs (dict, optional): additional keyword arguments to pass to the
             Task constructor
     """
@@ -72,6 +73,7 @@ class LambdaCreate(Task):
         function_tags: dict = None,
         tracing_config: str = "PassThrough",
         layers: list = None,
+        boto_kwargs: dict = None,
         **kwargs
     ):
         self.function_name = function_name
@@ -105,9 +107,14 @@ class LambdaCreate(Task):
         self.tracing_config = tracing_config
         self.layers = layers
 
+        if boto_kwargs is None:
+            self.boto_kwargs = {}
+        else:
+            self.boto_kwargs = boto_kwargs
+
         super().__init__(**kwargs)
 
-    def run(self, credentials: dict = None, boto_kwargs: dict = None):
+    def run(self, credentials: dict = None):
         """
         Task run method. Creates Lambda function.
 
@@ -117,17 +124,13 @@ class LambdaCreate(Task):
                 with two keys: `ACCESS_KEY` and `SECRET_ACCESS_KEY` which will be
                 passed directly to `boto3`.  If not provided here or in context, `boto3`
                 will fall back on standard AWS rules for authentication.
-            - boto_kwargs (dict, optional): additional keyword arguments to forward to the boto client.
 
         Returns:
             - json: response from AWS CreateFunction endpoint
         """
 
-        if boto_kwargs is None:
-            boto_kwargs = {}
-
         lambda_client = get_boto_client(
-            "lambda", credentials=credentials, **boto_kwargs
+            "lambda", credentials=credentials, **self.boto_kwargs
         )
 
         # create lambda function
@@ -161,16 +164,29 @@ class LambdaDelete(Task):
         - function_name (str): name of the Lambda function to delete
         - qualifier (str, optional): specify a version to delete, if not
             provided, the function will be deleted entirely
+        - boto_kwargs (dict, optional): additional keyword arguments to forward to the boto client.
         - **kwargs (dict, optional): additional keyword arguments to pass to the
             Task constructor
     """
 
-    def __init__(self, function_name: str, qualifier: str = "", **kwargs):
+    def __init__(
+        self,
+        function_name: str,
+        qualifier: str = "",
+        boto_kwargs: dict = None,
+        **kwargs
+    ):
         self.function_name = function_name
         self.qualifier = qualifier
+
+        if boto_kwargs is None:
+            self.boto_kwargs = {}
+        else:
+            self.boto_kwargs = boto_kwargs
+
         super().__init__(**kwargs)
 
-    def run(self, credentials: dict = None, boto_kwargs: dict = None):
+    def run(self, credentials: dict = None):
         """
         Task run method. Deletes Lambda function.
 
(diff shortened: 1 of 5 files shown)


src/prefect/tasks/aws/secrets_manager.py
```diff
@@ -38,6 +38,8 @@ def run(self, secret: str = None, credentials: str = None) -> dict:
                 with two keys: `ACCESS_KEY` and `SECRET_ACCESS_KEY` which will be
                 passed directly to `boto3`.  If not provided here or in context, `boto3`
                 will fall back on standard AWS rules for authentication.
+            - **kwargs (dict, optional): additional keyword arguments to pass to the
+                Task constructor
```
jcrist, line 42: These keyword arguments aren't for the task constructor, they're for boto.

Rather than taking in `**kwargs`, perhaps this task (and all other aws tasks) should get a `boto_kwargs=None` keyword argument for kwargs to forward to boto. This makes more sense to me, since we might want to do something else with `**kwargs` in the future, it's clearer where each argument is going.

If you want, it'd be nice to add `boto_kwargs` to all the aws tasks, but I'd also be happy to merge if you just added it here alone (I can take care of the rest later). Up to you.

Zaubeerer: Very good point! I will add it in a bit to all the aws tasks. 👍 

Just to be sure, you mean to have a boto_kwargs that is per default set to none and otherwise is a dictionary that is forwarded to the respective boto call as `**boto_kwargs` right?

jcrist: > Just to be sure, you mean to have a boto_kwargs that is per default set to none and otherwise is a dictionary that is forwarded to the respective boto call as `**boto_kwargs` right?

Yep, exactly that. Should probably store it on the `Task` always as a dict so we only worry about `None` checks in the constructor. Thanks!

Zaubeerer: @jcrist, what do you think about the changes?

src/prefect/tasks/aws/lambda_function.py
```diff
@@ -117,12 +117,19 @@ def run(self, credentials: dict = None):
                 with two keys: `ACCESS_KEY` and `SECRET_ACCESS_KEY` which will be
                 passed directly to `boto3`.  If not provided here or in context, `boto3`
                 will fall back on standard AWS rules for authentication.
+            - boto_kwargs (dict, optional): additional keyword arguments to pass to the
+                Task constructor that are forwarded as kwargs to the boto client
```
jcrist, line 121: Edit on the docstring, would be good to apply this everywhere (just commenting here).

```suggestion
            - boto_kwargs (dict, optional): additional keyword arguments to
                forward to the boto client.
```

jcrist (changes requested):
Can you add `boto_kwargs` to the `__init__` for each task as well? I'm not actually sure if it needs to be part of the `run` method (what kwargs would you pass at runtime rather than flow build time?), but if your use case needs it in `run` then feel free to keep it there. If not, best to limit the option for now.

src/prefect/tasks/aws/lambda_function.py
Zaubeerer: Just edited the boto_kwargs descriptions accordingly.

Zaubeerer:
> Can you add `boto_kwargs` to the `__init__` for each task as well? I'm not actually sure if it needs to be part of the `run` method (what kwargs would you pass at runtime rather than flow build time?), but if your use case needs it in `run` then feel free to keep it there. If not, best to limit the option for now.

Ah, you mean adding `boto_kwargs` as variable `self.boto_kwargs` and checking for `None` in the `__init__` and then calling it in the run method as follows?

```
lambda_client = get_boto_client("lambda", credentials=credentials, self.boto_kwargs)
```

Makes sense to me!

I also thought of using the parameter in the `__init__` method first.

jcrist:
Yeah, if that'd work, then that'd be my preference. The `defaults_from_attrs` decorator can be used to support both `__init__` and `run` usage, but if we don't need runtime configuration then I'd opt for `__init__` only for now.

Zaubeerer:
Works for my example with the secret manager :)

I get an error concerning black formatting, although I black formatted it locally...

Could you finalize those formatting issues?

Should be only minor changes.

Thanks for the support!

Zaubeerer:
@jcrist, 

1. could you either check out the black formatting issue or let me know whether prefect derives from the black standard formatting?

black is the only reason for code style to block...

2. What do you think about the changes?

joshmeek:
@Zaubeerer I pushed the black format fix. Prefect is using the most recent black release and if you see something different locally when running `black` you may need to upgrade its version 🙂

Zaubeerer:
> @Zaubeerer I pushed the black format fix. Prefect is using the most recent black release and if you see something different locally when running `black` you may need to upgrade its version 🙂

thanks a lot!

I just checked and I am at `black 19.10b0`, seems to be most recent 🤔

joshmeek:
@Zaubeerer my version is `20.8b1` 😉

Zaubeerer:
> @Zaubeerer my version is `20.8b1` 😉

Yep, `pip install --upgrade black` did the trick.

I guess, now all systems are go for merge?

jcrist (commented):
LGTM @Zaubeerer, would you mind adding a [changelog entry](https://docs.prefect.io/core/development/contributing.html#changelog) with yourself as a contributor?

Also, it looks like your `git` cli is not linked to your github profile. Not a big deal, but if you want to fix this (and link the commits to your profile) you might see:

- https://docs.github.com/en/enterprise/2.18/user/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user
- https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git

Zaubeerer:
> LGTM @Zaubeerer, would you mind adding a [changelog entry](https://docs.prefect.io/core/development/contributing.html#changelog) with yourself as a contributor?
> 
> Also, it looks like your `git` cli is not linked to your github profile. Not a big deal, but if you want to fix this (and link the commits to your profile) you might see:
> 
> * https://docs.github.com/en/enterprise/2.18/user/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user
> * https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git

Thanks for the hints, I took care of it :)

jcrist (approved):

PR #3281, opened 2020-09-08
Pause resume idiom

kingsleyb:
## Summary
Idiom for manual_only trigger, demonstrating how to pause a task that can only be resumed with manual intervention. 

## Checklist

This PR:


docs/core/idioms/pause-for-approval.md
```diff
@@ -0,0 +1,53 @@
+# Pause for Approval
```
joshmeek, line 1: Could you add this to the list on `idioms.md`?

docs/core/idioms/pause-for-approval.md
```diff
@@ -0,0 +1,53 @@
...
+
+A manual-only trigger that has been applied to a task will pause the task once it is reached. The task can only be resumed by an individual or event, either from the Prefect UI or GraphQL API.
```
cicdw, line 3: I suggest adding a little more context prior to this line; something like:

There are many situations where a task requires some form of manual approval in order to run successfully.  Prefect has two built-in mechanisms for achieving this pattern: 
- a manual-only trigger that can be configured on a per task basis, and prevents that task from running unless a user approves it
- a `PAUSE` signal that can be raised programmatically when certain conditions are met

Both of these mechanisms mean that the task can only be resumed by an individual or event, either from the Prefect UI or GraphQL API.

docs/core/idioms/pause-for-approval.md
```diff
@@ -0,0 +1,53 @@
...
+    return True
+
+@task
+def test(build_result):
+    print("test task")
+    build_result == True
+
+@task(trigger=manual_only)
+def deploy():
+    print("deploy task")
```
cicdw, line 28: I suggest a docstring on this task that reinforces that it will only run if someone approves it

cicdw (changes requested):

docs/core/idioms/pause-for-approval.md
kingsleyb: Added.

docs/core/idioms/pause-for-approval.md
kingsleyb: See changes.

cicdw (dismissed):

docs/core/idioms/pause-for-approval.md
kingsleyb: added

cicdw (approved):

PR #3307, opened 2020-09-13
fix logging in MySQL tasks

jameslamb:
## Summary

This PR fixes logging in MySQL tasks, which could lead to misleading error logs that obscure the true cause of errors being logged.

I found these with `pylint`. You can find them by running the following:

```shell
pylint src/prefect/ | grep logging-too-many-args
```

## Changes

This PR changes cases like this: `logging.debug("error: ", e)` to `logging.debug("error: %s", e)`.

## Importance

The missing formatting strings in the logging statements changed by this PR will result in confusing logs when they're triggered. Consider the following

**without format string**

```python
import logging

try:
    {}["a"]
except Exception as e:
    logging.error("Fetch Error: ", e)
```

```text
--- Logging error ---
  File "<stdin>", line 2, in <module>
KeyError: 'a'

  File "/Users/jlamb/miniconda3/lib/python3.6/logging/__init__.py", line 994, in emit
    msg = self.format(record)
  File "/Users/jlamb/miniconda3/lib/python3.6/logging/__init__.py", line 840, in format
    return fmt.format(record)
  File "/Users/jlamb/miniconda3/lib/python3.6/logging/__init__.py", line 577, in format
    record.message = record.getMessage()
  File "/Users/jlamb/miniconda3/lib/python3.6/logging/__init__.py", line 338, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "<stdin>", line 4, in <module>
Message: 'Fetch Error: '
Arguments: (KeyError('a',),)
```

**with format string**

```python
import logging

try:
    {}["a"]
except Exception as e:
    logging.error("Fetch Error: %s", str(e))
```

```text
ERROR:root:Fetch Error: 'a'
```

## Checklist

This PR:

### DIFF
--- a/src/prefect/contrib/tasks/mysql/mysql.py
+++ b/src/prefect/contrib/tasks/mysql/mysql.py
@@ -83,12 +83,12 @@ class MySQLExecute(Task):
                         conn.commit()
 
             conn.close()
-            logging.debug("Execute Results: ", executed)
+            logging.debug("Execute Results: %s", executed)
             return executed
 
         except (Exception, pymysql.MySQLError) as e:
             conn.close()
-            logging.debug("Execute Error: ", e)
+            logging.debug("Execute Error: %s", e)
             raise e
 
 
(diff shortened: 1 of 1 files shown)


cicdw (approved):
Great catch @jameslamb!  Thanks for taking the time to run `pylint` and look out for things like this - it is genuinely appreciated

PR #3324, opened 2020-09-15
Use value_

### LINKED ISSUE #3034: `value` can not be used as task argument name with prefect server
## Description

For the following workflow:
```
@prefect.task
def t(value):
    print(value)
    return value

with prefect.Flow("testf") as f:
    t(2)
```

When running with the server, I got

```
[2020-07-27 06:45:26] ERROR - prefect.CloudTaskRunner | Unexpected error: TypeError("write() got multiple values for argument 'value'")
  File "/home//.local/lib/python3.7/site-packages/prefect/engine/runner.py", line 48, in inner
  File "/home//.local/lib/python3.7/site-packages/prefect/engine/task_runner.py", line 839, in get_task_run_state
    result = self.result.write(value, **formatting_kwargs,)
TypeError: write() got multiple values for argument 'value'
```

It seems that `value` is conflicting with an internal parameter.

prefect version: 0.12.5

cicdw:
## Summary

Once we began templating using task keyword argument names as well, we introduced a minor bug where any task with an argument named "value" would fail at runtime; this was caused by our Result classes all using `value` as the kwarg name for `result.write`, and so the two kwargs were competing. 

## Changes

This PR updates the internally used kwarg name to `value_` to prevent overshadowing user kwargs.

## Importance

Closes #3034 

## Checklist

This PR:

### DIFF
--- a/changes/issue3034.yaml
+++ b/changes/issue3034.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix issue with result templating that failed on task arguments named 'value' - [#3034](https://github.com/PrefectHQ/prefect/issues/3034)"
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -257,12 +257,12 @@ class Result(ResultInterface):
             "see https://docs.prefect.io/core/concepts/results.html"
         )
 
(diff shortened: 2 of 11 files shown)


jlowin (approved):

Marlin-Na:
Thanks for the fix!

PR #2760, opened 2020-06-10
DateTime parameter

jlowin:
**Thanks for contributing to Prefect!**

Please describe your work and make sure your PR:

- [ ] adds new tests (if appropriate)
- [ ] add a changelog entry in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

Note that your PR will not be reviewed unless all three boxes are checked.

PR #3327, opened 2020-09-16
Add DateTimeParameter

jlowin:
## Summary

Adds a DateTime parameter that returns Python datetime objects, even if provided a string, and also safely serializes them as PrefectResults.

Supersedes #2760

## Changes

- Introduces `parameters.DateTimeParameter`
- Introduces `serializers.DateTimeSerializer`, which serializes datetimes in a human-readable format
- Allows PrefectResult to accept a `DateTimeSerializer` in addition to `JSONSerializer`

## Checklist

This PR:


jlowin:
cc @bnaul btw

docs/outline.toml
```diff
@@ -210,7 +210,7 @@ classes = ["PrefectResult", "GCSResult", "LocalResult", "S3Result", "AzureResult
 [pages.engine.serializers]
 title = "Result Serializers"
 module = "prefect.engine.serializers"
-classes = ["Serializer", "PickleSerializer", "JSONSerializer", "PandasSerializer"]
+classes = ["Serializer", "PickleSerializer", "JSONSerializer", "DateTimeSerializer", "PandasSerializer"]
```
cicdw, line 213: Could you also add the new `DateTimeParameter` class to the docs?

src/prefect/core/parameter.py
```diff
@@ -112,3 +114,29 @@ def serialize(self) -> Dict[str, Any]:
             - dict representing this parameter
         """
         return prefect.serialization.task.ParameterSchema().dump(self)
+
+
+class DateTimeParameter(Parameter):
+    """
+    A DateTimeParameter that casts its input as a DateTime
+    Args:
```
cicdw, line 122: Once this is added to our API docs, I think you'll need a newline above `Args:`

src/prefect/core/parameter.py
```diff
@@ -112,3 +114,29 @@ def serialize(self) -> Dict[str, Any]:
...
+        - name (str): the Parameter name.
+        - required (bool, optional): If True, the Parameter is required and the default
+            value is ignored.
+        - default (any, optional): A default value for the parameter. If the default
```
cicdw, line 126: Hmmm can you actually provide a `default` to this class?  This could be confusing - I think you can provide a default but it needs to be an already serialized datetime object.  We should either document this or just not let you provide a default to this parameter type?

cicdw (changes requested):
Good stuff - I have a few minor requests and one larger question about default values

src/prefect/core/parameter.py
bnaul: I'm sure this is too wacky for y'all but in my custom version of this I had it default to `datetime.datetime.now()` 😄

jlowin: At one time we did allow or plan to allow `callable` defaults, or am I off?

jcrist: > I think you can provide a default but it needs to be an already serialized datetime object.

Why is this true? Couldn't the serializer be modified to take care of this automatically? I'd find forcing the default to be different than the wrapped type non-intuitive.

cicdw: With the implementation in this PR, this attribute will be serialized to Cloud using this logic: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/serialization/task.py#L123 which requires the default to be JSON compatible. 

We could introduce a whole new serializer for this new class, but that seems like a slippery slope to me.

And yea, I agree that the behavior of a `default` on this class would be wonky as-currently-implemented - I propose removing the ability to specify a default on this class.

jlowin: I removed the ability to specify any custom default (but allowed `None` so the parameter can be optional)

cicdw:
LGTM @jlowin but you have a doc test failure

cicdw (approved):

PR #3330, opened 2020-09-16
Month end and specific day filters

znicholasbrown:
## Summary

Adds 2 new schedule filters, `is_day_of_week` and `is_month_start`, on a passed day of the week and the first day of the month, respectively.

## Changes

* Adds 2 filter methods to `schedules.filters`
* Imports these filters to `schedule.NewScheduleSchema`
* Adds tests for both filters

## Importance

More filters! (Requested by someone in the community)

## Checklist

This PR:


src/prefect/schedules/filters.py
```diff
@@ -181,3 +181,40 @@ def is_weekend(dt: datetime) -> bool:
         - bool: True if the date is a weekend; False otherwise
     """
     return pendulum.instance(dt).weekday() > 4
+
+
+def is_month_start_or_specific_day(day_of_week: int) -> Callable[[datetime], bool]:
```
jcrist, line 186: Since we support OR operations for filters, why combine these into a single filter? Seems to me it might be more composable/easier to have:

- `is_month_start` - is it the start of a month. We already have `is_month_end`, so this follows the existing pattern
- `is_day_of_week(n)` - is it a specific day of the week.

`is_month_start_or_specific_day` could then be built out of these smaller blocks.

jcrist (changes requested):

src/prefect/schedules/filters.py
znicholasbrown: Yeah that's fair, I can separate them @jcrist

znicholasbrown: Ok that's done and tests are updated, just waiting on those to run 👍

changes/pr3330.yaml
```diff
@@ -0,0 +1,2 @@
+enhancement:
+  - 'Adds schedule filters for month end or month start and specific day  - [#3330](https://github.com/PrefectHQ/prefect/pull/3330)'
```
cicdw, line 2: FYI I think these need to be double quotes to get parsed correctly

jcrist: The description also is no longer accurate (`is_month_end` already is in prefect).

tests/schedules/test_filters.py
```diff
@@ -126,3 +126,58 @@ def test_is_weekend(dt):
...
+            )  # day of the week also acts as an offset for each day, which starts at Sunday (0)
+            next_day = day + 1
+
+            assert filter_fn(dt=pendulum.datetime(year=year, month=month, day=day))
+            assert not filter_fn(
+                dt=pendulum.datetime(year=year, month=month, day=next_day)
+            )
+
+    for day in days_week:
+        test_day_of_week(day)
```
jcrist, line 179: Small nit - missing trailing newline (usually there is an editor configuration to fix this automatically for you).

jcrist (commented):
Overall this looks good to me, just one small nit.

tests/schedules/test_filters.py
znicholasbrown: Oops, sorry! Fixed 👍

cicdw:
@znicholasbrown it looks like your test for `DateTime(1971, 2, 21, 0, 0, 0, tzinfo=Timezone('UTC'))` is failing

znicholasbrown:
@cicdw - right you are, TIL Python datetime index 0 is Monday instead of Sunday 😄

jcrist (approved):

Issue #3329, opened 2020-09-16
Brittle Test


cicdw:
I just saw this test failure on a CI job that only updated docs:

```python
    def test_gzip_compression(self, monkeypatch):
        task = S3Upload("bucket")
        byte_string = b"col1,col2,col3\nfake,data,1\nfalse,info,2\n"
        gzip_data = gzip.compress(byte_string)
    
        client = MagicMock()
        boto3 = MagicMock(client=MagicMock(return_value=client))
        monkeypatch.setattr("prefect.utilities.aws.boto3", boto3)
    
        task.run(byte_string, key="key", compression="gzip")
        args, kwargs = client.upload_fileobj.call_args_list[0]
        gzip_data_stream = args[0]
>       assert gzip_data_stream.read() == gzip_data
E       assert (b'\x1f\x8b\x08\x00\xff"b_\x02\xffK\xce\xcf1\xd4I\xce\xcf1\x02\x11\xc6\\i'\n b'\x89\xd9\xa9:)\x89%\x89:\x86@vNq\xaaNf^Z\xbe\x8e\x11\x17\x00Q\x15\x7fk'\n b'(\x00\x00\x00') == (b'\x1f\x8b\x08\x00\xfe"b_\x02\xffK\xce\xcf1\xd4I\xce\xcf1\x02\x11\xc6\\i'\n b'\x89\xd9\xa9:)\x89%\x89:\x86@vNq\xaaNf^Z\xbe\x8e\x11\x17\x00Q\x15\x7fk'\n b'(\x00\x00\x00')
E         At index 4 diff: b'\xff' != b'\xfe'
E         Full diff:
E           (
E         -  b'\x1f\x8b\x08\x00\xfe"b_\x02\xffK\xce\xcf1\xd4I\xce\xcf1\x02\x11\xc6\\i'
E         ?                       ^
E         +  b'\x1f\x8b\x08\x00\xff"b_\x02\xffK\xce\xcf1\xd4I\xce\xcf1\x02\x11\xc6\\i'
E         ?                       ^
E            b'\x89\xd9\xa9:)\x89%\x89:\x86@vNq\xaaNf^Z\xbe\x8e\x11\x17\x00Q\x15\x7fk'
E            b'(\x00\x00\x00',
E           )

tests/tasks/aws/test_s3.py:91: AssertionError
```

It appears that there are many different byte strings that all decompress back to the same original string:
```python
[...]

PR #3331, opened 2020-09-16
Fix S3 gzip compression test

madelgi:
## Summary

Fixes a broken test for the S3Upload class. See [3329](https://github.com/PrefectHQ/prefect/issues/3329).

## Changes

I guess `gzip` uses the current timestamp when compressing data -- looks like `compress` offers an `mtime` argument to override this, but I think the better option would be to just decompress the result and compare it with the original bytes, rather than comparing compressed data.

## Importance

Good to not have a test that sometimes breaks!

## Checklist

This PR:


joshmeek (approved):
Thanks @madelgi!

PR #3338, opened 2020-09-17
Support configuring executor on the flow

jcrist:
(*Split off of #3333*)

This moves the `executor` configuration to the `Flow` itself, rather than on the "environment". This is nice, as it improves parity between `flow.register()` and `flow.run()`, where the same configured executor is used. Several users have been confused in the past about why `flow.run()` doesn't use the configuration in the `environment`.

Another benefit here is that Flows always have to be loaded (usually unpickled), but environments are sometimes unpickled and sometimes loaded from json from cloud (on the agent). This distinction between a fully loaded environment and a half-loaded environment (that might be missing some pickled components) is confusing, removing the need for a half-loaded environment is a goal of #3333. Moving one more piece of pickled state off of environments simplifies this.

### DIFF
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -229,7 +229,12 @@ class FlowRunner(Runner):
         task_contexts = dict(task_contexts or {})
         parameters = dict(parameters or {})
         if executor is None:
-            executor = prefect.engine.get_default_executor_class()()
+            # Use the executor on the flow, if configured
+            executor = getattr(self.flow, "executor", None)
+            if executor is None:
+                executor = prefect.engine.get_default_executor_class()()
+
+        self.logger.debug("Using executor type %s", type(executor).__name__)
 
         try:
             state, task_states, context, task_contexts = self.initialize_run(
--- a/tests/engine/test_flow_runner.py
+++ b/tests/engine/test_flow_runner.py
@@ -708,6 +708,16 @@ class TestRunCount:
         assert isinstance(state3.result[t2], Failed)
 
 
+def test_flow_runner_uses_default_executor_on_flow_if_present():
+    t = SuccessTask()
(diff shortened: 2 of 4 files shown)


src/prefect/engine/flow_runner.py
```diff
@@ -229,7 +229,12 @@ def run(
         task_contexts = dict(task_contexts or {})
         parameters = dict(parameters or {})
         if executor is None:
-            executor = prefect.engine.get_default_executor_class()()
+            # Use the executor on the flow, if configured
+            executor = getattr(self.flow, "executor", None)
+            if executor is None:
+                executor = prefect.engine.get_default_executor_class()()
+
+        self.logger.debug("Using executor type %s", type(executor).__name__)
```
cicdw, line 237: 👏

tests/engine/test_flow_runner.py
```diff
@@ -708,6 +708,16 @@ def test_run_count_tracked_via_retry_states(self):
...
 
 
+def test_flow_runner_uses_default_executor_on_flow_if_present():
+    t = SuccessTask()
+    with Flow(name="test", executor=Executor()) as flow:
+        result = t()
+
+    with raise_on_exception():
+        with pytest.raises(NotImplementedError):
+            FlowRunner(flow=flow).run(executor=Executor())
```
cicdw, line 718:
```suggestion
            FlowRunner(flow=flow).run(executor=None)
```
otherwise you don't know whether it was using the passed in Executor class or the Flow executor class, right?

cicdw (changes requested):
I _think_ there was a typo in your test - let me know, but otherwise LGTM!  The first step is now taken...

tests/engine/test_flow_runner.py
jcrist: Oop, good catch. Fixed.

cicdw (approved):
LGTM

PR #3346, opened 2020-09-18
A few mypy fixups

### LINKED ISSUE #3056: Use of @curry with task decorator breaks type checking
## Description

If I use the task decorator with arguments, pylint complains that I am not passing `fn`.

e.g.
```python
@task(tags = ['foo'])
def foo():
    return True
```
Results in:
image

## Expected Behavior

No linting error

## Reproduction

(perhaps not needed)

## Environment

I'm not exactly sure how best to fix this; typing `@curry` seems to be tricky. Still I'm not sure if this is an upstream bug either (or what, exactly, upstream, is broken).

### LINKED ISSUE #3191: Using static type checking as qa on a Flow library
## Current behavior
As a user developing `Flow`s, I struggle with debugging flows, since each of the tasks represents delayed execution. Running a flow is the most complete test, but since flows can change the state of external systems, it would be helpful to have some qa that my flow is set-up correctly before running/registering. This is especially impactful in the situation where I have a [flow library](https://docs.prefect.io/orchestration/tutorial/multiple.html) and want to implement this qa as part of a CI pipeline.

## Proposed behavior
For flows with `Task`s that have data dependencies between them, I could qa on the types passed between different tasks. Right now, tasks are typed to hint at the types encountered when the flow is registered. Imagine a scenario, where the types hinted at the task execution result (the result of the delayed execution). You could run mypy and ensure that the dependencies between flows have sensible types (plus your IDE will be more helpful)

## Example
```
from typing import Any, Callable, TypeVar

import prefect
import prefect.tasks.core.function


F = TypeVar('F', bound=Callable[..., Any])


def typed_task(fn: F, **task_init_kwargs: Any) -> F:
    """Same dynamic code as `prefect.task`, but preserves function signature for debugging in a Flow context.
    """
    return prefect.tasks.core.function.FunctionTask(fn=fn, **task_init_kwargs)  # type: ignore


@typed_task
def task_one() -> str:
    return "1"


@typed_task
def task_two(arg: str) -> int:
    return int(arg)


with prefect.Flow("test_type_hints") as flow:
    result_one: str = task_one()  # typed to hint at the execution result 
    result_two: int = task_two(result_one)
flow.register()
```

jcrist:
- Fix `__signature__` of `Task` types to work on both the class and on
  instances.
- Fix `task` decorator to support type checking. mypy doesn't like
  decorators that accept optional arguments, to handle this we have to
  add additional annotations.
- *Almost* fix `resource_manager` decorator to support type checking.
  This is annoying, mypy assumes (incorrectly) that the output of a
  class decorator is always an instance of the class, even if the
  decorator type signature says otherwise.

Fixes #3056.
Related to, but doesn't fix #3191.

### DIFF
--- a/changes/pr3346.yaml
+++ b/changes/pr3346.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix mypy type checking for tasks created with `prefect.task` - [#3346](https://github.com/PrefectHQ/prefect/pull/3346)"
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -83,6 +83,9 @@ class instance_property:
     def __init__(self, func: Callable):
         self.func = func
 
+    def __getattr__(self, k: str) -> Any:
+        return getattr(self.func, k)
+
     def __get__(self, obj: Any, cls: Any) -> Any:
         if obj is None:
             raise AttributeError
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -77,6 +77,18 @@ class SignatureValidator(type):
         return type.__new__(cls, name, parents, methods)  # type: ignore
 
 
+class instance_property:
+    """Like property, but only available on instances, not the class"""
+
+    def __init__(self, func: Callable):
+        self.func = func
+
+    def __get__(self, obj: Any, cls: Any) -> Any:
+        if obj is None:
+            raise AttributeError
+        return self.func(obj)
+
+
 class Task(metaclass=SignatureValidator):
     """
     The Task class which is used as the full representation of a unit of work.
@@ -426,7 +438,7 @@ class Task(metaclass=SignatureValidator):
 
         return new
 
-    @property
+    @instance_property
     def __signature__(self) -> inspect.Signature:
         """Dynamically generate the signature, replacing ``*args``/``**kwargs``
         with parameters from ``run``"""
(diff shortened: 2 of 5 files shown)


jcrist:
The type checking this supports only checks that the output of a `task` decorated function is a `Task` (before this would be `Any`). This should also work for `resource_manager` decorated classes, but doesn't due to a bug in mypy. I'd be fine with deleting the changes to the `resource_manager` file, as they don't really help anything, but they should and it took a bit of experimenting to get even this far.

Demo:

```python


@task
def hello(name: str) -> str:
    print("Hello %s" % name)
    return name


@task(name="hello-2")
def hello2(name: str) -> None:
    print("Hello %s" % name)


reveal_type(hello)
reveal_type(hello2)


with Flow("greeter") as flow:
    res = hello("world")
    reveal_type(res)
    res2 = hello2(res)
    reveal_type(res2)

flow.run()
```

```shell
$ mypy test.py
test2.py:15: note: Revealed type is 'prefect.tasks.core.function.FunctionTask'
test2.py:16: note: Revealed type is 'prefect.tasks.core.function.FunctionTask'
test2.py:21: note: Revealed type is 'prefect.core.task.Task'
test2.py:23: note: Revealed type is 'prefect.core.task.Task'
```

cicdw (dismissed):
I'll admit I'm not 100% in the loop on `overload` here, but the changes seem safe and valuable, so LGTM!

cicdw (approved):

jcrist:
> I'll admit I'm not 100% in the loop on overload here, but the changes seem safe and valuable, so LGTM!

I am also not a mypy user, but per googling on how to do this, this appears to be the recommended approach.

Issue #3335, opened 2020-09-17
LocalModule Storage Option


jacksund:
Hi all,

This is a cross posting of discussion #3304. This implementation will allow for easier sharing of python workflow libraries that are installed via pip/conda in any type of environment, including LocalEnvironment.

## Proposed behavior
LocalModule will serve just like Local with stored_as_script=True, where instead of a path='/path/to/my/flows/exampleflow.py', you could use module_path='from myrepo.flows import exampleflow' or 'myrepo.flows.exampleflow'. Thus the flow is accessible in any environment where that module import works, not just the environments that share the same path structure. 

In an example workflow library, workflows are separated into scripts and can thus be imported and used. One can easily do the following psuedo code...
```
from myrepo.workflows import exampleflow
exampleflow.flow.register(...)
```
or alternatively... While I import the entire script above, it may be easier to simply import the single instance (the flow variable).
```
from myrepo.workflows.exampleflow import flow
flow.register(...)
```
where the exampleflow.py looks like...

```
import prefect
from prefect import task, Flow

@task
def hello_task():
    logger = prefect.context.get("logger")
    logger.info("Hello, Cloud!")

flow = Flow("hello-flow", tasks=[hello_task])

from prefect.environments.storage import LocalModule
flow.storage = LocalModule(module_path=__name__) # __name__ will give 'myrepo.workflows.exampleflow'
```

## The LocalModule Class

Just to start with implementation, LocalModule should be pretty simple to implement in terms of serialization and dynamic loading.

The serialization of a module is simply saving the path and version:
```
info = {'module': 'myrepo.workflows.exampleflow',
'version': 'v0.0.1',
'class_str': 'flow', # (optional) this is what I made my variable name for my Flow instance above
}
```
And dynamic loading can be done with...
```
module_path = info['module'] # grab the string from the serialized info
class_str = info['class_str'] # grab the name of the variable for Flow instance
# (optional) some version check -- maybe warn or raise error if there is a difference...?
# and now use the string to import
import importlib
module = importlib.import_module(module_path)
# and if the Flow is saved as a variable instance named 'flow', we can grab it using
flow = getattr(module, class_str)
```

jcrist:
Hi @jacksund, the above design makes sense to me. Is this something you'd be interested in working on?

jacksund:
[...]

PR #3351, opened 2020-09-18
add localmod storage

jacksund:
## Summary
This is the start to a LocalModule storage class. It is currently working with Prefect Server along with a LocalAgent, but full testing and features have not been added yet. Thus, this is a draft PR for discussion of code. Prior discussion of this topic is within #3304 and #3335 threads. 

## Changes

- adds localmod.py (with LocalModule class) to prefect.environments.storage
- adds LocalModule storage schema to prefect.serialization.storage
- adds LocalModule storage as a default label and support for LocalAgent
- adds tests for the above changes _[incomplete at the moment]_

## Importance
Local module based storage allows for easier sharing of workflow libraries, where flows are accessible in the python path. LocalModule will serve just like Local with stored_as_script=True, where instead of a path='/path/to/my/flows/exampleflow.py', you could use module_path='from myrepo.flows import exampleflow' or 'myrepo.flows.exampleflow'. Thus the flow is accessible in any environment where that module import works, not just the environments that share the same path structure.

## Checklist

_NOTE -- this is a draft PR so the tasks below are still ongoing._

This PR:


cicdw (changes requested):
Hi @jacksund - first off, apologies for the extended delay on reviewing.  Now that I've seen your implementation (which is great!) I realize that we don't need to create an entirely new class for this storage pattern.

In particular, I think we can expand the types of `location` strings we accept on the pre-existing `Local` storage class to allow for module imports instead of file paths.  And note that we will only use this functionality when `store_as_script=True`.  In that case we can do something like
```python
module_path, class_str = location.split(".")[:-1], location.split(".")[-1]
```
and recreate your functionality here.

Let me know what you think about that, and apologies for not identifying this earlier!

jacksund:
No worries on the slow reply :). And yeah, I can definitely merge the functionality, so I'll give this a go.

Thanks for taking a look!

jacksund:
Alright so I have a working version if you take a look at my branch again. It ended up being extremely little code, so it's nice and easy to read through. There are a couple things to note though: (1) I haven't updated the doc-strings to describe the new functionality, (2) I didn't change the 'path' parameter to 'location' as you suggest because I'm scared that will cause breaks in the code elsewhere.

While everything looks to be running correctly on my end, I'm completely new to circleci, so I don't understand the build errors that I'm getting. Would you be willing to walk me through these?

src/prefect/utilities/storage.py
```diff
@@ -82,3 +83,35 @@ def extract_flow_from_file(
...
     raise ValueError("No flow found in file.")
+
+def extract_flow_from_module(
+    module_str: str, flow_name: str = None
+) -> "Flow":
+    """
+    Extract a flow object from a python module.
+
+    Args:
+        - module_str (str, optional): A module path pointing to a .py file containing a flow.
```
cicdw, line 94:
```suggestion
        - module_str (str): A module path pointing to a .py file containing a flow.
```

src/prefect/utilities/storage.py
```diff
@@ -82,3 +83,35 @@ def extract_flow_from_file(
...
+            If not set then the first flow object retrieved from file will be returned.
+            The "first" flow object will be based on the dir(module) ordering, which
+            is alphabetical and capitalized letters come first.
+
+    Returns:
+        - Flow: A flow object extracted from a file
+    """
+
+    # load the module
+    module = importlib.import_module('prefecttest.flow')
```
cicdw, line 106: Looks like some testing hardcoded values - probably meant `module_str` here?

cicdw (changes requested):
Two minor requests.

Your CircleCI job is faliing because of our formatting rules - if you install and run `black` on the codebase you should be good; here's how to do that from the CLI:
```
pip install -U black
black . # note the period
```

Otherwise, if you could add a changelog entry to the `changes/` directory with your contributor info I think we're good to go!

src/prefect/utilities/storage.py
jacksund: oh wow, I can't believe I missed this. Thanks for catching it. Yes, it's supposed to be module_str!

src/prefect/utilities/storage.py
jacksund: Another good catch! I'll go back through the docs on the Local storage class as well and update everything there.

jacksund:
I think that covers the docs and typos. Let me know if I need to do anything else.

src/prefect/utilities/storage.py
```diff
@@ -113,5 +112,5 @@ def extract_flow_from_module(
         for var in dir(module):
             if isinstance(getattr(module, var), prefect.Flow):
                 return getattr(module, var)
-                
+
     raise ValueError("No flow found in file.")
```
cicdw, line 116: @jacksund this isn't a big deal but maybe we should update this error message to say `module` instead of `file`

cicdw: Better yet, the name of the module even: `f"No flow found in {module_str}"`

jacksund: I left it thinking this would portray the same idea, but I think you're right. I'll change it.

cicdw: Thanks for being so responsive!

cicdw (approved):
LGTM thank you for the cool contribution @jacksund !  Hope it's your first of many :D

PR #3352, opened 2020-09-18
Dont use asserts for control flow

### LINKED ISSUE #2918: Problem visualizing a Flow
When I call flow.visualize(), there is no graph created. Debugging this I found that the problem was the following code in flow.py, from line 1238
       if filename:
            graph.render(filename, view=False, format=format)
        else:
            try:
                from IPython import get_ipython
                
                assert get_ipython().config.get("IPKernelApp") is not None              <----------------- the problem
            except Exception:

Now in order to see if the value was really None I added this line of code:
        if filename:
            graph.render(filename, view=False, format=format)
        else:
            try:
                from IPython import get_ipython
                test = get_ipython().config.get("IPKernelApp")   <-----------------------line added
                assert get_ipython().config.get("IPKernelApp") is not None
            except Exception:

and it worked. I know can visualize the graph. Without the line  it does not work.
I have no idea why this worked.

jcrist:
Python assert statements may be dropped dropped by the compiler if running with `-O`/`PYTHONOTIMIZE=1`. As such, they shouldn't be used for control flow, as they might not be executed.

Fixes a bug in `flow.visualize()` where no output would be generated if asserts are optimized out.

Also fixes the `flow.visualize` tests to properly use `pytest.skipif`, rather than xfailing them unconditionally (which makes them useless as tests).

Fixes #2918.

### DIFF
--- a/tests/core/test_flow.py
+++ b/tests/core/test_flow.py
@@ -1232,7 +1232,16 @@ def test_skip_validation_in_init_with_kwarg():
     assert Flow(name="test", edges=[e1, e2], validate=False)
 
 
-@pytest.mark.xfail(raises=ImportError, reason="viz extras not installed.")
+try:
+    import graphviz
+
+    graphviz.pipe("dot", "png", b"graph {a -- b}", quiet=True)
+    no_graphviz = False
+except Exception:
+    no_graphviz = True
+
+
+@pytest.mark.skipif(no_graphviz, reason="viz extras not installed.")
 class TestFlowVisualize:
     def test_visualize_raises_informative_importerror_without_python_graphviz(
         self, monkeypatch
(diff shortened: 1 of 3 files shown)


tests/core/test_flow.py
```diff
@@ -1232,7 +1232,16 @@ def test_skip_validation_in_init_with_kwarg():
...
+try:
+    import graphviz
+
+    graphviz.pipe("dot", "png", b"graph {a -- b}", quiet=True)
+    no_graphviz = False
+except Exception:
+    no_graphviz = True
+
+
+@pytest.mark.skipif(no_graphviz, reason="viz extras not installed.")
```
cicdw, line 1244: This is way clearer (and faster!), but isn't it true that `xfail(raises=ImportError)` will properly `FAIL` if the error type is something different?

cicdw (approved):
Nice, looks good - I left a question that's independent of merging this PR

tests/core/test_flow.py
jcrist: That's a good point, I missed that kwarg. Note that you can have the python `graphviz` module installed without the cli `graphviz` library, so an `ImportError` isn't the only reason to skip these tests.

PR #3360, opened 2020-09-21
Task library documentation and structure clean up

joshmeek:
## Summary
This PR cleans up the task library documentation and moves around some files that were previously in `/contrib`.

## Changes
- Removes duplicate task library documentation
- Adds a small overview and contributing document to the task library docs
- Deprecates importing task library tasks from `/contrib`
- Moves Databricks and MySQL tasks from `prefect.contrib.tasks` to `prefect.tasks` (warning raise if imported from old location)

## Importance
Task library was missing a true home for documentation. The `contrib` directory is not useful and it doesn't benefit having some tasks separate from the rest.

## Checklist

This PR:


.flake8
```diff
@@ -4,6 +4,7 @@ exclude =
     versioneer.py,
     _version.py
     src/prefect/contrib/*
+    src/prefect/tasks/*
```
jcrist, line 7: Before flake8 was passing fine on this, now it's not. Can you fix things so that it passes again?

docs/core/task_library/contributing.md
```diff
@@ -0,0 +1,88 @@
...
+containing all of the [task library code](https://github.com/PrefectHQ/prefect/tree/master/src/prefect/tasks).
+
+For more information on contributing to the Prefect library as whole check out the
+[development documentation](/core/development/overview.html).
+
+## Testing
+
+Due to the nature of the tasks in the task library interacting with a wide range of services from various
+contributors it is not condusive for Prefect integration tests to maintain tests that communicate with
+all of these services. Because of this, it is integral that users who are contributing tasks to the task
```
jcrist, line 82:
```suggestion
contributors it is not always easy for Prefect to maintain tests that communicate with
all of these services. Because of this, it is integral that users who are contributing tasks to the task
```

docs/core/task_library/overview.md
```diff
@@ -0,0 +1,115 @@
...
+| <img src="/logos/sqlite.png" height=128 width=128 style="max-height: 128px; max-width: 128px;"> [<p>SQLite</p>](https://docs.prefect.io/api/latest/tasks/sqlite.html) | <img src="/logos/tlogo.png" height=128 width=128 style="max-height: 128px; max-width: 128px;"> [<p>Twitter</p>](https://docs.prefect.io/api/latest/tasks/twitter.html) |
+
+## Task library in action
+
+Just like any other Prefect [task](/core/concepts/tasks.html), tasks in the task library can easily be
```
jcrist, line 27: Saying things are easy or simple when they might not be for everyone can be frustrating for some users (there's loads of articles about this, but here's one: https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher). Here the word "easily" doesn't add anything; I suggest we remove it.

```suggestion
Just like any other Prefect [task](/core/concepts/tasks.html), tasks in the task library can be
```

.flake8
joshmeek: Flake8 appears to be passing fine, it's codecov which is failing now

jcrist: Sorry, I mean can you drop the changes to the config so that the task library is still linted by flake8? Only the new code you moved over hear should be affecting this, things were passing fine before.

joshmeek:
Codecov has been corrected 👍 @jcrist @cicdw

.flake8
```diff
@@ -11,7 +11,7 @@ ignore =
     E203,
     # Comparing types instead of isinstance
     E721,
-    # Assign a lambda
+    # Assign a lam\bda
```
jcrist, line 14: why?

jcrist (approved):
Overall this looks good to me.

.flake8
joshmeek: why not?

jcrist (approved):

docs/core/task_library/contributing.md
```diff
@@ -0,0 +1,88 @@
...
+}
+</script>
+
+<Tweet id="1298298873878847490"></Tweet>
+
+## Task Structure
+
+This snippet below is the general structure of a task contained in the task library. All you need to do
+is define the task's `__init__` and `run` functions! The `run` function is where any of your task's logic
+will live.
```
cicdw, line 32: We might want to rehash some of the differences between init and run (as in https://docs.prefect.io/core/advanced_tutorials/task-guide.html) here, "your initialization will be called before you register and run your Flow" etc.

docs/core/task_library/contributing.md
```diff
@@ -0,0 +1,88 @@
...
+
+    def __init__(
+        self,
+        your_kwarg: str = None,
+        **kwargs: Any
+    ):
+        self.your_kwarg = your_kwarg
+        super().__init__(**kwargs)
+
+    # Allows init kwargs to be passed to the run function if they are not overridden
```
cicdw, line 55: This isn't super clear to me - maybe
> Allows for certain attributes to optionally be provided at runtime, which will override the values provided at init

/shrug

cicdw (commented):
Left some comments - we can merge this as-is and revisit the comments for the full post + as follow-up PRs.  In general I think we might want to expand on "design" for tasks (e.g., what to include in init vs run) and maybe provide some examples but this is still a great start

joshmeek:
Was just in a conversation with @kingsleyb  and I realized we need to document things like our desired pattern for interacting with secrets and also some more information WRT init and run kwargs, stand by

docs/core/task_library/contributing.md
joshmeek: 👍 going to spruce this up a bit

docs/core/task_library/contributing.md
joshmeek: Going to actually expand on this outside of the code

joshmeek:
@cicdw @jcrist Made some of the changes that I mentioned above 👍

docs/core/task_library/contributing.md
```diff
@@ -0,0 +1,168 @@
...
+def __init__(self, your_kwarg: str = None, **kwargs: Any):
+    self.your_kwarg = your_kwarg
+    super().__init__(**kwargs)
+
+    @defaults_from_attrs("your_kwarg")
+    def run(self, your_kwarg: str = None) -> str:
+        use_your_library(your_kwarg)
+```
+
+is equivalent to
```
cicdw, line 104: Actually these aren't 100% equivalent - if a value is provided at runtime it is _always_ used regardless of the truthiness.  It can even be `None` (test for reference: https://github.com/PrefectHQ/prefect/blob/master/tests/utilities/test_tasks.py#L540-L546)

docs/core/task_library/contributing.md
```diff
@@ -0,0 +1,168 @@
...
+        ...
+
+    # Allows init kwargs to be passed to the run function if they are not overridden
+    @defaults_from_attrs("your_kwarg")
+    def run(self, your_kwarg: str = None, your_secret: str = None) -> str:
+        authenticate_with_your_service(your_secret)
+        ...
+```
+
+This allows users of the task to use Prefect Secrets to securely pass sensitive information to the task:
```
cicdw, line 144:
```suggestion
This allows users of the task to use Prefect Secrets to securely pass sensitive information to the task using whatever secret storage mechanism they prefer:
```

cicdw (changes requested):
Two small suggestions

docs/core/task_library/contributing.md
joshmeek: Okay I'm actually going to remove these snippets

cicdw (approved):

PR #3364, opened 2020-09-22
Expose flow_run_name to .run() method for local runs

veseln:
When running using Server, it is possible to  set `flow_run_name`, which is currently not
possible when doing local runs. This change allows the user to specify  `flow_run_name` when
calling run, in the same way as it is currently done for `flow_run_id`.

## Summary

This change allows the user to specify  `flow_run_name` when doing local runs in the same way as it is currently done for `flow_run_id`.

## Changes

This PR adds an option of specifying `flow_run_name` when running prefect locally in the same way that 
specifying flow_run_id  is possible. This allows the flow_run_name to be settable and not necessarily auto-generated.   

## Importance

Allowing the user to set the flow_run_name for local runs is important because it brings closer prefect functionality when running locally and when using server. It also allows the user to set `flow_run_name` which enables one to use `flow_run_name` as one of the identifiers when defining target for  persistance. 

` target='{flow_run_name}/...'` (as an example)   

## Tests 
Tests were not added, as I couldn't find where the same  functionality for `flow_run_id` (that is already implemented) was tested,  I would gladly add tests if I could get some help identifying where I should add the tests. 

## Checklist

This PR:

### DIFF
--- a/changes/pr3364.yaml
+++ b/changes/pr3364.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Expose flow_run_name to .run() method for local runs - [#3364](https://github.com/PrefectHQ/prefect/pull/3364)"
+
+contributor:
+  - "[Nejc Vesel](https://github.com/veseln)"
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1009,6 +1009,11 @@ class Flow:
             "flow_run_id", kwargs.pop("flow_run_id", str(uuid.uuid4()))
         )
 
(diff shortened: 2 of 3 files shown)


jcrist:
Thanks @veseln, this makes sense to me. Would you mind adding a test in `tests/core/test_flow.py`? Easiest way would be to create a flow with a task that pulls the flow run name from the context, then run the flow with a specified flow run name.

Could you also add a changelog entry with yourself as the contributor? https://docs.prefect.io/core/development/contributing.html#changelog

jcrist (approved):
LGTM, thanks!

veseln:
Great, thanks!

Issue #3306, opened 2020-09-13
Use 'from' to explicitly chain exceptions


jameslamb:
## Current behavior

I wanted to spend a little time contributing this weekend, so I ran `pylint` over this repo. I see this project has a `.pylintrc` but it looks like `pylint` isn't run in CI, so some things do show up.

I found one class of warning from `pylint` that I think is worth addressing: explicitly re-raising exceptions in try-catches with `from`.

The newest version of `pylint` added a check for this. You can see details in ["Explicit Exception Chaining"](https://www.python.org/dev/peps/pep-3134/#explicit-exception-chaining). That PEP describes a way to improve stack traces from try-catched exceptions.

**without `from`**

```python
try:
    {}["a"]
except KeyError:
    raise RuntimeError("hey this does not exist")
```

```text
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
KeyError: 'a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError: hey this does not exist
```

**with `from`**

```python
try:
    {}["a"]
except KeyError  as err:
    raise RuntimeError("hey this does not exist") from err
```

```text
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
KeyError: 'a'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError: hey this does not exist
```

I think this language of "was the direct cause of the following exception" is clearer, and helps you to understand exactly where things started going wrong.

## Proposed behavior

Would you be open to a pull request that changes try-catched exceptions in `prefect` to use this pattern? I wanted to ask first because right now there are 1334 places where that happens so it would touch a lot of code.

you can see them all like this:

```shell
pylint src/prefect/ | grep raise-missing-from > pylint.txt
```

I think this change would improve `prefect` and help save users some debugging time. Especially users who are not very experienced in Python.

Thanks for your time and consideration.

## Notes for Hacktoberfest Contributors

This issue has been left open for [Hacktoberfest 2020](https://hacktoberfest.digitalocean.com/) contributors. If that describes you, welcome!

1. Leave a comment below claiming a sub-module
1. Identify all cases of this issue in that submodule by running code like the following in a terminal:
    ```shell
    pylint src/prefect/engine/cloud | grep "raise-missing-from"
    ```
1. Fix all of these cases.
    - see #3320 for an example
[...]

PR #3383, opened 2020-09-24
use explicit exception chaining in core and engine

brainless:
## Summary
Adds explicit exception chaining so that exception stack traces are a bit easier to comprehend. See #3306 for a lot more detail and background discussion on this.

*note*: Issue body is copied from #3320, created by @jameslamb.

## Changes
Uses explicit exception chaining like `raise RuntimeError("whatever") from exc` in all the places in `src/prefect/core/*` and `src/prefect/engine/*` that are missing it.

You can confirm that they were all caught by running the following from the root of the repo:

```
python -m pylint src/prefect/core --disable=all --enable=W0707
python -m pylint src/prefect/engine --disable=all --enable=W0707
```

I have intentionally left out - `src/prefect/engine/serializers.py:84` since that usage seems valid as it is to me.

## Importance
This pull request will make some of `prefect`'s stack traces a bit easier to understand, to improve the user experience in debugging.

## Checklist

This PR:

### DIFF
--- a/changes/issue3306.yaml
+++ b/changes/issue3306.yaml
@@ -1,0 +1,5 @@
+enhancement:
+    - "Use explicit exception chaining [#3306](https://github.com/PrefectHQ/prefect/issues/3306)"
+
+contributor:
+    - "[Sumit Datta](https://github.com/brainless)"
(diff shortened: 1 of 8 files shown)


joshmeek:
Thanks @brainless!

joshmeek (approved):

PR #3387, opened 2020-09-24
Monday task library

kingsleyb:
## Summary
Adding first task to task library for monday.com that allows you create an item in a board. More tasks to come. 

## Checklist

This PR:

### DIFF
--- a/docs/outline.toml
+++ b/docs/outline.toml
@@ -372,6 +372,11 @@ title = "Jira Tasks"
 module = "prefect.tasks.jira"
 classes = ["JiraTask", "JiraServiceDeskTask"]
 
+[pages.tasks.monday]
+title = "Monday Tasks"
+module = "prefect.contrib.tasks.monday"
+classes = ["CreateItem"]
+
 [pages.tasks.mysql]
 title = "MySQL Tasks"
 module = "prefect.contrib.tasks.mysql"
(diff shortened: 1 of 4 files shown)


src/prefect/tasks/monday/monday.py
```diff
@@ -0,0 +1,101 @@
+import requests
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+from typing import Any
+
+import json
+
+
+class CreateItem(Task):
```
joshmeek, line 9: Can you add this to the docs `outline.toml` so it gets picked up by the API doc generation https://github.com/PrefectHQ/prefect/blob/dec10320dacf2ab7e90881feb09fb71c62729a69/docs/outline.toml#L241-L244

Also realized I need to add a note about adding it to the docs in the contributing doc 👍

kingsleyb: done.

docs/outline.toml
```diff
@@ -372,6 +372,11 @@ title = "Jira Tasks"
 module = "prefect.tasks.jira"
 classes = ["JiraTask", "JiraServiceDeskTask"]
 
+[pages.tasks.monday]
+title = "Monday Tasks"
+module = "prefect.contrib.tasks.monday"
```
cicdw, line 377: Incorrect import path, I think this is causing your tests to fail:
```suggestion
module = "prefect.tasks.monday"
```

joshmeek (approved):
Lgtm @kingsleyb!

PR #3393, opened 2020-09-25
Add option to specify run_name for FlowRUnTask

veseln:
## Summary

Add an option of specifying run_name when using FlowRunTask 

## Changes

It changes FlowRunTask by adding an optional parameter that allows setting run_name

## Importance

It is important, because having an option to set the name for your flow is beneficial for people that rely on flow names for tracking the flows. Having an option to set the flow name, allows more flexibility. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -48,7 +48,7 @@ class FlowRunTask(Task):
             kwargs.setdefault("name", f"Flow {flow_name}")
         super().__init__(**kwargs)
 
-    @defaults_from_attrs("flow_name", "project_name", "parameters", "new_flow_context")
+    @defaults_from_attrs("flow_name", "project_name", "parameters", "new_flow_context", "run_name")
     def run(
         self,
         flow_name: str = None,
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -54,6 +57,7 @@ def run(
         parameters: dict = None,
         idempotency_key: str = None,
         new_flow_context: dict = None,
+        run_name: str = None,
```
cicdw, line 60: Since this option is available both at initialization and at run time, it should be included in the `@defaults_from_attrs` decorator on the `run` method.  Otherwise the value at initialization will never be used

cicdw (changes requested):
Nice enhancement @veseln !  Two small requests:
- add `run_name` to the `@default_from_attrs` decorator
- add a changelog entry (with the "task" header) along with your contributor info

src/prefect/tasks/prefect/flow_run.py
veseln: Thanks, good catch. I added this when I was looking at how this works, but then it seems to have slipped when making the commit.

cicdw (approved):
Thanks @veseln !

PR #3407, opened 2020-09-29
Tweaks to task library documentation

joshmeek:
Adds logo for Monday tasks and adds note about documentation to the task library contributing page


cicdw (approved):
Good stuff - test failure is a one-off one that we've seen before

PR #3408, opened 2020-09-30
Quickfix to solve docker related windows shutil.copytree permissions error

Zaubeerer:
## Summary
This quickfixes a bug where `prefect.Docker` raises the following error:

```
shutil PermissionError: [WinError 5] Access is denied .
```
when trying to push the image:
```
INFO - prefect.Docker | Pushing image to the registry... 
```
## Changes

Adds `copy_function=shutil.copyfile` to `shutil.copytree`.

## Importance

The last release brought more powerful docker capabilities. 

This PR increases probability that these work properly on windows.

## Checklist

This PR:


jcrist:
Per the docs, the only difference here is that `copyfile` doesn't copy over file metadata like permissions. However, `shutil.copy2` will copy file permissions, and `shutil.copytree` still will for directories - both are also used in this code section. It would be good to be uniform here about maintaining permissions, otherwise we risk confusing berhavior. Also note that docker's `COPY` will copy permissions, so we likely want to avoid dropping these if possible.

Any idea why you're running into a permission error? The error was for `.` which should be your current working directory, is there anything special about that directory or the files/directories that are being temporarily copied into it?

Zaubeerer:
OK, for the sake of completeness, this is where I found the solution explaining also shortly what you just mentioned about the difference in copying over permissions: https://stackoverflow.com/questions/53728200/shutil-copytree-permission-denied-when-copied-sys32-config

I will investigate on that either tomorrow or later this week. We had this issue on several machines and I primarily just needed a quickfix. But would also be happy to get a deeper understanding of the cause.

PR #3411, opened 2020-10-01
Config docs

zhen0:
## Summary

Add Server configuration info to the docs

## Changes

Adds a configuration options section to the Server FAQ section

## Importance

We've seen several questions about configuration and the correct configuration options/environment variables for Server in the community slack.  There was also a question about connecting to a pre-exisiting postgres db and I couldn't see a great place to add that info.  This section adds a link to the prefect repo config.toml and gives info about how to set a different API endpoint and how to run Prefect with a pre-exisiting postgres database. It also links to the more detailed explanation about configuration in the Core docs (which are relevant to server but not so easy to find if one is looking for something server specific.)

Ideally when we do a docs refactor, we can create one configuration page. 

## Checklist

This PR:

### DIFF
--- a/docs/orchestration/faq/config.md
+++ b/docs/orchestration/faq/config.md
@@ -1,0 +1,13 @@
+# Configuration Options
+
+A full list of configuation options can be seen in the Prefect [config.toml](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/config.toml). 
+
+## Running Prefect with a pre-exisiting postgres database
+If you have a postgres instance set up elsewhere then providing a connection string like this through the config var:
+
+```
+[server.database]
+connection_url = ...
+```
+
+will override the default.  You could also update the database host. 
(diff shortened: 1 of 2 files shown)


docs/orchestration/faq/config.md
```diff
@@ -0,0 +1,30 @@
...
+
+As with all configuration options, you can update these using environment variables as well:
+
+```
+PREFECT__SERVER__HOST
+PREFECT__SERVER__PORT
+```
+
+## Running Prefect with a pre-exisiting postgres database
+If you are running Prefect Server and have a postgres instance set up elsewhere then providing a `server.database.connection_url` or `server.database.host` will allow you to connect to it:
```
joshmeek, line 24: Do you think you could include some information here about the `connection_url` format where it also takes in port, username, password, db_name, etc.

zhen0:
Thanks for the feedback @joshmeek - I've added that in.

joshmeek (approved):

PR #3420, opened 2020-10-01
Add option to template task run name at runtime when using backend API

### LINKED ISSUE #2100: Option to include item string in mapped task name

joshmeek:
## Summary
This PR supersedes #2974
Closes #2100 

With the recent introduction of the `set_task_run_name` mutation in the GraphQL API we can now use this to set the name of task runs dynamically at run time. This means that a code snippet like this:

```python
@task
def get_values():
    return ['a', 's', 'd', 'f']

@task(log_stdout=True, task_run_name=lambda **kwargs: f"{kwargs['v']}")
def p(v):
    print(v)

with Flow("task_run_names") as flow:
    vals = get_values()
    p.map(vals)
```

Will automatically name all of the mapped tasks based on their inputs `a`, `s`, `d`, `f` instead of the index!

image

## Changes
Adds a new kwarg called `task_run_name` to the Task class that can now take in a string or callable to update the task run name at runtime.

## Importance
Lots of users have requested this!

## Checklist

This PR:

### DIFF
--- a/changes/issue2100.yaml
+++ b/changes/issue2100.yaml
@@ -1,2 +1,4 @@
 enhancement:
   - "Add option to template task run name at runtime when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+  - "Add `set_task_run_name` Client function - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+  - "Update mypy typing for `target` kwarg on base Task class - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
--- a/changes/issue2100.yaml
+++ b/changes/issue2100.yaml
@@ -1,4 +1,6 @@
 enhancement:
   - "Add option to template task run name at runtime when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
   - "Add `set_task_run_name` Client function - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
-  - "Update mypy typing for `target` kwarg on base Task class - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+
+fix:
+  - "Fix mypy typing for `target` kwarg on base Task class - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
--- a/changes/issue2100.yaml
+++ b/changes/issue2100.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Allow for templatable task run names when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
--- a/changes/issue2100.yaml
+++ b/changes/issue2100.yaml
@@ -1,2 +1,2 @@
 enhancement:
-  - "Allow for templatable task run names when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+  - "Add option to template task run name at runtime when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -12,6 +12,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Union,
 )
 
 import prefect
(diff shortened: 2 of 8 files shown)


changes/issue2100.yaml
```diff
@@ -0,0 +1,4 @@
+enhancement:
+  - "Add option to template task run name at runtime when using backend API - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+  - "Add `set_task_run_name` Client function - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
+  - "Update mypy typing for `target` kwarg on base Task class - [#2100](https://github.com/PrefectHQ/prefect/issues/2100)"
```
cicdw, line 4: "enhancement"

src/prefect/core/task.py
```diff
@@ -174,7 +175,7 @@ def run(self, x, y):
             is provided, it should have signature `callable(**kwargs) -> str` and at write
             time all formatting kwargs will be passed and a fully formatted location is
             expected as the return value.  Can be used for string formatting logic that
-            `.format(**kwargs)` doesn't support
+            `.format(**kwargs)` supports.
```
cicdw, line 178: I think the intent of this docstring was to say, if you can't do it within `.format()`, you can pass a callable instead

src/prefect/core/task.py
```diff
@@ -188,6 +189,13 @@ def run(self, x, y):
             Task enters a failure state
         - log_stdout (bool, optional): Toggle whether or not to send stdout messages to
             the Prefect logger. Defaults to `False`.
+        - task_run_name (Union[str, Callable], optional): a name to set for this task at runtime.
+            `task_run_name` strings can be templated formatting strings which will be
+            formatted at runtime with values from `prefect.context`. If a callable function
```
cicdw, line 194: also task inputs and parameters!

src/prefect/core/task.py
```diff
@@ -188,6 +189,13 @@ def run(self, x, y):
...
+            is provided, it should have signature `callable(**kwargs) -> str` and at write
+            time all formatting kwargs will be passed and a fully formatted location is
+            expected as the return value. Can be used for string formatting logic that
+            `.format(**kwargs)` supports. Note: this only works for tasks running against a backend API.
```
cicdw, line 198: doesn't support, or we can come up with better wording here

src/prefect/core/task.py
```diff
@@ -216,7 +224,8 @@ def __init__(
         on_failure: Callable = None,
         log_stdout: bool = False,
         result: "Result" = None,
-        target: str = None,
+        target: Union[str, Callable] = None,
```
cicdw, line 227: ENHANCE

cicdw (changes requested):
minor docstring suggestions

src/prefect/core/task.py
joshmeek: Ahhh

cicdw (approved):
LGTM!  Can't wait to see this in action

PR #3429, opened 2020-10-05
Added FROM to explicitly chain exceptions in src/prefect/utilities

gabrielcalderon:
## Summary
* Added from on each raised exception for `src/prefect/utilities`.
* This PR is part of Hacktoberfest.

## Changes
* Changes based on https://github.com/PrefectHQ/prefect/issues/3306 for `src/prefect/utilities` .

## Checklist

This PR:


gabrielcalderon:
It seems like this simple change will break the tests?

src/prefect/utilities/executors.py
```diff
@@ -261,11 +261,11 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
             except RecursiveCall as exc:
                 try:
                     call_func = getattr(exc.func, "__wrapped_func__")
-                except AttributeError:
+                except AttributeError as exc:
                     raise RecursionError(
                         "function has not been wrapped to provide tail recursion (func={})".format(
                             exc.func
-                        )
+                        ) from exc
```
joshmeek, line 268: I think this needs to be moved down one parenthesis

gabrielcalderon: Ahhh! I see it. Let me fix that quick.

joshmeek:
Also @gabrielcalderon one more, looks like you'll need to run black formatting 🙂

gabrielcalderon:
> Also @gabrielcalderon one more, looks like you'll need to run black formatting 🙂

Should be good now :-)

joshmeek (approved):

Issue #3285, opened 2020-09-09
Prefect runs on Prefect task library additions


jlowin:
Based on user feedback, there are a number of tasks we could add to the task library to dramatically enhance the ability of users to affect their orchestration backend while still running Prefect on Prefect.

This is a tracking issue for some common requests.

- [x] Rename Flow Run: to set the name of a flow run from runtime information like schedule or parameters
- [ ] Rename Task Run: to set the name of a task run from runtime information like schedul, parameters, or inputs
- [ ] Schedule Flow Run: to schedule a new run of a flow at an arbitrary time
- [x] Run Flow - execute a new flow run, optionally blocking for completion. [Task](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/prefect/flow_run.py#L11)

joshmeek:
A couple more I think would be beneficial:

- [ ] Set a flow run / task run state
- [ ] Cancel a flow run

PR #3431, opened 2020-10-05
Add FlowRunRename task 

madelgi:
## Summary
Adds a `FlowRunRename` task for renaming running flows

## Changes
Update the `prefect` task module to allow users to rename a currently running flow.

## Importance
Expands the `prefect` task module in a useful way, as specified in #3285

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/prefect/flow_run_rename.py
+++ b/src/prefect/tasks/prefect/flow_run_rename.py
@@ -1,0 +1,53 @@
+from typing import Any
+
+from prefect import Task
+from prefect.client import Client
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class FlowRunRenameTask(Task):
+    """
+    Task used to rename a running flow.
+
+    Args:
+        - flow_run_id (str, optional): The ID of the flow run to rename.
+        - flow_run_name (str, optional): The new flow run name.
+    """
+
+    def __init__(
+        self,
+        flow_run_id: str = None,
+        flow_run_name: str = None,
+        **kwargs: Any,
+    ):
+        self.flow_run_id = flow_run_id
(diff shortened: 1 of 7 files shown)


src/prefect/tasks/prefect/flow_run_rename.py
```diff
@@ -0,0 +1,54 @@
+from typing import Any
+
+from prefect import Task
+from prefect.client import Client
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class FlowRunRenameTask(Task):
```
jcrist, line 8: Would you mind changing this to `RenameFlowRunTask`? This matches our other task naming conventions, where the verb goes first (e.g. `CreateGithubPR`). The presence/absence of the `Task` suffix is a bit all over the place currently, but since this module uses `Task` suffixes that makes sense here too.

jcrist (commented):
Thanks @madelgi, overall this looks great to me! Would like to get a second +1 from @joshmeek before merge, but I'm pretty happy with this.

src/prefect/tasks/prefect/flow_run_rename.py
joshmeek: Yeah this looks great to me, I am +1 on the name change 🙂

madelgi: Should be updated! Let me know if you want me to rename the file as well, i.e., `rename_flow_run.py`. For now I left it as is.

jcrist: Nah, that should be fine. Users import from `prefect.tasks.prefect`, so the actual file name matters less.

jcrist (approved):
Will merge on tests pass. Thanks!

PR #3461, opened 2020-10-09
Add GitLab storage

madelgi:
## Summary
Adds a `Gitlab` storage option

## Changes
Adds a `Gitlab` storage option. Very similar to the `Github` storage option, though it allows two additional arguments:
- `host`: For GitLab server users, can point to a local instance of GitLab
- `ref`: Required for the python GitLab client, points to the branch or commit hash where the file is located. Defaults to `master` if not specified 

## Importance
Expands the list of git-based storage options.

## Checklist

This PR:


joshmeek:
Awesome @madelgi! I'm going to test this against a GitLab repo to verify everything works but otherwise this looks really solid 🚀

joshmeek:
@madelgi Tested and is worked flawlessly! Barring passing tests I think I am good to go on this! 🙂 

https://gitlab.com/joshmeek/prefect-flows

image

image

joshmeek:
@madelgi Could you also add this to the API reference documentation outline? https://github.com/PrefectHQ/prefect/blob/master/docs/outline.toml#L228

src/prefect/environments/storage/gitlab.py
```diff
@@ -0,0 +1,161 @@
...
+        from gitlab.exceptions import GitlabAuthenticationError, GitlabGetError
+
+        try:
+            project = self._gitlab_client.projects.get(quote_plus(self.repo))
+            contents = project.files.get(file_path=flow_location, ref=ref)
+        except GitlabAuthenticationError as exc:
+            self.logger.error(
+                "Unable to authenticate Gitlab account. Please check your credentials."
+            )
+            raise exc
```
jcrist, line 101: When re-raising an exception, you should use `raise` rather than `raise exc`, as the former doesn't add a new chained exception, while the latter does (resulting in a larger traceback).

```suggestion
            raise
```

src/prefect/environments/storage/gitlab.py
```diff
@@ -0,0 +1,161 @@
...
+        except GitlabGetError as exc:
+            self.logger.error(
+                f"Error retrieving file contents at {flow_location} in {self.repo}@{ref}. "
+                "Ensure the project and file exist."
+            )
+            raise exc
```
jcrist, line 107: Same here

```suggestion
            raise
```

jcrist (commented):
Two small nits and this looks good to me. Thanks!

jcrist (approved):

joshmeek (approved):

Issue #3037, opened 2020-07-27
DbtShellTask does not work without profiles_dir parameter


fertek:
Hi,
profiles_dir parameter is marked as optional, but without that parameter DbtShellTask fails:
`TypeError('expected str, bytes or os.PathLike object, not NoneType')`

It happens because of this line:
https://github.com/PrefectHQ/prefect/blob/6ef1f36362a46762985dfcab27a3c4ea0ec8084a/src/prefect/tasks/dbt/dbt.py#L146

Zaubeerer:
As @mhmcdonald, mentioned to me in [this Slack thread](https://prefect-community.slack.com/archives/C0106HZ1CMS/p1595937514075500?thread_ts=1592850043.041100&cid=C0106HZ1CMS):

> you need to provide a profiles_dir to let dbt know where your profiles.yml file lives. Perhaps profiles.yml located in the same directory as where your flow files are located, but also very possible it's not.

This also indicates that it should not be an optional parameter, right?

fertek:
dbt docs at https://docs.getdbt.com/dbt-cli/configure-your-profile/#using-a-custom-profile-directory says:
> By default, dbt expects your profiles.yml file to be located in the ~/.dbt/ directory.

Therefore, I think that without `profiles_dir` parameter specified the `~/.dbt/` directory should be used.

Zaubeerer:
And it should be ensured, that the `~/.dbt/` folder is created, if not existant. 
I got several `folder or file does not exist` warnings...

mhmcdonald:
@fertek @Zaubeerer It's been a while since I wrote this but I believe I left it as optional because you might have the location already set through the dbt envar `DBT_PROFILES_DIR`. If it wasn't possible to set the dir this way, I would have made it required. 

I think defaulting the profiles_dir path to ~/.dbt/ is a great idea, just as long as the DBT_PROFILES_DIR envar gets priority in determining the path location. This would be in keeping with how dbt checks for the profiles.yml file.

joshmeek:
Closed by #3462

PR #3462, opened 2020-10-09
Fix/adding default behavior to dbt task

zmac12:
## Summary
Addresses #3037 
Adding default profile directory creation behavior similar to what dbt expects, adding a default directory if none is given.

## Changes
dbt task, where profile_dir is now explicitly optional.

## Importance
If someone wishes to invoke the DbtShellTask but is unaware that the task will not create the default directory for them, they will receive an error message they will have not expected.

## Checklist

This PR:


src/prefect/tasks/dbt/dbt.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         env: dict = None,
         environment: str = None,
         overwrite_profiles: bool = False,
-        profiles_dir: str = None,
+        profiles_dir: str = os.path.join(os.path.expanduser('~'), '.dbt'),
```
cicdw, line 71: This is a little subtle, but if a user builds this task on machine A but runs it on machine B, this directory could be different.  Because of that, I think we should keep this default as `None` and only use this default when none is set (which is what I think your current `run` method implementation handles 👍 )

src/prefect/tasks/dbt/dbt.py
```diff
@@ -143,7 +148,15 @@ def run(
                 }
             }
 
-            profile_path = os.path.join(self.profiles_dir, "profiles.yml")
+            if not self.profiles_dir:
+                try:
+                    os.mkdir(DEFAULT_PROFILES_DIR)
+                except OSError:
+                    print("Creation of directory %s has failed" % DEFAULT_PROFILES_DIR)
```
cicdw, line 155: This print statement might be better as a warning log (`self.logger.warning()`)

cicdw (changes requested):
Nice, thanks for the PR @zmac12 ! I left a few minor comments.  

Note that when you are done addressing these comments, you'll need to run the `black` formatting tool on the code from the command line:
```
pip install -U black
black . # note the period specifying the present working directory
```

zmac12:
Thanks for checking my PR and thanks for the feedback! I'll make these changes and run black. 👍

cicdw (approved):
LGTM!

PR #3463, opened 2020-10-09
Extend module storage capabilities

m1so:
## Summary

Extends the module storage capabilities and calling ergonomics to match the likes of Celery/Gunicorn, etc.
Builds on top of #3351 

## Changes

Allows nested attribute access when loading flows from the module storage. Additionally, zero argument callables / factories are supported.

## Importance

Creates a more consumer friendly interface and mimics the rest of the Python ecosystem conventions.

## Checklist

This PR:


src/prefect/utilities/storage.py
```diff
@@ -102,15 +105,37 @@ def extract_flow_from_module(module_str: str, flow_name: str = None) -> "Flow":
...
+    )
+
+    module = importlib.import_module(module_name)
 
     # if flow_name is specified, grab it from the module
     if flow_name:
-        return getattr(module, flow_name)
+        attr = attrgetter(flow_name)(module)
+
+        if callable(attr):
```
jcrist, line 125: Why support callable objects? Is this something you need? I'm fine with adding it, but would like to hear a use case beforehand.

src/prefect/utilities/storage.py
```diff
@@ -102,15 +105,37 @@ def extract_flow_from_module(module_str: str, flow_name: str = None) -> "Flow":
...
+    module_parts = module_str.split(":", 2)
+
+    if len(module_parts) == 2 and flow_name is not None:
+        raise ValueError(
+            "Provide either `module_str` without an attribute specifier or remove `flow_name`."
+        )
+
+    module_name, flow_name = (
+        module_parts if len(module_parts) == 2 else [module_str, flow_name or ""]
+    )
```
jcrist, line 117: Would you mind de-clevering this logic here, I find the one-liner a bit hard to follow. Merging this with the `if len(module_parts) == 2` logic above would make things clearer IMO.

tests/utilities/test_storage.py
```diff
@@ -89,3 +93,70 @@ def test_extract_flow_from_file_raises_on_run_register(tmpdir):
...
+        f"{module_name}:test_extract_flow_from_module.multi_level_flow"
+    )
+    multi_level_arg_flow = extract_flow_from_module(
+        module_name, "test_extract_flow_from_module.multi_level_flow"
+    )
+    callable_flow = extract_flow_from_module(
+        f"{module_name}:test_extract_flow_from_module.callable_flow"
+    )
+
+    assert (
```
jcrist, line 129: Would you mind using different `flow` objects for each of these? Right now the logic could all be bypassed and use only `module.flow` and we'd never catch it.

jcrist (commented):
Thanks @m1so, this looks pretty good. Just a few comments.

tests/utilities/test_storage.py
m1so: Alright, should be incorporated in https://github.com/PrefectHQ/prefect/pull/3463/commits/4d06d46ac6e8480868cbdd9064844eb8858a8339

Please let me know if there are any additional changes needed 👍

src/prefect/utilities/storage.py
m1so: Mostly to defer the object/flow creation and have the possibility to wrap it in a class if needed. 

I.e. not relying on import side effects and module level variable that could potentially be mutated

jcrist (approved):
LGTM, thanks!

PR #3467, opened 2020-10-09
Better error for `flow.register` in flow context

jcrist:
Previously this would raise an error due to pickle failing. We could
have fixed it so pickle worked, but I'd argue that the flow context
should always exit before registering the flow (we do correctness checks
for the user on flow context exit). We now raise a nicer error message.

Reported via slack: https://prefect-community.slack.com/archives/C014Z8DPDSR/p1602257752136900

### DIFF
--- a/changes/pr3467.yaml
+++ b/changes/pr3467.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Raise a better error when calling `flow.register()` from within a `Flow` context - [#3467](https://github.com/PrefectHQ/prefect/pull/3467)"
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -357,7 +357,6 @@ class Flow:
         with prefect.context(flow=self, _unused_task_tracker=unused_task_tracker):
             yield self
 
-        # constants are not tracked at the flow level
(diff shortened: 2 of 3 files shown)


jcrist:
New error message:

```python
In [1]: from prefect import Flow

In [2]: with Flow("test") as flow:
   ...:     flow.register()
   ...:
ValueError                                Traceback (most recent call last)
<ipython-input-2-69d428f4efc3> in <module>
      1 with Flow("test") as flow:
----> 2     flow.register()
      3

~/Code/prefect/src/prefect/core/flow.py in register(self, project_name, build, labels, set_schedule_active, version_group_id, no_url, **kwargs)
   1583         """
   1584         if hasattr(self, "_ctx"):
-> 1585             raise ValueError(
   1586                 "Don't call `flow.register()` from within a `Flow` context manager.\n\n"
   1587                 "Do:\n\n"

ValueError: Don't call `flow.register()` from within a `Flow` context manager.

Do:

  with Flow(...) as flow:
      ...
  flow.register(...)

Don't:

  with Flow(...) as flow:
      ...
      flow.register(...)
```

joshmeek (approved):
Love the error message, very informative 👌

PR #3474, opened 2020-10-09
Don't use `signal.raise` when cancelling tasks

### LINKED ISSUE #3424: Cancelled flow run still has running task runs
## Description

This may be specific for the weird task and flow I use for testing a lot of things in the UI but wanted to raise in case it's not. (Especially because it has implications for restart from cancelled in the UI).  
I have a task that involves time.sleep() and sleeps for 10+ minutes.  If I cancel a flow run that includes that task, the flow run enters a cancelled state while the task run is still running.  Eventually the task run is picked up by the zombiekiller as it has no heartbeat and fails. 

## Expected Behavior

Expect that the task run would enter a finished state (cancelled or the state from that run) before (or at least at the same time as) the flow run does. 

## Reproduction

1. Register the following flow: 
```
import time

@task
def sleep():
    time.sleep(750)

with Flow(name='simple-sleep') as flow:
    sleep()

flow.register('PROJECTNAME')
```

2. Run the flow 
3. Cancel it (using the UI or the cancel_flow_run mutation).

jcrist:
This doesn't actually interrupt long running calls like `time.sleep`,
better to use `signal.pthread_kill` everywhere. Long running system
blocking calls like `time.sleep` are unlikely, but still possible,
especially for users trying things out.

Fixes #3424 (partially).

### DIFF
--- a/changes/pr3474.yaml
+++ b/changes/pr3474.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix task cancellation on Python 3.8 to properly interrupt long blocking calls - [#3474](https://github.com/PrefectHQ/prefect/pull/3474)"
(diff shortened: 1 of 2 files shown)


cicdw (approved):
Nice!

PR #3475, opened 2020-10-10
Informative error

cicdw:
## Summary

When helping a user debug, I realized that the new import path logic for Local Storage can cause misleading errors whenever a user uses pickle-based storage + the file doesn't exist in the Agent's environment.  In particular, our storage logic first checks if the location is a known path, and if not, it tries to treat the location as an import string.  This causes the ultimate error to be a `ModuleNotFound` error, which looks confusing.

This is a minimal change to add a helpful log, but I'm open to other ideas for how to avoid the spurious import attempt!

## Changes

Adds a log + traceback that explicitly says what the issue was, and re-raises the error for the Agent CLI command to capture.

## Importance

To help users self debug.

## Checklist

This PR:

### DIFF
--- a/src/prefect/environments/storage/local.py
+++ b/src/prefect/environments/storage/local.py
@@ -104,9 +104,9 @@ class Local(Storage):
             # otherwise the path is given in the module format
             else:
                 return extract_flow_from_module(module_str=flow_location)
-        except Exception as exc:
+        except Exception:
             self.logger.exception(f"Failed to load Flow from {flow_location}")
-            raise exc
+            raise
 
     def add_flow(self, flow: "Flow") -> str:
         """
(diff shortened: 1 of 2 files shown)


src/prefect/environments/storage/local.py
```diff
@@ -95,14 +95,18 @@ def get_flow(self, flow_location: str = None) -> "Flow":
...
+            # otherwise the path is given in the module format
             else:
-                return prefect.core.flow.Flow.load(flow_location)
-        # otherwise the path is given in the module format
-        else:
-            return extract_flow_from_module(module_str=flow_location)
+                return extract_flow_from_module(module_str=flow_location)
+        except Exception as exc:
+            self.logger.exception(f"Failed to load Flow from {flow_location}")
+            raise exc
```
jcrist, line 109:
```suggestion
        except Exception:
            self.logger.exception(f"Failed to load Flow from {flow_location}")
            raise
```

Small tweak to use `raise` not `raise exc` (which will add a chained exception). Slightly cleaner tracebacks.

jcrist (commented):
Small comment, otherwise this looks good to me.

jcrist (approved):

Issue #3280, opened 2020-09-08
DBTShellTask dbt_kwargs


zhen0:
## Description

From community slack: https://prefect-community.slack.com/archives/CL09KU1K7/p1599551263372600

From that slack comment: 

__I am trying to run a DbtShellTask with a dynamic set of dbt_kwargs that are defined at run time. The task works when I define the dbt_kwargs in the task initialization, but not when they are provided at runtime._

Looks like the dbt_kwargs needs an empty dictionary at initialization to be modified at runtime. 

_The documentation or the task default arguments should be changed.
Either change the documentation to say that it requires an empty dict at initialization but that this dict can be modified at runtime, or change the default arguments to not be None so that the dictionary modification I mentioned above works as expected.
Thanks so much!__

## Expected Behavior

Ideally, you don't need to provide an empty dict at initialization.  If not that, the docs define what is need at initialization correctly.

nathaniel-md:
Which do people think sounds better from a design perspective? Either changing the docs, or the default arguments, or the line [here](https://github.com/PrefectHQ/prefect/blob/24cb38fe661baa5b91f1506a74709eff60a334cc/src/prefect/tasks/dbt/dbt.py#L136)... I tend to think that it would be okay to initialize the `dbt_kwargs` argument to be an empty dict instead of `None`...

zmac12:
With the PR merged to resolve this issue, is there anything I need to do to get it closed out?

joshmeek:
Nope @zmac12 it should be good to close!

PR #3477, opened 2020-10-10
fixing default DbtKwargs arg to default to empty dict

zmac12:
## Summary
Documentation did not specify that an empty dictionary was required for DbtKwargs to function properly in the dbt task. This change creates an empty dict when none is given.

## Changes
Default behavior in dbt task as requested in #3280 

## Importance
Improves user experience with dbt task.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/dbt/dbt.py
+++ b/src/prefect/tasks/dbt/dbt.py
@@ -70,7 +70,7 @@ class DbtShellTask(ShellTask):
         overwrite_profiles: bool = False,
         profiles_dir: str = None,
         set_profiles_envar: bool = True,
-        dbt_kwargs: dict = {},
+        dbt_kwargs: dict = None,
         helper_script: str = None,
         shell: str = "bash",
         return_all: bool = False,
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/dbt/dbt.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         overwrite_profiles: bool = False,
         profiles_dir: str = None,
         set_profiles_envar: bool = True,
-        dbt_kwargs: dict = None,
+        dbt_kwargs: dict = {},
```
cicdw, line 73: Mutable defaults in Python are [generally not a good idea](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments); if this kwarg should default to an empty dictionary, the place to do it would be in line 86 below:
```python
self.dbt_kwargs = dbt_kwargs or {}
```

cicdw (changes requested):
Request to change where the default is created

cicdw (approved):
Thanks for another useful PR @zmac12 !

Issue #3200, opened 2020-08-21
Update note about storage requirements in orchestration tutorial


shaunc:
Hello!

I've found an issue here:`In order to deploy your flow to Kubernetes the flow is required to have a Docker Storage with the image pushed to a registry that your Kubernetes cluster has access to.`

Page reference: https://docs.prefect.io/orchestration/tutorial/k8s.html

Is it in fact true that a flow deployed to Kubernetes EKS can't use s3 storage? If so, an explicit affirmation of this limitation would be helpful, as it is a bit counterintuitive: `S3 storage is not supported even for Kubernetes clusters running on AWS.` Otherwise, the documentation should be amended.

joshmeek:
The orchestration tutorial has been updated to account for the new run configuration pattern https://docs.prefect.io/orchestration/tutorial/flow_config.html#flow-configuration

PR #3478, opened 2020-10-10
Updating K8s docs

zmac12:
## Summary
By default, AWS deployments of K8s do not have access to S3 for persistent storage unless permissions are given. Updated the K8s docs to describe that behavior.

## Changes
Scaling out with Kubernetes docs. Addresses #3200 

## Importance
Users getting started with K8s on AWS might not understand the need for persistent storage and this update to the docs may simplify their getting started process.

## Checklist

This PR:


jcrist (commented):
Thanks for the PR @zmac12. This feels like the wrong place to put this info (not all users use AWS, and the paragraph you edited was talking about docker image registry access, not S3). Looking through our docs, I think the proper place would be in the k8s agent docs. Could you add a new section on permissions before the `labels` section here: https://docs.prefect.io/orchestration/agents/kubernetes.html#labels?

zmac12:
@jcrist Thanks for reviewing my PR! Let me know if this location is preferable or not and I can change it if needs be.

docs/orchestration/agents/kubernetes.md
```diff
@@ -163,6 +163,10 @@ prefect-agent-845798bb59-s7wxg   1/1     Running   0          5s
 
 You are now ready to run some flows!
 
+#### Permissions
+
+In AWS deployments using Amazon Elastic Kubernetes Service, S3 is not accessible by default to Amazon EKS. To enable S3 access by your kubernetes cluster on EKS, add the necessary permissions (AmazonS3FullAccess or AmazonS3ReadOnlyAccess) directly to the NodeInstanceRole used by aws-auth-cm.yaml after launching worker nodes and before applying aws-auth-cm.yaml with kubectl.
```
jcrist, line 168: Can you add something in here to qualify that S3 isn't required? Something like "If you need S3 access, note that..." or something else. Not all users will be affected by this.

jcrist (commented):
One last tweak, then I think this is good to go. Thanks!

zmac12:
@jcrist Thanks for reviewing that again! I just submitted the changes you requested, let me know if it looks good.

jcrist (approved):
Thanks!

PR #3481, opened 2020-10-12
Add GetAttr task

m1so:
## Summary

Helper task that retrieves a specific attribute of an upstream task's result.

## Changes

-

## Importance

Allows one to extract deeply nested attributes of objects instead of passing potentially large structures (such as configuration) to downstream tasks.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/core/operators.py
+++ b/src/prefect/tasks/core/operators.py
@@ -5,7 +5,7 @@ indexing, and logical comparisons.
 In general, users will not instantiate these tasks by hand; they will automatically be
 applied when users apply inline Python operators to a task and another value.
 """
-
+from operator import attrgetter
 from typing import Any
 
 from prefect import Task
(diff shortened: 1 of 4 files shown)


src/prefect/tasks/core/operators.py
```diff
@@ -32,6 +32,27 @@ def run(self, task_result: Any, key: Any) -> Any:  # type: ignore
...
+
+    Args:
+        - *args (Any): positional arguments for the `Task` class
+        - **kwargs (Any): keyword arguments for the `Task` class
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+
+    def run(self, task_result: Any, item: Any) -> Any:  # type: ignore
```
jcrist, line 47: Mind renaming `item` to `attr`?

src/prefect/tasks/core/operators.py
```diff
@@ -32,6 +32,27 @@ def run(self, task_result: Any, key: Any) -> Any:  # type: ignore
...
+        """
+        Args:
+            - task_result (Any): a value
+            - item (Any): the attribute to retrieve as `task_result.attr`
```
jcrist, line 51: Would be good to also note that nested parameter access also works (e.g. `attr="a.b.c.d"`).

jcrist (commented):
Thanks @m1so, overall this looks good to me, just a few small fixes. You'll also need to add this to the docs config in `docs/outline.toml`.

jcrist (approved):

PR #3482, opened 2020-10-12
Enable upload for file based storage when using S3 and GCS

joshmeek:
## Summary
It is the expected behavior for some users that when they call

```bash
prefect register flow -f my_flow.py -p Project
```

That this will also upload their flow to their desired storage location in S3 and GCS. While this was previously true with pickle-based storage we did not perform the automatic upload for these cloud storage options and this PR adds that.

## Changes
Adds a `script_path` kwarg to the S3 and GCS storage classes to be used in tandem with `store_as_script`. The value can either be set manually in a script:

```python
flow.storage = S3(
    bucket="my-flow-bucket",
    stored_as_script=True,
    script_path="my_flow.py"  # Local file that you want uploaded to the bucket
)
```

or it will pull the value of the flow file when calling the `register flow` CLI command from context.

## Importance
Enables some expected behavior for these classes.

## Checklist

This PR:

### DIFF
--- a/docs/core/idioms/file-based.md
+++ b/docs/core/idioms/file-based.md
@@ -162,7 +162,7 @@ not set then a key will be automatically generated for the storage of the flow)
 flow.storage = S3(
     bucket="my-flow-bucket",
     stored_as_script=True,
-    script_path="my_flow.py"  # Local file that you want uploaded to the bucket
+    local_script_path="my_flow.py"  # Local file that you want uploaded to the bucket
 )
 
 # or
--- a/docs/core/idioms/file-based.md
+++ b/docs/core/idioms/file-based.md
@@ -170,7 +170,7 @@ flow.storage = S3(
 flow.storage = GCS(
     bucket="my-flow-bucket",
     stored_as_script=True,
-    script_path="my_flow.py"  # Local file that you want uploaded to the bucket
+    local_script_path="my_flow.py"  # Local file that you want uploaded to the bucket
 )
 ```
 
--- a/docs/core/idioms/file-based.md
+++ b/docs/core/idioms/file-based.md
@@ -22,7 +22,6 @@ pip install 'prefect[github]'
 ```
 :::
 
-
 In this example we will walk through a potential workflow you may use when registering flows with
 [GitHub](/api/latest/environments/storage.html#github) storage. This example takes place in a GitHub
 repository with the following structure:
@@ -81,7 +80,7 @@ Now that the file exists on the repo the flow needs to be registered with a Pref
 Core's server or Prefect Cloud).
 
 ```bash
-prefect register flow -f flows/my_flow.py
+prefect register flow -f flows/my_flow.py -p MyProject
 Result check: OK
 Flow: http://localhost:8080/flow/9f5f7bea-186e-44d1-a746-417239663614
 ```
@@ -95,7 +94,7 @@ If you change any of the structure of your flow such as task names, rearrange ta
 will need to reregister that flow.
 :::
 
-::: tip GitLab users 
+::: tip GitLab users
 This example applies to GitLab as well. To use GitLab storage, install the `gitlab` extra:
 
 ```bash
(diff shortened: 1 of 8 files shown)


src/prefect/environments/storage/base.py
```diff
@@ -29,6 +29,10 @@ class Storage(metaclass=ABCMeta):
             configuration at `flows.defaults.storage.add_default_labels`.
         - stored_as_script (bool, optional): boolean for specifying if the flow has been stored
             as a `.py` file. Defaults to `False`
+        - script_path (str, optional): the path to a local script to upload when `stored_as_scipt` is set
```
jcrist, line 32:
```suggestion
        - script_path (str, optional): the path to a local script to upload when `stored_as_script` is set
```

src/prefect/environments/storage/gcs.py
```diff
@@ -36,6 +36,10 @@ class GCS(Storage):
             if not provided, the project will be inferred from your Google Cloud credentials.
         - stored_as_script (bool, optional): boolean for specifying if the flow has been stored
             as a `.py` file. Defaults to `False`
+        - script_path (str, optional): the path to a local script to upload when `stored_as_scipt` is set
```
jcrist, line 39:
```suggestion
        - script_path (str, optional): the path to a local script to upload when `stored_as_script` is set
```

src/prefect/environments/storage/s3.py
```diff
@@ -31,6 +31,10 @@ class S3(Storage):
             is only useful when storing a single Flow using this storage object.
         - stored_as_script (bool, optional): boolean for specifying if the flow has been stored
             as a `.py` file. Defaults to `False`
+        - script_path (str, optional): the path to a local script to upload when `stored_as_scipt` is set
```
jcrist, line 34:
```suggestion
        - script_path (str, optional): the path to a local script to upload when `stored_as_script` is set
```

src/prefect/environments/storage/base.py
```diff
@@ -38,10 +42,13 @@ def __init__(
         labels: List[str] = None,
         add_default_labels: bool = None,
         stored_as_script: bool = False,
+        script_path: str = None,
```
jcrist, line 45: Since this doesn't make sense for all storage types, perhaps this shouldn't be added to the base class? AFAICT it only makes sense for `S3` and `GCS` storage (currently).

jcrist: Also wondering if `local_script_path` would be help disambiguate from a path in remote storage? No strong thoughts.

joshmeek: Good call yeah I'll move it to those classes only and change the name. Was originally thinking about doing something like that instead

docs/core/idioms/file-based.md
```diff
@@ -130,3 +130,53 @@ flow.storage = Docker(
...
+Storing flows this way is similar to the git-based flow storage where users manually have to upload the
+flows to the buckets and then set a key to match. There is another option where flows scripts can be
+automatically uploaded to the buckets by providing a file path to the storage object. (Note: if `key` is
+not set then a key will be automatically generated for the storage of the flow)
+
+```python{4,12}
+flow.storage = S3(
+    bucket="my-flow-bucket",
+    stored_as_script=True,
+    script_path="my_flow.py"  # Local file that you want uploaded to the bucket
```
jcrist, line 165:
```suggestion
    local_script_path="my_flow.py"  # Local file that you want uploaded to the bucket
```

docs/core/idioms/file-based.md
```diff
@@ -130,3 +130,53 @@ flow.storage = Docker(
...
+)
+
+# or
+
+flow.storage = GCS(
+    bucket="my-flow-bucket",
+    stored_as_script=True,
+    script_path="my_flow.py"  # Local file that you want uploaded to the bucket
```
jcrist, line 173:
```suggestion
    local_script_path="my_flow.py"  # Local file that you want uploaded to the bucket
```

jcrist (changes requested):

jcrist (approved):

PR #3466, opened 2020-10-09
Parameter default value's allowed types

raphael-riel:
## Summary
Parameter default value must be JSON-Serializable.
I tried to pass a `Decimal` object and it failed at registration.
Type hint had mislead me 😄 

## Changes
- Strict Type hints for the default value.
- Test for Failures (Non-JSON-Serializable values) and Success (Native values)
- Adapted doc for the Arg.

## Importance
Helping developers avoid errors is crucial for adoption!
(Yet, this PR is not that important 😝  )

## Checklist

This PR:

- [x] adds new tests (if appropriate)
- [x] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #3489, opened 2020-10-13
Add defaults for GetItem and GetAttr

m1so:
## Summary

Adds option to specify default values for GetItem and GetAttr tasks

## Changes

-

## Importance

Convenience methods so the end users don't have to handle upstream exceptions and can specify sensible defaults if desired.

## Checklist

This PR:


m1so:
@jcrist @joshmeek seems like mypy is failing due to https://github.com/PrefectHQ/prefect/pull/3466/commits/552f163ceb114e805bae73a9389413599ad3d934 which was merged, however mypy check wasn't triggered for this commit

(you could make it required in the branch settings)

jcrist:
> however mypy check wasn't triggered for this commit

Wait, why wasn't it triggered? It normally runs for every commit in a PR? Looks like none of the CircleCI tests ran on that last commit in the PR, but they did run on all commits before then.

m1so:
> > however mypy check wasn't triggered for this commit
> 
> Wait, why wasn't it triggered? It normally runs for every commit in a PR? Looks like none of the CircleCI tests ran on that last commit in the PR, but they did run on all commits before then.

I might be overlooking something, but it seems like there was only one commit for which CircleCI ran (https://app.circleci.com/pipelines/github/PrefectHQ/prefect?branch=pull%2F3466). 

It used to happen for GH Actions on some of our internal repos as well, so we had to mark those checks as required in https://github.com/PrefectHQ/prefect/settings/branches to prevent "false positive" merges.

jcrist:
Thanks for the tip, we'll definitely update that.

> I might be overlooking something, but it seems like there was only one commit for which CircleCI ran

Ooop, You're right. I can't find another PR where that's happened though, not sure why it happened here.

src/prefect/core/task.py
```diff
@@ -764,6 +779,40 @@ def serialize(self) -> Dict[str, Any]:
...
+            - item (str): the (possibly nested) attribute to retrieve as `self.attr`.
+                Nested attributes should be accessed via `.`-delimited strings.
+            - default (object): the object to use as the default.
+
+        Returns:
+            - Task
+        """
+        args = (self, item) if isinstance(default, NoDefault) else (self, item, default)
+
+        return prefect.tasks.core.operators.GetAttr().bind(*args)
```
jcrist, line 814: We already support `__getitem__` on a task for creating `GetItem` tasks. In a different issue automatic support for `GetAttr` was discussed but shot down, as typos in task attributes would silently create new tasks rather than erroring.

I don't think we want to add these accessor methods here, as I don't see this as a common pattern (prefect tasks generally aren't worth it for something as quick as a getitem/getattr).

src/prefect/core/task.py
```diff
@@ -34,6 +34,21 @@
...
+
+
+class SerializableNoDefault(NoDefault):
+    __slots__ = ()
+
+    def __repr__(self) -> str:
+        return "no_default"
+
+    def __reduce__(self) -> str:
+        return "no_default"
```
jcrist, line 49: Apologies, looks like we both tried to fix the mypy issue (fixed in #3490), so you now have a few merge conflicts.

Should be able to reuse the `NoDefault` class from `prefect.core.parameter` in your changes in `prefect.tasks.core.operators`, or just duplicate the code over there for a new sentinel. IMO there's no harm in duplicating that pattern in a few places - it's not user facing and there's a mild mental cost to indirection.

src/prefect/core/task.py
m1so: Understandable, although the proposed implementation doesn't use magic methods at all, rather a clearly defined api. The only reason I've added these two accessors was for convenience when wanting to define defaults if the given key doesn't exist. Although I get that tasks aren't worth creating for attribute access, but also sharing and ser/de large objects isn't desirable either when reusing the intermediate result in various parts of the workflow.

I can remove these and move them to my BaseTask that wraps Prefect's Task.

jcrist (approved):

Issue #2634, opened 2020-05-21
Flow not starting or throw an error after a while [multiprocessing, cloud]


gbreeze:
## Code

``` python
import prefect
import time
from prefect import task, Flow
from datetime import timedelta
from prefect.schedules import IntervalSchedule

from multiprocessing import Pool


def long_task(x):
    print("long_task")
    time.sleep(10)
    print("long_task done")


@task
def hello_task():
    logger = prefect.context.get("logger")
    logger.info("Starting Pools...")
    thread_pools = Pool()

    thread_pools.map(long_task, range(4))

    thread_pools.close()
    thread_pools.join()
    logger.info("all tasks done")


if __name__ == "__main__":
    if True:
        schedule = IntervalSchedule(interval=timedelta(minutes=2))
        flow = Flow("hello-flow", schedule, tasks=[hello_task])

        flow.register(project_name="test")
        flow.run_agent(
            token="..."
        )
    else:
        flow = Flow("hello-flow", tasks=[hello_task])
        flow.run()

```

## Exception (in cloud logs)

``` error
Unexpected error: PicklingError("Can't pickle <function long_task at 0x0000019E18510A68>: attribute lookup long_task on __main__ failed")
Traceback (most recent call last):
  File "c:\users\mail\appdata\local\programs\python\python37\lib\site-packages\prefect\engine\runner.py", line 48, in inner
    new_state = method(self, state, *args, **kwargs)
  File "c:\users\mail\appdata\local\programs\python\python37\lib\site-packages\prefect\engine\task_runner.py", line 943, in get_task_run_state
    self.task.run, timeout=self.task.timeout, **raw_inputs
  File "c:\users\mail\appdata\local\programs\python\python37\lib\site-packages\prefect\utilities\executors.py", line 182, in timeout_handler
    return fn(*args, **kwargs)
  File "e:/Temp/prefect/test1/task1.py", line 22, in hello_task
    thread_pools.map(long_task, range(4))
  File "c:\users\mail\appdata\local\programs\python\python37\lib\multiprocessing\pool.py", line 268, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "c:\users\mail\appdata\local\programs\python\python37\lib\multiprocessing\pool.py", line 657, in get
    raise self._value
  File "c:\users\mail\appdata\local\programs\python\python37\lib\multiprocessing\pool.py", line 431, in _handle_tasks
    put(task)
[...]

PR #3511, opened 2020-10-15
Fix multiprocessing timeout

madkinsz:
## Summary

The value retrieval in the `multiprocessing_timeout` was using a locally defined helper function
which could not be serialized by the native Python pickler resulting in an error running tasks with
a timeout while using a process based scheduler.

#### Example
```python
from random import randint

@task
def foo(x):
    return [randint(0, i) for i in range(x)]

@task(timeout=10)
def bar(y):
    return y * 2

with Flow("foobar") as flow:
    y = foo(10)
    z = bar.map(y) 

if __name__ == "__main__":
    state = flow.run(executor=LocalDaskExecutor())
```

Results in the error `AttributeError: Can't pickle local object 'multiprocessing_timeout.<locals>.retrieve_value'`

Also fixes a timeout issue with function pickling similar to https://github.com/PrefectHQ/prefect/issues/2634

## Changes

- Moves the `multiprocessing_timeout.retrieve_value` function to a top-level function
- Renames `retrieve_value` to `multiprocessing_safe_retrieve_value`
- `multiprocessing_safe_retrieve_value` takes a cloudpickle serialized callable so tasks defined in scripts can be passed
- Adds the `mproc` and `mproc_local` executors to the flow timeout test
- Minor documentation improvements in the `multiprocessing_timeout` function
- Adds a `__processes` attribute to distributed test executors
- Adds a local multiprocess test executor
- Fixes the synchronous test executor to use the correct scheduler
- Adds test coverage for daemon process / soft limit timeout enforcement

## Importance

- Fixes a bug in multiprocessing timeouts
- Improves user experience when defining timeout tasks in scripts

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -115,7 +115,7 @@ def multiprocessing_safe_retrieve_value(
     _container: multiprocessing.Queue,
     _ctx_dict: dict,
     _fn_pkl: bytes,
-    **kwargs: Any
+    **kwargs: Any,
 ) -> None:
     """
     Gets the return value from a function and puts it in a multiprocessing-safe container
@@ -238,19 +238,16 @@ def timeout_handler(
         elif multiprocessing.current_process().daemon is False:
             return multiprocessing_timeout(fn, *args, timeout=timeout, **kwargs)
 
-        msg = (
-            "This task is running in a daemonic subprocess; "
-            "consequently Prefect can only enforce a soft timeout limit, i.e., "
-            "if your Task reaches its timeout limit it will enter a TimedOut state "
-            "but continue running in the background."
-        )
+        soft_timeout_reason = "in a daemonic subprocess"
     else:
-        msg = (
-            "This task is running on Windows; "
-            "consequently Prefect can only enforce a soft timeout limit, i.e., "
-            "if your Task reaches its timeout limit it will enter a TimedOut state "
-            "but continue running in the background."
-        )
+        soft_timeout_reason = "on Windows"
+
+    msg = (
+        f"This task is running {soft_timeout_reason}; "
+        "consequently Prefect can only enforce a soft timeout limit, i.e., "
+        "if your Task reaches its timeout limit it will enter a TimedOut state "
+        "but continue running in the background."
+    )
 
     warnings.warn(msg, stacklevel=2)
     executor = ThreadPoolExecutor()
@@ -266,7 +263,10 @@ def timeout_handler(
     try:
         return fut.result(timeout=timeout)
     except FutureTimeout as exc:
-        raise TimeoutError("Execution timed out.") from exc
+        raise TimeoutError(
+            f"Execution timed out but was executed {soft_timeout_reason} and will "
+            "continue to run in the background."
+        ) from exc
 
 
 class RecursiveCall(Exception):
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -10,7 +10,7 @@ import warnings
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import TimeoutError as FutureTimeout
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Union, Sequence
 
 import cloudpickle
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -51,27 +56,40 @@ def local():
...
+    "Multi-processing executor using dask distributed"
     with Client(
         processes=True, scheduler_port=0, dashboard_address=":0", n_workers=2
     ) as client:
-        yield DaskExecutor(client.scheduler.address)
+        executor = DaskExecutor(client.scheduler.address)
+        # Since the cluster can't be inspected until the client is started we can't
+        # know if processes are being used before `flow.run` is called so we patch
+        # the class with an indicator
+        executor.__processes = True
```
madkinsz, line 78: I don't love attaching an attribute to the executor but don't see another way to grab this config value during the test

madkinsz:
The referenced pickling error does not appear to occur on Linux just OSX 🤷

jcrist:
Nice find, I'll give this a longer review after lunch.

> The referenced pickling error does not appear to occur on Linux just OSX 🤷

In python 3.8 osx switched to using spawn by default, while linux still uses fork (where pickling won't be needed).

src/prefect/utilities/executors.py
```diff
@@ -108,20 +110,61 @@ def error_handler(signum, frame):  # type: ignore
...
+        Callers are expected to re-raise any exceptions.
+    """
+    try:
+        with prefect.context(_ctx_dict):
+            fn = cloudpickle.loads(_fn_pkl)
+            # We could check if the `fn` is callable here but it will get caught by
+            # the try-except anyway. If the given error message is confusing, we
+            # can do an explicit check and provide more details
+            val = fn(*args, **kwargs)
+        _container.put(val)
```
jcrist, line 146: We should also use `cloudpickle.dumps` on the result (or the exception), as not all results may be pickleable either.

tests/core/test_flow.py
```diff
@@ -2960,13 +2978,18 @@ def slow_fn():
         assert not os.path.exists(FILE)
 
         start_time = time.time()
-        state = flow.run(executor=executor)
+        with assert_daemon_warning if daemon_process else nullcontext():
```
madkinsz, line 2981: This is failing to capture the warning -- cannot capture it using `capfd` either. It's being logged though. Hmm.

tests/core/test_flow.py
```diff
@@ -2943,6 +2948,19 @@ def test_timeout_actually_stops_execution(executor):
...
     # This will drastically reduce the brittleness of the test (but not completely).
 
+    # The amount of time to sleep before writing 'invalid' to the file
+    # lower values will decrease test time but increase chances of intermittent failure
+    SLEEP_TIME = 3
+
+    # Determine if the executor is distributed and using daemonic processes which
+    # cannot be cancelled and throw a warning instead. The `__processes` property
+    # is injected in `conftest.py`
+    daemon_process = isinstance(executor, DaskExecutor) and executor.__processes
```
jcrist, line 2958: This would be equivalent, but avoids the need to modify the executor in the tests.

```suggestion
    daemon_process = (
        isinstance(executor, DaskExecutor)
        and executor.client.scheduler.address.startswith("inproc")
    )
```

tests/conftest.py
jcrist: Yeah, I'd prefer to avoid this.

tests/core/test_flow.py
madkinsz: Ah great that's exactly what I was looking for!

tests/core/test_flow.py
madkinsz: Resolved by adding it to the Timeout error message

src/prefect/utilities/executors.py
madkinsz: Good idea! Done.

src/prefect/utilities/executors.py
```diff
@@ -108,20 +110,64 @@ def error_handler(signum, frame):  # type: ignore
         signal.alarm(0)
 
 
+def multiprocessing_safe_retrieve_value(
+    *args: Any,
+    _container: multiprocessing.Queue,
+    _ctx_dict: dict,
+    _fn_pkl: bytes,
+    **kwargs: Any,
```
jcrist, line 118: Just realized we also should cloudpickle the args/kwargs (and probably context) to be safe. Also should probably avoid conflating function arguments with the stuff prefect needs to pass around (as is done here).

What about the following signature:

```python
def multiprocessing_safe_retrieve_value(queue: multiprocessing.Queue, payload: bytes):
    request = cloudpickle.loads(payload)
    func = request["func"]
    args = request["args"]
    kwargs = request["kwargs"]
    context = request["context"]
    ...
```

- everything is sent through cloudpickle
- no mixing of user args/kwargs with stuff the timeout handler needs to function

src/prefect/utilities/executors.py
```diff
@@ -196,19 +241,16 @@ def timeout_handler(
...
-            "but continue running in the background."
-        )
+        soft_timeout_reason = "on Windows"
+
+    msg = (
+        f"This task is running {soft_timeout_reason}; "
+        "consequently Prefect can only enforce a soft timeout limit, i.e., "
+        "if your Task reaches its timeout limit it will enter a TimedOut state "
+        "but continue running in the background."
+    )
```
jcrist, line 253: Nice

jcrist (commented):
One final tweak and this looks good to me.

src/prefect/utilities/executors.py
madkinsz: That seems proper -- glad to have some encouragement for a more aggressive signature change.

jcrist (approved):

PR #3512, opened 2020-10-15
Submit flow runs based on order of scheduled start time

### LINKED ISSUE #3165: Agents should submit flow runs in order of scheduled times
Right now the agents pull a list of flow runs in [arbitrary order](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/agent/agent.py#L543) and submit them. In practice this doesn't matter because only a few runs are in the queue on every check, but when the work queue contains late work (say, because an agent wasn't running) it should be cleared in order - the oldest runs submitted first, then the newer ones.

joshmeek:
## Summary
`get_runs_in_queue` will return flow runs in order however when we do the additional query to get flow run information order is not guaranteed. This updates that function to sort the flow runs after returned from the query

Closes #3165 

## Changes
Adds `sorted` call on flow run information query in the agent deployment process, ordered by `scheduled_start_time`

## Importance
It is ideal to have work run in the order it is scheduled under circumstances (such as) where agents go down and come back online

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -21,7 +21,7 @@ from prefect.engine.state import Failed, Submitted
 from prefect.serialization import state
 from prefect.utilities.context import context
 from prefect.utilities.exceptions import AuthorizationError
-from prefect.utilities.graphql import EnumValue, GraphQLResult, with_args
+from prefect.utilities.graphql import GraphQLResult, with_args
 
 ascii_name = r"""
  ____            __           _        _                    _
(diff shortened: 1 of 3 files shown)


src/prefect/agent/agent.py
```diff
@@ -554,7 +554,10 @@ def query_flow_runs(self) -> list:
                                     },
                                 },
                             ],
-                        }
+                        },
+                        "order_by": {
```
cicdw, line 558: Hmmm couldn't we just as easily do this client-side?  E.g., make the original query w/ an added `scheduled_start_time` attribute and then do a sort on `scheduled_start_time` in this method?

cicdw (approved):

Issue #3451, opened 2020-10-07
FargateTaskEnvironment - containerDefinition's environment should be validated by sorting the list values


AlessandroLollo:
As described in this [Slack thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1602060735083800?thread_ts=1602003197.043500&cid=CL09KU1K7)
Since the `environment` of a task definition is a `list`, it should be checked using `sorted`, otherwise the validation fails even if the existing task definition and the flow task definition have the same values for `environment`

Changes should only touch the following lines of code

https://github.com/PrefectHQ/prefect/blob/f554d1313638286d4183b136f97b56bfbfae58db/src/prefect/environments/execution/fargate/fargate_task.py#L255-L256

https://github.com/PrefectHQ/prefect/blob/f554d1313638286d4183b136f97b56bfbfae58db/src/prefect/environments/execution/fargate/fargate_task.py#L268

I believe this is a **blocking** issue for all those who wants to use Prefect together with Fargate.

jcrist:
Fixed by #3452. Thanks!

PR #3514, opened 2020-10-16
Fix Fargate task definition environment and role validation

zpencerq:
## Summary
Expands and fixes upon #3451

Properly fixes previous FargateTaskEnvironment task definition validation around `Environment` by refactoring the comparisons.

Expands the validation to include ARNs as well (you can reference them by name when created  but the name gets converted to the ARN form in the taskDefinition in AWS and will clash in validation later). To play it safe, I also went through the AWS documentation and reformat the fields that could possibly be re-ordered by AWS incase users are using task definitions with more than one container.

Tests are included to verify the implementation.

### The root issue
Essentially, the data structure storing environment variables is a hashmap that isn't ordered, so each `describe_task_definition` call can potentially return the environment variables out of order, breaking the validation. Furthermore, roles can be provided as names which are saved as `ARN`s.

## Changes
Makes the task definition validation in the FargateTaskEnvironment more correcter.
* Updates ENV comparison by converting the (name, value) `list` into a `dict` (applies this change to all `object array` type fields before doing any comparison to avoid future errors)
* Adds alternative equality for `roleArns` since they can be provided as role names which get realized as fully qualified `ARN`s when saved in the ECS `taskDefinition`.

## Importance
The existing validation is incorrectly blocking flow runs 😓 
This change should actually fix the validation to work properly.

## Checklist

This PR:

### DIFF
--- a/changes/pr3514.yaml
+++ b/changes/pr3514.yaml
@@ -1,0 +1,5 @@
+fix:
+ - "Update Fargate task definition validation - [#3514](https://github.com/PrefectHQ/prefect/pull/3514)"
+
+contributor:
+  - "[Spencer Ellinor](https://github.com/zpencerq)"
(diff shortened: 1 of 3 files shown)


joshmeek (approved):
Lgtm @zpencerq! And thanks for providing tests 🙂

PR #3521, opened 2020-10-16
Catch & log errors in heartbeat thread

jcrist:
Previously an error in the heartbeat thread would silently kill the
thread. We now catch and log them, and continue looping.

This fixes a bug where a non-api error (e.g. a socket error) in the k8s
agent resource manager would kill the resource manager.

This PR:

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -218,7 +218,10 @@ class KubernetesAgent(Agent):
         """
         Check status of jobs created by this agent, delete completed jobs and failed containers.
         """
-        self.manage_jobs()
+        try:
+            self.manage_jobs()
+        except Exception:
+            self.logger.error("Error while managing existing k8s jobs", exc_info=True)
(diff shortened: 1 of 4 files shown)


src/prefect/agent/kubernetes/agent.py
```diff
@@ -218,7 +218,10 @@ def heartbeat(self) -> None:
         """
         Check status of jobs created by this agent, delete completed jobs and failed containers.
         """
-        self.manage_jobs()
+        try:
+            self.manage_jobs()
+        except Exception:
+            self.logger.error("Error while managing existing k8s jobs", exc_info=True)
```
madkinsz, line 224: I did not know about `exc_info` -- that's handy!

madkinsz (approved):

joshmeek:
@jcrist Do you want a changes entry for this?

jcrist:
Was just adding one :)

joshmeek (approved):

PR #3530, opened 2020-10-17
[client] close #3515 Introduce the property active_tenant_id which handles the tenant initialization

### LINKED ISSUE #3515: Client in Server mode makes a call to the server before having to setup the attached_headers.
## Description

In server mode the `prefect.client.Client` make a call to the server to obtain the tenant-id. However, in our case the server is protected with BasicAuth. `Client.attach_headers` is here to help us! However since this call is performed inside the `__init__` we can't pass the required headers (attached_headers have not been set yet).

https://github.com/PrefectHQ/prefect/blob/9721bb103ba85eedab863258588671e1cba8fd58/src/prefect/client/client.py#L124

```python
class Client:
    def __init__(self, api_server: str = None, api_token: str = None):
        self._access_token = None
        self._refresh_token = None
        self._access_token_expires_at = pendulum.now()
        self._active_tenant_id = None
        self._attached_headers = {}  # type: Dict[str, str]
        self.logger = create_diagnostic_logger("Diagnostics")

        # store api server
        self.api_server = api_server or prefect.context.config.cloud.get("graphql")

        # store api token
        self._api_token = api_token or prefect.context.config.cloud.get(
            "auth_token", None
        )

        if prefect.config.backend == "cloud":
          ...
        else:
            # TODO: Separate put this functionality and clean up initial tenant access handling
            if not self._active_tenant_id:
                # This call is performed without self._attached_headers set up
                tenant_info = self.graphql({"query": {"tenant": {"id"}}})
                if tenant_info.data.tenant:
                    self._active_tenant_id = tenant_info.data.tenant[0].id
```

In our case this is pretty annoying. Since that our server is protected with BasicAuth this example will raise an error at the constructor step:

```python
client = Client() # Here will have requests.exceptions.HTTPError: 401 Client Error: Unauthorized 
headers = {'Authorization': "Basic {}".format(PREFECT__SERVER__BASIC_AUTH)}
client.attach_headers(headers)
```

It will happens for all the classes depending on a client:

```python
agent = DockerAgent(name="trololo") # Here will have requests.exceptions.HTTPError: 401 Client Error: Unauthorized 
headers = {'Authorization': "Basic {}".format(PREFECT__SERVER__BASIC_AUTH)}
agent.client.attach_headers(headers)
```

## Expected Behavior

Move the initialization of `_active_tenant_id` further in order to let it enjoy our `attached_headers`.

```python
client = Client() #No call to the server here and no error.
headers = {'Authorization': "Basic {}".format(PREFECT__SERVER__BASIC_AUTH)}
client.attach_headers(headers)
```

## Environment
```json
{
  "config_overrides": {
    "flows": {
      "checkpointing": true
    }
  },
  "env_vars": [
    "PREFECT__SERVER__BASIC_AUTH"
  ],
  "system_information": {
    "platform": "Linux-4.15.0-118-generic-x86_64-with-debian-stretch-sid",
    "prefect_backend": "server",
    "prefect_version": "0.13.10",
    "python_version": "3.7.2"
  }
}
```

EmGarr:
close #3515

## Summary

Developpers using prefect server needs sometimes to attach_headers to the
queries in order to communicate with the server (e.g Basic Auth).

However, the first call to the server happened inside the constructor with the
tenant_id initialization. Since the headers couldn't been attached it results to failures.
This commit fix this issue.

Server protected with BasicAuth
```python
 # Here will have requests.exceptions.HTTPError: 401 Client Error: Unauthorized 
# It is because of the tenant initialization inside the constructor
client = Client()
headers = {'Authorization': "Basic {}".format(PREFECT__SERVER__BASIC_AUTH)}
client.attach_headers(headers)
```

After the commit will have this:

```python
client = Client() # No errors
headers = {'Authorization': "Basic {}".format(PREFECT__SERVER__BASIC_AUTH)}
client.attach_headers(headers)
client.graphql({"query": {"tenant": {"id"}}})
```

## Changes

1. ~tenant initialization is performed during the first call of the public method graphql~
2. ~graphql is split in public and private methods to avoid recursion errors~ 
1. Introduce the property `active_tenant_id` which handle the tenant initialization 

## Importance

Limit the usage of prefect server and cannot be deployed to production with BasicAuth

## Checklist

This PR:

- [x] ~adds a change file in the `changes/` directory (if appropriate)~

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -98,9 +98,10 @@ class Client:
         # store api server
         self.api_server = api_server or prefect.context.config.cloud.get("graphql")
 
-        # store api token
-        self._api_token = api_token or prefect.context.config.cloud.get(
-            "auth_token", None
+        self._api_token = (
+            api_token
+            or prefect.context.config.cloud.get("auth_token", None)
+            or self._load_local_settings().get("api_token")
         )
 
     def create_tenant(self, name: str, slug: str = None) -> str:
@@ -239,20 +240,16 @@ class Client:
         protected with auth like BasicAuth do not forget to `attach_headers` before any call.
         """
         if prefect.config.backend == "cloud":
-            if not self._api_token:
-                # if no api token was passed, attempt to load state from local storage
-                settings = self._load_local_settings()
-                self._api_token = settings.get("api_token")
-
-                if self._api_token:
-                    self._active_tenant_id = settings.get("active_tenant_id")
-                if self._active_tenant_id:
-                    try:
-                        self.login_to_tenant(tenant_id=self._active_tenant_id)
-                    except AuthorizationError:
-                        # if an authorization error is raised, then the token is invalid and should
-                        # be cleared
-                        self.logout_from_tenant()
+            # if no api token was passed, attempt to load state from local storage
+            settings = self._load_local_settings()
+            self._active_tenant_id = settings.get("active_tenant_id")
+            if self._active_tenant_id:
+                try:
+                    self.login_to_tenant(tenant_id=self._active_tenant_id)
+                except AuthorizationError:
+                    # if an authorization error is raised, then the token is invalid and should
+                    # be cleared
+                    self.logout_from_tenant()
         else:
             tenant_info = self.graphql({"query": {"tenant": {"id"}}})
             if tenant_info.data.tenant:
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -101,8 +101,11 @@ class Client:
         self._api_token = (
             api_token
             or prefect.context.config.cloud.get("auth_token", None)
-            or self._load_local_settings().get("api_token")
         )
+
+        # Initialize the tenant and api token if not yet set
+        if not self._api_token and prefect.config.backend == "cloud":
+            self._init_tenant()
 
     def create_tenant(self, name: str, slug: str = None) -> str:
         """
@@ -242,6 +245,10 @@ class Client:
         if prefect.config.backend == "cloud":
             # if no api token was passed, attempt to load state from local storage
             settings = self._load_local_settings()
+
+            if not self._api_token:
+                self._api_token = settings.get("api_token")
+
             self._active_tenant_id = settings.get("active_tenant_id")
             if self._active_tenant_id:
                 try:
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -98,9 +98,8 @@ class Client:
         # store api server
         self.api_server = api_server or prefect.context.config.cloud.get("graphql")
 
-        self._api_token = (
-            api_token
-            or prefect.context.config.cloud.get("auth_token", None)
+        self._api_token = api_token or prefect.context.config.cloud.get(
+            "auth_token", None
         )
 
         # Initialize the tenant and api token if not yet set
@@ -248,8 +247,9 @@ class Client:
 
             if not self._api_token:
                 self._api_token = settings.get("api_token")
+            if self._api_token:
+                self._active_tenant_id = settings.get("active_tenant_id")
 
-            self._active_tenant_id = settings.get("active_tenant_id")
             if self._active_tenant_id:
                 try:
                     self.login_to_tenant(tenant_id=self._active_tenant_id)
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -91,7 +91,7 @@ class Client:
         self._access_token = None
         self._refresh_token = None
         self._access_token_expires_at = pendulum.now()
-        self._active_tenant_id: Optional[str] = None
+        self._active_tenant_id = None
         self._attached_headers = {}  # type: Dict[str, str]
         self.logger = create_diagnostic_logger("Diagnostics")
 
@@ -529,7 +529,7 @@ class Client:
             - str: the access token
         """
         if not self._access_token:
-            return self._api_token
+            return self._api_token  # type: ignore
 
         expiration = self._access_token_expires_at or pendulum.now()
         if self._refresh_token and pendulum.now().add(seconds=30) > expiration:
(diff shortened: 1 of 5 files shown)


src/prefect/client/client.py
```diff
@@ -241,6 +219,36 @@ def post(
...
+                    self._active_tenant_id = settings.get("active_tenant_id")
+                if self._active_tenant_id:
+                    try:
+                        self.login_to_tenant(tenant_id=self._active_tenant_id)
+                    except AuthorizationError:
+                        # if an authorization error is raised, then the token is invalid and should
+                        # be cleared
+                        self.logout_from_tenant()
+        else:
+            # TODO: Separate put this functionality and clean up initial tenant access handling
```
madkinsz, line 247: Is this todo relevant now? Perhaps during this change `_init_tentant` should be split into its two separate modes

src/prefect/client/client.py
```diff
@@ -253,6 +261,43 @@ def graphql(
...
+        raise_on_error: bool = True,
+        headers: Dict[str, str] = None,
+        variables: Dict[str, JSONLike] = None,
+        token: str = None,
+        retry_on_api_error: bool = True,
+    ) -> GraphQLResult:
+        """
+        Convenience function for running queries against the Prefect GraphQL API
+
         Args:
```
madkinsz, line 301: I'm not sure I would replicate all of the docstring here since it would always be the same as `self.graphql` and need to be maintained to match. I'm not sure how this works with our doc enforcement though.

src/prefect/client/client.py
```diff
@@ -253,6 +261,43 @@ def graphql(
         """
         Convenience function for running queries against the Prefect GraphQL API
```
madkinsz, line 262: Perhaps with noting that we may initialize the tenant with the first request here

madkinsz (commented):
Hi! Thanks for your contribution :)

I just have a few comments, we'll have to wait for review from Jim or Josh since I'm new.

Do we want to run `Client._init_tenant` in the initializer still but wrapped in a try/catch with a warning log so we preserve the old behavior? Is the first `graphql` call initializing the tenant going to be a significant issue for anything? If so, should it be a public method?

Do we need to initialize the tenant before other calls too? What about `Client.post`?

src/prefect/client/client.py
EmGarr: I don't know the scope of it. So I'll let the decision to the maintainer  it :+1:

src/prefect/client/client.py
EmGarr: I replicated it for the doc enforcement ^^

src/prefect/client/client.py
joshmeek: This PR actually addresses this TODO so it can be removed 🙂

EmGarr:
> Do we want to run `Client._init_tenant` in the initializer still but wrapped in a try/catch with a warning log so we preserve the old behavior?

I agree. I don't all the implication of the tenant. If there are other changes I won't mind to do them.

>  Is the first `graphql` call initializing the tenant going to be a significant issue for anything? If so, should it be a public method?
> Do we need to initialize the tenant before other calls too? What about `Client.post`?

From what I read the tenant_id is only used in two methods in client.py

1. `create_project`

https://github.com/EmGarr/prefect/blob/63c0ed4aec2e652ae3f98b3a96494d83c581f910/src/prefect/client/client.py#L924

2. `register_agent`

https://github.com/EmGarr/prefect/blob/63c0ed4aec2e652ae3f98b3a96494d83c581f910/src/prefect/client/client.py#L1595

Since it is a private attribute. I assume we could move the initialization of the tenant to this part and let the graphql_method be a public one (without the private) ?

joshmeek:
@EmGarr I like the idea of changing it to a property the way you have it! Especially because we use it outside of the client. Seems like a much cleaner approach

madkinsz:
> @EmGarr I like the idea of changing it to a property the way you have it! Especially because we use it outside of the client. Seems like a much cleaner approach

I agree! This would be great and address any concerns I have

EmGarr:
@joshmeek  @madkinsz What about this commit ? If it is okay for you I'll rebase and make a clearer explanation for the commit message.

madkinsz:
> What about this commit ? If it is okay for you I'll rebase and make a clearer explanation for the commit message.

@EmGarr that looks good to me

EmGarr:
Rebase done

joshmeek:
This looks good @EmGarr! Let me test this with a few different setups to make sure the previous tenant init at client creation isn't expected anywhere (I don't think it is but want to make sure)

madkinsz:
> Rebase done

@EmGarr if you could update the title to match the new implementation that'd be great too! :)

EmGarr:
@joshmeek @madkinsz do you have any news regarding this MR ?

joshmeek:
@EmGarr in its current state this actually will fail when running against cloud. I think we should move the `if prefect.config.backend == "cloud"` part back to the init and leave the else block in the `_init_tenant`

EmGarr:
@joshmeek Sorry I had work to do! Do you have any information regarding what does not work with the cloud ?

EmGarr:
I guess this is a problem with the local agent and this line:

https://github.com/PrefectHQ/prefect/blob/99f5701e209f78366f85f39ec82b5fd0590c3c65/src/prefect/agent/local/agent.py#L241

I'll make an update.

joshmeek:
@EmGarr when communicating with Cloud as the backend the API token needs to be used for all calls in order to infer authentication. With this setup the `_init_tenant` is only being called when `active_tenant_id` is retrieved and that is only being retrieved in `create_project` and `register_agent`. Therefore when running something like `prefect get flows` this fails because the client has not yet loaded the API token

EmGarr:
@joshmeek what about this fix ?

joshmeek:
@EmGarr Works now! Looks like there's a small mypy error then this should be good

joshmeek:
@EmGarr I pushed some changes that I believe should resolve the errors I was seeing 👍

EmGarr:
@joshmeek thanks I haven't been able to work on it lately.

joshmeek (approved):

PR #3532, opened 2020-10-19
Info about storing SLACK_WEBHOOK_URL as a secret

magdzikk:
## Summary
Add information about storing SLACK_WEBHOOK_URL as a Prefect Secret for production usage. 

## Changes
Documentation change

## Importance
Currently the docs suggest that it's not possible to securely store the slack webhook url. However, I tested it and it seems that when you pass the SLACK_WEBHOOK_URL secret, the slack_notifier correctly picks this up.


jcrist (approved):
Thanks!

Issue #3531, opened 2020-10-19
Enable custom texts for notifications


Zaubeerer:
## Current behavior
Slack, email and other notifications provide a standardized text and backend info.

## Proposed behavior
Users can add custom texts to the notifications.

## Example

```
Prefect
Failed
awesome-flow is now in a Failed state
Message
Some reference tasks failed.
Prefect notification
@colleague, please analyze possible root causes.
```

## Ideas for Implementation
add kwargs to functions in notifications.py to support a "custom_text"

madkinsz:
Hi! I've implemented a rough draft of what this could look like in https://github.com/PrefectHQ/prefect/pull/3534 but I think some discussion on the correct API for this would be prudent! Please let me know if something like that would fit your use-case.

@joshmeek thoughts?

This will take additional work to add to all of the notifications and update documentation, this would be a good issue for someone from the community!

joshmeek:
@madkinsz I'm guessing we would want the template to be "templatable" so I think something similar to what we do for task [targets](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/engine/task_runner.py#L679) would be a nice implementation. Task targets can take in templatable strings like `"{task_name}-{today}"` but for the notifications we would run with a set amount of templatable variables. Looks like `tracked_obj`, `backend_info` and `state` are the reserved values that are available to the notifier at runtime. The notifier _also_ has context available so we could use variables from there.

madkinsz:
Hi all, there are some comments in the relevant PoC PR and this is ready for anyone in the community to work on.

Zaubeerer:
Hey @madkinsz, that looks already like a great framework in which to bring this functionality to live indeed!

Unfortunately, I am still pretty busy at the moment, but if nobody else looks into this, I might be able to finish it in ~10 days :)

zhen0:
I think we should consider this stale!  Closing but happy to re-open if anyone disagrees.

PR #3534, opened 2020-10-19
[PoC] Slack notification templates

madkinsz:
## Summary

Implements notification message templates for just slack notifications for API discussion.

When fully implemented, closes https://github.com/PrefectHQ/prefect/issues/3531

## Changes

- Adds `template` and `template_vars` to slack notifications

## Importance

Users want to amend notifications to include extra information but the best API for this is not entirely clear. This PoC will help us determine the best way to implement this as a feature.

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/notifications/notifications.py
+++ b/src/prefect/utilities/notifications/notifications.py
@@ -8,7 +8,7 @@ import smtplib
 from email.header import Header
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
-from typing import TYPE_CHECKING, Any, Callable, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Union, cast, Mapping
 
 from toolz import curry
 
@@ -125,7 +125,12 @@ def slack_message_formatter(
     tracked_obj: TrackedObjectType,
     state: "prefect.engine.state.State",
     backend_info: bool = True,
+    template: str = None,
+    template_vars: Mapping = None,
 ) -> dict:
+    template = template or f"{tracked_obj} is now in a {state} state"
+    template_vars = template_vars or {}
+
     # see https://api.slack.com/docs/message-attachments
     fields = []
     if isinstance(state.result, Exception):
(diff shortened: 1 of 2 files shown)


src/prefect/utilities/notifications/notifications.py
```diff
@@ -125,7 +125,12 @@ def slack_message_formatter(
     tracked_obj: TrackedObjectType,
     state: "prefect.engine.state.State",
     backend_info: bool = True,
+    template: str = None,
```
madkinsz, line 128: `None` here so the default value can still be defined in this function but also updated from `slack_notifier` -- not great but preserves the existing location of the default message.

src/prefect/utilities/notifications/notifications.py
```diff
@@ -125,7 +125,12 @@ def slack_message_formatter(
...
+    template_vars: Mapping = None,
 ) -> dict:
+    template = template or f"{tracked_obj} is now in a {state} state"
+    template_vars = template_vars or {}
```
madkinsz, line 132: We may want to update `template_vars` with `tracked_obj` and `state` here so there cannot be collisions in the `.format(...)` call

src/prefect/utilities/notifications/notifications.py
```diff
@@ -125,7 +125,12 @@ def slack_message_formatter(
     tracked_obj: TrackedObjectType,
     state: "prefect.engine.state.State",
     backend_info: bool = True,
+    template: str = None,
+    template_vars: Mapping = None,
 ) -> dict:
+    template = template or f"{tracked_obj} is now in a {state} state"
```
madkinsz, line 131: Where should we define what names will be provided by default? What if a user wants to add to the template without having to recreate this string?

madkinsz: Note to include prefect context variables

joshmeek: I think that mentioning it in the docstring will suffice. The only things we have available to pass into the format call are those two vars and context

madkinsz:
Closing this as it's only a proof of concept and I'm worried this is discouraging contribution from others.

PR #3540, opened 2020-10-20
Raise better error for invalid functions used with cloud

### LINKED ISSUE #3495: Raise a better error when using custom schedule filters with cloud
Currently custom filters on schedules aren't supported, since the filters run on cloud's servers. Right now the error message raised is a long `ValidationError` traceback that just notes that the function is "invalid" - we should provide a nicer error message to let users know some context around why.

jcrist:
When using Prefect with Cloud/Server, custom functions aren't supported
for fields in schedules (e.g. filters), as these functions are run as
part of Cloud/Server's infrastructure. Previously if a user tried to use
a custom function here, registering the flow would raise an opaque
"invalid function" error message, we now provide a more helpful message.

Fixes #3495.

This PR:

### DIFF
--- a/changes/pr3540.yaml
+++ b/changes/pr3540.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Raise a better error message when trying to register a flow with a schedule using custom filter functions - [#3450](https://github.com/PrefectHQ/prefect/pull/3540)"
--- a/src/prefect/utilities/serialization.py
+++ b/src/prefect/utilities/serialization.py
@@ -394,7 +394,7 @@ class StatefulFunctionReference(fields.Field):
 
         if not valid_bases and self.reject_invalid:
             raise ValidationError(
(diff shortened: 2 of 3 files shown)


jcrist:
Example:

```python
In [1]: from datetime import timedelta

In [2]: from prefect import Flow

In [3]: from prefect.schedules import Schedule

In [4]: from prefect.schedules.clocks import IntervalClock

In [5]: sched = Schedule(
   ...:     clocks=[IntervalClock(timedelta(hours=1))],
   ...:     filters=[lambda x: x.weekday() == 1],
   ...: )

In [6]: flow = Flow("example", schedule=sched)

In [7]: flow.serialize()
...
ValidationError: When running wth Prefect Cloud/Server, custom functions aren't supported for `Schedule.filters`. Unable to serialize `<function <lambda> at 0x7fc983656550>`.
```

joshmeek (approved):

PR #3546, opened 2020-10-20
Add network mode

AlessandroLollo:
## Summary
This PR remove the fixed "awsvpc" value for networkMode, which might not be always the best option (i.e: when running flow with launch type EC2)

## Changes

Now networkMode has to be provided along with other task definition arguments

## Importance

More networking flexibility for running flows on EC2.
For example, if you set networkMode to `bridge`, then the Docker container will be able to use the network defined for the EC2 instance.

## Checklist

This PR:


src/prefect/agent/fargate/agent.py
```diff
@@ -682,7 +683,6 @@ def _create_task_definition(
         self.boto3_client.register_task_definition(
             family=task_definition_name,  # type: ignore
             containerDefinitions=container_definitions,
-            networkMode="awsvpc",
```
joshmeek, line 685: This needs to default to `awsvpc` otherwise it's a breaking change

src/prefect/agent/fargate/agent.py
```diff
@@ -796,14 +819,31 @@ def validate_configuration(self) -> None:
         # Register task definition
         flow_task_definition_kwargs = copy.deepcopy(self.task_definition_kwargs)
 
+        # If networkMode is not provided, default to awsvpc
+        networkMode = (
```
joshmeek, line 823: Same comment as above

src/prefect/agent/fargate/agent.py
```diff
@@ -671,6 +672,23 @@ def _create_task_definition(
...
+            else "awsvpc"
+        )
+
+        # Remove networkMode if provided in flow task definition
+        try:
+            flow_task_definition_kwargs.pop("networkMode")
+        except KeyError:
+            self.logger.info(
+                "networkMode not passed in task definitions args, default to awsvpc"
+            )
```
joshmeek, line 688: I'm not sure if all of this is needed and instead you could do something like:

```
self.boto3_client.register_task_definition(..., networkMode=flow_task_definition_kwargs.get("networkMode", "awsvpc"), ...
```

joshmeek:
@AlessandroLollo Apologies my comment captured too much of the code block 😅! And I now see why you were removing from the dict to avoid duplicate passing to the boto3 calls. Let's keep the `pop` in so we have a workflow like this:

```python
network_mode = flow_task_definition_kwargs.pop("networkMode", "awsvpc")

register_task_definition(..., networkMode=network_mode, ...)
```

AlessandroLollo:
@joshmeek @jcrist do you guys have any suggestions on how to resolve the issue related to code style checking and docs which is failing?

madkinsz:
> do you guys have any suggestions on how to resolve the issue related to code style checking and docs which is failing?

@AlessandroLollo perhaps this was changed between when you forked and now? Can you try merging `PrefectHQ:master` into your branch? If that does not fix it, we can explicitly ignore this type error with a mypy `# type: ignore`

joshmeek:
Looks like mypy is out of sync but IIRC I fixed this one a week or so ago 🤔

AlessandroLollo:
I merged from master before making any changes, so I think we can ignore that failure

madkinsz:
> I merged from master before making any changes, so I think we can ignore that failure

At https://github.com/AlessandroLollo/prefect/tree/add-networkMode you can see `This branch is 11 commits ahead, 340 commits behind PrefectHQ:master.` and your `task_runner.py` has not been updated in 19 days and is notably missing these recent commits at https://github.com/PrefectHQ/prefect/commits/master/src/prefect/engine/task_runner.py which fix this error.

AlessandroLollo:
> > I merged from master before making any changes, so I think we can ignore that failure
> 
> At https://github.com/AlessandroLollo/prefect/tree/add-networkMode you can see `This branch is 11 commits ahead, 340 commits behind PrefectHQ:master.` and your `task_runner.py` has not been updated in 19 days and is notably missing these recent commits at https://github.com/PrefectHQ/prefect/commits/master/src/prefect/engine/task_runner.py which fix this error.

Hi @madkinsz,
my bad, didn't notice it.
No more code style checks and docs issues after following your suggestion, thanks!
However, now there are tests failing for Python 3.6 and Python 3.7.

joshmeek:
@AlessandroLollo You can ignore those, some weird non-deterministic test failures happening in CI currently 🙂 

Do you mind if I push some stylistic changes to this branch WRT how we're popping the value off of the dict?

AlessandroLollo:
Hey @joshmeek thanks for the update!
Regarding stylistic changes, feel free to refactor as you want

joshmeek (approved):

PR #3560, opened 2020-10-21
Fix a few flaky tests

jcrist:
Hard to know if this is effective without a few tests runs

- Reduce resource usage for dask clusters, to hopefully fix issues on resource constrained CI systems
- Fixup some of the timeout tests to hopefully be less flaky on Circle CI. Note that the changes shouldn't make the tests take any longer (the longer timeouts are for tests that should never timeout).


cicdw (approved):
Nice, this LGTM

PR #3562, opened 2020-10-22
fix(azure): need to use BlobServiceClient with the new version

kfirstri:
## Summary
azure-blob-storage library used to have a BlockBlobService class but it was changed to BlobServiceClient, so i changed the Task to create the blob_service using the new class

## Changes
Instead of using `BlockBlobService` I used `BlobServiceClient.from_connection_string(conn_str=azure_credentials)`

## Importance
Today when using the azure tasks the task fails because `BlockBlobService` does not exists

## Checklist

This PR:

### DIFF
--- a/changes/issue3562.yaml
+++ b/changes/issue3562.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Use `BlobServiceClient` instead of `BlockBlobService` to connect to azure blob in azure tasks - [#3562](https://github.com/PrefectHQ/prefect/pull/3562)"
+
+contributor:
+  - "[Kfir Stri](https://github.com/kfirstri)"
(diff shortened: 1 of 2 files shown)


jcrist:
Thanks @kfirstri. Do you know what version of `azure-blob-storage` made this change? Would it make sense to support both versions (with a version check), or should we bump our dependency requirements?

jcrist:
It also looks like our linting tests failed for you, could you run `black` on the file(s) you changed? You can find instructions here: https://docs.prefect.io/core/development/style.html#black-formatting.

kfirstri:
Hi @jcrist , your requirements for `azure-blob-storage` package is already `"azure-storage-blob >= 12.1.0, < 13.0"`
and `BlockBlobService` exited in the 2.x version if im not mistaken.. So I think you don't need to support the old Class.

jcrist:
This looks good to me @kfirstri. Can you add a changelog entry to the `changes/` directory? Instructions here: https://docs.prefect.io/core/development/contributing.html#changelog

madkinsz (approved):

PR #3563, opened 2020-10-22
Fix CloudHandler's "Failed to write log with error: Object of type bytes is not JSON serializable"

raphael-riel:
## Summary
I'm using a library that (wrongly) emits Click's coloured output `click.secho`when outside of a CLI context.
Adding color to a tty is done via Control characters, which make the log being bytes instead of a str.

This was resulting in CloudHandler getting the error `Failed to write log with error: Object of type bytes is not JSON serializable`

image

In local in my terminal we can confirm it was bytes causing the problem:
image

## Changes
Convert bytes to string before sending the line to CloudHandler.
Character that cannot be converted are ignored. 

## Importance
Missing some valuable logs and getting a truckload of CloudHandler Errors.

## Checklist

This PR:


changes/pr3563.yaml
```diff
@@ -0,0 +1,5 @@
+fix:
+  - "Fix CloudHandler failing on bytes - [#3563](https://github.com/PrefectHQ/prefect/pull/3563)"
```
jcrist, line 2:
```suggestion
  - "Tasks with `log_stdout=True` work with non-utf8 output - [#3563](https://github.com/PrefectHQ/prefect/pull/3563)"
```

src/prefect/utilities/logging.py
```diff
@@ -305,14 +305,20 @@ class RedirectToLog:
...
-    def write(self, s: str) -> None:
+    def write(self, s: Union[str, bytes]) -> None:
         """
         Write message from stdout to a prefect logger.
         Note: blank newlines will not be logged.
 
         Args:
             s (str): the message from stdout to be logged
         """
+        if isinstance(s, bytes):
```
jcrist, line 316: I'm a bit confused by this. When we capture `stdout`, all we do is replace `sys.stdout` in the task context with a new file-like object. Normally `sys.stdout` is a `io.TextIOWrapper`, which takes in `str` (not `bytes`) as input. No file type should ever get a mix of `str` or `bytes` under proper operation, I'm confused why `click` is sending it bytes. Perhaps our mock file class here is missing an interface method that click relies on to infer whether to send bytes or str?

Can you provide a reproducible example so we can investigate deeper? While the above might fix it, I don't think it's the best fix.

raphael-riel:
@jcrist Here's a Flow that I can confirm failed at logging output:
```python

import click
from prefect import Flow, Task


class Run(Task):

    def run(self) -> None:
        click.secho("Test", fg="green")


run = Run(log_stdout=True)


with Flow("bytes-flow") as flow:
    res = run()


if __name__ == '__main__':
    flow_state = flow.run()


```

image

And not on Cloud
image


I will check what's click doing that creates such behaviour.

raphael-riel:
@jcrist Got it.
https://github.com/pallets/click/blob/97de2c97275f52adc5c6ef8bb8aa993ce997fe7e/src/click/_compat.py#L159

jcrist:
I think it is. `sys.stdout` (what we are replacing) is a text writer, not a binary writer. To match that interface, we should also raise on receiving non-str values.

raphael-riel:
Ok! Let me change my PR to fit this approach!

raphael-riel:
@jcrist Change done. I reused StringIO's way to handle bytes.
image

jcrist:
LGTM, would you mind adding a test in `tests/utilities/test_logging.py`? Click is a dependency of prefect, so something like this should work (actual `assert` left up to you)

```python
def test_redirect_to_log_is_textwriter(caplog):
    log_stdout = utilities.logging.RedirectToLog()
    with contextlib.redirect_stdout(log_stdout):
        click.secho(...)
    log_stdout.flush()

    # check something about the logs in `caplog`
    logs = [r.message for r in caplog.records if r.levelname == "INFO"]
    ...
```

jcrist (approved):
Thanks!

PR #3574, opened 2020-10-27
add cursor_type param to task, unit test

mmarinaccio:
## Summary
Add support for the Pymysql `DictCursor` class in `MySQLFetch`.

## Changes
This PR adds an extra `cursor_type` argument to `MySQLFetch`, for which the user can define `"cursor"`, `"dict_cursor"`, or leave blank. A new conditional has also been added to the task's `run` method to choose the corresponding cursor class which is then set in `pymysql.connect`.

## Importance
There are cases where it's useful to have column names as part of the mysql fetch response, especially in the context of task mapping where downstream tasks behave or build new queries according to the particular column / values inputted. In such a case, without the dict cursor, downstream tasks must be aware of the schema or simply have hardcoded the index value of the column being observed.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -196,9 +196,7 @@ class MySQLFetch(Task):
         elif isinstance(cursor_type, str):
             cursor_class = cursor_types.get(cursor_type.lower())
         else:
-            raise TypeError(
-                f"'cursor_type' must be of type str or a full cursor class, got {type(cursor_type)}"
-            )
+            cursor_class = None
 
         if not cursor_class:
             raise TypeError(
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -108,7 +108,8 @@ class MySQLFetch(Task):
         - query (str, optional): query to execute against database
         - commit (bool, optional): set to True to commit transaction, defaults to false
         - charset (str, optional): charset of the query, defaults to "utf8mb4"
-        - cursor_type (Union[str, Callable], optional): cursor type you want to use (defaults to standard Cursor class)
+        - cursor_type (Union[str, Callable], optional): The cursor type to use.
+            Can be `'cursor'` (the default), `'dictcursor'`, or a full cursor class.
         - **kwargs (Any, optional): additional keyword arguments to pass to the
             Task constructor
     """
@@ -164,7 +165,8 @@ class MySQLFetch(Task):
             - query (str, optional): query to execute against database
             - commit (bool, optional): set to True to commit transaction, defaults to false
             - charset (str, optional): charset of the query, defaults to "utf8mb4"
-            - cursor_type (Union[str, Callable], optional): cursor type you want to use (defaults to standard Cursor class)
+            - cursor_type (Union[str, Callable], optional): The cursor type to use.
+                Can be `'cursor'` (the default), `'dictcursor'`, or a full cursor class.
 
         Returns:
             - results (tuple or list of tuples): records from provided query
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -109,7 +109,8 @@ class MySQLFetch(Task):
         - commit (bool, optional): set to True to commit transaction, defaults to false
         - charset (str, optional): charset of the query, defaults to "utf8mb4"
         - cursor_type (Union[str, Callable], optional): The cursor type to use.
-            Can be `'cursor'` (the default), `'dictcursor'`, `'sscursor'`, `'ssdictcursor'`, or a full cursor class.
+            Can be `'cursor'` (the default), `'dictcursor'`, `'sscursor'`, `'ssdictcursor'`,
+            or a full cursor class.
         - **kwargs (Any, optional): additional keyword arguments to pass to the
             Task constructor
     """
@@ -166,7 +167,8 @@ class MySQLFetch(Task):
             - commit (bool, optional): set to True to commit transaction, defaults to false
             - charset (str, optional): charset of the query, defaults to "utf8mb4"
             - cursor_type (Union[str, Callable], optional): The cursor type to use.
-                Can be `'cursor'` (the default), `'dictcursor'`, `'sscursor'`, `'ssdictcursor'`, or a full cursor class.
+                Can be `'cursor'` (the default), `'dictcursor'`, `'sscursor'`, `'ssdictcursor'`,
+                or a full cursor class.
 
         Returns:
             - results (tuple or list of tuples): records from provided query
@@ -200,7 +202,8 @@ class MySQLFetch(Task):
 
         if not cursor_class:
             raise TypeError(
-                f"'cursor_type' should be one of {list(cursor_types.keys())} or a full cursor class, got {cursor_type}"
+                f"'cursor_type' should be one of {list(cursor_types.keys())} or a "
+                f"full cursor class, got {cursor_type}"
             )
 
         conn = pymysql.connect(
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -34,7 +34,7 @@ class MySQLExecute(Task):
         query: str = None,
         commit: bool = False,
         charset: str = "utf8mb4",
-        **kwargs: Any
+        **kwargs: Any,
     ):
         self.db_name = db_name
         self.user = user
@@ -126,7 +126,7 @@ class MySQLFetch(Task):
         commit: bool = False,
         charset: str = "utf8mb4",
         cursor_type: Union[str, Callable] = "cursor",
-        **kwargs: Any
+        **kwargs: Any,
     ):
         self.db_name = db_name
         self.user = user
@@ -141,7 +141,9 @@ class MySQLFetch(Task):
         self.cursor_type = cursor_type
         super().__init__(**kwargs)
 
-    @defaults_from_attrs("fetch", "fetch_count", "query", "commit", "charset", "cursor_type")
+    @defaults_from_attrs(
+        "fetch", "fetch_count", "query", "commit", "charset", "cursor_type"
+    )
     def run(
         self,
         query: str,
@@ -180,16 +182,18 @@ class MySQLFetch(Task):
 
         cursor_types = {
             "cursor": pymysql.cursors.Cursor,
-            "dictcursor": pymysql.cursors.DictCursor
+            "dictcursor": pymysql.cursors.DictCursor,
         }
 
         if callable(cursor_type):
             cursor_class = cursor_type
         else:
             cursor_class = cursor_types.get(cursor_type.lower())
-        
+
         if cursor_class is None or cursor_class not in cursor_types.values():
-            raise TypeError(f"'cursor_type' should be one of ('cursor', 'dictcursor') or the callable equivalent, got {cursor_type}")
+            raise TypeError(
+                f"'cursor_type' should be one of ('cursor', 'dictcursor') or the callable equivalent, got {cursor_type}"
+            )
 
         conn = pymysql.connect(
             host=self.host,
@@ -198,7 +202,7 @@ class MySQLFetch(Task):
             db=self.db_name,
             charset=self.charset,
             port=self.port,
-            cursorclass=cursor_class
+            cursorclass=cursor_class,
         )
 
         try:
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -178,15 +178,18 @@ class MySQLFetch(Task):
                 "The 'fetch' parameter must be one of the following - ('one', 'many', 'all')"
             )
 
-        if cursor_type not in {"cursor", "dict_cursor"}:
-            raise ValueError(
-                f"Unsupported cursor_type '{cursor_type}'. Please choose one of the following - ('cursor', 'dict_cursor')"
-            )
+        cursor_types = {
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/mysql/mysql.py
```diff
@@ -173,13 +178,24 @@ def run(
...
 
+        if cursor_type not in {"cursor", "dict_cursor"}:
+            raise ValueError(
+                f"Unsupported cursor_type '{cursor_type}'. Please choose one of the following - ('cursor', 'dict_cursor')"
+            )
+
+        if cursor_type == "cursor":
+            cursor_class = pymysql.cursors.Cursor
+        else:
+            cursor_class = pymysql.cursors.DictCursor
```
madkinsz, line 189: Hi! Thanks for contributing to Prefect :) 

I'm curious why you choose to take strings and then do this match rather than just taking the class you're interested in directly? e.g. `cursor_type: pymysql.cursors.Cursor = pymysql.cursors.Cursor`

mmarinaccio: Hey there! Thanks for the prompt response. Good question. My initial thinking was to avoid the user having to import `pymysql.cursors` in their flow. I thought it might be cleaner to reserve all pymysql imports for the pymysql task module, but one could argue that less code wins here, which your suggestion certainly enables. Let me know if you'd like me to make that change.

madkinsz: I think that it'd be more maintainable and if you're trying to customize your cursors it seems reasonable to import the option you want. I'd wait until a code owner weighs in though. @jcrist

jcrist: I'm not familiar with `pymysql`. Looks like there's currently 4 types of cursors: https://pymysql.readthedocs.io/en/latest/modules/cursors.html. Perhaps support both? I'd write this as:

```python
cursor_types = {
    "cursor": pymysql.cursors.Cursor,
    "sscursor": pymysql.cursors.SSCursor,
    ...
}
if callable(cursor_type):
    cursor_class = cursor_type
else:
    try:
        cursor_class = cursor_types[cursor_type.lower()]
    except KeyError:
        raise TypeError(f"`cursor_type` should be a callable or one of {'cursor', 'sscursor', 'dictcursor', 'ssdictcursor'}, got {cursor_type}") from None
```

Users can then pass in either a short string name, or a callable class itself. Slight preference for no `_` in the string names, just lowercase the class name.

madkinsz: Unless this is a common use case supporting multiple input types for this parameter seems like more work than it's worth, but I also don't have experience with `pymysql` 🤷

mmarinaccio: I deliberately avoided the `SS*` cursors, since these are built to provider iterators for looping through sql results pulled from the SQL server 'on-demand'. E.g.
```python
import pymysql.cursors

conn = pymysql.connect(
host=self.host,
user=self.user,
password=self.password,
db=self.db_name,
charset=self.charset,
port=self.port,
cursorclass=pymysql.cursors.SSCursor
)

with conn.cursor() as cursor:
    cursor.execute(query)

    for row in cursor:
        # each iteration fetches a new result from the sql server, as opposed to all results buffered in memory at once.
        print(row)
```

I'm skeptical that this paradigm has a place in this task class (or if it's generally a broader Prefect anti-pattern), since as of now we're just fetching one/some/all results and returning at once.

jcrist: Makes sense to me. I'm fine with the shorthand names, provided:
- the names match the class names sans capitalization (no underscore, just check `name.lower()` matches `cursor` or `dictcursor`)
- You can also pass in a class directly
- We check the full set and error on unknown names (right now if it's not `cursor` it's `dictcursor`, even if the user types something else).

mmarinaccio: Thanks @jcrist. I'll have another commit pushed by tomorrow.

mmarinaccio: @jcrist @madkinsz I implemented the changes discussed above. Let me know what you think.

src/prefect/tasks/mysql/mysql.py
```diff
@@ -108,6 +108,7 @@ class MySQLFetch(Task):
         - query (str, optional): query to execute against database
         - commit (bool, optional): set to True to commit transaction, defaults to false
         - charset (str, optional): charset of the query, defaults to "utf8mb4"
+        - cursor_type (Union[str, Callable], optional): cursor type you want to use (defaults to standard Cursor class)
```
jcrist, line 111: Can you mention the set of possible string names? Something like:

```suggestion
        - cursor_type (Union[str, Callable], optional): The cursor type to use.
            Can be `'cursor'` (the default), `'dictcursor'`, or a full cursor class.
```

src/prefect/tasks/mysql/mysql.py
```diff
@@ -173,13 +178,27 @@ def run(
...
+            "dictcursor": pymysql.cursors.DictCursor
+        }
+
+        if callable(cursor_type):
+            cursor_class = cursor_type
+        else:
+            cursor_class = cursor_types.get(cursor_type.lower())
+        
+        if cursor_class is None or cursor_class not in cursor_types.values():
+            raise TypeError(f"'cursor_type' should be one of ('cursor', 'dictcursor') or the callable equivalent, got {cursor_type}")
```
madkinsz, line 192:
```suggestion
        if not cursor_class:
            raise TypeError(f"'cursor_type' should be one of ('cursor', 'dictcursor') or a Cursor class type, got {cursor_type}")
```

I don't think Jim meant to limit the cursors to just these two. The callable should allow custom cursors to be passed in.

madkinsz: Also having this here with the `.get` you wrote is a great way to check for missing types. Nice :)

src/prefect/tasks/mysql/mysql.py
```diff
@@ -136,16 +138,18 @@ def __init__(
...
-    @defaults_from_attrs("fetch", "fetch_count", "query", "commit", "charset")
+    @defaults_from_attrs("fetch", "fetch_count", "query", "commit", "charset", "cursor_type")
     def run(
         self,
         query: str,
         fetch: str = "one",
         fetch_count: int = 10,
         commit: bool = False,
         charset: str = "utf8mb4",
+        cursor_type: Union[str, Callable] = "cursor",
```
madkinsz, line 152: I would be a little more specific about the type here
```suggestion
        cursor_type: Union[str, pymysql.cursors.Cursor] = pymysql.cursors.Cursor,
```
All of the other cursors inherit from the `Cursor` class and we can specify that this is what we are looking for

I'd also prefer to default to the real class type because it's easier for a user to know what we mean by the default value and it's already imported

src/prefect/tasks/mysql/mysql.py
```diff
@@ -158,6 +162,7 @@ def run(
             - query (str, optional): query to execute against database
             - commit (bool, optional): set to True to commit transaction, defaults to false
             - charset (str, optional): charset of the query, defaults to "utf8mb4"
+            - cursor_type (Union[str, Callable], optional): cursor type you want to use (defaults to standard Cursor class)
```
madkinsz, line 165:
```suggestion
            - cursor_type (Union[str, Cursor], optional): cursor type you want to use (defaults to standard Cursor class)
```

jcrist, line 165: Same here

```suggestion
            - cursor_type (Union[str, Callable], optional): The cursor type to use.
                Can be `'cursor'` (the default), `'dictcursor'`, or a full cursor class.
```

jcrist (commented):
Looks like you're failing on `black` tests, can you run black on your changes (instructions here: https://docs.prefect.io/core/development/style.html)?

src/prefect/tasks/mysql/mysql.py
```diff
@@ -173,13 +178,27 @@ def run(
                 "The 'fetch' parameter must be one of the following - ('one', 'many', 'all')"
             )
 
+        cursor_types = {
+            "cursor": pymysql.cursors.Cursor,
+            "dictcursor": pymysql.cursors.DictCursor
+        }
+
+        if callable(cursor_type):
```
madkinsz, line 186: I believe @jcrist checks for a callable here because it's quick and does not limit cursors to things that explicitly  nherit from `Cursor` but I think it's probably worth doing this
```suggestion
        if isinstance(cursor_type, pymysql.cursors.Cursor):
```

src/prefect/tasks/mysql/mysql.py
```diff
@@ -173,13 +178,27 @@ def run(
...
+            cursor_class = cursor_type
+        else:
+            cursor_class = cursor_types.get(cursor_type.lower())
```
madkinsz, line 189: Note that was written, this does not cover the case where the cursor is not callable and not a string (which was something Jim alluded to) so if someone passed a list in it'd fail on this `.lower()` call and be a confusing error.

```suggestion
        elif isinstance(cursor_type, str):
            cursor_class = cursor_types.get(cursor_type.lower())
        else:
            cursor_class = None  # will be raised below
```

madkinsz (changes requested):

src/prefect/tasks/mysql/mysql.py
madkinsz: Also note the TypeError can use your dict.keys() to display the possible string cursor values

src/prefect/tasks/mysql/mysql.py
madkinsz: This could also just remain `callable` and we can let the pymysql library determine if it's a valid Cursor as it pleases

src/prefect/tasks/mysql/mysql.py
jcrist: This type signature isn't correct, you don't pass in a `Cursor` type, you pass in a callable that might create a cursor type. IMO the `Callable` signature was fine.

src/prefect/tasks/mysql/mysql.py
jcrist: I wouldn't check the types, minimal work on our end. I'd only validate `cursor_type` is in the set of string shorthand names, or is a callable. Let the mysql library error if needed in other cases.

src/prefect/tasks/mysql/mysql.py
madkinsz: Ah yep my bad, I definitely conflated this during the review. I think (for annotation purposes) `Type[Cursor]` is more informative to users than `Callable` unless you want to extend the `Callable` annotation to indicate a return type of `Cursor` but however this is done I think it's not a big deal and not worth getting further into the weeds -- I'll be happy with whatever is submitted.

src/prefect/tasks/mysql/mysql.py
madkinsz: See Jim's comment at https://github.com/PrefectHQ/prefect/pull/3574#discussion_r513530489

src/prefect/tasks/mysql/mysql.py
madkinsz: @mhmcdonald thanks again for taking the time to work on this! I've marked a few conversations about type annotations as resolved as I don't think it's worth fretting over unless you wanted to. Following a couple edge case fixes this looks great.

src/prefect/tasks/mysql/mysql.py
mmarinaccio: > I don't think Jim meant to limit the cursors to just these two. The callable should allow custom cursors to be passed in.

I'll remove this check. I initially included to avoid potentially confusing or unexpected behavior resulting from the `SS*` cursor classes, as I described here: https://github.com/PrefectHQ/prefect/pull/3574#discussion_r512978435. If a user sets `cursor_type=pymysql.cursors.SSCursor`.

However, I tested against an actual db and it seems to work just fine, even though the current `MySQLFetch` implementation doesn't use it effectively (ideally it would iterate over the cursor and yield rows).

With that said, I think it makes sense to support all of the cursors.

mmarinaccio:
@madkinsz @jcrist This is ready for another look. Let me know if you'd like to see anything else changed. Thanks!

src/prefect/tasks/mysql/mysql.py
```diff
@@ -173,13 +184,36 @@ def run(
...
+        }
+
+        if callable(cursor_type):
+            cursor_class = cursor_type
+        elif isinstance(cursor_type, str):
+            cursor_class = cursor_types.get(cursor_type.lower())
+        else:
+            raise TypeError(
+                f"'cursor_type' must be of type str or a full cursor class, got {type(cursor_type)}"
+            )
```
madkinsz, line 201: If you do `cursor_class = None` here the TypeError does not need to be duplicated

madkinsz (dismissed):
Minor nit but this lgtm!

src/prefect/tasks/mysql/mysql.py
mmarinaccio: Ah good catch. I'm happy to incorporate if it's not more trouble than it's worth on your side. Additionally, I see a CircleCI failure here:

https://app.circleci.com/pipelines/github/PrefectHQ/prefect/6569/workflows/e5016f4d-91a2-4462-9ab0-9caa0004aafd/jobs/102603

Although, the failure cases seem orthogonal to the files changed in this PR. Is this a concern?

madkinsz: It's not more trouble on our side! 

I'm not sure what's up with that test. I would ensure that your forked repo is up to date with `master` and merged into this branch.

madkinsz (approved):

jcrist (approved):

PR #3581, opened 2020-10-28
Artifacts API

joshmeek:
## Summary
Core side of the Artifacts API to be used in collaboration with the `artifact_api` branch on server (https://github.com/PrefectHQ/server/pull/121). This creates a low level set of functions on the Client for creating, updating, and deleting task run artifacts. A top level API is also defined for ease of use / enforcing "schema" of data being posted. I believe this to be the cleanest approach.

Example use:

```python
from prefect import artifacts

@task
def post_some_artifacts():
    mkdown = """
    # This is markdown

    [apparently](https://prefect.io)

    Am I markdown?

    -----

    #### We'll find out
    """

    # Create
    artifact_id = artifacts.create_link("https://prefect.io")
    print(artifact_id)

    artifact_id = artifacts.create_markdown(mkdown)
    print(artifact_id)

    # Update
    artifacts.update(artifact_id, "https://prefect.io/cloud")
    
    # Delete
    artifacts.delete_link(artifact_id)
```

Now those artifacts of that task can be queried for in the backend. Once merged and promoted this can be added to some of the tasks in the task library such as publishing a link artifact for the `CreateFlowRunTask`.

## Checklist

This PR:


madkinsz:
I envisioned an object oriented approach where we have

```
Artifact:
     id: UUID
     data: dict
     _update() 
     _create()-> calls `create_task_run_artifact` with self.data
 
Link(Artifact):
    data: dict
    update(link: str) -> update self.data, super._update()
    create()

Markdown(Artifact):
    ....
```

Then you could create an artifact

`my_md = artifacts.Markdown("foo").create()`

and update it

`my_md.update("bar")`

I also envisioned the api looking like
```
    def create_task_run_artifact(
        self, task_run_id: str, artifact: Artifact, tenant_id: str = None
    ) -> str:
```

I'm not sure there's a clear benefit to this approach though, it opens a can of worms of expectations around 'state' within the object.

joshmeek:
@madkinsz Yeah I don't see a benefit in going with an object based approach as a few small functions would provide the same functionality without the extra cruft.

madkinsz:
> Yeah I don't see a benefit in going with an object based approach as a few small functions would provide the same functionality without the extra cruft.

Without getting into it too much, it seems like the benefit of the object based approach would be that the interface for working with an artifact is unified so you don't have to find the function for whatever kind you're interest in. You choose a kind then you can submit it to the client/api using a single, consistent function. At this scale it probably doesn't matter though.

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
+from typing import Union
+
+from prefect import context, Client
```
madkinsz, line 3: Isn't it dangerous to import the context this way?

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def running_with_backend() -> bool:
```
madkinsz, line 6: Would this be useful / better placed elsewhere?

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
...
+    Determine if running in context of a backend. This is always true when running
+    using the `CloudTaskRunner`.
+
+    Returns:
+        - bool: if `running_with_backend` is set in context
+    """
+    return bool(context.get("running_with_backend"))
+
+
+def create_link(link: str) -> Union[str, None]:
```
madkinsz, line 17: nit: I'd prefer `Optional[str]` generally here and elsewhere

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Args:
+        - task_run_artifact_id (str): the ID of an existing task run artifact
+        - link (str): the new link to update the artifact with
+
+    Returns:
+        - bool: whether or not the request was successful
+    """
+    if not running_with_backend():
+        return False
```
madkinsz, line 55: Could be nice to have as a decorator? Should there be a UserWarning involved?

joshmeek: Yeah can make a decorator, was getting some weird documentation errors so switched to this but on biggie

src/prefect/artifacts.py
joshmeek: Yeah I can make an artifacts utility file

src/prefect/artifacts.py
joshmeek: Hmm doesn't seem to affect my runs at all whether it's imported here or inside the functions that use it

src/prefect/artifacts.py
madkinsz: From talking to @cicdw there was an intention to stop the utilities from creeping larger. I was curious if it fit into any of the existing utility groups or modules.

src/prefect/artifacts.py
madkinsz: I remember seeing that it would always be fully imported / it changes when the context is loaded? Perhaps it only really matters within Tasks

src/prefect/artifacts.py
madkinsz: I think it mostly makes sense if it's more than one line of repeated handling

joshmeek: Discussed w/ @jcrist, skipping decorator for now

src/prefect/artifacts.py
cicdw: To be super clear, the utility comment I mentioned previously was because it was a proposed new utility to DRY out a single line of code, which is not a good tradeoff.

jcrist: My 2c on things like these is to keep the helper functions local to the module they're used until there's a good reason for another module to need it. Since they're non-public, moving helpers around doesn't harm anything. But adding things to a utility module adds indirection (I have to go look at another file to understand some code), and can harden interfaces too early (can I change this helper function? what if something else relies on its current signature?).

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
...
+    Delete an existing link artifact
+
+    Note: The functionality here is experimental, and may change between
+    versions without notice. Use at your own risk.
+
+    Args:
+        - task_run_artifact_id (str): the ID of an existing task run artifact
+
+    Returns:
+        - bool: whether or not the request was successful
```
jcrist, line 77: Under what circumstances would the request be unsuccessful. Api failures will error, I assume this just means "did the id exist before"?

src/prefect/artifacts.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Note: The functionality here is experimental, and may change between
+    versions without notice. Use at your own risk.
+
+    Args:
+        - task_run_artifact_id (str): the ID of an existing task run artifact
+        - link (str): the new link to update the artifact with
+
+    Returns:
+        - bool: whether or not the request was successful
```
jcrist, line 52: Same here. What does "unsuccessful" mean?

src/prefect/artifacts.py
joshmeek: The API returns a boolean whether or not the row in the DB was updated

```python
return bool(result.affected_rows)
```

I can update the docstring to make this clearer

jcrist: If `task_run_artifact_id` doesn't exist, what happens (error? or returns false?).

I personally don't see a user for a return value here, dropping it from `update` and `delete` makes the most sense to me.

joshmeek: If it doesn't exist then an error is raised

src/prefect/artifacts.py
```diff
@@ -0,0 +1,148 @@
...
+    if task_run_artifact_id is None:
+        raise ValueError("The ID of an existing task run artifact must be provided.")
+
+    client = Client()
+    client.update_task_run_artifact(
+        task_run_artifact_id=task_run_artifact_id, data={"markdown": markdown}
+    )
+
+
+def delete_markdown(task_run_artifact_id: str) -> None:
```
jcrist, line 131: Since all the delete functions are the same (just take a `task_run_artifact_id`), do we really need/want multiple versions or would a single `delete`/`delete_artifact` function be sufficient? If we want a `delete_*` method for each type, could at least share the common implementation in a `_` prefixed helper and call that in e.g. `delete_markdown`.

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -266,6 +266,7 @@ def run(
             - State: `State` representing the final post-run state of the `Flow`.
         """
         context = context or {}
```
jcrist, line 268: This currently mutates the passed in `context`, which we ideally shouldn't do.

```suggestion
        context = (context or {}).copy()
```

src/prefect/client/client.py
```diff
@@ -1649,3 +1649,89 @@ def get_agent_config(self, agent_config_id: str) -> dict:
...
+    def update_task_run_artifact(self, task_run_artifact_id: str, data: dict) -> bool:
+        """
+        Update an artifact that corresponds to a specific task run
+
+        Args:
+            - task_run_artifact_id (str): the task run artifact id
+            - data (dict): the artifact data
+
+        Returns:
+            - bool: if the artifact update was successful
```
jcrist, line 1699: Since the return value here is something internal to the function of the database, should we remove it and return `None` instead (same for `delete_task_run_artifact`)? The meaning of "successful" here is a bit odd, since most reasons it should fail would error instead of returning `False`.

jcrist (commented):
Overall this looks good to me, just a few small comments.

src/prefect/artifacts.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Args:
+        - task_run_artifact_id (str): the ID of an existing task run artifact
+        - markdown (str): the new markdown to update the artifact with
+    """
+    if not running_with_backend():
+        return
+
+    if task_run_artifact_id is None:
+        raise ValueError("The ID of an existing task run artifact must be provided.")
```
jcrist, line 123: Could this check be put in the client? Then the shared code between `update_link` and `update_markdown` is very small.

src/prefect/artifacts.py
joshmeek: I'm cool with a single delete artifact call 👍

jcrist (approved):

Issue #2928, opened 2020-07-07
Remove the need for `Environment` classes entirely


jcrist:
Currently a registered flow is parametrized by the follow:

- An `Environment`, containing an `Executor` and some other configuration, specifying how and where to run the flow
- A `Storage` object specifying where to get the flow from
- An `Environment.metadata` dict specifying general metadata about the flow. There's a convention that the `image` key in that mapping specifies the image to use to the agent
- The `Agent` that kicks off the flow run.

This is a *lot* of different concepts, and the different options can be confusing to beginners and experienced users alike.

Say I'm a user who wants to run a flow as a K8s Job. I see there's a `KubernetesJobEnvironment` - that sounds like what I want. But there's also a `KubernetesAgent`. Do I need both? Only one? Why would I pick one or the other? Too many options and combinations of options can be overwhelming.

With #2805 we deprecated (or plan to deprecate) all of the environments except for:

- `LocalEnvironment`
- `FargateTaskEnvironment`
- `KubernetesJobEnvironment`

I believer with making the agents more configurable we can drop the last two entirely, removing environment configuration from something a user needs to think about.

The `FargateTaskEnvironment` and `KubernetesJobEnvironment` exist to allow users to configure a k8s job or fargate task without exposing that configuration to Prefect cloud (the configuration is stored with the flow in the flow in the `Storage` object, not in Cloud's DB). If users only want a single job spec for all k8s jobs, they can use a `LocalEnvironment` with the `KubernetesAgent` and things work out fine.

A different way to expose configurable specs to users would be to let them customize how the agent generates these specs from the flow run information. Currently this is hardcoded into each agent (template in a few flow-specific things, a few user-configured things, then kick off the k8s job/fargate task). We might make this a configurable callable, to let users fully customize this process, and also provide a few builtin versions for common patterns.

A possible proposal for the `KubernetesAgent`:

- We add a configurable callback `generate_job_spec(flow_run) -> V1JobSpec`. This defaults to an implementation similar to the one provided already.
- If they want to customize the job spec in a simple-ish way, we can let them override the template used by the builtin callback. All jobs would still look the same, but would be customized in some static way (say adding a static `Secret` to all pods).
- If a user wants super-custom-behavior, they can configure their own callback to be used. In the callback they'd have access to the `metadata` dict associated with the flow, which can contain whatever info they want.

With the third option, they could implement behaviors like:

[...]

PR #3596, opened 2020-10-29
allow use of multiple image pull secrets in KubernetesAgent, DaskKubernetesEnvironment

jameslamb:
## Summary 

Today, you can customize the value of `imagePullSecrets` (the secrets used to pull docker images) used by `KubernetesAgent` and `DaskKubernetesAgent`. However, `prefect` only allows you to specify a single secret.

This PR proposes expanding that, to allow multiple image pull secrets to be provided.

## Changes

Changes `KubernetesAgent` and `DaskKubernetesAgent` to accept multiple `imagPullSecrets` values. What I'm proposing here is backwards compatible with how `prefect` works today

* one secret: `"my-secret"`
* 2 secrets: `"my-secret,other-secret"`

## Importance

I started looking into this because of this message that Docker sent out last night:

> You are receiving this email because of a policy change to Docker products and services you use. On Monday, November 2, 2020 at 9am Pacific Standard Time, Docker will begin enforcing rate limits on container pulls for Anonymous and Free users. Anonymous (unauthenticated) users will be limited to 100 container image pulls every six hours, and Free (authenticated) users will be limited to 200 container image pulls every six hours, when enforcement is fully implemented...

More details are available in [this blog post](https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/).

This change means that people using `prefect` kubernetes components that pull public images from Docker Hub might want to now make *authenticated* (instead of anonymous) calls to Docker Hub, to reduce the risk of hitting that rate limit.

This PR makes it possible to run these components with multiple secrets so that, for example, an agent could pull images from either Docker Hub or a private registry like AWS ECR.

## Checklist

This PR:

### Notes for reviewers

* I searched for places to update like this:
    ```shell
    git grep -i imagePullSec src/
    git grep -i image_pull_sec src/
    ```
* I know that environments might go away in the future (#2928) but thought they should continue to get patches like this until they do. Let me know if you disagree and I can remove the `DaskKubernetesEnvironment` code from this

### Questions for reviewers

* is it intentional that there is no "customize imagePullSecrets" functionality for `KubernetesJobEnvironment`? Is the thinking the idea there that the customization should just go into `job_spec_file` instead of adding new environment variables / kwargs to that class?

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -303,7 +303,10 @@ class KubernetesAgent(Agent):
         - `IMAGE_PULL_POLICY`: policy for pulling images. Defaults to `"IfNotPresent"`.
         - `IMAGE_PULL_SECRETS`: name of an existing k8s secret that can be used to pull
                 images. This is necessary if your flow uses an image that is in a non-public
-                container registry, such as Amazon ECR.
+                container registry, such as Amazon ECR, or in a public registry that requires
+                authentication to avoid hitting rate limits. To specify multiple image pull
+                secrets, provide a comma-delimited string with no spaces, like
+                `"some-secret,other-secret"`.
         - `SERVICE_ACCOUNT_NAME`: name of a service account to run the job as.
                 By default, none is specified.
         - `YAML_TEMPLATE`: a path to where the YAML template should be loaded from. defaults
@@ -366,9 +369,15 @@ class KubernetesAgent(Agent):
         # Use image pull secrets if provided
         image_pull_secrets = os.getenv("IMAGE_PULL_SECRETS")
         if image_pull_secrets:
-            job["spec"]["template"]["spec"]["imagePullSecrets"][0][
-                "name"
-            ] = image_pull_secrets
+            secrets = image_pull_secrets.split(",")
+            for idx, secret_name in enumerate(secrets):
+                # this check preserves behavior from previous releases,
+                # where prefect would only overwrite the first entry in
+                # imagePullSecrets
+                if idx == 0:
+                    job["spec"]["template"]["spec"]["imagePullSecrets"][0] = {"name": secret_name}
+                else:
+                    job["spec"]["template"]["spec"]["imagePullSecrets"].append({"name": secret_name})
         else:
             del job["spec"]["template"]["spec"]["imagePullSecrets"]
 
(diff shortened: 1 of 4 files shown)


changes/pr3596.yaml
```diff
@@ -0,0 +1,5 @@
+enhancement:
+  - "allow use of multiple image pull secrets in KubernetesAgent, DaskKubernetesEnvironment - [#3596](https://github.com/PrefectHQ/prefect/pull/3596)"
```
jcrist, line 2:
```suggestion
  - "Allow use of multiple image pull secrets in `KubernetesAgent`, `DaskKubernetesEnvironment` - [#3596](https://github.com/PrefectHQ/prefect/pull/3596)"
```

jcrist (approved):
Thanks @jameslamb! This looks good to me.

Note that with the new `KubernetesRun` pattern (we should be rolling out docs/release in the next couple weeks) this is already possible - the full k8s job spec is exposed and configurable on either agent or as part of the `KubernetesRun` object itself.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -366,9 +369,19 @@ def generate_job_spec_from_environment(
...
-            ] = image_pull_secrets
+            secrets = image_pull_secrets.split(",")
+            for idx, secret_name in enumerate(secrets):
+                # this check preserves behavior from previous releases,
+                # where prefect would only overwrite the first entry in
+                # imagePullSecrets
+                if idx == 0:
+                    job["spec"]["template"]["spec"]["imagePullSecrets"][0] = {
+                        "name": secret_name
+                    }
```
madkinsz, line 380: @jcrist just curious why we do this

madkinsz (approved):

src/prefect/agent/kubernetes/agent.py
jcrist: The old way of managing the agent made a lot of assumptions on the underlying template (for better or worse). This will all go away with the new `KubernetesRun` pattern (once everything else is deprecated), but we can't break compatibility for now.

jameslamb:
> Thanks @jameslamb! This looks good to me.
> 
> Note that with the new `KubernetesRun` pattern (we should be rolling out docs/release in the next couple weeks) this is already possible - the full k8s job spec is exposed and configurable on either agent or as part of the `KubernetesRun` object itself.

oooo cool! First I've heard of that, I'm looking forward to trying it out.

PR #3599, opened 2020-10-30
added `tasks.jupyter.ExecuteNotebook`

bachsh:
## Summary
A prefect task for running jupyter notebooks

## Changes
Added a new task `tasks.jupyter.ExecuteNotebook`

## Importance
I use it myself for generating a visual report at the end of a prefect flow run. Got some positive feedback on it in the prefect slack workspace and decided to add it to prefect task library.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/jupyter/jupyter.py
+++ b/src/prefect/tasks/jupyter/jupyter.py
@@ -6,19 +6,20 @@ from prefect import Task
 from prefect.utilities.tasks import defaults_from_attrs
 
 
-class JupyterTask(Task):
+class ExecuteNotebook(Task):
     """
     Task for running Jupyter Notebooks.
     In order to parametrize the notebook, you need to mark the parameters cell as described in
         the papermill documentation: https://papermill.readthedocs.io/en/latest/usage-parameterize.html
 
     Args:
-        - notebook_path (string, optional): path to fetch the notebook from.
+        - path (string, optional): path to fetch the notebook from.
             Can be a cloud storage path.
             Can also be provided post-initialization by calling this task instance
-        - notebook_params (dict, optional): dictionary of parameters to use for the notebook
+        - parameters (dict, optional): dictionary of parameters to use for the notebook
             Can also be provided at runtime
-        - as_html (bool, optional): whether to output as HTML. (default: True)
+        - output_format (str, optional): Notebook output format.
+            Currently supported: json, html (default: json)
         - kernel_name (string, optional): kernel name to run the notebook with.
             If not provided, the default kernel will be used.
         - **kwargs: additional keyword arguments to pass to the Task constructor
@@ -26,41 +27,43 @@ class JupyterTask(Task):
 
     def __init__(
         self,
-        notebook_path: str = None,
-        notebook_params: dict = None,
-        as_html: bool = True,
+        path: str = None,
+        parameters: dict = None,
+        output_format: str = 'json',
         kernel_name: str = None,
         **kwargs
     ):
-        self.notebook_path = notebook_path
-        self.notebook_params = notebook_params
-        self.as_html = as_html
+        self.path = path
+        self.parameters = parameters
+        self.output_format = output_format
         self.kernel_name = kernel_name
         super().__init__(**kwargs)
 
-    @defaults_from_attrs("notebook_path", "notebook_params", "as_html")
+    @defaults_from_attrs("path", "parameters", "output_format")
     def run(
         self,
-        notebook_path: str = None,
-        notebook_params: dict = None,
-        as_html: bool = None,
+        path: str = None,
+        parameters: dict = None,
+        output_format: str = None,
     ) -> str:
         """
         Run a Jupyter notebook and output as HTML or JSON
 
         Args:
-        - notebook_path (string, optional): path to fetch the notebook from; can also be
+        - path (string, optional): path to fetch the notebook from; can also be
             a cloud storage path
-        - notebook_params (dict, optional): dictionary of parameters to use for the notebook
-        - as_html (bool, optional): whether to output as HTML
+        - parameters (dict, optional): dictionary of parameters to use for the notebook
+        - output_format (str, optional): Notebook output format.
+            Currently supported: json, html (default: json)
         """
         nb: nbformat.NotebookNode = pm.execute_notebook(
-            notebook_path, "-", parameters=notebook_params, kernel_name=self.kernel_name
+            path, "-", parameters=parameters, kernel_name=self.kernel_name
         )
-        if as_html:
+        if output_format == 'json':
+            return nbformat.writes(nb)
+        if output_format == 'html':
             html_exporter = nbconvert.HTMLExporter()
             (body, resources) = html_exporter.from_notebook_node(nb)
             return body
 
-        # return JSON format
-        return nbformat.writes(nb)
+        raise NotImplementedError("Notebook output %s not supported", output_format)
--- a/src/prefect/tasks/jupyter/jupyter.py
+++ b/src/prefect/tasks/jupyter/jupyter.py
@@ -1,0 +1,57 @@
+from typing import List, Tuple
+
+import nbformat
+from nbformat import NotebookNode
+
+from prefect import Task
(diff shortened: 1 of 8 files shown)


src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -0,0 +1,57 @@
...
+            notebook_path: str = None,
+            notebook_params: dict = None,
+            as_html: bool = True,
+            kernel_name: str = None,
+            **kwargs
+    ):
+        self.notebook_path = notebook_path
+        self.notebook_params = notebook_params
+        self.as_html = as_html
+        self.kernel_name = kernel_name
```
madkinsz, line 42: This appears to be unused

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -0,0 +1,57 @@
...
+            notebook_path,
+            "-",
+            parameters=notebook_params,
+        )
+        if as_html:
+            html_exporter = nbconvert.HTMLExporter()
+            (body, resources) = html_exporter.from_notebook_node(nb)
+            return body
+        else:  # return JSON format
+            return nbformat.writes(nb)
```
madkinsz, line 57: I tend to skip leave out `else` statements following `return` but this is just stylistic and I'm okay with either!

```suggestion

        # return JSON format
        return nbformat.writes(nb)
```

tests/tasks/jupyter/test_jupyter.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+from prefect.tasks.jupyter import JupyterTask
+
+pytest.importorskip("papermill")
+
+
+def test_jupyter():
```
madkinsz, line 8: Can you cover your JSON case as well?

tests/tasks/jupyter/test_jupyter.py
```diff
@@ -0,0 +1,15 @@
...
+    task = JupyterTask(
+        "tests/tasks/jupyter/sample_notebook.ipynb",
```
madkinsz, line 10: Did you leave this file out? I think I'd prefer if you used the pytest `tmpdir` fixture and wrote this file to disk for the test.

madkinsz (changes requested):
Hi! This looks like a good start, thanks for your contribution :) 

A few comments! Mostly looks like it just needs some polish and test improvements.

jcrist:
Since this executes a Jupyter notebook using papermill, should this be `prefect.tasks.papermill.PapermillTask` instead?

tests/tasks/jupyter/test_jupyter.py
bachsh: Not sure how to use the `tmpdir` fixture for this. Can you point me to an example? And where should I put the file if I do that

bachsh:
> Since this executes a Jupyter notebook using papermill, should this be `prefect.tasks.papermill.PapermillTask` instead?

@jcrist what the task does is execute Jupyter notebooks, and since not many people know what papermill is I named it Jupyter. Otherwise it would be similar to naming the RSS tasks `FeedparserTask`. WDYT?

tests/tasks/jupyter/test_jupyter.py
bachsh: added a test for the JSON case

bachsh:
Tests are failing because the requirements for the new task are not installed, how can I get them installed?

tests/tasks/jupyter/test_jupyter.py
madkinsz: Now that you've attached the file, it's small enough that I don't mind it being in the repo although it's possible someone else will have a different opinion. Unless someone else thinks this would be better, don't worry about it. Anyway, here's an example of the tmpdir fixture for future reference: https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture

madkinsz:
> Tests are failing because the requirements for the new task are not installed, how can I get them installed?

They are being installed for some of the tests but not others where we explicitly exclude extras. Take a look at https://github.com/PrefectHQ/prefect/blob/master/tests/tasks/dropbox/__init__.py

tests/tasks/jupyter/test_jupyter.py
bachsh: The fixture is cool, thanks!
Nevertheless, I'd still need a place to put the actual data, and then load it into the fixture, wouldn't I?

bachsh:
> > Tests are failing because the requirements for the new task are not installed, how can I get them installed?
> 
> They are being installed for some of the tests but not others where we explicitly exclude extras. Take a look at https://github.com/PrefectHQ/prefect/blob/master/tests/tasks/dropbox/__init__.py

Ohh I get it. I added the relevant part but the skipping command appeared after the part where I import my task, so the import failed 👍

tests/tasks/jupyter/test_jupyter.py
madkinsz: You'd put it in a string literal in the test which would make it faster for someone to see what's being run in the notebook. That said, the file is a little long for that so your solution is probably ideal :) There's probably  a function (in jupyter?) that generates most of the notebook metadata for you so it'd be more concise but I don't see a reason to try to track that down.

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -0,0 +1,66 @@
+import nbconvert
+import nbformat
+import papermill as pm
+
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class JupyterTask(Task):
```
jcrist, line 9: We're moving away from tasks with `Task` suffixes (which feel redundant). Perhaps `RunNotebook` or `ExecuteNotebook` instead?

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -0,0 +1,66 @@
...
+    In order to parametrize the notebook, you need to mark the parameters cell as described in
+        the papermill documentation: https://papermill.readthedocs.io/en/latest/usage-parameterize.html
+
+    Args:
+        - notebook_path (string, optional): path to fetch the notebook from.
+            Can be a cloud storage path.
+            Can also be provided post-initialization by calling this task instance
+        - notebook_params (dict, optional): dictionary of parameters to use for the notebook
+            Can also be provided at runtime
+        - as_html (bool, optional): whether to output as HTML. (default: True)
```
jcrist, line 21: Instead of using `as_html`, perhaps make this a string category so other nbconvert formats could be used (markdown, latex, etc...)? For now you could only support the two you currently support (json and html), but it leaves an opportunity for expansion later where a boolean flag doesn't. Additionally, I think the default should be `json`, as that allows more flexibility for downstream tasks to postprocess the results if needed (and most users are familiar with executed notebooks as json, while fewer may be familiar with them as html).


```python
output_type="json"
```

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -0,0 +1,66 @@
...
+    Task for running Jupyter Notebooks.
+    In order to parametrize the notebook, you need to mark the parameters cell as described in
+        the papermill documentation: https://papermill.readthedocs.io/en/latest/usage-parameterize.html
+
+    Args:
+        - notebook_path (string, optional): path to fetch the notebook from.
+            Can be a cloud storage path.
+            Can also be provided post-initialization by calling this task instance
+        - notebook_params (dict, optional): dictionary of parameters to use for the notebook
+            Can also be provided at runtime
```
jcrist, line 20: I think the `notebook_` prefix here is redundant, it's already a notebook task. I'd suggest switching to `path` and `parameters` (not `params`) would be clearer.

jcrist (commented):
Thanks @bachsh , just a few cleanups on the interface and this looks good to me.

bachsh:
> Thanks @bachsh , just a few cleanups on the interface and this looks good to me.

Thanks for the review @jcrist . Code and tested were updated following your comments.

jcrist (approved):
Thanks!

PR #3600, opened 2020-10-30
Log Record UTC Offset & Timezone Display

### LINKED ISSUE #3556: Prefect logger doesn't use local Python logger timezone configuration

**dcshah88**: Hello Everyone, 

Have started looking into prefect core library recently and exploring it to use for simple use cases.

One thing I am struggling with right now is to provide a different timezone to prefect instead of default UTC.

By default, prefect is generating UTC timestamps in logs but wanted to change it to my local timezone. 

Could you please help ?

**znicholasbrown**: Hi <@U01CWF05D6Y> - Prefect uses Python's built-in logging module; please look at the <https://docs.python.org/3/library/logging.html|Python logging documentation> for info on how to configure your local Python logger.

**dcshah88**: Thanks <@UN6FTLFAS> : what I am confused about is, if I use python logging without prefect, it prints logs using local timezone by default but when used with prefect it prints timestamps in UTC.

looks like prefect is overriding default python logging behavior.

So was wondering if prefect has any config option to provide which timezone to use while printing logs ?

**znicholasbrown**: Ah, that's a good flag - it looks like we're explicitly defaulting to UTC instead of local in the logger, which is probably unintentional. Let me see if there's a reason for that and if not we can PR it :slightly_smiling_face:

**znicholasbrown**: I think this is something we can open a ticket for <@U01CWF05D6Y> - <https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/logging.py#L220|this line> is where the timezone is set in the logger; I'll use <@ULVA73B9P> to open the ticket from this thread :slightly_smiling_face:

**znicholasbrown**: <@ULVA73B9P> open "Prefect logger doesn't use local Python logger timezone configuration"

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1603234095357600?thread_ts=1603234095.357600&cid=CL09KU1K7).

speedyturkey:
## Summary

This PR adds some useful configuration for log timezone display.

## Changes

1. Adds UTC offset (`%z`) to the default `logging.datefmt` setting.
2. Adds new attribute `logging.display_tz` with default value of UTC, which allows user to set the timezone used in formatted log records.

## Importance

The UTC offset adds useful, explicit information to log records. The `logging.display_tz` attribute may be appreciated by some users, particularly for local development and testing, since UTC is sometimes difficult for humans to think in.

It is possible that the adding the UTC offset is a good enough fix for the original issue. If the reviewer(s) think the extra complexity of a custom formatter is not worth it for this feature, I'm definitely open to that conversation. Thanks for considering my PR!

## Checklist

This PR:

Closes #3556

### DIFF
--- a/src/prefect/config.toml
+++ b/src/prefect/config.toml
@@ -93,6 +93,9 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 # e.g., log_attributes = "['context_var']"
 log_attributes = "[]"
 
+# name of the timezone in which to display log timestamps
+display_tz = "UTC"
+
 # the timestamp format
 datefmt = "%Y-%m-%d %H:%M:%S%z"
 
--- a/src/prefect/config.toml
+++ b/src/prefect/config.toml
@@ -94,7 +94,7 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 log_attributes = "[]"
 
 # the timestamp format
-datefmt = "%Y-%m-%d %H:%M:%S"
+datefmt = "%Y-%m-%d %H:%M:%S%z"
 
 # Send logs to Prefect Cloud
 log_to_cloud = false
--- a/src/prefect/config.toml
+++ b/src/prefect/config.toml
@@ -93,7 +93,8 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 # e.g., log_attributes = "['context_var']"
 log_attributes = "[]"
 
-# name of the timezone in which to display log timestamps
+# name of the timezone in which to display log timestamps, in tzdata/IANA format
+# For list of valid timezones, see `pytzdata.timezones`
 display_tz = "UTC"
 
 # the timestamp format
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -45,20 +45,28 @@ class PrefectFormatter(logging.Formatter):
         method for display in log records.
     """
 
-    def __init__(self, *args, display_tz: str = "UTC", **kwargs) -> None:
+    def __init__(self, *args: Any, display_tz: str = "UTC", **kwargs: Any) -> None:
         self.display_tz = display_tz
         super().__init__(*args, **kwargs)
 
     def converter(self, timestamp: float) -> pendulum.DateTime:
+        """
+        Return timezone-aware timestamp using specified timezone.
+        Default `converter` implementation returns a `struct_time` object,
+        which does not readily support timezones.
+        """
         return pendulum.from_timestamp(timestamp, tz=self.display_tz)
 
-    def formatTime(self, record, datefmt=None) -> str:
+    def formatTime(self, record: logging.LogRecord, datefmt: str = None) -> str:
+        """
+        Return the creation time of the specified LogRecord as formatted text.
+
+        Differs from `logging.Formatter` implementation due to the need to
+        call `datetime.strftime` instead of `time.strftime`.
+        """
         dt = self.converter(record.created)
-        if datefmt:
-            s = dt.strftime(datefmt)
-        else:
-            s = dt.isoformat()
-        return s
+        formatted_time = dt.strftime(datefmt) if datefmt else dt.isoformat()
+        return formatted_time
 
 
 class CloudHandler(logging.StreamHandler):
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -33,16 +33,44 @@ PREFECT_LOG_RECORD_ATTRIBUTES = (
 )
 
 
+class PrefectFormatter(logging.Formatter):
+    """
+    Custom formatter implementing timezone localization.
+    Creates a logger with a `StreamHandler` that has level and formatting
+    set from `prefect.config`.
+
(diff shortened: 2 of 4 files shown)


src/prefect/utilities/logging.py
```diff
@@ -33,16 +33,44 @@
...
+    def converter(self, timestamp: float) -> pendulum.DateTime:
+        return pendulum.from_timestamp(timestamp, tz=self.display_tz)
+
+    def formatTime(self, record, datefmt=None) -> str:
+        dt = self.converter(record.created)
+        if datefmt:
+            s = dt.strftime(datefmt)
+        else:
+            s = dt.isoformat()
+        return s
```
madkinsz, line 60: I'd generally try to avoid single letter variables and take advantage of Python's one-lining for the conditional
```suggestion
        timestr = dt.strftime(datefmt) if datefmt else dt.isoformat()
        return timestr
```
I'm not sure `timestr` is the best name choice so feel free to choose something better yet :)

src/prefect/config.toml
```diff
@@ -93,8 +93,11 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 # e.g., log_attributes = "['context_var']"
 log_attributes = "[]"
 
+# name of the timezone in which to display log timestamps
```
madkinsz, line 96: A note here about which timezone name format we expect would be nice

src/prefect/utilities/logging.py
```diff
@@ -33,16 +33,44 @@
...
+    """
+
+    def __init__(self, *args, display_tz: str = "UTC", **kwargs) -> None:
+        self.display_tz = display_tz
+        super().__init__(*args, **kwargs)
+
+    def converter(self, timestamp: float) -> pendulum.DateTime:
+        return pendulum.from_timestamp(timestamp, tz=self.display_tz)
+
+    def formatTime(self, record, datefmt=None) -> str:
```
madkinsz, line 54: A docstring may be nice explaining how this specifically differs from the parent class' implementation. For reference for other reviewers

```python
        ct = self.converter(record.created)
        if datefmt:
            s = time.strftime(datefmt, ct)
        else:
            t = time.strftime(self.default_time_format, ct)
            s = self.default_msec_format % (t, record.msecs)
        return s
```

madkinsz (commented):
Thanks @speedyturkey, this looks like a good start! We'll have to wait for a maintainer to take a look and make a decision about the UTC offset but I only have minor comments -- the functionality looks solids.

Also note you're failing our static analysis:

```
src/prefect/utilities/logging.py:47: error: Function is missing a type annotation for one or more arguments
src/prefect/utilities/logging.py:51: error: Argument 1 of "converter" is incompatible with supertype "Formatter"; supertype defines the argument type as "Optional[float]"
src/prefect/utilities/logging.py:51: note: This violates the Liskov substitution principle
src/prefect/utilities/logging.py:51: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
src/prefect/utilities/logging.py:51: error: Return type "DateTime" of "converter" incompatible with return type "struct_time" in supertype "Formatter"
src/prefect/utilities/logging.py:54: error: Function is missing a type annotation for one or more arguments
```

speedyturkey:
@madkinsz Thank you so much for the detailed and thoughtful review. I hope the changes are responsive to your feedback, and static analysis is now passing. I did need to do one `# type: ignore` because the whole point of the new `converter` method is to change the return type.

I do see that a test failed but it seems to be something to do with DaskExecutors and the test environment itself, rather than any code changes I made.

madkinsz:
> @madkinsz Thank you so much for the detailed and thoughtful review. I hope the changes are responsive to your feedback, and static analysis is now passing. I did need to do one `# type: ignore` because the whole point of the new `converter` method is to change the return type.

You're welcome. Your changes look great! I think you've addressed everything I mentioned quite well. I especially appreciate the documentation you added. Now we'll just have to wait for a maintainer to take a look!

> I do see that a test failed but it seems to be something to do with DaskExecutors and the test environment itself, rather than any code changes I made.

I see that, I'll restart that round of tests. Some of them can be a bit flaky.

madkinsz (approved):

jcrist:
The implementation here looks good to me, but I wonder if a more intuitive fix (without requiring extra config variables) would be to:

- Always display logs in local time with the default `StreamHandler`, matching python's default logging behavior. Logs with `flow.run()` would show up in the local time.
- Add the timezone formatting to cloud logs (in the `CloudHandler`) to add the UTC offset, skipping the timezone conversion for display. The UI at some point could try to localize these on display to the user's timezone, but we'd always store them as UTC.

This avoids the addition of an extra config variable, and does what I think most users would want.

jcrist:
Yeah, on further thought I think a simpler fix would be:

- Add the `%z` format to the default datetime format
- Delete this line: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/logging.py#L220%7Cthis

Local logs would display in the local timezone, but with a UTC offset added for localization, and cloud logs would still work as is.

speedyturkey:
Thanks @madkinsz and @jcrist. That makes a ton of sense and I agree with your fix. I'm closing this PR and have opened #3607.

PR #3607, opened 2020-11-02
Log Record UTC Offset & Converter

### LINKED ISSUE #3556: Prefect logger doesn't use local Python logger timezone configuration

speedyturkey:
## Summary

This PR makes logging timestamp format and behavior easier to understand and better aligned with default Python behavior. This PR is an update to #3600, incorporating feedback received.

## Changes

1. Adds UTC offset (%z) to the default logging.datefmt setting.
2. Removes override of `logging.Formatter.converter` so that `time.localtime` is used instead of `time.gmtime`.

## Importance

The UTC offset adds useful, explicit information to log records. The logging.display_tz attribute may be appreciated by some users, particularly for local development and testing, since UTC is sometimes difficult for humans to think in.

Removing the `logging.Formatter.converter` override will lead to more expected behavior locally, and will not change cloud logging functionality.

## Checklist

This PR:

Closes #3556.

### DIFF
--- a/changes/pr3607.yaml
+++ b/changes/pr3607.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+enhancement:
+  - "Add UTC offset to default logging.datefmt; logging timestamp converter now follows Python default behavior  - [#3607](https://github.com/PrefectHQ/prefect/pull/3607)"
+
+contributor:
+  - "[Billy McMonagle](https://github.com/speedyturkey)"
--- a/src/prefect/config.toml
+++ b/src/prefect/config.toml
@@ -94,7 +94,7 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
 log_attributes = "[]"
 
 # the timestamp format
-datefmt = "%Y-%m-%d %H:%M:%S"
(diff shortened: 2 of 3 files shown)


madkinsz:
🤦 glad it was this easy. Thanks again for the effort you've put into this @speedyturkey !

Kudos to @jcrist for finding the simplest fix.

speedyturkey:
The work that ended up not being used was still interesting and worth doing for me - I learned a few things about `logging`, `pytest` and `mypy`, but I agree with @jcrist that this is the simplest and best fix.

jcrist (approved):

jcrist:
Thanks @speedyturkey!

PR #3610, opened 2020-11-02
Restructure prefect agent cli

jcrist:
- Deprecates `prefect agent start <kind>` in favor of `prefect agent <kind> start`
- Deprecates `prefect agent install <kind>` in favor of `prefect agent <kind> install`
- Updates the documentation to use the new commands only
- Adds the new agent commands to the api docs
- Updates the test suite to provide better coverage of the agent CLI
- Fix a few bugs in the agent CLI

---

### DIFF
--- a/src/prefect/cli/agent.py
+++ b/src/prefect/cli/agent.py
@@ -4,41 +4,386 @@ from prefect import config
 from prefect.utilities.configuration import set_temporary_config
 from prefect.utilities.serialization import from_qualified_name
 
+COMMON_START_OPTIONS = [
+    click.option(
+        "--token",
+        "-t",
+        required=False,
+        help="A Prefect Cloud API token with RUNNER scope.",
+    ),
+    click.option("--api", "-a", required=False, help="A Prefect API URL."),
+    click.option(
+        "--agent-config-id",
+        help="An agent ID to link this agent instance with",
+    ),
+    click.option(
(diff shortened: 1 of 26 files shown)


src/prefect/cli/agent.py
```diff
@@ -4,41 +4,386 @@
...
+
+    start_agent(ECSAgent, **kwargs)
+
+
+#############################
+# Deprecated Agent Commands #
+#############################
+
+_agents = {
+    "fargate": "prefect.agent.fargate.FargateAgent",
```
cicdw, line 382: Since it seems `prefect agent start fargate` is still supported, would this be a good time to transition this directly to `prefect agent ecs start` instead of going through a second deprecation?

cicdw (commented):
I have one question about the transition from `fargate` -> `ecs` but otherwise this LGTM!

src/prefect/cli/agent.py
jcrist: I don't want to start raising a deprecation warning about the fargate agent until we're sure the ECS agent is 100% ready to take over all user's use cases. I think 2 deprecation cycles is fine:

- start warning users over to the new CLI pattern, which should be easy on their end
- once docs are up, start warning users over to the ECS agent, which is a much more involved move

cicdw (approved):

PR #3612, opened 2020-11-03
pass day_or croniter argument to CronClock & CronSchedule

sp1thas:
## Summary

The purpose of this PR is to allow to `day_or` croniter on `CronClock` and `CronSchedule` instantiation.

## Changes

This PR changes the `CronClockSchema` and `CronScheduleSchema` by adding the new `day_or` field. `day_or` has been added to `CronClock.__init__` and `CronSchedule`. Some tests for have also been added.

## Importance

This change makes every cron clock more configurable, therefore, the whole procedure of scheduling can become more flexible.

This PR was inspired by the following case:

Let's suppose that we have a scheduler with multiple clocks. We want to add a clock that creates events every first Thursday of the month, this periodicity can be achived by this cron: `0 0 1-7 * 4`.

Unfortunately, this cron will not work out of the box:
```
>>> import datetime
>>> import croniter
>>> cron = croniter.croniter('0 0 1-7 * 4')
>>> cron.get_next(datetime.datetime)
datetime.datetime(2020, 11, 1, 0, 0)
>>> cron.get_next(datetime.datetime)
datetime.datetime(2020, 11, 2, 0, 0)
>>> cron.get_next(datetime.datetime)
datetime.datetime(2020, 11, 3, 0, 0)
```

because of legacy/problematic behavior: [so link](https://stackoverflow.com/questions/34357126/why-crontab-uses-or-when-both-day-of-month-and-day-of-week-specified).

To fix this issue, `day_or=False` should be pass on initialization.

```
>>> cron = croniter.croniter('0 0 1-7 * 4', day_or=False)
>>> cron.get_next(datetime.datetime)
datetime.datetime(2020, 11, 5, 0, 0)
>>> cron.get_next(datetime.datetime)
datetime.datetime(2020, 12, 3, 0, 0)
```

## Checklist

This PR:

### DIFF
--- a/src/prefect/schedules/clocks.py
+++ b/src/prefect/schedules/clocks.py
@@ -236,8 +236,11 @@ class CronClock(Clock):
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
-        - croniter_kwargs (dict, optional): an optional dictionary with extra croniter
-            keyword arguments
+        - day_or (bool, optional): Control how croniter handles `day` and `day_of_week` entries.
+            Default option is the cron behaviour, which connects those values using OR.
+            If the switch is set to False, the values are connected using AND. This behaves like
+            fcron and enables you to e.g. define a job that executes each 2nd friday of a month
+            by setting the days of month and the weekday.
 
     Raises:
         - ValueError: if the cron string is invalid
@@ -250,13 +253,13 @@ class CronClock(Clock):
         end_date: datetime = None,
         parameter_defaults: dict = None,
         labels: List[str] = None,
-        croniter_kwargs: dict = None,
+        day_or: bool = None,
     ):
         # build cron object to check the cron string - will raise an error if it's invalid
         if not croniter.is_valid(cron):
             raise ValueError("Invalid cron string: {}".format(cron))
         self.cron = cron
-        self.croniter_kwargs = croniter_kwargs if croniter_kwargs else {}
+        self.day_or = day_or
         super().__init__(
             start_date=start_date,
             end_date=end_date,
@@ -307,7 +310,7 @@ class CronClock(Clock):
         if after_localized.microsecond:
             after_localized = after_localized + timedelta(seconds=1)
 
-        cron = croniter(self.cron, after_localized, **self.croniter_kwargs)
+        cron = croniter(self.cron, after_localized, day_or=self.day_or)
         dates = set()  # type: Set[datetime]
 
         while True:
--- a/src/prefect/schedules/clocks.py
+++ b/src/prefect/schedules/clocks.py
@@ -236,6 +236,8 @@ class CronClock(Clock):
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
+        - croniter_kwargs (dict, optional): an optional dictionary with extra croniter
+            keyword arguments
 
     Raises:
         - ValueError: if the cron string is invalid
--- a/src/prefect/schedules/clocks.py
+++ b/src/prefect/schedules/clocks.py
@@ -248,11 +248,13 @@ class CronClock(Clock):
         end_date: datetime = None,
         parameter_defaults: dict = None,
         labels: List[str] = None,
+        croniter_kwargs: dict = None,
     ):
         # build cron object to check the cron string - will raise an error if it's invalid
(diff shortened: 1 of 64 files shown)


jcrist (commented):
Thanks @sp1thas. Looking at the `croniter` docs, it looks like the only kwarg to `croniter` that it makes sense for us to expose is `day_or` - perhaps adding that as a top-level kwarg to `CronClock`/`CronSchedule` would make more sense than exposing all of cron-iter's kwargs?

Other than that, this seems fine to me. cc'ing @cicdw for a 2nd glance-over just to be sure (schedules also need to be supported by `cloud`, so there's some extra nuance here when making changes to them).

cicdw:
Yup this can be supportable by Cloud (once the serializer changes have been released) with no additional work needed as all of the work is handled by the (de)serialization of the schedule / clock itself. 

I think I agree with you @jcrist that `day_or` might be better than general kwargs; if they ever introduce new kwargs it could cause confusion between the version of `croniter` running in Cloud vs locally so an explicit list seems better (and allows people to discover the functionality easier).

src/prefect/schedules/clocks.py
```diff
@@ -303,7 +310,7 @@ def events(self, after: datetime = None) -> Iterable[ClockEvent]:
         if after_localized.microsecond:
             after_localized = after_localized + timedelta(seconds=1)
 
-        cron = croniter(self.cron, after_localized)
+        cron = croniter(self.cron, after_localized, day_or=self.day_or)
```
jcrist, line 313: Looks like mypy is failing due to an inaccurate type definition for `croniter`, can just ignore types here

```suggestion
        cron = croniter(self.cron, after_localized, day_or=self.day_or)  # type: ignore
```

src/prefect/schedules/clocks.py
```diff
@@ -236,6 +236,11 @@ class CronClock(Clock):
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
+        - day_or (bool, optional): Control how croniter handles `day` and `day_of_week` entries.
+            Default option is the cron behaviour, which connects those values using OR.
```
jcrist, line 240:
```suggestion
            Defaults to True, matching cron which connects those values using OR.
```

src/prefect/schedules/schedules.py
```diff
@@ -255,14 +258,19 @@ def CronSchedule(
         - cron (str): a valid cron string
         - start_date (datetime, optional): an optional start date for the clock
         - end_date (datetime, optional): an optional end date for the clock
+        - day_or (bool, optional): Control how croniter handles `day` and `day_of_week` entries.
+            Default option is the cron behaviour, which connects those values using OR.
```
jcrist, line 262:
```suggestion
            Defaults to True, matching cron which connects those values using OR.
```

src/prefect/schedules/clocks.py
```diff
@@ -248,11 +253,13 @@ def __init__(
         end_date: datetime = None,
         parameter_defaults: dict = None,
         labels: List[str] = None,
+        day_or: bool = None,
     ):
         # build cron object to check the cron string - will raise an error if it's invalid
         if not croniter.is_valid(cron):
             raise ValueError("Invalid cron string: {}".format(cron))
         self.cron = cron
+        self.day_or = day_or
```
jcrist, line 262: We'll want to default to `True` to match the existing behavior. Right now with it set to `None` croniter will interpret that as `False`. We want to do this within the code (rather than as the default kwarg) so that our serializers do the right thing if the value is missing.

```suggestion
        self.day_or = True if day_or is None else day_or
```

tests/schedules/test_clocks.py
```diff
@@ -445,6 +449,20 @@ def test_cron_clock_events_with_after_argument(self):
             start_date.add(days=3),
         ]
 
+    def test_cron_clock_events_with_day_or(self):
```
jcrist, line 452: It would be good to test with `day_or` as `None` and `True` as well - this would have caught the bug that was introduced above. You might use `pytest.mark.parametrize` to handle both cases with the same test code.

jcrist (changes requested):

tests/schedules/test_clocks.py
sp1thas: Good catch! On it

jcrist (approved):
Thanks! This looks good to me.

PR #3615, opened 2020-11-04
Better GraphQL error message

cicdw:
## Summary

Many users experience 400 status codes when issuing client requests to the Prefect API.  This is _almost_ always caused by a bad GraphQL request (schema mismatch, bad type, etc.) and would be much easier to debug if the GraphQL was displayed.  This also allows users to copy / paste the GraphQL when we support.

## Changes

Captures 400 status codes and expands the error message.

## Importance

Easier debugging experience.

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -340,7 +340,7 @@ class Client:
                 raise ClientError(msg)
             except ClientError:
                 raise
-            except:
+            except Exception:
                 pass
         response.raise_for_status()
         return response
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -330,6 +330,18 @@ class Client:
             )
 
         # Check if request returned a successful status
+        if response.status_code == 400:
+            try:
+                msg = (
+                    f"400 Client Error: Bad Request for url: {url}\n\n"
+                    f"This is likely caused by a poorly formatted GraphQL query or mutation. "
+                    f"GraphQL sent:\n\n{parse_graphql(params)}"
+                )
+                raise ClientError(msg)
+            except ClientError:
+                raise
+            except:
+                pass
         response.raise_for_status()
         return response
 
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -331,17 +331,17 @@ class Client:
 
         # Check if request returned a successful status
         if response.status_code == 400:
+            msg = (
+                f"400 Client Error: Bad Request for url: {url}\n\n"
+                f"This is likely caused by a poorly formatted GraphQL query or mutation."
+            )
             try:
-                msg = (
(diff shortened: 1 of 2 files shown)


src/prefect/client/client.py
```diff
@@ -330,6 +330,18 @@ def _send_request(
...
+                msg = (
+                    f"400 Client Error: Bad Request for url: {url}\n\n"
+                    f"This is likely caused by a poorly formatted GraphQL query or mutation. "
+                    f"GraphQL sent:\n\n{parse_graphql(params)}"
+                )
+                raise ClientError(msg)
+            except ClientError:
+                raise
+            except Exception:
+                pass
```
madkinsz, line 344: Would this 400 be handled by the caller then? Seems dangerous to silence exceptions in the message creation bit.

src/prefect/client/client.py
```diff
@@ -330,6 +330,18 @@ def _send_request(
...
 
         # Check if request returned a successful status
+        if response.status_code == 400:
+            try:
+                msg = (
+                    f"400 Client Error: Bad Request for url: {url}\n\n"
+                    f"This is likely caused by a poorly formatted GraphQL query or mutation. "
+                    f"GraphQL sent:\n\n{parse_graphql(params)}"
+                )
+                raise ClientError(msg)
```
madkinsz, line 340: Any consideration to attach additional metadata directly to ClientErrors then just have a robust string formatter within the class?

src/prefect/client/client.py
jcrist, line 344: I find the logic above confusing (we try to raise an error, and ignore the error if it's raised properly). Under what reasons would the formatting fail, and if it does is there another nice error message we could give rather than the one raised by `raise_for_status()`? If not,  falling back to `response.raise_for_status()` seems fine to me - I'd find the following logic cleaner though:

```suggestion
        if response.status_code == 400:
            try:
                msg = (
                    f"400 Client Error: Bad Request for url: {url}\n\n"
                    f"This is likely caused by a poorly formatted GraphQL query or mutation. "
                    f"GraphQL sent:\n\n{parse_graphql(params)}"
                )
            except Exception:
                # failed to format, raise as normal below
                pass
            else:
                raise ClientError(msg)
```

cicdw: In what sense is that dangerous?  The message is intended to add more info than already is presented, but if for some reason that's not possible it falls back to the original behavior

cicdw: Makes sense - will update

src/prefect/client/client.py
cicdw: Not sure I see any benefit to that

jcrist (approved):

PR #3630, opened 2020-11-09
Skip project creation if the project already exists

amanda-wee:
## Summary

Change `Client.create_project` and `prefect create project`to skip creating the project if it already exists, returning/printing as if the project had been created

## Changes

- Change `Client.create_project` such that if the project to be created already exists, then it has no net effect and the id of the existing project is returned.
- Change `prefect create project` to regard an attempt to create a project that already exists as successful, skipping the creation of the project.

## Importance

Enables users to easily automate project creation (e.g., at startup in an ECS task) while continuing to receive errors unrelated to the project having been previously created. This meshes nicely with the introduction of the idempotency key for `flow.register` and `flow.serialized_hash()` in 0.13.14.

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -929,7 +929,7 @@ class Client:
             - skip_if_exists (bool, optional): skip creating the project if it already exists
 
         Returns:
-            - str: the ID of the newly-created project or an empty string if it already exists
+            - str: the ID of the newly-created or pre-existing project
 
         Raises:
             - ClientError: if the project creation failed
@@ -953,7 +953,15 @@ class Client:
             )  # type: Any
         except ClientError as exc:
             if skip_if_exists and "'Uniqueness violation.'" in str(exc):
-                return ""
+                project_query = {
+                    "query": {
+                        with_args(
+                            "project", {"where": {"name": {"_eq": project_name}}}
+                        ): {"id": True}
+                    }
+                }
+                res = self.graphql(project_query)
+                return res.data.project[0].id
             raise
 
         return res.data.create_project.id
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -914,16 +914,22 @@ class Client:
             slug = res.get("data").tenant[0].slug
         return slug
 
-    def create_project(self, project_name: str, project_description: str = None) -> str:
+    def create_project(
+        self,
+        project_name: str,
+        project_description: str = None,
+        skip_if_exists: bool = False,
+    ) -> str:
         """
         Create a new Project
 
         Args:
             - project_name (str): the project that should contain this flow
             - project_description (str, optional): the project description
+            - skip_if_exists (bool, optional): skip creating the project if it already exists
 
         Returns:
-            - str: the ID of the newly-created project
+            - str: the ID of the newly-created project or an empty string if it already exists
 
         Raises:
             - ClientError: if the project creation failed
@@ -934,16 +940,21 @@ class Client:
             }
         }
 
-        res = self.graphql(
-            project_mutation,
-            variables=dict(
-                input=dict(
-                    name=project_name,
-                    description=project_description,
-                    tenant_id=self.active_tenant_id,
-                )
-            ),
-        )  # type: Any
+        try:
+            res = self.graphql(
+                project_mutation,
+                variables=dict(
+                    input=dict(
+                        name=project_name,
+                        description=project_description,
+                        tenant_id=self.active_tenant_id,
+                    )
+                ),
+            )  # type: Any
+        except ClientError as exc:
+            if skip_if_exists and "'Uniqueness violation.'" in str(exc):
+                return ""
+            raise
 
         return res.data.create_project.id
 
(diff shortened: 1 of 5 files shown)


src/prefect/client/client.py
```diff
@@ -934,16 +940,21 @@ def create_project(self, project_name: str, project_description: str = None) ->
...
+                variables=dict(
+                    input=dict(
+                        name=project_name,
+                        description=project_description,
+                        tenant_id=self.active_tenant_id,
+                    )
+                ),
+            )  # type: Any
+        except ClientError as exc:
+            if skip_if_exists and "'Uniqueness violation.'" in str(exc):
```
joshmeek, line 955: I think instead of catching this on error we should instead make a query for a matching project with that name and if it exists then return there instead of having to rely on a substring in the exception raised. Something like:

```python
if skip_if_exists:
  query_project = {
            "query": {
                with_args("project", {"where": {"name": {"_eq": project_name}}}): {
                    "id": True
                }
            }
        }
   project = self.graphql(query_project).data.project
   if project:
       return None
```

src/prefect/client/client.py
```diff
@@ -914,16 +914,22 @@ def get_default_tenant_slug(self, as_user: bool = True) -> str:
...
         Create a new Project
 
         Args:
             - project_name (str): the project that should contain this flow
             - project_description (str, optional): the project description
+            - skip_if_exists (bool, optional): skip creating the project if it already exists
 
         Returns:
-            - str: the ID of the newly-created project
+            - str: the ID of the newly-created project or an empty string if it already exists
```
joshmeek, line 932: Let's change this to returning None instead of an empty string. Will need to update the type of the function from 

```
-> str
```
to
```
-> Optional[str]
```

src/prefect/client/client.py
jcrist: That leaves a possibility for a race condition if multiple commands happen to be running concurrently - while parsing the error message is unfortunate I think the existing logic is less error prone.

src/prefect/client/client.py
jcrist: I think we should return the id regardless - if it already exists we could lookup and return the existing ID.

Thinking about how this might be used, I think most users probably want this function to ensure a project exists, and are less likely to need to branch based on if the project exists already or not. Being consistent in return value (always returning the project id) makes it easier for the user to write their logic without having to check for a `None` result.

We might even make the idempotency the default (or only behavior)? I can't think of a reason I'd want to have `create_project` error if the project already existed.

src/prefect/client/client.py
joshmeek: 👍

src/prefect/client/client.py
amanda-wee: I have changed it to always return the id, so the return type hint remains `str`.

I like the idea of making idempotency the only behaviour. Shall I go ahead with that?

amanda-wee: Also, I have no idea why the code style checks are failing: both `black .` and `mypy src` are fine for me.

amanda-wee: Ah okay, I checked the circleci config and figured out that `flake8 src` was being run, and turns out I needed to fix something with that. I think `dev-requirements.txt` might need to include `flake8` and the contributing docs might need to be updated to mention running `flake8 src`.

jcrist:
> I like the idea of making idempotency the only behaviour. Shall I go ahead with that?

I also like that, and can't think of a downside. There's a complexity cost to adding more options, so I'm always for things that remove knobs we might not need. Would want to get buy-in from another maintainer though beforehand (cc @joshmeek).

joshmeek:
@jcrist Yeah that sounds good to me 👍

amanda-wee:
Great, I have made a commit that does that. Shall update the PR accordingly.

amanda-wee:
Hmm... it looks like there's an error detected by `tests/utilities/test_logging.py`. I tried going back to a commit where the tests pass though, and they still fail for me locally. Any idea what to do about this? It looks unrelated:
```
caplog = <_pytest.logging.LogCaptureFixture object at 0x7f39c520f970>, datefmt = '%Y -- %D'

    @pytest.mark.parametrize("datefmt", ["%Y", "%Y -- %D"])
    def test_root_logger_datefmt_responds_to_config(caplog, datefmt):
        try:
            with utilities.configuration.set_temporary_config({"logging.datefmt": datefmt}):
                logger = utilities.logging.configure_logging(testing=True)
                logger.error("badness")
                logs = [r for r in caplog.records if r.levelname == "ERROR"]
>               assert logs[0].asctime == datetime.datetime.utcnow().strftime(datefmt)
E               AssertionError: assert '2020 -- 11/11/20' == '2020 -- 11/10/20'
E                 - 2020 -- 11/10/20
E                 ?             ^
E                 + 2020 -- 11/11/20
E                 ?             ^
```

jcrist:
Don'y worry about that - looks like we have a flaky test, reran and things passed as intended. Things look good to me, thanks @amanda-wee!

jcrist (approved):

PR #3649, opened 2020-11-11
Add `stream_output` and a result on failure to `ShellTask`

### LINKED ISSUE #3584: shellTask logs all stdout when `ShellTask(return_all=True)`
**Context**
Currently, when `prefect.tasks.shell.ShellTask` is invoked with `return_all=True`, the shell task will not log the stdout through logger.debug.

This is defined in doc 
https://docs.prefect.io/api/latest/tasks/shell.html#shelltask
and implement in the code.
https://github.com/PrefectHQ/prefect/blob/73fe1eb57945fc13bdd36b361a6a17c3f324b1ba/src/prefect/tasks/shell.py#L107-L111

**Problem**

Sometimes, we execute a long-running shell process, for example, we are running dbt, which might take 20 to 40 minutes to run. the task would need to return the entire stdout log as its output, for the next task to parse and interpret the result. Meanwhile, We would also like to see the shell stdout logging out in real-time, which makes it easy for us to inspect issues and monitoring / debug shell runs on-the-go, instead of waiting for 20 minutes not knowing what is going on.

**proposed change**

```
                for raw_line in iter(sub_process.stdout.readline, b""):
                    line = raw_line.decode("utf-8").rstrip()
                    if self.return_all:
                        lines.append(line)
                    # we still log every line
                    self.logger.debug(line)
```

### LINKED ISSUE #3614: allow ShellTask to return stdout when it fails
**The context**
Currently, if the shell process returns non-zero exit code, It will raise a FAIL signal, without any result attached. 

https://github.com/PrefectHQ/prefect/blob/b939c7d043f5de910a54ad95e07873a835f48b31/src/prefect/tasks/shell.py#L113-L122

**The probem**
Sometime, the downstream task need to interpret the output of a failed shell task (might be from both stdout and stderr) to handle the the failure properly. 

In my use case, I use DBTShellTask (which is a subclass of ShellTask) for execute a  `dbt run` as shell sub-process. I then use another task, called `parse_dbt_result`, to take the output of the dbt task and handle both success and failure results, such as knowing which models in the dbt-run process failed.

When some models fails in a `dbt run`, those model errors are shown in the stdout, and  the shell subprocess returns non-zero exit code (1). The ShellTask raises a FAIL signal without any stdout result.  As a result, the `parse_dbt_result` task is not able to interpret the log of failed run and not know which model fails, hereby not be able to handle the failure properly. 

**Solution**
please provide a flag that allows ShellTask to return stdout(or/and stderror) result when the shell process fails and FAIL signal is raised.

madkinsz:
## Summary

Adds some better handling of output to `ShellTask`

Closes #3614
Closes #3584 

## Changes

- Adds the output result to the `FAIL` signal state for inspection on failure cases
- Adds notes about inconsistencies / handling that need to be fixed but are breaking changes
- Adds `stream_output` to log printed lines as they come independent of return settings

## Importance

- The `ShellTask` is a bit neglected and was hard to inspect in failure cases.
- Setup for future `ShellCommand` but fixes pressing problems

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/shell.py
+++ b/src/prefect/tasks/shell.py
@@ -1,15 +1,23 @@
 import os
 import tempfile
 from subprocess import PIPE, STDOUT, Popen
-from typing import Any
+from typing import Any, TypedDict, List, Union, Optional
 
 import prefect
 from prefect.utilities.tasks import defaults_from_attrs
 
 
+class ShellResult(TypedDict):
+    output: List[str]
+    returncode: int
+
+
 class ShellTask(prefect.Task):
     """
     Task for running arbitrary shell commands.
+
+    NOTE: This task combines stderr and stdout because reading from both
+          streams without blocking is tricky.
 
     Args:
         - command (string, optional): shell command to be executed; can also be
@@ -25,8 +33,11 @@ class ShellTask(prefect.Task):
             should return all lines of stdout as a list, or just the last line
             as a string; defaults to `False`
         - log_stderr (bool, optional): boolean specifying whether this task
-            should log the output from stderr in the case of a non-zero exit code;
+            should log the output in the case of a non-zero exit code;
             defaults to `False`
+        - stream_output (bool, optional): specifies whether this task should log
+            the output as it occurs. If enabled, `log_stderr` will be ignored as the
+            output will have already been logged. defaults to `False`
         - **kwargs: additional keyword arguments to pass to the Task constructor
 
     Example:
@@ -52,6 +63,7 @@ class ShellTask(prefect.Task):
         shell: str = "bash",
         return_all: bool = False,
         log_stderr: bool = False,
+        stream_output: bool = False,
         **kwargs: Any
     ):
         self.command = command
@@ -59,11 +71,17 @@ class ShellTask(prefect.Task):
         self.helper_script = helper_script
         self.shell = shell
         self.return_all = return_all
+
+        # TODO: log_stderr should be deprecated and renamed for clarity this may be
+        #       worth retaining here for compat reasons and using a new task instead
         self.log_stderr = log_stderr
+        self.stream_output = stream_output
         super().__init__(**kwargs)
 
     @defaults_from_attrs("command", "env")
-    def run(self, command: str = None, env: dict = None) -> str:
+    def run(
+        self, command: str = None, env: dict = None
+    ) -> Optional[Union[str, List[str]]]:
         """
         Run the shell command.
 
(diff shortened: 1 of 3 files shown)


madkinsz:
Alas `TypedDict` is 3.8+ only :( @joshmeek I'd rather not return a dict directly and a class seems heavy. We don't have `pydantic` or `attrs` in our requirements so I'm not sure what's best. Let me know what you think.

madkinsz:
I'm interested in adding `attrs` to our requirements as it's useful for things like this and lightweight. @cicdw ?

src/prefect/tasks/shell.py
```diff
@@ -101,26 +119,34 @@ def run(self, command: str = None, env: dict = None) -> str:
...
 
-                    raise prefect.engine.signals.FAIL(msg) from None  # type: ignore
-        if self.return_all:
-            return lines
-        else:
-            return line
+                    # Return a exit code and lines so the failure can be inspected
+                    raise prefect.engine.signals.FAIL(
+                        msg,
+                        result=ShellResult(
```
joshmeek, line 143: I'm not sure I follow the need have a `ShellResult` typed dict as opposed to a standard dictionaty :thinking:

I also think this functionality of attaching the lines should be gated behind some kwarg like `return_stdout_on_error` (probably a better name than this) and keep with only attaching the error code and message. As an example on why I think it should be gated: I recall some users performing terraform commands with this task and if there's an error there would be quite a bit of lines returned in the output.

madkinsz: Well I think just returning a dictionary without information about what's in it isn't a pattern I like. I'd prefer to have the return type of the function be meaningful (hence a typed dict or a lightweight class).

Are the lines being returned in the signal result displayed or something? Wouldn't they just not access them if they don't need them?

joshmeek: I don't feel strongly about it so no preference but a dictionary with static keys

```python
{
  output: [],
  return_code: ...
}
```

that are documented in the docstring, to me, feels like is provides the same benefit.

Thinking on this a bit more I actually am more in favor of also using the `return_all` logic here (instead of the kwarg to gate it) that is present at the end of the task. i.e. if I don't opt in to returning all of my lines then I am expecting the FAIL signal's return value to only have the last line.

madkinsz: It's missing the benefit of:
- Understanding the function's return by looking at the type signature
- mypy type checking of the return value and access to its members
- Enforcing and indicating types of the members of the dict

`return_all` makes sense to me! I'll implement that.

jcrist: I think josh is right here. Since the `ShellResult` is only part of a raised `FAIL` signal, a dict here is best. We need it json-serializable, and users are unlikely to deal with it programmatically (and if they do, documenting the fields in the docstring makes the most sense to me).

joshmeek (approved):
Lgtm 🚀

PR #3651, opened 2020-11-11
Pass PREFECT_CORE_VERSION to Server

madkinsz:
## Summary

Passes `PREFECT_CORE_VERSION` during `prefect server start` for https://github.com/PrefectHQ/server/pull/126

## Changes

- Passes PREFECT_CORE_VERSION to docker-compose
- Adds PREFECT_CORE_VERSION to the server container env

## Importance

Allow core version display in the UI

## Checklist

This PR:

Manually verified that this value populates when running `prefect server start`

I do not think a changes/ entry is meaningful here.

### DIFF
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -325,6 +325,9 @@ def start(
         )
         env.update(PREFECT_SERVER_DB_CMD=cmd)
 
+    # Pass the Core version so the Server API can return it
+    env.update(PREFECT_CORE_VERSION=prefect.__version__)
+
     proc = None
     try:
         if not skip_pull:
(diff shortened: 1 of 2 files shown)


src/prefect/cli/server.py
```diff
@@ -325,6 +325,9 @@ def start(
         )
         env.update(PREFECT_SERVER_DB_CMD=cmd)
 
+    # Pass the Core version so the Server API can return it
+    env.update(PREFECT_CORE_VERSION=prefect.__version__)
```
madkinsz, line 329: This could also use `base_version[0]` from above where `base_version = prefect.__version__.split("+")` or we could use `default_tag` which is either `master` if there are extra commits or `core-VERSION` but it seems useful to know if the prefect launching the server is dirty/has extra commits. Hmm.

cicdw: Yea I agree, I like using `__version__` unprocessed 👍

cicdw (approved):

Issue #3655, opened 2020-11-12
Flow.replace sets reference tasks even if they weren't set previously in the flow


benfogelson:
## Description

`Flow.replace` manually sets a flow's reference tasks whether or not the flow actually had manual reference tasks set or not.  This means that new terminal tasks added to the flow after the call to `Flow.replace` won't become reference tasks.

I'm not actually sure this is a bug, because there is a test that seems to actually check for and enforce this behavior: https://github.com/PrefectHQ/prefect/blob/b085657965a9c483c2671e9cee6a66075f31a9fe/tests/core/test_flow.py#L1646-L1647

This behavior is so counter-intuitive that I can't help but wonder if enforcing it is an oversight.

## Expected Behavior

I would have expected that if a flow doesn't have specified reference tasks (i.e. if `Flow._reference_tasks` is empty), then it should default to terminal tasks being reference tasks, both before and after a call to `Flow.replace`.

## Reproduction

```python
from prefect import Flow, Task

with Flow(name="test") as f:
    t1 = Task(name="t1")()
    t2 = Task(name="t2")(upstream_tasks=[t1])
t3 = Task(name="t3")
f.replace(t1, t3)
t4 = Task(name="t4")
f.add_task(t4)
assert f.reference_tasks() == {t2, t4}
```

cicdw:
Closed with #3656

PR #3656, opened 2020-11-12
Stopped Flow.replace from setting unset reference tasks

benfogelson:
## Summary

Changes `Flow.replace` so that it doesn't set reference tasks on a flow that didn't already have reference tasks set.

## Changes

This PR modifies `Flow.replace` by adding an `if` statement to check whether `Flow._reference_tasks` is set before it updates reference tasks by replacing the old task with the new task.

## Importance

This is discussed in https://github.com/PrefectHQ/prefect/issues/3655

Briefly, the current behavior of `Flow.replace` does not make intuitive sense since it freezes the reference tasks of a flow that didn't have frozen reference tasks.

## Checklist

This PR:

### DIFF
--- a/changes/issue3655.yaml
+++ b/changes/issue3655.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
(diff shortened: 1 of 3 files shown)


cicdw (approved):
Great find and clean solution - thank you!

PR #3664, opened 2020-11-13
Allow clocks to pass parameterized flow run names

znicholasbrown:
## Summary

Allow clocks to dictate auto-scheduled flow runs' names based on parameters. 

## Changes

Adds a new arg to the `Clock` class, `flow_run_name_template`, which is a templated string that can reference parameter defaults defined on the `Clock` object. 

## Importance

It allows users to define their own flow run name template, rather than using the auto-generated animal names from Prefect Server/Cloud.

## Checklist

This PR:

### DIFF
--- a/src/prefect/schedules/clocks.py
+++ b/src/prefect/schedules/clocks.py
@@ -133,6 +133,9 @@ class IntervalClock(Clock):
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
+        - flow_run_name_template (str, optional): a templated string referencing any parameterized
+            values which the scheduler can use to create scheduled flow runs
+
 
     Raises:
         - TypeError: if start_date is not a datetime
@@ -246,6 +249,8 @@ class CronClock(Clock):
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
+        - flow_run_name_template (str, optional): a templated string referencing any parameterized
+            values which the scheduler can use to create scheduled flow runs
         - day_or (bool, optional): Control how croniter handles `day` and `day_of_week` entries.
             Defaults to True, matching cron which connects those values using OR.
             If the switch is set to False, the values are connected using AND. This behaves like
@@ -272,7 +277,7 @@ class CronClock(Clock):
         self.cron = cron
         self.day_or = True if day_or is None else day_or
 
-        print(f'Creating clock with flow run template {flow_run_name_template}')
+        print(f"Creating clock with flow run template {flow_run_name_template}")
         super().__init__(
             start_date=start_date,
             end_date=end_date,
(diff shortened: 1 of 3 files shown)


src/prefect/schedules/clocks.py
```diff
@@ -65,6 +67,8 @@ class Clock:
             Flow Runs which are run on this clock's events
         - labels (List[str], optional): a list of labels to apply to all flow runs generated
             from this Clock
+        - flow_run_name_template (str, optional): a templated string referencing any parameterized
+            values which the scheduler can use to create scheduled flow runs
```
madkinsz, line 71: Is it worth discussing the available values a bit more and what happens when this is `None`?

src/prefect/schedules/clocks.py
```diff
@@ -254,17 +269,21 @@ def __init__(
...
         day_or: bool = None,
+        flow_run_name_template: str = None,
     ):
         # build cron object to check the cron string - will raise an error if it's invalid
         if not croniter.is_valid(cron):
             raise ValueError("Invalid cron string: {}".format(cron))
         self.cron = cron
         self.day_or = True if day_or is None else day_or
+
+        print(f"Creating clock with flow run template {flow_run_name_template}")
```
cicdw, line 280: debug print?

src/prefect/schedules/clocks.py
cicdw: If it's `None`, the current logic will kick in which sets a random adj-animal run name.  

In terms of available values, it's a good question - I don't think I see a use case for templating the name based on properties that are also known at Clock creation time.  It seems hard-coded names here would allow for the same amount of functionality?

cicdw (changes requested):
Stray print + question about whether templating is actually useful here?

pedromachados:
Templating would be useful to name flows based on the scheduled run time.

src/prefect/schedules/clocks.py
madkinsz: > If it's `None`, the current logic will kick in which sets a random adj-animal run name.

I mean is that worth explaining in the docstring :)

src/prefect/schedules/clocks.py
znicholasbrown: Oops yeah, removed 😬

src/prefect/schedules/clocks.py
znicholasbrown: @cicdw your proposal is that we remove templating in favor of a static string? I think we should offer some dynamism to this even if it's not parameters (though I think there's a value in only writing your parameter value once instead of multiple times when it changes).

cicdw: Yea, and in fact I actually think we should resolve https://github.com/PrefectHQ/server/issues/125 first; that way we can template via scheduled start time, parameters for the run (which might include flow group defaults once that issue is closed), and also labels for the run (which could come from flow group defaults as well)

znicholasbrown:
Since this has been sitting for awhile and I don't have the bandwidth atm, anyone can feel free to take this over 👍🏻

PR #3666, opened 2020-11-14
Write agent command line warnings to stderr

bnaul:
Currently if you run `prefect agent install kubernetes ... | kubectl apply -f -` you get a fairly confusing error:
```
error: error parsing STDIN: error converting YAML to JSON: yaml: found character that cannot start any token
```

The reason is that `Warning: `prefect agent install kubernetes` is deprecated, use `prefect agent kubernetes install` instead`  shows up at the beginning of the stdout block and gets piped through to `kubectl apply`. Writing to stderr instead seems more appropriate here.

### DIFF
--- a/src/prefect/cli/agent.py
+++ b/src/prefect/cli/agent.py
@@ -616,13 +616,14 @@ def start(
         retrieved_agent = _agents.get(agent_option, None)
 
         if not retrieved_agent:
-            click.secho("{} is not a valid agent".format(agent_option), fg="red")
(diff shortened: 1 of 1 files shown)


cicdw (changes requested):
@bnaul could you run `black .` on the file and add an `enhancement` (or `fix`) `changes/` entry?

Otherwise LGTM!

cicdw (approved):

bnaul:
oops sorry forgot about the `changes` entry...

cicdw:
@bnaul no worries, we can add one

PR #3679, opened 2020-11-18
Use exist_ok flag in os.makedirs to avoid race condition

rmax:
## Summary

Related to https://github.com/PrefectHQ/prefect/discussions/3678

There is a potential race condition between `os.path.exists` and `os.makedirs`.
`os.makedirs` provides the `exist_ok` flag and already handles race condition case when other thread may
have created the directory between the exists check and the makedirs
call. See
https://github.com/python/cpython/blob/48b069a003ba6c684a9ba78493fbbec5e89f10b8/Lib/os.py#L214-L218

## Changes

Replaces `os.path.exists` by `exist_ok` flag.

## Importance

I faced this issue when writing tests for a custom agent that used local storage. More details in https://github.com/PrefectHQ/prefect/discussions/3678

Workaround is to use `validate=False` and create the directory before instantiating the local storage class.

## Checklist

This PR:

### DIFF
--- a/changes/pr3679.yml
+++ b/changes/pr3679.yml
@@ -1,0 +1,5 @@
+fix:
+  - "Use `exist_ok` flag in `os.makedirs` to avoid race condition in local storage class - [#3679](https://github.com/PrefectHQ/prefect/pull/3679)"
+
+contributor:
+  - "[R Max Espinoza](https://github.com/rmax)"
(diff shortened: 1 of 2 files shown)


rmax:
Moved back to draft as it's missing changelog updates.

joshmeek:
Thanks for making the PR @rmax! Lgtm once you add the changes entry 🙂

rmax:
@joshmeek done!

joshmeek (approved):

PR #3682, opened 2020-11-18
Serialize task, edges, etc. in deterministic order

madkinsz:
## Summary

Much to my shame, `serialized_hash` was still not creating a consistent output. I've now expanded the test cases further and fixed the ordering issues that I thought would be fully resolved in #3654 

Alternate implemntation in #3683 abandoned because of typo anagrams.

## Changes

- Adds helpers for `FlowSchema` that sorts tasks and parameters
- `FlowSchema` helpers for reference tasks and parameters return sorted lists instead of sets
- Adds note about serialized_hash not necessarily changing when the code does
- Adds `SortedList` derivative of `marshmallow.fields.List`
- Sorts lists for labels in `run_configs` and `environments`

## Importance

Fixes additional bugs with `serialized_hash`

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1472,6 +1472,11 @@ class Flow:
         determining if the flow has changed. If this hash is equal to a previous hash,
         no new information would be passed to the server on a call to `flow.register()`
 
+        Note that this will not always detect code changes since the task code is not
+        included in the serialized flow sent to the server. That said, as long as the
+        flow is "built" during registration, the code changes will be in effect even
+        if a new version is not registered with the server.
+
         Args:
             - build (bool, optional):  if `True`, the flow's environment is built
                 prior to serialization. Passed through to `Flow.serialize()`.
@@ -1479,8 +1484,15 @@ class Flow:
         Returns:
             - str: the hash of the serialized flow
         """
+        serialized_flow = self.serialize(build)
+
+        # Sort the tasks by slug name for determinism
+        serialized_flow["tasks"] = sorted(
+            serialized_flow["tasks"], key=lambda task: task["slug"]
+        )
+
         return hashlib.sha256(
-            json.dumps(self.serialize(build), sort_keys=True).encode()
+            json.dumps(serialized_flow, sort_keys=True).encode()
         ).hexdigest()
 
     # Diagnostics  ----------------------------------------------------------------
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1445,18 +1445,6 @@ class Flow:
         for task, slug in flow_copy.slugs.items():
             task.slug = slug
 
-        # Convert tasks and edges from sets to sorted lists for determinism
-        # mypy will complain that t.slug is `Optional[str]` but we have enforced this
-        # above so it should be safe. We also need to ignore types here because we are
-        # pushing a list into a `Set` typed attribute
-        flow_copy.tasks = sorted(
-            list(flow_copy.tasks), key=lambda t: t.slug  # type: ignore
-        )  # type: ignore
-        flow_copy.edges = sorted(
-            list(flow_copy.edges),
-            key=lambda e: f"{e.upstream_task.slug}-{e.downstream_task.slug}",
-        )  # type: ignore
-
         serialized = schema(exclude=["storage"]).dump(flow_copy)
 
         if build:
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1446,11 +1446,16 @@ class Flow:
             task.slug = slug
 
         # Convert tasks and edges from sets to sorted lists for determinism
-        flow_copy.tasks = sorted(list(flow_copy.tasks), key=lambda t: t.slug)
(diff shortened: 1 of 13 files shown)


src/prefect/core/flow.py
```diff
@@ -1472,15 +1480,22 @@ def serialized_hash(self, build: bool = False) -> str:
...
                 prior to serialization. Passed through to `Flow.serialize()`.
 
         Returns:
             - str: the hash of the serialized flow
         """
+        serialized_flow = self.serialize(build)
+
         return hashlib.sha256(
-            json.dumps(self.serialize(build), sort_keys=True).encode()
+            json.dumps(serialized_flow, sort_keys=True).encode()
```
madkinsz, line 1498: This pulled out because I was transforming the serialized_flow but switched to transforming the flow copy before serialization. Still useful pulled out for debugging.

tests/core/test_flow.py
```diff
@@ -1900,6 +1901,42 @@ def dummy_task():
         assert hashes[0]  # Ensure we don't have an empty string or None
         assert len(set(hashes)) == 1
 
+    def test_task_order_is_deterministic(self):
```
madkinsz, line 1904: These tests could be duplicated (or moved) into the flow serialization tests

src/prefect/core/flow.py
cicdw: Question - instead of messing around with the serialization logic, couldn't you do:
```python
"".join(sorted(json.dumps(serialized_flow)).encode()
```
which would detect any changes?

cicdw: This PR would work it seems, but will be really hard to maintain as the changes are so diverse

madkinsz: I did try sorting after serializing but it's a bit of a pain. The issue here are there are lists of dictionaries and the dictionaries cannot be sorted in the list without identifying which key you'd like to sort by. I think it's more clear to sort at the schema level and actually I think if I move the sorting next to the `FlowSchema` definition and add a note that nested schemas with `many=True` will need to be sorted for determinism it'll be pretty easy to maintain.

madkinsz: Chris just schooooooled me. I'll be updating this with that recommendation and a comment.

madkinsz:
Closing in favor of https://github.com/PrefectHQ/prefect/pull/3683

madkinsz:
I've noticed that we have our own implementation of `Nested` and could technically automatically sort all collections that use `many` within there but I think that might be a bit opaque.

src/prefect/serialization/flow.py
```diff
@@ -16,21 +16,43 @@
...
+
+def get_tasks(obj: prefect.Flow, context: dict) -> List:
+    return list(sorted(obj.tasks, key=lambda t: t.slug))
+
+
+def get_edges(obj: prefect.Flow, context: dict) -> List:
+    return list(
+        sorted(
+            obj.edges,
+            key=lambda e: f"{e.upstream_task.slug}-{e.downstream_task.slug}",
```
jcrist, line 47: No need to format this, tuples are sortable just fine.

```suggestion
            key=lambda e: (e.upstream_task.slug, e.downstream_task.slug),
```

jcrist (commented):
Both `flow.environment.labels` and `flow.run_config.labels` (when present) are sets that should also be sorted when serialized.  There may be others as well.

madkinsz:
I clearly don't have enough flows in my repetoire

> Both `flow.environment.labels` and `flow.run_config.labels` (when present) are sets that should also be sorted when serialized. There may be others as well.

This should be fixed now. I'll do another check for other sets

madkinsz:
@jcrist I believe I got them and we can rely on user reports for any edge cases here? Maintaining test coverage for this in `TestSerializedHash` is feeling a bit unreasonable. Perhaps we can write a test that captures all the sets we know about?

src/prefect/utilities/serialization.py
```diff
@@ -231,6 +231,23 @@ def _serialize(self, value, attr, obj, **kwargs):  # type: ignore
...
+    An extension of the Marshmallow List field that sorts the serialized object for
+    determinism
+
+    Args:
+        - cls_or_instance (type): the inner field class
+        - **kwargs (Any): the keyword arguments accepted by `marshmallow.Field`
+    """
+
+    def __init__(self, cls_or_instance: Union[fields.Field, type], **kwargs: Any):
+        super().__init__(cls_or_instance, **kwargs)
```
jcrist, line 245: If a class doesn't override the `__init__` at all, there's no need to define one.

```suggestion
```

jcrist:
The test failures are due to changes in `dask-cloudprovider` that were already fixed in master - if you merge in master the tests should be fixed.

jcrist:
> Perhaps we can write a test that captures all the sets we know about?

What I would do would be add a test in `tests/serialization/*.py` for the appropriate classes that checks that serializing them results in sorted lists. For example, add a few tags to an `Environment`, then check that the serialized environment tags are sorted. Not a perfect test for the `serialized_hash`, but good enough IMO.

jcrist (approved):

PR #3683, opened 2020-11-18
Ensure `serialized_hash` is resilient to ordering differences

madkinsz:
## Summary

`Flow.serialized_hash()` would change depending on the order of tasks (and edges, parameters, reference tasks) which was dynamic because they are stored in a `Set`

See initial implementation #3682 

## Changes

- Adds disclaimer about code changes to `serialized_hash`
- Fixes `serialized_hash` handling of unordered sets of objects within the flow

## Importance

Makes `serialized_hash` meaningfully finally.

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1472,6 +1472,11 @@ class Flow:
         determining if the flow has changed. If this hash is equal to a previous hash,
         no new information would be passed to the server on a call to `flow.register()`
 
+        Note that this will not always detect code changes since the task code is not
+        included in the serialized flow sent to the server. That said, as long as the
+        flow is "built" during registration, the code changes will be in effect when you
+        run your flow even if a new version is not registered with the server.
+
         Args:
             - build (bool, optional):  if `True`, the flow's environment is built
                 prior to serialization. Passed through to `Flow.serialize()`.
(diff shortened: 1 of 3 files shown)


src/prefect/core/flow.py
```diff
@@ -1472,16 +1472,24 @@ def serialized_hash(self, build: bool = False) -> str:
...
             - str: the hash of the serialized flow
         """
-        return hashlib.sha256(
-            json.dumps(self.serialize(build), sort_keys=True).encode()
-        ).hexdigest()
+        serialized_flow = self.serialize(build)
+        # Sort all the characters from the serialized flow json to get a unique key for
+        # this flow. This eliminates all concerns about the order of items in sets
+        # within the serialized flow but retains a unique key.
+        key = "".join(sorted(json.dumps(serialized_flow)))
```
jcrist, line 1491: This is prone to a non-negligible set of hash collisions, and some that might be common-enough for them to hit users.

I'm specifically thinking of anagram typos, which aren't infrequent:

```python
@task
def some_tsak(a, b):
    pass

# edit to fix the typo would still hash the same
@task
def some_task(a, b):
    pass
```

or updates that swap fields like

```python
@task(checkpoint=False)
def task_a():
    pass

@task(checkpoint=True)
def task_b():
    pass

# edit to switch the checkpoints to True and False respectively
# in one go would still hash the same
@task(checkpoint=True)
def task_a():
    pass

@task(checkpoint=False)
def task_b():
    pass
```

If possible, I think it'd be better to ensure that all list-like-things in the serialized representation (output of `flow.serialize()` have a deterministic order. I don't *think* this should be too hard, but I might be mistaken. Then `json.dumps(sort_keys=True)` would be all that was needed for a deterministic json output.

madkinsz: 😢 You're right, those cases do seem relevant. @cicdw do you have a rebuttal before I return to a cleaned up version of #3682 ?

cicdw: I'm a pessimist with unwritten rules that affect downstream features, so I expect that if this key relies on deterministic key order then we will one day "break" it without knowing, but ultimately this feature is a convenience more than anything, so a changing key doesn't have any bad impact other than mild annoyance.  So I'm OK taking the other approach 👍 

Also, super clever observation @jcrist !

madkinsz:
Closing in favor of #3682 🤣

PR #3690, opened 2020-11-19
Fix dask-cloudprovider tests

jcrist:
The import path changed to local-only (instead of top-level) in the
latest release. Local imports are backwards compatible, so this should
work for all versions.


joshmeek (approved):

jcrist:
Yeah, looks like a bad changelog entry was merged in a prior PR (also needed to update the requirements to include AWS extra).

joshmeek (approved):

PR #3691, opened 2020-11-20
Nice warning if passing task to Task constructor

jcrist:
A common user mistake when usint class-based tasks is to pass dependent
tasks as arguments to the task constructor, rather than when calling the
task (or using methods like `.map`/`.bind`/`.set_dependencies`). We now
raise a nice warning in this case. We don't error outright in case the
user actually intended this (although my guess is this case is rare).

Also cleans up a few linting errors in the `test_task.py` file.

This PR:

### DIFF
--- a/changes/pr3691.yaml
+++ b/changes/pr3691.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Raise nice warning if user passes `Task` instance to `Task` constructor, rather than when calling the `Task` (or using `Task.map`/`Task.set_dependencies`) - [#3691](https://github.com/PrefectHQ/prefect/pull/3691)"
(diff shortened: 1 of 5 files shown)


jcrist:
Example:

```python
In [2]: from prefect import Flow, Task

In [3]: class MyInc(Task):
   ...:     def __init__(self, a=None, **kwargs):
   ...:         self.a = a
   ...:         super().__init__(**kwargs)
   ...:     def run(self, a=None):
   ...:         return (a if a is not None else self.a) + 1
   ...:

In [4]: with Flow("test") as flow:
   ...:     a = MyInc(1)
   ...:     b = MyInc(a)  # user error
   ...:     flow.add_task(a)
   ...:     flow.add_task(b)
   ...:
<ipython-input-4-6f8d6507c31f>:3: UserWarning: A Task was passed as an argument to MyInc, you likely want to first initialize MyInc with any static (non-Task) arguements, then call the initialized task with any dynamic (Task) arguements instead. For example:

  my_task = MyInc(...)  # static (non-Task) args go here
  res = my_task(...)  # dynamic (Task) args go here

see https://docs.prefect.io/core/concepts/flows.html#apis for more info.
```

joshmeek (approved):

PR #3692, opened 2020-11-20
Add kwarg for creating link artifact in StartFlowRun task

joshmeek:
## Summary
Now that there is an artifact API for creating things like links we can take advantage of it in some places like the task library. This serves as an initial addition where we can create a link artifact for the `StartFlowRun` task that links to the created flow run in the UI.

## Changes
Adds a new kwarg `create_link_artifact` to the `StartFlowRun` task. When set to `True` this will post a link artifact that looks something like `https://cloud.prefect.io/my-tenant-slug/flow-run/424242-ca-94611-111-55` (or localhost for server users).

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -846,9 +846,7 @@ class Client:
 
         return flow_id
 
-    def get_cloud_url(
-        self, subdirectory: str, id: str, as_user: bool = True, hostname: bool = True
-    ) -> str:
+    def get_cloud_url(self, subdirectory: str, id: str, as_user: bool = True) -> str:
         """
         Convenience method for creating Prefect Cloud URLs for a given subdirectory.
 
@@ -857,8 +855,6 @@ class Client:
             - id (str): the ID of the page
             - as_user (bool, optional): whether this query is being made from a USER scoped token;
                 defaults to `True`. Only used internally for queries made from RUNNERs
-            - hostname (bool, optional): whether or not to include the hostname in the URL. e.g.
-                `cloud.prefect.io` or `localhost:8080`. Defaults to `True`
 
         Returns:
             - str: the URL corresponding to the appropriate base URL, tenant slug, subdirectory
@@ -880,17 +876,15 @@ class Client:
         tenant_slug = self.get_default_tenant_slug(as_user=as_user and using_cloud_api)
 
         # For various API versions parse out `api-` for direct UI link
-        base_url = ""
-        if hostname:
-            base_url = (
-                (
-                    re.sub("api-", "", prefect.config.cloud.api)
-                    if re.search("api-", prefect.config.cloud.api)
-                    else re.sub("api", "cloud", prefect.config.cloud.api)
-                )
-                if using_cloud_api
-                else prefect.config.server.ui.endpoint
+        base_url = (
+            (
+                re.sub("api-", "", prefect.config.cloud.api)
+                if re.search("api-", prefect.config.cloud.api)
+                else re.sub("api", "cloud", prefect.config.cloud.api)
             )
+            if using_cloud_api
+            else prefect.config.server.ui.endpoint
+        )
 
         return "/".join([base_url.rstrip("/"), tenant_slug, subdirectory, id])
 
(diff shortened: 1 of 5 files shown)


jcrist:
> Making this behavior opt-in because we don't want tasks to post artifacts by default without user knowledge.

IMO a link to another flow run is innocuous and makes the UI experience when using `StartFlowRun` much better - I'm not sure we need this to be opt-in.

> When set to True this will post a link artifact that looks something like https://cloud.prefect.io/my-tenant-slug/flow-run/424242-ca-94611-111-55 (or localhost for server users). 

Hmmm, I'm not sure about this. We don't know what host the user is viewing the UI at. This could be `cloud.prefect.io`, could be `localhost`, could be a self-hosted server at some other address, could even be a self-hosted UI still proxying through to prefect-cloud's backend (since we allow that). Perhaps we should use a relative link here instead of an absolute one? This way the link should work regardless of how the user's browser is getting to the UI.

jcrist:
> Perhaps we should use a relative link here instead of an absolute one? 

Looking closer, I see you're using `Client.get_cloud_url` to do this, which is already used elsewhere. I guess if there's precedence then maybe we don't care about this for now, not sure. I do think a relative link is less likely to be an issue, but it also would limit the link to being used from the UI (which IMO is fine, since I think that's how we see these artifacts being used). Up to you.

src/prefect/client/client.py
```diff
@@ -870,25 +874,25 @@ def get_cloud_url(self, subdirectory: str, id: str, as_user: bool = True) -> str
...
-        else:
-            tenant_slug = ""
 
-        base_url = (
-            re.sub("api-", "", prefect.config.cloud.api)
-            if re.search("api-", prefect.config.cloud.api)
-            else re.sub("api", "cloud", prefect.config.cloud.api)
-        )
+        # Search for matching cloud API because we can't guarantee that the backend config is set
+        using_cloud_api = bool(re.search(".prefect.io", prefect.config.cloud.api))
```
jcrist, line 879:
```suggestion
        using_cloud_api = ".prefect.io" in prefect.config.cloud.api
```

src/prefect/tasks/prefect/flow_run.py
```diff
@@ -29,7 +30,9 @@ class StartFlowRun(Task):
             task will wait until the flow run is complete, and then reflect the corresponding
             state as the state of this task.  Defaults to `False`.
         - scheduled_start_time (datetime, optional): the time to schedule the execution
-                for; if not provided, defaults to now
+            for; if not provided, defaults to now
+        - create_link_artifact (bool, optional): create a link artifact that links to the
+            created flow run page in the UI. Defaults to `True`.
```
jcrist, line 35: Is there a reason a user would ever set this to False? If not, we might skip the option entirely until its requested.

src/prefect/client/client.py
```diff
@@ -855,6 +857,8 @@ def get_cloud_url(self, subdirectory: str, id: str, as_user: bool = True) -> str
             - id (str): the ID of the page
             - as_user (bool, optional): whether this query is being made from a USER scoped token;
                 defaults to `True`. Only used internally for queries made from RUNNERs
+            - hostname (bool, optional): whether or not to include the hostname in the URL. e.g.
```
jcrist, line 860: Instead of adding the option here, you might always return with the hostname, and strip out just the path using `urllib.parse.urlparse`:

```python
full_url = Client().get_cloud_url(...)
just_path = urllib.parse.urlparse(full_url).path
```

keeps the `Client` method simpler, and doesn't leave the user wondering if the result has a leading slash or not. Since this is currently only needed in one place this might be simpler. No strong thoughts either way though.

I also find the kwarg `hostname` confusing, since it reads like I can pass the hostname in (as a string), rather than a bool. If you do keep the option, might be clearer as `relative_url=False` instead.

jcrist (commented):
A few small nits/questions, otherwise this looks good to me.

src/prefect/tasks/prefect/flow_run.py
joshmeek: Yeah that's fair, no real need for it until requested

jcrist (approved):
LGTM (assuming tests pass).

PR #3694, opened 2020-11-20
Cleanup docker storage dockerfile formatting

madkinsz:
## Summary

Automatically quote passed dockerfile environment variables so that something like

```python
{"MY_ENV_VAR": "Hello world!"}
```

does not result in a syntax error when written to the dockerfile

## Changes

- Adds quotes around relevant environment variable declarations using their repr format
- Fixes indentation for dockerfile environment variables
- Cleanup of dockerfile building, especially formatting

## Importance

- Less likely to encounter an unexpected error when using env variables
- A bit clearer file formatting if you need to inspect the dockerfile

## Checklist

This PR:

### DIFF
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -464,7 +464,7 @@ class Docker(Storage):
         if self.env_vars:
             white_space = " " * 4
             env_vars = "ENV " + " \\ \n{}".format(white_space).join(
-                '{k}="{v}"'.format(k=k, v=v) for k, v in self.env_vars.items()
+                f"{k}={v!r}" for k, v in self.env_vars.items()
             )
 
         # Copy user specified files into the image
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -391,7 +391,8 @@ class Docker(Storage):
             dockerfile_path = self.create_dockerfile_object(directory=tempdir)
             client = self._get_client()
 
-            # Verify that a registry url has been provided for images that should be pushed
+            # Verify that a registry url has been provided for images that should be
+            # pushed
             if self.registry_url:
                 full_name = str(PurePosixPath(self.registry_url, self.image_name))
             elif push is True:
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -464,7 +464,7 @@ class Docker(Storage):
         if self.env_vars:
             white_space = " " * 20
             env_vars = "ENV " + " \\ \n{}".format(white_space).join(
-                "{k}={v}".format(k=k, v=v) for k, v in self.env_vars.items()
+                '{k}="{v}"'.format(k=k, v=v) for k, v in self.env_vars.items()
             )
 
         # Copy user specified files into the image
--- a/src/prefect/environments/storage/docker.py
+++ b/src/prefect/environments/storage/docker.py
@@ -23,6 +23,14 @@ if TYPE_CHECKING:
     import docker
 
 
+def multiline_indent(string: str, spaces: int) -> str:
+    """
+    Utility to indent all but the first line in a string to the specified level,
+    especially useful in textwrap.dedent calls where multiline strings are formatted in
+    """
+    return string.replace("\n", "\n" + " " * spaces)
+
+
 class Docker(Storage):
     """
     Docker storage provides a mechanism for storing Prefect flows in Docker
@@ -462,10 +470,9 @@ class Docker(Storage):
         # Generate ENV variables to load into the image
         env_vars = ""
         if self.env_vars:
-            white_space = " " * 4
-            env_vars = "ENV " + " \\ \n{}".format(white_space).join(
-                f"{k}={v!r}" for k, v in self.env_vars.items()
-            )
+            # Format with repr to get proper quoting
+            formatted_vars = [f"{k}={v!r}" for k, v in self.env_vars.items()]
+            env_vars = "ENV " + " \\\n    ".join(formatted_vars)
 
         # Copy user specified files into the image
         copy_files = ""
@@ -520,7 +527,7 @@ class Docker(Storage):
         final_commands = (
             ""
             if self.extra_dockerfile_commands is None
-            else str.join("\n", self.extra_dockerfile_commands)
+            else "\n".join(self.extra_dockerfile_commands)
         )
 
         # Write a healthcheck script into the image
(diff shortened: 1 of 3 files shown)


src/prefect/environments/storage/docker.py
```diff
@@ -464,7 +464,7 @@ def create_dockerfile_object(self, directory: str) -> str:
         if self.env_vars:
             white_space = " " * 20
```
jcrist, line 465: This huge whitespace is odd - if the point is to align the set environment variables, 

```suggestion
            white_space = " " * 4
```

would make more sense

src/prefect/environments/storage/docker.py
```diff
@@ -464,7 +464,7 @@ def create_dockerfile_object(self, directory: str) -> str:
...
             env_vars = "ENV " + " \\ \n{}".format(white_space).join(
-                "{k}={v}".format(k=k, v=v) for k, v in self.env_vars.items()
+                '{k}="{v}"'.format(k=k, v=v) for k, v in self.env_vars.items()
```
jcrist, line 467: This won't work if the environment variable has double quotes in it already. Since docker accepts both single or double quotes, we should be able to rely on repr here instead:

```suggestion
                f"{k}={v!r}" for k, v in self.env_vars.items()
```

src/prefect/environments/storage/docker.py
madkinsz: I think they actually weren't aligned at all -- I'll commit this then take a look locally

src/prefect/environments/storage/docker.py
madkinsz: Ooh that's fun, I'll add coverage for this case as well.

madkinsz: Hm that gives `STR_WITH_QUOTES="Hello "friend""`

madkinsz: We may be best off just dropping this and relying on the user to quote their variables as needed

jcrist: ? I'm unable to reproduce:

```python
In [39]: v = '"has quotes"'

In [40]: k = 'FOO'

In [41]: print(f"{k}={v!r}")
FOO='"has quotes"'
```

madkinsz: You're right that is working as expected, there are some other wonky things going on with the dockerfile formatting that I'm digging into now

src/prefect/environments/storage/docker.py
```diff
@@ -23,6 +23,14 @@
     import docker
 
 
+def multiline_indent(string: str, spaces: int) -> str:
```
madkinsz, line 26: This is reproduced from https://github.com/PrefectHQ/prefect/pull/3632/files#diff-4c6587fe7d1728fde257726e6d238f5f098c62f56cf2226c013642ae2063a5f7R22 and would be useful as a general util for formatting

jcrist: Until it's needed elsewhere, I'd prefer to keep this here. There's a mental indirection cost to having to look up a utility in a different file, and this one's simple enough that pulling it out doesn't feel worth it.

jcrist (commented):
This now touches a bunch of other stuff, mind giving a rundown of what the other changes were for?

src/prefect/environments/storage/docker.py
```diff
@@ -453,19 +462,30 @@ def create_dockerfile_object(self, directory: str) -> str:
         Returns:
             - str: the absolute file path to the Dockerfile
         """
+        # Either load the base commands from the specified dockerfile or define a FROM
+        if self.dockerfile:
+            with open(self.dockerfile, "r") as contents:
+                base_commands = contents.read()
+        else:
+            base_commands = "FROM {base_image}".format(base_image=self.base_image)
+
```
madkinsz, line 471: Moved from the bottom of the variable setup here because it's the first templated item

`contents.read()` is no longer indented using `textwrap.indent` here because the number of spaces to indent by is dependent on the indentation of the code further down.

src/prefect/environments/storage/docker.py
```diff
@@ -511,16 +533,11 @@ def create_dockerfile_object(self, directory: str) -> str:
                     "A `path` must be provided to show where flow `.py` file is stored in the image."
                 )
 
-        # Write all extra commands that should be run in the image
-        installation_commands = ""
-        for cmd in self.installation_commands:
-            installation_commands += "RUN {}\n".format(cmd)
```
madkinsz, line 517: Moved up without change

src/prefect/environments/storage/docker.py
```diff
@@ -453,19 +462,30 @@ def create_dockerfile_object(self, directory: str) -> str:
...
         # Generate single pip install command for python dependencies
         pip_installs = "RUN pip install "
         if self.python_dependencies:
             for dependency in self.python_dependencies:
                 pip_installs += "{} ".format(dependency)
 
+        # Write all extra commands that should be run in the image
+        installation_commands = ""
+        for cmd in self.installation_commands:
+            installation_commands += "RUN {}\n".format(cmd)
```
madkinsz, line 481: Moved up without change to match order of template

src/prefect/environments/storage/docker.py
```diff
@@ -453,19 +462,30 @@ def create_dockerfile_object(self, directory: str) -> str:
...
         # Generate ENV variables to load into the image
         env_vars = ""
         if self.env_vars:
-            white_space = " " * 20
-            env_vars = "ENV " + " \\ \n{}".format(white_space).join(
-                "{k}={v}".format(k=k, v=v) for k, v in self.env_vars.items()
-            )
+            # Format with repr to get proper quoting
+            formatted_vars = [f"{k}={v!r}" for k, v in self.env_vars.items()]
+            env_vars = "ENV " + " \\\n    ".join(formatted_vars)
```
madkinsz, line 488: Functional change for environment variables, includes relative indent but not indent dependent on template indentation further down

src/prefect/environments/storage/docker.py
```diff
@@ -500,9 +520,11 @@ def create_dockerfile_object(self, directory: str) -> str:
...
                     cloudpickle.dump(self._flows[flow_name], f)
                 copy_flows += "COPY {source} {dest}\n".format(
-                    source=flow_path.replace("\\", "/")
-                    if self.dockerfile
-                    else "{}.flow".format(clean_name),
+                    source=(
+                        flow_path.replace("\\", "/")
+                        if self.dockerfile
+                        else "{}.flow".format(clean_name)
+                    ),
```
madkinsz, line 527: Readability change, just wrapped in parens so it's clear it's all for the `source` arg

src/prefect/environments/storage/docker.py
```diff
@@ -511,16 +533,11 @@ def create_dockerfile_object(self, directory: str) -> str:
...
-        for cmd in self.installation_commands:
-            installation_commands += "RUN {}\n".format(cmd)
-
-        # Write final user commands that should be run in the image
+        # Write final extra user commands that should be run in the image
         final_commands = (
             ""
             if self.extra_dockerfile_commands is None
-            else str.join("\n", self.extra_dockerfile_commands)
+            else "\n".join(self.extra_dockerfile_commands)
```
madkinsz, line 540: Be consistent with the rest of the file for `join` calls

src/prefect/environments/storage/docker.py
```diff
@@ -533,59 +550,40 @@ def create_dockerfile_object(self, directory: str) -> str:
...
+
+        # Generate the command to run the healthcheck
+        healthcheck_run = ""
+        if not self.ignore_healthchecks:
+            flow_file_paths = ", ".join(['"{}"'.format(k) for k in self.flows.values()])
+            python_version = (sys.version_info.major, sys.version_info.minor)
+            healthcheck_run = (
+                f"RUN python {self.prefect_directory}/healthcheck.py "
+                f"'[{flow_file_paths}]' '{python_version}'"
+            )
```
madkinsz, line 566: Consolidate healthcheck code, reduce nesting, use f-strings. Functionality unchanged.

src/prefect/environments/storage/docker.py
```diff
@@ -533,59 +550,40 @@ def create_dockerfile_object(self, directory: str) -> str:
...
 
         file_contents = textwrap.dedent(
-            """\
-            {base_commands}
+            f"""
+            {multiline_indent(base_commands, 12)}
```
madkinsz, line 570: Here and below, use a f-string instead of a format call with a bunch of repeated variables and use `multiline_indent` to indent multiline strings rather than indenting with confusing numbers above

src/prefect/environments/storage/docker.py
```diff
@@ -533,59 +550,40 @@ def create_dockerfile_object(self, directory: str) -> str:
...
-                    prefect_dir=self.prefect_directory,
-                )
-            )
+            {multiline_indent(env_vars, 12)}
+            {multiline_indent(final_commands, 12)}
+            {healthcheck_run}
+            """
+        )
 
-        file_contents = "\n".join(line.lstrip() for line in file_contents.split("\n"))
```
madkinsz, line 588: No longer strip left spaces because we've fixed the templating

src/prefect/environments/storage/docker.py
```diff
@@ -533,59 +550,40 @@ def create_dockerfile_object(self, directory: str) -> str:
...
-
-                RUN python {prefect_dir}/healthcheck.py '[{flow_file_paths}]' '{python_version}'
-                """.format(
-                    flow_file_paths=", ".join(
-                        ['"{}"'.format(k) for k in self.flows.values()]
-                    ),
-                    python_version=(sys.version_info.major, sys.version_info.minor),
-                    prefect_dir=self.prefect_directory,
-                )
-            )
```
madkinsz, line 586: This string is now always templated but is empty if we're ignoring healthchecks which reduces the nesting and makes it more readable

madkinsz (commented):
Here are some clarifying comments, we can hop on a chat too

jcrist (approved):
LGTM, thanks for the clarifying comments.

PR #3696, opened 2020-11-20
Allow chaining of imperative task dependency calls

### LINKED ISSUE #3695: Chain `set_upstream` with Task Results
## Current behavior

The `set_upstream` method is available on the functional API for tasks to explicitly set upstream dependencies like so:
```


@task
def foo():
    return "foo"


@task
def bar():
    return "bar"


@task
def oops(x):
    return "oops"


with Flow("test") as flow:
    bar_result = bar()
    foo_result = foo()
    foo_result.set_upstream(bar_result)
    oops_result = oops(foo_result)
```

## Proposed behavior

It would be intuitive to use the API with chaining, like so:

```


@task
def foo():
    return "foo"


@task
def bar():
    return "bar"


@task
def oops(x):
    return "oops"


with Flow("test") as flow:
    bar_result = bar()
    foo_result = foo().set_upstream(bar_result)
    oops_results = oops(foo_result)
```

madkinsz:
## Summary

Returns `self` from imperative task dependency calls so they can be changed

Closes #3695 

## Changes

- Changes return types of `set_upstream`, `set_downstream`, and `set_dependencies` from `None` to `Task`
- Returns `self` from all the above

## Checklist

This PR:

### DIFF
--- a/changes/issue3695.yaml
+++ b/changes/issue3695.yaml
@@ -1,0 +1,3 @@
+enhancement:
+  - "Allow chaining of `Task` imperative dependency calls - [#3696](https://github.com/PrefectHQ/prefect/pull/3696)"
+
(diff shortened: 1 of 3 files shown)


madkinsz:
This feels a little clunky but I haven't been a big user of the imperative API.

jcrist (approved):

PR #3697, opened 2020-11-20
Nicer support for tasks return multiple results

jcrist:
Adds support for annotating tasks that return multiple results, allowing
for result-unpacking. This is a convenience to avoid explicitly
indexing out tasks.

If `nout` isn't provided, it's inferred from the task return type (if
possible), falling back to `None`.

Tasks that have a non-None `nout` are iterable, so they can be used in
result unpacking or iteration, since Prefect knows how long they are. By
default (`nout=None`) tasks are still non-iterable, keeping the expected
behavior.

This makes the following work:

```python

@task(nout=2)
def inc_and_dec(x):
    return x + 1, x - 1

@task
def double_and_triple(x : int) -> tuple[int, int]:
    return x * 2, x * 3

with Flow("example"):
    inc, dec = inc_and_dec(1)
    double, triple = double_and_triple(inc)
```

This PR:

### DIFF
--- a/changes/pr3697.yaml
+++ b/changes/pr3697.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Improved support for Tasks returning multiple results - [#3697](https://github.com/PrefectHQ/prefect/pull/3697)"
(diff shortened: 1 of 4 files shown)


cicdw (dismissed):
This is an _excellent_ idea @jcrist - nicely done

madkinsz (dismissed):

jlowin:
Wow. 🤩

joshmeek (dismissed):

joshmeek (approved):

PR #3703, opened 2020-11-22
Add postgres execute many

FarisALSaleem:
## Summary

Add a new postgres task PostgresExecuteMany.

## Changes

Added a new Task subclass called PostgresExecuteMany which makes multiple queries against a Postgres database given a list of tuples.

## Importance

Makes writing Tasks that push lists of rows into a db much cleaner and faster.

## Checklist

This PR:

### DIFF
--- a/changes/pr3795.yaml
+++ b/changes/pr3795.yaml
@@ -1,5 +1,5 @@
-enhancement:
-  - "Add a new postgres task PostgresExecuteMany"
+task:
+  - "Add a new postgres task `PostgresExecuteMany` - [#3703](https://github.com/PrefectHQ/prefect/pull/3703)"
 
 contributor:
   - "[Faris ALSaleem](https://github.com/FarisALSaleem)"
--- a/changes/pr3795.yaml
+++ b/changes/pr3795.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Add a new postgres task PostgresExecuteMany"
+
+contributor:
+  - "[Faris ALSaleem](https://github.com/FarisALSaleem)"
--- a/src/prefect/tasks/postgres/postgres.py
+++ b/src/prefect/tasks/postgres/postgres.py
@@ -100,6 +100,104 @@ class PostgresExecute(Task):
             raise
 
 
+class PostgresExecuteMany(Task):
+    """
+    Task for executing many queries against a Postgres database.
+
+    Args:
+        - db_name (str): name of Postgres database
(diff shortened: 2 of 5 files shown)


src/prefect/tasks/postgres/postgres.py
```diff
@@ -100,6 +100,105 @@ def run(
...
+    @defaults_from_attrs("query", "data", "commit")
+    def run(
+        self,
+        query: str = None,
+        data: list = None,
+        commit: bool = False,
+        password: str = None,
+    ):
+        """
+        Task run method. Executes a query against Postgres database.
```
joshmeek, line 151:
```suggestion
        Task run method. Executes many queries against Postgres database.
```

src/prefect/tasks/postgres/postgres.py
```diff
@@ -100,6 +100,105 @@ def run(
...
+
+        Args:
+            - query (str, optional): query to execute against database
+            - data (list): list of values to use in query, must be specified using placeholder
```
joshmeek, line 155:
```suggestion
            - data (List[tuple], optional): list of values to use in query, must be specified using placeholder
```

joshmeek (commented):
Couple nit comments but otherwise looks good @FarisALSaleem

changes/pr3795.yaml
```diff
@@ -0,0 +1,5 @@
+enhancement:
+  - "Add a new postgres task PostgresExecuteMany"
```
joshmeek, line 2:
```suggestion
task:
  - "Add a new postgres task `PostgresExecuteMany` - [#3703](https://github.com/PrefectHQ/prefect/pull/3703)"
```

joshmeek (approved):

src/prefect/tasks/postgres/postgres.py
```diff
@@ -100,6 +100,106 @@ def run(
...
+                    else:
+                        conn.rollback()
+
+            conn.close()
+            return executed
+
+        # pass through error, and ensure connection is closed
+        except (Exception, pg.DatabaseError):
+            conn.close()
+            raise
```
amanda-wee, line 200: I'm way too late to the party, but it seems to me that this could be improved with:
```
try:
    with conn, conn.cursor() as cursor:
        executed = cursor.executemany(query=query, vars_list=data)
        if commit:
            conn.commit()
        else:
            conn.rollback()

    return executed

# ensure connection is closed
finally:
    conn.close()
```

PR #3708, opened 2020-11-23
Run config docs

jcrist:
- Adds initial docs on configuring flows using `RunConfig` objects
- Marks docs on using `Environment` objects as legacy, recommends users start transitioning over. We don't warn yet, since (IMO) this still isn't sufficient documentation to put users in a good place to succeed. Just enough for now to get people noticing and moving forward.
- Removes the "experimental" label from the run-config and ECS agent docs


joshmeek (approved):
These all look great to me, second opinion welcome because I may have missed a typo or two (or it's actually flawless) 🚀

jcrist:
Going to merge these for now, fixups can come in later PRs as needed.

PR #3714, opened 2020-11-25
Refactor execute query code for postgres tasks

amanda-wee:
## Summary

Refactor execute query code for `PostgresExecute`, `PostgresExecuteMany`, and `PostgresFetch` tasks.

## Changes

- Instead of nesting `with` statements, use a single `with` statement with two items.
- Instead of nested repeating `conn.close()` in both try and except blocks and explicitly listing exception types, use the finally block.

## Importance

- Avoid an unnecessary indentation level with no loss of readability.
- Avoid unnecessarily listing exception types only to re-raise the exception when the finally block is perfect for the aim of ensuring the connection is always closed.

## Checklist

This PR:

(Note that neither new tests nor docstring update is applicable.)


jcrist (approved):
Thanks!

PR #3715, opened 2020-11-25
Stream log output from Kubernetes RunNamespacedJob

joelluijmes:
## Summary

This PR opens a watch on `read_namespaced_pod_log` such that the logging output of the created job is forwarded to prefect. No more black box ⬛!

## Changes

Replaces the polling for status on a job, to a watch to forward the logging from the job.

## Importance

Gives insight in what the job is doing.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/kubernetes/job.py
+++ b/src/prefect/tasks/kubernetes/job.py
@@ -2,7 +2,7 @@ import time
 from typing import Any, Optional, cast
 
 from kubernetes import client
-from multiprocessing.pool import ThreadPool
+from concurrent.futures import ThreadPoolExecutor
 
 from prefect import Task
 from prefect.engine import signals
@@ -722,11 +722,7 @@ class RunNamespacedJob(Task):
         )
         self.logger.info(f"Job {job_name} has been created.")
 
-        # TODO: if a job has more active pods than the threadpool has threads
-        # logs will no longer be streamed live for all pods. Using an async
-        # k8s client would help remedy this.
-        # https://github.com/PrefectHQ/prefect/pull/3715#discussion_r538324514
-        pool = ThreadPool()
+        pool = ThreadPoolExecutor()
         pod_log_streams = dict()
 
         try:
@@ -763,7 +759,7 @@ class RunNamespacedJob(Task):
                         )
 
                         self.logger.info(f"Started following logs for {pod_name}")
-                        pod_log_streams[pod_name] = pool.apply_async(read_pod_logs.run)
+                        pod_log_streams[pod_name] = pool.submit(read_pod_logs.run)
 
                 if job.status.active:
                     time.sleep(job_status_poll_interval)
@@ -784,5 +780,5 @@ class RunNamespacedJob(Task):
                 self.logger.info(f"Job {job_name} has been deleted.")
         finally:
             if pool:
-                pool.close()
+                pool.shutdown()
                 pool = None
--- a/src/prefect/tasks/kubernetes/job.py
+++ b/src/prefect/tasks/kubernetes/job.py
@@ -1,12 +1,14 @@
 import time
-from typing import Any, cast
+from typing import Any, Optional, cast
 
 from kubernetes import client
+from multiprocessing.pool import ThreadPool
 
 from prefect import Task
 from prefect.engine import signals
 from prefect.utilities.tasks import defaults_from_attrs
 from prefect.utilities.kubernetes import get_kubernetes_client
+from prefect.tasks.kubernetes.pod import ReadNamespacedPodLogs
 
 
 class CreateNamespacedJob(Task):
@@ -613,9 +615,12 @@ class RunNamespacedJob(Task):
         - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
             which stored your Kubernetes API Key; this Secret must be a string and in
             BearerToken format
-        - job_status_pool_interval (int, optional): The interval given in seconds
+        - job_status_poll_interval (int, optional): The interval given in seconds
             indicating how often the Kubernetes API will be requested about the status
             of the job being performed, defaults to the `5` seconds
+        - log_level (str, optional): Log level used when outputting logs from the job
+            should be one of `debug`, `info`, `warn`, `error`, 'critical' or `None` to
+            disable output completely. Defaults to `None`.
         - delete_job_after_completion (bool, optional): boolean value determining whether
             resources related to a given job will be removed from the Kubernetes cluster
             after completion, defaults to the `True` value
@@ -629,7 +634,8 @@ class RunNamespacedJob(Task):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        job_status_pool_interval: int = 5,
+        job_status_poll_interval: int = 5,
+        log_level: Optional[str] = None,
         delete_job_after_completion: bool = True,
         **kwargs: Any,
     ):
@@ -637,7 +643,8 @@ class RunNamespacedJob(Task):
         self.namespace = namespace
         self.kube_kwargs = kube_kwargs or {}
         self.kubernetes_api_key_secret = kubernetes_api_key_secret
-        self.job_status_pool_interval = job_status_pool_interval
+        self.job_status_poll_interval = job_status_poll_interval
+        self.log_level = log_level
         self.delete_job_after_completion = delete_job_after_completion
 
         super().__init__(**kwargs)
@@ -647,7 +654,8 @@ class RunNamespacedJob(Task):
         "namespace",
         "kube_kwargs",
         "kubernetes_api_key_secret",
-        "job_status_pool_interval",
+        "job_status_poll_interval",
+        "log_level",
         "delete_job_after_completion",
     )
     def run(
@@ -656,7 +664,8 @@ class RunNamespacedJob(Task):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        job_status_pool_interval: int = 5,
+        job_status_poll_interval: int = 5,
+        log_level: Optional[str] = None,
         delete_job_after_completion: bool = True,
     ) -> None:
         """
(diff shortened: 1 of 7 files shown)


src/prefect/tasks/kubernetes/job.py
```diff
@@ -613,9 +614,9 @@ class RunNamespacedJob(Task):
         - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
             which stored your Kubernetes API Key; this Secret must be a string and in
             BearerToken format
-        - job_status_pool_interval (int, optional): The interval given in seconds
```
jcrist, line 616: Just noticed the typo here (should be `job_status_poll_interval`).

src/prefect/tasks/kubernetes/job.py
```diff
@@ -613,9 +614,9 @@ class RunNamespacedJob(Task):
...
-            indicating how often the Kubernetes API will be requested about the status
-            of the job being performed, defaults to the `5` seconds
+        - log_level (str, optional): Log level used when outputting logs from the job
```
jcrist, line 617: Some jobs will generate lots of logs, and not all users will want logs stored in prefect's backend. This should default to `None` IMO (so it's opt-in).

src/prefect/tasks/kubernetes/job.py
```diff
@@ -695,34 +698,68 @@ def run(
...
+        job = api_client_job.read_namespaced_job_status(
+            name=job_name, namespace=namespace
+        )
+
+        pod_selector = f"controller-uid={job.metadata.labels['controller-uid']}"
+        pods_list = api_client_pod.list_namespaced_pod(
+            namespace=namespace, label_selector=pod_selector
+        )
+
+        if len(pods_list.items) != 1:
```
jcrist, line 723: Some jobs might run multiple pods

src/prefect/tasks/kubernetes/job.py
```diff
@@ -695,34 +698,68 @@ def run(
...
+        pod_name = pod.metadata.name
+        self.logger.info(pod)
+        while pod.status.phase != "Running":
+            time.sleep(1)
+            pod = api_client_pod.read_namespaced_pod(name=pod_name, namespace=namespace)
+
+        self.logger.info(f"Pod {pod_name} is running.")
+
+        # The watch stream is automatically stopped when the job is completed or failed. So we don't
+        # have to poll the job its status.
```
jcrist, line 739: Watch streams can complete for other reasons (stale event requiring a restart, api error, etc...). Most code I've used them required an extra loop around `stream` calls to handle restarts/spurious errors.

src/prefect/tasks/kubernetes/job.py
```diff
@@ -695,34 +698,68 @@ def run(
...
+
+        api_client_job.create_namespaced_job(
+            namespace=namespace, body=body, **kube_kwargs
+        )
         self.logger.info(f"Job {job_name} has been created.")
 
-        completed = False
-        while not completed:
-            res = api_client.read_namespaced_job_status(
-                name=job_name, namespace=namespace
```
jcrist, line 708: Since:
- not all users will want to stream logs
- relying on a log stream for job completion isn't sufficient

I think we'll still want to check for job completion via the job status. The polling loop is one way, starting a watch stream on the job status would be another (I could go either way here).

jcrist (changes requested):
Thanks @joelluijmes. I agree that streaming the logs from a job into prefect sounds useful, but I think we'll want to make it opt-in.

src/prefect/tasks/kubernetes/job.py
joelluijmes: Hmm, I was thinking about this, but I thought it was impossible, how can a job create multiple pods?

joelluijmes:
> Thanks @joelluijmes. I agree that streaming the logs from a job into prefect sounds useful, but I think we'll want to make it opt-in.

Thanks for the quick review, I'll update the PR tomorrow with requested chagnes 👍

src/prefect/tasks/kubernetes/job.py
jcrist: By specifying `parallelism` or `completions`: https://kubernetes.io/docs/concepts/workloads/controllers/job/#parallel-jobs

joelluijmes:
Just worked on it, and reimplemented the polling loop in order to support multiple pods. However I'm now running into how `read_namespaced_pod_log` works, ideally we want to have pagination. Or an interruptible stream (such that it doesn't keep hanging). 

I tried to paginate the logs with `since_seconds` but that is not precise enough and skips logs if multiple logs are  happen in milliseconds. I could add an overlap second, but then we have to filter on duplicate lines but that is tricky because we don't know the job is making duplicates or not.

The possible solution as I now see:
1. Don't stream the logs, only retrieve them on completion (most easy, but I'd like to see the job logs in action). 
2. For each pod start a watch in a new thread. 

Do you have any suggestions or insights on this? 

--

I'll try to see what I can do with 2, honestly I always shudder a bit with threads as I'm used to co-routines. But I see that api_client of kubernetes internally also use a ThreadPool when param `async_req=True`, so maybe that is a direction I could take a look at.

joshmeek:
Hey @joelluijmes would you mind merging the latest from master into your branch? We merged in a PR yesterday that fixes some of the issues we have been seeing in the CI tests!

joelluijmes:
> Hey @joelluijmes would you mind merging the latest from master into your branch? We merged in a PR yesterday that fixes some of the issues we have been seeing in the CI tests!

Sure! I reverted the merge because I suspected it as the cause for the build failing 😬

joelluijmes:
Thought the diff was weird after the master merge. Probably did something wrong 🤷 
Anyhow, rebased the branch on to latest master and squashed into meaningful commits 😄 

Should be ready for review.

joelluijmes:
Any remarks on this?

src/prefect/tasks/kubernetes/job.py
```diff
@@ -695,34 +709,76 @@ def run(
...
+
+                        read_pod_logs = ReadNamespacedPodLogs(
+                            pod_name=pod_name,
+                            namespace=namespace,
+                            kubernetes_api_key_secret=kubernetes_api_key_secret,
+                            func_stream=lambda log: func_log(f"{pod_name}: {log}"),
+                        )
+
+                        self.logger.info(f"Started following logs for {pod_name}")
+                        pod_log_streams[pod_name] = pool.apply_async(read_pod_logs.run)
```
jcrist, line 762: Eventually the thread pool will fill up, so active streaming of logs will only work for jobs with n-pods <= n-threads. This is a bit unfortunate - it'd be easy to support more if we were using an async kubernetes client but we're not.

I think I'm fine with this for now, but in the future I may want to fix this.

src/prefect/tasks/kubernetes/pod.py
```diff
@@ -571,3 +573,118 @@ def run(
...
+    an override for the remote connection.
+    2. Attempt in-cluster connection (will only work when running on a Pod in a cluster)
+    3. Attempt out-of-cluster connection using the default location for a kube config file
+
+    Args:
+        - pod_name (str, optional): The name of a pod to replace
+        - namespace (str, optional): The Kubernetes namespace to patch this pod in,
+            defaults to the `default` namespace
+        - func_stream (Callable, optional): Stream the logs to provided function,
+            for each line the function is called.
```
jcrist, line 602: I'd find the following kwarg name (and docstring) clearer. Same in the `run` method below.

```
        - on_log_entry (Callable, optional): If provided, will stream the pod logs
            calling the callback for every line (and the task returns `None`). If not
            provided, the current pod logs will be returned immediately from the task.
```

src/prefect/tasks/kubernetes/pod.py
```diff
@@ -571,3 +573,118 @@ def run(
         api_client.replace_namespaced_pod(
             name=pod_name, namespace=namespace, body=body, **kube_kwargs
         )
+
+
+class ReadNamespacedPodLogs(Task):
```
jcrist, line 578: Can you add this to the api docs (add to the list of classes [here](https://github.com/PrefectHQ/prefect/blob/928d2b9e9b54847e3d021a544c9cdeb8ea1ba10b/docs/outline.toml#L312)).

jcrist (commented):
Apologies for the delayed review, this is looking pretty good! Only a few small fixes needed.

src/prefect/tasks/kubernetes/job.py
jcrist: Actually, would you mind adding a todo comment above this line with that info? Something like

```
# TODO: if a job has more active pods than the threadpool has threads
# logs will no longer be streamed live for all pods. Using an async
# k8s client would help remedy this.
```

joelluijmes: Aha alright, I'm not that well versed in Python multiprocessing, but good remark, I'll add it! An async kubernetes client would probably make this bit easier 😬 

To what size does Python's threadpool grow?

joelluijmes:
> Apologies for the delayed review, this is looking pretty good! Only a few small fixes needed.

No worries, just like to see this merged so I can use it 🤓 
Great, updated with the requested changes!

src/prefect/tasks/kubernetes/job.py
jcrist: > To what size does Python's threadpool grow?

It's static, and equal to the number of cores.

Actually we might want to switch to a [`concurrent.futures.TheadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor) instead. This will grow in size but starts no threads by default (the default cap is `min(32, (os.cpu_count() or 1) + 4)`). You'd use `submit` instead of `apply_async` and `shutdown` instead of `close`, but everything else should be equivalent.

joelluijmes:
Okay, better late then never, but completed the requested changes 👍 

One thing we may need to look out for is following. The kubernetes library list this as comment on the stream
```
Note that watching an API resource can expire. The method tries to
        resume automatically once from the last result, but if that last result
        is too old as well, an `ApiException` exception will be thrown with
        ``code`` 410. In that case you have to recover yourself, probably
        by listing the API resource to obtain the latest state and then
        watching from that state on by setting ``resource_version`` to
        one returned from listing.
```

However, when passing `resource_version`, None or the actual value, the generator doesn't yield any logs. Maybe something is broken in the Kubernetes library, or this is just not supported for `read_namespaced_pod_log`. In my experience, it just doesn't work. I tried it with two different Kubernetes clusters (GKE at 1.19 and a bare metal at 1.17).

joelluijmes:
Just noticed that the build wasn't successful after my last commit. I've rebased it on latest master and now it's green. Hope it can be merged soon 👍

jcrist:
Oop, apologies for letting this linger. I'll try to finish this up today and get it in. Thanks @joelluijmes!

jcrist (approved):

PR #3693, opened 2020-11-20
Fix display of docker errors during storage build

madkinsz:
## Summary

Fixes missing display of docker error messages during image build. On docker-py version 4.3.1 and Docker version 19.03.13, build 4484c46d9d the message for a syntax error in the dockerfile was in a format we did not check for. This gave me a `ValueError` suggesting a healthcheck had failed when really there was a dockerfile syntax error that was hidden.

## Changes

- Adds a check for "message" in the docker build output parsing

## Importance

Helps debug bad Docker storage setups

## Checklist

This PR:

- ~[ ] adds new tests (if appropriate)~
- [x] adds a change file in the `changes/` directory (if appropriate)
- ~[ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)~

PR #3717, opened 2020-11-25
Fix docker storage bug when output is `None`

madkinsz:
## Summary

Fixes bug introduced in #3693 where an error is raised due to `None` values during docker storage build output parsing.

## Changes

- Adds a default for `output` as an empty string instead of `None`
- Changes output parsing to a static method (does not use `self`)
- Adds test coverage for output parsing

## Importance

Bugfix

## Checklist

This PR:

### DIFF
--- a/changes/pr3717.yaml
+++ b/changes/pr3717.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Fix bug with docker storage throwing exception while trying to display output - [#3717](https://github.com/PrefectHQ/prefect/pull/3717)"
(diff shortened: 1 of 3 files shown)


madkinsz:
Going to add test coverage as well

changes/pr3717.yaml
```diff
@@ -0,0 +1,2 @@
+enhancement:
```
jcrist, line 1: Should be a `fix:` instead.

madkinsz: Yep!

jcrist (approved):

PR #3719, opened 2020-11-25
Hide private methods in api docs

jcrist:
Previously any non `_`-prefixed method in a class was documented as part
of the api docs. For things like agents (which are more an application
than a class), there were several non-public methods that aren't
`_`-prefixed. Having these in the api docs led some users to be
confused, thinking they could call those internal methods directly.

We now add a way to specify a set of methods to document, and apply this
to the agent api docs to only document the `start` methods.

### DIFF
--- a/docs/generate_docs.py
+++ b/docs/generate_docs.py
@@ -3,7 +3,9 @@ Functionality for auto-generating markdown documentation.
 
 Each entry in `OUTLINE` is a dictionary with the following key/value pairs:
     - "page" -> (str): relative path to the markdown file this page represents
-    - "classes" -> (list, optional): list of classes to document
+    - "classes" -> (list or dict, optional): list of classes to document. If a
+        dict, the keys are class names and the values are lists of methods to
+        document.
     - "functions" -> (list, optional): list of standalone functions to document
     - "title" -> (str, optional): title of page
     - "top-level-doc" -> (object, optional): module object that contains the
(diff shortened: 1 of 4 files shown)


docs/generate_docs.py
```diff
@@ -51,6 +53,11 @@ def patched_import(*args, **kwargs):
         builtins.__import__ = real_import
 
 
+def import_object(name):
+    module, attr = name.rsplit(".", 1)
+    return getattr(importlib.import_module(module), attr)
```
madkinsz, line 58: @cicdw this is what you'd want to do in the event pr where I mentioned importlib

cicdw: Yup made those changes a while back

docs/generate_docs.py
```diff
@@ -354,12 +350,18 @@ def format_subheader(obj, level=1, in_table=False):
...
-        obj, predicate=lambda x: inspect.isroutine(x) and obj.__name__ in x.__qualname__
-    )
-    public_members = [method for (name, method) in members if not name.startswith("_")]
-    return public_members
+def get_class_methods(obj, methods=None):
+    if methods is None:
+        members = inspect.getmembers(
+            obj,
+            predicate=lambda x: inspect.isroutine(x) and obj.__name__ in x.__qualname__,
+        )
```
madkinsz, line 358: Can you add a comment explaining what this predicate filters?

madkinsz (commented):
Sweet!

docs/generate_docs.py
jcrist: to be honest, I'm not sure. This code predates me, just moved it into a branch.

joshmeek (commented):
Do you think you could add a note about this new option to https://docs.prefect.io/core/development/documentation.html#api-reference?

jcrist:
Done

Issue #3657, opened 2020-11-12
Returning BigQuery LoadJob causes pickling issue when running flow via agent


fwinokur:
## Description
When running a flow via an agent that calls `BigQueryLoadFile`, the returned object (of type google.cloud.bigquery.job.LoadJob) causes a `_pickle.PicklingError` despite a successful run of this task.
```
[2020-11-12 15:08:48-0500] INFO - prefect.CloudTaskRunner | Task 'BigQueryLoadFile': Starting task run...
[2020-11-12 15:08:49-0500] INFO - prefect.CloudTaskRunner | Task 'BigQueryLoadFile': Finished task run for task with final state: 'Mapped'
[2020-11-12 15:08:49-0500] INFO - prefect.CloudTaskRunner | Task 'BigQueryLoadFile[0]': Starting task run...
[2020-11-12 15:09:10-0500] ERROR - prefect.CloudTaskRunner | Unexpected error: PicklingError('Pickling client objects is explicitly not supported.\nClients have non-trivial state that is local and unpickleable.')
Traceback (most recent call last):
  File "/path/to/virtualenv/lib/python3.8/site-packages/prefect/engine/runner.py", line 48, in inner
    new_state = method(self, state, *args, **kwargs)
  File "/path/to/virtualenv/lib/python3.8/site-packages/prefect/engine/task_runner.py", line 891, in get_task_run_state
    result = self.result.write(value, **formatting_kwargs)
  File "/path/to/virtualenv/lib/python3.8/site-packages/prefect/engine/results/gcs_result.py", line 75, in write
    binary_data = new.serializer.serialize(new.value)
  File "/path/to/virtualenv/lib/python3.8/site-packages/prefect/engine/serializers.py", line 70, in serialize
    return cloudpickle.dumps(value)
  File "/path/to/virtualenv/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/path/to/virtualenv/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 563, in dump
    return Pickler.dump(self, obj)
  File "/path/to/virtualenv/lib/python3.8/site-packages/google/cloud/client.py", line 163, in __getstate__
    raise PicklingError(
_pickle.PicklingError: Pickling client objects is explicitly not supported.
Clients have non-trivial state that is local and unpickleable.
```

[...]

PR #3724, opened 2020-11-27
Fix PicklingError in BigQueryLoadFile and BigQueryLoadGoogleCloudStorage #3657

yukihira1992:
## Summary
Fix issue #3657 

After remove `_client` and `_completion_lock`, the LoadJob object is pickleable.

```python
import cloudpickle
from google.cloud import bigquery

client = bigquery.Client()
table_ref = client.dataset('dataset').table('table')
job_config = bigquery.LoadJobConfig(
    source_format=bigquery.SourceFormat.CSV,
)
with open('data.csv', 'rb') as f:
    load_job = client.load_table_from_file(
        f,
        table_ref,
        job_config=job_config,
    )

load_job.result()

# remove unpickleable attributes
load_job._client = None
load_job._completion_lock

# this code works
cloudpickle.dumps(load_job)
```

## Changes
Override unpickleable attributes with None before return from the run method.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/gcp/bigquery.py
+++ b/src/prefect/tasks/gcp/bigquery.py
@@ -382,6 +382,10 @@ class BigQueryLoadGoogleCloudStorage(Task):
         load_job = client.load_table_from_uri(uri, table_ref, job_config=job_config)
         load_job.result()  # block until job is finished
 
+        # remove unpickleable attributes
+        load_job._client = None
+        load_job._completion_lock = None
+
         return load_job
 
 
@@ -535,6 +539,10 @@ class BigQueryLoadFile(Task):
             raise IOError(f"Can't open and read from {path.as_posix()}.")
 
         load_job.result()  # block until job is finished
+
+        # remove unpickleable attributes
+        load_job._client = None
+        load_job._completion_lock = None
 
         return load_job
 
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/gcp/bigquery.py
```diff
@@ -382,6 +382,10 @@ def run(
         load_job = client.load_table_from_uri(uri, table_ref, job_config=job_config)
         load_job.result()  # block until job is finished
 
+        # remove unpickleable attributes
+        load_job._client = None
+        load_job._completion_lock = None
```
jcrist, line 387: I'm a bit wary of modifying private attributes on objects in non-prefect libraries. If the implementation in the bigquery library changes, this may no longer work, and it may also lead to errors for users relying on methods in the `load_job` object. Perhaps pickling support should be fixed upstream in the bigquery library instead?

yukihira1992: As you say, this is a temporary patch and these methods are relying on the `_client` and `_completion_lock` attribute.
- `exists`
- `reload`
- `cancel`

> Perhaps pickling support should be fixed upstream in the bigquery library instead?

Hmm, I think prefect is responsible for serialization, not bigquery library.
If we want to fix it correctly, I think prefect should define an interface of pickleable return value instead of the LoadJob object itself.

jcrist: > Hmm, I think prefect is responsible for serialization, not bigquery library.

We use `pickle`, which I'd argue most objects in Python should support. I think an issue report to the upstream bigquery library for supporting pickling here would be useful and valid.

That said, the fix here seems fine for now.

jcrist (approved):

PR #3725, opened 2020-11-27
Allow custom exporter for ExecuteNotebook task 

### LINKED ISSUE #3687: Allow exporter arguments in Jupyter ExecuteNotebook task
## Current behavior

When running the `jupyter.jupyter.ExecuteNotebook` task with `output_format='html'` the default settings for the HTMLExporter are used. There is no way to pass arguments to this exporter.

## Proposed behavior

Allow passing arguments to the HTMLExporter.

## Implementation suggestion

Something like `html_exporter = nbconvert.HTMLExporter(**exporter_kwargs)` on the following line:
https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/jupyter/jupyter.py#L65

## Example usecase

This allows you to exclude code cells, only showing their output, in the exported html document by passing the `exclude_input=True` argument to the exporter.

swierh:
## Summary

ExecuteNotebook task now takes an optional `exporter` argument that (when provided) is used to export the notebook.

Fixes #3687

## Changes

The ExecuteNotebook task now takes an optional `exporter` argument that (when provided) is used to export the notebook.

## Importance

This allows for custom exporters to any format, and not just html or json. In addition, it allows for an html-export with non-default HTMLExporter arguments.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/jupyter/jupyter.py
+++ b/src/prefect/tasks/jupyter/jupyter.py
@@ -42,9 +42,10 @@ class ExecuteNotebook(Task):
         self.parameters = parameters
         self.output_format = output_format
         self.kernel_name = kernel_name
+        self.exporter_kwargs = exporter_kwargs
         super().__init__(**kwargs)
 
-    @defaults_from_attrs("path", "parameters", "output_format")
+    @defaults_from_attrs("path", "parameters", "output_format", "exporter_kwargs")
     def run(
         self,
         path: str = None,
--- a/src/prefect/tasks/jupyter/jupyter.py
+++ b/src/prefect/tasks/jupyter/jupyter.py
@@ -22,7 +22,7 @@ class ExecuteNotebook(Task):
             nbconvert Exporter name.
             Valid exporter names: asciidoc, custom, html, latex, markdown,
             notebook, pdf, python, rst, script, slides, webpdf. (default: notebook)
-        - exporter_kwargs (dict, optional): The arguments used for initializing 
+        - exporter_kwargs (dict, optional): The arguments used for initializing
             the exporter.
         - kernel_name (string, optional): kernel name to run the notebook with.
             If not provided, the default kernel will be used.
@@ -63,7 +63,7 @@ class ExecuteNotebook(Task):
             nbconvert Exporter name.
             Valid exporter names: asciidoc, custom, html, latex, markdown,
             notebook, pdf, python, rst, script, slides, webpdf. (default: notebook)
-        - exporter_kwargs (dict, optional): The arguments used for initializing 
+        - exporter_kwargs (dict, optional): The arguments used for initializing
             the exporter.
         """
         nb: nbformat.NotebookNode = pm.execute_notebook(
--- a/src/prefect/tasks/jupyter/jupyter.py
+++ b/src/prefect/tasks/jupyter/jupyter.py
@@ -19,7 +19,7 @@ class ExecuteNotebook(Task):
         - parameters (dict, optional): dictionary of parameters to use for the notebook
             Can also be provided at runtime
         - output_format (str, optional): Notebook output format, should be a valid
-            nbconvert Exporter name.
+            nbconvert Exporter name. 'json' is treated as 'notebook'.
             Valid exporter names: asciidoc, custom, html, latex, markdown,
             notebook, pdf, python, rst, script, slides, webpdf. (default: notebook)
         - exporter_kwargs (dict, optional): The arguments used for initializing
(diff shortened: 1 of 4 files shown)


src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -19,7 +19,10 @@ class ExecuteNotebook(Task):
         - parameters (dict, optional): dictionary of parameters to use for the notebook
             Can also be provided at runtime
         - output_format (str, optional): Notebook output format.
+            This argument is gnored if `exporter` is provided.
```
jcrist, line 22: Typo (if we keep the interface as implemented)

```suggestion
            This argument is ignored if `exporter` is provided.
```

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -19,7 +19,10 @@ class ExecuteNotebook(Task):
...
             Currently supported: json, html (default: json)
+        - exporter (nbconvert.Exporter, optional): The exporter that is used for
+            exporting the notebook, regardless of `output_format`.
```
jcrist, line 25: Interesting, I had originally envisioned supporting this as a combination of `output_format` (for specifying the exporter class) and optional `exporter_kwargs` for instantiating that class. Docstring would be something like:

```python
"""
output_format (str, optional): Notebook output format. One of {'json', 'html'},
  or a full import path to a `nbconvert.Exporter` class.
exporter_kwargs (dict, optional): Extra arguments to use when instantiating
  the `nbconvert.Exporter`.
"""
```

I think this is nice, since many users won't need to configure the exporters, so using a simple name alias (like "html") is sufficient. And for those that need to configure extra options, they can pass them in as a simple dict rather than having to import from `nbconvert`.

I saw that @madkinsz suggested the approach you're using, but I don't find that argument convincing - really the `json` output format should be passing through a `nbconvert.exporters.notebook.NotebookExporter` anyway.

Could go either way on this though, just my 2 cents.

jcrist (commented):
Thanks for working on this @swierh, just a few comments.

src/prefect/tasks/jupyter/jupyter.py
madkinsz: Yeah I don't like that the `output_format` gets ignored in this implementation. I think we should actually be using https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/base.py#L91 to retrieve the exporter passed as `output_format` and then initialize it with `exporter_kwargs`. This would automatically support a bunch of short names as well as full import paths.

```
In [8]: get_export_names()
Out[8]: 
['asciidoc',
 'custom',
 'html',
 'latex',
 'markdown',
 'notebook',
 'pdf',
 'python',
 'rst',
 'script',
 'slides']

In [9]: get_exporter('html')
Out[9]: nbconvert.exporters.html.HTMLExporter
``` 

It's not 100% clear to me how the json format should be handled though.

jcrist: Ah, didn't know about `get_exporter`, yeah we should move to use that. For `json`, this matches `notebook` - we should change the default to `notebook` and keep around `json` as an alias.

swierh: I agree with @madkinsz that `export` in combination with `get_exporter` is the way to go.

Two things I'm still not entirely sure about:
1. Should you still be able to provide your own Exporter instance or are short names and import paths for `get_exporter` handles enough?
2. How to document the valid output_format options, for now I've just listed all of them.

src/prefect/tasks/jupyter/jupyter.py
```diff
@@ -53,17 +59,20 @@ def run(
...
-        if output_format == "html":
-            html_exporter = nbconvert.HTMLExporter()
-            (body, resources) = html_exporter.from_notebook_node(nb)
-            return body
 
-        raise NotImplementedError("Notebook output %s not supported", output_format)
+        if exporter_kwargs is None:
+            exporter_kwargs = {}
+
+        exporter = nbconvert.get_exporter(output_format)
```
jcrist, line 76: For backwards compatibility we'll still need to handle `"json"` as an alias of `notebook`. No need to document it in the docstring, just something like:

```
if output_format == "json":
    output_format = "notebook"
exporter = nbconvert.get_exporter(output_format)
```

jcrist (commented):
> Should you still be able to provide your own Exporter instance or are short names and import paths for get_exporter handles enough?

I think they're enough. If people need more flexibility later we can add it then.

> How to document the valid output_format options, for now I've just listed all of them.

How you've done it now seems fine to me.

src/prefect/tasks/jupyter/jupyter.py
madkinsz: re 2. I'd point them to the nbconvert `get_export_names()` as well but it's not a big deal.

tests/tasks/jupyter/test_jupyter.py
```diff
@@ -13,6 +14,18 @@ def test_jupyter_html_output():
     assert "a*b=10" in output
 
 
+def test_jupyter_custom_exporter():
+    exporter = HTMLExporter()
+
+    task = ExecuteNotebook(
+        "tests/tasks/jupyter/sample_notebook.ipynb",
+        parameters=dict(a=5),
+        exporter=exporter,
```
jcrist, line 23: Looks like you need to update this test.

jcrist (approved):
Thanks!

bachsh:
Apologies for jumping into the party so late, but just wanted to say thanks to @swierh; that's a great addition to the task.

PR #3735, opened 2020-11-30
Drop dask_cloudprovider from install in testing CI jobs and pin pip version

joshmeek:
Attempting to fix an install issue we are seeing on the 3.6 circle jobs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,9 @@ extras["all_extras"] = sum(extras.values(), [])
 # CI extras to control dependencies for tests
 extras["test_ci"] = sum(extras.values(), [])
 if sys.version_info < (3, 7):
-    extras["test_ci"].remove("dask_cloudprovider")
+    extras["test_ci"] = [
+        r for r in extras["test_ci"] if not r.startswith("dask_cloudprovider")
+    ]
 
 cmdclass = {
     "verify_version": VerifyVersionCommand,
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -71,6 +71,9 @@ def run(self):
 if sys.version_info < (3, 6):
     extras["dev"].remove("black")
 
+if sys.version_info < (3, 7):
+    del extras["dask_cloudprovider"]
```
jcrist, line 75: I don't think we want to remove the extra from users (some users may be relying on this, though at this point 3.6 users are in the minority). Rather, we just want to drop it from our 3.6 test builds for now (though that may be tricky with our ci setup).

yukihira1992:
FYI [pip 20.3 (2020-11-30)](https://pip.pypa.io/en/stable/news/#id1)

> The new resolver now resolves packages in a deterministic order. (#9100)

jcrist (dismissed):

jcrist (approved):

setup.py
```diff
@@ -73,6 +73,11 @@ def run(self):
 
 extras["all_extras"] = sum(extras.values(), [])
 
+# CI extras to control dependencies for tests
+extras["test_ci"] = sum(extras.values(), [])
```
jcrist, line 77:
```suggestion
```

PR #3750, opened 2020-12-02
Support for flows with `flow.environment = None`

jcrist:
This adds support for flows with no `environment` AND no `run_config`.

Currently all flow's have an environment created by default. If a user explicitly sets a `run_config` on a flow, the new `run_config` based configuration is used, otherwise the default `environment` is used.

In 0.14.0 we plan to remove the default `environment`, so flow's will by default have `run_config=None, environment=None`. This change adds support for that, without removing the default (yet).

The new logic is now:

- If a flow has a run-config configured, use that. Agents check that the run-config set on the flow matches the type they support.
- If a flow has no environment, the agent will create the default `RunConfig` for that agent type. This means that a flow registered with no `run_config` configured on it is runable by all agents (storage and labels dependent), matching the existing behavior. A `UniversalRun` run-config is created on the client side in this case, so that a flow with nothing else created can still store the auto-generated labels present on `Storage` objects.
- If a flow has an `environment`, the old environment-based code path is taken.

Some other changes were needed throughout to get registration to work (mostly in `flow.register`.

Note that the `UniversalRun` run config is exposed as a new run config option, for those who want to configure labels on a flow but don't want to bind that flow to a specific agent type. This can be useful if all config can be handled on the agent, with nothing backend-specific on the flow. I wouldn't expect this to be used much (it was only added to be the default), but some users may find it useful. Backend-specific functionality still requires a backend-specific run config.

This PR:

### DIFF
--- a/src/prefect/agent/local/agent.py
+++ b/src/prefect/agent/local/agent.py
@@ -9,7 +9,6 @@ from prefect.agent import Agent
 from prefect.environments.storage import Docker
 from prefect.run_configs import LocalRun
 from prefect.serialization.storage import StorageSchema
-from prefect.serialization.run_config import RunConfigSchema
 from prefect.utilities.agent import get_flow_run_command
 from prefect.utilities.graphql import GraphQLResult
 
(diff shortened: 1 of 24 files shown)


jcrist:
I'm able to register a flow with `flow.environment=None` and no explicitly set `run_config`, but something has gone wrong in the backend (only tested on cloud so far). Flow runs are created, but not pulled down by the agent, and I can't cancel a created flow run due to a backend error. I'll take a look at this more tomorrow.

jcrist:
Ok, this has been fixed (already worked in server, cloud needed a small patch). Should be good-to-go.

src/prefect/agent/local/agent.py
```diff
@@ -147,21 +146,7 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
-                    flow_run.id,
-                    type(run_config).__name__,
-                )
-                raise TypeError(
-                    "Unsupported RunConfig type: %s" % type(run_config).__name__
-                )
-        else:
-            run_config = None
-
+        run_config = self._get_run_config(flow_run, LocalRun)
```
joshmeek, line 149: Is there a reason why this one doesn't also have an `assert` like the other agents?

jcrist: we disable `mypy` on this file for some reason. The `assert` is just to appease mypy.

joshmeek (approved):
This all looks good to me 👍

PR #3751, opened 2020-12-02
Black formatting 

joshmeek:
Running `black .` this suddenly happened for me locally, not sure if there was an update or not 🤔


joshmeek:
Hmm my local black version is `black-20.8b1` which matches that used in the tests here and I am getting a different result. This is happening for me on 3.8 but the tests are using 3.7 so that could be it. Has anyone else run into this before?

joshmeek:
Closing for now, can't seem to pinpoint what's going wrong but it appears to be some version mismatch error

joshmeek:
Appears to have been a cache issue, running `rm -r ~/Library/Caches/black` solved it!

Issue #3747, opened 2020-12-01
Increase transparency around K8s Agent deployed jobs


joshmeek:
## Current behavior
Currently there are a few places where there exists a slight black box around jobs deployed by the kubernetes agent. A couple small enhancements / PRs could be made in order to remove some of these issues with visibility.

## Proposed behavior
- [x] When the agent cleans up jobs we find those that have killed/failed pods and under those circumstances we send a flow run log with some information (e.g. last log, kubernetes reason) that leads to an explanation of why the pod was killed
- [x] Audit which agent level logs we currently surface for jobs, find out which ones are useful, need to be cleaned up/removed, and what needs to be added
- [x] Increase the amount of relevant information that we surface from agent logs that are focused around flow run deployment level events
- [x] Display information about what is happening when a flow run is stuck in submitted. This could be something such as watching job events from creation time to the time it starts running.

PR #3752, opened 2020-12-02
Surface pod failures and container errors in jobs deployed with the Kubernetes Agent

joshmeek:
## Summary
Related to some of the points outlined in #3747 

This adds a step to the k8s agent's `manage_jobs` to report on failed jobs and container errors. Previously if there were a container errors the flow run would stay in a running state until it was picked back up by lazarus and repeated multiple times. Now if there are container errors causing to pod failure it will be reported. Also the container status is reported as a log in this case.

Example:
image

image

Open to enhancements that would make this particular action more beneficial / any other information we can expose.

## Changes
Adds a step to the k8s agent's `manage_jobs` to report on failed jobs and container errors.

## Importance
More efforts to remove any black boxes in flow deployment!

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -215,7 +215,7 @@ class KubernetesAgent(Agent):
                             self.client.set_flow_run_state(
                                 flow_run_id=flow_run_id,
                                 state=Failed(
-                                    message="Kubernetes Error: the pods {} failed for this job".format(
+                                    message="Kubernetes Error: pods {} failed for this job".format(
                                         failed_pods
                                     )
                                 ),
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -1,6 +1,7 @@
 import os
 import time
 import uuid
+import textwrap
 from typing import Iterable, List, Any
 
 import json
@@ -208,6 +209,28 @@ class KubernetesAgent(Agent):
 
                             failed_pods.append(pod.metadata.name)
 
+                            for status in pod.status.container_statuses:
+                                if status.state.terminated:
+                                    msg = textwrap.dedent(
+                                        f"""
+                                        Found failed container: {status.name}
+                                            Exit code: {status.state.terminated.exit_code}
+                                            Message: {status.state.terminated.message}
+                                            Reason: {status.state.terminated.reason}
+                                            Signal: {status.state.terminated.signal}
+                                        """
+                                    )
+                                    self.client.write_run_logs(
+                                        [
+                                            dict(
+                                                flow_run_id=flow_run_id,
+                                                name=self.name,
+                                                message=msg,
+                                                level="ERROR",
+                                            )
+                                        ]
+                                    )
+
                         if failed_pods:
                             self.logger.debug(
                                 f"Failing flow run {flow_run_id} due to the failed pods {failed_pods}"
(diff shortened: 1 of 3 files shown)


src/prefect/agent/kubernetes/agent.py
```diff
@@ -191,6 +193,58 @@ def manage_jobs(self) -> None:
...
+
+                        failed_pods = []
+                        for pod in pods.items:
+                            if pod.status.phase != "Failed":
+                                continue
+
+                            failed_pods.append(pod.metadata.name)
+
+                            for status in pod.status.container_statuses:
+                                if status.state.terminated:
```
jcrist, line 213: Not all terminated containers are indications of failure - if a user has a sidecar container for a pod that exited successfully that doesn't necessarily need to be logged.

Perhaps instead of writing one log record for each container in a pod, we write one log record that includes info on all containers in the pod?

We might want to also include only non-null fields (everything but `exit_code` below is optional and may be null.

So something like:

```
Pod <pod name> failed.
  Container <container name>:
    all non-null fields as you have them
  Container <container name>:
    ...
```

I think including successful containers seems fine (more information for the user), but like not having an ERROR log that only mentions a successful container (grouping the log message together fixes that, since one of those containers will be a failure).

jcrist (commented):
Overall this looks pretty good to me.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -191,6 +192,72 @@ def manage_jobs(self) -> None:
...
+                                        level="ERROR",
+                                    )
+                                ]
+                            )
+
+                        if failed_pods:
+                            self.logger.debug(
+                                f"Failing flow run {flow_run_id} due to the failed pods {failed_pods}"
+                            )
+                            self.client.set_flow_run_state(
```
jcrist, line 251: I could see cases where the flow run already completed successfully (per the prefect backend) but something happened between reporting that to the backend and the container exiting. I'm not sure if that should trigger a flow run to be marked as failed.

Likewise, I think there are cases where a flow run might have already reported itself as failed (for a flow-level reason) before having it exit with a non-zero exit code. I'm not sure if we want to update the state to failed again.

Perhaps we only want to set the flow run state if it hasn't moved to a finished state already? Not sure if there's an easy way to do this, but it seems like a better user experience to me.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -191,6 +192,72 @@ def manage_jobs(self) -> None:
...
+                                if status.state.terminated:
+                                    msg += f"\t\tExit Code:: {status.state.terminated.exit_code}\n"
+                                    if status.state.terminated.message:
+                                        msg += f"\t\tMessage: {status.state.terminated.message}\n"
+                                    if status.state.terminated.reason:
+                                        msg += f"\t\tReason: {status.state.terminated.reason}\n"
+                                    if status.state.terminated.signal:
+                                        msg += f"\t\tSignal: {status.state.terminated.signal}\n"
+
+                                pod_failure_message += msg
```
jcrist, line 233: Additively building a string like this can be inefficient (it might not matter for N this small though). You might instead collect a list of rows, then join them at the end with `'\n'.join(msg_rows)`. This also drops the need to `.strip()`  before reporting the message, since there won't be a trailing `'\n'`.

jcrist (approved):
LGTM!

PR #3753, opened 2020-12-02
Enhance the RunGreatExpectationsCheckpoint task

### LINKED ISSUE #2436: Design/discuss integration of Great Expectations with Prefect (potentially in Results feature)
Discussion issue to figure out a plan to make [Great Expectations](https://github.com/great-expectations/great_expectations) integrate seamlessly with Prefect.

In general we are assuming that people want to be able to configure Great Expectations data validators to track onto pieces of data passed through a flow, and want the pipeline to manage actually calling the GE API for them and surfacing errors/alerts/failures in some way that feels totally integrated into Prefect.

Some user experience questions to consider:

a) What is an example of the Python API a core user would use to attach a single great expectation assertion to a task? Multiple? The same one to many tasks?
a1) Do they even attach them to tasks? Do they attach them to something else (ie Results?)
a2) What about a Core server/Cloud user: is there ever a world where GE validators are configured directly through the UI?

b) How are the validation results from great expectations surfaced in the Prefect logs? Are they visualized somehow in the UI?

c) Are there assumptions or conventions Prefect should make/support to autodetect GE assertions on disk? How does this relate to the "Expectations on rails" framework in beta in GE?
c1) how does Prefect (and --dun dun dun-- dask) play with the fact that Great Expectations is mainly configured using file based configuration?

d) Where/how in the pipeline do we do validation checks for people? Where/how should users configure to turn this check on or off, besides removing the validator from the code (for example, in a global configuration toggle?)

e) If people call great expectation validators themselves in a task via a Prefect API such as `Result.validate()`, do we do anything special for them with the output?

f) can we provide better, Prefect-based semantics for what to do on failure of a GE validation since the pipeline can control the execution flow in reaction to the validation failure (ex potentially allow flow configuration for retrying up a task tree when a downstream validator fails)

g) can we integrate the GE data docs metadata into our UI somehow (simplest case is to link out, though this can get infinitely more fancy)

**Curveball question: Is there a need (either in addition or in replacement of integrated pipeline checks) of an abstract GE task in the task library that can be easily used as a terminal/reference task?

### LINKED ISSUE #3057: Improve or create new GE task library task to run a checkpoint against in-memory config
## Current behavior
Today there is a [great expectations task in the task library to run GE checkpoints](https://docs.prefect.io/api/latest/tasks/great_expectations.html#prefect-tasks-great_expectations-checkpoint-rungreatexpectationscheckpoint), but it is dependent on being run in the correct environment to find the local data configuration files specific to the expectation suite being run, so it is not super great in a distributed environment. In addition, it doesn't meaningfully utilize the dataflow semantics of Prefect since the GE checkpoint infrastructure re-reads the data into memory based on config of where it is stored (disk/sql/spark).

## Proposed behavior
Preferably, you should be able to pass all the relevant configuration to run a GE checkpoint as Python objects in-memory so we don't have to depend on the local filesystem matching up on a given worker, including
1. any config GE needs to know about the expectation suite store
2. any config GE needs to know about *which* actions it should take (including whether to store to a validation store or rebuild data docs)
3. any config GE needs to know to *actually do* the actions it should take (such as where the validation store IS and how to connect to it)
4. the actual data object to validate in memory

After talking some with the GE folks (cc @jcampbell, @abegong) there is a way to configure your [expectations suites](https://docs.greatexpectations.io/en/latest/how_to_guides/configuring_metadata_stores/how_to_configure_an_expectation_store_in_amazon_s3.html) and [validation store](https://docs.greatexpectations.io/en/latest/how_to_guides/configuring_metadata_stores/how_to_configure_a_validation_result_store_in_s3.html) in S3. In regular usage, you still need a `great_expectations.yml` file that points to those locations and to configure the data contexts that you are actually running those expectations against. BUT also from talking with them it sounds that by instantiating the [BaseDataContext class](https://github.com/great-expectations/great_expectations/blob/4f2e634502d81bbeaf5e719b8e6611f26a12a4a4/great_expectations/data_context/data_context.py#L84) directly, we should be able to pass a config object in memory that can be composed of all the relevant stuff we need from `great_expectations.yml` (we will just have to reverse engineer it a bit I think because I can't find any good API docs on what that config object should look like). According to them this class was originally made to make testing easier to do in memory, so there should be some inspiration in the test files (in particular [here is a datacontextconfig](https://github.com/great-expectations/great_expectations/blob/4f2e634502d81bbeaf5e719b8e6611f26a12a4a4/tests/data_context/test_data_context.py#L591) used for testing that should be good inspo for what we need to cover from `great_expectations.yml` in memory).

Finally, the last piece to make this most Prefect-y is probably to pass the actual piece of data as an in-memory object instead of as a lot of configuration around files/sql/spark dfs. I don't have a great idea of how to do this yet so this may need to be punted to another issue. I had originally tried to do this by instantiating batches and passing them directly into [`DataContext.run_validation_operator.assets_to_validate`](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/great_expectations/checkpoints.py#L110) but those actually must be [`DataAsset`s or arguments to get_batch](https://github.com/great-expectations/great_expectations/commit/9a31ff2b0596c57d02fa2c36919e86a12b493e49) which will be re-read by GE infra and which also need a ton of config related not just to the data but the expectation suite.

## Example

- Users who want to do GE validation on distributed (and especially ephemeral) infrastructure that would be annoying to keep in sync with a local file system that has the GE directories on it

joshmeek:
## Summary
Closes #2436 
Closes #3057

Enhances the `RunGreatExpectationsCheckpoint` task to allow for a lot more functionality! This also adds a call to the new artifacts API.

Example flow:

```python
from prefect import task, Flow, Parameter
from prefect.tasks.great_expectations import RunGreatExpectationsCheckpoint

checkpoint_task = RunGreatExpectationsCheckpoint()


@task
def get_batch_kwargs(datasource_name, dataset):
    dataset = ge.read_csv(dataset)
    return {"dataset": dataset, "datasource": datasource_name}


with Flow("ge_test") as flow:
    datasource_name = Parameter("datasource_name", default="data__dir")
    dataset = Parameter("dataset", default="data/yellow_tripdata_sample_2019-01.csv")
    batch_kwargs = get_batch_kwargs(datasource_name, dataset)

    expectation_suite_name = Parameter(
        "expectation_suite_name", default="yellow_tripdata_sample_2019-01.warning"
    )
    checkpoint_name = Parameter("checkpoint_name", default="action_list_operator")
    checkpoint_task(
        checkpoint_name=checkpoint_name,
        batch_kwargs=batch_kwargs,
        expectation_suite_name=expectation_suite_name,
    )

flow.register(project_name="Demo")
```

Screenshot of artifact in UI (click to zoom):

image

## Changes
Adds a bunch of new kwargs to the task to align more with in-memory great expectations validations. Task signature now looks like the following:

```
        - checkpoint_name (str, optional): the name of the checkpoint; should match the filename of
            the checkpoint without .py
        - context (DataContext, optional): an in-memory GE DataContext object. e.g.
            `ge.data_context.DataContext()` If not provided then `context_root_dir` will be used to
            look for one.
        - assets_to_validate (list, optional): A list of assets to validate when running the
            validation operator. If not provided then `batch_kwargs` and `expectation_suite_name`
            will be used if context is provided. Also, if not provided and
            `get_checkpoint_from_context` is True then the assets will be loaded from that context.
        - batch_kwargs (dict, optional): a dictionary of batch kwargs to be used when validating
            assets.
        - expectation_suite_name (str, optional): the name of an expectation suite to be used when
            validating assets.
        - get_checkpoint_from_context (bool, optional): get the checkpoint from context. Defaults to
            `False`
        - context_root_dir (str, optional): the absolute or relative path to the directory holding
            your `great_expectations.yml`
        - runtime_environment (dict, optional): a dictionary of great expectation config key-value
            pairs to overwrite your config in `great_expectations.yml`
        - run_name (str, optional): the name of this  Great Expectation validation run; defaults to
            the task slug
        - run_info_at_end (bool, optional): add run info to the end of the artifact generated by this
            task. Defaults to `True`.
        - disable_markdown_artifact (bool, optional): toggle the posting of a markdown artifact from
            this tasks. Defaults to `False`.
```

## Checklist

This PR:


madkinsz:
Whoa sweet!

src/prefect/tasks/great_expectations/checkpoints.py
```diff
@@ -10,108 +10,247 @@
...
+            task. Defaults to `True`.
+        - disable_markdown_artifact (bool, optional): toggle the posting of a markdown artifact from
+            this tasks. Defaults to `False`.
         - **kwargs (dict, optional): additional keyword arguments to pass to the Task constructor
     """
 
     def __init__(
         self,
         checkpoint_name: str = None,
+        context: "DataContext" = None,  # noqa
```
madkinsz, line 104: Why noqa?

src/prefect/tasks/great_expectations/checkpoints.py
```diff
@@ -10,108 +10,247 @@
...
+        if not context:
+            context = ge.DataContext(
+                context_root_dir=context_root_dir,
+                runtime_environment=runtime_environment,
+            )
+
+        # Retrieve the assets to validate if not already provided
+        if not assets_to_validate:
+            assets_to_validate = []
+            if get_checkpoint_from_context:
```
madkinsz, line 211: Can you add a comment to this if-else explaining the two different actions?

src/prefect/tasks/great_expectations/checkpoints.py
madkinsz: Isn't this just `"ge.DataContext"` ?

src/prefect/tasks/great_expectations/checkpoints.py
```diff
@@ -10,108 +10,249 @@
...
-            checkpoint["validation_operator_name"],
-            assets_to_validate=batches_to_validate,
-            run_id={"run_name": prefect.context.get("task_slug")},
+            checkpoint_name or ge_checkpoint["validation_operator_name"],
+            assets_to_validate=assets_to_validate,
+            run_id={"run_name": run_name or prefect.context.get("task_slug")},
         )
 
         if results.success is False:
             raise signals.VALIDATIONFAIL(result=results)
```
cicdw, line 235: This is the only place we use this signal / state; I do remember the original motivation but now I'm inclined to think we should enter a standard `Failed` state instead

cicdw (changes requested):
I think we should use `signals.FAIL` instead of `VALIDATIONFAIL` (not a change you made in the PR, but it reminded me)

cicdw (approved):

Issue #3640, opened 2020-11-10
Env Vars in Docker Storage cannot be used in the build process


zhen0:
## Description

From Community Slack: https://prefect-community.slack.com/archives/C014Z8DPDSR/p1605012711434000
env_vars in the Docker Storage, are variables that are added to the end of a Dockerfile. Since the ENV command is added to the end of the file, you cannot actually pass any variables that are used in the build process itself.

## Expected Behavior

variables can be used in the build or we better explain in the docs that they're added to the end of the dockerfile.

cicdw:
👍 in this instance, we can easily rearrange the order of the internally-generated `Dockerfile` no problem, but in general users who need control of the `Dockerfile` commands should write their own and use the `dockerfile` kwarg.

PR #3755, opened 2020-12-02
Move docker storage env declaration up

madkinsz:
## Summary

Moves docker storage environment variables up in the generated Dockerfile so they can be used downstream

## Importance

Allows the environment variables to be used in the commands
Closes https://github.com/PrefectHQ/prefect/issues/3640

## Checklist

This PR:

### DIFF
--- a/changes/pr3755.yaml
+++ b/changes/pr3755.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Allow Docker storage environment vars to be used in commands - [#3755](https://github.com/PrefectHQ/prefect/pull/3755)"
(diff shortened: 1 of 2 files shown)


madkinsz:
This is kind of blocked while I wait for a way to test this effectively although we could merge it without automated testing.

joshmeek:
@madkinsz Also worth noting that your branch needs to be updated with master, the CI pip loop is happening which we fixed the other day 👍

madkinsz:
@jcrist I don't think there's a great way to test this before we can actually run docker tests in CI. Thoughts?

jcrist:
I think that's fine. The code puts the env vars in the file before the commands by inspection, any test now would be just testing the string output. You could add a test that the generated env strings are before the generated cmd strings, but I'm not sure that's necessary (and those kind of tests are generally brittle). Happy to merge with or without a string-munging test - up to you.

jcrist (approved):

PR #3766, opened 2020-12-05
Make RunGreatExpectationsCheckpoint task accept different parameters

spbail:
## Summary
Some tweaks to the RunGreatExpectationsCheckpoint task (including renaming...) to clarify the types of parameters that can be used to run validation. See discussion for context: https://github.com/PrefectHQ/prefect/discussions/3558

**This is untested, I just wanted to propose the change and discuss!**

## Changes

I think there was some confusion around the use of "checkpoint"! A checkpoint is basically just a pre-configured "bundle" of batch_kwargs + suite name, so it can be used instead of those pairs to run validation. It looks like it was confused with a "validation_operator", which is basically just a pointer to some actions that happen after validation (by default: store the results, render data docs). I made some tweaks to reflect this better in the code:

- Renamed the task to RunGreatExpectationsValidation, since it doesn't just run checkpoints
- Clarified the different possible parameters in the docstring
- Dropped the get_checkpoint_from_context parameter since it's not needed
- Added the validation_operator parameter with a sensible default "action_list_parameter" (that's the default that's setup when running great_expectations init)
- Added a check to ensure that the parameters are mutually exclusive
- Replaced "checkpoint" with "validation_operator" where appropriate

TODO: I just commented out the current test for checkpoint_name, would be awesome if someone could help write a sensible test there.

## Importance
The current version of the task doesn't clearly separate the different ways to run validation and mixes up checkpoints and validation operators, which doesn't match the Great Expectations terminology. I also think the renaming makes sense since it runs more than just checkpoints.

## Checklist

This PR:

### DIFF
--- a/examples/task_library/great_expectations/great_expectations_flow.py
+++ b/examples/task_library/great_expectations/great_expectations_flow.py
@@ -1,7 +1,7 @@
 from prefect import Flow, Parameter
-from prefect.tasks.great_expectations import RunGreatExpectationsCheckpoint
+from prefect.tasks.great_expectations import RunGreatExpectationsValidation
 
-ge_task = RunGreatExpectationsCheckpoint()
+ge_task = RunGreatExpectationsValidation()
 
 with Flow("great expectations example flow") as flow:
     checkpoint_name = Parameter("checkpoint_name")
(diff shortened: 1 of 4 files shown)


joshmeek:
Thanks for this @spbail! I'll give it a test today and see how everything works out 🙂

joshmeek:
@spbail I was having issues pushing to the remote branch so I opened my changes in a new PR #3776 🙂

PR #3780, opened 2020-12-08
Update path to config file in docs

jackdesert:
## Summary

Minor documentation change

## Changes

Shows the correct path to the config file

## Importance

Important because if your config file is in the wrong location, the server won't pick it up. And you might have to clear data from your browser before if will pick it up from the correct location in the future if you do it wrong the first time.

## Checklist

This PR:

(No checkboxes needed for this documentation change)


jcrist (approved):
Thanks, nice catch.

Issue #2468, opened 2020-05-01
Use `async` code in agents instead of threads


jcrist:
Currently the agent code is using threads (mostly `concurrent.futures`). While fine, this makes a couple things tricky:

- It's hard to test the agents robustly, as they're each designed to run as the main process. Many of the tests rely on heavy monkeypatching, which decreases test effectiveness. The threaded nature of the code also makes it hard to test specific behaviors, as any code masquerading as the backend api needs to run in a background thread rather than in the same event loop.
- It's hard to add new behaviors, since the main thread is blocking. For example, in #2406 I added an optional HTTP server for use with health checks - this had to be started in its own backing thread (with complicated thread management logic), rather than just-one-more-thing running in an event loop.

I'd argue that the above problems would are easier to solve if we make use of python's async features. I'm especially interested in making it easier to add background behaviors that don't match the standard `deploy_flow` model. Currently there's some non-ideal workarounds for these cases - for example, the docker agent currently needs to start a temporary process per flow to handle rerouting logs, something that could be done easier in-process if running asynchronously.

Since the backing infrastructure that the agent manages is inherently asynchronous (e.g. docker, k8s, fargate, etc..), it feels a bit more natural to me to make the `deploy_flow` calls non blocking async methods. Several other Python projects that abstract over and manage infrastructure take the same stance and are written in an asynchronous manner. A few examples:

- [JupyterHub Spawners](https://jupyterhub.readthedocs.io/en/stable/reference/spawners.html#spawner-start)
- [Dask-Gateway backends](https://github.com/dask/dask-gateway/blob/master/dask-gateway-server/dask_gateway_server/backends/base.py)
- [Dask ClusterManagers](https://github.com/dask/distributed/blob/master/distributed/deploy/spec.py)

PR #3783, opened 2020-12-08
Log pod events for pending pods created by Kubernetes Agent

joshmeek:
## Summary

In an increasing effort to remove any ambiguity around flow run deployments (#3747) this PR adds a new step to the Kubernetes Agent's job management process to log pod events for pending pods. Resulting information looks like this in flow run logs:

image

In some cases when users see that their flow runs are stuck in a submitted state it can be confusing. Often times this can occur due to things like node scaling in k8s. This is an important set of information that can be elevated to users.

## Changes
Adds a check to the Kubernetes Agent's job management process that looks for pods in a pending state and sends flow run logs for new events.

## Importance
This is a huge potential black box that we can surface information about in order to improve visibility into the health of flow runs!

## Checklist

This PR:


jcrist (commented):
I like the idea of using the events api here - this surfaces a bunch of useful info on pod/job operations.

Polling for every pod (and re-listing on each iteration) may be a large overhead. It also will duplicate all events on startup. A more efficient way might be to:

- On startup, list all events for all pods managed by the agent. This should be doable with a label selector. Grab the most recent `resource_version` and stash it.
- On each iteration, relist events for all prefect pods starting at the most recent `resource_version`. We'll need to handle resource version expiration (do a relist and start over), but otherwise we should only get new events. Log each new event for each pod. We could batch them as you're doing here, or log them as individual rows (I think individual rows are clearer to read, but may be noisy).

This should be more efficient in the number of api calls to the k8s api, and keeps are local state smaller (only hold onto `resource_version`). I think this should work.

At some point if we move to using async code in agents instead of threads (https://github.com/PrefectHQ/prefect/issues/2468) it'd be easier to use a watch on events/pods to track these things in a background task. We could do this with threads as well, but it's a bit messier.

jcrist:
When we merge this, we'll also need to add events to the RBAC permissions given to the agent (both in the docs, and in the generated manifests).

joshmeek:
> This should be doable with a label selector.

We can't seem to do this, the events don't appear to have labels attached to them. That is why I went with the field selector for the implementation. All of the pod and job events show this for labels:

```
Labels:           <none>
```

> We'll need to handle resource version expiration (do a relist and start over)

What do you mean by resource version expiration?

> We could batch them as you're doing here, or log them as individual rows

I'm starting to feel individual events would be a better option 👍 

> When we merge this, we'll also need to add events to the RBAC permissions given to the agent

Ah right, this would _technically_ be breaking in that case. Should we wait until 0.14.0?

jcrist:
Discussed offline - what's here is probably fine, especially since we're only tracking pending pods. We're going to switch to a different strategy for de-duplicating events, but this is pretty close as is.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -183,6 +188,60 @@ def manage_jobs(self) -> None:
...
+                                        break
+
+                                    self.job_pod_event_timestamps[job_name][
+                                        pod_name
+                                    ] = event.last_timestamp
+
+                                    pod_event_logs = [
+                                        f"Event: '{event.reason}' on pod '{pod_name}'"
+                                    ]
+                                    pod_event_logs.append(f"\tMessage: {event.message}")
```
jcrist, line 230: No need to build a list and join below, could build it all at once:

```suggestion
                                    log_msg = f"Event: {event.reason!r} on pod {pod_name!r}\n\tMessage: {event.message}"
```

src/prefect/agent/kubernetes/agent.py
```diff
@@ -183,6 +188,60 @@ def manage_jobs(self) -> None:
...
+
+                                    # Send pod failure information to flow run logs
+                                    self.client.write_run_logs(
+                                        [
+                                            dict(
+                                                flow_run_id=flow_run_id,
+                                                name=self.name,
+                                                message="\n".join(pod_event_logs),
```
jcrist, line 238: If you take my above suggestion

```suggestion
                                                message=log_msg,
```

src/prefect/agent/kubernetes/agent.py
```diff
@@ -183,6 +188,60 @@ def manage_jobs(self) -> None:
...
+                                    pod_events.items, key=lambda x: x.last_timestamp
+                                ):
+                                    # Break out of loop if there are no new events
+                                    if (
+                                        event.last_timestamp
+                                        < self.job_pod_event_timestamps[job_name][
+                                            pod_name
+                                        ]
+                                    ):
+                                        break
```
jcrist, line 221: I think this check is incorrect - if we get a list of `[old_event, old_event, new_event]`, the check here will see the old event first and break out, dropping the chance to display the new event. I think you want to `continue` instead to skip displaying the event.

```suggestion
                                    # Skip old events
                                    if (
                                        event.last_timestamp
                                        < self.job_pod_event_timestamps[job_name][
                                            pod_name
                                        ]
                                    ):
                                        continue
```

joshmeek: The loop sorts based on the timestamp and it will have the newest event first. The break prevents us from having to loop over everything in that case

jcrist (approved):
LGTM (if tests pass)

PR #3785, opened 2020-12-08
Replace characters in API doc build that cause rendering issues

joshmeek:
Currently in some of the API docs when rendering defaults with the class's `repr` it will cause it to not render due to the use of `<` and `>`. 

Current:
image

With fix:
image

### DIFF
--- a/docs/generate_docs.py
+++ b/docs/generate_docs.py
@@ -15,6 +15,7 @@ Each entry in `OUTLINE` is a dictionary with the following key/value pairs:
 On a development installation of Prefect, run `python generate_docs.py` from inside the `docs/` folder.
 """
 import builtins
+import html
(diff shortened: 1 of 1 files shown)


docs/generate_docs.py
```diff
@@ -290,6 +290,9 @@ def get_call_signature(obj):
                 default = f'"{default}"'
             else:
                 default = repr(default)
+
+            # Replace from repr because it can cause HTML errors in rendering
+            default = default.replace("<", "&lt;").replace(">", "&gt;")
```
jcrist, line 295: You should use [`html.escape`](https://docs.python.org/3/library/html.html#html.escape) instead.

jcrist (approved):

PR #3796, opened 2020-12-09
Move storage classes to `prefect.storage`

jcrist:
This moves the `Storage` classes from `prefect.environments.storage` to `prefect.storage`. Since environments are being deprecated, and `Storage` objects are so commonly used, it makes sense to bring them up a level.

Thin subclasses of the existing types have been kept at the old import paths. These subclasses will raise a warning on initialization, but should otherwise function the same.

All references in the docs, examples, etc... have been updated to use the new import paths.


joshmeek (approved):
🚀

PR #3800, opened 2020-12-10
Remove default labels on storage types and local agent

joshmeek:
## Summary
`0.14.0`: Removes the default labels on all storage options that had them and on the local agent. This does not, however, remove the hostname label on the local storage and local agent.

## Changes
- Removed setting of default labels on storage objects and the local agent
- Removed `storage_labels` boolean kwarg from local agent
- Removed `--storage-labels` option from agent `start` CLI command

## Importance
This has been a slightly confusing safety net that is no longer necessary now that all storage options can be used with all agents.

## Checklist

This PR:

### DIFF
--- a/changes/pr3800.yaml
+++ b/changes/pr3800.yaml
@@ -1,6 +1,6 @@
 breaking:
   - "Removed setting of default labels on storage objects and the local agent - [#3800](https://github.com/PrefectHQ/prefect/pull/3800)"
 
-deprecated:
+deprecation:
   - "Deprecated use of `storage_labels` boolean kwarg on local agent - [#3800](https://github.com/PrefectHQ/prefect/pull/3800)"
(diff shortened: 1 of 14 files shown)


jcrist (dismissed):
This seems fine to me.

Do we want to do this as a full on breaking change though (as done here), or just a breaking change in what the defaults are? As written, if users are using the removed kwargs/flags, this will lead to an error if they re-register using prefect 0.14.0. We could do this by keeping the flags (with a deprecation warning) and just changing the defaults, which might be a little smoother transition. In practice though, the upgrade strategy here is pretty straightforward (drop the flag/pass in labels explicitly, depending on what the user's doing), so I'm also fine doing this as a full breaking change. Up to you I guess.

joshmeek:
@jcrist Yeah I considered doing that and since you point it out I will make that change / deprecation warnings 👍

jcrist (approved):
LGTM.

PR #3804, opened 2020-12-10
Refactor dependencies included in Docker images

joshmeek:
## Summary
With `0.14.0` approaching we have the opportunity to better package dependencies in our base image. This PR makes various updates to some CI steps (including tests and image builds). It also updates places where the previous `all_extras` image was being used.

**Note**: We should cut a release candidate of `0.14.0` to ensure that the image building jobs function properly.

## Changes
- Adds CI job for only testing the task library
- Ignores testing the task library in the 3.6-9 jobs
- Removes building the `all_extras` Docker image
- Adds a build of the `core` image that only contains the prefect core base dependencies
- Adds all orchestration extras to the standard `prefect/prefect` image
- Docker storage now installs all orchestration dependencies
- Deployed flows now default to using the `prefect/prefect` image with the orchestration extras installed instead of `all_extras`

## Importance
This gives us a better packaging experience of what we consider important deployment dependencies. Those include packages that service things like storage and agents.

## Checklist

This PR:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,7 @@ extras["all_orchestration_extras"] = [
 # CI extras to control dependencies for tests
 extras["task_library_ci"] = sum(extras.values(), [])
 extras["task_library_ci"] = [
-    r for r in extras["test_ci"] if not r.startswith("dask_cloudprovider")
+    r for r in extras["task_library_ci"] if not r.startswith("dask_cloudprovider")
 ]
 
 extras["base_library_ci"] = extras["all_orchestration_extras"] + extras["test"]
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ extras["task_library_ci"] = [
     r for r in extras["task_library_ci"] if not r.startswith("dask_cloudprovider")
 ]
 
-extras["base_library_ci"] = extras["all_orchestration_extras"] + extras["test"]
+extras["base_library_ci"] = extras["all_orchestration_extras"] + extras["dev"]
 
 cmdclass = {
     "verify_version": VerifyVersionCommand,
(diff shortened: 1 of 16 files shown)


jcrist (changes requested):
One small fix and this looks good to me.

setup.py
```diff
@@ -70,17 +70,31 @@ def run(self):
...
+# Extras for docker image builds to include for orchestration
+extras["all_orchestration_extras"] = [
+    "atlassian-python-api >= 2.0.1",
+    "azure-storage-blob >= 12.1.0, < 13.0",
+    "boto3 >= 1.9, < 2.0",
+    "kubernetes >= 9.0.0a1, <= 11.0.0b2",
+    "google-cloud-storage >= 1.13, < 2.0",
+    "python-gitlab >= 2.5.0, < 3.0",
+    "PyGithub >= 1.51, < 2.0",
+]
```
jcrist, line 88: Instead of duplicating these requirements here, we should build them programmatically out of a set of extras we want to include (similar to how `task_library_ci` is done). Would merge deps for each needed extra e.g. [`aws`, `github`, etc...].

jcrist (changes requested):
Oops, for some reason the comment didn't take last time.

setup.py
joshmeek: I considered that but we have extra dependencies in there for those names. e.g. azure has three, gcp/google has two, kubernetes has two, etc..

jcrist: Ah, hmmm. My main concern is different versions listed for each - perhaps assign the ones listed here as variables and then re-use them in both locations? I just don't want to have to bump the boto3 version in two spots.

jcrist (approved):

PR #3829, opened 2020-12-14
Move creation of markdown artifact in RunGreatExpectationsValidation up so it is created even when validation fails

cwerner:
## Summary

Hi. I wanted to receive artifact reports for my GreatExpectations validation even if the validation fails. 

## Changes
Solution is to simply move the creation of markdown artifact in RunGreatExpectationsValidation up so it is created even when success check fails and thus returns with state FAIL.

## Importance
Creates the artifact even if validation fails (which I think is important).

## Checklist

This PR:

### DIFF
--- a/changes/pr3829.yaml
+++ b/changes/pr3829.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
+#   - fix
+#   - deprecation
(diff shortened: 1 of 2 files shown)


joshmeek (approved):
Thanks @cwerner!

cwerner:
No worries. However, seems I have the wrong PR number mentioned in the changes yml file. Not sure if this causes a problem? The filename is correct though...

joshmeek:
@cwerner As not a problem, we'll fix it before we make the release 🙂

PR #3833, opened 2020-12-14
Properly pass labels to register call in CLI command

joshmeek:
Updates the `prefect register flow` command to pass labels to the register call. Currently calling w/o any run config or environment would error because the UniversalRun isn't added until later. Simply passing in the labels this way fixes that and achieves the same behavior as before. (also updates a test)

No changes entry necessary since this was brought about due to a change in master


jcrist (approved):
Nice catch.

PR #3834, opened 2020-12-14
Load run config from backend on execute

joshmeek:
In the case where there is no run config on a flow object and the flow is stored using file-based storage then the execute command will previously error. This is due to the fact that the run config needs to be loaded off of the backend metadata (just like storage is) in order to retrieve the `UniversalRun` default.

No changes entry necessary since this was brought about due to a change in master

### DIFF
--- a/src/prefect/cli/execute.py
+++ b/src/prefect/cli/execute.py
@@ -39,7 +39,7 @@ def flow_run():
     query = {
         "query": {
             with_args("flow_run", {"where": {"id": {"_eq": flow_run_id}}}): {
-                "flow": {"name": True, "storage": True},
+                "flow": {"name": True, "storage": True, "run_config": True},
                 "version": True,
             }
         }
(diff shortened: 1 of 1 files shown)


src/prefect/cli/execute.py
```diff
@@ -66,6 +66,12 @@ def flow_run():
...
             flow = storage.get_flow(storage.flows[flow_data.name])
 
+        # Load run config from backend if not present on flow object
+        if flow.run_config is None and flow_data.run_config:
+            flow.run_config = prefect.serialization.run_config.RunConfigSchema().load(
+                flow_data.run_config
+            )
+
         with prefect.context(secrets=secrets):
             if getattr(flow, "run_config", None) is not None:
```
jcrist, line 76: Since the only reason we need the `run_config` in execution is to determine if the flow is using a legacy `environment` based configuration, we should be able to drop the load (and mutation) above and just modify the boolean condition here. This will also play nicer with future changes where the run-config info can be overridden in the backend (if we let legacy flows be overridden). Something like:

```suggestion
            if flow_data.run_config is not None:
```

jcrist (approved):

PR #3837, opened 2020-12-15
Fetch flow parameters by calling parameters()

bbtfr:
## Summary

Call `parameters()` to fetch flow parameters.

## Importance

It's better to call `parameters()` to fetch parameters of a flow, so user can inherit Flow class and rewrite parameters method.

For me, my flow contains a huge number of parameters, since Parameter is a Task, it takes 20s ~ 30s to run these tasks to fetch parameters.
So I create a new Parameter class (just a namedtuple) and a subclass of Flow, rewrite the `parameters` method, a subclass of FunctionTask to fetch parameters directly from `context['parameters']`.
It works fine, except I can't register to the server, because the flow serialize with `get_parameters` and it fetch parameters without calling `parameters()`

## Checklist

This PR:

### DIFF
--- a/src/prefect/serialization/flow.py
+++ b/src/prefect/serialization/flow.py
@@ -18,9 +18,7 @@ from prefect.utilities.serialization import (
 
 def get_parameters(obj: prefect.Flow, context: dict) -> List:
     if isinstance(obj, prefect.Flow):
-        params = {
-            p for p in obj.tasks if isinstance(p, prefect.core.parameter.Parameter)
-        }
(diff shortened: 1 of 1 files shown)


jcrist (approved):
Thanks @bbtfr, this looks good to me.

Regarding your larger issue though, we'd like Prefect itself to be able to handle your workload. Would you mind opening an issue with a reproducible example of something you'd like to be more performant, and what your expected behavior is? Thanks!

PR #3838, opened 2020-12-15
Remove Result handlers and associated stale code

cicdw:
## Summary

As a part of 0.14.0, we are removing all result handler related code; the only sense in which this will be a breaking change for users is that the `result_handler` kwarg will cease working on Tasks, Flows and Results, but otherwise functionality should be preserved.

## Changes

Removes:
- Result Handlers
- Safe results
- No Results

and associated code paths / tests.

## Importance

Removing stale / deprecated code.

## Checklist

This PR:


tests/serialization/test_states.py
```diff
@@ -338,22 +298,70 @@ def test_new_result_with_location_deserializes_correctly(self):
...
+                        "dir": "/src",
+                        "type": "LocalResultHandler",
+                        "__version__": "0.11.5+59.gdba764390",
+                    },
+                }
+            },
+        },
+    ],
+)
+def test_can_deserialize_old_no_result(old_json):
```
cicdw, line 363: This test is a new test that actually confirms we can deserialize old state payloads still

src/prefect/serialization/result.py
```diff
@@ -152,9 +116,13 @@ class StateResultSchema(OneOfSchema):
         "PrefectResult": PrefectResultSchema,
         "S3Result": S3ResultSchema,
         "SecretResult": SecretResultSchema,
-        "ResultHandlerResult": ResultHandlerResultSchema,
     }
 
+    def _load(self, data, *args, **kwargs):
+        if data.get("type") not in self.type_schemas:
+            data["type"] = "CustomResult"
```
cicdw, line 123: these are the two key lines for ensuring we can deserialize old-style results / result handlers

jcrist (approved):
🚀🚀

PR #3861, opened 2020-12-16
Fix window referencesto prevent errors being thrown during build

znicholasbrown:
## Summary

Fixes references to `window` in the Node context

## Changes

^

## Importance

Streamlines build output

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #3862, opened 2020-12-16
Move 0.10.7 API docs to legacy docs

joshmeek:
With `0.14.0` release on the horizon the `0.10.7` docs have been added to the legacy API documentation (https://github.com/PrefectHQ/legacy-api-docs/pull/9) and can be removed from this repo.

https://docs-legacy.prefect.io/api/0.10.7/

This branched off of https://github.com/PrefectHQ/prefect/pull/3861 and can wait to be merged until that is good to go 👍


jcrist (approved):
Thanks Josh! Yeah, this seems necessary. Sometimes the build completes, sometimes it just sits there spinning silently forever. Per Nicholas people run vuepress on sites much larger than ours, so perhaps there's something dumb in our custom docs setup? (another push towards not rolling our own docs tooling :)). If this fixes things I'm 👍 for now.

PR #3875, opened 2020-12-18
Support GCS versions < 1.31.0

### LINKED ISSUE #3870: Google Cloud Storage <= 1.31 no longer supported
#3741 changed the GCS result to use a function that was only released a couple of months ago in 1.31.0 (see https://github.com/googleapis/python-storage/commits/master?after=9d5bb5622f17a0143fa6670406596e53d0068286+34&branch=master).

In my opinion it's a bit too aggressive to require a very new version of a low-level requirement like this, but regardless the listed requirement should be bumped correspondingly (right now it's at 1.13.0 https://github.com/PrefectHQ/prefect/blob/b133b44cffab79735a8fe86e16c81d3e6c74d1ab/setup.py#L35 which is from October 2018). Would a `try: ... except AttributeError:` block be acceptable as a way to maintain backwards compatibility?

jcrist:
Use `download_as_string` if `download_as_bytes` isn't available.

Fixes #3870.

This PR:

### DIFF
--- a/changes/issue3870.yaml
+++ b/changes/issue3870.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Add support for `google-cloud-storage` < 1.31.0 - [#3875](https://github.com/PrefectHQ/prefect/pull/3875)"
--- a/src/prefect/engine/results/gcs_result.py
+++ b/src/prefect/engine/results/gcs_result.py
@@ -94,7 +94,13 @@ class GCSResult(Result):
 
         try:
             self.logger.debug("Starting to download result from {}...".format(location))
-            serialized_value = self.gcs_bucket.blob(location).download_as_bytes()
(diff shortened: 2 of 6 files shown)


bnaul:
Thanks @jcrist! Fyi I noticed that there's a call in `src/prefect/storage/gcs.py` that was missed in the original PR (so it still works but gives the deprecation warning), dunno if you want to update that as well for consistency.

jcrist:
Thanks, I only grepped for `download_as_bytes`, should be resolved.

madkinsz (approved):
LGTM -- is there no longer test coverage of `download_as_string` ?

jcrist:
> is there no longer test coverage of download_as_string

No. We could add a test for this, but given it'd have to be used everywhere google stuff was used (which would be hard to enforce) I'm fine skipping it for now.

PR #3892, opened 2020-12-23
Fix `StartFlowRun` idempotency_key

jcrist:
- Adds `idempotency_key` as a kwarg to `StartFlowRun` constructor for
  parity with its use in `run`.
- Changes the default `idempotency_key` to use `task_run_id` instead of
  `flow_run_id`. The previous default would lead to only a single flow
  run being created if multiple `StartFlowRun` tasks were created for
  the same flow without relying on mapping.

This PR:

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -30,10 +30,6 @@ class StartFlowRun(Task):
             state as the state of this task.  Defaults to `False`.
         - new_flow_context (dict, optional): the optional run context for the new flow run
         - run_name (str, optional): name to be set for the flow run
-        - idempotency_key (str, optional): a unique idempotency key for scheduling the
-            flow run. Duplicate flow runs with the same idempotency key will only create
-            a single flow run. This is useful for ensuring that only one run is created
-            if this task is retried. If not provided, defaults to the active `task_run_id`.
         - scheduled_start_time (datetime, optional): the time to schedule the execution
             for; if not provided, defaults to now
         - **kwargs (dict, optional): additional keyword arguments to pass to the Task constructor
@@ -47,7 +43,6 @@ class StartFlowRun(Task):
         wait: bool = False,
         new_flow_context: dict = None,
         run_name: str = None,
-        idempotency_key: str = None,
         scheduled_start_time: datetime.datetime = None,
         **kwargs: Any,
     ):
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -43,6 +47,7 @@ def __init__(
         wait: bool = False,
         new_flow_context: dict = None,
         run_name: str = None,
+        idempotency_key: str = None,
```
cicdw, line 50: I don't think I agree with this change - if utilized, it means that one and only one flow run will ever be created across all runs of the parent flow, which can always be achieved with a hardcoded string, but this feels like it will invite confusion if people use it

src/prefect/tasks/prefect/flow_run.py
```diff
@@ -147,20 +155,15 @@ def run(
...
 
-        idem_key = None
-        if context.get("flow_run_id"):
-            map_index = context.get("map_index")
-            default = context.get("flow_run_id") + (
-                f"-{map_index}" if map_index else ""
-            )
-            idem_key = idempotency_key or default
+        if idempotency_key is None:
+            idempotency_key = context.get("task_run_id", None)
```
cicdw, line 159: 👏

cicdw (changes requested):
Love the task run ID idea; request to remove the key from init though

src/prefect/tasks/prefect/flow_run.py
jcrist: I don't follow? Parameters can be passed to tasks two ways:

- As static parameters to the constructor
- As dynamic parameters to `__call__`/`bind` (usually from tasks)

I don't see a good reason to restrict a parameter to one or the other, and frequently find users confused when a parameter is only available in the constructor (or as part of `.run`). Restricting to a `run` parameter doesn't stop a user from incorrectly using `idempotency_key`, and the default behavior remains `None` as before.

jcrist: Oh wait, I see what you're saying. Sure, I can remove it.

cicdw: I think there's a good reason to restrict whenever a static value for something is misleading or not particularly useful; if I built a flow and used a static key here, then what's the point of the task?  It will create a single flow run once, and never create a second one.  Users will almost certainly ask "Why is my task not creating new runs"?  For users that (for some reason) actually _do_ want that functionality, they can opt-in via a hardcoded static string in the run method, but at least in that case they'll know what they're getting themselves into.

cicdw (approved):

PR #3893, opened 2020-12-23
Add container labels to docker agent

### LINKED ISSUE #3819: Use custom container names when useing docker storage with docker agent
Hi,

I trying to figure out how to set the name for the container that docker agent spans when the flow is deployed from the cloud.

We are using the docker agent running on the local network together with the Prefect Could as an orchestrator. There are multiple jobs that run scheduled to be run by one of the containerized docker agents in a separated container. Our problem with the approach is that every time a new container with the same flow is started it gets an autogenerated name by docker which makes monitoring hard. It would be great if we could use the image name or something similar for the containers.

jcrist:
Adds the following labels to all flows deployed by the docker agent.
This is useful tools that monitor docker deployments by label.

- `io.prefect.flow-name`: the full flow name
- `io.prefect.flow-id`: the full flow id
- `io.prefect.flow-run-id`: the full flow run id

See https://docs.docker.com/config/labels-custom-metadata/ for more
information on docker labels.

Fixes #3819.

This PR:

### DIFF
--- a/changes/pr3893.yaml
+++ b/changes/pr3893.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Add labels to all containers managed by the docker agent - [#3893](https://github.com/PrefectHQ/prefect/pull/3893)"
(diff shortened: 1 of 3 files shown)


cicdw (approved):
nice

Issue #3660, opened 2020-11-13
Add command: prefect server stop


joelluijmes:
## Current behavior

We can locally start server with `prefect server start`, docs: https://docs.prefect.io/api/latest/cli/server.html
But unless I'm mistaken, I miss the reverse, the stop command.

## Proposed behavior

Add stop command which stops prefect.

rmax:
I see that the CLI uses [tempfile.gettempdir()](https://github.com/PrefectHQ/prefect/blob/0e9fa02c7acac586610c9894bec82e90fdf57c8b/src/prefect/cli/server.py#L266) as temporary location of the modified `docker-compose.yml`.

That location is consistent between runs (at least in OSX) so it's possible to reference the file for a `docker-compose down` command. Although, it might be good to use a subidrectory rather than the root temp dir to avoid potential side effects.

By using the `--file` option I can stop the containers. Notice the containers are prefixed by 't_' which comes from the 'T' directory name. This could be more friendly by using `--project prefect` or similar in both `up` and `down` commands to get the container names more descriptive.

```
❯ docker ps --format '{{.Names}}\t{{.Image}}'
t_ui_1  prefecthq/ui:core-0.13.14
t_towel_1       prefecthq/server:core-0.13.14
t_apollo_1      prefecthq/apollo:core-0.13.14
t_graphql_1     prefecthq/server:core-0.13.14
t_hasura_1      hasura/graphql-engine:v1.3.0
t_postgres_1    postgres:11
```

Here is how I stop the containers:

```
❯ python -c 'import tempfile; print(tempfile.gettempdir())'
/var/folders/nv/7pj04vn56wlf1470pj2_3pth0000gn/T                                    

❯ ls /var/folders/nv/7pj04vn56wlf1470pj2_3pth0000gn/T | head
9B8F6828-49B5-46D9-B42F-085082C0528A
AppTranslocation
AudioComponentRegistrar
DMD01241D7D-D786-4D42-89D1-519D259E13F0
DMD01AD8FA0-ADE9-4D14-8801-3298810A7185
DMD02472AF6-CC86-416A-8EF5-CAA7206C2AF4
DMD02C59915-9D0C-45AA-BDC8-1C058BC9B822
DMD0418B6C5-C841-46C3-8C33-107D40CCFB53
DMD048794A7-3598-462C-9B98-F74B7E44D3CA                                             

[...]

Issue #2348, opened 2020-04-17
prefect server stop


tekumara:
## Current behavior

There doesn't appear to be a way to stop the prefect server and all its containers.

## Proposed behavior

`prefect server stop` will stop all containers.

cicdw:
Hi @tekumara - you can gracefully stop all containers with a `SIGINT` signal (e.g., CTRL+C).

Because `prefect server start` is a blocking command with a graceful shutdown mode, I don't think there's a benefit to adding a new CLI endpoint for stopping the containers independently of this.

tekumara:
I've hung up on the original terminal session that started prefect, and so now don't know how to shut the server containers down.

znicholasbrown:
Hi @tekumara - since all of the services run with Docker, you can run `docker ps` to get a list of your running containers and then `docker kill ${container-id}` for each of the containers you'd like to shut down. If you've used something like `nohup` to redirect output, you can instead use `ps aux | grep prefect` to get the PID of the prefect process and run `kill ${PID}`. I hope that helps!

tekumara:
The prefect process doesn't exist anymore. The hang up occurred when I upgraded macOS. When my machine rebooted, Docker Desktop started the containers again.

I killed the 6 containers using `docker kill`. The `prefect-server` docker network was left behind. So I reran `prefect server start` and then hit CTRL-C which removed the network. A little complicated but it worked, thanks!

chrisgervang:
Manually doing this work to clean up a prefect session isn't ideal since I switch between projects. I use `prefect server start >> server.log 2>&1 &` to run in the background. Can prefect provide a clean-up command as proposed?

PR #3899, opened 2020-12-28
Add server stop command and detached option to start command

joshmeek:
## Summary
A recurring question is about how to stop the docker-compose server. This command adds that option with a new command that shuts down the `prefect-server` network and all associated containers. It also adds a detached option to the start command for running the server in the background using docker's detached mode.

## Changes
`prefect server stop` will shut down the running prefect server instance
`prefect server start --detach` will start the server in the background

## Checklist

This PR:


src/prefect/cli/server.py
```diff
@@ -360,6 +372,8 @@ def start(
                     time.sleep(0.5)
                     pass
             while True:
+                if detach:
+                    return
```
jcrist, line 376:
```suggestion
            if detach:
                return
            while True:
```

src/prefect/cli/server.py
```diff
@@ -400,6 +414,32 @@ def ascii_welcome(ui_port="8080"):
...
+    client = docker.APIClient()
+
+    network_id = client.networks(names=["prefect-server"])[0].get("Id")
+
+    if not network_id:
+        click.echo("No running Prefect Server found")
+        return
+
+    click.echo("Stopping Prefect Server containers and network")
+    for container in client.inspect_network(network_id).get("Containers").items():
```
jcrist, line 435: Is this a `dict.items` call? If so, it looks like you could use `keys()` instead of indexing out `container[0]` below.

src/prefect/cli/server.py
```diff
@@ -400,6 +414,36 @@ def ascii_welcome(ui_port="8080"):
...
+
+    networks = client.networks(names=["prefect-server"])
+
+    network_id = None
+    if networks:
+        network_id = networks[0].get("Id")
+
+    if not network_id:
+        click.echo("No running Prefect Server found")
+        return
```
jcrist, line 436:
```suggestion
    if not networks:
        click.echo("No running Prefect Server found")
        return
    network_id = networks[0].get("Id")
```

cicdw:
NB: closes / address both https://github.com/PrefectHQ/prefect/issues/3660 and https://github.com/PrefectHQ/prefect/issues/2348

jcrist (approved):

rajashekarmakam:
If I use the prefect server start --detach then the service hangs forever. Any reason ??

madkinsz:
> If I use the prefect server start --detach then the service hangs forever. Any reason ??

Hi! Please open an issue with more details or post in our community slack.

PR #3927, opened 2021-01-06
POC - Flow builder function

madkinsz:
Note: Holding off on tests and documentation until discussion

## Summary

I'd like to improve the flow development experience in a project format which benefits from some additional utilities.

This introduces a `flow_builder` decorator for functions that will initialize a flow named after the wrapped function and enter the context.

Tasks called from within the function are added to the flow
```python
from prefect import task
from prefect.utilities.flows import flow_builder

@task(log_stdout=True)
def foo():
    print("Foobar!")

@flow_builder
def my_flow(flow):
    foo()
```

Flow initializer arguments can be passed
```python
import prefect
from prefect import task
from prefect.utilities.flows import flow_builder

def say_name():
    prefect.context.logger.info(f"Hi, my name is {prefect.context.flow_id!r}")

@flow_builder(name="my-custom-name")
def my_flow(flow):
    say_name()
```

It also introduces a `run_or_register` utility which takes the place of 

```python
if __name__ == "__main__":
     flow.run(...)
     flow.register(...)
```

where users frequently have to comment out one of the calls. This allows you to do

```python
run_or_register(flow, run_kwargs=dict(for_example=True))
```
then

`python my-flow-file.py <run/register>` to run or register it. 

It also supports callables that produce flows so flows can be functional e.g.

```python
def my_flow() -> Flow:
    with Flow("test") as flow:
        pass

run_or_register(my_flow)
```

The `flow_builder` decorator calls this automatically so it doesn't need to be included in every file and args can be passed to it still, e.g.

```python
from prefect.utilities.flows import flow_builder

@flow_builder(
    register_kwargs=dict(project_name="default"),
    run_kwargs=dict(executor=DaskExecutor()),
)
def my_flow(flow):
    pass
```

Flow builder functions can be imported and treated normally as well, this doesn't make use of `run/register_kwargs` though -- it may be nice to have these defaults be stored in the `Flow` object?
```
from my_project import fn_that_uses_flow_builder

flow = fn_that_uses_flow_builder()
flow.register()
```

## Changes

- Adds new `flows` utilities

## Importance

- Defining flows in functions keeps module import times fast
- Improves flow development experience without introducing new CLI
- Provides a future pathway to tracking flows
- Reduces user confusion around Python `__main__` usage

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/flows.py
+++ b/src/prefect/utilities/flows.py
@@ -1,0 +1,157 @@
+import inspect
+import sys
+from functools import wraps
+from typing import Callable, Union, Dict, Any
+
+from toolz import curry
+
+import prefect
+
+# A reusable type definition for flows or a function that creates a flow
+ResolvesToFlow = Union[prefect.Flow, Callable[[], prefect.Flow]]
+
+
+@curry
+def flow_builder(
+    fn: Callable[..., None],
+    run_kwargs: dict = None,
+    register_kwargs: dict = None,
+    **flow_kwargs,
+) -> Callable[..., prefect.Flow]:
+    """
+    Wraps a function with a Prefect Flow context and calls `run_or_register` so the
+    flow file can be run as a script. The function *must* take `flow` as the first
+    argument. The flow name defaults to the name of the function with '_' replaced with
+    '-'.
+
+    Args:
+        fn: The callable to wrap
+        run_kwargs: A dict of kwargs to pass to `run_or_register` for script run calls
+        register_kwargs: A dict of kwargs to pass to `run_or_register` for script
+            register calls
+        **flow_kwargs: Additional kwargs as passed to the `Flow` initializer
+
+    Returns:
+        A wrapped function with the flow context.
+
+    Examples:
+
+        Tasks called from within the function are added to the flow
+        ```python
+        from prefect import task
+        from prefect.utilities.flows import flow_builder
+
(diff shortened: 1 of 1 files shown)


madkinsz:
Also considered having the decorator be named `flow` and having the first arg be `self` but I'm hesitant to make it so people can't have `flow` in their global namespace because it'd collide with this utility.

madkinsz:
Another possible name that does not collide and does not introduce another "named thing" is `@with_flow`. We could also consider making the `Flow` object work as a decorator

```python
class Flow:
    def __init__(self, name_or_fn, *, name: str = None):
        if callable(name_or_fn):
            self._function = name_or_fn
            self.name = name or self._function.__name__.replace("_", "-")
        else:
            if name_or_fn and name:
                raise ValueError("Multiple values received for Flow name")
            self._function = None
            self.name = name_or_fn

    def __call__(self, *args, **kwargs):
        if not self._function:
            raise ValueError("Flow cannot be called unless used as a decorator")
        with self:
            return self._function(self)

    def __enter__(self):
        print("Enter context...")

    def __exit__(self, exc_type, exc_val, exc_tb):
        print("Exit context.")


@Flow
def test(self):
    print(self)


test()
```

madkinsz:
Closing for now, perhaps a use-case in the future.

PR #3940, opened 2021-01-08
Better error messages on failure to load flow

### LINKED ISSUE #3457: Add version info to serialized flow
Currently when a user registers a flow, we don't save any info regarding the python environment it was serialized with. This can result in opaque errors on flow load if the registration and execution environments don't match. See #3447 for example.

I'd argue that usually this is user error, and we'll likely be unable to catch all environment mismatch issues (when using pickle-based storage, any library that differs in the execution environment may result in a failure to load a flow). Still, we could check for the most common culprits:

- Python version difference
- Prefect version difference

The logic might be:

```python
try:
    flow = unpickle_or_import_flow(...)
except Exception:
    if python_versions_differ or prefect_versions_differ:
        raise SomeInformativeError
    raise
```

I'm not sure where we'd want to store this info. We already store the prefect version implicitly as part of all serialized objects (`__version__`), but we lack the python version. Should this go on the `Flow` schema? Or the storage schema? Since the python version should only become important when using pickle-backed storage, we might also store this info as part of the binary payload:

```python
def serialize_flow_as_pickle(flow):
    flow_bytes = cloudpickle.dumps(flow)
    return json.dumps({"flow": flow, "versions": {...}})
```

jcrist:
Currently flows are stored using `cloudpickle` by default. While
flexible, the error messages produced on failure can be opaque
(especially for new users). Since `pickle` makes the internal
datastructures part of the public api, changes in core library versions
can lead to subtle errors - it's hard for us to provide better error
messages without knowing these version differences exist.

This commit changes the storage representation to include extra metadata
(currently only version info), which allows for generating better error
messages. We currently encode the versions for things likely to lead to
pickling errors:

- `prefect`
- `python`
- `cloudpickle`

If the flow fails to load, version differences between the build and
execution environments are checked and added to the error message to
provide better information to the user. Additionally, if an
`ImportError` is raised we note that they may need to install additional
dependencies.

If the flow successfully loads, we also check that the build and
execution versions of `prefect` match, and warn if they don't.
Differences in prefect version can lead to subtle bugs (due to the
pickled representation of a class changing), but things *may* work
anyway. A warning provides a nudge to the user to keep their
dependencies aligned, but doesn't break things unnecessarily.

We also pin to pickle protocol 4, which is supported by all versions of
Python prefect supports. Previously we'd use cloudpickle's default,
which would default to version 5 on Python 3.8, causing flows to
fail to load on 3.7.

Fixes #3457.

This PR:

### DIFF
--- a/src/prefect/storage/_healthcheck.py
+++ b/src/prefect/storage/_healthcheck.py
@@ -39,6 +39,7 @@ def cloudpickle_deserialization_check(flow_file_paths: list):
     for flow_file in flow_file_paths:
         with open(flow_file, encoding="utf-8") as f:
             try:
+                # Match implementation in prefect.utilities.storage.flow_from_bytes_pickle
                 flow_bytes = binascii.a2b_base64(json.loads(f.read())["flow"])
                 flows.append(cloudpickle.loads(flow_bytes))
             except ModuleNotFoundError:
--- a/src/prefect/storage/_healthcheck.py
+++ b/src/prefect/storage/_healthcheck.py
@@ -5,7 +5,9 @@ deployment issues early on.
 """
 
 import ast
+import binascii
 import importlib
+import json
 import sys
 import warnings
 
@@ -35,9 +37,10 @@ def system_check(python_version: str):
 def cloudpickle_deserialization_check(flow_file_paths: list):
     flows = []
     for flow_file in flow_file_paths:
-        with open(flow_file, "rb") as f:
+        with open(flow_file, encoding="utf-8") as f:
             try:
-                flows.append(cloudpickle.load(f))
+                flow_bytes = binascii.a2b_base64(json.loads(f.read())["flow"])
+                flows.append(cloudpickle.loads(flow_bytes))
             except ModuleNotFoundError:
                 warnings.warn(
                     "Flow uses module which is not importable. Refer to documentation "
--- a/src/prefect/utilities/storage.py
+++ b/src/prefect/utilities/storage.py
@@ -1,8 +1,16 @@
+import binascii
 import importlib
+import json
+import sys
+import warnings
 from operator import attrgetter
 from typing import TYPE_CHECKING
+from distutils.version import LooseVersion
+
+import cloudpickle
 
 import prefect
+from prefect.utilities.exceptions import StorageError
 
 if TYPE_CHECKING:
     from prefect.core.flow import Flow  # pylint: disable=W0611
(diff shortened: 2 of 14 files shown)


jcrist:
Example error message:

```python-traceback
StorageError                              Traceback (most recent call last)
<ipython-input-11-11be1b3b1bf1> in <module>
----> 1 flow_from_bytes_pickle(s)

~/Code/prefect/src/prefect/utilities/storage.py in flow_from_bytes_pickle(data)
    224                 "dependencies installed."
    225             )
--> 226         raise StorageError("\n".join(parts)) from exc
    227
    228     run_prefect = run_versions["prefect"]

StorageError: An error occurred while unpickling the flow:
  ValueError('Oh no!')
This may be due to one of the following version mismatches between the flow build and execution environments:
  - cloudpickle: (flow built with '1.6.0', currently running with '1.5.0')
  - prefect: (flow built with '0.13.19+294.g4fb785aa2', currently running with '0.13.0')
```

src/prefect/storage/_healthcheck.py
```diff
@@ -35,9 +37,10 @@ def system_check(python_version: str):
 def cloudpickle_deserialization_check(flow_file_paths: list):
     flows = []
     for flow_file in flow_file_paths:
-        with open(flow_file, "rb") as f:
+        with open(flow_file, encoding="utf-8") as f:
             try:
-                flows.append(cloudpickle.load(f))
+                flow_bytes = binascii.a2b_base64(json.loads(f.read())["flow"])
```
madkinsz, line 42: A comment that this is matching the implementation of `flow_from_bytes_pickle` would be nice

src/prefect/utilities/storage.py
```diff
@@ -143,3 +151,87 @@ def extract_flow_from_module(module_str: str, flow_name: str = None) -> "Flow":
...
+
+    run_prefect = run_versions["prefect"]
+    reg_prefect = reg_versions.get("prefect")
+    if reg_prefect and LooseVersion(reg_prefect) != run_prefect:
+        warnings.warn(
+            f"This flow was built using Prefect {reg_prefect!r}, but you currently "
+            f"have Prefect {run_prefect!r} installed. We recommend loading flows "
+            "with the same Prefect version they were built with, failure to do so "
+            "may result in errors."
+        )
```
madkinsz, line 236: I'm sure I could figure it out if I looked long enough but I read through this three times and I'm confused that this is not covered by L206 where you loop over the versions to check for mismatches `for name, v1 in sorted(reg_versions.items()):` -- Can you add a comment to clarify?

madkinsz (commented):
Sweet! This feels really well done.

src/prefect/utilities/storage.py
jcrist: This is checking for a warning if we didn't error above (mismatches in prefect version can lead to subtle bugs, but may also work). Added a comment to clarify.

src/prefect/utilities/storage.py
```diff
@@ -143,3 +151,92 @@ def extract_flow_from_module(module_str: str, flow_name: str = None) -> "Flow":
...
+                    mismatches.append(
+                        f"  - {name}: (flow built with {v1!r}, currently running with {v2!r})"
+                    )
+        if mismatches:
+            parts.append(
+                "This may be due to one of the following version mismatches between "
+                "the flow build and execution environments:"
+            )
+            parts.extend(mismatches)
+        if isinstance(exc, ImportError):
```
joshmeek, line 219: Is there a way to relay what package is actually causing the error?

src/prefect/utilities/storage.py
```diff
@@ -143,3 +151,92 @@ def extract_flow_from_module(module_str: str, flow_name: str = None) -> "Flow":
...
+        flow_bytes = binascii.a2b_base64(info["flow"])
+        reg_versions = info["versions"]
+
+    run_versions = _get_versions()
+
+    try:
+        flow = cloudpickle.loads(flow_bytes)
+    except Exception as exc:
+        parts = ["An error occurred while unpickling the flow:", f"  {exc!r}"]
+        # Check for mismatched versions to provide a better warning if possible
```
madkinsz, line 204:
```suggestion
        # Check for mismatched versions to provide a better error if possible
```

src/prefect/utilities/storage.py
jcrist: Not in a way I'd like to do. The original error is still displayed, but inferring the package from the message is error prone. Further, the user may be missing several dependencies - elevating one (the first import error) feels incorrect to me.

src/prefect/utilities/storage.py
madkinsz: Ah that's helpful, thanks!

madkinsz (approved):

joshmeek (approved):

PR #3947, opened 2021-01-11
Document auto-labelling behavior.

j3pic:
## Summary

Document the esoteric auto-labelling behavior that happens
when you create a flow run for a Local Storage flow.

## Changes

Documentation only.

## Importance

After first setting up Prefect, I found that the agent I set up was just ignoring the flows I created. Someone in Slack was having the same problem, and it had to do with mismatching labels. The person in Slack did not understand why his flow run had a label.

Based on the view from the CLI, I expected my flow run to have no labels, but when I looked at it through the Web interface, my flow run also inexplicably had a label. The presence of this label is not explained by the documentation.

Then I discovered that other people were *so* surprised by this behavior, that an issue got filed asking for a flag to disable it. 

This behavior seems to be an endless source of surprise and confusion.

## Checklist

This PR:

- [N/A] adds new tests (if appropriate)
- [N/A] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

### DIFF
--- a/changes/pr3947.yaml
+++ b/changes/pr3947.yaml
@@ -1,2 +1,2 @@
 fix:
-  - "Document the auto-labelling behavior for Flow Runs of local Flows."
+  - "Document the auto-labelling behavior for Flow Runs of local Flows, and other stuff."
--- a/changes/pr3947.yaml
+++ b/changes/pr3947.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Document the auto-labelling behavior for Flow Runs of local Flows."
--- a/docs/orchestration/flow_config/run_configs.md
+++ b/docs/orchestration/flow_config/run_configs.md
@@ -58,6 +58,21 @@ register a flow with no labels it will only be picked up by Agents which also
 do not have labels specified.
 :::
 
+:::tip Automatic Labels
+If your Flow is configured for [Local Storage](storage.md#local), which is the
+default, then flow runs generated from it will automatically be given a label
+matching the hostname of the machine where the flow run was created. This behavior can be disabled
+by adding the following to your `config.toml`:
+
+    [flows.defaults]
+        [flows.defaults.storage]
+
+        add_default_labels = false
+
+However, doing this does not enable you to use Local Storage with any agent
+besides the [local agent](/orchestration/agents/local.md).
+:::
+
 ## Types
 
 Prefect has a number of different `RunConfig` implementations - we'll briefly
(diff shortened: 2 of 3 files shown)


changes/pr3947.yaml
```diff
@@ -0,0 +1,2 @@
+fix:
+  - "Document the auto-labelling behavior for Flow Runs of local Flows."
```
cicdw, line 2: Documentation PRs do not need changelog entries

docs/orchestration/flow_config/run_configs.md
```diff
@@ -58,6 +58,22 @@ register a flow with no labels it will only be picked up by Agents which also
...
+by adding the following to your `config.toml`:
+
+    [flows.defaults]
+        [flows.defaults.storage]
+
+        add_default_labels = false
+
+However, doing this does not enable you to use Local Storage with any agent
+besides the [local agent](/orchestration/agents/local.md), which expects
+the hostname label to be there by default.
```
cicdw, line 74: This blurb is strictly true but in order to be useful it needs to provide more context on why this behavior is currently implemented -- the auto-labeling is a feature to prevent non-local agents from submitting flows that are stored in a local filesystem.  

Using `LocalStorage` + a non-local agent is a supported pattern but requires additional setup (e.g., mounting volumes) that are agent and platform dependent.  This should be called out here so that people don't turn off autolabeling without understanding what additional work will be required.

cicdw (changes requested):
Thanks for the PR @j3pic - left a note that we'll need to expand the commentary so that users who decide to turn off auto labeling know what to do next

docs/orchestration/flow_config/run_configs.md
```diff
@@ -58,6 +58,22 @@ register a flow with no labels it will only be picked up by Agents which also
 do not have labels specified.
 :::
 
+:::tip Automatic Labels
+If your Flow is configured for [Local Storage](storage.md#local), which is the
+default, then flow runs generated from it will automatically be given a label
+matching the hostname of the machine where the flow run was created. This behavior can be disabled
+by adding the following to your `config.toml`:
```
cicdw, line 65: An additional comment on this - this is valid, but we should also call out the less blunt approach:
```python
flow.storage = Local(add_default_labels=False)
```
which can be configured on a flow-by-flow basis.

madkinsz:
Closing in favor of https://github.com/PrefectHQ/prefect/pull/4659 which does the final request changes. Thanks!

PR #3953, opened 2021-01-13
Avoid race condition when creation task run artifacts

jcrist:
Currently when running with Prefect Cloud, there's a rare race condition
that may occur when creating a task run artifact against a mapped task,
leading to errors. For now, we catch this error and retry a few times
with a fixed delay in between attempts.

No test added (hard to test, temporary fix), but I've confirmed this fix
works locally.

This PR:

### DIFF
--- a/src/prefect/artifacts.py
+++ b/src/prefect/artifacts.py
@@ -1,6 +1,8 @@
+import time
 from typing import Optional
 
 from prefect import context, Client
+from prefect.utilities.exceptions import ClientError
 
 
 def _running_with_backend() -> bool:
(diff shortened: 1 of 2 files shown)


src/prefect/artifacts.py
```diff
@@ -14,6 +16,27 @@ def _running_with_backend() -> bool:
     return bool(context.get("running_with_backend"))
 
 
+def _create_task_run_artifact(kind: str, data: dict) -> str:
+    client = Client()
+    task_run_id = context.get("task_run_id")
+    # XXX: there's a race condition in the cloud backend for mapped tasks where
```
madkinsz, line 22: XXX?

madkinsz (approved):

src/prefect/artifacts.py
joshmeek: I was wondering the same thing lol

joshmeek (approved):

src/prefect/artifacts.py
jcrist: https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments

I have my editor setup to highlight these comments, drawing attention to them as things that could be better (but aren't really todos).

PR #3976, opened 2021-01-19
Add CI step for running a Cloud flow on release

joshmeek:
This adds a new step to the UI for kicking off a flow run upon release. The flow this is set to run pins the netlify doc build to the most recent production master deploy.

cc @madkinsz If you want to also add your release flow to this branch


joshmeek:
Looks like I need to adjust how the token is being used in the step

jcrist (approved):

Issue #3974, opened 2021-01-19
GCSUpload task marks "Success" when trying to upload a non-string or byte type but does not upload anything


LoicEm:
## Description

Per the title,  a [prefect.tasks.gcp.storage.GCSUpload](https://docs.prefect.io/api/latest/tasks/gcp.html#gcsupload), when given a non string type to save, still marks "Success" without uploading anything.

## Expected Behavior

The documentation states that the data accepted must be either string or bytes.
I'd expect trying to feed another type of data should raise an error.

## Reproduction
I was able to reproduce the bug with this snippet of code:

```python
import prefect
from prefect.tasks.gcp.storage import GCSUpload
from prefect.run_configs import  LocalRun
from prefect.storage import Local

results_upload = GCSUpload(bucket="{your-bucket-here}")

run_config = LocalRun()
storage = Local()


with prefect.Flow("minimal_reproducible_example",
                  run_config=run_config,
                  storage=storage) as flow:
    results_upload("[1, 2, 3]", blob="test_file_str.txt")

    results_upload([1, 2, 3], blob="test_file_list.txt")

    results_upload({1: "2", 3: "4"}, blob="test_file_dict.txt")

flow.run()
```

With this run, only the "test_file_str.txt" appears, while all tasks are marked as success.

## Environment

I have set my GOOGLE_APPLICATION_CREDENTIALS to a path containing the needed service account key.

Optionally run `prefect diagnostics` from the command line and paste the information here. -->

joshmeek:
Thanks for raising @LoicEm in the code for GCSUpload it looks like the if/else chain for checking the type of the data needs an `else` block that raises a ValueError if the type does not match either string or bytes. Would you like to make this change? 🙂

LoicEm:
Created a PR #3978 ! First contribution to open source, your feedback is more than welcome :)

PR #3978, opened 2021-01-19
Raise error when invalid data type is uploaded with GCSUpload

LoicEm:
## Summary

Added a case when data uploaded is neither string nor bytes in `prefect.tasks.gcp.storage.GCSUpload`, raising an explicit `ValuError` in that case.

Solving issue #3974 

## Changes

Create an `else` case when determining the data type in `prefect.tasks.gcp.storage.GCSUpload`, raising a `ValueError` if type is neither string nor bytes.

Created an associated test in `tests/tasks/gcp/storage/test_gcs_upload_and_download.py`: I have created a `TestRunTimeValidation` object with this test a single function, took the naming from `test_gcs_copy.py` in the same folder.

## Importance

GCSUpload is at the moment failing silently when data uploaded is neither string nor bytes.
With this PR it will explicitely fail.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/gcp/storage.py
+++ b/src/prefect/tasks/gcp/storage.py
@@ -22,7 +22,7 @@ class GCSBaseTask(Task):
         chunk_size: int = 104857600,  # 1024 * 1024 B * 100 = 100 MB
         encryption_key_secret: str = None,
         request_timeout: Union[float, Tuple[float, float]] = 60,
-        **kwargs
+        **kwargs,
     ):
         self.bucket = bucket
         self.blob = blob
@@ -112,7 +112,7 @@ class GCSDownload(GCSBaseTask):
         chunk_size: int = None,
         encryption_key_secret: str = None,
         request_timeout: Union[float, Tuple[float, float]] = 60,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(
             bucket=bucket,
@@ -121,7 +121,7 @@ class GCSDownload(GCSBaseTask):
             chunk_size=chunk_size,
             encryption_key_secret=encryption_key_secret,
             request_timeout=request_timeout,
-            **kwargs
+            **kwargs,
         )
 
     @defaults_from_attrs(
@@ -232,7 +232,7 @@ class GCSUpload(GCSBaseTask):
         create_bucket: bool = False,
         encryption_key_secret: str = None,
         request_timeout: Union[float, Tuple[float, float]] = 60,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(
             bucket=bucket,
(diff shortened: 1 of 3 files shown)


joshmeek:
Looks great @LoicEm! Would you mind also adding a [changes](https://github.com/PrefectHQ/prefect/tree/master/changes) update for this task library fix?

madkinsz (commented):
This lgtm, a `match` in the `with pytest.raises(ValueError):` to make sure that it's the one you expect may be nice.

src/prefect/tasks/gcp/storage.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         chunk_size: int = 104857600,  # 1024 * 1024 B * 100 = 100 MB
         encryption_key_secret: str = None,
         request_timeout: Union[float, Tuple[float, float]] = 60,
-        **kwargs
+        **kwargs,
```
LoicEm, line 25: Weird behavior from `black` here, should I keep it that way or remove the change ?

madkinsz: A trailing comma is standard from `black` although this will break in py2.7 which doesn't matter since we require python3 -- so this change should be fine.

src/prefect/tasks/gcp/storage.py
```diff
@@ -335,6 +336,8 @@ def run(
             if content_encoding:
                 gcs_blob.content_encoding = content_encoding
             gcs_blob.upload_from_file(io.BytesIO(data), timeout=request_timeout)
+        else:
+            raise ValueError(f"data must be str or bytes: got {type(data)} instead")
```
jcrist, line 340:
```suggestion
            raise TypeError(f"data must be str or bytes: got {type(data)} instead")
```

tests/tasks/gcp/test_gcs_upload_download.py
```diff
@@ -126,3 +126,20 @@ def test_blob_name_can_be_overwritten_at_runtime_by_download(self, monkeypatch):
...
+
+        blob = MagicMock()
+        client = MagicMock()
+        client.return_value = MagicMock(
+            get_bucket=MagicMock(return_value=MagicMock(blob=blob))
+        )
+        monkeypatch.setattr("prefect.tasks.gcp.storage.get_storage_client", client)
+
+        with pytest.raises(
+            ValueError, match="data must be str or bytes: got .* instead"
```
jcrist, line 143:
```suggestion
            TypeError, match="data must be str or bytes: got .* instead"
```

src/prefect/tasks/gcp/storage.py
```diff
@@ -300,6 +300,7 @@ def run(
                 Can also be passed as a tuple (connect_timeout, read_timeout).
 
         Raises:
+            - ValueError: if data is neither string nor bytes.
```
jcrist, line 303:
```suggestion
            - TypeError: if data is neither string nor bytes.
```

jcrist (commented):
One small change and this looks good to me.

jcrist (approved):

PR #3988, opened 2021-01-21
POC - Infer Github storage values from git

madkinsz:
This is proof of concept and requires more helpful error messages, documentation updates, and tests.

## Summary

Adds inference to GitHub storage so the user does not need to provide a `repo` and `path` when they are using source control to write their flow.

## Changes

`repo` is now an optional argument to `GitHub`. If it is not provided, but a `path` is, and that `path` exists on the file system, we can infer that we were passed `__file__` and run `git` commands to infer the repo of that file and the path of the file within the repo to set default values.

Note: We could also update `ref` which I think is quite useful to do when developing but since it *has* a default value this would require more changes to the interface.

## Importance

Simplifies usage of Github storage to

```python
flow.storage = GitHub(path=__file__)
```

rather than repeating the information that we already know e.g. the name of the file that the code is in

## Checklist

This PR:

### DIFF
--- a/src/prefect/storage/github.py
+++ b/src/prefect/storage/github.py
@@ -1,3 +1,5 @@
+import os
+import subprocess
 from typing import TYPE_CHECKING, Any, Dict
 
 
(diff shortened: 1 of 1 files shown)


mahiki:
This could lead to a very nice cleanup of a problem area for teams with 100s of flows in git repo file storage.

joshmeek:
@madkinsz I like the idea here! With documentation this would be really useful for a lot of users with many flows in a repo as @mahiki said 🙂

madkinsz:
Closing as #4256 provides a better route for these inferences.

PR #3989, opened 2021-01-21
Refactor logging

### LINKED ISSUE #3820: extra_loggers doesn't work for "processes" scheduler in LocalDaskExecutor
## Description
When the `extra_loggers` config is set and the processes scheduler is used in `LocalDaskExecutor`, log messages don't ultimately make it to the flow.

## Expected Behavior
Log outputs in tasks are expected to make it to the main process and to stdout (when running the flow directly from the flow script).

## Reproduction
```
import sys
import logging

from prefect.environments import LocalEnvironment


LOGGER = logging.getLogger("my-named-logger")
LOGGER.addHandler(logging.StreamHandler(sys.stdout))
LOGGER.setLevel(logging.DEBUG)

@task()
def log_message():
    LOGGER.info("Hello world!")
    LOGGER.error("Foo!")


with Flow('flow-with-extra-logger') as flow:
    log_message()
flow.environment = LocalEnvironment(executor=LocalDaskExecutor(scheduler="processes"))

flow.run()
```

## Environment
The local agent can be used with `extra_loggers` set with an env variable:
```
prefect agent local start --env PREFECT__LOGGING__EXTRA_LOGGERS="['my-named-logger']"
```

jcrist:
This does several things, with the goal of making our logging more
robust:

- Changes the implementation of `CloudHandler` to use a single (lazily
  started) background thread, rather than a thread per handler. This
  makes it cheaper to register `CloudHandler` handlers for custom
  loggers.
- Better ensures that all logs are flushed to cloud on process exit
- Pre-checks all log messages for a max length (logging a warning if
  that length is exceeded, and truncating the offending log). This
  provides better user-facing warnings, rather than relying on an error
  message from cloud.
- Batches log messages to cloud by a max size, preventing a huge amount
  of logs from being sent in a single batch. With this and the above,
  only network/intermitent errors should result in logs failing to
  upload.
- In the case of intermitent errors, ensure that log messages aren't
  silently dropped. Previously if a single upload call failed, all logs
  in that batch would go missing. We now retry again later.
- When running with a `LocalDaskExecutor(scheduler="processes")`, catch
  and handle SIGTERM signals, ensuring the standard interpreter shutdown
  procedure is run (we rely on this for flushing all logs to cloud).
- Drops the `info` dict from log messages. Cloud is already ignoring it,
 no reason to use the extra bandwidth.

Fixes #3820.

This PR:

### DIFF
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -9,11 +9,11 @@ Note that Prefect Tasks come equipped with their own loggers.  These can be acce
 When running locally, log levels and message formatting are set via your Prefect configuration file.
 """
 import atexit
-import json
 import logging
 import sys
 import threading
 import time
+import warnings
 from queue import Empty, Queue
 from typing import Any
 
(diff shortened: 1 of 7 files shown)


src/prefect/utilities/logging.py
```diff
@@ -32,147 +32,146 @@
     "task_run_id",
 )
 
+MAX_LOG_LENGTH = 1_000_000  # 1 MB - max length of a single log message
+MAX_BATCH_LOG_LENGTH = 50_000_000  # 50 MB - max total batch size for log messages
```
jcrist, line 36: cc @cicdw, can you confirm these numbers are sufficient with cloud?

Note that the max size that can be sent in a batch is actually `MAX_BATCH_LOG_LENGTH + MAX_LOG_LENGTH - 1`. Since we can't (easily) peek to see if the next entry will exceed the total batch size, we batch all logs until empty or the limit is exceeded (since we check earlier to prevent entries larger than `MAX_LOG_LENGTH`, that's the max a single log can increment the total).

cicdw: Yea these are compatible with Cloud, although 50MB seems large to me (I know there's a small chance of hitting it, but I wouldn't be surprised if we reduced our payload limits at some point).

jcrist: What batch size would make more sense? 20 MB? 10?

cicdw: I vote 20

src/prefect/utilities/logging.py
```diff
@@ -32,147 +32,146 @@
...
+                pass
+
+    def stop(self) -> None:
+        """Flush all logs and stop the background thread"""
+        if hasattr(self, "thread"):
+            self._stopped = True
+            self.thread.join()
+            self._write_logs()
+            del self.thread
+            del self.client
```
madkinsz, line 82: Just curious why deleting the references matters

madkinsz (commented):
This looks a lot nicer than before!

src/prefect/utilities/logging.py
```diff
@@ -32,147 +32,146 @@
...
-        logs = []
+    def __init__(self) -> None:
+        self.queue = Queue()  # type: Queue[dict]
+        self.pending_logs = []  # type: List[dict]
+        self.pending_length = 0
+        self._stopped = False
+
+    def ensure_started(self) -> None:
+        """Ensure the log manager is started"""
+        if not hasattr(self, "thread"):
```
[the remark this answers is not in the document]
jcrist: That's what I did before (see first commit), but mypy didn't like that. I don't like mypy.

Happy to go back to how it was before, but I'll have to add a bunch of mypy annotations/asserts to make mypy pass.

madkinsz (approved):

PR #3998, opened 2021-01-22
A few changes to GitHub storage

jcrist:
- Default `ref` is now `None` (instead of `"master"`). This defaults to
  using the default branch for the repo, allowing for alternate default
  branch names than `"master"`.
- Makes `path` a required parameter for `GitHub` storage. This was
  always implicitly required (things would break if it was missing), but
  the code and serializers still allowed for it.
- Adds improved log messages when failing to load a flow from `GitHub`
  storage
- Adds more information to the github storage logs, including the commit
  sha for the code that's loaded in all cases. This makes it easier to
  determine the exact source used when running a flow.
- Improved test coverage

This PR:

### DIFF
--- a/changes/pr3998.yaml
+++ b/changes/pr3998.yaml
@@ -1,0 +1,3 @@
+enhancement:
+  - "`GitHub` storage now logs the commit sha used when loading a flow - [#3998](https://github.com/PrefectHQ/prefect/pull/3998)"
+  - "`GitHub` storage now loads from a repo's default branch, allowing default branch names other than 'master' - [#3998](https://github.com/PrefectHQ/prefect/pull/3998)"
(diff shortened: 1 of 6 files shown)


jcrist:
Example logs:

```
[2021-01-22 20:30:06,847] INFO - agent | Deploying flow run a8121b2f-b5b7-43e6-8248-e904b5191b05
[2021-01-22 14:30:09-0600] INFO - prefect.GitHub | Downloading flow from GitHub storage - repo: 'jcrist/prefect-hacking', path: 'test_github.py'
[2021-01-22 14:30:10-0600] INFO - prefect.GitHub | Flow successfully downloaded. Using commit: d856f499589ae73b98a5924298eef20e457e62a4
[2021-01-22 14:30:10-0600] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'test-github'
```

madkinsz:
This is nice! I don't think this affects #3988 other than making it easier to change the default for `ref` but I'm still kind of lost on the best way to make the inference interface clear if you want to weigh in while this is fresh in your mind.

joshmeek (approved):

PR #4006, opened 2021-01-25
Handle missing flow runs in k8s agent resource management

joshmeek:
## Summary
The resource management loop in the Kubernetes agent has the potential to error if it attempts to set the state for a missing flow run. This can happen in the case where a job is started, the user deletes the flow run, then the agent attempts to clean up the job (when the job is in a state where the agent marks the flow run as failed). To avoid this we handle the case where a `ClientError` is raised by the call to the client and the loop continues with its resource management cycle.

## Changes
Handles `ClientError`s when calling the `set_flow_run_state` mutation in the Kubernetes agent resource management loop

## Importance
Avoids a possible edge case

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -199,10 +199,11 @@ class KubernetesAgent(Agent):
                                                     )
                                                 ),
                                             )
-                                        except ClientError:
+                                        except ClientError as exc:
                                             self.logger.error(
                                                 "Error attempting to set flow run state for "
-                                                f"{flow_run_id}"
+                                                f"{flow_run_id}:"
+                                                f"{exc}"
                                             )
 
                                         delete_job = True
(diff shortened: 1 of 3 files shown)


src/prefect/agent/kubernetes/agent.py
```diff
@@ -320,14 +327,19 @@ def manage_jobs(self) -> None:
...
+                                    flow_run_id=flow_run_id,
+                                    state=Failed(
+                                        message="Kubernetes Error: pods {} failed for this job".format(
+                                            failed_pods
+                                        )
+                                    ),
+                                )
+                            except ClientError:
+                                self.logger.error(
+                                    f"Error attempting to set flow run state for {flow_run_id}"
```
madkinsz, line 341: Can you attach the `exc_info` ?

madkinsz: (and in the other place where this is logged)

madkinsz (approved):

PR #4013, opened 2021-01-26
Support `EXTRA_PIP_PACKAGES` in docker image

jcrist:
This adds support for a `EXTRA_PIP_PACKAGES` environment variable in our
official `prefect` docker images. This takes a space delimited list of
packages to pass to `pip install` on image startup, allowing for easier
runtime installation of extra packages. For production deployments users
should still build images with their required dependencies, but for
development this can make things a bit easier. The official dask images
also support this, so there's some precedence here.

Also extends our default entrypoint to drop the user into a shell if no
CMD is specified (with a header explaining what the image is). We've had
a few users confused with what the prefect image was for (mostly
assuming it would start prefect server automatically), so hopefully this
helps remove this confusion.

This PR:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,6 @@ RUN apt update && \
     apt clean && apt autoremove -y && \
     rm -rf /var/lib/apt/lists/*
 
-ENTRYPOINT ["tini", "-g", "--"]
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
+ENTRYPOINT ["tini", "-g", "--", "entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


jcrist:
For example, if developing a flow that requires some extra packages not found in the official images (`my-extra-package1`, `my-extra-package2`), you might use the following run-config to have those packages automatically installed at runtime:

```python
flow.run_config = DockerRun(
    env={"EXTRA_PIP_PACKAGES": "my-extra-package1 my-extra-package2"}
)
```

madkinsz (approved):

joshmeek (approved):

PR #4017, opened 2021-01-27
Docs issue: Fix option for generating a RUNNER scope token

VioletVivirand:
## Summary

The [Tokens doc](https://docs.prefect.io/orchestration/concepts/tokens.html#cli) illustrates that the way we generate a RUNNER token with the CLI is to add a `-r` option, but now it's `-s` option (I guess `s` stands for **scope**?).

## Changes

Change the option for generating RUNNER scope token from `-r` to `-s`, so the command becomes:

```bash
$ prefect auth create-token -n my-runner-token -s RUNNER
```

## Importance

The command can be executed successfully now.

## Checklist

This PR:


joshmeek (approved):
Thanks @VioletVivirand!

PR #4021, opened 2021-01-27
Add option to provide flow ID when calling run flow CLI command

joshmeek:
This PR adds the option to provide a flow ID to the `prefect run flow` CLI command.

### DIFF
--- a/src/prefect/cli/run.py
+++ b/src/prefect/cli/run.py
@@ -94,9 +94,9 @@ def flow(
 
     \b
     Options:
-        --name, -n                  TEXT        The name of a flow to run [required]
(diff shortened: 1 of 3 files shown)


src/prefect/cli/run.py
```diff
@@ -110,6 +114,10 @@ def flow(
         --no-url                                Only output the flow run id instead of a
                                                 link
 
+    \b
+    Either `id` or both `name` and `project` must be provided to run a flow. If all three
+    are provided then `id` will be used when calling the mutation.
```
jcrist, line 119: Perhaps instead of using the flow id and ignoring the project & name, providing both should error?

jcrist (approved):

Issue #3986, opened 2021-01-21
Support multiple networks for containers deployed by DockerAgent


peterroelants:
## Current behavior

Currently the `DockerAgent` takes as `network` only a single network as string. See [`prefect/agent/docker/agent.py97`](https://github.com/PrefectHQ/prefect/blob/0adc072d1f6628905c644f32cf921c5b54948a0c/src/prefect/agent/docker/agent.py#L97). 

## Proposed behavior

Change `DockerAgent`'s `network` parameter to `networks`, which can take in multiple networks to be added to the container (similar to the `volumes` parameter).

Adapt the `prefect agent` CLI behaviour to take in a list of network names. See [`prefect/cli/agent.py:516`](https://github.com/PrefectHQ/prefect/blob/98c46d0a3f96a9998c3460ad32b43a731a06341d/src/prefect/cli/agent.py#L516). For example the CLI could look like:
```
prefect agent docker start \
  --network prefect-server kafka
```

## Example

This would allow for a more complex deployment where the deployed task container can communicate with multiple containers on multiple networks. For example in my use case I have a prefect server and a kafka endpoint deployed, each on different networks, the agent needs to communicate with instances on both networks. Currently the agent can only communicate with instances from one network, so everything has to be deployed under the prefect-server network.

peterroelants:
If this is deemed useful I might have a stab at creating a PR to implement this myself.

joshmeek:
@peterroelants Yeah this seems like a useful enhancement! Instead of replacing `network` on the `DockerAgent` class let's add the new `networks` kwarg and instead raise a `UserWarning` in the case someone provides values to `network` mentioning that it is deprecated. Mainly because I don't want to break any users who might be using that kwarg and instead it can follow the standard deprecation cycle.

peterroelants:
Hi @joshmeek , I found some time to take a first stab at a PR: https://github.com/PrefectHQ/prefect/pull/4038 . I would love to get some feedback on this.

As you suggested I decided to keep the `network` argument in the Docker Agent and deprecate it in favor of the new `networks` argument. There is however a slight breaking change using the `prefect agent docker start` CLI command, this will assign the arguments provided to `--network` to `networks` in the Docker Agent. Otherwise I would've needed to introduce a new command line option `--networks`, which I didn't want to do to keep consistency with the `--volume` argument.

peterroelants:
Resolved in https://github.com/PrefectHQ/prefect/pull/4038

PR #4038, opened 2021-01-31
Allow Docker Agent to connect container to multiple Docker networks

peterroelants:
## Summary
Allow Docker Agent to connect to multiple Docker networks (https://github.com/PrefectHQ/prefect/issues/3986).

## Changes
- `prefect agent docker start` allows connecting to multiple networks, e.g. `--network network1 --network network2`.
- `prefect agent docker start --network <network-name>` assigns `<network-name>` to `networks` argument instead of `network`.
- `DockerAgent` has new `networks` argument, the `network` argument is set to deprecated.

## Importance
A Docker container started by a Docker Agent might want to connect to multiple docker networks (e.g. in my use case a Kafka network, a database network, and the prefect network). This was previously not possible, this PR makes this possible.

## Checklist
This PR:

### DIFF
--- a/changes/3986_docker_networks.yaml
+++ b/changes/3986_docker_networks.yaml
@@ -3,3 +3,6 @@ enhancement:
 
 contributor:
   - "[Peter Roelants](https://github.com/peterroelants)"
+
+deprecation:
+  - "Docker agent `network` kwarg deprecated in favor of `networks`"
--- a/changes/3986_docker_networks.yaml
+++ b/changes/3986_docker_networks.yaml
@@ -5,4 +5,4 @@ contributor:
   - "[Peter Roelants](https://github.com/peterroelants)"
 
 deprecation:
-  - "Docker agent `network` kwarg deprecated in favor of `networks`"
+  - "Docker agent `network` kwarg deprecated in favor of `networks` - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
--- a/changes/3986_docker_networks.yaml
+++ b/changes/3986_docker_networks.yaml
@@ -1,0 +1,8 @@
+enhancement:
+  - "Support multiple docker networks with Docker Agent - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
+
+breaking:
+  - "Docker agent CLI `--network` parameter gets assigned to `networks` argument"
+
+contributor:
+  - "[Peter Roelants](https://github.com/peterroelants)"
--- a/changes/3986_docker_networks.yaml
+++ b/changes/3986_docker_networks.yaml
@@ -1,8 +1,5 @@
 enhancement:
   - "Support multiple docker networks with Docker Agent - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
 
-breaking:
-  - "Docker agent CLI `--network` parameter gets assigned to `networks` argument"
-
 contributor:
   - "[Peter Roelants](https://github.com/peterroelants)"
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -673,7 +673,7 @@ class Agent:
             version=flow_run.version,
             state=Failed(message=str(exc)),
         )
-        self.logger.error("Error while deploying flow: {}".format(repr(exc)))
+        self.logger.error("Error while deploying flow: {}".format(repr(exc)), exc_info=exc)
 
     def _get_run_config(
         self, flow_run: GraphQLResult, run_config_cls: Type[RunConfig]
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -673,7 +673,7 @@ class Agent:
             version=flow_run.version,
             state=Failed(message=str(exc)),
         )
-        self.logger.error("Error while deploying flow: {}".format(repr(exc)), exc_info=exc)
+        self.logger.error("Error while deploying flow", exc_info=exc)
 
     def _get_run_config(
         self, flow_run: GraphQLResult, run_config_cls: Type[RunConfig]
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -5,6 +5,7 @@ import re
 import sys
 from sys import platform
 from typing import TYPE_CHECKING, Dict, Iterable, List, Tuple
+import warnings
 
 from prefect import config, context
 from prefect.agent import Agent
@@ -74,6 +75,8 @@ class DockerAgent(Agent):
         - volumes (List[str], optional): a list of Docker volume mounts to be attached to any
             and all created containers.
         - network (str, optional): Add containers to an existing docker network
+            (deprecated in favor of `networks`).
+        - networks (List[str], optional): Add containers to existing Docker networks.
         - docker_interface (bool, optional): Toggle whether or not a `docker0` interface is
             present on this machine.  Defaults to `True`. **Note**: This is mostly relevant for
             some Docker-in-Docker setups that users may be running their agent with.
@@ -95,6 +98,7 @@ class DockerAgent(Agent):
         volumes: List[str] = None,
         show_flow_logs: bool = False,
         network: str = None,
+        networks: List[str] = None,
         docker_interface: bool = True,
         reg_allow_list: List[str] = None,
     ) -> None:
@@ -130,9 +134,22 @@ class DockerAgent(Agent):
             self.host_spec,
         ) = self._parse_volume_spec(volumes or [])
 
-        # Add containers to a docker network
+        # Add containers to the given Docker networks
+        if networks and network:
+            raise ValueError(
+                "Only provide either `network` or `networks` argument, not both!"
+            )
+        if network:
+            warnings.warn(
+                "DockerAgent `network` argument is deprecated and will be removed from Prefect. "
+                "Use `networks` instead.",
+                UserWarning,
+            )
+        # TODO: Only keep `self.networks` when `network` parameter gets removed.
         self.network = network
-        self.logger.debug("Docker network set to {}".format(self.network))
+        self.logger.debug(f"Docker network set to {self.network}")
+        self.networks = networks
+        self.logger.debug(f"Docker networks set to {self.networks}")
 
         self.docker_interface = docker_interface
         self.logger.debug(
@@ -156,11 +173,10 @@ class DockerAgent(Agent):
                 "Issue connecting to the Docker daemon. Make sure it is running."
             )
             raise exc
-
         self.logger.debug(f"Base URL: {self.base_url}")
         self.logger.debug(f"No pull: {self.no_pull}")
         self.logger.debug(f"Volumes: {volumes}")
-        self.logger.debug(f"Network: {self.network}")
+        self.logger.debug(f"Networks: {self.networks}")
         self.logger.debug(f"Docker interface: {self.docker_interface}")
 
     def _get_docker_client(self) -> "docker.APIClient":
@@ -399,17 +415,25 @@ class DockerAgent(Agent):
             host_config.update(extra_hosts={"host.docker.internal": docker_internal_ip})
 
         networking_config = None
+        # At the time of creation, you can only connect a container to a single network,
+        # however you can create more connections after creation.
+        # Connect first network in the creation step. If no network is connected here the container
+        # is connected to the default `bridge` network.
+        # The rest of the networks are connected after creation.
+        if self.networks:
+            networking_config = self.docker_client.create_networking_config(
+                {self.networks[0]: self.docker_client.create_endpoint_config()}
+            )
+        # Try fallback on old, deprecated, behaviour.
         if self.network:
             networking_config = self.docker_client.create_networking_config(
                 {self.network: self.docker_client.create_endpoint_config()}
             )
-
         labels = {
             "io.prefect.flow-name": flow_run.flow.name,
             "io.prefect.flow-id": flow_run.flow.id,
             "io.prefect.flow-run-id": flow_run.id,
         }
-
         container = self.docker_client.create_container(
             image,
             command=get_flow_run_command(flow_run),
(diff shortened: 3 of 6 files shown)


src/prefect/agent/docker/agent.py
```diff
@@ -399,17 +415,25 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
         networking_config = None
+        # At the time of creation, you can only connect a container to a single network,
+        # however you can create more connections after creation.
+        # Connect first network in the creation step. If no network is connected here the container
+        # is connected to the default `bridge` network.
+        # The rest of the networks are connected after creation.
+        if self.networks:
+            networking_config = self.docker_client.create_networking_config(
+                {self.networks[0]: self.docker_client.create_endpoint_config()}
+            )
```
peterroelants, line 426: Interestingly it is not possible to add multiple networks while creating the container. The docker-py documentation mentions under "Networking":

>At the time of creation, you can only connect a container to a single networking, but you can create more connections by using connect_container_to_network().

peterroelants: Note that adding all networks via `connect_container_to_network` below results in slightly different behaviour where also the default `bridge` network is added to the container (seemingly as a default when no `networking_config` is passed to `create_container`.

src/prefect/agent/docker/agent.py
```diff
@@ -419,11 +443,18 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
             networking_config=networking_config,
             labels=labels,
         )
```
peterroelants, line 445: Small note: During debugging this method was giving some difficult to interpret `Error while deploying flow: APIError(HTTPError('400 Client Error: Bad Request for url: http+docker://localhost/v1.41/containers/create'))` errors. The logging seems to remove the stack-trace.

To make debugging easier I surrounded this with a try-except block that would log the exception with stack-trace and re-raise it. Is this something we would want to put in by default to make logging of exceptions coming from the Docker API more verbose?

peterroelants:
Another remark: The current unit tests seem to mock the Docker client and thus container creation. The unit test didn't capture the issue that its not possible to link more than one network when creating a container (see https://github.com/PrefectHQ/prefect/pull/4038#discussion_r567474534), I only figured this out by actually running the code on my use case.

changes/3986_docker_networks.yaml
```diff
@@ -0,0 +1,8 @@
+enhancement:
+  - "Support multiple docker networks with Docker Agent - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
+
+breaking:
```
joshmeek, line 4: I don't think this would be a breaking change, doesn't have any effect on users who are starting the agent strictly through the CLI 👍

src/prefect/agent/docker/agent.py
```diff
@@ -130,9 +134,22 @@ def __init__(
...
+            raise ValueError(
+                "Only provide either `network` or `networks` argument, not both!"
+            )
+        if network:
+            warnings.warn(
+                "DockerAgent `network` argument is deprecated and will be removed from Prefect. "
+                "Use `networks` instead.",
+                UserWarning,
+            )
+        # TODO: Only keep `self.networks` when `network` parameter gets removed.
```
joshmeek, line 148: Not sure if the TODO comment is needed, in the standard deprecation cycle we would probably know what to remove when this comes around 🙂

src/prefect/agent/docker/agent.py
joshmeek: What information does the original stack track contain that helped you diagnose? Do you have an example you could post?

src/prefect/agent/docker/agent.py
joshmeek: I don't think this should be a problem. Is there a case where you wouldn't want to default bridge network?

src/prefect/agent/docker/agent.py
```diff
@@ -419,11 +443,18 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
+            for network in self.networks[1:]:
+                self.docker_client.connect_container_to_network(
+                    container=container, net_id=network
+                )
         # Start the container
         self.logger.debug(
             "Starting Docker container with ID {}".format(container.get("Id"))
         )
+        if self.networks:
+            self.logger.debug(f"Adding container to docker networks: {self.networks}")
```
joshmeek, line 457: Could you also log the container ID here in this debug statement?

madkinsz:
> Another remark: The current unit tests seem to mock the Docker client and thus container creation. The unit test didn't capture the issue that its not possible to link more than one network when creating a container (see [#4038 (comment)](https://github.com/PrefectHQ/prefect/pull/4038#discussion_r567474534)), I only figured this out by actually running the code on my use case.

This is because of the unfortunate limitation in our current CI where we are not running a docker server and consequently have weird test coverage here.

src/prefect/agent/docker/agent.py
peterroelants: I don't know of a case where you wouldn't want to connect to the default network.
However not adding the `bridge` network as a default has the advantage that:
- It is consistent with previous behaviour of just passing in a single network.
- Still allows to pass in the bridge network explicitly.

That being said, I don't have a strong personal preference and happy to implement either here.

src/prefect/agent/docker/agent.py
peterroelants: Printing the original stacktrace with `self.logger.exception` prints the following for example:
```
400 Client Error for http+docker://localhost/v1.41/containers/create: Bad Request ("Container cannot be connected to network endpoints: prefect-server, kafka-network")
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http+docker://localhost/v1.41/containers/create

  File "/home/peter/workspace/prefect/src/prefect/agent/docker/agent.py", line 433, in deploy_flow
    container = self.docker_client.create_container(
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/api/container.py", line 430, in create_container
    return self.create_container_from_config(config, name)
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/api/container.py", line 441, in create_container_from_config
    return self._result(res, True)
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/api/client.py", line 274, in _result
    self._raise_for_status(response)
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/api/client.py", line 270, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/home/peter/miniconda3/envs/prefect/lib/python3.8/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.APIError: 400 Client Error for http+docker://localhost/v1.41/containers/create: Bad Request ("Container cannot be connected to network endpoints: prefect-server, kafka-network")
```

joshmeek: Yeah let's expose this stack trace it has much more detail!

peterroelants: Modified the code to expose the stacktrace

changes/3986_docker_networks.yaml
peterroelants: Ok, removed the `breaking`

src/prefect/agent/docker/agent.py
peterroelants: Removed the TODO

src/prefect/agent/docker/agent.py
peterroelants: Updated to also log the container ID.

src/prefect/agent/agent.py
```diff
@@ -673,7 +673,7 @@ def mark_failed(self, flow_run: GraphQLResult, exc: Exception) -> None:
             version=flow_run.version,
             state=Failed(message=str(exc)),
         )
-        self.logger.error("Error while deploying flow: {}".format(repr(exc)))
+        self.logger.error("Error while deploying flow: {}".format(repr(exc)), exc_info=exc)
```
peterroelants, line 676: More helpful logs by printing the stack trace:

Old behaviour (Example):
```
ERROR - SharepointDockerAgent | Error while deploying flow: APIError(HTTPError('500 Server Error: Internal Server Error for url: http+docker://localhost/v1.41/containers/create'))
```

New behaviour (example):
```
prefect-agent            | [2021-02-02 09:16:06,419] ERROR - SharepointDockerAgent | Error while deploying flow: APIError(HTTPError('500 Server Error: Internal Server Error for url: http+docker://localhost/v1.41/containers/create'))
prefect-agent            | Traceback (most recent call last):
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/api/client.py", line 268, in _raise_for_status
prefect-age
```
nt            |     response.raise_for_status()
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
prefect-agent            |     raise HTTPError(http_error_msg, response=self)
prefect-agent            | requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http+docker://localhost/v1.41/containers/create
prefect-agent            | 
prefect-agent            | During handling of the above exception, another exception occurred:
prefect-agent            | 
prefect-agent            | Traceback (most recent call last):
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/prefect/agent/agent.py", line 406, in deploy_and_update_flow_run
prefect-agent            |     deployment_info = self.deploy_flow(flow_run)
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/prefect/agent/docker/agent.py", line 436, in deploy_flow
prefect-agent            |     container = self.docker_client.create_container(
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/api/container.py", line 430, in create_container
prefect-agent            |     return self.create_container_from_config(config, name)
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/api/container.py", line 441, in create_container_from_config
prefect-agent            |     return self._result(res, True)
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/api/client.py", line 274, in _result
prefect-agent            |     self._raise_for_status(response)
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/api/client.py", line 270, in _raise_for_status
prefect-agent            |     raise create_api_error_from_http_exception(e)
prefect-agent            |   File "/opt/runtime/python_env/lib/python3.8/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
prefect-agent            |     raise cls(e, response=response, explanation=explanation)
prefect-agent            | docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.41/containers/create: Internal Server Error ("invalid volume specification: '/home/peter/prefect_run/download:.:rw': invalid mount config for type "bind": invalid mount path: '.' mount path must be absolute")

```

File: src/prefect/agent/docker/agent.py (hunk 1 above)
Comment (joshmeek) [reply src/prefect/agent/docker/agent.py:425]: I think having bridge added by default should be fine and we can adjust this behavior later if someone needs it to function otherwise 👍

File: src/prefect/agent/agent.py (hunk 6 above)
Comment (joshmeek) [reply src/prefect/agent/agent.py:676]: Nice! Much more descriptive

File: changes/3986_docker_networks.yaml [hunk 7]
```diff
@@ -0,0 +1,5 @@
+enhancement:
+  - "Support multiple docker networks with Docker Agent - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
+
```
joshmeek, line 3: Let's add one more to this under `deprecation` mentioning the deprecated Docker agent `network` kwarg

src/prefect/agent/agent.py
jcrist, line 676: If using `exc_info`, you can drop the repr'ing of the exception in the log, since it's automatically included.

```suggestion
        self.logger.error("Error while deploying flow", exc_info=exc)
```

joshmeek:
@jcrist Any idea why the unit tests might not be running on this PR?

jcrist:
Maybe github/circleci is/was having an issue? Not sure. You might close and immediately reopen, sometimes that can fix broken CI state.

peterroelants:
I see there is an issue with the code formatting. [From CircleCI](https://app.circleci.com/pipelines/github/PrefectHQ/prefect/7627/workflows/85da5ec5-b944-40f0-8ad7-d700c92b96e4/jobs/112060/parallel-runs/0/steps/0-104):
```
would reformat /root/project/src/prefect/agent/agent.py
Oh no! 💥 💔 💥
1 file would be reformatted, 506 files would be left unchanged.
```

However, when running `black` locally I don't see this, I get:
```
❯ black --check .
All done! ✨ 🍰 ✨
507 files would be left unchanged.
```

Any idea what could cause this discrepancy? Any idea how to get more detailed logs from running black on CircleCI?

src/prefect/agent/agent.py
peterroelants: Simplified the exception logging as suggested

changes/3986_docker_networks.yaml
peterroelants: Added deprecation info

peterroelants:
> I see there is an issue with the code formatting. 
And now it seems to be resolved ¯\_(ツ)_/¯

changes/3986_docker_networks.yaml
```diff
@@ -0,0 +1,8 @@
...
+contributor:
+  - "[Peter Roelants](https://github.com/peterroelants)"
+
+deprecation:
+  - "Docker agent `network` kwarg deprecated in favor of `networks`"
```
joshmeek, line 8:
```suggestion
  - "Docker agent `network` kwarg deprecated in favor of `networks` - [#3986](https://github.com/PrefectHQ/prefect/issues/3986)"
```

joshmeek (approved):

joshmeek:
Looks good thanks @peterroelants!

PR #4053, opened 2021-02-02
Fix task decorator chaining by using inspect.unwrap instead of __wrap…

marwan116:
…ped__

## Summary

We came across a bug when chaining more than one decorator with the prefect task decorator basically the method `_validate_run_signature` tries to unwrap the function provided to get to the underlying function but it does so by accessing `__wrapped__` instead of using something like `inspect.unwrap` that would traverse the `__wrapped__ ` chain …

so a simple decorator like this:
```
def simple_dec(func):
    @functools.wraps(func)
    def wrapped(*args, **kwargs):
        return func(*args, **kwargs)
    return wrapped
```
would work fine when chained with a task decorator - i.e.
```
@task
@simple_dec
def test(a: int, b:int) -> int:
    return a + b
```
but chaining the simple decorator more than once would fail -i.e.
```
@task
@simple_dec
@simple_dec
def test(a: int, b:int) -> int:
    return a + b
```
with the following error
```
  File "./_tmp/r.py", line 16, in <module>
    def test(a: int, b: int) -> int:
  File "/Users/marwansarieddine/.pyenv/versions/3.8.5/envs/pyinfima/lib/python3.8/site-packages/prefect/utilities/tasks.py", line 408, in task
    return prefect.tasks.core.function.FunctionTask(fn=fn, **task_init_kwargs)
  File "/Users/marwansarieddine/.pyenv/versions/3.8.5/envs/pyinfima/lib/python3.8/site-packages/prefect/core/task.py", line 158, in init
    old_init(self, *args, **kwargs)
  File "/Users/marwansarieddine/.pyenv/versions/3.8.5/envs/pyinfima/lib/python3.8/site-packages/prefect/tasks/core/function.py", line 60, in __init__
    prefect.core.task._validate_run_signature(fn)  # type: ignore
  File "/Users/marwansarieddine/.pyenv/versions/3.8.5/envs/pyinfima/lib/python3.8/site-packages/prefect/core/task.py", line 66, in _validate_run_signature
    raise ValueError(
ValueError: Tasks with variable positional arguments (*args) are not supported, because all Prefect arguments are stored as keywords. As a workaround, consider modifying the run() method to accept **kwargs and feeding the values to *args.
 ```
    

## Changes

A simple change is proposed to use `inspect.unwrap` instead of trying to access `__wrapped__`

## Importance

This would allow further flexibility in composing prefect tasks by chaining decorators, or using complex decorators that have calls to other decorators.

## Checklist

This PR:

### DIFF
--- a/changes/pr4053.yaml
+++ b/changes/pr4053.yaml
@@ -1,0 +1,5 @@
+fix:
+  - Fix task decorator chaining by using inspect.unwrap instead of __wrap__ - [#4053](https://github.com/PrefectHQ/prefect/pull/4053)
+contributor:
+  - "[Alex P.](https://github.com/alexifm)" 
+  - "[Marwan S.](https://github.com/marwan116)"
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -48,7 +48,7 @@ class NoDefault(enum.Enum):
 
 
 def _validate_run_signature(run: Callable) -> None:
-    func = getattr(run, "__wrapped__", run)
+    func = inspect.unwrap(run)
     try:
         run_sig = inspect.getfullargspec(func)
     except TypeError as exc:
--- a/tests/core/test_task.py
+++ b/tests/core/test_task.py
@@ -1,5 +1,6 @@
 import inspect
 import logging
+import functools
 from datetime import timedelta
 from typing import Any, Tuple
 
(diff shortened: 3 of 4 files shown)


jcrist (commented):
Thanks for the fix! Ideally we'd fix our signature validation to use `inspect.signature` instead of `inspect.getfullargspec` (`inspect.signature` follows `__wrapped__` automatically), but the fix you have is good for now. 

Would you mind adding a test in `tests/utilities/test_tasks.py` to check that multiple decorators can be chained? Should go somewhere near https://github.com/PrefectHQ/prefect/blob/9adf14306723edc7a5ae1dcc7d733c46326ff3c2/tests/utilities/test_tasks.py#L83.

alexifm:
@jcrist That's a fair point but `inspect.Signature` doesn't have nice attributes like `args` and `varargs` while the result of `inspect.getfullargspec` does.  So it's throw in the `unwrap` call or reimplement the logic in `inspect.getfullargspec`.  An unfortunate side effect of backward compatibility:

>         # getfullargspec() historically ignored __wrapped__ attributes,
>         # so we ensure that remains the case in 3.3+

jcrist:
Using `inspect.unwrap` is fine and sufficient, no changes needed there. Just a test (and [a changelog entry](https://docs.prefect.io/core/development/contributing.html#changelog)) and this is good to go.

marwan116:
@jcrist - I just added testing and a changelog entry

jcrist (approved):
LGTM. Either location for tests would have been fine, the tests for this functionality just happened to already be in the `utilities/test_tasks.py` file. Merging, thanks for the fix!

PR #4059, opened 2021-02-03
A few storage cleanups

jcrist:
- Remove unused `Storage.get_env_runner` method
- Move more common functionality to base class
- Simplify serializers to not copy-paste so much
- Make access token secret name used at runtime for `GitHub`, `GitLab`, and `Bitbucket` storage configurable. This lets users use multiple different access tokens within the same tenant. It also makes the configuration more clear (rather than passing a list of secrets, a single secret name is provided). Docs have been updated to reflect this.
- Remove unneeded `prefect/utilities/git.py`. These were pulled out, but only used by storage classes. It's simpler (and clearer IMO) to keep this functionality within the storage classes if they're the only consumers. [Indirection has a cost](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection).
- Updates tests to make better use of fixtures.


jcrist:
> Make access token secret name used at runtime for GitHub, GitLab, and Bitbucket storage configurable

This was the main purpose of this PR, the other changes kind of spun out as I started cleaning things up. I plan to do this for the other storage classes that make use of secrets, but since those have results (and sometimes tasks) that also depend on the hardcoded secret names this is a larger change. The classes changed here could be done without wider changes.

changes/pr4059.yaml
```diff
@@ -0,0 +1,2 @@
+enhancement:
+  - "Add `access_token_secret` to `GitHub`, `GitLab`, and `Bitbucket` storage, making the Prefect secret containing an access token to these services configurable - [#4059](https://github.com/PrefectHQ/prefect/pull/4059)"
```
joshmeek, line 2: Even though they're mainly private methods and probably weren't accessed I think it's worth calling out the removal of the get_env_runner functions and that the git utilities was absorbed by the respective storage classes.

joshmeek (approved):

PR #4061, opened 2021-02-03
Minor CLI enhancements

joshmeek:
## Summary
This PR adds two new options to the CLI:
- `--detect-changes` when calling `prefect register flow` that uses the flow's serialized hash as the idempotency key when registering (allowing the flow version to not be bumped if the flow has not changed)
- `--skip-if-exists` when calling `prefect create project` to safely query for the project by name and skip if it has already been created

cc @cicdw 

## Importance
This comes out of a direct request from some users that would make their CI/CD flow building and registration process smoother!

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/register.py
+++ b/src/prefect/cli/register.py
@@ -61,7 +61,8 @@ def register():
     hidden=True,
     multiple=True,
 )
-def flow(file, name, project, label):
+@click.option("--detect-changes", is_flag=True, help="Detect changes to the flow")
+def flow(file, name, project, label, detect_changes):
     """
     Register a flow from a file. This call will pull a Flow object out of a `.py` file
     and call `flow.register` on it.
@@ -74,15 +75,19 @@ def flow(file, name, project, label):
         --project, -p   TEXT    The name of a Prefect project to register this flow
         --label, -l     TEXT    A label to set on the flow, extending any existing labels.
                                 Multiple labels are supported, eg. `-l label1 -l label2`.
+        --detect-changes        If toggled, passes a serialized hash of the flow to the register
+                                function's idempotency key. This allows you to avoid bumping the
+                                registered flow's version if the flow has not changed.
 
     \b
     Examples:
         $ prefect register flow --file my_flow.py --name My-Flow -l label1 -l label2
     """
-
     # Don't run extra `run` and `register` functions inside file
     file_path = os.path.abspath(file)
     with prefect.context({"loading_flow": True, "local_script_path": file_path}):
         flow = extract_flow_from_file(file_path=file_path, flow_name=name)
 
-    flow.register(project_name=project, labels=label)
+    idempotency_key = flow.serialized_hash() if detect_changes else None
+
+    flow.register(project_name=project, labels=label, idempotency_key=idempotency_key)
--- a/src/prefect/cli/register.py
+++ b/src/prefect/cli/register.py
@@ -62,12 +62,12 @@ def register():
     multiple=True,
 )
 @click.option(
-    "--detect-changes",
+    "--skip-if-flow-metadata-unchanged",
     is_flag=True,
-    help="Detect changes to the flow on registration",
+    help="Skips registration if flow metadata is unchanged",
     hidden=True,
 )
-def flow(file, name, project, label, detect_changes):
+def flow(file, name, project, label, skip_if_flow_metadata_unchanged):
     """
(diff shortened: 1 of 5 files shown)


src/prefect/cli/register.py
```diff
@@ -74,15 +80,19 @@ def flow(file, name, project, label):
         --project, -p   TEXT    The name of a Prefect project to register this flow
         --label, -l     TEXT    A label to set on the flow, extending any existing labels.
                                 Multiple labels are supported, eg. `-l label1 -l label2`.
+        --detect-changes        If toggled, passes a serialized hash of the flow to the register
+                                function's idempotency key. This allows you to avoid bumping the
+                                registered flow's version if the flow has not changed.
```
madkinsz, line 85: We have to be really careful here because the serialized hash is **not** a great representation of whether or not the flow itself has changed.

jcrist: Agreed. Perhaps this should be renamed to something more verbose/descriptive of what it does. Perhaps

```
prefect register flow --skip-if-flow-structure-unchanged
```

joshmeek: Right right, we weren't exactly sure the best way to phrase this / what would make it more clear. It is specifically for the case where the metadata has not changed. Open for suggestions!

jcrist: Metadata may be a good word. Perhaps

```
prefect register flow --skip-if-flow-metadata-unchanged
```

madkinsz: `--only-on-metadata-changes` ? 🤷 tricky stuff

jcrist: I like `skip` as the prefix, since we use that elsewhere already. No thoughts on the suffix.

cicdw: What if we used `--idempotency-key` (which is very explicit), and if no special key is provided we default to `flow.serialized_hash()`?  That way we match the register kwarg without misleading anyone.

jcrist: `--idempotency-key` sounds like it takes a value, so it'd be weird to have it optionally not take a value. It's also a bit uncommon for a CLI flag to take 0 or 1 args, so this may confuse users. I can't think of a reason why a user would pass in an explicit idempotency key that wasn't the flow hash, and it's also not clear how a user would write a bash script that made use of a flow hash (how would they hash the flow from the CLI? hardcoding the value feels wrong).

I'd prefer `--skip-if-flow-metadata-unchanged` (or something like it), since this is what most users would want and is explicit about the behavior. If you're concerned about api compatibility, we might add an explicit flag to `flow.register` as well with the same name, so the apis exist in both python code and CLI.

amanda-wee:
Concerning the `--skip-if-exists` option. Refer to:
https://github.com/PrefectHQ/prefect/pull/3630
That is, project creation is always skipped if the project already exists; what that CLI option does is skip even _attempting_ to create the project if it is found to already exist.

joshmeek:
> Concerning the `--skip-if-exists` option. Refer to:
> #3630
> That is, project creation is always skipped if the project already exists; what that CLI option does is skip even _attempting_ to create the project if it is found to already exist.

Yeah the intention is to not bother attempting to call the create mutation if the project already exists. Currently the creation call will capture any `ClientError` and log it back which doesn't appear friendly when a project already exists so I think doing a check prior to that will lead to a better experience. (and it will still safely handle the error when a user doesn't opt into skipping if it exists)

src/prefect/cli/register.py
```diff
@@ -75,14 +81,22 @@ def flow(file, name, project, label):
         --label, -l     TEXT    A label to set on the flow, extending any existing labels.
                                 Multiple labels are supported, eg. `-l label1 -l label2`.
 
+        --skip-if-flow-metadata-unchanged       If toggled, passes a serialized hash of the flow to the
+                                                register function's idempotency key. This allows you to
+                                                avoid bumping the registered flow's version if the flow
+                                                metadata (structure) has not changed.
```
jcrist, line 87: I don't think we need to explain how this works behind the scenes, the user will generally only care what the effect is.

```suggestion
        --skip-if-flow-metadata-unchanged       If set, the flow will only be re-registered if its
                                                metadata or structure has changed.
```

jcrist (commented):
Overall this looks good to me, just one small fix.

jcrist (approved):

PR #4063, opened 2021-02-04
Implement CompressedSerializer

amanda-wee:
## Summary

- Adds a `Serializer` wrapper class named `CompressedSerializer` that allows any `Serializer` subclass object to be augmented with compression functionality.

## Changes

- Implements the `CompressedSerializer` class
- Implements the constructor for `CompressedSerializer` to take another `Serializer` subclass object, and either a compression format name (which will cause the predetermined compression/decompression functions to be used), or a compression function and a decompression function. The constructor also takes optional keyword argument dictionaries to pass to the compression function and decompression function respectively.
- Implements `__eq__` method for `CompressedSerializer`
- Implements `serialize` method for `CompressedSerializer` to forward the serialization job to the wrapped `Serializer`, then compresses the result with the compression function and returns it.
- Implements `deserialize` method for `CompressedSerializer` to decompress the value with the decompression function, then forwards the deserialization job to the wrapped `Serializer`, and returns the result. 

## Importance

Presently, if it is desired to add compression to a serializer, one option is to derive from that `Serializer` subclass and override `serialize` and `deserialize` to make use of compression/decompression. However, this can lead to a multiplicity of subclasses corresponding to each possible pair of `Serializer` subclass and compression/decompression function pair.

This wrapper `Serializer` allows any concrete `Serializer` subclass to be combined with any compression/decompression function pair that takes a `bytes` object and returns a `bytes` object (e.g., as provided by the `zlib`, `gzip`, `bz2`, and `lzma` standard modules), with optional keyword arguments. Hence, compression can be added to all current and future concrete `Serializer` subclasses, with only one new subclass. For convenience, standard module compression/decompression pairs can be specified by a compression format name instead of having to explicitly pass them.

## Checklist

This PR:


amanda-wee:
It looks like the tests failed for an unrelated reason ("Too long with no output (exceeded 10m0s): context deadline exceeded") but I do not have the permission to re-run the failed test.

jcrist (commented):
Thanks @amanda-wee, overall this looks to me.

I'm wondering if a more user-friendly api could/should be added? I think supporting arbitrary `compress`/`decompress` functions is useful, but most users will probably use something pretty standard. What do you think about making `compress`/`decompress` optional and adding a `format` kwarg for choosing common options? The user would need to pass in either `format` or a `compress`/`decompress` pair. Options might be `gzip`, `lzma`, `bz2`, `zip`:

```python
import lzma
import zlib
import gzip
import bz2

formats = {
    "lzma": (lzma.compress, lzma.decompress),
    "zip": (zlib.compress, zlib.decompress),
    "gzip": (gzip.compress, gzip.decompress),
    "bz2": (bz2.compress, bz2.decompress)
}
```

Thoughts?

amanda-wee:
Yes, I like that idea, and hopefully I'll have some time to implement it after lunch. As for format names, it may be better to stick to "zlib" instead of "zip". Reading Mark Adler's reply here:
https://stackoverflow.com/questions/20762094/how-are-zlib-gzip-and-zip-related-what-do-they-have-in-common-and-how-are-they

So zlib provides an implementation of the same Deflate compression algorithm as used by zip (and gzip), but it's possible that say, the header, checksum, and other wrapping metadata might differ, plus zip technically can use other compression algorithms (although that's just a historical curiosity).

I found another article in which the author mentioned removing the header and footer introduced by `zlib.compress` to obtain the same raw Deflated result as used by zip:
http://jon.glass/blog/has-fun-with-zlib/

amanda-wee:
I have implemented the suggested functionality. A few notes:
- the new parameter is named `compression_format` rather than `format` because it turns out that `format` is the name of a built-in function. Maybe no big deal to have just shadowed the name, so let me know if you think it should be named `format` after all.
- the new parameter is placed just after the `serializer` parameter. In a way this reduces the impact of the long-winded `compression_format` name because it could be used as a positional argument instead of a keyword argument, although I have written the tests to demonstrate its use as a keyword argument. We don't seem to be enforcing keyword arguments elsewhere, so I chose not to do so, but maybe it is a good idea.
- the tests failed, but it looks like that has nothing to do with my changes but rather are just brittle tests

amanda-wee:
@jcrist would you like me to do the parameter rename, and is there anything else I need to do before this can be merged? Thanks!

jcrist:
Hi Amanda, apologies for the delay here. Yeah, I think `format` as a kwarg makes more sense (many function's use this kwarg, shadowing a seldom-used builtin isn't that big of a deal). Thanks for the PR, I'll give this a larger review first thing tomorrow morning.

amanda-wee:
I've renamed the parameter and added a sanity check in case someone tries to pass both `format` and `compress` and `decompress` parameters.

src/prefect/engine/serializers.py
```diff
@@ -249,3 +261,95 @@ def _get_serialize_method(self, dataframe: "pd.DataFrame" = None) -> Callable:
...
+
+        if format:
+            try:
+                self._compress, self._decompress = COMPRESSION_FORMATS[format]
+            except KeyError as e:
+                raise ValueError("unknown compression format: {}".format(format)) from e
+
+            if (compress and compress != self._compress) or (
+                decompress and decompress != self._decompress
+            ):
```
jcrist, line 302: I don't think we should check for equality here. Rather, if `format` is provided, `compress`/`decompress` must not be provided.

src/prefect/engine/serializers.py
```diff
@@ -249,3 +261,95 @@ def _get_serialize_method(self, dataframe: "pd.DataFrame" = None) -> Callable:
...
+                    "Both compression format and custom compression/decompression "
+                    "functions provided but are incompatible."
+                )
+        elif compress and decompress:
+            self._compress = compress
+            self._decompress = decompress
+        else:
+            raise ValueError(
+                "CompressedSerializer must be provided the compression format or the "
+                "compression/decompression functions."
```
jcrist, line 313: The traceback will already indicate that this is in `CompressedSerializer`. It's also clearer (IMO) to specify the kwarg names required rather than their meanings.

```suggestion
                "You must specify either `format` or `compress`/`decompress`"
```

jcrist (commented):
Just two small nits, otherwise this looks good to me. Thanks for all your work here!

amanda-wee:
Thanks, the suggestions have been implemented. I have also improved the error message for an invalid format along similiar lines.

jcrist (approved):
Looks good to me! Thanks for all your work on this, will merge once tests pass.

PR #4075, opened 2021-02-05
Speedup imports

jcrist:
Delay importing some optional dependencies to speedup import time. Drops
`import prefect` time from 1.2 s to 0.34 s on my machine.

### DIFF
--- a/src/prefect/engine/cache_validators.py
+++ b/src/prefect/engine/cache_validators.py
@@ -13,7 +13,6 @@ Note that _all_ validators take into account cache expiration.
 
 A cache validator returns `True` if the cache is still valid, and `False` otherwise.
 """
-from dask.base import tokenize
 from typing import Any, Callable, Dict, Iterable
 
 import pendulum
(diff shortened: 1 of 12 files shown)


jcrist:
Side note - why do we import almost all of prefect on `import prefect`? Most users won't use the jira tasks, but they're still always imported. Only importing the items used in the top-level namespace on `import prefect` would speedup time to load significantly, and let us avoid this delayed-import pattern (since not every file would need to load initially).

I'm not sure we can step back from this at this point, since users may be relying on the following to work:

```python
import prefect

notifier = prefect.utilities.notifications.jira_notifications.jira_notifier(...)
```

Really, they should be importing this functionality directly

```python
import prefect
from prefect.utilities.notifications.jira_notifications import jira_notifier

notifier = jira_notifier(...)
```

We could hack around this by implementing a module-level `__getattr__` and doing a delayed import then, but this is more complicated than just not importing at all.

cicdw:
There wasn't any interesting reason for that mass import pattern other than convenience so we can definitely refactor that (I agree with your reasoning) - because the fix on the user side is so simple (changing attribute access to an import), I think it makes sense to consider as a breaking change for a future release.

PR #4076, opened 2021-02-05
Forward backend to deployed env

### LINKED ISSUE #4051: DockerAgent on Prefect Server tries to pull secrets from API
## Description

If a secret is not available in the context the API is queried for the secret (when the backend is `cloud`) or an error is raised (when the backend is `server`). The configured backend is not passed from the agent environment into the docker container so a confusing graphql error is raised instead of the nicer 'Local secret not found' when you're using the `server` backend.

```python
# secrets.py
        try:
            value = secrets[self.name]
        except KeyError:
           # This is the code path we want to hit
            if prefect.config.backend != "cloud":
                raise ValueError(
                    'Local Secret "{}" was not found.'.format(self.name)
                ) from None
           # Query the API...
```

## Expected Behavior

The backend is passed through to the container so the helpful error is raised

## Reproduction

A `DockerRun` flow with a secret that is not passed to the agent environment explicitly

jcrist:
Some logic in core does different behavior if using `server` or `cloud`
(for example, server doesn't support hosted secrets). This adds the
backend to the env forwarded by the agent ensuring deployed flow runs
know what kind of backend they're running against.

Fixes #4051.

This PR:


madkinsz (approved):
Thanks for getting to this so quickly :)

PR #4077, opened 2021-02-05
Drop default resource limits on k8s agent

### LINKED ISSUE #3942: Increase Kubernetes Agent CPU Limit Defaults
## Current behavior

Either due to increased scale in our tenant or enhanced abilities in the agent, we've begun experiencing CPU throttling at startup in Kubernetes, leading to failed health checks at start.  With proper resource limits, 40 seconds delay and 40 seconds period with 2 failure threshold (current health check defaults) should be plenty safe

## Proposed behavior

 100millicores is a very low limit for CPU for anything running in k8s, so I propose we increase that to a saner default in the k8s agent install deployment.yaml

jcrist:
These were extremely low, and there's not a good reason to have them
(IMO). If a user wants to ensure the agent doesn't exceed a limit they
can add one themself (or add a request), but we shouldn't assume any by
default.

Fixes #3942.

### DIFF
--- a/changes/pr4077.yaml
+++ b/changes/pr4077.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Drop resource limits from manifest generated using `prefect agent kubernetes install` - [#4077](https://github.com/PrefectHQ/prefect/pull/4077)"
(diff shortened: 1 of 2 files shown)


madkinsz:
Note that there ought to be a corresponding patch at https://github.com/PrefectHQ/server/blob/master/helm/prefect-server/values.yaml#L285

joshmeek (approved):

PR #4081, opened 2021-02-05
Add backend custom server config

steph-ben:
## Summary

Extend `prefect backend` to setup a custom server host.

## Changes

Easiest way to setup custom prefect server host, by using :

```shell
prefect backend http://my-beautiful-prefect-server
```

Will result :

```shell
$ cat ~/.prefect/backend.toml
backend = "server"
 
$ cat ~/.prefect/config.toml
[server]
host = "http://my-beautiful-prefect-server"
```

## Importance

In our test environment, we have two prefect servers:
 - one on my laptop for testing
 - one running on our infra

It's more easy to change the server host through the cli.

## Checklist

This PR:

### DIFF
--- a/docs/outline.toml
+++ b/docs/outline.toml
@@ -573,7 +573,7 @@ functions = ["merge_dicts", "as_nested_dict", "dict_to_flatdict", "flatdict_to_d
 [pages.utilities.configuration]
 title = "Configuration"
 module = "prefect.utilities.configuration"
-functions = ["set_temporary_config"]
+functions = ["set_temporary_config", "set_permanent_user_config"]
 
 [pages.utilities.context]
 title = "Context"
--- a/src/prefect/cli/__init__.py
+++ b/src/prefect/cli/__init__.py
@@ -17,7 +17,7 @@ from .run import run as _run
 from .server import server as _server
 from .heartbeat import heartbeat as _heartbeat
 from .register import register as _register
-
+from ..utilities.configuration import set_permanent_user_config
 
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
@@ -119,11 +119,29 @@ def diagnostics(include_secret_names):
 @click.argument("api")
 def backend(api):
     """
-    Switch Prefect API backend to either `server` or `cloud`
-    """
-    if api not in ["server", "cloud"]:
-        click.secho("{} is not a valid backend API".format(api), fg="red")
-        return
+    \b
+    Switch Prefect API backend :
+        - `cloud` for using prefect cloud
+        - `server` for using a localhost server
+        - any other string for specifying a custom server host
+            (~/.prefect/config.toml will be updated)
 
-    backend_util.save_backend(api)
-    click.secho("Backend switched to {}".format(api), fg="green")
+    """
+    if api == "cloud":
+        backend_util.save_backend(api)
+        click.secho("Backend switched to {}".format(api), fg="green")
+
+    if api == "server":
+        backend_util.save_backend(api)
+        click.secho("Backend switched to {}".format(api), fg="green")
+
+        # Set back to default localhost
+        fp = set_permanent_user_config({'server': {'host': "localhost"}})
+
+    if api not in ["server", "cloud"]:
+        # Switch backend to server
+        backend_util.save_backend("server")
+
+        # Expect a server hostname
+        fp = set_permanent_user_config({'server': {'host': api}})
+        click.secho(f"User config {fp} updated with server.host = {api}", fg="green")
--- a/src/prefect/cli/__init__.py
+++ b/src/prefect/cli/__init__.py
@@ -123,7 +123,7 @@ def backend(api):
     Switch Prefect API backend :
         - `cloud` for using prefect cloud
         - `server` for using a localhost server
-        - any other string for specifying a custom server host
+        - any other string for specifying a custom server host, eg. http://my-prefect-server.com
             (~/.prefect/config.toml will be updated)
 
     """
@@ -136,7 +136,7 @@ def backend(api):
         click.secho("Backend switched to {}".format(api), fg="green")
 
         # Set back to default localhost
-        fp = set_permanent_user_config({'server': {'host': "localhost"}})
+        set_permanent_user_config({'server': {'host': "http://localhost"}})
 
     if api not in ["server", "cloud"]:
         # Switch backend to server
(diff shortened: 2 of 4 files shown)


sm-Fifteen:
Wouldn't this lead to unexpected behavior in case of a typo like `prefect backend srever`? Not to mention it only allows setting the host component, not the port.

In fact, looking at the default config, prefect would probably behave weirdly if someone tried to set their port that way, using something like `prefect backend http://myserver:1234`:

https://github.com/PrefectHQ/prefect/blob/ec5790319c9b726c378479fd4a5d9abda3539513/src/prefect/config.toml#L9-L13

That is, it would resolve to `http://myserver:1234:4200`

Using parameters like `--host` and `--port` that apply only to `prefect backend server` would better mirror the configuration format.

steph-ben:
Thanks for the feedback, I think you're right.

Do you think we should update the apollo endpoint as well ?
As described here https://github.com/PrefectHQ/prefect/issues/4082#issuecomment-775204829

joshmeek:
Thanks for making this PR! We agree that this type of configuration is something that should be supported however there is probably a better implementation that we can take. The implementation you wrote takes automatic action on the `~/.prefect/config.toml` which is not the intended use of that file. The user config is intended for manual writing only as using the toml library to write new variables to it will override user formatting/comments/etc. When a user runs `prefect backend OPTION` that option is written to `~/.prefect/backend.toml` and that file is merged with config at runtime. `backend.toml` is intended for this type of automatic use and is not something that users should ever care about or manually edit. 

A better implementation for this that would be more resilient is to introduce a new type of backend context switching. So you could imagine in the `backend.toml` there would be sets of backend configurations like so (implementation details TBD):

```
backend = {name_of_current_active_backend}

[backends]
  [backends.server1]
  host, api, port, etc.

  [backends.server2]
  host, api, port, etc.
```

That way a user could come in and change their backend by running something like `prefect backend server2` (switches the backend to what is configured for the one named _server2_). Is this a change that you would be interested in making? If not I'm sure the prefect team would gladly implement it (it's been on our minds for a while).

sm-Fifteen:
> A better implementation for this that would be more resilient is to introduce a new type of backend context switching. So you could imagine in the `backend.toml` there would be sets of backend configurations like so (implementation details TBD):
> 
> ```
> backend = {name_of_current_active_backend}
> 
> [backends]
>   [backends.server1]
>   host, api, port, etc.
> 
>   [backends.server2]
>   host, api, port, etc.
> ```
> 
> That way a user could come in and change their backend by running something like `prefect backend server2` (switches the backend to what is configured for the one named _server2_). Is this a change that you would be interested in making? If not I'm sure the prefect team would gladly implement it (it's been on our minds for a while).

Wouldn't that require some changes elsewhere? On top of my head, I can think of at least one place where the backend type is validated by name:

https://github.com/PrefectHQ/prefect/blob/f1584b77e7e24e37f1f4d4cf3969535473d9d301/src/prefect/client/secrets.py#L139

If the names were to be made arbitrary and no longer be connected to the server type, that could end up requiring more complex changes than the current scope of this PR, which is about simply writing to a config file.

joshmeek:
> If the names were to be made arbitrary and no longer be connected to the server type, that could end up requiring more complex changes than the current scope of this PR, which is about simply writing to a config file.

Correct, this would be a more complex change to benefit users and supersedes the change currently proposed. As it stands now this PR won't be accepted since it overrides the user config. Since this PR is about writing to a config file, in the meantime, it is preferred that users take that action manually (by editing the `config.toml` or setting the host as an environment variable). I do agree that providing information such as the host when setting the backend would be a beneficial enhancement but I feel like the more complex approach is the road we should go down to better adapt this behavior.

steph-ben:
Thanks for you interest in this MR !

> Thanks for making this PR! We agree that this type of configuration is something that should be supported however there is probably a better implementation that we can take. The implementation you wrote takes automatic action on the `~/.prefect/config.toml` which is not the intended use of that file. The user config is intended for manual writing only as using the toml library to write new variables to it will override user formatting/comments/etc. When a user runs `prefect backend OPTION` that option is written to `~/.prefect/backend.toml` and that file is merged with config at runtime. `backend.toml` is intended for this type of automatic use and is not something that users should ever care about or manually edit.

Actually, by reading the following documentation :
* [orchestration tutorial](https://docs.prefect.io/orchestration/tutorial/overview.html#select-an-orchestration-backend)
* [connecting to a different Endpoint](https://docs.prefect.io/orchestration/faq/config.html#connecting-to-a-different-api-endpoint)
 
I was really expecting something simple to switch the server url, and I did the MR in this way (somehow it's just make [this part](https://docs.prefect.io/orchestration/faq/config.html#connecting-to-a-different-api-endpoint) easier).

I understand there can be a wide range of server deployment configuration, however I feel that newcomers (or most users?) will just follow the doc and simple use `prefect server start` method, that is deploying everything (ui, server, graphql, hasura) on one host (and [expect server.ui](https://github.com/PrefectHQ/prefect/issues/4082#issuecomment-775204829) to be updated).

Once said that, I understand the needs of specifying all config variable for each custom deployement.


> A better implementation for this that would be more resilient is to introduce a new type of backend context switching. So you could imagine in the `backend.toml` there would be sets of backend configurations like so (implementation details TBD):
> 
> ```
> backend = {name_of_current_active_backend}
> 
> [backends]
>   [backends.server1]
>   host, api, port, etc.
> 
>   [backends.server2]
>   host, api, port, etc.
> ```
> 
> That way a user could come in and change their backend by running something like `prefect backend server2` (switches the backend to what is configured for the one named _server2_). Is this a change that you would be interested in making? If not I'm sure the prefect team would gladly implement it (it's been on our minds for a while).

From the user point of view, I would still miss the ability to simply specify "one" server hostname from the cli, and everything will work smoothly. But I guess I can live with that.

From the technical point of view, I didn't dig in the current code to see how `backend.toml` is used, but from the few I've seen it seems a bit tricky. If no one has a better suggestion, I will try to have a look at it ?

madkinsz:
Hey @steph-ben -- If you're still interested in implementing this, feel free to take a look at using the `backend` config file. The ability to switch between backends by name would be useful to me as well :) you'll need to make sure that the backends are specified as 'cloud' or 'server' type backends so that `core` still knows what kind of requests to send (authenticated or not). 

Since the server.endpoint is generated with `endpoint = "${server.host}:${server.port}"` you could also support passing a backend via `prefect backend` (although I'm not sure this is the best option) e.g. `prefect backend my-server --create --endpoint <...>`

madkinsz:
Closing this for now since it's stale. Feel free to ping me if you're interested in pursuing this configuration.

Issue #4062, opened 2021-02-03
Add project name to context


kevin-smartersorting:
## Current behavior

Currently https://docs.prefect.io/api/latest/utilities/context.html#context the context does not expose the name of the project that the flow/task is being run in.

## Proposed behavior

Add key/value pair to the context that exposes this value

## Example

Metric-ing run stats in datadog by project would be trivialized if this context can be accessed.

cicdw:
This was completed in the PR above

PR #4083, opened 2021-02-06
add project name and id to context

heyitskevin:
## Summary

References this issue https://github.com/PrefectHQ/prefect/issues/4062
This PR seeks to add `project_name` and `project_id` to the context object.
NOTE: This PR leverages the existing Prefect `Client` object which means this feature is constrained to Prefect Cloud. I've noticed from a cursory scan of the code that Prefect Server does not appear to use the `Client` object. If this isn't the case I can add some basic exception handling to the code.

My GQL is a bit rusty so I may be slightly deviant in the query definition but I did my best to follow the docs available on the GQL Schema in Cloud.

## Changes

Modifies an existingGQL query to grab project id/name, surfaces that query result in the`update_context()` method

## Importance

Allows for further context introspection by seeing the referenced project (ex: PROD/STAGING) at a flow level context.

## Checklist

This PR:


jcrist (approved):

jcrist:
Thanks! To clarify, this will work for all flows executed by Cloud or Server, only local flows (done with `flow.run()`) will lack an associated project.

PR #4100, opened 2021-02-10
Add option to provide version group ID to `prefect run flow` CLI command

joshmeek:


src/prefect/cli/run.py
```diff
@@ -154,7 +162,7 @@ def flow(
 
     client = Client()
     flow_id = id
-    if not flow_id:
+    if not flow_id and not version_group_id:
```
jcrist, line 165: (github won't let me comment on lines too far above this change)

You'll need to update the logic on lines 135/142 to make `version_group_id` one of the 3 ways to specify what flow to run. Right now providing `version_group_id` alone will error, but should work fine.

src/prefect/cli/run.py
```diff
@@ -140,16 +140,16 @@ def flow(
...
+    if not id and not (name and project) and not version_group_id:
         click.secho(
-            "A flow ID or some combination of flow name and project must be provided.",
+            "A flow ID, version group ID,  or some combination of flow name and project must be provided.",
             fg="red",
         )
         return
 
-    if id and (name or project):
+    if (id or version_group_id) and (name or project):
```
jcrist, line 150: This will still let both `id` and `version_group_id` be passed, when we probably only want to accept one or the other.

jcrist (approved):

PR #4104, opened 2021-02-10
patch S3Result to use a global client

### LINKED ISSUE #4085: S3Result causes a memory leak in DaskExecutor
## Description
When using `S3Result` as a checkpoint container, more memory is consumed than expected. In the simple flow below, 1000 small jobs are mapped but the dask worker reports > 6GB of ram consumed. This memory persists until the dask-worker is restarted. I suspect that this is related to some boto3 resource that is not  being garbage collected.

## Expected Behavior
I expect that very little memory should be consumed from using a S3Result as a checkpoint container. If the container is switched to `LocalResult`, the memory consumption is much lower as expected. 

## Reproduction
```python

import time
from prefect.engine.results import S3Result


@task(result=S3Result(bucket='my-bucket'), target="jobs/{flow_name}/cache/{task_slug}-{map_index}/{task_slug}-{map_index}.cpt" )
def gpu_task(x):
    time.sleep(0.1)
    return x * 2

with Flow("MemLeak") as flow:
    x = range(1000)
    y = gpu_task.map(x)  # 100 gpu_map

flow.executor = DaskExecutor('0.0.0.0:8786')

flow.run()

```

xuevin:
## Summary
Patch the S3Result to initialize the boto client once. Resolves #4085 

## Changes

S3Result causes a memory leak in DaskExecutor. Instead of creating a client with every S3Result, create a global client and reuse it.

## Importance

If many tasks are cached, a Dask worker will eventually run out of memory and the job will stall. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/engine/results/s3_result.py
+++ b/src/prefect/engine/results/s3_result.py
@@ -9,6 +9,7 @@ if TYPE_CHECKING:
 
 S3_CLIENT = None
 
+
 class S3Result(Result):
     """
     Result that is written to and retrieved from an AWS S3 Bucket.
@@ -46,6 +47,7 @@ class S3Result(Result):
         Initializes an S3 Client.
         """
         from prefect.utilities.aws import get_boto_client
+
         global S3_CLIENT
         if S3_CLIENT:
             self.client = S3_CLIENT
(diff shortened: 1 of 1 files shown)


jcrist:
Thanks for the fix @xuevin, I can see how the original code could lead to excessive resource usage. I think we'll want to be a bit stricter about where and how these clients are created, stored, and cleaned up. Do you mind if I push an alternative fix up?

xuevin:
Yes! Please do! I look forward to seeing the changes and learning more. Thanks you!

jcrist:
See #4115.

PR #4105, opened 2021-02-10
A few additions to the storage docs

jcrist:
Expands the storage docs a bit:

- Standardizes use of `script based` flow storage, instead of `file based`. The kwarg is named `stored_as_script`, and not all backends refer to things as "files", so I think this makes more sense.

- Adds section to storage docs clarifying the difference between pickle and script based flow storage, and why you might pick one or the other.

- Adds section to the docs on choosing a storage class.

- A few small typo fixers.

### DIFF
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -71,7 +71,7 @@ This has a few nice properties:
   range of Prefect/Python/dependency versions.
 
 The downside is you may have to do a bit more configuration to tell prefect
-where your script is located (since it can't always automatically inferred).
+where your script is located (since it can't always be automatically inferred).
 
 Some storage classes (e.g. `GitHub`, `Bitbucket`, `GitLab`, ...) only support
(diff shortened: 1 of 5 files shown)


docs/orchestration/flow_config/storage.md
```diff
@@ -26,11 +26,89 @@ with Flow("example") as flow:
...
+  or edges) or the flow's metadata (e.g. updating the run config) will require
+  re-registration, but editing the definitions for individual tasks is fine.
+
+- Pickle based flows are prone to breakage if the internals of Prefect or a
+  dependent library changes (even if the public-facing API remains the same).
+  Using a script based flow storage your flow is likely to work across a larger
+  range of Prefect/Python/dependency versions.
+
+The downside is you may have to do a bit more configuration to tell prefect
+where your script is located (since it can't always automatically inferred).
```
joshmeek, line 74:
```suggestion
where your script is located (since it can't always be automatically inferred).
```

joshmeek (approved):

PR #4111, opened 2021-02-12
AWS Policy information for ecs

kingsleyb:
## Summary
Documentation to show the AWS policies required on the task-role-arn and execution-role-arn to use the ECS agent with a temporary fargate cluster and using ECR for storage. 

## Checklist

This PR:


docs/orchestration/agents/ecs.md
```diff
@@ -131,6 +131,117 @@ prefect agent ecs start --task-role-arn my-task-role-arn
 Flows can override this agent default by passing the `task_role_arn` option to
 their respective [ECSRun](/orchestration/flow_config/run_configs.md#ecsrun) `run_config`.
 
+The following policy is needed for the task-role-arn if using a Temporary Fargate Cluster. This allows the
+DaskExecutor to create the temporary cluster. 
+
+```json
```
madkinsz, line 137: Is there somewhere we can store this in the code (or just alongside here?) and refer to it? Does that make sense? Seems like a big blob to have in the docs.

kingsleyb: I'm open to suggestions

mashun4ek (commented):
It's awesome! Thanks Kingsley for putting it all together! I would also add a brief description about task-role and execution role. For example, use execution role if you need to pull an image from ECR and enable logs in CloudWatch, and use task-role if your tasks need make AWS API calls, but it's up to you. LGTM. Not sure where to move IAM policy as @madkinsz suggested.

docs/orchestration/agents/ecs.md
jcrist: I think I'd rather link to the dask-cloudprovider docs for this. This is specific to users using a very specific tool that may change implementation details without us noticing (since we don't maintain it). They already provide this in their docs, and presumably this would get updated if things change: https://cloudprovider.dask.org/en/latest/aws.html#fargate.

I do think we should provide an example for something common, just so users get used to what a taskrole might entail. Perhaps one for accessing S3 storage from within a task (which should only be a few permissions).

cicdw: Could we possibly just delete this section on the Fargate Cluster and merge the rest?  This is a common question and it'd be nice to have some of the basics in our official docs

kingsleyb: Removed for now and follow Jim's suggestion in a different PR.

cicdw (approved):

PR #4115, opened 2021-02-12
Reduce number of boto3 clients created

### LINKED ISSUE #4085: S3Result causes a memory leak in DaskExecutor

jcrist:
Prefect relies on `boto3` for aws integration. Previously we were relying on creating a new session for parallel requests to avoid thread-safety issues - this worked, but was expensive (creating a new session is an order of magnitude slower). This was compounded by some faulty logic in the `S3Result` code that looked like it was supposed to result in shared boto3 clients, but in practice didn't.

Turns out boto3 clients are thread safe to use, the thread-safety issues only apply to instantiation (or if you're using the higher-level `boto3.resource` clients). We now use a single lock around boto3 client creation, and have a shared weakref cache of clients, allowing use across threads. This results in significantly fewer boto3 clients being created (possibly only one session per process), reducing memory usage and speeding up aws operations.

Fixes #4085, supersedes #4104.

### DIFF
--- a/src/prefect/utilities/aws.py
+++ b/src/prefect/utilities/aws.py
@@ -14,7 +14,7 @@ if TYPE_CHECKING:
     import botocore.client
 
 
-_CACHE = (
+_CLIENT_CACHE = (
     weakref.WeakValueDictionary()
 )  # type: MutableMapping[tuple, botocore.client.BaseClient]
 _LOCK = threading.Lock()
@@ -29,6 +29,10 @@ def get_boto_client(
 ) -> "botocore.client.BaseClient":
     """
     Utility function for loading boto3 client objects from a given set of credentials.
+
+    Note: this utility is threadsafe, and will cache _active_ clients to be
+    potentially reused by other concurrent callers, reducing the overhead of
+    client creation.
 
     Args:
         - resource (str): the name of the resource to retrieve a client for
@@ -97,8 +101,8 @@ def get_boto_client(
         )
 
         # If no extra kwargs and client already created, use the cached client
-        if not kwargs and cache_key in _CACHE:
-            return _CACHE[cache_key]
+        if not kwargs and cache_key in _CLIENT_CACHE:
+            return _CLIENT_CACHE[cache_key]
 
         if profile_name or botocore_session:
             session = boto3.session.Session(
@@ -119,6 +123,6 @@ def get_boto_client(
         )
         # Cache the client if no extra kwargs
         if not kwargs:
-            _CACHE[cache_key] = client
+            _CLIENT_CACHE[cache_key] = client
 
     return client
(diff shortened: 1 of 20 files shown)


src/prefect/utilities/aws.py
```diff
@@ -1,19 +1,32 @@
...
 
 import boto3
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Optional, MutableMapping
+
+if TYPE_CHECKING:
+    import botocore.client
+
+
+_CACHE = (
```
madkinsz, line 17: minor preference for `_CLIENT_CACHE` or `_CLIENTS`

src/prefect/utilities/aws.py
```diff
@@ -54,27 +70,55 @@ def get_boto_client(
...
 
-    if use_session:
-        # see https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?#multithreading-multiprocessing  # noqa
-        region_name = kwargs.pop("region_name", None)
+    # Boto3 clients _are_ threadsafe, but the creation of a client or a session
+    # isn't thread safe. Creating a client has a small cost (~5-10 ms), while a
+    # session has a larger cost (50-120 ms). Creating a new session also
+    # renegotiates auth - apparently AWS can throttle these requests if there
+    # are too many of them, so we really want to minimize session creation if
+    # we can.
```
madkinsz, line 78: Very good sleuthing !

src/prefect/utilities/aws.py
```diff
@@ -1,19 +1,32 @@
...
     resource: str,
     credentials: Optional[dict] = None,
-    use_session: Optional[bool] = False,
+    region_name: Optional[str] = None,
     profile_name: Optional[str] = None,
     **kwargs: Any
-) -> "boto3.client":
+) -> "botocore.client.BaseClient":
     """
     Utility function for loading boto3 client objects from a given set of credentials.
```
madkinsz, line 31: Is it worth noting in the docstring that we'll cache clients here?

madkinsz (approved):

PR #4116, opened 2021-02-12
Initial release of the FivetranSyncTask, to manage your Fivetran connector sync process.

crw:
## Summary
Initial release of the FivetranSyncTask, to manage your [Fivetran](https://fivetran.com) connector sync process.

## Changes
Adds a `fivetran` folder (package) to `tasks`
Adds a FivetranSyncTask, starts and monitors a Fivetran connector job. 

## Importance
This is Fivetran's first effort at integrating our APIs directly into a scheduling tool. We are very much looking to see if this helps with more complicated use cases that involved loading data through Fivetran. 

## Checklist

This PR:

Tests have not actually been added yet. We would love to work with the team on the best way to test this new Task.

### DIFF
--- a/src/prefect/tasks/fivetran/__init__.py
+++ b/src/prefect/tasks/fivetran/__init__.py
@@ -1,9 +1,3 @@
 """
 This module contains a task for starting and monitoring Fivetran connector sync jobs
 """
-try:
-    from prefect.tasks.fivetran.fivetran import FivetranSyncTask
-except ImportError as import_error:
-    raise ImportError(
-        'Using `prefect.tasks.fivetran` requires Prefect to be installed with the "fivetran" extra.'
-    ) from import_error
--- a/src/prefect/tasks/fivetran/__init__.py
+++ b/src/prefect/tasks/fivetran/__init__.py
@@ -1,0 +1,9 @@
+"""
+This module contains a task for starting and monitoring Fivetran connector sync jobs
+"""
+try:
+    from prefect.tasks.fivetran.fivetran import FivetranSyncTask
+except ImportError as import_error:
+    raise ImportError(
+        'Using `prefect.tasks.fivetran` requires Prefect to be installed with the "fivetran" extra.'
+    ) from import_error
--- a/src/prefect/tasks/fivetran/fivetran.py
+++ b/src/prefect/tasks/fivetran/fivetran.py
@@ -1,0 +1,148 @@
+import json
+import time
+
+import requests
+import pendulum
+
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class FivetranSyncTask(Task):
+    """
+    Task for running Fivetran connector sync jobs.
+
+    Args:
+        - connector_id (str, optional): Default connector id to use for sync jobs, if none is
+            specified to `run()`.
+        - **kwargs (Any, optional): additional kwargs to pass to the base Task constructor
+    """
+    def __init__(
+        self,
+        connector_id: str = None,
+        **kwargs
+    ):
+        self.connector_id = connector_id
+        super().__init__(**kwargs)
+
+
+    @defaults_from_attrs("connector_id")
+    def run(
+        self,
+        api_key: str,
+        api_secret: str,
+        connector_id: str = None,
+        poll_status_every_n_seconds: int = 15
+    ) -> dict:
+        """
+        Task run method for Fivetran connector syncs.
+
+        An invocation of `run` will attempt to start a sync job for the specified `connector_id`. `run` will
+        poll Fivetran for connector status, and will only complete when the sync has completed or when it
+        receives an error status code from an API call.
+
+        Args:
+            - api_key (str): `API key` per https://fivetran.com/account/settings; should be secret!
+            - api_secret (str): `API secret` per https://fivetran.com/account/settings; should be secret!
+            - connector_id (str, optional): if provided, will overwrite value provided at init.
+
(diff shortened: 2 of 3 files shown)


src/prefect/tasks/fivetran/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+This module contains a task for starting and monitoring Fivetran connector sync jobs
+"""
+try:
```
jlowin, line 4: Unless I'm mistaken, your task doesn't require any special dependencies, so it will actually be available automatically to every Prefect user! You could remove this `try/except` if you prefer.

crw: I see! I was a little unclear on the purpose of this section of code. I'll pull that out.

madkinsz:
Sweet! One of the benefits of using a `Task` subclass like this is that you can provide some helper methods to make the `run()` method easier to parse. I'd recommend splitting it up if you have the time so it's easier for users and future maintainers to understand.

crw:
I completely agree, this could use a refactor. The focus with this initial release is on getting a minimal viable product in front of users to see if more effort is indicated. Let me know if you would like to see those changes before we go forward. There are a few other changes I would like to make (particular around the request.Session settings and API error retries) as an early v2 to this code, so I was already planning some improvements for the near future.

madkinsz:
> I completely agree, this could use a refactor. The focus with this initial release is on getting a minimal viable product in front of users to see if more effort is indicated. Let me know if you would like to see those changes before we go forward. There are a few other changes I would like to make (particular around the request.Session settings and API error retries) as an early v2 to this code, so I was already planning some improvements for the near future.

That makes sense. That's more @jcrist's call than mine though.

src/prefect/tasks/fivetran/fivetran.py
```diff
@@ -0,0 +1,160 @@
...
+        URL_SETUP = "https://fivetran.com/dashboard/connectors/{}/{}/setup".format(
+            connector_details["service"], connector_details["schema"]
+        )
+        setup_state = connector_details["status"]["setup_state"]
+        if setup_state != "connected":
+            EXC_SETUP: str = (
+                'Fivetran connector "{}" not correctly configured, status: {}; '
+                + "please complete setup at {}"
+            )
+            raise Exception(EXC_SETUP.format(connector_id, setup_state, URL_SETUP))
```
cicdw, line 100: Because most exceptions inherit from the base `Exception` class, just to avoid a weird edge case we should probably choose a more specific exception here and down below - maybe a `ValueError` for now?

src/prefect/tasks/fivetran/fivetran.py
```diff
@@ -0,0 +1,160 @@
...
+        succeeded_at = parse_timestamp(connector_details["succeeded_at"])
+        failed_at = parse_timestamp(connector_details["failed_at"])
+        previous_completed_at = succeeded_at if succeeded_at > failed_at else failed_at
+        # URL for connector logs within the UI
+        self.logger.info(
+            "Connector type: {}, connector schema: {}".format(
+                connector_details["service"], connector_details["schema"]
+            )
+        )
+        self.logger.info("Connectors logs at {}".format(URL_LOGS))
```
cicdw, line 111: We can reserve this enhancement for v2, but as an FYI note that you can publish "link artifacts" that will get rendered in the UI in a special way; this `URL_LOGS` could be a good candidate for that.  Example code here that publishes a link artifact: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/prefect/flow_run.py#L191

cicdw (changes requested):
This looks great - thanks for the quick turnaround @crw and @PubChimps !

I have one request on the raised exception type, but otherwise LGTM.  Once merged, we can make sure that this gets added to the API reference docs 👍

src/prefect/tasks/fivetran/fivetran.py
crw: Ha, I was reading through the [list of built-in exceptions](https://docs.python.org/3/library/exceptions.html) and nothing really seemed to fit. I'll look again, but yeah maybe just ValueError.

src/prefect/tasks/fivetran/fivetran.py
crw: Are there docs for this? I am a little confused how this works just from this code sample.

crw: I agree the existing implementation is a little hokey. I'm slightly concerned about them from the POV that Fivetran's UI URLs aren't necessarily a user-facing, supported feature in the way I am using them; they could change without notice.

cicdw: Gotcha, that makes sense; our artifacts API is also [very lightly documented](https://docs.prefect.io/api/latest/artifacts/artifacts.html) at the moment.  For a link artifact, the UI produces a clickable link in a convenient place on the task run / flow run pages for each run.

I'd vote to hold off on this enhancement for now given what you've described 👍

cicdw (approved):

src/prefect/tasks/fivetran/fivetran.py
cicdw: Yea, maybe we could use a [Prefect FAIL signal](https://docs.prefect.io/api/latest/engine/signals.html#fail) which simply indicates failure without having to pick an underlying exception.

Happy to merge without that, or happy to make that change myself post-merge, whatever you'd like

PR #4139, opened 2021-02-18
Filehandling Tasks

sti0:
## Summary

Currently there are no standardized Prefect tasks for file handling. I like to contribute these tasks. 

- One to move files / directories to a new location. 
- One to unzip data
- One to zip one or more files / directories.

## Changes

This adds a new section `Filehandling` and starts up with three new tasks (`Move`, `Unzip`, `Zip`)

## Importance

I think this are common basic tasks for interacting with files in the file system.

## Checklist

This PR:


src/prefect/tasks/filehandling/operations.py
```diff
@@ -0,0 +1,89 @@
+from pathlib import Path
+from typing import Union, Any
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+from prefect.tasks.filehandling.base import FileBase
+
+
+class Copy(Task):
+    pass
```
madkinsz, line 9: Did you mean to remove this one? :)

sti0: Ups...this is another one to come in the future. Will remove it for this PR

src/prefect/tasks/filehandling/compression.py
```diff
@@ -0,0 +1,249 @@
...
+    Args:
+        - zip_file_path (Union[str, Path], optional): path to the zip file
+        - target_directory (Union[str, Path], optional): path to the target directory
+        - zip_password (str, optional): password for extraction. This must be utf-8 encoded.
+        - create_target_if_not_exists (bool, optional): Create the target directory if it
+            doesn't exist (default: False)
+        - use_filename_as_target_dir (bool, optional): Extracts data within a folder named
+            like the zip archive (without extension). If True, the basename of the zip archive will
+            be attached to the target_directory and may create a new folder. (default: False)
+        - remove_after_unzip (bool, optional): Removes the zip archive after extraction (default: False)
```
jcrist, line 29: There's a lot of options here, some of which seem unnecessary. Adding flexibility is good when useful, but can also confuse users and increases our maintenance burden. If I were writing this I'd go with:

```python
"""
- source (Union[str, Path], optional): the path to the zip file
- target_dir (Union[str, Path], optional): directory to extract the zip file into.
    If not provided, the current working directory will be used. This directory
    must already exist.
"""

The `run` method would support all of the above, as well as

```python
"""
- password (str, optional): password for unzipping a password-protected zip file.
"""
```

Supporting `password` in the run method alone better encourages users to store passwords in secrets rather than hardcoding the parameter in the constructor.

This design better matches the functionality of the `unzip` shell command (or `shutil.unpack_archive`), both existing interfaces a user is likely already familiar with. The removed options would better be handled by other tasks, or can be inferred automatically from the parameters above.

src/prefect/tasks/filehandling/compression.py
```diff
@@ -0,0 +1,249 @@
...
+                    => basename of the source file without extension
+                2. if the first object in the source_path is a directory
+                    => parent name of the directory
+        - create_target_if_not_exists (bool, optional): Create the target directory if it
+            doesn't exist (default: False)
+        - compression: (int, optional): pass in a supported compression method
+            https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile-objects
+        - compression_level(int, optional): pass in a supported compression level
+            https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile-objects
+            Only supported on Python >= 3.7
```
jcrist, line 139: Same here. This is a lot of options, some of which I find confusing (`zip_file_name` and `target_directory` in particular). I'd write this as (in both the constructor and `run` method):

```python
"""
- source (Union[str, Path, List[str], List[Path]], optional): a path or paths to compress
     into a single zip archive.
- target (Union[str, Path], optional): path to the output archive file. Any parent directories
    of `target` must already exist.
- compression_method (str, optional): the compression method to use. Options are
    "deflate", "store", "bzip2", and "lzma". Defaults to `"deflate"`.
- compression_level (int, optional). Compression level to use,
    see https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile for more info.
"""
```

A few notes:
- We don't support inferring the directory name from the inputs. Inference is unlikely to satisfy most users, and doesn't really make sense for archives created from a list of paths. If a user wants to determine the archive path from the input, they can write a short `@task` decorated task to do this themselves.
- All directories must already exist. Again, an `@task` decorated task can do this external to this task if needed.
- We use string codes instead of integers for `compression_method`. This is much friendlier than the stdlib int-based interface.
- We default to `deflate` instead of `store`. This better matches the existing `zip` shell function (which compresses by default).

sti0:
@jcrist thanks for your suggestions. I made some changes and removed the parameters according to your suggestion. I thought to big. It's better to break down the functionality into smaller pieces (like you said).

Looking forward to here from you.

abkfenris:
If you're going to decompose the tasks into smaller functions, how about using [fsspec `OpenFiles`](https://filesystem-spec.readthedocs.io/en/latest/index.html) as the core file type? 

That will give you serialisability for passing files between tasks, and the ability for the same functions to work against many different types of file systems. Then those functions can be wrapped up in specific file system tasks that present a simpler interface to library users, while allowing advanced users to build upon them (and giving them a template of how to get started doing so).

jcrist (approved):

jcrist:
Thanks!

Issue #4114, opened 2021-02-12
Server healthchecks create curl zombies


jschnare:
## Description
When I run the local server I get zombie curl processes, about 3 per second.  They get cleaned up when I stop the server.

I suspect that it's the health checks added in #4041.

#2418 refers to running the agent under tini, which should fix the server containers as well.

## Expected Behavior
Containers shouldn't generate zombie processes.  Given enough time they'll fill the host's process list.

## Reproduction
```bash
$ prefect backend server`
$ prefect server start --detach
$ while true; do echo -ne "$SECONDS\t"; pgrep -u root curl | wc -l; sleep 3; done
61802	47
61805	53
61808	58
61811	63
61814	69
61817	73
61821	79
61824	85
61827	89
61830	95
61833	99
```

## Environment

```bash
$ docker -v
Docker version 20.10.3, build 48d30b5
$ docker-compose -v
docker-compose version 1.28.2, build 67630359
```
```json
{
  "config_overrides": {
    "context": {
      "secrets": false
    },
    "server": {
      "telemetry": {
        "enabled": true
      }
    }
  },
  "env_vars": [],
  "system_information": {
    "platform": "Linux-5.4.0-58-generic-x86_64-with-debian-bullseye-sid",
    "prefect_backend": "server",
    "prefect_version": "0.14.8",
    "python_version": "3.7.9"
  }
}
```

madkinsz:
Thanks for the thorough report @jschnare!

tothandor:
I can confirm that. Host falls over withhin a few hours.

madkinsz:
We are working on a fix for this, it'll be in the next release.

madkinsz:
Hey @jschnare -- could you give https://github.com/PrefectHQ/prefect/pull/4142 a quick test? I (weirdly) cannot reproduce this on my dev machine. If not I will try on another machine. Thanks!

asmhack:
Have the same situation. 
prefect version: 0.14.9 
GCP 4.19.0-14-cloud-amd64 1 SMP Debian 4.19.171-2 (2021-01-30) x86_64 GNU/Linux

```
CONTAINER ID   NAME                       CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
a1468dabae7f   cli_ui_1                   0.00%     4.863MiB / 18.87GiB   0.03%     1.13kB / 0B       4.1kB / 4.1kB     5
[...]

PR #4142, opened 2021-02-19
Add init to apollo/graphql deployment to reap healthcheck zombies

madkinsz:
## Summary

Reaps the zombies introduced by #4041 

Alternative to https://github.com/PrefectHQ/server/pull/191 which installs tini into the server images.

Per the [tini docs](https://github.com/krallin/tini#using-tini)
> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just pass the --init flag to docker run.

Simply adding the `init` flag should solve this without modifying `Server` images.

## Changes

- Adds `init: true` to apollo/graphql images

## Importance

Fixes https://github.com/PrefectHQ/prefect/issues/4114

## Checklist

This PR:

### DIFF
--- a/changes/pr4142.yaml
+++ b/changes/pr4142.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Add init to apollo/graphql deployment to reap healthcheck zombies - [#4142](https://github.com/PrefectHQ/prefect/pull/4142)"
--- a/src/prefect/cli/docker-compose.yml
+++ b/src/prefect/cli/docker-compose.yml
@@ -57,6 +57,7 @@ services:
   # GraphQL: the server's business logic that exposes GraphQL mutations
   graphql:
     image: "prefecthq/server:${PREFECT_SERVER_TAG:-latest}"
+    init: true
     ports:
       - "${GRAPHQL_HOST_PORT:-4201}:4201"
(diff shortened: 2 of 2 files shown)


jcrist:
This will work for docker-compose, but not those deploying on docker swarm or other platforms. We really should be running our containers with an init process everywhere, relying on an external tini install feels less reliable to me than just packaging it as part of our images.

TylerWanner:
yea along @jcrist's point, I think we should include tini in the image itself, which will also provide a better k8s runtime experience, rather than doing this at the docker-compose layer

madkinsz:
> This will work for local docker-compose, but not those deploying on docker swarm or other platforms. We really should be running our containers with an init process everywhere, relying on an external tini install feels less reliable to me than just packaging it as part of our images.

This is just such a minimal change compared to the weird github install we're forced to do on our apollo image and doesn't require a Server release to patch bug. I'm also not excited that the apollo image is broken on my development machine with the Server fix. I'm okay with either solution though, I agree having it installed in the images makes the most sense in the long run.

madkinsz:
Closing in favor of the PR in Server -- this is a functional solution for anyone who needs a quick patch though.

PR #4150, opened 2021-02-22
Dynamically load compression libraries for serializers

madkinsz:
Not all users have them installed and it breaks the import of core if they are imported at the top level.

## Summary

If a user does not have a compression library required by the new `CompressedSerializer` (#4063) then `prefect` cannot be imported without exceptions. To resolve this, we 

## Changes

Instead of importing all the libraries at module load, we import the library requested for compression as needed.

## Importance

This reduces assumptions about the user's python environment.

## Checklist

This PR:

### DIFF
--- a/changes/pr4150.yaml
+++ b/changes/pr4150.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Dynamically import compression libraries for `CompressedSerializer` - [#4150](https://github.com/PrefectHQ/prefect/pull/4150)"
--- a/src/prefect/engine/serializers.py
+++ b/src/prefect/engine/serializers.py
@@ -1,11 +1,9 @@
 import base64
-import bz2
-import gzip
 import io
 import json
-import lzma
+import importlib
+
 from typing import TYPE_CHECKING, Any, Callable, Dict, Tuple
-import zlib
 
 import cloudpickle
 import pendulum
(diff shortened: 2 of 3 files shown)


madkinsz:
Needs an update to tests but I want to nail down the implementation first as dynamically importing like this is a bit tricky to get right. If ya'll could take a look @jcrist @joshmeek 

I _need_ these changes to develop locally and can't seem to get lzma built into my python. I imagine that some users will have a similar problem.

jcrist (commented):
Hmmm, it wasn't apparent from the python docs that these could ever be absent (the python docs don't list them as optional, only the cpython build docs). In practice this hasn't bit us in other libraries, so I suspect your setup is a bit odd compared to most. Still, the fix here seems fine provided tests pass.

madkinsz:
> Hmmm, it wasn't apparent from the python docs that these could ever be absent (the python docs don't list them as optional, only the cpython build docs). In practice this hasn't bit us in other libraries, so I suspect your setup is a bit odd compared to most. Still, the fix here seems fine provided tests pass.

Yeah it is a little weird. I think it's most frequently encountered while building python on non-unix systems where it can struggle to find the compression libs to link to.

There's a brief issue at https://bugs.python.org/issue34895 and something like https://github.com/pandas-dev/pandas/issues/27575 shows that it's not _too_ obscure.

jcrist (approved):

PR #4157, opened 2021-02-23
EmailTask: accept msg_plain, email_to_cc and email_to_bcc

joelluijmes:
## Summary

Expends possibilities of `EmailTask` by accepting the following parameters: msg_plain, email_to_cc and email_to_bcc
- `msg_plain`: allows to add plain text version of email
- `email_to_cc`: additional addresses to put in CC
- `email_to_bcc`: additional addresses to put in BCC

## Changes

Additional parameters.

## Importance

More flexibility :) 

## Checklist

This PR:


src/prefect/tasks/notifications/email_task.py
```diff
@@ -90,25 +107,39 @@ def run(
...
-        contents["From"] = email_from
-        contents["To"] = email_to
+        # https://docs.python.org/3/library/email.examples.html
+        # Add the html version. This converts the message into a multipart/alternative
+        # container, with the original text message as the first part and the new html
+        # message as the second part.
+        if msg_plain:
+            message.set_content(msg_plain, subtype="plain")
+        if msg:
+            message.add_alternative(msg, subtype="html")
```
madkinsz, line 140: Sweet :D

madkinsz: Starting with "Add the html version" then setting the plain message could be a little confusing

madkinsz (commented):
Hey @joelluijmes, this is a nice upgrade for this task! Can you confirm that switching to this different email API isn't going to break anything for existing users?

joelluijmes:
> Hey @joelluijmes, this is a nice upgrade for this task! Can you confirm that switching to this different email API isn't going to break anything for existing users?

Yes should be backwards compatible. As in, didn’t rename the param or changed the order :)

src/prefect/tasks/notifications/email_task.py
joelluijmes: Hm, yea agreed, kinda confusing now. I just copied that text, the order does actually matter. You cant first use add_alternative followed by set_content. 

Ill update it tomorrow!

madkinsz:
> > Hey @joelluijmes, this is a nice upgrade for this task! Can you confirm that switching to this different email API isn't going to break anything for existing users?
> 
> Yes should be backwards compatible. As in, didn’t rename the param or changed the order :)

My concern is that the appearance of people's emails will change since the content is being specified differently now and I don't want to break anyone's flows using email tasks.

joelluijmes:
> My concern is that the appearance of people's emails will change since the content is being specified differently now and I don't want to break anyone's flows using email tasks.

Understandably, but fortunately this doesn't change anything. Only specifiying the `msg` (`add_alternative`) with html yields a html email 👍

madkinsz (dismissed):
If you merge `master` in you'll fix the codecov error. You'll need to run `black` on whatever file it's complaining about. Otherwise this LGTM, I like the rephrased comment 👍

madkinsz (approved):

jcrist (approved):

PR #4158, opened 2021-02-23
Pre-fetch runs to minimize agent submission delays

jlowin:
## Summary

Minimize latency by pre-fetching flow runs to start them as close to their scheduled start times as possible

## Changes

This has been kicking around in my head for a long time, so I wanted to present it for comment. 

Currently, the agent polls for flow runs whose `scheduled_start_time` is before `now()`, and submits them for execution. This means that there can be up to a 10-second delay in starting a run because of loop interval timing. For example, if the agent runs at 9:59:59 and sleeps 10 seconds, a 10:00 run won't be queried or submitted until 10:00:09. (Note: this is a worst-case scenario, as when runs are available loop times can be as short as 0.25 seconds. However, for people running hourly work, the agent will have probably backed off to its maximum 10-second interval)

This PR makes a minor change: it pre-fetches runs whose `scheduled_start_time` is before `now() + 10 seconds`, and adds a brief `time.sleep()` when submitting them for execution. This means that if an agent runs at 9:59:59, it will retrieve the 10:00 run and submit it immediately with a 1-second delay. The user experience is improved because the loop-induced delay is eliminated.

## Checklist

This PR:


madkinsz (commented):
This makes a lot of sense to me. It seems like there will be increased risk of race conditions, I presume these are caught by the `update_state` try/except in `deploy_and_update_flow_run` ?

jlowin:
@madkinsz that try / except will help when users are running competing agents with version locking enabled, and the primary mechanism (with a single matched agent) is the `submitting_flow_runs` set that prevents duplicate submission.

madkinsz:
This makes sense to me, do we want to get this released? cc @cicdw @zangell44

zangell44 (approved):
Looks good to me

cicdw (approved):
LGTM too

Issue #4151, opened 2021-02-22
prefect server start should offer to dump resulting docker-compose.yml and environment for standalone deployment


sm-Fifteen:
## Current behavior
`prefect server start` is currently just a wrapper around a few docker and docker-compose functions to create a custom docker-compose.yml based on [a reference file](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/docker-compose.yml) and run it. While this is convenient for local testing, it can make actual core server deployment a bit cumbersome. For instance, even though the entirety of the server is going to be running inside docker containers, one still needs to have python3 installed and either

* run `sudo pip install prefect` to install prefect globally (which pip even warns that you shouldn't do that) with a prefect config in the /root directory
* use a virtual environment and override `VIRTUAL_ENV` and `PREFECT__USER_CONFIG_PATH` when running `prefect server start` as a service
* use `pip install --user prefect` to install prefect in a user's home and have the service run `prefect server start` as that user, given that user is part of the `docker` group

In all cases, all of this is merely so a python program can slice together a YAML file at `/tmp/docker-compose.yml` and create a few environment variables before basically running `docker-compose up` on it. This isn't so great for standalone deployments.

I got the idea of putting together my own `docker-compose.yml` from [flavienbwk/prefect-docker-compose](https://github.com/flavienbwk/prefect-docker-compose), although going back and forth between [prefect/cli/docker-compose.yml](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/docker-compose.yml) and [prefect/cli/server.py](https://github.com/PrefectHQ/prefect/blob/559330bb549b9c92c413948b615509b6c6d33620/src/prefect/cli/server.py) to figure out what each one was doing and what each one needed was rather tedious.

## Proposed behavior

It seems to me like a pure docker-compose deployment is something that users of prefect core would have a number of reasons to be interested in (easier to compose with additional services, lots of environment variables to currently manage, avoids "polluting" the host machine, etc.) and have indeed shown interest towards in the past (PrefectHQ/prefect#2715, PrefectHQ/prefect#3557, among others). Given how `prefect server start` is already just a wrapper for that, I could see a `prefect server dump` or `prefect server generate` command taking the same options as `prefect server start`, but creating a `docker-compose.yml` and a `.env` file in the current directory (or anywhere else) instead, so that the same configuration can be launched later without having the prefect  client or even python installed.

## Additional information
[...]

PR #4176, opened 2021-02-25
Add `prefect server config` command to output a compose file

madkinsz:
## Summary

Instead of starting the server, provide the configured file for a user to spin it up themselves.

`prefect server config` writes a configure docker-compose yaml to standard output using `docker-compose config` to resolve environment variables. It takes all the same (non pull/start) options as `prefect server start`

## Changes

- Adds `config` to `prefect server` CLI
- Refactors docker-compose file and environment generation
- Warns on collisions between the current environment and CLI commands

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/4151

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -210,15 +210,15 @@ def setup_compose_file(
 
 
 def setup_compose_env(
-    version,
-    ui_version,
-    no_upgrade,
-    postgres_port,
-    hasura_port,
-    graphql_port,
-    ui_port,
-    server_port,
-    volume_path,
+    version=None,
+    ui_version=None,
+    no_upgrade=None,
+    postgres_port=None,
+    hasura_port=None,
+    graphql_port=None,
+    ui_port=None,
+    server_port=None,
+    volume_path=None,
 ):
     # Pull current version information
     base_version = prefect.__version__.split("+")
@@ -292,14 +292,15 @@ def setup_compose_env(
             "environment variable value (={env['PREFECT_SERVER_DB_CMD']}) will be "
             "ignored."
         )
-    else:
-        # Allow a more complex database command to be passed via env
-        env.setdefault(
-            "PREFECT_SERVER_DB_CMD",
-            "prefect-server database upgrade -y"
-            if not no_upgrade
-            else "echo 'DATABASE MIGRATIONS SKIPPED'",
-        )
+        env.pop("PREFECT_SERVER_DB_CMD")
+
+    # Allow a more complex database command to be passed via env
+    env.setdefault(
+        "PREFECT_SERVER_DB_CMD",
+        "prefect-server database upgrade -y"
+        if not no_upgrade
+        else "echo 'DATABASE MIGRATIONS SKIPPED'",
+    )
 
     env.update(**PREFECT_ENV, **APOLLO_ENV, **POSTGRES_ENV, **UI_ENV, **HASURA_ENV)
 
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -13,7 +13,7 @@ from prefect import config
 from prefect.utilities.configuration import set_temporary_config
 
 
-def make_env(fname=None):
+def make_env(include_local: bool = True):
     # replace localhost with postgres to use docker-compose dns
     PREFECT_ENV = dict(
         DB_CONNECTION_URL=config.server.database.connection_url.replace(
@@ -54,15 +54,9 @@ def make_env(fname=None):
 
     HASURA_ENV = dict(HASURA_HOST_PORT=config.server.hasura.host_port)
 
-    ENV = os.environ.copy()
+    ENV = os.environ.copy() if include_local else {}
     ENV.update(**PREFECT_ENV, **APOLLO_ENV, **POSTGRES_ENV, **UI_ENV, **HASURA_ENV)
 
-    if fname is not None:
-        list_of_pairs = [
-            f"{k}={v!r}" if "\n" in v else f"{k}={v}" for k, v in ENV.items()
-        ]
-        with open(fname, "w") as f:
-            f.write("\n".join(list_of_pairs))
     return ENV.copy()
 
 
@@ -127,6 +121,15 @@ def server():
     "-d",
     help="Detached mode. Runs Server containers in the background",
     is_flag=True,
+    hidden=True,
+)
+@click.option(
+    "--dump",
+    help=(
+        "Instead of running the Server, dump the generated docker-compose.yml file and "
+        ".env file to the given directory"
+    ),
+    type=click.Path(exists=True, file_okay=False, writable=True),
     hidden=True,
 )
 @click.option(
@@ -211,6 +214,7 @@ def start(
     no_upgrade,
     no_ui,
     detach,
+    dump,
     postgres_port,
     hasura_port,
     graphql_port,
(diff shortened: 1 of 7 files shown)


sm-Fifteen:
`prefect agent local` has a similar feature, `local install`, which takes (most of) the same parameters as `local start` but generates a `supervisord.conf` file. I would argue that making `--dump` a subcommand instead of an option would be better, as far as CLI uniformity is concerned, though being able to specify a target directory is a nice plus.

Aside from that, this all looks great.

src/prefect/cli/server.py
```diff
@@ -340,6 +348,20 @@ def start(
...
     env.update(PREFECT_CORE_VERSION=prefect.__version__)
 
+    if dump:
+        dump_dir = Path(dump)
+        Path(compose_dir_path).joinpath("docker-compose.yml").rename(
+            dump_dir.joinpath("docker-compose.yml")
+        )
+        with dump_dir.joinpath(".env").open("w") as fout:
+            env_vars = [
+                f"{k}={v!r}" if "\n" in v else f"{k}={v}" for k, v in env.items()
```
sm-Fifteen, line 358: I don't believe docker supports multiline strings at all in its environment file, quoted or not. The doc says it [doesn't treat quotes in any special way](https://docs.docker.com/compose/env-file/), so I doubt using `repr()` on the string would help.

madkinsz: Oh that's good to know! I just pulled that code from above presuming you'd `source .env` then run `docker-compose up`. Seems kind of silly that they auto-pull from a `.env` file with their own opinions about syntax but 🤷

madkinsz:
Yeah I could be convinced to do a sub-command instead. I just didn't want to replicate all the code in `prefect server start` or refactor it all to be in a separate function. Thoughts @jcrist ?

jcrist:
I think a separate command makes sense. I'd like to transition `prefect agent local install`/`prefect agent kubernetes install` to commands with more specific names (`install` indicates that it installs something but it actually doesn't, it just outputs a file). Would be good to start with a good name here, perhaps `prefect server gen-compose`, or something similar?

src/prefect/cli/server.py
sm-Fifteen: I'm actually running ` sudo /usr/local/bin/docker-compose --env-file=/opt/python-prefect/docker-compose-env.env up -d` instead of leaving it up to docker to automatically load a hidden file.

And as for the dotenv file syntax, yeah, quoting and multiline is [all over the place](https://github.com/theskumar/python-dotenv/issues/170), that's what tends to happen with loosely-defined formats.

jcrist:
`persist-config` sounds like it's writing a file somewhere. I think this should output text, and can be piped to a file as needed (or just piped into `docker-compose directly). Config also doesn't really indicate what the "config" file is, whereas something with `compose` in the name better indicates it's a docker-compose setup. No strong thoughts on a name beyond that.

sm-Fifteen:
> `persist-config` sounds like it's writing a file somewhere. I think this should output text, and can be piped to a file as needed (or just piped into `docker-compose directly). Config also doesn't really indicate what the "config" file is, whereas something with `compose` in the name better indicates it's a docker-compose setup. No strong thoughts on a name beyond that.

That could work if this didn't need writing content to two files to work, the generated docker-compose.yml file won't work without a dozen mandatory env variables plucked from Prefect's config.toml, which need to be saved to a separate file for docker-compose to load alongside it.

Though I guess one could always tell people to do something like `prefect server gen-compose 1> ./docker-compose.yml 3> ./.env` or something, it's not like there's a limit to the amount of output streams a program is allowed to have.

jcrist:
> the generated docker-compose.yml file won't work without a dozen mandatory env variables plucked from Prefect's config.toml, which need to be saved to a separate file for docker-compose to load alongside it.

Oh, I hadn't realized that. Why not write the values directly in the generated compose file? I was viewing this as effectively "freezing" the config for later use. A single generated file feels simpler to work with to me than having to move around two files with the proper paths.

madkinsz:
I think it makes sense to inject the variables into the file as well. I'm honestly quite confused at the way the docker-compose is written right now where there are `${VAR-default}` everywhere but the defaults are actually determined by the click CLI option defaults.

jcrist:
Yeah, I think that might just be historical and could all be removed/redone. Feel free to change the compose setup however you want to make things simpler.

src/prefect/cli/server.py
madkinsz: Wow that issue is wild. Anyway, I'll make sure to test that it loads from the file correctly :)

madkinsz:
It turns out there's a docker-compose command for this! `docker-compose config` so I've named this to match: `prefect server config`

This means that I do not have to touch the `docker-compose` file. We may want to do this eventually/later but this makes this change less likely to break things for users.

src/prefect/utilities/compatibility.py
```diff
@@ -21,3 +20,17 @@ def nullcontext() -> Iterator[None]:
...
+
+def Call(call):
+    # Takes a `unittest.mock.call` and adds args/kwargs properties
+
+    if sys.version_info < (3, 8):
+        # Properties were added in 3.8
+        call.args = call[1]
+        call.kwargs = call[2]
+
+    return call
```
madkinsz, line 36: I am not enthused about this but it 
- feels clunky no matter how we handle this
- it's only used in tests
- I think it improves test readability at the expense of a more confusing util

🤷

madkinsz:
Still needs a test for `prefect server config` -- I'm not sure how I want to test it yet. Any tests other than "this command runs" would basically replicate `prefect server start` tests unless we want to actually call "docker-compose config". I'm not sure it's unreasonable to do the same thing as the server start setup/teardown test to ensure variables are passed correctly

src/prefect/utilities/compatibility.py
```diff
@@ -21,3 +20,18 @@ def nullcontext() -> Iterator[None]:
 else:
 
     from contextlib import nullcontext  # noqa: F401
+
+
+# Provide 3.6/3.7 `call` with `.args` and `.kwargs`
+
+
+def Call(call):  # type: ignore
```
jcrist, line 28: Since this is only used in a helper function in a few tests in one file, it doesn't really feel worth it to me to pull it out into a separate file. Honestly, I'd just keep indexing the `call_args` tuple rather than making a helper like this (we do the indexing everywhere in our tests, consistency matters). When we drop support for older Python versions we can use the attribute accessors.

jcrist (commented):
One nit, otherwise this looks good to me.

Can you also add the `config` command to the CLI docs here: https://github.com/PrefectHQ/prefect/blob/641287898a680afd9f02884f0863f6cdf9640ac4/docs/outline.toml#L74

jcrist:
> I'm not sure it's unreasonable to do the same thing as the server start setup/teardown test to ensure variables are passed correctly

I think doing a test with no flags (things run fine by default) and one with lots of flags/env-vars (test args get forwarded properly) would be sufficient. By inspection the same helpers are used, so if the actual env/flag handling logic is tested elsewhere already we only need to ensure things are hooked together properly.

madkinsz:
Docs look good: https://deploy-preview-4176--prefect-docs.netlify.app/api/latest/cli/server.html#config

Messed up a test... should be an easy fix.

jcrist (approved):

Issue #3789, opened 2020-12-09
Allow naming tasks produced by `nout`


bnaul:
The `nout` flag that @jcrist added in #3697 is a great new feature! But one thing that hampers its utility a little bit is the fact that every resulting task has the name `GetItem`; this means that for a flow that makes liberal use of this new feature, the Cloud UI is nearly useless as it's impossible to distinguish one task from another.
![image](https://user-images.githubusercontent.com/903655/101665448-98a98180-3a12-11eb-8ffa-6cceaee98036.png)

I'm guessing @jcrist or @cicdw might have a more general idea, but a couple of possible solutions could be:
a) rename `nout` (`out` ?) and accept either an int (like the current behavior) or an iterable of labels (the length would implicitly provide the number of tasks)
b) add a new optional arg just to label the tasks

Curious to hear people's thoughts; I can see an argument for keeping things simpler, but at least for me the lack of names will definitely keep me from using this exciting cool new feature which is too bad 🙂 .

jcrist:
What kind of names would you want? What if we auto-named the tasks, but gave them a better name? Perhaps take the name of the generating task plus an index into it - like `my_task[0]`.

madkinsz:
> What kind of names would you want? What if we auto-named the tasks, but gave them a better name? Perhaps take the name of the generating task plus an index into it - like `my_task[0]`, `my_task[1]`.

Adding the indexing makes sense as a first step. I'm not sure if we want to parse it but we could maybe inspect `NamedTuple` which is kind of a pain

```
>>> from typing import NamedTuple
>>> 
>>> def foo() -> NamedTuple('FooOut', [('a', int), ('b', int), ('c', int)]):
...         return 1, 2, 3
... 
>>> print(foo.__annotations__['return']._fields)
('a', 'b', 'c')
```

bnaul:
> What kind of names would you want? What if we auto-named the tasks, but gave them a better name? Perhaps take the name of the generating task plus an index into it - like `my_task[0]`, `my_task[1]`.

that would be a good start! ideally I'd just want to use the same var names I'm using at unpacking time, just like your example has `double, triple = double_and_triple(inc)`; `double_and_triple[0]` is better than `GetItem` but not as good as `double` 🙂 

> we could maybe inspect NamedTuple which is kind of a pain

agree it looks gross but I personally would definitely use this in order to get a better UI experience....clean Cloud UI + messy code is a very worthwhile tradeoff for my use case

[...]

PR #4183, opened 2021-02-26
Improve indexed task naming

JacobHayes:
## Summary

Use the parent task name + key for the `GetItem` task returned by indexing.

First pass at fixing #3789 with just task + key naming (didn't try `NamedTuple`)

## Importance

`GetItem` tasks are noisy and hard to follow in the UI schematic, this makes them a bit easier to work with.

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -953,7 +953,7 @@ class Task(metaclass=TaskMetaclass):
             - Task
         """
         return prefect.tasks.core.operators.GetItem(
-            checkpoint=self.checkpoint, name=f"{self.name}[{key}]", result=self.result
+            checkpoint=self.checkpoint, name=f"{self.name}[{key!r}]", result=self.result
         ).bind(self, key)
 
     def __or__(self, other: object) -> object:
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -953,7 +953,7 @@ class Task(metaclass=TaskMetaclass):
             - Task
         """
         return prefect.tasks.core.operators.GetItem(
-            checkpoint=self.checkpoint, result=self.result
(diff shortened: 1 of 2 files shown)


src/prefect/core/task.py
```diff
@@ -953,7 +953,7 @@ def __getitem__(self, key: Any) -> "Task":
             - Task
         """
         return prefect.tasks.core.operators.GetItem(
-            checkpoint=self.checkpoint, result=self.result
+            checkpoint=self.checkpoint, name=f"{self.name}[{key}]", result=self.result
```
jcrist, line 956: Suggestion to use `repr` instead of `str` for the key, so that string-based indices still display with their quotes.

```suggestion
            checkpoint=self.checkpoint, name=f"{self.name}[{key!r}]", result=self.result
```

jcrist (commented):
This seems like a great idea. Can you add a test as well? It'd go in `tests/core/test_task_operators.py`.

tests/core/test_task_operators.py
```diff
@@ -85,6 +85,15 @@ def test_getitem_preserves_result_info(self):
...
         assert y.result.dir.endswith("home")
 
+    def test_getitem_name(self):
+        with Flow(name="test") as f:
+            x = Parameter("x")[0]
+            assert x.name == "x[0]"
+            y = Parameter("y")["a"]
+            assert y.name == "y['a']"
+            z = Parameter("z")[Parameter("a")]
+            assert z.name == "z[<Parameter: a>]"
```
jcrist, line 95: Hmmm, this is a good callout, I think we could do better here. Perhaps?

```python
if isinstance(key, Task):
    name = f"{name}[{key.name}]"
else:
    name = f"{name}[{key!r}]"
```

Use the name of the task (as a non-repr) if the key is a task, otherwise repr the key as a constant.

PR #4191, opened 2021-03-02
Add new docs page on docker images

jcrist:
Adds a new docs page on docker images and prefect. Covers dependency requirements, different options for building images, and some recommendations on which option to choose.


cicdw (approved):
This is top notch, nicely done

PR #4195, opened 2021-03-03
Pass S3 client option from storage to result

davzucky:
## Summary
The initialization of storage.s3 initialize as well the S3Result property for S3. however the client_options dict is not passed to the result boto3_kwargs. 

## Changes

Pass the property to the S3Result as below
```
result = S3Result(bucket=bucket, boto3_kwargs=client_options)
```

## Importance

This is important to ensure the storage and the result are initialized with the same param when calling S3

## Checklist

This PR:

### DIFF
--- a/changes/pr4195.yaml
+++ b/changes/pr4195.yaml
@@ -1,5 +1,5 @@
 fix:
-  - "Ensure the S3 storage properly initialize the S3Result with client_option"
+  - "Forward `client_options` to `S3Result` from `S3` storage - [#4195](https://github.com/PrefectHQ/prefect/pull/4195"
 
 
 contributor:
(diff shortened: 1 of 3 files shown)


changes/pr4195.yaml
```diff
@@ -0,0 +1,6 @@
+fix:
+  - "Ensure the S3 storage properly initialize the S3Result with client_option"
```
jcrist, line 2:
```suggestion
  - "Forward `client_options` to `S3Result` from `S3` storage - [#4195](https://github.com/PrefectHQ/prefect/pull/4195"
```

jcrist (approved):
Thanks!

PR #4202, opened 2021-03-03
Copy and Remove

sti0:
## Summary

This introduces two new tasks to the `files` section.

## Changes

 Adds the tasks `Copy` and `Remove` 

## Importance

Doesn't exists...

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/files/operations.py
+++ b/src/prefect/tasks/files/operations.py
@@ -1,6 +1,6 @@
 import shutil
 from pathlib import Path
-from typing import Union, Any
+from typing import Any, Union
 
 from prefect import Task
 from prefect.utilities.tasks import defaults_from_attrs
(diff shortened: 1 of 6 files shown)


src/prefect/tasks/files/operations.py
```diff
@@ -64,3 +64,110 @@ def run(
...
+    Args:
+        - remove_path (Union[str, Path], optional): file or directory to be removed
+            If deleting a directory, the directory must be empty.
+        - **kwargs (dict, optional): additional keyword arguments to pass to the
+            Task constructor
+    """
+
+    def __init__(
+        self,
+        remove_path: Union[str, Path] = "",
```
jcrist, line 144: Since `Remove` only takes one path, I think the `remove` prefix is unnecessary. Could just call it `path`.

jcrist (commented):
One small comment, otherwise this looks good to me. Thanks!

src/prefect/tasks/files/operations.py
sti0: Done

jcrist (approved):
Thanks!

PR #4211, opened 2021-03-04
A few ECS agent cleanups

jcrist:
- Move `command` and `environment` for the `flow` container to be set at ECS task run time rather than as part of the task definition. This makes it simpler to provide a `task_definition_arn`, as prefect can set more of the required parameters for you.
- Improve our docs a bit on what's required for a custom task definition.

### DIFF
--- a/src/prefect/agent/ecs/agent.py
+++ b/src/prefect/agent/ecs/agent.py
@@ -1,6 +1,6 @@
 import os
 from copy import deepcopy
-from typing import Iterable, Dict, Optional, Any
+from typing import Iterable, Dict, Any
 
 import slugify
 import yaml
@@ -192,9 +192,6 @@ class ECSAgent(Agent):
         )  # type: Dict[str, Any]
 
         self.ecs_client = get_boto_client("ecs", **self.boto_kwargs)
-        self.rgtag_client = get_boto_client(
-            "resourcegroupstaggingapi", **self.boto_kwargs
-        )
 
         # Load default task definition
         if not task_definition_path:
@@ -302,8 +299,7 @@ class ECSAgent(Agent):
         run_config = self._get_run_config(flow_run, ECSRun)
         assert isinstance(run_config, ECSRun)  # mypy
 
-        taskdef_arn = self.get_task_definition_arn(flow_run, run_config)
-        if taskdef_arn is None:
+        if run_config.task_definition_arn is None:
             # Register a new task definition
             self.logger.debug(
                 "Registering new task definition for flow %s", flow_run.flow.id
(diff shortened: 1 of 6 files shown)


src/prefect/agent/ecs/agent.py
```diff
@@ -508,16 +497,22 @@ def get_run_task_kwargs(
             container = {"name": "flow"}
             container_overrides.append(container)
 
+        # Set flow run command
+        container["command"] = ["/bin/sh", "-c", get_flow_run_command(flow_run)]
```
madkinsz, line 501: Should this be 
```suggestion
        # Set flow run command
        container.setdefault("command", ["/bin/sh", "-c", get_flow_run_command(flow_run)])
```

jcrist: We don't let users in other execution environments override this, I don't think we shouldn't here. If a user wants to extend what happens at startup time they can set an `entrypoint` (either as part of the image, or via the task definition).

madkinsz: Perhaps with 

```python
if get_flow_run_command(flow_run) not in container["command"]:
    warnings.warn("...")
```

madkinsz: Alrighty. Perhaps a warning or exception if it's already set then?

jcrist: Sure. Can add tomorrow.

src/prefect/run_configs/ecs.py
```diff
@@ -36,7 +39,8 @@ class ECSRun(RunConfig):
             `agent` (for paths local to the runtime agent)).
         - task_definition_arn (str, optional): A pre-registered task definition
             ARN to use (either `family`, `family:version`, or a full task
-            definition ARN).
+            definition ARN). This task definition must include a container
```
TaliaSRTR, line 42: 👍

jcrist:
This has grown a bit in scope, after noticing that our current design wouldn't work with per-run run configs (since before we were debouncing the registration of new task definitions based on the flow name & version). We now register (and deregister) a new task definition for every flow run, unless a pre-registered `task_definition_arn` is provided. Since task definitions don't cost anything to the user, and are quick to create, I don't expect this to cause issues for users.

Full changes:

- Move `command`, `environment`, `cpu`, `memory`, `execution_role_arn`, and `task_role_arn` configuration for ECS tasks from definition time to run time in the ECS agent
- Register (and deregister) a new task definition for every flow run in ECS agent
- Cleanup docs on `execution_role_arn`
- Error if both `task_definition_arn` and `image` are provided to `ECSRun`
- Error if `Docker` storage is mixed with an explicit `task_definition_arn`

madkinsz (approved):
LGTM as long as it's been given a good QA.

PR #4218, opened 2021-03-05
Task ListDirectory

sti0:
## Summary

Adds a new task to the `files` section to receive all file/subdirectory paths of given path within the file system.

## Changes

Adds new task `ListDirectory` and unittests.

## Importance

Doesn't exists..

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/files/operations.py
+++ b/src/prefect/tasks/files/operations.py
@@ -171,3 +171,52 @@ class Remove(Task):
             path.unlink()
         else:
             path.rmdir()
+
+
+class ListDirectory(Task):
+    """
+    This task returns the content of a given directory within the file system.
+
+    Args:
+        - path (Union[str, Path], optional): directory path
+        - recursive (bool, optional): If True, all subdirectories are returned
+            recursively. (defaults to False)
+        - **kwargs (dict, optional): additional keyword arguments to pass to the
+            Task constructor
+    """
+
+    def __init__(
+        self,
+        path: Union[str, Path] = "",
+        recursive: bool = False,
+        **kwargs: Any,
+    ):
+        self.path = path
+        self.recursive = recursive
+        super().__init__(**kwargs)
+
+    @defaults_from_attrs(
+        "path",
+    )
+    def run(self, path: Union[str, Path] = "") -> None:
+        """
+        Task run method.
+
+        Args:
+            - path (Union[str, Path], optional): directory path
+
+        Returns:
+            - List: content of the given path as Path objects
+        """
+        if not path:
+            raise ValueError("No `path` provided.")
+
+        path = Path(path)
+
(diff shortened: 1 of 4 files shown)


src/prefect/tasks/files/operations.py
```diff
@@ -171,3 +171,52 @@ def run(self, path: Union[str, Path] = "") -> None:
...
+
+        path = Path(path)
+
+        if not path.is_dir():
+            raise ValueError(f"Path ('{path}') is not a directory")
+
+        if self.recursive:
+            return list(path.rglob("*"))
+        else:
+            return list(path.glob("*"))
```
jcrist, line 222: Since with `recursive` true this more of a file glob than an `ls` (which only provides the top-level files), and `glob` is a more general `ls` anyway, how about renaming this task to `Glob` instead and making it more general? You'd use `glob.glob` instead of `pathlib`, but other than that the implementation should be fairly straightforward. Something like:

```python
    @defaults_from_attrs(
        "path",
        "recursive",
    )
    def run(self, path: Union[str, Path] = "", recursive: bool = False) -> List[Path]:
        if not path:
            raise ValueError("No `path` provided.")

        return [pathlib.Path(f) for f in glob.glob(path, recursive=recursive)]
```

sti0: I like the idea of making it more general but there is a problem, I think.

`glob.glob` only provides a `str` or `byte-like` object as path. It's necessary to put a glob pattern at the end of the path. In this case we can't simply pass a `Path` object in and get all the files within (which was my intention).

I suggest to introduce a `pattern` parameter. In this case the default parameter can be overwritten. Instead of `glob` i still like to prefer `pathlib.glob / pathlib.rglob` because `glob.glob` needs a double asterisks for recursive results while `pathlib.rglob` doesn't.

I pushed a new version and looking forward to here from you.

sti0: @jcrist This has been open for a while now. Any suggestions on this PR?

madkinsz: Hey @sti0 -- Jim's out of office, sorry this has been sitting.

Personally, it feels like `recursive` shouldn't be an option at all and the user would just pass `pattern="**"` to descend into sub-directories. I feel like in this case you'd just use `glob.glob` with `recursive` always set to `True`. If you want to use `pathlib.glob` that's fine too, it supports double asterisks. Why are you opposed to having to use double asterisks? `rglob` is just placing `**/` in front of the pattern for you. I think it makes sense to match the typical shell interface here.

madkinsz:
Hey @sti0 -- do you mind if take this over and implement my suggestion from https://github.com/PrefectHQ/prefect/pull/4218#discussion_r620672053 ? You're welcome to if you'd like but I'd like to close out this PR.

Issue #4234, opened 2021-03-09
`Task.__signature__` fails if run has a variadic keyword-only parameter


benfogelson:
## Description
`Task.__signature__` is an instance property that inspects the task's `run` method and adds task-specific keyword-only arguments to the call signature for instances of the class. This addition of keyword-only tasks fails if the `run` method has a `**kwargs` parameter, because the implementation of `Task.__signature__` does not correctly order the modified list of parameters.

## Expected Behavior
I expected `Task.__signature__` to run successfully.

## Reproduction
```python
>>> from prefect import Task
>>> class KwargTask(Task):
...     def run(self, **kwargs):
...         pass
>>> task_instance = KwargTask()
>>> task_instance.__signature__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ben.fogelson/repos/prefect/src/prefect/core/task.py", line 191, in __get__
    return self.func(obj)
  File "/Users/ben.fogelson/repos/prefect/src/prefect/core/task.py", line 565, in __signature__
    self._cached_signature = inspect.Signature(
  File "/Users/ben.fogelson/miniconda3/envs/prefect-dev/lib/python3.9/inspect.py", line 2823, in __init__
    raise ValueError(msg)
ValueError: wrong parameter order: variadic keyword parameter before keyword-only parameter
>>>
```

jcrist:
Fixed by #4235.

PR #4235, opened 2021-03-09
Test and bugfix for variadic keyword-only task signature

benfogelson:
## Summary
This PR fixes a bug (https://github.com/PrefectHQ/prefect/issues/4234) where task signatures could not be generated for subclasses of `Task` where `.run` includes `**kwargs`.

## Changes
This PR changes `prefect.core.Task.__signature__` to fix the bug by correctly accounting for parameter kind and order when modifying the task instance's call signature.

## Importance
`task` is a core piece of prefect's functionality, and fixing this bug is necessary for it to work as documented.

## Checklist

This PR:


jcrist (approved):
Thanks for the fix!

PR #4238, opened 2021-03-10
Add raw parameter to S3Download

tonal:
## Summary
see https://github.com/PrefectHQ/prefect/discussions/4226

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/aws/s3.py
+++ b/src/prefect/tasks/aws/s3.py
@@ -43,6 +43,7 @@ class S3Download(Task):
         credentials: str = None,
         bucket: str = None,
         compression: str = None,
+        raw: bool = False,
     ):
         """
         Task run method.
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/aws/s3.py
```diff
@@ -43,6 +43,7 @@ def run(
         credentials: str = None,
         bucket: str = None,
         compression: str = None,
+        raw: bool = False,
```
madkinsz, line 46: I think I would prefer `as_bytes` but it may be worth waiting for @jcrist to weigh in.

madkinsz (commented):
Thanks for contributing! This will also need a test and a changes/ entry.

src/prefect/tasks/aws/s3.py
jcrist: Yeah, `as_bytes` would be preferred here. This probably should have been the default originally, but that ship has sailed.

jcrist (commented):
Can you also add a changelog entry? See https://docs.prefect.io/core/development/contributing.html#changelog.

jcrist (approved):
I pushed the rename to `as_bytes` and added a changelog entry. Thanks for the PR!

PR #4245, opened 2021-03-11
Return flow run information in `StartFlowRun`

madkinsz:
## Summary

This PR is a draft of returning flow run information from successful `StartFlowRun` tasks on success when `wait=True`. Previously, it would return the object `SUCCESS('<flow-run-id> finished in state <Success: "All reference tasks succeeded.">')` with no additional result / information. The `flow_run_id` cannot be retrieved without some arbitrary string parsing and this return value was undocumented. 

To prevent breaking compatibility we could try to return a `SUCCESS` state with the `result` attached as an attribute but it seems like the `Task/FlowRunner` unrolls success states into their `result` attribute if it exists.

## Changes

- `StartFlowRun` return type is documented with `wait` toggle
- `StartFlowRun` returns `prefect.client.client.FlowRunInfoResult`

Also a few usability fixes
- Change some logs to info
- Adds logs for subflow state changes
- Adds logs clarifying `wait` behavior
- Allows `wait` to be passed to `.run(...)`

## Importance

Improves usability of `StartFlowRun` and prepares for more robust result passing.

## Future work

`FlowRunInfoResult` contains `TaskRunInfoResult` for each task in the flow which contains the task states which will hold `Result` references that can be used to retrieve persisted task results. Upgrading the `FlowRunInfoResult` from a `NamedTuple` to a class that supports `__getattr__` like task indexing may allow us to create subflow data passing patterns.

## Example

```python
from prefect.client import Client
from prefect.tasks.prefect.flow_run import StartFlowRun


# Define a very simple child flow -----------------------------------------


@task(log_stdout=True)
def say_hello():
    print("Hello")
    return "Hello"


with Flow("hello-flow") as hello_flow:
    say_hello()

# Register the flow to the API
hello_flow.register("default", idempotency_key="ONLY_REGISTER_ONCE")

# A local agent will need to be started for this flow to run
run_hello_flow = StartFlowRun(flow_name="hello-flow", project_name="default")


# Define our orchestrator flow --------------------------------------------


@task(log_stdout=True)
def do_something_with_run_id(flow_run_id):
    client = Client()
    flow_info = client.get_flow_run_info(flow_run_id)
    # Here we'll have to manually wait if we want the final flow result...
    # Otherwise the flow will likely just be in a Scheduled state
    print(flow_info)


@task(log_stdout=True)
def do_something_with_run_result(flow_result):
    # Here we have the 'terminal' information about the flow if it succeeded
    # If it failed, this task will have automatically failed as its upstream
    # task will have been marked as a failure. We could change this task's
    # triggers to _always_ process the result of the flow
    for task_run in flow_result.task_runs:
        print(f"Subflow task {task_run.task_slug!r} has final state {task_run.state}")


with Flow("orchestrator") as flow:
    # A subflow that we do not wait for finish
    hello_run_id = run_hello_flow(wait=False)
    do_something_with_run_id(hello_run_id)

    # Another subflow that we wait for finish
    hello_info = run_hello_flow(wait=True, run_name="hello-with-wait")
    do_something_with_run_result(hello_info)


# Test --------------------------------------------------------------------

if __name__ == "__main__":
    # We'll run this flow locally without the agent for ease of testing
    flow.run()
```

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -1,12 +1,12 @@
 import time
 import datetime
 import warnings
-from typing import Any
+from typing import Any, Union
 from urllib.parse import urlparse
 
 from prefect import context, Task
 from prefect.artifacts import create_link
-from prefect.client import Client
+from prefect.client.client import Client, FlowRunInfoResult
 from prefect.engine.signals import signal_from_state
 from prefect.run_configs import RunConfig
 from prefect.utilities.graphql import EnumValue, with_args
@@ -95,7 +95,7 @@ class StartFlowRun(Task):
         idempotency_key: str = None,
         scheduled_start_time: datetime.datetime = None,
         wait: bool = None,
-    ) -> str:
+    ) -> Union[str, FlowRunInfoResult]:
         """
         Run method for the task; responsible for scheduling the specified flow run.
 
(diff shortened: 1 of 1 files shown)


madkinsz:
Closing in favor of #4434

PR #4260, opened 2021-03-16
Basic redis result storage - code complete but no tests

mianos:
## Summary
Add a redis Result type to store prefect results in redis.
Incomplete PR. I'll update the rest below when I get to it, or if people are actually interested in this I'll do it sooner.
Using this in our lab environment.

## Changes
Adds RedisResult type.

## Importance
If you use docker storage there is no easy local storage option. This adds it.
In my use case I want to be able to re-run short but complex tasks.
If the data is lost and I have to restart the whole task that is not a huge problem.

## Interim Changes
RedisResult() takes:
  -  connection_string: Redis URI used for Redis.from_url default to REDIS_STORAGE_CONNECTION_STRING
   -  ttl_seconds: key time to live in seconds if the result is to expire

## Checklist

Not for review yet.

This PR:

### DIFF
--- a/src/prefect/engine/results/redis_result.py
+++ b/src/prefect/engine/results/redis_result.py
@@ -1,0 +1,96 @@
+import os
+from typing import Any
+from prefect.engine.result import Result
+import prefect
+
+
+class RedisResult(Result):
+    """
+    Hook for storing and retrieving constant Python objects in redis.
+
+    connection_string: Redis URI used for Redis.from_url
+        default to REDIS_STORAGE_CONNECTION_STRING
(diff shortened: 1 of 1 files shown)


mianos:
Needs work, exists was wrong and some work for mypy. Onto this later today.

madkinsz:
Hi! We're working on redesigning some parts of our results API so we won't be able to accept this. I appreciate having a reference though and I'll keep redis storage in mind :)

mianos:
All good. We have been running this is production for 6 months with the same code but as a module with our flows. No real need for it to be in the core.

PR #4268, opened 2021-03-18
Exasol tasks credential assignment at runtime

sti0:
## Summary

According to #3936 this change will move the `user` and `password` assignment of Exasol tasks to the run() method to work with PrefectSecret.

## Changes

- Remove `user` and `password` from the `__init__()` and move it to `run()`
- `dsn` is optional at `__init__()` but can also assigned at runtime

## Importance

Currently the tasks are not working with PrefectSecret

## Checklist

This PR:


jcrist (commented):
Thanks for the PR. I agree these should be runtime parameters, but to not break backwards compatibility we should keep the args in the constructor for a few releases (with a warning that this is deprecated). Something like:

```
if user is not None or password is not None:
    warnings.warn(
        "Passing `user` or `password` to the `ExasolExecute` constructor "
        "is deprecated. These should be passed as runtime arguments instead."
    )

self.user = user
self.password = password
```

After a few releases we can remove the warning and constructor args.

sti0:
@jcrist added a deprecated warning as requested :)

changes/pr4268.yml
```diff
@@ -0,0 +1,5 @@
+task:
+  - "Exasol tasks: add `user` and `password` as runtime parameter. Both are deprecated for init() method - [#4268](https://github.com/PrefectHQ/prefect/pull/4268)"
```
jcrist, line 2:
```suggestion
task:
  - "Add `user` and `password` as runtime parameters to Exasol tasks - [#4268](https://github.com/PrefectHQ/prefect/pull/4268)"
  
deprecation:
  - "Deprecate `user` and `password` arguments to Exasol task constructors in favor of runtime parameters - [#4268](https://github.com/PrefectHQ/prefect/pull/4268)"
```

jcrist (commented):
Just a couple small edits, otherwise this looks good.

jcrist (approved):

PR #4282, opened 2021-03-22
Add support to CLI for pre-building flows

jcrist:
This PR:

- Adds a new `prefect build` CLI command for building flows and creating/updating a `.json` file containing their representations.
- Adds a `--json` flag to `prefect register` for registering flows from a `.json` file. Note that this file may be local, or hosted elsewhere (currently only `https`/`http`, `s3`, and `gcs` are supported).

Combined, these let users build flows once and host the `json` results somewhere for later registration into one or more projects.

For example, you might build your flows as part of CI (and host the output `flows.json` file on github pages as part of your docs). Users can then register flows from your repo as:

```
$ prefect register --json https://username.github.io/your-project/flows.json --project their-project-name
```

With pre-built flows, the client doesn't need access to the source or dependencies, making registration very lightweight.

### DIFF
--- a/src/prefect/cli/build_register.py
+++ b/src/prefect/cli/build_register.py
@@ -613,6 +613,10 @@ REGISTER_EPILOG = """
 
 \b    $ prefect register --project my-project -m "myproject.flows"
 
+\b  Register all pre-built flows from a remote JSON file.
+
+\b    $ prefect register --project my-project --json https://some-url/flows.json
+
 \b  Watch a directory of flows for changes, and re-register flows upon change.
 
 \b    $ prefect register --project my-project -p myflows/ --watch
@@ -651,7 +655,8 @@ REGISTER_EPILOG = """
     "json_paths",
     help=(
         "A path or URL to a json file containing the flow(s) to register. "
-        "May be passed multiple times to specify multiple paths to register."
+        "May be passed multiple times to specify multiple paths to register. "
+        "Note that this path may be a remote url (e.g. https://some-url/flows.json)."
     ),
     multiple=True,
 )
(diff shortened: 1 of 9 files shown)


src/prefect/cli/build_register.py
```diff
@@ -0,0 +1,972 @@
...
+from prefect.utilities.filesystems import read_bytes_from_path, parse_path
+from prefect.utilities.graphql import with_args, EnumValue, compress
+from prefect.storage import Local, Module
+from prefect.run_configs import UniversalRun
+
+
+FlowLike = Union[box.Box, "prefect.Flow"]
+
+
+class SimpleFlowSchema(marshmallow.Schema):
```
cicdw, line 29: Any reason we shouldn't use pydantic here, or is it not worth the new dependency?

src/prefect/cli/build_register.py
```diff
@@ -0,0 +1,972 @@
...
+        "passed multiple times to specify multiple modules to register."
+    ),
+    multiple=True,
+)
+@click.option(
+    "--json",
+    "-j",
+    "json_paths",
+    help=(
+        "A path or URL to a json file containing the flow(s) to register. "
```
cicdw, line 657: The fact that the schema of this file is not a dump of `flow.serialize()` took me a minute to work through and seems like something we should call out here -- not sure of the simplest way to say "we expect this file to be the output of `prefect build`

cicdw (changes requested):
Two questions, overall this looks great

src/prefect/cli/build_register.py
jcrist: I want to use pydantic, but sticking with marshmallow for now. We should maybe prioritize transitioning though.

src/prefect/cli/build_register.py
jcrist: Sure, I can expand the help string here.

jcrist: Done

cicdw (approved):

PR #4283, opened 2021-03-22
Improving logging of Jupyter task to capture cell output

kvnkho:
## Summary
This is a response to [#4265](https://github.com/PrefectHQ/prefect/issues/4265), which was raised on Slack. The user wanted to capture Jupyter notebook cell outputs in the Prefect logs.

## Changes
There are two things needed to get this to work. One is to pass the `log_output` bool into `papermill`'s execute_notebook function. By default this is False. It needs to be True. For consistency with the previous implementation, I kept the default asFalse, though I feel there might be benefits in making it True.

The second thing is to get the `papermill` logger and change the level to `INFO` or `DEBUG`. By default, it is at `WARN` so it does not log cell outputs. I think we should not do this change on the task side. Instead, Prefect users should use the (extra loggers)[https://docs.prefect.io/core/concepts/logging.html#extra-loggers] part of the TOML config and add the `papermill` logger.

There is an added test to see if the `papermill` logs are being captured. There are some edits to the .ipynb to see more output from the `papermill` logs.

## Checklist

This PR:

### DIFF
--- a/changes/issue4265.yaml
+++ b/changes/issue4265.yaml
@@ -1,0 +1,6 @@
+task:
+  - "Adds logging of cell outputs to Jupyter task - [#4265]
+  (https://github.com/PrefectHQ/prefect/issues/4265)"
+
+contributor:
+  - "[Kevin Kho](https://github.com/kvnkho)"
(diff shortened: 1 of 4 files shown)


cicdw (approved):
LGTM and congrats on your first PR!

PR #4287, opened 2021-03-23
Respect global context in flow runner

### LINKED ISSUE #4276: FlowRunner won't preserve all items from prefect.context
## Description

FlowRunner does not preserve all items from prefect.context (e.g. dates: today, yesterday, etc.), because `initialize_run()` method prepares an empty context if not specified in `context` parameter. This could be okay, but contradicts to the [documentation](https://docs.prefect.io/core/concepts/execution.html#modifying-context-at-runtime), where it is suggested that for "backfills", dates could be overridded (e.g. today). 

## Expected Behavior

Items already set in prefect.context should be respected.

## Reproduction

Both runs should output the same value "YESTERDAAAAAY" for `yesterday`, but the first outputs the actual date (2021-03-18 as of this writing).

```python
import prefect

CTX = dict(test_variable="TEST_VALUE", yesterday="YESTERDAAAAAY")

@task(log_stdout=True)
def print_env():
    print('TEST_VARIABLE: %s' % prefect.context.get('test_variable', 'MISSING'))
    print('YESTERDAY: %s' % prefect.context.get('yesterday', 'MISSING'))

with Flow("Context override test") as flow:
    print_env()

## Run 1
with prefect.context(**CTX):
    flow.run()

## Run 2
flow.run(context=CTX)
```

## Environment

```json
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Linux-4.18.0-240.10.1.el8_3.x86_64-x86_64-with-centos-8.3.2011",
    "prefect_backend": "server",
    "prefect_version": "0.14.10",
    "python_version": "3.6.8"
  }
}
```

## Suggested Resolution

A single line added to FlowRunner could solve that. 

```diff
--- prefect/engine/flow_runner.py.orig  2021-02-16 16:08:28.204208267 +0100
+++ prefect/engine/flow_runner.py       2021-03-19 16:36:56.278923592 +0100
@@ -152,6 +152,7 @@
         for param, value in (parameters or {}).items():
             context_params[param] = value
 
+        context.update(prefect.context)
         context.update(flow_name=self.flow.name)
         context.setdefault("scheduled_start_time", pendulum.now("utc"))
```

jcrist:
When adding default fields to context during a flow run, several check
if there's an existing value and only override if it doesn't already
exist (e.g. `"date"`). However, this would only work if the context was
passed explicitly to `.run(context=context)`, rather than being part of
the implicit global context (since that wasn't added until the very
end). This PR fixes that bug.

Fixes #4276.

### DIFF
--- a/changes/pr4287.yaml
+++ b/changes/pr4287.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix bug where sometimes the global `prefect.context` wouldn't be respected during a flow run - [#4287](https://github.com/PrefectHQ/prefect/pull/4287)"
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -224,10 +224,13 @@ class FlowRunner(Runner):
         self.logger.info("Beginning Flow run for '{}'".format(self.flow.name))
 
         # make copies to avoid modifying user inputs
-        task_states = dict(task_states or {})
-        context = dict(context or {})
-        task_contexts = dict(task_contexts or {})
         parameters = dict(parameters or {})
+        task_states = dict(task_states or {})
+        task_contexts = dict(task_contexts or {})
+        # Default to global context, with provided context as override
+        run_context = dict(prefect.context)
+        run_context.update(context or {})
+
         if executor is None:
             # Use the executor on the flow, if configured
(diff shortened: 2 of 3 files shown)


cicdw (approved):
LGTM

PR #4293, opened 2021-03-25
Fixing Snowflake Query Task

kvnkho:
## Summary

This is to fix this [issue](https://github.com/PrefectHQ/prefect/issues/3744) about Snowflake output not being pickleable. This is because the original code just returned the cursor.

I spun up a free trial for Snowflake and tested this change with the default databases. I then used `cloudpickle` to make sure results are serializable. This is hard to reflect in the unit tests.

## Changes

I added the `fetchall` call, and then I added a new unit test that ensures the Task has the `fetchall` call.

## Checklist
This PR:

### DIFF
--- a/tests/tasks/snowflake/test_snowflake.py
+++ b/tests/tasks/snowflake/test_snowflake.py
@@ -49,7 +49,8 @@ class TestSnowflakeQuery:
     def test_execute_fetchall(self, monkeypatch):
         """
         Tests that the SnowflakeQuery Task calls the fetchall method on the
-        cursor. This is to prevent future code edits from returning the cursor.
+        cursor. This is to prevent future code edits from returning the cursor
+        object because that cursors are not pickleable.
         """
         snowflake_module_connect_method = MagicMock()
         connection = MagicMock(spec=sf.SnowflakeConnection)
--- a/tests/tasks/snowflake/test_snowflake.py
+++ b/tests/tasks/snowflake/test_snowflake.py
@@ -45,3 +45,29 @@ class TestSnowflakeQuery:
 
         with pytest.raises(sf.errors.DatabaseError, match="Invalid query"):
             task.run(query="SELECT * FROM foo")
+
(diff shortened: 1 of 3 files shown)


tests/tasks/snowflake/test_snowflake.py
```diff
@@ -45,3 +45,29 @@ def test_execute_error_must_pass_through(self, monkeypatch):
 
         with pytest.raises(sf.errors.DatabaseError, match="Invalid query"):
             task.run(query="SELECT * FROM foo")
+
+    def test_execute_fetchall(self, monkeypatch):
```
cicdw, line 49: For tests like this, a good practice is to add a docstring explaining the test -- in this case, you don't _actually_ care that `fetchall` is called, you care that the result is cloudpickleable.  The docstring ensures that someone who changes to another method call and this test fails, they know what they need to validate to safely delete / change the test.

cicdw (changes requested):
Awesome!  Minor request for a docstring on the new test clarifying why the test exists

tests/tasks/snowflake/test_snowflake.py
kvnkho: Can definitely add the docstring but just to clarify, this is really intended for the `fetchall` call because it means that the data was pulled from the Snowflake Query cursor and it would prevent future code edits from just returning the cursor.

I can't really test retrieving data from Snowflake and pickling that because I can't spin up a test database. In my tests though, `fetchall` returns a list.

cicdw: Right -- but if the API changes or someone uses a different method call here for whatever reason to something else that also doesn't return a cursor, they might delete this test when it inevitably fails, but the docstring will soft-enforce that the test is in place for pickleability, not because we have an opinion on using `fetchall`.

cicdw (dismissed):

cicdw (approved):

PR #4295, opened 2021-03-25
Use custom `date` in context to determine dependent context vars

madkinsz:
## Summary

Previously, if you overrode `date` in the context for a flow, the dependent variables such as `tomorrow` and `today` would still be based on `pendulum.now()` making it hard for a user to do a simple override and see expected context changes. This will attempt to parse `date` if set to determine the "now" `DateTime` to be used to set the dependent variables. If unparsable, a warning will be displayed.

## Changes

Dependent variables are automatically updated if `date` is set and is parsable as a `DateTime`.

## Importance

Improves usability for testing / backfilled schedules.

## Checklist

This PR:

### DIFF
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -155,8 +155,26 @@ class FlowRunner(Runner):
         context.update(flow_name=self.flow.name)
         context.setdefault("scheduled_start_time", pendulum.now("utc"))
 
+        # Determine the current time, allowing our formatted dates in the context
+        # to rely on a manually set value
+        now = context.get("date")
+        if isinstance(now, str):
+            # Attempt to parse into a `DateTime` object
+            try:
+                now = pendulum.parse(now)
+            except:
+                pass
+        if not isinstance(now, pendulum.DateTime):
+            if now is not None:
+                self.logger.warning(
+                    "`date` was set in the context manually but could not be parsed "
+                    "into a pendulum `DateTime` object. Additional context variables "
+                    "that rely on the current date i.e `today` and `tomorrow` will be "
+                    "based on the current time instead of the `date` context variable."
+                )
+            now = pendulum.now()
+
         # add various formatted dates to context
-        now = pendulum.now("utc")
         dates = {
             "date": now,
             "today": now.strftime("%Y-%m-%d"),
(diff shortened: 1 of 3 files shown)


src/prefect/engine/flow_runner.py
```diff
@@ -155,8 +155,26 @@ def initialize_run(  # type: ignore
...
+                pass
+        if not isinstance(now, pendulum.DateTime):
+            if now is not None:
+                self.logger.warning(
+                    "`date` was set in the context manually but could not be parsed "
+                    "into a pendulum `DateTime` object. Additional context variables "
+                    "that rely on the current date i.e `today` and `tomorrow` will be "
+                    "based on the current time instead of the `date` context variable."
+                )
+            now = pendulum.now()
```
jcrist, line 175: For full compatibility, this should still pass in `utc` I think:

```suggestion
            now = pendulum.now('utc')
```

src/prefect/engine/flow_runner.py
```diff
@@ -155,8 +155,26 @@ def initialize_run(  # type: ignore
         context.update(flow_name=self.flow.name)
         context.setdefault("scheduled_start_time", pendulum.now("utc"))
 
+        # Determine the current time, allowing our formatted dates in the context
+        # to rely on a manually set value
+        now = context.get("date")
+        if isinstance(now, str):
+            # Attempt to parse into a `DateTime` object
+            try:
+                now = pendulum.parse(now)
```
jcrist, line 164: When parsing the date, should we also convert to `'utc'` so the `date` context is always in utc? No strong thoughts either way.

src/prefect/engine/flow_runner.py
madkinsz: Good catch. Will do

src/prefect/engine/flow_runner.py
madkinsz: I think not. It wouldn't override their `date` variable since we're using `setdefault` below and it'd make `date` be in whatever tz they used and all the dependent variables in `utc`

jcrist: >  It wouldn't override their date variable since we're using setdefault below

Hmmm, I'm wondering if we should override `date` in context if we successfully parse it. Values set via cloud are sent as json (so it's a string), but if `date` is normally a pendulum object it'd be nice to keep that interface uniform.

madkinsz: I think it makes sense to override it in just the case where we parse from string -> pendulum object, I'll add that.

madkinsz:
Lookout for the ci skipped changes update.

jcrist (approved):

PR #4296, opened 2021-03-25
Always load latest task run info prior to running task

jlowin:
## Summary

Generally, we pre-load task run information when the flow run starts. The only time we load task run information when the task itself starts is for dynamic tasks. However, in certain fast-moving situations, the flow may actually begin running before all task-level detail is available. This PR modifies the task run info logic to ALWAYS load when the task run starts.

With this in place, we could further improve performance by completely eliminating the initial batch task run load, but I'm not including that here.

## Checklist

This PR:

### DIFF
--- a/changes/pr4296.yaml
+++ b/changes/pr4296.yaml
@@ -1,0 +1,20 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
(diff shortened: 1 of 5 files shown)


cicdw (changes requested):
Could you add a `fix` changelog entry?  Also looks like you're failing some style checks.

Otherwise this LGTM!

cicdw (approved):

PR #4305, opened 2021-03-26
Modified resource manager to allow success and failure cleanup

benfogelson:
## Summary
This PR modifies the resource manager so that the user can supply `success_cleanup()` and `failure_cleanup()` functions in addition to `cleanup()`. When present, `success_cleanup()` is executed before `cleanup()` if all tasks in the resource manager context succeeded, and `failure_cleanup()` is executed if at least on task in the context failed.

## Changes
Changes `resource_manager.py` by adding `ResourceSuccessCleanupTask` and `ResourceFailureCleanupTask`, adds custom triggers for each class type, modifies `ResourceContext.__exit__` to set downstream tasks with the new conditional cleanup tasks, and `ResourceManager` and `resource_manager` to support these cleanup tasks.

## Importance
Resource managers are important, because they allow many tasks to make use of a shared resource while ensuring that that resource is only cleaned up after all tasks that need it have finished. It is commonly the case that different cleanup steps need to be performed depending on the outcome of tasks that depend on the resource. For example, the resource could be a database transaction, which we want to commit or rollback depending on the success or failure of the queries that are part of that transaction.

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -407,6 +407,8 @@ class Flow:
         from prefect.tasks.core.resource_manager import (
             ResourceInitTask,
             ResourceCleanupTask,
+            ResourceSuccessCleanupTask,
+            ResourceFailureCleanupTask,
         )
 
         # Select all tasks that aren't a ResourceInitTask/ResourceCleanupTask
@@ -419,11 +421,26 @@ class Flow:
         return {
             t
             for t in self.tasks
-            if not isinstance(t, (ResourceInitTask, ResourceCleanupTask))
+            if not isinstance(
+                t,
+                (
+                    ResourceInitTask,
+                    ResourceCleanupTask,
+                    ResourceSuccessCleanupTask,
+                    ResourceFailureCleanupTask,
+                ),
+            )
             and not any(
                 t
                 for t in self.downstream_tasks(t)
-                if not isinstance(t, ResourceCleanupTask)
+                if not isinstance(
+                    t,
+                    (
+                        ResourceCleanupTask,
+                        ResourceSuccessCleanupTask,
+                        ResourceFailureCleanupTask,
+                    ),
+                )
             )
         }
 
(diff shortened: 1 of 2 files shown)


jcrist:
Hmmm, interesting idea. I think I'd rather better expose the state (success/failure) inside the single cleanup task and have you branch inside it rather than add more callbacks to the resource manager. Feels cleaner to keep all your cleanup logic together, where you can better control and view the order it runs in.

This could be done by adding an extra (optional) parameter to the cleanup function that gets e.g the upstream states, or by adding some extra metadata to the cleanup function's `prefect.context`.

Potential pseudocode:

```python
@resource_manager
class ExtraParameter:
    ...
    def cleanup(self, resource, upstream_states):
        if any_failed(upstream_states):
            ...
        else:
            ...
```

Thoughts?

benfogelson:
Thanks for the comment @jcrist. Your suggestion sounds way cleaner to me as well. I like the context idea: I'm thinking that `resource_cleanup_trigger` could set the context, since it gets the states and is run right before the cleanup task runs. Does that sound reasonable to you?

jcrist:
That could work, but I'm now wondering if we'd want to add all upstream states to context anyway in the `TaskRunner` (they're already always loaded) - cc @cicdw for thoughts here.

One worry I have is that `prefect.context`-based apis are hard to change once added, since the `prefect.context` object is an untyped nested dict. If we wanted to deprecate the api we'd have to add features to our `prefect.context` object just to support deprecation, while an explicit (optional) parameter could be handled local to the `resource_manager` code.

cicdw:
There are other use cases for putting them all in context for the duration of the task runner pipeline (for example, some users have requested this to be used within state handlers).  I also get nervous about bloating context further, and ideally Prefect would expose hooks for all places where state-based logic could be implemented, but in this case upstream task states do feel "contextual" so it does make some sense.

madkinsz:
If all the upstream states are in the context does that make #4311 unnecessary?

madkinsz:
Closing as stale. I think it is unlikely that we will extend this in the near future :)

PR #4307, opened 2021-03-28
Add new task KubernetesSecret

davzucky:
## Summary
This is a new task that allows us to extract Kubernetes Secret and transform it into a Prefect secret.
The idea is to have another secret store from the one which already exists. Prefect, env var or AWSvault.

## Changes
This PR adds a new task KubernetesSecret that inherit from SecretBase

## Importance
This PR allows users who are using Prefect Server to centralize all their secret on a Kubernetes Secret object

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/kubernetes/secrets.py
+++ b/src/prefect/tasks/kubernetes/secrets.py
@@ -69,8 +69,6 @@ class KubernetesSecret(SecretBase):
         "namespace",
         "kube_kwargs",
         "kubernetes_api_key_secret",
-        "cast_function",
-        "raise_if_missing",
     )
     def run(
         self,
@@ -79,8 +77,6 @@ class KubernetesSecret(SecretBase):
         namespace: str = "default",
         kube_kwargs: dict = None,
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
-        cast_function: Callable[[Any], Any] = None,
-        raise_if_missing: bool = False,
     ):
         """
         Returns the value of an kubenetes secret after applying an optional `cast` function.
@@ -94,11 +90,6 @@ class KubernetesSecret(SecretBase):
             - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
                 which stored your Kubernetes API Key; this Secret must be a string and in
                 BearerToken format
-            - cast_function (Callable[[Any], Any]): A function that will be called on the Parameter
-                value to coerce it to a type.
-            - raise_if_missing (bool): if True, an error will be raised if the env var is not found.
-        Returns:
-            - Any: the (optionally type-cast) value of the Kubenetes secret
 
         Raises:
             - ValueError: if `raise_is_missing` is `True` and the kubernetes secret was not found.
@@ -120,7 +111,7 @@ class KubernetesSecret(SecretBase):
         ).data
 
         if secret_key not in secret_data:
-            if raise_if_missing:
+            if self.raise_if_missing:
                 raise ValueError(f"Cannot find the key {secret_key} in {secret_name} ")
             else:
                 return None
@@ -128,5 +119,7 @@ class KubernetesSecret(SecretBase):
         decoded_secret = base64.b64decode(secret_data[secret_key]).decode("utf8")
 
         return (
-            decoded_secret if cast_function is None else cast_function(decoded_secret)
+            decoded_secret
+            if self.cast_function is None
+            else self.cast_function(decoded_secret)
         )
--- a/src/prefect/tasks/kubernetes/secrets.py
+++ b/src/prefect/tasks/kubernetes/secrets.py
@@ -81,7 +81,7 @@ class KubernetesSecret(SecretBase):
         kubernetes_api_key_secret: str = "KUBERNETES_API_KEY",
         cast_function: Callable[[Any], Any] = None,
         raise_if_missing: bool = False,
-    ) -> Any:
+    ):
         """
         Returns the value of an kubenetes secret after applying an optional `cast` function.
 
--- a/src/prefect/tasks/kubernetes/secrets.py
+++ b/src/prefect/tasks/kubernetes/secrets.py
@@ -12,11 +12,11 @@ from prefect.utilities.kubernetes import get_kubernetes_client
 class KubernetesSecret(SecretBase):
     """
     Task for loading a Prefect secret from a kubernetes secret.
-    
+
     This task will read a secret from kubernetes, returning the decoded value
     associated with `secret_key`. All initialization arguments can optionally
     be provided or overwritten at runtime.
-    
+
     Note that depending on cluster configuration, you may need to ensure you
     have the proper RBAC permissions to read the secret.
 
--- a/src/prefect/tasks/kubernetes/secrets.py
+++ b/src/prefect/tasks/kubernetes/secrets.py
@@ -14,8 +14,8 @@ class KubernetesSecret(SecretBase):
     Task for creating a Prefect secret from the kubernetes secret.
 
     This task will read the secret object from kubernetes and extract the value of the secret_key
-    and decode it. The kubernetes secret can be created by sealedsecret or vault and read here. 
-    
+    and decode it. The kubernetes secret can be created by sealedsecret or vault and read here.
+
     Note that you need to ensure that you have the right RBAC from your cluster admin to read the secret
     Note that all initialization arguments can optionally be provided or overwritten at runtime.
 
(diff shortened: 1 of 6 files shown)


src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+
+class KubernetesSecret(SecretBase):
+    """
+    Task for creating a Prefect secret from the kubernetes secret.
+
+    This task will read the secret object from kubernetes and extract the value of the secret_key
+    and decode it. The kubernetes secret can be created by sealedsecret or vault and read here.
+
+    Note that you need to ensure that you have the right RBAC from your cluster admin to read the secret
+    Note that all initialization arguments can optionally be provided or overwritten at runtime.
```
jcrist, line 20:
```suggestion
    Task for loading a Prefect secret from a kubernetes secret.

    This task will read a secret from kubernetes, returning the decoded value
    associated with `secret_key`. All initialization arguments can optionally
    be provided or overwritten at runtime.

    Note that depending on cluster configuration, you may need to ensure you
    have the proper RBAC permissions to read the secret.
```

src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+    `kubernetes_api_key_secret` = `None` then it will attempt the next two connection
+    methods. By default the value is `KUBERNETES_API_KEY` so providing `None` acts as
+    an override for the remote connection.
+    2. Attempt in-cluster connection (will only work when running on a Pod in a cluster)
+    3. Attempt out-of-cluster connection using the default location for a kube config file
+
+
+    Args:
+        - secret_name (string, optional): The name of the kubernetes secret object
+        - secret_key (string, optional): The key to look for in the kubernetes data
```
jcrist, line 32: What do you think about making this optional, in which case a mapping of `{key: decoded_value}` is returned for all items in the k8s secret?

src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+        - namespace (str, optional): The Kubernetes namespace to read the secret from
```
jcrist, line 33:
```suggestion
        - namespace (str, optional): The Kubernetes namespace to read the secret from,
             defaults to the `default` namespace.
```

src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+        - kube_kwargs (dict, optional): Optional extra keyword arguments to pass to the
+            Kubernetes API (e.g. `{"pretty": "...", "dry_run": "..."}`)
+        - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
+            which stored your Kubernetes API Key; this Secret must be a string and in
+            BearerToken format
+        - cast_function (Callable[[Any], Any]): A function that will be called on the Parameter
+            value to coerce it to a type.
```
jcrist, line 40:
```suggestion
        - cast_function (Callable[[Any], Any], optional): If provided, this will
            be called on the secret to transform it before returning. An example
            use might be passing in `json.loads` to load values from stored JSON.
```

jcrist: What do you think about renaming this to `cast` instead to maintain consistency with the api of `EnvVarSecret`?

src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+        - raise_if_missing (bool): if True, an error will be raised if the env var is not found.
```
jcrist, line 41:
```suggestion
        - raise_if_missing (bool): if True, an error will be raised if the secret is not found.
```

src/prefect/tasks/kubernetes/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Args:
+            - secret_name (string, optional): The name of the kubernetes secret object
+            - secret_key (string, optional): The key to look for in the kubernetes data
+            - namespace (str, optional): The Kubernetes namespace to read the secret from
+            - kube_kwargs (dict, optional): Optional extra keyword arguments to pass to the
+                Kubernetes API (e.g. `{"pretty": "...", "dry_run": "..."}`)
+            - kubernetes_api_key_secret (str, optional): the name of the Prefect Secret
+                which stored your Kubernetes API Key; this Secret must be a string and in
+                BearerToken format
```
jcrist, line 92: Same edits I suggested above also apply here.

jcrist (commented):
Overall this seems good to me. Made a few edits to the docstrings, and asked a few api design questions.

One question I have though is around intended use. If the flow run is deployed inside K8s, wouldn't it make more sense to map the secret into the flow run pod and load it from an environment variable/file (rather than giving the pod full access to the k8s api)? I guess I'm curious what use case you have in mind for this.

davzucky:
@jcrist Thank you for your comment.
folllowing on your question
```One question I have though is around intended use. If the flow run is deployed inside K8s, wouldn't it make more sense to map the secret into the flow run pod and load it from an environment variable/file (rather than giving the pod full access to the k8s api)? I guess I'm curious what use case you have in mind for this.```
Yes, this is what we do at the moment using env var. However, we want to have our dask worker pod to not have to worried about k8s secret been mounted. Using this secret would allow us to have a single secret store and pod extract the secret using the key they want and store it on the prefect result store.
This is another option similar to the env var one.

File: src/prefect/tasks/kubernetes/secrets.py (hunk 2 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py:32]: Could make it optional. However, I'm not sure of the value. In this case, we would return a Dict[str. str]. but when I think of secret I think of it as one key. The secret_name map to the k8s secret (the store) and the secret key return the value we want to extract.
If we allow it to be optional we would return for example
{"PASSWORD_DB":  "132456", "PASSWORD_S3": "654321"} 
we would have to map them later on. In this case, we can be strict about ensuring the key exists 

What do you think ? I prefer been strict here

File: src/prefect/tasks/kubernetes/secrets.py (hunk 1 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py]: updated

File: src/prefect/tasks/kubernetes/secrets.py (hunk 3 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py]: updated

File: src/prefect/tasks/kubernetes/secrets.py (hunk 4 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py]: it was cast before, however I use ```cast``` from ```typing``` in this case it changes the function. This is why I had to give it another name

File: src/prefect/tasks/kubernetes/secrets.py (hunk 5 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py]: updated

File: src/prefect/tasks/kubernetes/secrets.py (hunk 6 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py:96]: updated

**Comment by davzucky** [comment]
> @jcrist Thank you for your comment.
> folllowing on your question
> ```One question I have though is around intended use. If the flow run is deployed inside K8s, wouldn't it make more sense to map the secret into the flow run pod and load it from an environment variable/file (rather than giving the pod full access to the k8s api)? I guess I'm curious what use case you have in mind for this.```
> Yes, this is what we do at the moment using env var. However, we want to have our dask worker pod to not have to worried about k8s secret been mounted. Using this secret would allow us to have a single secret store and pod extract the secret using the key they want and store it on the prefect result store.
> This is another option similar to the env var one.

To add as well. This allow us to retrieve different secret for different env in the same cluster without having to mount the kubernetes secret

File: src/prefect/tasks/kubernetes/secrets.py (hunk 4 above)
Comment (davzucky) [reply src/prefect/tasks/kubernetes/secrets.py]: I renamed the parameter to cast. My bad about the previous comment

File: src/prefect/tasks/kubernetes/secrets.py (hunk 2 above)
Comment (jcrist) [reply src/prefect/tasks/kubernetes/secrets.py:33]: The use case I was thinking of was some prefect secrets expect JSON values - using a k8s secret with with multiple keys feels like a natural way to write this, rather than doing nesting below a single key. Leaving this out for now is fine though, can always add it later if users want it.

**Review by jcrist — APPROVED** [review]
LGTM. Thanks!

**Comment by kst83** [comment]
Is it possible to fall back to a different secret type if the KubernetesSecret isn't available, e.g. because the flow runs on a local computer without access to the cluster?
E.g. is there a way to basically put a nvl function around the secret? Kind of like this:
```
my_secret = nvl(
    KubernetesSecret(secret_name="secret-name", secret_key="valide_secret"),
    EnvVarSecret("env-name"),
)
```

**Comment by kst83** [comment]
Seems this is already possible, though instead of nvl or coalesce it is called merge:

```
from prefect.tasks.control_flow import merge
my_secret = merge(
    KubernetesSecret(secret_name="secret-name", secret_key="valide_secret"),
    EnvVarSecret("env-name"),
)
```

davzucky:
> 
> 
> Seems this is already possible, though instead of nvl or coalesce it is called merge:
> 
> ```
> from prefect.tasks.control_flow import merge
> my_secret = merge(
>     KubernetesSecret(secret_name="secret-name", secret_key="valide_secret"),
>     EnvVarSecret("env-name"),
> )
> ```

Thank you @xyxz-web for this input. We were using a flag at flow build time to do that. Your solution is nice as well.

PR #4309, opened 2021-03-28
Fix `DatabricksRunNow` task attribute override behavior

greglu:
## Summary

`DatabricksRunNow.run` was not properly overriding task attributes, which resulted in unexpected job json objects being generated for the Databricks API calls.

## Changes

* Fixes task attribute override behavior when calling `DatabricksRunNow.run`.
* Improves `test_databricks.py` to allow mocked responses from Databricks API.
* Implement tests for `DatabricksRunNow.run` and generated job json object for Databricks API calls.

## Importance

Using the `DatabricksRunNow` task resulted in some unexpected behavior when attempting to override task attributes during `DatabricksRunNow.run`. The issues are best demonstrated with this simple test (which currently fails):

```python
def test_run_with_attribute_overrides():
    task_template = DatabricksRunNow(
        databricks_conn_secret={
            "host": "https://cloud.databricks.com",
            "token": "databricks-token",
        },
        json={
            "job_id": 1,
            "notebook_params": {
                "dry-run": "false"
            }
        },
    )

    task_template.run(job_id="42", notebook_params={"dry-run": "true"})

    print(task_template.json)
    # {'job_id': '1', 'notebook_params': {'dry-run': 'false'}}

    assert task_template.json.get("job_id") == "42"
    assert task_template.json.get("notebook_params", {}).get("dry-run") == "true"
```

Aside from these 2 attributes, the same issues also occurred with all other task attributes.

Test coverage was also sparse and there was no ability to actually test the results of `DatabricksRunNow.run` calls without triggering outgoing API requests. By subclassing and overriding specific methods, the task and run method behavior can be better tested now.

## Checklist

This PR:

### DIFF
--- a/changes/pr4309.yaml
+++ b/changes/pr4309.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix `DatabricksRunNow` task attribute override behavior - [#4309](https://github.com/PrefectHQ/prefect/pull/4309)"
+
+contributor:
+  - "[Greg Lu](https://github.com/greglu)"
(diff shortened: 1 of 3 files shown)


jcrist (approved):
Thanks for the fix!

Issue #4198, opened 2021-03-03
Allow the terminal flow state handler to be overriden


madkinsz:
## Current behavior

The terminal flow state handler is a `FlowRunner` function that cannot be overridden using the `Flow` api. While users can create `state_handlers` at the flow level which allow them to mutate states in-transition, the state handler is only given the flow object, the old flow state, and the new flow state. In contrast, the terminal flow state handler is given task state information which is very helpful for determining custom messages depending on the tasks that have failed. This is particularly useful to people using Hooks to notify on flow failure.

## Proposed behavior

- Add the attribute `terminal_state_handler: Optional[Callable]` to `Flow`
- At the end of `FlowRunner.determine_final_state` if the flow has `terminal_state_handler`, call it to mutate the final state and return its return value instead

*signature of terminal state handling in FlowRunner*
```python
    def determine_final_state(
        self,
        state: State,
        key_states: Set[State],
        return_states: Dict[Task, State],
        terminal_states: Set[State],
    ) -> State:
```

Alternative solutions include
- Using a normal state handler and encouraging the user to query GraphQL for their task state information (likely not performant)
- Adding task state data to the normal state handler (seems significantly more complex to collect)

## Example

This enhancement would allow custom messages such as iterating through the return states and indicating which tasks have failed.

zangell44:
Do Server/Cloud need to know about the `terminal_state_handler` property as well?

(Attempting to answer my own question)
No.

The code associated with a Flow, such as existing state handlers, is packaged up with the rest of your Python code when the Flow is registered.

Anything outside of the Python code, such as run config, needs to be stored in the database.

Do I have that right?

madkinsz:
> Do Server/Cloud need to know about the `terminal_state_handler` property as well?
> 
> (Attempting to answer my own question)
> No.

Not really.
 
> The code associated with a Flow, such as existing state handlers, is packaged up with the rest of your Python code when the Flow is registered.
> 
> Anything outside of the Python code, such as run config, needs to be stored in the database.
> 
> Do I have that right?

Roughly yes. The run config is included in the serialized flow because we allow it to be overridden by the backend afaik. @jcrist would have a better answer about what belongs in the serialized `FlowSchema`

jcrist:
Yeah, there's no need to modify serializers here - this will all be handled core-side only.

PR #4311, opened 2021-03-29
Allow terminal flow state handler override

zangell44:
## Summary
Introduces a new `terminal_state_handler` to `Flow` objects. If present, the state handler mutates the final Flow state in `FlowRunner.determine_final_state`.

## Changes

- Adds `terminal_state_handler` property and logic
- Adds relevant tests

## Importance
Resolves [#4198](https://github.com/PrefectHQ/prefect/issues/4198).

## Checklist

This PR:

TODO
- [x] Look into adding something to the docs? Perhaps at https://docs.prefect.io/core/concepts/states.html#flow-state-transitions or https://docs.prefect.io/core/concepts/flows.html#state-handlers
- [x] Ensure that a hook to call this doesn't need to be added to the CloudFlowRunner as well


madkinsz (commented):
Great! 

Can you:
- Look into adding something to the docs? Perhaps at https://docs.prefect.io/core/concepts/states.html#flow-state-transitions or https://docs.prefect.io/core/concepts/flows.html#state-handlers
- Ensure that a hook to call this doesn't need to be added to the `CloudFlowRunner` as well

changes/issue4198.yaml
```diff
@@ -0,0 +1,20 @@
...
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+enhancement:
+  - "Add terminal flow state handler override  - [#4198](https://github.com/PrefectHQ/prefect/issues/4198)"
```
hui-zheng, line 20: just want to note that, I also see other change-logs putting PR# here.

madkinsz: Both are accepted patterns. I'm actually curious if we want to consolidate to just using the PR though @cicdw ?

cicdw: If there's an open issue it generally provides more context than the PR so I tend to link to issues if they exist (personal preference), otherwise I use the PR link.  There's no need to pick a unifying rule here.

jcrist: I always use the PR #, but don't think it matters too much. Having something to link the change back to is the important part.

zangell44: Good point, I'll try to get some clarity on which reference is preferred

src/prefect/core/flow.py
```diff
@@ -143,6 +143,12 @@ def my_task():
             the flow (e.g., presence of cycles and illegal keys) after adding the edges passed
             in the `edges` argument. Defaults to the value of `eager_edge_validation` in
             your prefect configuration file.
+        -  terminal_state_handler (Callable, optional): A state handler that mutates the final state
+            of the flow, with signature `state_handler(state: State, key_states: Set[State],
+            return_states: Dict[Task, State], terminal_states: Set[State],) -> State`.
+            `state` is the current state of the Flow. `key_states` are the states that will determine
+            the success / failure of the flow run. `return_states` contains states for Flow Tasks.
+            `terminal_states` are the states of terminal tasks for this Flow.
```
jcrist, line 151: Currently this is written in a way that exposes the internal workings of the `FlowRunner` in a way that the user probably doesn't care about (and we wouldn't want to guarantee wouldn't change). For example - `return_states` is all states that are requested to return (No states by default). `key_states` are known to users as "reference task states" (`key_states` is just an internal variable name, not a term a user would recognize). I'd argue that we want to present an API here that is easier for a user to understand without exposing internal details. We'd also want something closer to the existing api for other state handlers (for consistency).

Perhaps:

```python
def terminal_state_handler(flow: Flow, state: State, task_states: Dict[Task, State]) -> Optional[State]:
    """Determine the final state for a flow.

    Args:
        - flow (Flow): the flow itself.
        - state (State): the current flow state.
        - task_states (Dict[Task, State]): A mapping from all tasks to their
            corresponding states.
    Returns:
        - state (State, optional): Optionally return a new state if different
            than the original state. If no state is returned, the original state
            will be used.
    """
```

If a user wants to to make a decision based on reference tasks (possible) or terminal tasks (unlikely) they can query those from the `flow` object directly and then use the `task_states` mapping to map tasks to their corresponding states. But it also provides access to all included task states under a uniform and more understandable api.

Thoughts?

madkinsz: I agree the simplification may be nice since it's user-facing but we'll probably want to create an example of making the decision based on reference tasks then.

zangell44: Yep this makes sense, I'll update the signature and docs accordingly

zangell44: Updated and rebased to clean up the rewrite. Let me know what you think. Is there a better example I could use for the docs?

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,38 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
 The handler is called anytime the flow's state changes, and receives the flow itself, the old state, and the new state. The state that the handler returns is used as the flow's new state.
 
 If multiple handlers are provided, they are called in sequence. Each one will receive the "true" `old_state` and the `new_state` generated by the previous handler.
+
+## Terminal State Handlers
+
+Flows allow users to provide custom logic for determining the final State of a Flow.
```
madkinsz, line 198:
```suggestion
Flows allow users to provide custom logic for determining the final State of a Flow.
Unlike normal handlers, the terminal handler has access to the final states of all of the tasks in the flow.
```

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,38 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
...
+
+```python
+terminal_state_handler(flow: Flow, state: State, task_states: Dict[Task, State]) -> Optional[State]
+```
+
+`flow` is the current Flow
+`state` is the current state of the Flow
+`task_states` contains states for Flow Tasks
+
+An example use case would be iterating through return states and indicating which reference tasks that have failed.
```
madkinsz, line 210:
```suggestion
For example, you may want to determine which reference tasks that have failed and add them to your final state message.
```

madkinsz: Note that this is particularly useful for Cloud Hooks (/ Automations) which report the state message. Now users can customize the success/failure message so their notifications include useful information about what has failed.

src/prefect/core/flow.py
```diff
@@ -160,6 +165,7 @@ def __init__(
         on_failure: Callable = None,
         validate: bool = None,
         result: Optional[Result] = None,
+        terminal_state_handler: Optional[Callable] = None,
```
madkinsz, line 168: Might be worth including the signature of `[Flow, State, List[State]], Optional[State]` 🤷 since it's relatively simple

tests/core/test_flow.py
```diff
@@ -3252,3 +3254,70 @@ def test_slugs_robust_to_task_name_changes(self):
...
+            state: State,
+            task_states: Dict[Task, State],
+        ) -> Optional[State]:
+            return None
+
+        with Flow("test", terminal_state_handler=fake_terminal_state_handler) as flow:
+            fake_task = Task("fake")
+            flow.add_task(fake_task)
+
+        assert flow.run().message == "All reference tasks succeeded."
```
madkinsz, line 3296: I have a minor preference for asserting the final state type is still `Success` as well instead of just the message

src/prefect/engine/flow_runner.py
```diff
@@ -717,6 +717,16 @@ def determine_final_state(
...
             state = Success(message="No reference tasks failed.", result=return_states)
 
+        if self.flow.terminal_state_handler:
+            state = (
+                self.flow.terminal_state_handler(
+                    flow=self.flow,
+                    state=state,
+                    task_states=return_states,
+                )
+                or state
```
madkinsz, line 727: Note this will do weird things if they return falsey values. I'm not sure it matters but an explicit `is not None` may expose mistakes as errors more clearly here.

zangell44:
Thanks for the detailed review @madkinsz ! Just pushed commits with the requested changes.

madkinsz:
No problem! This looks good. @zangell44 mypy is complaining because you're passing kwargs to the callable. I think you shouldn't do that anyway, passing them as positional makes sense -- the user can call them whatever they want.

zangell44:
I take back my previous comment, will update now

madkinsz (dismissed):
I think this qualifies as one of the few times mypy caught something that would have been confusing later on.

Nice work on this!

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,42 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
...
+
+Terminal state handlers must have the following signature:
+
+```python
+terminal_state_handler(flow: Flow, state: State, task_states: Dict[Task, State]) -> Optional[State]
+```
+
+`flow` is the current Flow
+`state` is the current state of the Flow
+`task_states` contains states for Flow Tasks
```
jcrist, line 210: Single newlines are ignored in markdown, you probably want this to be a list:

```suggestion
where:

- `flow` is the current Flow
- `state` is the current state of the Flow
- `task_states` contains states for Flow Tasks
```

src/prefect/core/flow.py
```diff
@@ -143,6 +144,11 @@ def my_task():
             the flow (e.g., presence of cycles and illegal keys) after adding the edges passed
             in the `edges` argument. Defaults to the value of `eager_edge_validation` in
             your prefect configuration file.
+        -  terminal_state_handler (Callable, optional): A state handler that mutates the final state
+            of the flow, with signature `state_handler(flow: Flow, state: State,
+            task_states: Dict[Task, State]) -> Optional[State]`. `flow` is the current Flow.
+            `state` is the current state of the Flow. `task_states` contains states for Flow Tasks.
+            If no state is returned, the original state will be used.
```
jcrist, line 151: Updated to clarify wording.

```suggestion
        -  terminal_state_handler (Callable, optional): A state handler that can be used to
            inspect or modify the final state of the flow run. Expects a callable with signature
            `handler(flow: Flow, state: State, task_states: Dict[Task, State]) -> Optional[State]`,
            where `flow` is the current Flow, `state` is the current state of the Flow run, and
            `task_states` is a mapping from task -> state for all tasks in the flow. It should
            return either a new state for the flow run, or `None` (in which case the existing
            state will be used).
```

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,42 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
 The handler is called anytime the flow's state changes, and receives the flow itself, the old state, and the new state. The state that the handler returns is used as the flow's new state.
 
 If multiple handlers are provided, they are called in sequence. Each one will receive the "true" `old_state` and the `new_state` generated by the previous handler.
+
+## Terminal State Handlers
+
+Flows allow users to provide custom logic for determining the final State of a Flow.
+
+Unlike normal state handlers, the terminal handler has access to the final states of all tasks in the flow.
```
jcrist, line 200:
```suggestion
A flow's `terminal_state_handler` allow users to provide custom logic for determining the
final State of a Flow. Unlike normal state handlers, the `terminal_state_handler` has access
to the final states of all tasks in the flow.
```

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,42 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
...
+terminal_state_handler(flow: Flow, state: State, task_states: Dict[Task, State]) -> Optional[State]
+```
+
+`flow` is the current Flow
+`state` is the current state of the Flow
+`task_states` contains states for Flow Tasks
+
+For example, you may want to determine which reference tasks that have failed and add them to your final state message.
+
+This is particularly useful for Cloud Hooks that report the state message.
```
jcrist, line 214:
```suggestion
For example, you may want to determine which reference tasks have failed and add them
to your final state message. This could be used to improve the message sent out by Cloud
Hooks that report the flow run's final state.
```

docs/core/concepts/flows.md
```diff
@@ -192,3 +192,42 @@ state_handler(flow: Flow, old_state: State, new_state: State) -> State
...
+    failed_tasks = []
+    for task, task_state in task_states.items():
+        if task_state.is_failed() and task in flow.reference_tasks():
+            failed_tasks.append(task.name)
+    # update the terminal state of the Flow and return
+    state.message = "The following tasks failed: {}".format(failed_tasks)
+    return state
+
+# create a new flow using the terminal state handler
+f = Flow("my flow with custom terminal state handler", terminal_state_handler=custom_terminal_state_handler)
```
jcrist, line 232:
```suggestion
flow = Flow(
    "my flow with custom terminal state handler",
    terminal_state_handler=custom_terminal_state_handler
)
```

jcrist (commented):
A few edits to docs, otherwise this looks good to me.

jcrist (approved):

PR #4313, opened 2021-03-29
update docs for switch from tokens to API keys

whimsicallyson:
## Summary
Updates the docs for the new auth API keys feature

## Changes
Docs!
**Note: since most of my work was on the UI side, I'm less confident that the CLI instructions are accurate. Please take special care reviewing those sections!

## Importance
With changes in auth/tokens/API keys, we want to make sure our users have a good reference of how to use them

## Checklist

This PR:


docs/orchestration/concepts/api.md
```diff
@@ -63,28 +59,24 @@ client.graphql(
 
 ### Authenticating the Client with Cloud <Badge text="Cloud"/>
 
-If using Prefect Cloud, authentication is required. For interactive use, the most common way to use the Cloud Client is to generate a `USER`-scoped token and provide it to the client. After doing so, users can save the token so it persists across all Python sessions:
+If using Prefect Cloud, authentication is required. For interactive use, the most common way to use the Cloud Client is to generate a user-based API key and provide it to the client. After doing so, users can save the key so it persists across all Python sessions:
 
 ```python
 import prefect
-client = prefect.Client(api_token="YOUR_USER_TOKEN")
+client = prefect.Client(api_token="YOUR_USER_KEY")
```
znicholasbrown, line 66: Maybe here (and elsewhere) we should call this `"API_KEY"` instead

docs/orchestration/concepts/api.md
```diff
@@ -63,28 +59,24 @@ client.graphql(
...
 
-Note that a token can be provided by environment variable (`PREFECT__CLOUD__AUTH_TOKEN`) or in your Prefect config (under `cloud.auth_token`).
-
-:::tip Using `USER` tokens
-The steps shown here were designed to be used with `USER`-scoped tokens. They will work with `TENANT` scoped tokens as well, but unexpected errors could result.
-:::
+Note that a key can be provided by environment variable (`PREFECT__CLOUD__AUTH_TOKEN`) or in your Prefect config (under `cloud.auth_token`).
 
-Once provisioned with a `USER` token, the Cloud Client can query for available tenants and login to those tenants. In order to query for tenants, call:
+Once provisioned with a user-based API key, the Cloud Client can query for available tenants and login to those tenants. In order to query for tenants, call:
```
znicholasbrown, line 79:
```suggestion
Once provisioned with an API key, the Cloud Client can query for available tenants and login to those tenants. In order to query for tenants, call:
```

docs/orchestration/concepts/api.md
```diff
@@ -136,7 +128,7 @@ For ease of interacting with the GraphQL API, the UI contains a full GraphQL cli
 
 ### Authentication <Badge text="Cloud"/>
 
-The Cloud API also supports direct GraphQL access. While you can still use `USER`-scoped tokens to access and log in to tenants, you will need to manage the short-lived auth and refresh tokens yourself. Therefore, we recommend using the Python client for `USER`-scoped access. You can make GraphQL queries directly from the client:
+The Cloud API also supports direct GraphQL access. While you can still use user-based API keys to access and log in to tenants, you will need to manage the short-lived auth and refresh tokens yourself. Therefore, we recommend using the Python client for user-scoped access. You can make GraphQL queries directly from the client:
```
znicholasbrown, line 131:
```suggestion
The Cloud API also supports direct GraphQL access. While you can still use API keys to access and log in to tenants, you will need to manage the short-lived auth and refresh tokens yourself. Therefore, we recommend using the Python client for user-scoped access. You can make GraphQL queries directly from the client:
```

docs/orchestration/concepts/api.md
whimsicallyson: is it confusing that both user and service accounts are now using API keys? Or do you think there's enough context that you can decide which one should be used?

docs/orchestration/concepts/api.md
whimsicallyson: same question here re: user vs service account API key, that's definitely why I worded it this way initially

docs/orchestration/concepts/api.md
```diff
@@ -150,10 +142,10 @@ client.graphql(
 )
 ```
 
-To use `TENANT`-scoped tokens in programmatic requests, include the token as the authorization header of your GraphQL requests:
+To use `TENANT`-scoped keys in programmatic requests, include the key as the authorization header of your GraphQL requests:
```
whimsicallyson, line 145: this line is still not 100% correct, since it references tenant-scoped. Should it be user or service account?

docs/orchestration/concepts/api.md
madkinsz: Can we do `API_KEY` and `SERVICE_ACCOUNT_API_KEY` where necessary? It seems like having the default be the user api key and the service account ones be the special case would help with clarity. Service accounts are just special users 🤷

whimsicallyson: that seems like a good heuristic to follow! 👍

docs/orchestration/agents/fargate.md
```diff
@@ -40,12 +40,12 @@ $ prefect agent fargate start
...
-There are a few ways in which you can specify a `RUNNER` API token:
+::: tip API Keys <Badge text="Cloud"/>
+There are a few ways in which you can specify a service account API key:
 
-- command argument `prefect agent fargate start -t MY_TOKEN`
-- environment variable `export PREFECT__CLOUD__AGENT__AUTH_TOKEN=MY_TOKEN`
-- token will be used from `prefect.config.cloud.auth_token` if not provided from one of the two previous methods
+- command argument `prefect agent fargate start -t SERVICE_ACCOUNT_API_KEY`
+- environment variable `export PREFECT__CLOUD__AGENT__AUTH_TOKEN=SERVICE_ACCOUNT_API_KEY`
+- key will be used from `prefect.config.cloud.auth_token` if not provided from one of the two previous methods
```
jcrist, line 48: "key" on its own doesn't mean much to me on its own - could you use `API key` everywhere a key is expected?

docs/orchestration/concepts/api.md
```diff
@@ -63,28 +59,24 @@ client.graphql(
 
 ### Authenticating the Client with Cloud <Badge text="Cloud"/>
 
-If using Prefect Cloud, authentication is required. For interactive use, the most common way to use the Cloud Client is to generate a `USER`-scoped token and provide it to the client. After doing so, users can save the token so it persists across all Python sessions:
+If using Prefect Cloud, authentication is required. For interactive use, the most common way to use the Cloud Client is to generate an API key and provide it to the client. After doing so, users can save the key so it persists across all Python sessions:
```
jcrist, line 62: Same here. Anywhere `key` is used alone should be `API key` IMO.

docs/orchestration/concepts/api.md
```diff
@@ -150,10 +142,10 @@ client.graphql(
 )
 ```
 
-To use `TENANT`-scoped tokens in programmatic requests, include the token as the authorization header of your GraphQL requests:
+To use keys in programmatic requests, include the key as the authorization header of your GraphQL requests:
```
jcrist, line 145: Same here.

jcrist (commented):
Only one comment, otherwise this looks good. Everywhere `key` is used alone we probably want to use `API key` (or `Service Account key`) instead to clarify what type of key we're talking about.

This will also require some core code changes to remove "token" from our language and deprecate all kwargs/config options with those terms. I'll take care of that separately from this PR.

docs/orchestration/concepts/api.md
```diff
@@ -63,28 +59,24 @@ client.graphql(
...
-assert client._api_token == "YOUR_USER_TOKEN"  # True
+assert client._api_token == "API_KEY"  # True
 ```
 
-Note that a token can be provided by environment variable (`PREFECT__CLOUD__AUTH_TOKEN`) or in your Prefect config (under `cloud.auth_token`).
-
-:::tip Using `USER` tokens
-The steps shown here were designed to be used with `USER`-scoped tokens. They will work with `TENANT` scoped tokens as well, but unexpected errors could result.
-:::
+Note that an API key can be provided by environment variable (`PREFECT__CLOUD__AUTH_TOKEN`) or in your Prefect config (under `cloud.auth_token`).
```
mashun4ek, line 77: I believe it should be `PREFECT__CLOUD__AGENT__AUTH_TOKEN`, right?

jcrist (approved):

jcrist:
Thanks @whimsicallyson! I'm going to merge this now and follow up with some further changes to the core codebase to remove/deprecate the use of "token" everywhere.

PR #4321, opened 2021-03-31
Expand logs in `DaskExecutor`

jcrist:
Adds the following logs:
- Scheduler address used when connecting to an existing cluster
- Callable used when creating a new cluster
- Dashboard address used for new cluster (if available)

We cannot log the dashboard address for an existing cluster, as this
can't be determined from the client.

### DIFF
--- a/changes/pr4321.yaml
+++ b/changes/pr4321.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Expand logging for `DaskExecutor`, including the cluster dashboard address (if available) - [#4321](https://github.com/PrefectHQ/prefect/pull/4321)"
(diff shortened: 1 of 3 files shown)


madkinsz (approved):
Any reason you're using old string formatting? Seems like it'd be nice to be consistent about using the new style. Otherwise lgtm.

jcrist:
> Any reason you're using old string formatting?

I'm not using string formatting at all (note no manual `%` operator usage) - log calls delay string formatting unless they're needed, old-style format strings are the only format supported.

madkinsz:
Ah I missed that you were doing delayed formatting. I generally don't think the performance implications matter for our use-case and I prefer the readability of new-style formatting but this still makes sense 👍

jcrist:
I agree the perf probably doesn't matter here, just habit on my part from other applications where it does matter.

Issue #4231, opened 2021-03-09
Allow for ShellTask to stream output with logging level is INFO


pedrounas:
## Current behavior
The ShellTask supports output streaming using the `stream_output = True` flag and by setting the logging level to debug using the environment variable `PREFECT__LOGGING__LEVEL=DEBUG`.

## Proposed behavior
Allow for the output to be streamed when the logging level is set to info by switching the `stream_output` flag to work with INFO logging level and introducing an alternative flag like `debug_output` to stream the output in DEBUG logging level.

## Example
When calling R Scripts that contain loggers using Prefect its not ideal to have to change the debug level in Prefect as it clutters the output of the flow and can hide the R logs.

## Proposed change

```python
  with Popen(
      [self.shell, tmp.name], stdout=PIPE, stderr=STDOUT, env=current_env
  ) as sub_process:
      line = None
      lines = []
      for raw_line in iter(sub_process.stdout.readline, b""):
          line = raw_line.decode("utf-8").rstrip()

          if self.return_all:
              lines.append(line)

          if self.stream_output:
              self.logger.info(line)
         
          if self.debug_output:
              self.logger.debug(line)

      sub_process.wait()
```

pedrounas:
Relates to #3865

madkinsz:
I'm going to close this as a duplicate of #3865 as I think `stream_output` should remain a single kwarg. I think it'd be reasonable for the default to be changed to `INFO` as well. Thanks for taking the time to add to this though :) more use-case / context is valuable.

jcrist:
Actually closing :)

PR #4322, opened 2021-03-31
Add ability to stream ShellTask logs with custom log level

### LINKED ISSUE #3865: Custom log_level for ShellTask
## Current behavior
All `ShellTask` logs are logged at `DEBUG` level, preventing showing the command output in the UI logs unless the agent's logging threshold is set to DEBUG. But this is undesired for production workloads, as this applies to all log messages

streaming logging for the `ShellTask` was introduced in #3584 

## Proposed behavior
change existing `stream_logs` parameter to `Union[bool, int]` to accept the desired log_level as suggest by @madkinsz.
see discussion at the bottom of #3584

## Example
see #3584

TaliaSRTR:
There are certain circumstances under which we'd like to be able to stream the output of a `ShellTask` (or its child classes) without having to set the global logging level to `DEBUG`; for example, the process launched by `ShellTask` sometimes streams valuable output that a user might like to capture and stream using the base Prefect logger. Currently, this is not possible. This change gives the user more control in that they can change the log level from one `ShellTask` to the next within a given flow.

Fixes #3865.

## Summary

Add ability to stream ShellTask logs with custom log level

## Changes

- Changes the default behavior of `ShellTask.stream_output`, which will default to using a log level of `INFO` when enabled.
- Changes the type of `ShellTask.stream_output` to be a `Union[bool, int, str]`. This will allow a user to pass in a log level (either as an int like `logging.INFO` or as a string like `"INFO"`) as an argument to the `ShellTask` constructor, thereby letting them decide how they want Prefect to log the underlying Shell process' output from one task to the next, without adjusting the global log level.

## Importance

There are certain Shell tasks (the immediate use case here was for the `DbtShellTask`) where it would be helpful to have the task's output _always_ streamed to the Prefect output log; not just when the logging level is set to `DEBUG`. `dbt`, for example, logs a lot of useful information to STDOUT during normal execution of the `dbt run` or `dbt test` commands. None of this output is forwarded to the Prefect logs if you're not running in `DEBUG` mode.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/shell.py
+++ b/src/prefect/tasks/shell.py
@@ -34,6 +34,9 @@ class ShellTask(prefect.Task):
         - stream_output (bool, optional): specifies whether this task should log
             the output as it occurs. If enabled, `log_stderr` will be ignored as the
             output will have already been logged. defaults to `False`
+        - log_stream_as_info (bool, optional): boolean specifying whether the
+            streamed output should be logged using `INFO` as its log level. Will be
+            ignored if `stream_output` has not been set to `True`; defaults to `False`.
         - **kwargs: additional keyword arguments to pass to the Task constructor
 
     Example:
@@ -60,6 +63,7 @@ class ShellTask(prefect.Task):
         return_all: bool = False,
         log_stderr: bool = False,
         stream_output: bool = False,
+        log_stream_as_info: bool = False,
         **kwargs: Any
     ):
         self.command = command
@@ -69,6 +73,7 @@ class ShellTask(prefect.Task):
         self.return_all = return_all
         self.log_stderr = log_stderr
         self.stream_output = stream_output
+        self.log_stream_as_info = log_stream_as_info
         super().__init__(**kwargs)
 
     @defaults_from_attrs("command", "env", "helper_script")
@@ -124,7 +129,10 @@ class ShellTask(prefect.Task):
                         lines.append(line)
 
                     if self.stream_output:
-                        self.logger.debug(line)
+                        if self.log_stream_as_info:
+                            self.logger.info(line)
+                        else:
+                            self.logger.debug(line)
 
                 sub_process.wait()
                 if sub_process.returncode:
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/shell.py
```diff
@@ -34,6 +34,9 @@ class ShellTask(prefect.Task):
         - stream_output (bool, optional): specifies whether this task should log
             the output as it occurs. If enabled, `log_stderr` will be ignored as the
             output will have already been logged. defaults to `False`
+        - log_stream_as_info (bool, optional): boolean specifying whether the
+            streamed output should be logged using `INFO` as its log level. Will be
+            ignored if `stream_output` has not been set to `True`; defaults to `False`.
```
jcrist, line 39: What if instead of adding a new kwarg, we modify `stream_output` to optionally take a log level to stream logs at? `True` would still correspond to `debug` (or should we move this to `info`? cc @madkinsz since you recently touched this code).

This gives users more control over the log level used and keeps the options more localized (one option to set, not two).

madkinsz: Yeah we've already discussed this at https://github.com/PrefectHQ/prefect/issues/3865 and https://github.com/PrefectHQ/prefect/issues/4231#issuecomment-794090520

I think it'd be worth changing the default to `INFO` as well.

src/prefect/tasks/shell.py
```diff
@@ -123,8 +125,11 @@ def run(
                     if self.return_all:
                         lines.append(line)
 
-                    if self.stream_output:
-                        self.logger.debug(line)
+                    if isinstance(self.stream_output, bool):
+                        if self.stream_output:
+                            self.logger.info(line)
```
madkinsz, line 130: I would instead do this determination at class initialization

```python
self.stream_output = logging.INFO if stream_output is True else stream_output
```

then here you can do

```python
if self.stream_ouput:
    self.logger.log(level=self.stream_output, msg=line)
```

madkinsz: or if we want to support a user patching the stream output property live e.g

```python
my_task = ShellTask()
my_task.stream_output = False
```

then we'd want to use a property for `stream_output` and have that simple `if` logic exist in the setter

madkinsz: It will return a string if not given an existing name so you could just check if it successfully cast to an int at the end

```
if self.stream_logs and not isinstance(self.stream_logs, int):
     raise TypeError(...)
```

jcrist (approved):
LGTM, thanks!

PR #4323, opened 2021-03-31
Update ~ ECS Fargate Agent: +network configuration param

renzobecerra:
## Summary
ECS Fargate Agent failing to infer network configuration when a default configuration does not exist. 

## Changes
Adding network configuration input parameter to ECS Agent so that network configuration can be specified upon agent start.

     Update src/prefect/cli/agent ~ Add --network-configuration input param to allow ECS Fargate Agent to use defined config instead of inferring default.

     Update src/prefect/agent/ecs/agent ~ Add --network-configuration input param handling, before inferring default network configuration.

## Importance
These changes are necessary for specifying network configuration when a default does not exist and therefore cannot be inferred.

## Checklist
Formatting: black and mypy, passing

This PR:

### DIFF
--- a/src/prefect/agent/ecs/agent.py
+++ b/src/prefect/agent/ecs/agent.py
@@ -126,7 +126,8 @@ class ECSAgent(Agent):
         - botocore_config (dict, optional): Additional botocore configuration
             options to be passed to the boto3 client. See [the boto3
             configuration docs][2] for more information.
-
+        - network_configuration (str, optional): The network configuration
+            to use, if not provided the default will be used.
 
     [1]: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html
 
@@ -153,6 +154,7 @@ class ECSAgent(Agent):
         task_role_arn: str = None,
         execution_role_arn: str = None,
         botocore_config: dict = None,
+        network_configuration: str = None,
     ) -> None:
         super().__init__(
             agent_config_id=agent_config_id,
@@ -171,6 +173,7 @@ class ECSAgent(Agent):
         self.launch_type = launch_type.upper() if launch_type else "FARGATE"
         self.task_role_arn = task_role_arn
         self.execution_role_arn = execution_role_arn
+        self.network_configuration = network_configuration
 
         # Load boto configuration. We want to use the standard retry mode by
         # default (which isn't boto's default due to backwards compatibility).
@@ -229,9 +232,12 @@ class ECSAgent(Agent):
         if self.launch_type == "FARGATE" and not self.run_task_kwargs.get(
             "networkConfiguration"
         ):
-            self.run_task_kwargs[
-                "networkConfiguration"
-            ] = self.infer_network_configuration()
+            if isinstance(self.network_configuration, str):
+                network_config = yaml.safe_load(self.network_configuration)
+            else:
+                network_config = self.infer_network_configuration()
+
+            self.run_task_kwargs["networkConfiguration"] = network_config
 
     def infer_network_configuration(self) -> dict:
         """Infer default values for `networkConfiguration`.
--- a/src/prefect/cli/agent.py
+++ b/src/prefect/cli/agent.py
@@ -398,6 +398,11 @@ def ecs():
     "run_task_kwargs_path",
     help="Path to a yaml file containing extra kwargs to pass to `run_task`",
 )
+@click.option(
+    "--network-configuration",
+    "network_configuration",
+    help="The network configuration to use. If not provided, your default configuration will be used",
+)
 def start(**kwargs):
     """Start an ECS agent"""
     from prefect.agent.ecs import ECSAgent


jcrist:
Thanks for the PR @renzobecerra. You can already specify this config via the `--run-task-kwargs` flag (which points to a file containing any arbitrary configuration to set at runtime). Since `networkConfiguration` is a nested structure which would require passing in json/YAML on the commandline (something that's hard to do properly), I'd rather stick with `--run-task-kwargs` for handling this. Thoughts?

An example in the docs of providing network configuration via `--run-task-kwargs` would be nice though, would you be interested in adding that?

renzobecerra:
@jcrist 

> Thanks for the PR @renzobecerra. You can already specify this config via the `--run-task-kwargs` flag (which points to a file containing any arbitrary configuration to set at runtime). Since `networkConfiguration` is a nested structure which would require passing in json/YAML on the commandline (something that's hard to do properly), I'd rather stick with `--run-task-kwargs` for handling this. Thoughts?
> 
> An example in the docs of providing network configuration via `--run-task-kwargs` would be nice though, would you be interested in adding that?

Hey Jim! We've attempted to pass in these network configs via --run-task-kwargs using a yaml file. The agent starts up fine, but fails @ register_task_definitions() (Yaml is structured correctly and verified correct by prefect team). Not sure if merge_run_task_kwargs() is giving precedence/merging correctly here or if additional params are required in --run-task-kwargs during the creation of a task/service.

Because many of our users will be spinning up ecs agents with different network configs (we do not have a default config set in aws) it makes sense (from my perspective), to include a VPC network config parameter either at agent start via --network-configuration or (other param) in the same way we can specify clusters, role-arns and launch_types at agent start; thus increasing orthogonality for the agent.

Thoughts?

The vpc flag idea makes sense, however we still would not be able to infer defaults; being that we have multiple security groups and subnets per vpc.

renzobecerra:
@jcrist 

> Alternatively, are there non-yaml config flags we could add to the agent to make the inference process succeed as written? Perhaps a `--vpc` flag to infer from a specified vpc rather than the default?

If a yaml config doesn't make sense here, maybe we could specify list obj params to pass in --vpcs --subnets & --security-groups to be returned @ infer_network_configuration()

Thoughts?

jcrist:
> We've attempted to pass in network configs via --run-task-kwargs using a yaml file. The agent starts up fine, but fails @ register_task_definitions() (The Yaml file is structured correctly).

Can you provide the file config you used (you can replace anything you consider sensitive with xxxx, just looking for the structure of what you configured). Can you also reply with the error message you get and context on when/where it shows up? Providing network configuration in this way has worked fine for other users.

> If a yaml config doesn't make sense here, maybe we could specify list obj params to pass in (--vpc --subnets & --security-groups) to be passed to infer_network_configuration()

I could see an argument for adding these extra flags, but I'd like to ensure things work fine with `--run-task-kwargs` as well. AWS has a _huge_ number of options, we can't possibly add top-level flags for all of them. I'm fine adding specific flags for common things if it makes it easier for users, but it's a slippery slope and would prefer to avoid it when possible. It sounds like having CLI flags for these would be useful for your current team - in your experience do you think other users would also commonly want these flags, or is your case more unusual than most?

renzobecerra:
> > We've attempted to pass in network configs via --run-task-kwargs using a yaml file. The agent starts up fine, but fails @ register_task_definitions() (The Yaml file is structured correctly).
> 
> Can you provide the file config you used (you can replace anything you consider sensitive with xxxx, just looking for the structure of what you configured). Can you also reply with the error message you get and context on when/where it shows up? Providing network configuration in this way has worked fine for other users.
> 
> > If a yaml config doesn't make sense here, maybe we could specify list obj params to pass in (--vpc --subnets & --security-groups) to be passed to infer_network_configuration()
> 
> I could see an argument for adding these extra flags, but I'd like to ensure things work fine with `--run-task-kwargs` as well. AWS has a _huge_ number of options, we can't possibly add top-level flags for all of them. I'm fine adding specific flags for common things if it makes it easier for users, but it's a slippery slope and would prefer to avoid it when possible. It sounds like having CLI flags for these would be useful for your current team - in your experience do you think other users would also commonly want these flags, or is your case more unusual than most?

--netConfig yaml below-- Not sure if this is related to boto 0.14.12, but when I spoke with someone on the prefect team they mentioned this was related to https://github.com/PrefectHQ/prefect/pull/4325.

I completely understand the slippery slope concern-- In my experience, most companies have at least two subnets and multiple security groups. We're POC'ing a prefect cloud solution for multiple orgs at my company. For our use case, having the ability to set these vpc parameters at agent start would:
1. Simplify the creation and management of agents by eliminating the need to keep track of multiple yaml files
2. Make ECS Agents (service/tasks) more intuitive and practical to spin up (for individuals who are not familiar with ECS)
3. Leverage ECSRun to set run-task-kwargs params

I like the ability to specify a --run-task-kwargs config file at agent start, but I imagine a vpc param use case is a common one, especially for large corporations.

If we do set up these parameters, would it make sense to pass awsvpcConfiguration params individually ie. subnets and security groups as list objects; instead of passing and parsing a yaml string?


networkConfiguration:
    awsvpcConfiguration:
        assignPublicIp: 'DISABLED'
        subnets:
            - subnet-xxxxxxxxx
        securityGroups:
            - sg-xxxxxxxxx
            - sg-xxxxxxxxx

madkinsz:
Hey @renzobecerra -- Jim is out of office for a bit and I'm not sure I can keep moving this PR forward but I also would like to figure out why `run-task-kwargs` is failing for you. Could you provide the error message you encountered when the task definitions were registered?

madkinsz:
Closing this from lack of info

The ECS configuration has proven a bit tricky and will need to be revisted. See also
#4884
#4413 
#4704 
#4803

PR #4326, opened 2021-03-31
Add architecture diagram to docs

zangell44:
## Summary
Adds architecture diagram to documentation, along with a brief summary of the components.

Would love feedback on high level questions -

- Is this the right place to put this information?

Several places were suggested in the original issue thread. "Overview" of the Orchestration docs seems like a natural place to put it, with the other introductory info about Prefect Cloud.

- Is the explanation's level of abstraction correct?

My assumption here is that users can dig in on further detail as needed (and links have been included). I think it makes sense to briefly explain the WHAT of the main architecture components.

## Changes
Adds architecture diagram to documentation, along with a brief summary of the components.

## Importance
Our hybrid model can be confusing at first, this diagram and explanation will help users get up to speed faster.

## Checklist

This PR:


docs/orchestration/README.md
```diff
@@ -44,6 +44,22 @@ Prefect Cloud is a fully hosted, production-ready backend for Prefect Core. If y
 
 As you read through these docs, when you see the <Badge text="Cloud"/> badge, you'll know if a feature is supported only in Prefect Cloud or not.
 
+##### Prefect Cloud Architecture Overview
```
jcrist, line 47:
```suggestion
##### Architecture Overview
```

Since this also (mostly) applies to Server.

jcrist (approved):
One small tweak, otherwise this looks good. This isn't really the best place for a diagram like this (and the diagram is more marketing-y than technical, a deep user would likely want more detail). Still better than the nothing we have up now though :). Thanks!

jcrist (approved):

PR #4332, opened 2021-04-01
Support local directory imports in flow scripts from CLI

jcrist:
When using `prefect register`/`prefect build` we now support flows that
import files from other files in the same directory. Note that this does
_not_ ensure that these flows will run successfully (as the presence of
these "helper scripts" would have to be ensured by the storage/runtime
environment). All this does is ensure that running `prefect register
--path path/to/flow.py` has the same `sys.path` as `python
path/to/flow.py` would have.

We also move to using `runpy` to properly setup some other metadata
values in the script's namespace (e.g. `__file__`).

Reported on slack: https://prefect-community.slack.com/archives/CL09KU1K7/p1617276925312300?thread_ts=1617160663.217800&cid=CL09KU1K7

This PR:

### DIFF
--- a/changes/pr4332.yaml
+++ b/changes/pr4332.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Support imports from local directory when registering/building flows via CLI - [#4332](https://github.com/PrefectHQ/prefect/pull/4332)"
--- a/src/prefect/cli/build_register.py
+++ b/src/prefect/cli/build_register.py
@@ -1,11 +1,12 @@
 import functools
+import hashlib
 import importlib
 import json
 import multiprocessing
 import os
+import runpy
 import sys
 import time
-import hashlib
 import traceback
 from collections import Counter, defaultdict
 from typing import Union, NamedTuple, List, Dict, Iterator, Tuple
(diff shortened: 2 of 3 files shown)


madkinsz (approved):
I feel like it was almost good that it failed before because then the user learns they have to add the PYTHONPATH (which they will likely have to do for the agent) but this is still probably the right choice

semcha:
Hi @madkinsz! Do you mean `--import-path` option in `prefect agent local start`?

mianos:
If I am building a flow with "prefect build  -p flows -n ptest-flow -o deployment/ptest-flow.json" and my file is flows/ptest-flow.py (I have a bunch of them that all get read the same way, 
in 
``` 
namespace = runpy.run_path(path, run_name="<flow>")
```
path is 'flows/ptest-flow.py'
The path of the filename from the 'flows' directory, so I still have to add: sys.path.append(os.path.abspath(os.getcwd())) in every flow.
Is there some way around this with this change?

mianos:
My mistake. The run path is the flow directory. I'll move my extensions to the flow directory. They are cleaner down there anyway.

PR #4333, opened 2021-04-01
Add bitbucket cloud

wrightjonathan:
## Summary
Added support for Bitbucket cloud into Bitbucket storage class

## Changes
Added three class variables to Bitbucket class (workspace, cloud_username_secret, cloud_app_password_secret). When constructed with a workspace the class will switch to using the bitbucket cloud client; without it the behaviour for bitbucket server remains unchanged.

The retry adapter is not required and could be removed. Our Bitbucket cloud API gets a lot of requests and the limit is 5000 per hour.

## Importance
A lot of people and businesses will be using Bitbucket cloud and not hosting their own bitbucket servers. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -145,7 +145,9 @@ class Bitbucket(Storage):
                     ref = tag_hashes[ref]
 
             if not re.match(SHA1_REGEX, ref):
-                raise ValueError("ref not found in this Storage")
+                raise ValueError(
+                    "ref {ref!r} not found in '{self.workspace}/{self.repo}'"
+                )
 
             contents_url = (
                 f"repositories/{self.workspace}/{self.repo}/src/{ref}/{flow_location}"
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -5,11 +5,21 @@ from urllib.error import HTTPError
 
 import requests
 from requests.exceptions import HTTPError as RequestsHTTPError
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 import prefect
 from prefect.client import Secret
 from prefect.storage import Storage
 from prefect.utilities.storage import extract_flow_from_file
+
+
+BITBUCKET_CLOUD_RETRY = Retry(
+    total=10,
+    backoff_factor=0.5,
+    status_forcelist=[429],
+    method_whitelist=["HEAD", "GET", "OPTIONS"],
+)
 
 if TYPE_CHECKING:
     import atlassian
@@ -292,16 +302,7 @@ class Bitbucket(Storage):
 
         # Bitbucket clould api is rate limited, retry at this frequency
         # 0.25, 0.5, 1, 2, 4, 8, 16, 32, 64, 128
-        from requests.adapters import HTTPAdapter
-        from requests.packages.urllib3.util.retry import Retry
-
-        retry_strategy = Retry(
-            total=10,
-            backoff_factor=0.5,
-            status_forcelist=[429],
-            method_whitelist=["HEAD", "GET", "OPTIONS"],
-        )
-        adapter = HTTPAdapter(max_retries=retry_strategy)
+        adapter = HTTPAdapter(max_retries=BITBUCKET_CLOUD_RETRY)
         session = requests.Session()
         session.mount("https://", adapter)
 
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -145,9 +145,7 @@ class Bitbucket(Storage):
             contents_url = (
                 f"repositories/{self.workspace}/{self.repo}/src/{ref}/{flow_location}"
             )
-            self.logger.info(
-                f"Downloading flow from Bitbucket cloud - {contents_url}"
-            )
+            self.logger.info(f"Downloading flow from Bitbucket cloud - {contents_url}")
             contents = client.get(contents_url)
 
         except RequestsHTTPError as err:
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -278,19 +278,16 @@ class Bitbucket(Storage):
                 "Unable to import atlassian, please ensure you have installed the bitbucket extra"
             ) from exc
 
-        if self.cloud_username_secret is not None:
-            # If cloud username secret specified, load it
-            cloud_username = Secret(self.cloud_username_secret).get()
-        else:
-            if cloud_username is None:
-                cloud_username = os.getenv("BITBUCKET_CLOUD_USERNAME")
-
-        if self.cloud_app_password_secret is not None:
-            # If cloud app password secret specified, load it
-            cloud_app_password = Secret(self.cloud_app_password_secret).get()
-        else:
-            if cloud_app_password is None:
-                cloud_app_password = os.getenv("BITBUCKET_CLOUD_APP_PASSWORD")
+        cloud_username = (
+            Secret(self.cloud_username_secret).get()
+            if self.cloud_username_secret is not None
+            else os.getenv("BITBUCKET_CLOUD_USERNAME")
+        )
+        cloud_app_password = (
+            Secret(self.cloud_app_password_secret).get()
+            if self.cloud_app_password_secret is not None
+            else os.getenv("BITBUCKET_CLOUD_APP_PASSWORD")
+        )
 
         # Bitbucket clould api is rate limited, retry at this frequency
         # 0.25, 0.5, 1, 2, 4, 8, 16, 32, 64, 128
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -82,6 +82,11 @@ class Bitbucket(Storage):
         self.cloud_username_secret = cloud_username_secret
         self.cloud_app_password_secret = cloud_app_password_secret
 
+        if self.workspace and self.access_token_secret:
+            raise ValueError(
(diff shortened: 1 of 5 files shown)


wrightjonathan:
Hi, I can't see the reason for the static_analysis check failure. Please let me know what it is.

jcrist:
Can't see meaning you don't have access to view https://app.circleci.com/pipelines/github/PrefectHQ/prefect/8108/workflows/6e05cced-0dc3-453c-a43f-8a2ad7f98693/jobs/116628? Or don't know how to interpret the error logs here?

wrightjonathan:
Ah, I've logged in now and can see the issues. I incorrectly assumed I didn't have access. Thank you.

src/prefect/storage/bitbucket.py
```diff
@@ -38,35 +40,47 @@ class Bitbucket(Storage):
...
     ) -> None:
         self.project = project
         self.repo = repo
+        self.workspace = workspace
         self.host = host
         self.path = path
         self.ref = ref
         self.access_token_secret = access_token_secret
+        self.cloud_username_secret = cloud_username_secret
+        self.cloud_app_password_secret = cloud_app_password_secret
```
madkinsz, line 83: Could you add a warning if these are set and workspace is not and vice-versa? Basically

- if workspace is set but the cloud auth is not
- if workspace is not set but the cloud auth is

with the goal of indicating to the user which fields they'll need to set for things to work. If authentication is not _required_ for BitbucketCloud then nevermind. We may want to raise an exception instead of a warning if it'll always fail at runtime.

src/prefect/storage/bitbucket.py
```diff
@@ -80,14 +94,85 @@ def get_flow(self, flow_name: str) -> "Flow":
         Returns:
             - Flow: the requested flow
         """
+
+        return (
+            self._get_flow_cloud(flow_name)
```
madkinsz, line 99: I have a minor preference for `_get_flow_from_bitbucket_cloud` / `_get_flow_from_bitbucket_server` to make it clear we are not referring to Prefect Server/Cloud

src/prefect/storage/bitbucket.py
```diff
@@ -80,14 +94,85 @@ def get_flow(self, flow_name: str) -> "Flow":
...
+                )
+                tag_hashes = {
+                    tag["name"]: tag["target"]["hash"] for tag in tags["values"]
+                }
+
+                if ref in tag_hashes:
+                    ref = tag_hashes[ref]
+
+            if not re.match(SHA1_REGEX, ref):
+                raise ValueError("ref not found in this Storage")
```
madkinsz, line 143:
```suggestion
            if not re.match(SHA1_REGEX, ref):
                raise ValueError("ref {ref!r} not found in '{self.workspace}/{self.repo}'")
```

src/prefect/storage/bitbucket.py
```diff
@@ -80,14 +94,85 @@ def get_flow(self, flow_name: str) -> "Flow":
...
+            )
+            self.logger.info(f"Downloading flow from Bitbucket cloud - {contents_url}")
+            contents = client.get(contents_url)
+
+        except RequestsHTTPError as err:
+            if err.response.status_code == 403:
+                self.logger.error(
+                    "Access denied to repository. Please check credentials."
+                )
+                raise
```
madkinsz, line 156:
```suggestion
                err_msg = "Access denied to repository. Please check credentials."
                self.logger.error(err_msg)
                raise RuntimeError(err_msg) from exc
```

I would do something like this instead so the error is displayed with our message. This would also allow you to easily consolidate your log and re-raise and just set the `msg` in your `if/else`

src/prefect/storage/bitbucket.py
madkinsz: Ah after getting further down I see these will be loaded from the environment if not set :D 

I would change this to

- if self.workspace is set and self.access_token_secret is set _then_ throw an exception saying they've set the wrong secret variable

src/prefect/storage/bitbucket.py
```diff
@@ -167,3 +252,55 @@ def _get_bitbucket_client(self) -> "atlassian.Bitbucket":
...
+                cloud_username = os.getenv("BITBUCKET_CLOUD_USERNAME")
+
+        if self.cloud_app_password_secret is not None:
+            # If cloud app password secret specified, load it
+            cloud_app_password = Secret(self.cloud_app_password_secret).get()
+        else:
+            # Otherwise, fallback to loading from local secret or environment variable
+            cloud_app_password = prefect.context.get("secrets", {}).get(
+                "BITBUCKET_CLOUD_APP_PASSWORD"
+            )
```
madkinsz, line 282: This is handled by the `Secret` class and depends on a config flag (`use_local_secrets`). I think you should remove this fallback and the one for the username. I believe the `os.getenv` fallback is still in-line with our typical pattern for something like this so that's okay.

src/prefect/storage/bitbucket.py
```diff
@@ -167,3 +252,55 @@ def _get_bitbucket_client(self) -> "atlassian.Bitbucket":
...
+            if cloud_app_password is None:
+                cloud_app_password = os.getenv("BITBUCKET_CLOUD_APP_PASSWORD")
+
+        # Bitbucket clould api is rate limited, retry at this frequency
+        # 0.25, 0.5, 1, 2, 4, 8, 16, 32, 64, 128
+        from requests.adapters import HTTPAdapter
+        from requests.packages.urllib3.util.retry import Retry
```
madkinsz, line 289: These imports should be fine at the top of the module, any particular reason you inlined them?

src/prefect/storage/bitbucket.py
```diff
@@ -167,3 +252,55 @@ def _get_bitbucket_client(self) -> "atlassian.Bitbucket":
...
+
+        retry_strategy = Retry(
+            total=10,
+            backoff_factor=0.5,
+            status_forcelist=[429],
+            method_whitelist=["HEAD", "GET", "OPTIONS"],
+        )
+        adapter = HTTPAdapter(max_retries=retry_strategy)
```
madkinsz, line 297: Since this is a constant it may be worth declaring it at the top of the module as `RETRY_ADAPTER` to simplify this function.

madkinsz (changes requested):

madkinsz:
Thanks for contributing @wrightjonathan -- I've got some requests to simplify this a bit but this is a great start!

src/prefect/storage/bitbucket.py
wrightjonathan: Done

src/prefect/storage/bitbucket.py
wrightjonathan: Done

src/prefect/storage/bitbucket.py
wrightjonathan: Done

src/prefect/storage/bitbucket.py
wrightjonathan: I've removed this and the one for the username. 

I had copied this from _get_bitbucket_server_client() (which I didn't write and have not changed)

On our infrastructure (Prefect Server) we will be relying on `os.getenv`

src/prefect/storage/bitbucket.py
wrightjonathan: I've moved them to the top.

I was in two minds whether to remove the retry adapter completely.

src/prefect/storage/bitbucket.py
wrightjonathan: Done

src/prefect/storage/bitbucket.py
madkinsz: Ah sorry about that -- looks like it might just be old pattern there.

src/prefect/storage/bitbucket.py
madkinsz: Yeah I'm not sure how I feel about it either. It seems reasonable and a more robust implementation seems like the way to go.

src/prefect/storage/bitbucket.py
wrightjonathan: Shall I remove it from the other function?

src/prefect/storage/bitbucket.py
wrightjonathan: It does work. I suggest we keep it. Our Bitbucket Cloud API is under heavy load and it does improve reliability a bit. But of course it is possible all the retries also fail.

src/prefect/storage/bitbucket.py
madkinsz: I can do it in a follow-up PR. I think it's better to isolate a change to the existing bit.

wrightjonathan: That makes sense.

src/prefect/storage/bitbucket.py
```diff
@@ -38,35 +50,53 @@ class Bitbucket(Storage):
...
         self.path = path
         self.ref = ref
         self.access_token_secret = access_token_secret
+        self.cloud_username_secret = cloud_username_secret
+        self.cloud_app_password_secret = cloud_app_password_secret
+
+        if self.workspace and self.access_token_secret:
+            raise ValueError(
+                "Wrong secret variable set. Workspace variable implies Bitbucket Cloud "
+                "but an access token for Bitbucket Server was provided."
```
madkinsz, line 98:
```suggestion
                "`access_token_secret` and `workspace` are both set. Setting `workspace` "
                "is only for use with Bitbucket Cloud which uses `cloud_username_secret` "
                "and `cloud_app_password_secret` to provide authentication.
```

madkinsz:
One last requested change to clarify what they should do to resolve that error.

madkinsz (approved):

PR #4334, opened 2021-04-01
Speedup flushing of cloud logs on shutdown

jcrist:
When writing logs to prefect cloud, we batch uploads into batches by a
time period (defaults to 5 seconds). On shutdown, the intent was to
break out of this write-then-sleep loop and write all the remaining logs
before closing.  Unfortunately this accidentally relied on the shutdown
signal interrupting the sleeping thread, leading to a slower shutdown
process on some platforms. When running on dask, the worker process must
shutdown in 3 seconds before a kill signal is sent, and since 3 < 5,
some logs could be lost. We now use a `threading.Event` instead of
sleeping, ensuring that the write-then-sleep loop breaks early on
shutdown and all logs are written.

Issue reported on slack here: https://prefect-community.slack.com/archives/CL09KU1K7/p1616957545069800


cicdw (approved):
awesome catch

PR #4341, opened 2021-04-02
Refactor `Agent` base class for readability

madkinsz:
## Summary

The git diff for this is hard to parse, but this should have nearly no functional changes. See the whole new file at https://github.com/PrefectHQ/prefect/blob/agent-cleanup/src/prefect/agent/agent.py.

I renamed a lot of opaque methods like `agent_process` (-> `_submit_deploy_flow_run_jobs`) and `agent_connect` (-> `_setup_api_connection`) to describe what they are doing. The interface that would generally be worked with has been moved to the top of the class. Many methods that should not be touched by subclasses or callers have been given a leading underscore. The `start` method has been simplified in an attempt to make it very clear what agent setup looks like. 

```python

def start():
    try:
        self._setup_api_connection()

        # Call subclass hook
        self.on_startup()

        # Print some nice startup logs
        self._show_startup_display()

        # Start background tasks
        self._start_heartbeat_thread()
        if self.agent_address:
            self._start_agent_api_server()

        # Enter the main loop checking for new flows
        with exit_handler(self) as exit_event:
            self._enter_work_polling_loop(exit_event)

    finally:

        self.on_shutdown()
        self._stop_agent_api_server()
        self._stop_heartbeat_thread()
```

## Changes

The only functional changes should be
- If the API errors during setup, the agent exits immediately instead of attempting to query for flow runs. This means that token errors and such will be surfaced as a clear error instead of being logged and buried as a later process fails.

## Importance

Before tackling https://github.com/PrefectHQ/server/issues/213 I wanted to understand the base agent process. I found it incredibly difficult to parse so I began a refactor that should make it more maintainable in the future.

## Checklist

This PR:


src/prefect/agent/agent.py
```diff
@@ -124,196 +136,294 @@ def __init__(
...
-            raise AuthorizationError("No agent API token provided.")
 
-        # Check if RUNNER role
-        result = self.client.graphql(query="query { auth_info { api_token_scope } }")
-        if (
-            not result.data  # type: ignore
-            or result.data.auth_info.api_token_scope != "RUNNER"  # type: ignore
-        ):
-            raise AuthorizationError("Provided token does not have a RUNNER scope.")
+        try:
```
zangell44, line 176: do we need the `try` block here? if so, what is its function?

zangell44: edit: nvm not used to Github diffs yet lol

src/prefect/agent/agent.py
```diff
@@ -109,6 +110,17 @@ class Agent:
             and all deployed flow runs
     """
 
+    # Loop intervals for query sleep backoff
+    _loop_intervals = {
```
zangell44, line 114: is it common for users to write their own agent classes? I'm wondering how likely it is we'd break calls to `Agent.start` that provide `_loop_intervals` as an arg

zangell44:
One minor question but this looks good to me. The refactor makes this much, much more readable for me as someone who hasn't read through the agent code before!

src/prefect/agent/agent.py
madkinsz: afaik it is not a supported thing. That argument is "private" and exposed just for testing, we generally don't want users to be hitting the API at a different set of intervals.

jcrist (dismissed):
Seems like an improvement to me. Looks like you have a few small merge conflicts, then happy to merge.

madkinsz:
This now refactors the work in #4158 as well

madkinsz:
Although no agent sub-type should differ, I've QA'd the `LocalAgent` and `DockerAgent` extensively.

@mashun4ek can you report back when you get a chance to QA the `FargateAgent`?

@TylerWanner could you do a quick QA of the `KubernetesAgent` ? Notably, I'd like to make sure failure-cases (ie your flow fails to deploy to K8s correctly) are handled correctly still

TylerWanner:
yeah i can come up with and run some tests for the k8s agent with the CLI install and the terraform agent configs by this evening @madkinsz

mashun4ek:
- [x] ECSRun flow with Docker storage(image in docker hub)
- [ ] ECSRun flow with Docker storage(using ECR) having error `An error occurred (ClientException) when calling the RegisterTaskDefinition operation: Role is not valid`, but it might be an issue on my end, I'll update shortly
- [ ] With custom task definition
Let me know if you want to test additional edge cases

TylerWanner:
Tested the Kubernetes agent locally in certain scenarios including properly handling k8s failures, logs, using custom job templates, resource management, etc. Seems to be working well

zangell44 (approved):

jcrist (approved):

PR #4346, opened 2021-04-04
Add safe-to-evict annotation to k8s job

### LINKED ISSUE #3058: Kubernetes Cluster AutoScaler can result in failed Flow Runs
## Description
This only happens on K8 clusters with [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to dynamically provision the K8 nodes based on current workload demands in the cluster. In my case, I've observed this on AWS EKS with manual NodeGroups.

The Cluster Autoscaler will evaluate the utilization of each Node in the K8 cluster to determine if it can scale down the number of Nodes. By default this happens every 10 minutes (configurable by `--scale-down-unneeded-time flag`), and is determined by a number of factors, the main one being the sum of cpu and memory requests of all pods running on this node is smaller than 50% of the node's allocatable (configurable by `--scale-down-utilization-threshold`). There's also a back-off period after a Node is first added to the cluster of 10 minutes (configurable by `--scale-down-delay-after-add flag`).

This functionality of K8 assumes that all Pods created by a controller object (deployment, replica set, **job**, stateful set etc) are ephemeral, and can be relocated to a different Node, with zero impact to the functionality of those Pods. The Prefect Kubernetes Agent triggers Flow Runs via K8 Jobs (effectively declaring them as ephemeral), which spawn the Pods to perform the actual execution. 

The problem is intermittent, and only comes into play under the following situation:

1. A new, long-running (i.e. >21 minutes), Prefect Flow Run is scheduled, and picked up by a Prefect Agent running in K8 w/Cluster Autoscaler
2. Prefect Agent creates a K8 Job, but there are no available K8 Nodes to run this Job within 1-5 minutes
3. Cluster Autoscaler scales up the Nodes in K8. The Node it scales up has at least 2x cpu/memory allocatable than what the single Prefect Flow Run would consume.
4. The Prefect Flow Run is assigned to the new Node Cluster Autoscaler just started
5. Capacity frees up on the other Nodes in the K8 cluster, such that no other Pods are scheduled on the new Node the Cluster Autoscaler just started (in step 3)
6. After 20 minutes, the Cluster Autoscaler determines that the Node from step 3 is a candidate to scale down because it meets the [criteria](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) (i.e. sum of cpu/memory of all pods is smaller than 50%, the pods were created by a controller object, etc.)
7. The Pod (which is executing the Flow Run) is evicted from this Node and rescheduled to a different Node in K8, and the Cluster Autoscaler Node is removed from K8

Now the fun part:

1. Because K8 evicted the Pod, Prefect Cloud will attempt to keep track of the original Pod (which no longer exists)
2. When the evicted Pod starts again, it will begin the Flow Run execution again ... because this was a long-running Flow Run, it'll take it >21 minutes to complete
3. Eventually Prefect will encounter `No heartbeat detected from the remote task; marking the run as failed.` for the initial Pod it created, which was since relocated by the Cluster AutoScaler.
4. You can observe the newly started Pod is actually reporting logs to Prefect Cloud, so it's curious why it detected no heartbeat from that new process. I'm guessing it's looking for some metadata to identify a heartbeat from the original Pod IP.
5. Because no heartbeat was detected from that original Pod, the Lazarus process will attempt to reschedule the Flow Run .. but the Flow Run was already restarted by K8, it's just in a different Pod now
6. This confusion will result in the Flow Run being marked as "Failed"

While less likely, it's feasible for a Flow Run to be scheduled on a Node that has been up for awhile, but is a candidate to be scaled down in the next 1 minute. Then after ~1 minute of executing the Flow Run on this Node, capacity on other Nodes could free up, such that the Cluster AutoScaler can determine to relocate all the running Pods, which would result in the same type of failure. In this case, the failure would happen to any Flow Run, regardless of execution time.

## Expected Behavior

The root of the problem is that the Cluster AutoScaler is moving the Pod created by the Job the Prefect Agent generated for this Flow Run, before that Flow Run had time to complete. It seems the scheduler may not be capable of handling situations where the Flow Run is restarted by processes outside of it's control.

A simple way to prevent the Cluster AutoScaler from attempting to evict a Pod from a Node, would be to add this annotation to it's manifest:

```
"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
```

If the AutoScaler see's this annotation on a Pod, it will not consider the Node for scaling down. In theory you could add this to a `job_spec_file`, but it'd be exhausting to do this for every Flow you want to execute in a K8 w/Cluster AutoScaler.

## Reproduction
There's a public flow that simply sleeps for a few minutes:
https://hub.docker.com/repository/docker/szelenka/long-running-flow

In AWS, I have an EKS cluster with a simple NodeGroup of t3.2xlarge, where the Prefect Agent spawns Jobs with 2 vCPU and 2G RAM. Initially, we only have the Prefect Agent running on a single Node in K8.

Through Prefect Cloud, we start 3 Flow Runs to run at 5 minutes each. This fills up the capacity of the single Node in K8. 

Then we start another Flow Run, with a run time of 25 minutes. This causes the Cluster AutoScaler to scale up a Node in K8, and assign this Job to that new Node.

Because it takes longer than 20 minutes to complete, and the other Flow Runs have completed, it will be evicted and relocated to the other Node that capacity freed up on, and trigger the above failure.

## Environment
```json
{
  "config_overrides": {
    "cloud": {
      "use_local_secrets": true
    },
    "context": {
      "secrets": false
    }
  },
  "env_vars": [],
  "system_information": {
    "platform": "macOS-10.15.6-x86_64-i386-64bit",
    "prefect_version": "0.12.6",
    "python_version": "3.8.1"
  }
}
```

Xinyi2016:
## Summary

Closes #3058 

## Changes

Added `--safe-to-evict` CLI argument and `safe_to_evict` to Kubernetes agent. When the agent spawns the Pod, it checks for that variable to add that annotation to the job manifest.

## Importance

The Kubernetes Cluster Autoscaler will evaluate the utilization of each Node in the cluster to determine if it can scale down the number of Nodes. When a node is removed from service, Pods running on that node are relocated and restarted on another Node. This results in:
- Prefect will encounter `No heartbeat detected from the remote task`; marking the run as "Failed".
- The initial Pod was relocated by the Cluster AutoScaler and the newly started Pod is actually reporting logs to Prefect Cloud.
- Because no heartbeat was detected from that original Pod, the Lazarus process will attempt to reschedule the Flow Run. but the Flow Run was already restarted by K8, it's just in a different Pod now.

## Checklist

This PR:

### DIFF
--- a/changes/pr4346.yaml
+++ b/changes/pr4346.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Add safe-to-evict annotation to k8s job - [#4643](https://github.com/PrefectHQ/prefect/pull/4346)"
+
+contributor:
+  - "[Xinyi2016](https://github.com/xinyi2016)"
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -74,7 +74,7 @@ class KubernetesAgent(Agent):
         - volumes (list, optional): A list of volumes to make available to be mounted when a
             job is run. The volumes in the list should be specified as nested dicts.
             i.e `[{"name": "my-vol", "csi": {"driver": "secrets-store.csi.k8s.io"}}]`
-        - safe_to_evict (bool, optional): Allow Cluster Autoscaler to remove the nodes where jobs
+        - safe_to_evict (str, optional): Allow Cluster Autoscaler to remove the nodes where jobs
             are running on.
     """
 
@@ -93,7 +93,7 @@ class KubernetesAgent(Agent):
         no_cloud_logs: bool = False,
         volume_mounts: List[dict] = None,
         volumes: List[dict] = None,
-        safe_to_evict: bool = None,
+        safe_to_evict: str = None,
     ) -> None:
         super().__init__(
             agent_config_id=agent_config_id,
@@ -753,7 +753,7 @@ class KubernetesAgent(Agent):
         cpu_limit: str = None,
         image_pull_policy: str = None,
         service_account_name: str = None,
-        safe_to_evict: bool = None,
+        safe_to_evict: str = None,
         labels: Iterable[str] = None,
         env_vars: dict = None,
         backend: str = None,
@@ -781,7 +781,7 @@ class KubernetesAgent(Agent):
                 Job defaults to `IfNotPresent`.
             - service_account_name (str, optional): Name of a service account to use for
                 Prefect init job. Job defaults to using `default` service account.
-            - safe_to_evict (bool, optional): Allow Cluster Autoscaler to remove the nodes
+            - safe_to_evict (str, optional): Allow Cluster Autoscaler to remove the nodes
                 where jobs are running on.
             - labels (List[str], optional): a list of labels, which are arbitrary string
                 identifiers used by Prefect Agents when polling for work
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -74,6 +74,8 @@ class KubernetesAgent(Agent):
         - volumes (list, optional): A list of volumes to make available to be mounted when a
             job is run. The volumes in the list should be specified as nested dicts.
             i.e `[{"name": "my-vol", "csi": {"driver": "secrets-store.csi.k8s.io"}}]`
+        - safe_to_evict (bool, optional): Allow Cluster Autoscaler to remove the nodes where jobs
+            are running on.
     """
 
     def __init__(
@@ -91,6 +93,7 @@ class KubernetesAgent(Agent):
         no_cloud_logs: bool = False,
         volume_mounts: List[dict] = None,
         volumes: List[dict] = None,
+        safe_to_evict: bool = None,
     ) -> None:
         super().__init__(
             agent_config_id=agent_config_id,
@@ -106,6 +109,7 @@ class KubernetesAgent(Agent):
         self.service_account_name = service_account_name or os.getenv(
             "SERVICE_ACCOUNT_NAME"
         )
+        self.safe_to_evict = safe_to_evict or os.getenv("SAFE_TO_EVICT")
         if image_pull_secrets is None:
             image_pull_secrets_env = os.getenv("IMAGE_PULL_SECRETS")
             image_pull_secrets = (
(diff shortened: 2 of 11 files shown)


madkinsz:
Thanks for taking the time to contribute @Xinyi2016 -- my immediate tendency is to say this should just be implemented on the `RunConfig` (which did not exist during the suggestion in the original thread) instead of adding more flags to the `Agent`, is there a reason this is better set on the agent?

madkinsz:
Hi again @Xinyi2016. I've talked to the team about this and I do not think we want to add this as a flag to the agent or run config. It's too specific to be useful as a user-facing setting. Generally, we cannot support every infrastructure option with flags and for that reason expose the ability to pass custom definitions entirely. We can consider setting this value by default in the default job template but that will take careful consideration since it could affect many of our users. In the meantime, our suggestion is to use a custom job definition that sets this value (you can likely just copy the generated one and modify it slightly).

As far as determining if this belongs in the default job template, what are the consequences for users if the safe-to-evict annotation is added and they are not using an auto-scaler? If a pod is evicted, can we be certain K8s is going to handle rerunning the pod? Do we want to resolve this in the backend instead? (cc @TylerWanner, @cicdw)

TylerWanner:
While I agree with @madkinsz here that prefect config should not so directly address kubernetes autoscaler config, I can understand why the friction between k8s autoscaler scheduling and prefect scheduling is significantly confusing.  As mentioned, I think the real fix for 3058 should be in the heartbeat so that the Lazarus process does not feel the need to reschedule work that is already re-running successfully.

On a larger scale, I can completely understand reaching for the ability to set agent config that adds certain annotations on flowrun pods.  At the moment, using a custom job template requires adding the appropriate flag to the agent's run command, which defaults to `prefect kubernetes agent start`.  Common practice is to define annotations on the flow job in the k8s RunConfig instead.  I believe job template config in the agent is overwritten by any existing RunConfig on the flow.

madkinsz:
We probably shouldn't add the safe-to-evict annotation to the default template. I'm going to close this with the intention of investigating the heartbeat issue. I suggest users add safe-to-evict annotation to the template themselves in the meantime.

Let's move any further discussion back to the issue at #3058

Xinyi2016:
Hi @madkinsz @TylerWanner , supper excited to have your comments. I use Prefect Kubernetes agent and Tasks very often and like it very much. I'm keen on getting the heartbeat issue fixed.

Feel free to let me know if there's anything else you would like me to work on.

Issue #4357, opened 2021-04-06
ECS Agent: Bring back enable_task_revisions 


potter420:
## Current behavior
In the old day, FargateAgent have `enable_task_revisions` to make each version of a flow a task definitions. Right now this behavior was not implemented in the ECS agent.

## Proposed behavior
ECSAgent should also have enable_task_revisions as the deprecated FargateAgent has.

## Example
This would be very useful for audit guys to not be bombarded with thousand of task definitions.

madkinsz:
Ah sorry @potter420 -- Github got weird and I didn't see you'd linked a PR. I'll review that instead.

madkinsz:
Hey @potter420 -- I've checked in with the team and I do not think this is something we can support. Now that we've made it so run configs can be modified per flow run, the previous technique used in the `FargateAgent` actually won't work since any run can invalidate the cache. To avoid registering so many task definitions, I'd recommend pre-registering task definitions as needed, and specify a `task_definition_arn` on the flow instead. In this configuration, the agent will never register a new task definition.

potter420:
Agree with you, I have just looked at the new ECS run since we are migrating from 0.13, we can now specify custom task definition per run, that will certainly cause various problems, like hash checking the task and everything. I think we will write a function that produce sensible default task definition for flow run instead and have prefect agent register for us first run then fixed that version for the next until we deploy a new version of the flow. Me unable to create anything since it a managed account.

Thank you. Will also amend my pull request to undo the changes related to this.

madkinsz:
Glad that it makes sense to you! Thanks for taking the time to work through it :)

Issue #4356, opened 2021-04-06
ECSAgent Custom Capacity Provider Strategy


potter420:
## Description
Hello , I have tried to run the task with customCapacityProviderStrategy on my run config, as per configuration guide from AWS, but I found out that ECS run automatically append launchType to the run configuration, which causing errors.

I'm also working on a fix on this too.

## Expected Behavior
I should have run my flow with customCapacityProviderStrategy.

## Reproduction
```
flow.run_config = ECSRun(
      run_task_kwargs={
          "capacityProviderStrategy": [
              {
                  'capacityProvider': 'AnalyticsETLJobCP',
                  'weight': 1,
                  'base': 0
              },
          ]
      }
  )

```

## Environment
Linux

Optionally run `prefect diagnostics` from the command line and paste the information here. -->

madkinsz:
Hi @potter420 

I think this is reasonable to fix. We can drop defaults if they'll break passing in other options.

I believe the lines in your PR at #4361 

```
        if "capacityProviderStrategy" in out:
            del out['networkConfiguration'], out['platformVersion'], out['launchType']
```

are all that's necessary, although I'm curious why you have to drop these additional sections.

There should probably be a user-experience bit as well such as a log call

`logger.debug("Found user-provided 'capacityProviderStrategy', dropping 'networkConfiguration', 'platformVerison', and 'launchType'")`

potter420:
Hi @madkinsz 

I think I panic and remove also the `networkConfiguration` since my use case mostly use bridge mode, which mostly never expose any ports to other services. But in that case, `networkConfiguration` should only be available if we specify `awsvpc` networking mode, other mode may cause errors.

PlatformVersion is also a Fargate exclusive configuration, so if we provide `capacityProviderStategy`, this also have to be remove to avoid possible errors.

Thank you.

madkinsz:
I don't see `platformVersion` set anywhere so I don't think we should be dropping it.

Since the default networkConfiguration is `awsvpc` it seems like there would only be a problem if they overrode the default. I think we should not drop that either.

If all we're doing dropping the `launchType` this might be best implemented as an `if` where we set that instead.

potter420:
I agree with you, we should do an `if` statement instead rather than terrible delete statements. But since there are many scenario that can go wrong with this. I take for example in my current configuration:
- I set a default config on my agent to run every job as Fargate Instance, with default network settings, launchType and `platformVersion` coming from run config on that Fargate agent. 
[...]

PR #4361, opened 2021-04-06
Fix to issue #4356 and #4357

potter420:
- Add condition check for Custom Providers
- Add enable task revisions

## Summary

 An attempt to fix for ECS Agent with custom providers strategy by removing launch type and various network configuration from the final run config that get sent to ECS `run_task`

Also tried to add the option `enable_task_revision` back to the Agent to behave likely the same as the previous agent. I understand that this may not be appropriate for all use case

## Changes
Most

## Importance

1.  #4356 : Fixing this is important for me since I had multiple capacity providers on my ECS cluster to satisfy different work loads. M5 for ETL, C5 for ML training. And we don't want to create different clusters with different default strategy providers.
2. #4357 : This is some enhancement, not very important.

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/ecs/agent.py
+++ b/src/prefect/agent/ecs/agent.py
@@ -1,6 +1,7 @@
 import os
 from copy import deepcopy
-from typing import Iterable, Dict, Any
+from typing import Iterable, Dict, Any, Tuple
+from prefect.utilities.aws import get_boto_client
 
 import slugify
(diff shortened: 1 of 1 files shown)


madkinsz:
Discussed in the linked issues. https://github.com/PrefectHQ/prefect/issues/4356 will track a future fix for provider strategies.

PR #4368, opened 2021-04-07
Fix incorrect unused task tracking

madkinsz:
## Summary

When a task subclass is declared in the `with Flow..` context we instantiate a `Task()` object to check the reserved attributes. Unfortunately, this adds a `Task` to the `unused_task_tracker` set that will never be removed which will give users false warnings that they have task objects in their flow that are not called.

## Changes

- Disable the unused task tracker during reserved attribute retrieval

## Importance

Fixes incorrect warning

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -173,7 +173,11 @@ class TaskMetaclass(type):
         computed to avoid circular import issues.
         """
         if not hasattr(Task, "_cached_reserved_attributes"):
-            Task._cached_reserved_attributes = tuple(sorted(Task().__dict__))  # type: ignore
+            # Create a base task instance to determine which attributes are reserved
+            # we need to disable the unused_task_tracker for this duration or it will
+            # track this task
+            with prefect.context(_unused_task_tracker=set()):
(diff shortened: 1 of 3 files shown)


src/prefect/core/task.py
```diff
@@ -173,7 +173,11 @@ def _reserved_attributes(self) -> Tuple[str]:
         computed to avoid circular import issues.
         """
         if not hasattr(Task, "_cached_reserved_attributes"):
-            Task._cached_reserved_attributes = tuple(sorted(Task().__dict__))  # type: ignore
+            # Create a base task instance to determine which attributes are reserved
+            # we need to disable the unused_task_tracker for this duration or it will
+            # track this task
+            with prefect.context(_unused_task_tracker=set()):
+                Task._cached_reserved_attributes = tuple(sorted(Task().__dict__))  # type: ignore
```
madkinsz, line 180: We can also store the `Task()` object and `.remove()` it afterwards but this feels cleaner

zangell44 (approved):
looks good to me!

PR #4373, opened 2021-04-08
Improve messaging when a flow is skipped during registration

madkinsz:
## Summary

Explains to users who may not understand why their flow is being skipped during `prefect register` that skipping is okay with a couple small display changes.

e.g. https://prefect-community.slack.com/archives/CL09KU1K7/p1617827979177100

## Changes

- Adds metadata note to `--help` message
- Adds note after each `Skipped` message

## Importance

- Improves experience for non-power users / people unfamiliar with idempotency keys

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/build_register.py
+++ b/src/prefect/cli/build_register.py
@@ -462,7 +462,7 @@ def build_and_register(
                     click.echo(f"  └── Version: {flow_version}")
                     stats["registered"] += 1
                 else:
-                    click.secho(" Skipped", fg="yellow")
+                    click.secho(" Skipped (metadata unchanged)", fg="yellow")
                     stats["skipped"] += 1
     return stats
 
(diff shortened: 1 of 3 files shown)


cicdw (approved):

src/prefect/cli/build_register.py
```diff
@@ -705,7 +705,11 @@ def watch_for_changes(
 @click.pass_context
 @handle_terminal_error
 def register(ctx, project, paths, modules, json_paths, names, labels, force, watch):
-    """Register one or more flows into a project."""
+    """Register one or more flows into a project.
+
+    Flows with unchanged metadata will be skipped as registering again will only
+    change the version number.
```
zangell44, line 711: maybe worth noting that `force` will override this behavior and guarantee new registration

zangell44 (approved):
minor comment, looks good to me

src/prefect/cli/build_register.py
zangell44: (it is documented later in the help string, so we can live without it)

madkinsz: Yeah I had a similar thought. I want to keep the top-level description as simple as possible.

PR #4374, opened 2021-04-08
Display a more helpful error when flow context cannot be inferred

madkinsz:
## Summary

Flow context cannot be inferred if a task is called outside of the `with Flow()` block. This PR aims to help users quickly determine which task in a large script is causing this error.

e.g.

```
>>> from prefect import Flow, task
>>> @task
... def foo():
...    pass
... 
>>> foo()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect-core/lib/python3.8/site-packages/prefect/core/task.py", line 621, in __call__
    new.bind(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect-core/lib/python3.8/site-packages/prefect/core/task.py", line 674, in bind
    raise ValueError(
ValueError: Could not infer an active Flow context while creating edge to <Task: foo>. This often means you called a task outside a `with Flow(...)` block. Did you mean to call `this_task.run(...)`?
```

## Changes

- Identifies the task that is called when flow context cannot be inferred

## Importance

- Improves ability to quickly determine the task where this error occurs 
Prompted by https://prefect-community.slack.com/archives/CL09KU1K7/p1617891432223300

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -671,10 +671,18 @@ class Task(metaclass=TaskMetaclass):
 
         flow = flow or prefect.context.get("flow", None)
         if not flow:
+            # Determine the task name to display which is either the function task name
+            # or the initialized class where we can't know the name of the variable
+            task_name = (
+                self.name
+                if isinstance(self, prefect.tasks.core.function.FunctionTask)
+                else f"{type(self).__name__}(...)"
+            )
             raise ValueError(
                 f"Could not infer an active Flow context while creating edge to {self}."
                 " This often means you called a task outside a `with Flow(...)` block. "
-                "Did you mean to call `this_task.run(...)`?"
+                "If you're trying to run this task outside of a Flow context, you "
+                f"need to call `{task_name}.run(...)`"
             )
 
         self.set_dependencies(
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -671,7 +671,9 @@ class Task(metaclass=TaskMetaclass):
 
         flow = flow or prefect.context.get("flow", None)
         if not flow:
-            raise ValueError("Could not infer an active Flow context.")
+            raise ValueError(
+                f"Could not infer an active Flow context while creating edge to {self}."
+            )
 
         self.set_dependencies(
             flow=flow,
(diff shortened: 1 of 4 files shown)


src/prefect/core/task.py
```diff
@@ -671,7 +671,11 @@ def bind(
 
         flow = flow or prefect.context.get("flow", None)
         if not flow:
-            raise ValueError("Could not infer an active Flow context.")
+            raise ValueError(
+                f"Could not infer an active Flow context while creating edge to {self}."
+                " This often means you called a task outside a `with Flow(...)` block. "
+                "Did you mean to call `this_task.run(...)`?"
```
madkinsz, line 677: We could say `{self.__name__}.run(...)` but if their task is a class e.g. `FooBar` it will be misleading because it will not indicate it's instantiated / we can't know the name of their variable. Perhaps we should optimize for people using `@task` decorators instead since that would be a more helpful message.

madkinsz, line 677: e.g.
```suggestion
                f"Did you mean to call `{self.__name__}.run(...)`?"
```

whimsicallyson: if I saw this, I would assume that all I needed to do was add/replace `{self.__name__}.run(...)` into my code. Based on your comment above, it sounds like that might not always be the case? If it isn't, I would rather have the placeholder option in the error message. If I see code suggestions specific to what I'm writing, I assume they know what they're talking about 😝

madkinsz: It would be the case for most user written tasks but not for any of the task library task 🤦 like...

```python

@task
def foo():
   pass

@task 
def bar():
   foo()

with Flow("test") as flow:
    bar()

flow.run()
```

would say "Did you mean to call `foo.run(...)`?" -- which is correct


```python
from prefect import Flow, task, Task

class Foo(Task):
    def run():
         pass

@task 
def bar():
   foo = Foo()  # initialize
   foo()            # "run"


with Flow("test") as flow:
    bar()

flow.run()
```

would say "Did you mean to call `Foo.run(...)`?" which is wrong.

I could check if it's a function task or not and say

`foo.run(...)` in the first case and `Foo(...).run(...)` in the second case

whimsicallyson: I would be all for that (assuming its not incredibly tedious)!

madkinsz: Not tedious, I wonder if this is better or if it's too much

whimsicallyson: personally I like it

zangell44 (approved):

PR #4380, opened 2021-04-09
add task_definition_add_uuid as an agent option for ecs

joe1981al:
## Summary
Adds ability to append UUID to task definition for ECS Agent

## Changes
updates to ECS agent and agent CLI command

## Importance
AWS has a rate limit on RegisterTaskDefinition endpoint that is extremely low for the same task definition name (aka family). The error occurred when 10 runs of the same flow were scheduled and 3 agents starting trying to run the flows, 6 of the 10 failed with the message below.  

> An error occurred (ClientException) when calling the RegisterTaskDefinition operation: Too many concurrent attempts to create a new revision of the specified family.

## Checklist
This PR:

### DIFF
--- a/changes/pr4380.yaml
+++ b/changes/pr4380.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Adds ability to append UUID to task definition name (aka family) for the ECS agent preventing per family rate limiting on RegisterTaskDefinition api. [#4380](https://github.com/PrefectHQ/prefect/pull/4380)"
+
+contributor:
+  - "[Joe McDonald](https://github.com/joe1981al)"
--- a/docs/orchestration/agents/ecs.md
+++ b/docs/orchestration/agents/ecs.md
@@ -188,6 +188,17 @@ prefect agent ecs start --execution-role-arn my-execution-role-arn
 Flows can override this agent default by passing the `execution_role_arn` option to
 their respective [ECSRun](/orchestration/flow_config/run_configs.md#ecsrun) `run_config`.
 
+### Task Definition Add UUID
+
+The ECS agent when generating a task definition name (aka family) will default to using
+the pattern of `prefect-<flow_name>` which can lead to rate limiting if submit several
+flow runs in a short period.  This rate limit is per family, to bypass this set the 
+`--task-definition-add-uuid` option to true.
+
+```bash
+prefect agent ecs start --task-definition-add-uuid true
+```
+
 ### Custom Task Definition Template
 
 For deeper customization of the Tasks created by the agent, you may want to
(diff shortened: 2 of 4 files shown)


joe1981al:
From testing, this change prevented the error in description, but then hit general throttling limits.  I set two environment variables for rate limiting which now allowed me to submit 20 flow runs of the same flow within seconds.

Variables Set
```
AWS_RETRY_MODE=adaptive
AWS_MAX_ATTEMPTS=10
```

Links to official AWS documentation
[Boto3 Retries](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html)
[max_attempts SDK Docs](https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-max_attempts.html)
[retry_mode SDK Docs](https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-retry_mode.html)

dylanbhughes:
Hey @joe1981al! Thanks so much for taking the time to make Prefect even better 😄. We'll take a look at this PR as soon as possible 🚀

madkinsz:
Hey @joe1981al -- I think this is something we'd want to improve as a default experience i.e. the user should not have to set a flag or aws environment variables to get this working. I think that appending a short random id to the name by default may be reasonable (cc @cicdw @TylerWanner for thoughts) and we should be able to setup a retry via the boto3 client e.g.

```python
import boto3
from botocore.config import Config

config = Config(
   retries = {
      'max_attempts': 10,
      'mode': 'adaptive'
   }
)

ec2 = boto3.client('ec2', config=config)
```

I think that change may require a look from someone on our team so we can ensure that the config is handled properly across our uses of the boto client.

joe1981al:
@madkinsz the reason I did the flag was to not change the current behavior unexpectedly.  I wouldn't complain if it was the default...

madkinsz:
> @madkinsz the reason I did the flag was to not change the current behavior unexpectedly. I wouldn't complain if it was the default...

Yep that makes sense! I'm just saying this is something that we'll want to make _just work_ for everyone. I'll check in with the rest of the team and see if there's an ideal solution here. Do you mind if we take over this fix or do you want to do the back and forth to get this contribution in under your name?

joe1981al:
> > @madkinsz the reason I did the flag was to not change the current behavior unexpectedly. I wouldn't complain if it was the default...
> 
> Yep that makes sense! I'm just saying this is something that we'll want to make _just work_ for everyone. I'll check in with the rest of the team and see if there's an ideal solution here. Do you mind if we take over this fix or do you want to do the back and forth to get this contribution in under your name?

I'm good with however you want to handle it.  If you took over would you have an Issue# created?  If so, that would be good enough for me to track it's progress.

madkinsz:
We're going to try to look into this further this week. You can follow the issue at https://github.com/PrefectHQ/prefect/issues/4402 -- thanks for bringing this to us!

PR #4383, opened 2021-04-09
Run config logging level

zangell44:
## Summary
This PR adds support for environment variables in UniversalRunConfigs.

## Changes
- Updates `RunConfig` base class to include environment variables
- Updates run config serialization schema to include environment variables on universal run
- Updates Kubernetes Agent logging level logic to respect `PREFECT__LOGGING__LEVEL` if passed
- Adds test for universal run config env

## Importance
This rounds out changes necessary to make logging level easily configurable via the UI for individual flow runs.

## Checklist

This PR:


src/prefect/agent/kubernetes/agent.py
```diff
@@ -516,7 +516,9 @@ def generate_job_spec_from_environment(
         env[4]["value"] = self.namespace
         env[5]["value"] = str(self.labels)
         env[6]["value"] = str(self.log_to_cloud).lower()
-        env[7]["value"] = config.logging.level
+        env[7]["value"] = self.env_vars.get(
```
cicdw, line 519: Because we additionally populate all env vars in lines 524+ below, I'd vote to change this to 
```suggestion
        env[7]["value"] = self.env_vars.pop(
```

cicdw: I also think that we need to make similarly intentional changes to the other agents, e.g., here are some lines in the local agent that, if swapped, would break this feature: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/agent/local/agent.py#L193-L214
and similarly for the docker and ecs agents.

cicdw (changes requested):
A few requests

src/prefect/agent/kubernetes/agent.py
zangell44: Hmmm yep I can get behind the idea of more explicit logic here, I'll make some updates

zangell44: @cicdw re:
> make similarly intentional changes to the other agents, e.g., here are some lines in the local agent that, if swapped, would break this feature

So not only would changing the order of lines break this feature, it would break explicitly established precedence patterns, which are well described in comments. This precedence is well tested for Docker and Local Agents.

Would adding tests for ECS and K8s agents + adding comments to `env.update(run_config.env)` calls be sufficient here?

It feels clunky to write something like
```python
env.update(run_config.env.pop('PREFECT__LOGGING__LEVEL'), env['PREFECT__LOGGING__LEVEL'])
env.update(run_config.env)
```

madkinsz:
So the `DockerAgent` has 

```
        # 1. Logging config (optional)
        # We only set this on non-DockerRun runs for backwards compatibility.
        # In the future we don't want the agent to set logging config on a flow
        # run unless explicitly asked.  We do this early on so later config
        # sources can override
        if run_config is None:
            env.update({"PREFECT__LOGGING__LEVEL": config.logging.level})
```

and it seems like with this PR we are moving back to a behavior we were trying to move away from? (or at least that Jim was trying to move away from) -- am I misunderstanding what our intention is here?

src/prefect/agent/kubernetes/agent.py
cicdw: Because logging level is a uniquely important / special environment variable that we plan to elevate as a dedicated option in the UI, we need to treat it as such in code as well so that future changes to how we handle the `.env` attribute on run configs doesn't result in hidden breakages of this feature.  So in this case, every agent should explicitly set this variable with some precedence rule as the K8s agent does (note that environment variable at position `7` is set explicitly as logging level in the yaml template).  

What that looks like will most likely be agent-dependent.

zangell44:
> So the `DockerAgent` has
> 
> ```python
>         # 1. Logging config (optional)
>         # We only set this on non-DockerRun runs for backwards compatibility.
>         # In the future we don't want the agent to set logging config on a flow
>         # run unless explicitly asked.  We do this early on so later config
>         # sources can override
>         if run_config is None:
>             env.update({"PREFECT__LOGGING__LEVEL": config.logging.level})
> ```
> 
> and it seems like with this PR we are moving back to a behavior we were trying to move away from? (or at least that Jim was trying to move away from) -- am I misunderstanding what our intention is here?

"In the future we don't want the agent to set logging config on a flow run unless explicitly asked." matches our intention here. If `PREFECT__LOGGING__LEVEL` is not explicitly passed, these changes will have no effect

zangell44:
The alternative here is adding an explicit property `logging_level` on the RunConfig object and ignoring the environment variable, which I suppose I'm open to.

madkinsz:
> The alternative here is adding an explicit property `logging_level` on the RunConfig object and ignoring the environment variable, which I suppose I'm open to.

That may make sense if we're just trying to introduce a top-level option to adjust logging levels. Using an environment variable to set a config value which we then merge with a run config seems a bit convoluted.

cicdw:
At the end of the day, the end result will be an environment variable set on job -- if we broke it out into a separate run config attribute, we'd still be doing the same manipulations ("What happens if they set conflicting levels at the top level key and as an environment variable?") so having a single interface via env vars makes the most sense to me, we just need to make sure that we treat this env var in a special way in the code because it _is_ a supported feature.

zangell44:
All logic related to `PREFECT__LOGGING__LEVEL` has been made very verbose across agents. Updated a few relevant tests too. Let me know what you think.

src/prefect/agent/docker/agent.py
```diff
@@ -518,20 +518,27 @@ def populate_env_vars(
...
-
-        # 2. Values set on the agent via `--env`
+        # 1. Values set on the agent via `--env`
         env.update(self.env_vars)
 
+        # 2. Logging config - if logging level is passed as an environment
+        # variable on the run config, it should take precedence
+        existing_logging_level = env.get(
+            "PREFECT__LOGGING__LEVEL", config.logging.level
+        )
```
cicdw, line 530: I'm curious if there was a reason for looking in the env?  If you set this env var, it should populate `config.logging.level` so it seems to me that setting `existing_logging_level = config.logging.level` is sufficient here

cicdw: Also very minor preference for calling this `default_logging_level`

src/prefect/agent/ecs/agent.py
```diff
@@ -486,6 +486,14 @@ def get_run_task_kwargs(
         # - Values set using the `--env` CLI flag on the agent
         env = self.env_vars.copy()
         if run_config.env:
+            env.update(
+                {
+                    "PREFECT__LOGGING__LEVEL": run_config.env.pop(
+                        "PREFECT__LOGGING__LEVEL",
+                        env.get("PREFECT__LOGGING__LEVEL", config.logging.level),
```
cicdw, line 493: Same comment as above -- setting the env var populates the config, so defaulting to `config.logging.level` should be sufficient

src/prefect/agent/local/agent.py
```diff
@@ -209,8 +209,17 @@ def populate_env_vars(
...
 
-        # 5. Values set on a LocalRun RunConfig (if present
+        # 5. Values set on a LocalRun RunConfig (if present)
         if run_config is not None and run_config.env is not None:
+            # logging level passed via runconfig environment variable
+            # should always take precedence
+            env.update(
+                {
+                    "PREFECT__LOGGING__LEVEL": run_config.env.pop(
+                        "PREFECT__LOGGING__LEVEL", env["PREFECT__LOGGING__LEVEL"]
```
cicdw, line 219: does `env` always have this key?

cicdw (dismissed):
Love the spirit of the changes and this works as-is - left some suggestions for simplification

src/prefect/agent/docker/agent.py
zangell44: There seem to be three sources of logging level configs already implemented in the logic here, which I do not want to break.

1. `config.logging.level`, which aren't stored in `env` yet
2. The agent environment variables which now exist in `env` see `env.update(self.env_vars)` https://github.com/PrefectHQ/prefect/pull/4383/files/caacdb3e8693d1d1bf4e95663220e99d240a1ff2#diff-091912bb80855c71834fd22611ffb4d40919a954017024b5aa8588dac57b6ab4R524
3. `PREFECT__LOGGING__LEVEL` passed via the run config

Am I missing something here?

src/prefect/agent/ecs/agent.py
zangell44: Same comment as above https://github.com/PrefectHQ/prefect/pull/4383/files/caacdb3e8693d1d1bf4e95663220e99d240a1ff2#r613232254

src/prefect/agent/local/agent.py
zangell44: for the local agent it's set here https://github.com/PrefectHQ/prefect/pull/4383/files/caacdb3e8693d1d1bf4e95663220e99d240a1ff2#diff-83a67d1b4d4351a86132201d4d9794810d2b1a49aa051794d682e39daf9c67bdR193

Taking another look at it, it makes sense to make the logic a bit more uniform. After we clarify the comments above I'll do a little refactor here

src/prefect/agent/docker/agent.py
```diff
@@ -518,20 +518,27 @@ def populate_env_vars(
         else:
             api = config.cloud.api
 
-        env = {}
+        env = {}  # type: ignore
```
madkinsz, line 521: Why do we need to ignore the type?

src/prefect/agent/docker/agent.py
```diff
@@ -518,20 +518,27 @@ def populate_env_vars(
...
+        if run_config is not None and run_config.env is not None:
+            env.update(
+                {
+                    "PREFECT__LOGGING__LEVEL": run_config.env.pop(
+                        "PREFECT__LOGGING__LEVEL", existing_logging_level
+                    )
+                }
+            )
+        else:
+            env.update({"PREFECT__LOGGING__LEVEL": existing_logging_level})
```
madkinsz, line 540: Won't the top half of this occur on L544 with `env.update(run_config.env)` ? It seems like you could drop this entire if/else and just do `env.update({"PREFECT__LOGGING__LEVEL": existing_logging_level})` which will be overridden in the next block if it exists

src/prefect/agent/docker/agent.py
madkinsz: I agree `existing` is a confusing name

madkinsz: I'm a bit confused about the precedence here.

From low to high this looks like:
1. `--env` arguments passed to agent: may or may not be set
2. `config.logging.level`: will always have a value so the passed `--env PREFECT__LOGGING__LEVEL` is now ignored
3. `run_config.env`: may or may not be set, overrides all - this makes sense

Seems like 1/2 should be swapped

src/prefect/agent/ecs/agent.py
madkinsz: Now I'm wondering why the implementation of this is different per agent

src/prefect/agent/kubernetes/agent.py
```diff
@@ -516,7 +516,9 @@ def generate_job_spec_from_environment(
         env[4]["value"] = self.namespace
         env[5]["value"] = str(self.labels)
         env[6]["value"] = str(self.log_to_cloud).lower()
-        env[7]["value"] = config.logging.level
+        env[7]["value"] = self.env_vars.pop(
+            "PREFECT__LOGGING__LEVEL", config.logging.level
+        )
```
madkinsz, line 521: Probably should not be mutating `self.env_vars`

madkinsz (changes requested):

src/prefect/agent/docker/agent.py
zangell44: 1/2 are swapped in the get call, trying to get the env argument first and falling back to `config.logging.level`
```
existing_logging_level = env.get(
            "PREFECT__LOGGING__LEVEL", config.logging.level
        )
```

Seems like I just made things more confusing when I tried to make them more explicit, I'll take another pass at this

madkinsz: I see. I would load things in order of low -> high and just let them override.

src/prefect/agent/docker/agent.py
zangell44: lol

zangell44: that is roughly how I had it, Chris suggested making this logic more explicit because we're treating `PREFECT__LOGGING__LEVEL` in a first class way unlike other environment variables

src/prefect/agent/kubernetes/agent.py
zangell44: good catch

src/prefect/agent/docker/agent.py
zangell44: excellent question, I'll try to track down the underlying issue, but I believe this failed code check without the ignore

src/prefect/agent/docker/agent.py
madkinsz: I find this block far more complicated and we're not actually treating it differently here. The run config env overrides all other values regardless of whether it's logging related or not.

zangell44: @cicdw thoughts here? It also seems to me like these efforts are making things much more complicated

zangell44:
@cicdw @madkinsz given changes to make things more verbose were not well received, I dropped the last four commits related to these efforts.

My proposal is
1. Leave the logic as is, where environment variables override one another based on precedence
2. If the comments around the logic aren't crystal clear, update them
3. Add additional tests around `PREFECT__LOGGING__LEVEL` for all agents that both make the intention of the logic clear and ensure we don't break things in the future

madkinsz:
Sounds good to me. Some regression tests around that variable specifically should be sufficient.

src/prefect/agent/docker/agent.py
```diff
@@ -526,8 +526,7 @@ def populate_env_vars(
         # In the future we don't want the agent to set logging config on a flow
         # run unless explicitly asked.  We do this early on so later config
         # sources can override
-        if run_config is None:
-            env.update({"PREFECT__LOGGING__LEVEL": config.logging.level})
+        env.update({"PREFECT__LOGGING__LEVEL": config.logging.level})
```
madkinsz, line 529: So my only point of concern is this still
>  # In the future we don't want the agent to set logging config on a flow run unless explicitly asked.  We do this early on so later config sources can override

we are in direct violation of this idea here

`env.update({"PREFECT__LOGGING__LEVEL": config.logging.level})` 

where the logging level from the config.toml on the agent's machine is propagated to the flow run.

zangell44: What's the motivation behind not wanting the agent to propagate its logging level by default?

madkinsz: I don't know, this was a decision I presume was made by Jim and we're only maintaining that behavior when run configs are `None` for backwards compatibility. I could see the reasoning that the config.toml may not be expected to affect flow runs for that agent.

zangell44: Hmmm. From my perspective, it seems like the config.toml would provide a reasonable default + doesn't break any existing setups.

I can't imagine myself getting particularly upset about it, besides maybe having to redeploy an agent.

I have the least context here, so I'm open to persuasion.

If we're going to remove this logic, though, we should remove it from all agents and stop supporting it.

madkinsz: I also don't quite understand the reasoning. I'm fine with always pulling from the config for the base value, it makes sense to me.

madkinsz (approved):

src/prefect/agent/docker/agent.py
```diff
@@ -526,8 +526,7 @@ def populate_env_vars(
         # In the future we don't want the agent to set logging config on a flow
         # run unless explicitly asked.  We do this early on so later config
         # sources can override
```
madkinsz, line 528: This comment should be updated

madkinsz:
Seems like there should be somewhere in the docs to update for this too?

zangell44:
good point, I'll poke around in the docs a bit

madkinsz (approved):

madkinsz:
Did you mean to remove `docs/.vuepress/public/orchestration/ui/flow-run.png`?

zangell44:
I replaced it with an updated image showing logging level + adjustments to the Flow Run page

docs/orchestration/flow_config/run_configs.md
```diff
@@ -82,10 +85,10 @@ from prefect.run_configs import UniversalRun
 flow.run_config = UniversalRun()
 ```
 
-Configure labels for this flow:
+Configure environment variables and labels for this flow:
 
 ```python
-flow.run_config = UniversalRun(labels=["label-1", "label-2"])
+flow.run_config = UniversalRun(env={"SOME_VAR": "value"}, ["label-1", "label-2"])
```
cicdw, line 91: really awesome that you revisited all the docs for this :clap:

cicdw (approved):

PR #4384, opened 2021-04-09
Add error handling to timeouts that fail during result pickling or passing

madkinsz:
## Summary

A couple error cases in the `run_with_multiprocess_timeout` typically used by the `TaskRunner` when running a threaded dask scheduler were not captured and logged which could lead to opaque failures of the timeout handler. This PR adds logging and error capturing to try to improve introspection during failure cases.

## Changes

- Pickling errors are captured, logged, and returned so the user gets a clear description of result requirements while using this timeout handler. Previously, this would result in a `TimeoutError` because the queue would be empty.
- `Queue.put` errors are logged and re-raised. They cannot be returned to the main process because the queue is how we pass exceptions up. Logging with a traceback is our best way to improve this edge-case right now.

## Importance

Several users have encountered unknowable errors while using the timeout handler lately

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -174,11 +174,42 @@ def multiprocessing_safe_run_and_retrieve(
         with prefect.context(context):
             logger.debug(f"{name}: Executing...")
             return_val = fn(*args, **kwargs)
+            logger.debug(f"{name}: Execution successful.")
     except Exception as exc:
         return_val = exc
+        logger.debug(
+            f"{name}: Encountered a {type(exc).__name__}, "
+            f"returning details as a result..."
+        )
+
+    try:
+        pickled_val = cloudpickle.dumps(return_val)
+    except Exception as exc:
+        base_msg = (
+            f"Failed to pickle result of type {type(return_val).__name__!r} with "
+            f'exception: "{type(exc).__name__}: {str(exc)}".'
+        )
+        logger.error(f"{name}: {base_msg}")
+        pickled_val = cloudpickle.dumps(
+            RuntimeError(
+                f"{base_msg} This timeout handler requires your function return "
+                f"value to be serializable with `cloudpickle`. "
+                "If you must return a unserializable value, consider switching your "
+                "executor to use processes instead of threads to use a different "
+                "timeout handler."
+            )
+        )
 
     logger.debug(f"{name}: Passing result back to main process...")
-    queue.put(cloudpickle.dumps(return_val))
+
+    try:
+        queue.put(pickled_val)
+    except Exception:
+        logger.error(
+            f"{name}: Failed to put result in queue to main process!",
+            exc_info=True,
+        )
+        raise
 
 
 def run_with_multiprocess_timeout(
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -185,20 +185,13 @@ def multiprocessing_safe_run_and_retrieve(
     try:
         pickled_val = cloudpickle.dumps(return_val)
     except Exception as exc:
-        base_msg = (
+        err_msg = (
             f"Failed to pickle result of type {type(return_val).__name__!r} with "
-            f'exception: "{type(exc).__name__}: {str(exc)}".'
+            f'exception: "{type(exc).__name__}: {str(exc)}". This timeout handler "'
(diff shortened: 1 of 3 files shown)


src/prefect/utilities/executors.py
```diff
@@ -174,11 +174,42 @@ def multiprocessing_safe_run_and_retrieve(
         with prefect.context(context):
             logger.debug(f"{name}: Executing...")
             return_val = fn(*args, **kwargs)
+            logger.debug(f"{name}: Execution successful.")
     except Exception as exc:
         return_val = exc
+        logger.debug(
```
cicdw, line 180:
```suggestion
        logger.error(
```

cicdw: And also `exc_info =True`

src/prefect/utilities/executors.py
```diff
@@ -174,11 +174,42 @@ def multiprocessing_safe_run_and_retrieve(
...
+        )
+
+    try:
+        pickled_val = cloudpickle.dumps(return_val)
+    except Exception as exc:
+        base_msg = (
+            f"Failed to pickle result of type {type(return_val).__name__!r} with "
+            f'exception: "{type(exc).__name__}: {str(exc)}".'
+        )
+        logger.error(f"{name}: {base_msg}")
```
cicdw, line 192:
```suggestion
        logger.error(f"{name}: {base_msg}", exc_info=True)
```

cicdw (changes requested):
Some minor requests about logging details

src/prefect/utilities/executors.py
madkinsz: I'm not sure it makes sense to log the traceback here. It's just going to point them into this bit of code and a `Pickle.dumps` call. The returned error will include a decent traceback that should be sufficient and I worry printing several tracebacks will just be confusing.

madkinsz: (note I've already put the exception type and message into the `base_msg` so it's mostly redundant other than the traceback)

src/prefect/utilities/executors.py
```diff
@@ -174,11 +174,42 @@ def multiprocessing_safe_run_and_retrieve(
         with prefect.context(context):
             logger.debug(f"{name}: Executing...")
             return_val = fn(*args, **kwargs)
+            logger.debug(f"{name}: Execution successful.")
     except Exception as exc:
         return_val = exc
+        logger.error(
+            f"{name}: Encountered a {type(exc).__name__}, "
+            f"returning details as a result..."
+        )
```
madkinsz, line 183: Can't get your comment to come up here so replying in a new one -- adding `exc_info` here is unnecessary because this exception will be pickled and returned to be re-raised in the parent function.

cicdw: Hm I get your reasoning but that seems like a brittle way to lose this info if that ever changes.  I guess any logs here are better than the previous behavior so I won't push the point

madkinsz: 👍 I think the next time someone encounters issues here will tell us if this is sufficient or if we need to do more.  When I first started working on this, there were no logs at all and we've been trending gradually verbose.

If the behavior changes above this (such that the exception isn't raised) we'd have big problems because then any errors in the `task.run()` function would also be silenced, so I'm not concerned about brittleness there.

src/prefect/utilities/executors.py
```diff
@@ -174,11 +174,42 @@ def multiprocessing_safe_run_and_retrieve(
...
+        logger.error(f"{name}: {base_msg}")
+        pickled_val = cloudpickle.dumps(
+            RuntimeError(
+                f"{base_msg} This timeout handler requires your function return "
+                f"value to be serializable with `cloudpickle`. "
+                "If you must return a unserializable value, consider switching your "
+                "executor to use processes instead of threads to use a different "
+                "timeout handler."
+            )
```
madkinsz, line 200: Now that I think about it more -- won't dask fail with these kind of results anyway? I guess you could get away with using the `LocalExecutor` ? hmm.

madkinsz: I've removed this extra message because I do not think it will help.

cicdw (approved):

zangell44 (approved):

PR #4397, opened 2021-04-13
Fix `ShellTask` on win32

### LINKED ISSUE #1984: WindowsShellTask or Windows support for ShellTask
Our current `ShellTask` implementation is not Windows compatible.  We should _either_:

- make it windows compatible
- create a new shell task specifically for windows machines

madkinsz:
## Summary

The `ShellTask` does not work on win32 because it does not look for executables in the path by default. This sets `Popen(..., shell=True)` when using win32 which inspects the path for the given executable.

You still need `bash` installed and on your `PATH` for this to work by default. Other shells should work without issue as well, but `cmd.exe` does not seem to support running a script.

## Importance

Closes #1984 https://github.com/PrefectHQ/prefect/issues/4393

## Checklist

This PR:

### DIFF
--- a/changes/pr4397.yaml
+++ b/changes/pr4397.yaml
@@ -1,0 +1,3 @@
+
+enhancement:
+  - "Allow the `ShellTask` to be used on win32 - [#4397](https://github.com/PrefectHQ/prefect/pull/4397)"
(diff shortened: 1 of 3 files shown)


cicdw (approved):
nice detective work

zangell44 (approved):

PR #4409, opened 2021-04-15
Pass reference states to terminal handler

zangell44:
## Summary
This PR allows the flow runner to pass a set of reference task states instead of all task states.

## Importance
When running against Cloud or Server, `task_states` is not populated. Until we implement some sort of deferred task result retrieval in place, passing reference task states seems reasonable.

## Checklist

This PR:


madkinsz:
Did you make sure this works nicely when run against the API?

madkinsz (approved):

zangell44:
> Did you make sure this works nicely when run against the API?

Yep works as expected. `FlowRunner` is already using `key_states` to determine final state, so I wouldn't expect issues here.

PR #4414, opened 2021-04-18
Add column names to PostgresFetch

gmontanola:
## Summary

There are plenty of ways of doing this: 
* We could add other cursors like the MySQL task does and get column names through the results (DictCursor and NamedTupleCursor store field name values).
* Doing an extra query in order to fetch those values

I tried to keep it simple and use the `cursor` object that we've already have. I**f this change is out of scope** for the task, I think we should add different cursors support so the user can handle this without the need of extra querying,

## Changes

It adds a boolean flag to the `PostgresFetch` task.
If it's set to `True` the task fetches column names through the already created cursor and merge it with the result set.

Defaults to `False` to avoid breaking anything already built with the current behavior.

## Importance

It's a pretty common need when fetching data. Without this we need to make an extra query to fetch column names from `information_schema.columns` and merge it with your actual data.

## Checklist

This PR:

### DIFF
--- a/changes/pr4414.yaml
+++ b/changes/pr4414.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Adds column name fetching to PostgresFetch task - [#4414](https://github.com/PrefectHQ/prefect/pull/4414)"
+
+contributor:
+  - "[Gabriel Montañola](https://github.com/gmontanola)"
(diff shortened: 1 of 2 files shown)


madkinsz:
Thanks for contributing @gmontanola ! I think in the future we'll want to support custom cursors as well, but I think this is a reasonable implementation for now.

madkinsz (approved):

PR #4416, opened 2021-04-19
clear _stopped threading.Event for reuse

zhqu1148980644:
The original code does not cause the error if the logger manager only start once

## Summary
clear the `threading.Event` instance `_stopped`  for possible reuse.

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -82,6 +82,7 @@ class LogManager:
             self._write_logs()
             self.thread = None
             self.client = None
+            self._stopped.clear()
 
     def _write_logs_loop(self) -> None:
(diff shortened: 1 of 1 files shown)


zangell44 (changes requested):
Hi @zhqu1148980644 , thanks for contributing to Prefect!

When you say
> The original code does not cause the error if the logger manager only start once

Can you share a minimum reproducible example of the issue you're seeing?

---
Also, feel free to record your change + add yourself as a contributor via a Changelog entry. More info [here](https://docs.prefect.io/core/development/contributing.html#changelog)

zhqu1148980644:
There is no issue, I think the Event should be cleared by its design.

PR #4417, opened 2021-04-19
Check for AWS_RETRY_MODE variable before setting default in `ECSAgent`

madkinsz:
## Summary

The `botocore.Config` object does not load pre-existing settings so if a user tries to override this setting via environment variable, we will override them (which is not the intended behavior). Passing the variables via CLI is tricky because botocore expects a bunch of different types (ie we can't just take a (str, str) tuple of key value settings for to pass to `botocore_config`). This is the only setting we set a default for so this is an acceptable workaround for now. Users with more complex use-cases can start their agent via a custom python script rather than using the CLI. Note that this will likely not respect settings in an aws config file.

## Changes

- Do not force the retry mode to be 'standard' if it is set already via environment variable

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/ecs/agent.py
+++ b/src/prefect/agent/ecs/agent.py
@@ -180,7 +180,7 @@ class ECSAgent(Agent):
         # See https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html
         # for more info.
         boto_config = Config(**botocore_config or {})
-        if not boto_config.retries:
+        if not boto_config.retries and not os.environ.get("AWS_RETRY_MODE"):
             boto_config.retries = {"mode": "standard"}
 
         self.boto_kwargs = dict(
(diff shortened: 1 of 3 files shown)


src/prefect/agent/ecs/agent.py
```diff
@@ -180,7 +180,7 @@ def __init__(  # type: ignore
         # See https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html
         # for more info.
         boto_config = Config(**botocore_config or {})
-        if not boto_config.retries:
+        if not boto_config.retries and not os.environ.get("AWS_RETRY_MODE"):
```
zangell44, line 183: would we want to actually set the value from the environment in `boto_config` here? or is this environment variable picked up later in the process?

zangell44 (approved):

Issue #4376, opened 2021-04-08
Alter GitHub Storage to support using deploy keys and personal access tokenss


dylanbhughes:
## Intro
Several users have requested that we support passing Deploy Keys to GitHub storage. Example: https://prefect-community.slack.com/archives/CL09KU1K7/p1617659682480500

## Current behavior

We support passing a GitHub Personal Access Token.

## Proposed behavior

Also instead accept a Deploy Key as a way to authenticate to the repository where the Flow code lives.

## Example

Deploy Keys are scoped to a single repository and their use is in line with better security practices.

madkinsz:
To clarify this: Using a deploy key does not work with our current model because we pull a single file from your repository with a PAT but you can only clone full repos with deploy keys; this doesn't mean we can't do it just that it's not a simple change.

zhodowanec:
@madkinsz it's an odd observation that pulling a single file would require a PAT rather than a deploy key. As @dylanbhughes mentioned in the ticket submission 'Deploy Keys are more secure and where the industry is moving as they are limited to a single repository'.

The Prefect team supporting Deploy Keys is paramount to our team being able to adopt this tool.

madkinsz:
> @madkinsz it's an odd observation that pulling a single file would require a PAT rather than a deploy key. As @dylanbhughes mentioned in the ticket submission 'Deploy Keys are more secure and where the industry is moving as they are limited to a single repository'.

It's simply how the Github API works. We'll have to move to a method that no longer uses the API. 

> The Prefect team supporting Deploy Keys is paramount to our team being able to adopt this tool.

We will support them, we're tracking this issue in combination with #4328 which seems like a logical first step (moving away from API based git storage retrieval)

zhodowanec:
Thanks for the update. Any timeframe on release?

zangell44:
@zhodowanec Core 0.14.17 supports the use of Deploy Keys with a new `Git` storage class. More info here https://docs.prefect.io/orchestration/flow_config/storage.html#git

Issue #4328, opened 2021-04-01
Add support to Storages to pull whole repo


xyxz-web:
## Current behavior
If I'm not mistaken, currently only the flow file specified in the path attribute of GitHubStorage, GitLabStorage and similar is pulled.

## Proposed behavior
It would be great if there was an option to either pull the whole repo or at least certain directories, as an option for storing dependencies that are shared among different flows.

## Example
```
 GitHub(
        repo="my/repo",
        access_token_secret="GITHUB_ACCESS_TOKEN",
        path=source_path,
        pull_whole_repo=True,
    )
```

willianchan:
Same problem here, it would help a lot to be able to pull the entire repository

kst83:
Just wondering... since this ticket has been assigned, does it mean it is already being worked on?

madkinsz:
@kst83 we're looking at the feasibility of implementing this across our git storage options, the issue will be updated once we have more news.

sm-Fifteen:
One can currently use `Module` storage for that [to have the shared dependencies be recognized](https://github.com/PrefectHQ/prefect/issues/4233), though with the additional caveat that repo syncing has to be done by hand. If multiple flows are contained within the same module, and they're all running in the same working directory, this could get a bit race-y.

Syncing full repos instead of individual files could probably be done with a generic `GitModule` storage backend using something like libgit2 (`dulwich`, `pygit2`, etc.) instead of APIs specific to Bitbucket, Github, Gitlab, and so on. If individual flows and flow versions are tied to specific git commits ( instead of assuming that the latest registered flow version always matches the HEAD of the tracked branch), this might require a bit of extra magic to ensure flows can run on separate copies of the checked out files.

madkinsz:
> Syncing full repos instead of individual files could probably be done with a generic `GitModule` storage backend using something like libgit2 (`dulwich`, `pygit2`, etc.) instead of APIs specific to Bitbucket, Github, Gitlab, and so on. If individual flows and flow versions are tied to specific git commits ( instead of assuming that the latest registered flow version always matches the HEAD of the tracked branch), this might require a bit of extra magic to ensure flows can run on separate copies of the checked out files.

This is an interesting idea. Perhaps we could have a common base class but I'd be a bit concerned about supporting the various authentication patterns with a single implementation.

sm-Fifteen:
[...]

PR #4418, opened 2021-04-20
Add full repo storage

zangell44:
## Summary
This PR adds a generalized git storage class to Prefect.

## Changes

- Adds a new `Git` storage class which uses a pure python git implemenation
- Adds a new `TemporaryGitRepo` utility, allowing us to temporary clone and cleanup a repository

## Importance
These changes will
- Allow users to use Deploy Tokens and other ssh based git auth if they wish
- Allow us to become less reliant on third party libraries
- Allow users to pull info from non-python files at run time (python modules are another beast)

These changes address
- https://github.com/PrefectHQ/prefect/issues/4376
- https://github.com/PrefectHQ/prefect/issues/4328

These changes WILL NOT
- Allow users to import Python modules from the repo when cloning

## Example usage
```python
import prefect
from prefect.storage import Git

storage = Git(flow_path='flow.py', repo='zangell44/single-prefect-flow', repo_host='github.com')

@task
def say_hello(name):
        logger.info('Hi' + name)

with Flow('my-hello-flow') as flow:
        say_hello('Zach')

flow.storage = storage
```

## Checklist

This PR:

### DIFF
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -248,6 +248,29 @@ which means both upload (build) and download (local agent) times need to have
 the proper Google Application Credentials configuration.
 :::
 
+### Git
+[Git Storage](/api/latest/storage.md#git) is a storage option for referencing flows
+stored in a git repository as `.py` files.
+
+```python
+from prefect import Flow
+from prefect.storage import Git
+
+# using https by default
+storage = Git(
+    repo="org/repo",                            # name of repo
+    path="flows/my_flow.py",                    # location of flow file in repo
+    git_token_secret_name="MY_GIT_ACCESS_TOKEN" # name of personal access token secret
+)
+
+# using ssh (environment must be configured for ssh access to repo)
+storage = Git(
+    repo="org/repo",                            # name of repo
+    path="flows/my_flow.py",                    # location of flow file in repo
+    use_ssh=True                                # use ssh for cloning repo
+)
+```
+
 ### GitHub
 
 [GitHub Storage](/api/latest/storage.md#github) is a storage
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -285,6 +285,20 @@ You can find more information about configuring Deploy Keys for common providers
 - [BitBucket](https://bitbucket.org/blog/deployment-keys)
 :::
 
+:::tip GitLab Deploy Tokens
+To use `Git` storage with GitLab Deploy Tokens, first create a Secret storing your Deploy Token. Then, you can configure `Git` storage
+```
+storage = Git(
+    repo="org/repo",                            # name of repo
+    path="flows/my_flow.py",                    # location of flow file in repo
+    repo_host="gitlab.com",                     # repo host name, which may be custom
+    git_token_secret_name="MY_GIT_ACCESS_TOKEN",# name of Secret containing Deploy Token
+    git_token_username="myuser"                 # username associated with the Deploy Token
+)
+```
+
+:::
+
 ### GitHub
 
 [GitHub Storage](/api/latest/storage.md#github) is a storage
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -260,6 +260,7 @@ from prefect.storage import Git
 storage = Git(
     repo="org/repo",                            # name of repo
     path="flows/my_flow.py",                    # location of flow file in repo
+    repo_host="github.com",                     # repo host name
     git_token_secret_name="MY_GIT_ACCESS_TOKEN" # name of personal access token secret
 )
 
@@ -267,6 +268,7 @@ storage = Git(
 storage = Git(
     repo="org/repo",                            # name of repo
     path="flows/my_flow.py",                    # location of flow file in repo
+    repo_host="github.com",                     # repo host name
     use_ssh=True                                # use ssh for cloning repo
 )
 ```
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -252,6 +252,8 @@ the proper Google Application Credentials configuration.
 [Git Storage](/api/latest/storage.md#git) is a storage option for referencing flows
 stored in a git repository as `.py` files.
 
+This storage class uses underlying git protocol instead of specific client libaries (e.g. `PyGithub` for GitHub), superseding other git based storages
+
 ```python
 from prefect import Flow
 from prefect.storage import Git
@@ -264,7 +266,8 @@ storage = Git(
     git_token_secret_name="MY_GIT_ACCESS_TOKEN" # name of personal access token secret
 )
 
-# using ssh (environment must be configured for ssh access to repo)
+# using ssh, including Deploy Keys
+# (environment must be configured for ssh access to repo)
 storage = Git(
     repo="org/repo",                            # name of repo
     path="flows/my_flow.py",                    # location of flow file in repo
@@ -272,6 +275,15 @@ storage = Git(
     use_ssh=True                                # use ssh for cloning repo
 )
 ```
+
+:::tip Git Deploy Keys
+To use `Git` storage with Deploy Keys, ensure your environment is configured to use Deploy Keys. Then, create a `Git` storage class with `use_ssh=True`.
+
+You can find more information about configuring Deploy Keys for common providers here:
+- [GitHub](https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys)
+- [GitLab](https://docs.gitlab.com/ee/user/project/deploy_keys/)
+- [BitBucket](https://bitbucket.org/blog/deployment-keys)
+:::
 
 ### GitHub
 
--- a/src/prefect/storage/git.py
+++ b/src/prefect/storage/git.py
@@ -74,6 +74,12 @@ class Git(Storage):
                 "Either `tag` or `branch_name` can be specified, but not both"
             )
 
+        if use_ssh and git_token_secret_name is not None:
+            self.logger.warning(
+                "Git Storage initialized with `use_ssh = True` and `git_token_secret_name` provided. "
+                "SSH will be used to clone the repository. `git_token_secret_name` will be ignored"
+            )
+
         self.flow_path = flow_path
         self.repo = repo
         self.repo_host = repo_host
--- a/src/prefect/storage/git.py
+++ b/src/prefect/storage/git.py
@@ -27,7 +27,7 @@ class Git(Storage):
     - Call `prefect register -f flow.py` to register this flow with Git storage.
     Args:
         - flow_path (str): A file path pointing to a .py file containing a flow
-        - repo (str): the name of a GitHub repository to store this Flow
+        - repo (str): the name of a git repository to store this Flow
         - repo_host (str, optional): The site hosting the repo. Defaults to 'github.com'
         - flow_name (str, optional): A specific name of a flow to extract from a file.
             If not set then the first flow object retrieved from file will be returned.
@@ -137,7 +137,7 @@ class Git(Storage):
         access_token = Secret(self.git_token_secret_name).get()
         if self.format_access_token:
             return f"{self.git_token_username}:{access_token}"
-        return access_token
+        return str(access_token)
 
     @property
     def git_clone_url(self) -> str:
--- a/src/prefect/storage/git.py
+++ b/src/prefect/storage/git.py
@@ -15,16 +15,23 @@ class Git(Storage):
     """
     Git storage class. This class represents the Storage interface for Flows stored
     in `.py` files in a git repository.
+
     This class represents a mapping of flow name to file paths contained in the git repo,
-    meaning that all flow files should be pushed independently. A typical workflow using
-    this storage type might look like the following:
+    meaning that all flow files should be pushed independently.
+
+    A typical workflow using this storage type might look like the following:
+
     - Compose flow `.py` file where flow has Git storage:
+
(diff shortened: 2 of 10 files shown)


src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(
+        self,
+        flow_path: str,
+        repo: str,
+        repo_host: str = "github.com",
+        flow_name: str = None,
+        git_token_secret_name: str = None,
+        git_token_username: str = None,
+        branch_name: str = None,
+        tag: str = None,
```
madkinsz, line 59: Can I use a commit #?

src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+        - git_token_username (str, optional): the username associated with git access token,
+            if not provided it will default to repo owner
+        - branch_name (str, optional): branch name, if not specified and `tag` not specified,
+            repo default branch will be used
+        - tag (str, optional): tag name, if not specified and `branch_name` not specified,
+            repo default branch will be used
+        - clone_depth (int): the number of history revisions in cloning, defaults to 1
+        - use_ssh (bool): if True, cloning will use ssh. Ssh keys must be correctly
+            configured in the environment for this to work
+        - format_access_token (bool): if True, the class will attempt to format acess tokens
```
madkinsz, line 45:
```suggestion
        - format_access_token (bool): if True, the class will attempt to format access tokens
```

src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+            )
+
+        self.flow_path = flow_path
+        self.repo = repo
+        self.repo_host = repo_host
+        self.flow_name = flow_name
+        self.git_token_secret_name = git_token_secret_name
+        self.git_token_username = (
+            git_token_username if git_token_username else repo.split("/")[0]
+        )
```
madkinsz, line 77: This needs a comment at least

src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+            return f"{self.git_token_username}:{access_token}"
+        return str(access_token)
+
+    @property
+    def git_clone_url(self) -> str:
+        """
+        Build the git url to clone
+        """
+        if self.use_ssh:
+            return f"git@{self.repo_host}:{self.repo}.git"
```
madkinsz, line 148: Should probably warn if a secret is set here

src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+                'Name conflict: Flow with the name "{}" is already present in this storage.'.format(
+                    flow.name
+                )
+            )
+        self.flows[flow.name] = self.flow_path
+        self._flows[flow.name] = flow
+        return self.flow_path
+
+    @property
+    def git_token_secret(self) -> str:
```
madkinsz, line 129: Should probably support loading from an environment variable too so they don't need to use Prefect secrets to set this

madkinsz (changes requested):
Good start! Will definitely need to be added to the docs

src/prefect/storage/git.py
zangell44: not right now as implemented, I can do some more digging in `dulwich` to see what's possible here

is commit # a common enough pattern that it's must-have?

src/prefect/storage/git.py
```diff
@@ -0,0 +1,149 @@
...
+from prefect.storage import Storage
+from prefect.utilities.storage import extract_flow_from_file
+from prefect.utilities.git import TemporaryGitRepo
+
+if TYPE_CHECKING:
+    from prefect.core.flow import Flow
+
+
+class Git(Storage):
+    """
```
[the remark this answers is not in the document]
zangell44: oof this looked ugly in the docs

src/prefect/storage/git.py
zangell44: Hmm are there other good examples of how we'd want this pattern to work?

My cheeky response is, "just set the environment variable `export PREFECT__CONTEXT__SECRETS__MYSECRET="MY SECRET VALUE"`:".

src/prefect/storage/git.py
zangell44: added a warning in __init__, hopefully will allow users to catch it right away

src/prefect/storage/git.py
madkinsz: I guess every git tool I've worked with takes branch/commit hash/tag indiscriminately and I expect the same. If it's at all a pain I don't think it should hold this up but I'd make a follow up issue.

src/prefect/storage/git.py
madkinsz: Idk how strongly I feel but the pattern used in all the other storage's is like so
```python
        if self.access_token_secret is not None:
            # If access token secret specified, load it
            access_token = Secret(self.access_token_secret).get()
        else:
            # Otherwise, fallback to loading from local secret or environment variable
            access_token = prefect.context.get("secrets", {}).get("GITHUB_ACCESS_TOKEN")
            if access_token is None:
                access_token = os.getenv("GITHUB_ACCESS_TOKEN")
 ```

cc @cicdw @dylanbhughes on continuing to support this pattern

Comment (zangell44) [reply src/prefect/storage/git.py:146]: I would vote to stop supporting this pattern

Comment (dylanbhughes) [reply src/prefect/storage/git.py:146]: I defer to @cicdw. My feedback is that the user shouldn't have to stick with a particular name for a Prefect secret so that a single tenant can have more than one secret for GitHub storage and that's well-documented. As long as that's achieved I'm 👍

File: src/prefect/storage/git.py (hunk 1 above)
Comment (zangell44) [reply src/prefect/storage/git.py:68]: turned out to be easy enough, just added

File: src/prefect/storage/git.py [hunk 7]
```diff
@@ -0,0 +1,171 @@
...
+        return str(access_token)
+
+    @property
+    def git_clone_url(self) -> str:
+        """
+        Build the git url to clone
+        """
+        if self.use_ssh:
+            return f"git@{self.repo_host}:{self.repo}.git"
+        return f"https://{self.git_token_secret}@{self.repo_host}/{self.repo}.git"
```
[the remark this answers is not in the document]
zangell44: Fair point, I went back and forth on building url from params vs. letting the user provide their own url.
Is there a repo configuration you've worked with where this would run into compatibility issues?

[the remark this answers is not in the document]
madkinsz: I think having an option to provide a custom url may be reasonable but not strictly necessary in this first patch. We want this to be easy for people to use, we've got a lot of users that are data scientists.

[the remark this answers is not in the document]
sm-Fifteen: Well, if that's any better an argument, having to specify the URL in full would also make the `use_ssh` parameter redundant, since it would now be specified in the URL's protocol rather than added by the storage module.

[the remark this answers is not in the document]
zangell44: I agree with Michael here, I think this is reasonable to release and expand to include explicit URLs if necessary.

Thank you for taking a look at this PR in detail and detailed comments on the issue thread!

src/prefect/utilities/git.py
```diff
@@ -0,0 +1,83 @@
+from tempfile import TemporaryDirectory
+from typing import Any
+
+from dulwich.index import build_index_from_tree
+from dulwich.porcelain import clone
+
+
+class TemporaryGitRepo:
```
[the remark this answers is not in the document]
zangell44: Another fair point. Again, it's reasonable to release this as is and make updates as necessary if users run into performance issues.

I'll keep an eye out for any reports. Thanks for the detail here!

docs/orchestration/flow_config/storage.md
```diff
@@ -248,6 +248,31 @@ which means both upload (build) and download (local agent) times need to have
 the proper Google Application Credentials configuration.
 :::
 
+### Git
+[Git Storage](/api/latest/storage.md#git) is a storage option for referencing flows
+stored in a git repository as `.py` files.
```
madkinsz, line 253: Add a note on how this differs from other git based storages (ie how it basically supersedes them).

docs/orchestration/flow_config/storage.md
```diff
@@ -248,6 +248,31 @@ which means both upload (build) and download (local agent) times need to have
...
+```python
+from prefect import Flow
+from prefect.storage import Git
+
+# using https by default
+storage = Git(
+    repo="org/repo",                            # name of repo
+    path="flows/my_flow.py",                    # location of flow file in repo
+    repo_host="github.com",                     # repo host name
+    git_token_secret_name="MY_GIT_ACCESS_TOKEN" # name of personal access token secret
```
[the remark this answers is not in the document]
madkinsz: We should have a basic example of using a deploy key. Even something like

> To use a private repository, you'll have to authenticate. We recommend using a deploy key (link to git docs?) stored as a Prefect Secret which you'll reference in the storage initialization

Also, have you tested using a secret in Cloud? I feel like sometimes you have to include `secrets=["MY_GIT_ACCESS_TOKEN"]` in the storage init call which I feel like is a bad ux and we should fix if it's required here.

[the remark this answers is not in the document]
zangell44: On deploy keys, we can link to the docs here https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys. 

As far as I can tell, they're just slightly enhanced ssh tokens. Maybe worth noting in the docs the two are equivalent and to set `use_ssh = True`?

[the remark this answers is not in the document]
zangell44: confirmed this works with Cloud Secrets
`LocalAgent` + `use_local_secrets=false` + using a secret that doesn't exist locally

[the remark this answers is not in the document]
zangell44: docs also updated accordingly

docs/orchestration/flow_config/storage.md
```diff
@@ -273,6 +276,15 @@ storage = Git(
 )
 ```
 
+:::tip Git Deploy Keys
+To use `Git` storage with Deploy Keys, ensure your environment is configured to use Deploy Keys. Then, create a `Git` storage class with `use_ssh=True`.
```
madkinsz, line 280: Does this mean that if I use something like a K8s agent I have to setup my deploy key in my agent container? Why can't I use a Prefect Secret for that value?

zangell44: Yes you would have to configure it in the container (or use access tokens as a Secret).

Using a Secret for that value would involve manipulation of ssh configs, which I think is outside of Prefect's scope

madkinsz: Ah so the confusing part here is gitlab has deploy _tokens_

https://docs.gitlab.com/ee/user/project/deploy_tokens/#git-clone-a-repository

zangell44: Haha yes I was confused about this as well. Deploy Tokens will work like other access tokens without issues

madkinsz: If you don't add a note about that we're going to get some questions haha, just this one last thing and this is good to merge :D

zangell44: Added a note and example here https://github.com/PrefectHQ/prefect/pull/4418/commits/9ada00c0950984503385fb211a2a297750b587f2#diff-357fea1f44b9d331a93957d9d6b6408d2ffc07866872dfff6f91f2bb7af3a83cR288

madkinsz (approved):

mahiki:
Thus is a really helpful improvement

zangell44:
Glad to hear @mahiki! Please let us know if there's anything we can do to improve it.

jasonprado:
Hiya, I'm fairly new to Prefect. Can you explain why "Allow users to import Python modules from the repo when cloning" is not part of this–is it due to some structural way imports are handled inside the executor, or is it a non-goal in terms of design? I suspect the former as you said "python modules are another beast".

I have been surprised that getting script-based storage deployed is so easy but trying to share code between flows required deploying a custom docker image.

madkinsz:
Hey @jasonprado -- due to the myriad of execution environments we support, it's a more complex problem to ensure the rest of your code is accessible alongside your flow. This is something we want to address eventually, but running your flow in a container is the best way to have control over its environment. I think can use `EXTRA_PIP_PACKAGES` (see https://github.com/PrefectHQ/prefect/pull/4013) to install your module in the container still e.g. 

```python
flow.run_config = DockerRun(env=dict(EXTRA_PIP_PACKAGES="git+https://github.com/USER/REPO@COMMIT#egg=MODULE"))
```

I haven't given this a test but it should do what you want. Generalizing this pattern into this storage mode is something we're interested in pursuing once we understand the best way to support it across our deployment methods.

willianchan:
i am getting the following error using the Git Storage module:

`dulwich.client.HTTPUnauthorized: No valid credentials provided`

using the following code:

`flow.storage = Git(
    repo="GRG_GIT_CD/analytics/workflow-registration",
    flow_path="flow-test-clone-repo.py",
    repo_host="https://gitcorp.prod.cloud.ihf"
)`

i am using gitlab to store my flow, also, tried passing git_token_secret_name and git_token_username, am i doing something wrong?

al-yap:
@willianchan wondering if you have managed to resolve your issue above? I am currently troubleshooting the same error..

tyndyll:
> `flow.storage = Git( repo="GRG_GIT_CD/analytics/workflow-registration", flow_path="flow-test-clone-repo.py", repo_host="https://gitcorp.prod.cloud.ihf" )`

If GRG_GIT_CD is a URL it should probably be added as `repo_host`

PR #4419, opened 2021-04-20
Fix BigQuery: logging errors if exception is thrown during load_table_from_uri

joelluijmes:
…_from_uri

## Summary

Fixes logging of error when exception is thrown during BigQuery load_table_from_uri. Currently another exception is thrown due invalid handling.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/gcp/bigquery.py
+++ b/src/prefect/tasks/gcp/bigquery.py
@@ -379,16 +379,25 @@ class BigQueryLoadGoogleCloudStorage(Task):
         job_config = bigquery.LoadJobConfig(autodetect=autodetect, **kwargs)
         if schema:
             job_config.schema = schema
+
+        load_job = None
         try:
             load_job = client.load_table_from_uri(
                 uri,
                 table_ref,
                 location=location,
                 job_config=job_config,
-            ).result()  # block until job is finished
+            )
+
+            # Start the job and wait for it to complete and get the result
+            load_job.result()
         except Exception as exception:
-            for error in load_job.errors:
-                self.logger(error)
+            self.logger.error(exception)
+
+            if load_job is not None and load_job.errors is not None:
+                for error in load_job.errors:
+                    self.logger.error(error)
+
             raise FAIL(exception) from exception
         # remove unpickleable attributes
         load_job._client = None
--- a/changes/pr4419.yaml
+++ b/changes/pr4419.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "BigQuery: logging errors if exception is thrown during load_table_from_uri - [#4419](https://github.com/PrefectHQ/prefect/pull/4419)"
+
+contributor:
+  - "[Joël Luijmes](https://github.com/joelluijmes)"


src/prefect/tasks/gcp/bigquery.py
```diff
@@ -387,8 +387,9 @@ def run(
                 job_config=job_config,
             ).result()  # block until job is finished
```
zangell44, line 388: I think it makes sense to split up the initialization and `result` calls here

```python
load_job = client.load_table_from_uri(
                uri,
                table_ref,
                location=location,
                job_config=job_config,
            )
load_job.result()

File: src/prefect/tasks/gcp/bigquery.py [hunk 2]
```diff
@@ -387,8 +387,9 @@ def run(
...
         except Exception as exception:
-            for error in load_job.errors:
-                self.logger(error)
+            for error in exception.errors:
+                self.logger.error(error)
```
zangell44, line 391: After initializing `load_job` correctly (see above), `load_job.errors` might actually contain some useful information. See thread here https://github.com/googleapis/google-cloud-python/issues/6573.

Something like the following might work
```python
except Exception as exception:
            self.logger.error(exception)
            try:
                for error in load_job.errors:
                    self.logger.exception(error)
            except NameError:
                pass
            raise FAIL(exception) from exception
```

zangell44 (changes requested):
Thanks for taking time to update the error handling @joelluijmes!

A couple suggestions to make sure we also capture additional error info if present

src/prefect/tasks/gcp/bigquery.py
joelluijmes: Ah good suggestion! I didn't actually dive in to which errors[] the message was referring. But I opted to log everything :)

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -385,10 +385,21 @@ def run(
...
+            )
+
+            # Start the job and wait for it to complete and get the result
+            load_job.result()
         except Exception as exception:
-            for error in load_job.errors:
-                self.logger(error)
+            self.logger.error(exception)
+
+            if exception.errors is not None:
```
zangell44, line 395: I think we can remove this. I'm not sure what exception gets thrown here (and if it has an `errors` attribute, but we're already logging the full exception.

src/prefect/tasks/gcp/bigquery.py
```diff
@@ -385,10 +385,21 @@ def run(
...
+                for error in exception.errors:
+                    self.logger.error(error)
+
+            if load_job is not None and load_job.errors is not None:
```
zangell44, line 399: for this to work, we'll need to initalize`load_job = None` before calling `load_job = client.load_table_from_url(...)`

there are two possible sources of errors here
1. an error in initializing `load_job = client.load_table_from_url()`
2. an error in running `load_job` (`load_job.result()`)

if (1) occurs, `load_job` will be undefined, and we'll get a `NameError` here

zangell44 (changes requested):
A few minor feedback points, looking good!

joelluijmes:
Forgot about this PR 🙊 but I've removed the unnneeded for loop 👍

jcrist (approved):

PR #4426, opened 2021-04-21
Run Flow API: Add objects for interacting with flows, flow runs, and task runs

madkinsz:
See "Run Flow API" summary at #4434 

## Summary

This is a piece of my exploration with improving the flow run API in core. I've split my main exploration branch to facilitate review. This piece creates an object-based approach for interacting with flow runs and task runs.

Notably, task run results (return values) can be lazily loaded from flow runs via a simple interface.

## Changes

- Adds `prefect.backend` module
- Adds `FlowRunView`, `FlowView` and `TaskRunView` objects
- Adds `prefect.backend.flow_run.execute_flow_run` which basically replicates (and cleans up) the logic in `prefect execute flow-run` for future consolidation
- Some minor fixes to `EnumValue` to give it an equality operator and readable repr

## Importance

- Improves the user's ability to interact with the API from core
- Clarifies the concept of `FlowRun` and `TaskRun` in core

## Checklist

This PR:

## Example

```python
import os

from prefect import Flow, task, Task, Client
from prefect.executors import LocalDaskExecutor
from prefect.storage import Local
from prefect.backend.flow_run import execute_flow_run


# Define a simple flow -----------------------------------------------------------------


@task(log_stdout=True)
def say(message):
    print(message)


@task
def returns(x):
    return x


with Flow("Example: Cloud Run") as flow:
    hello = say("Hello world!")

    return_task = returns("A return value!")

    # Mapped return values
    mapped_return_task = returns.map([0, 1, 2])

    say(mapped_return_task[2])

# By default, local storage includes a timestamp which will break the cool lookup we
# do downstream. We'll want to change the default to use a consistent location instead
flow.storage = Local(
    path=os.path.expanduser(
        os.path.join("~", ".prefect", "flows", flow.serialized_hash())
    )
)
flow.executor = LocalDaskExecutor()

# Demo flow run and execution ----------------------------------------------------------

flow_id = flow.register("default", idempotency_key=flow.serialized_hash())

# We can create a flow run using the flow object
# Note: We lookup the flow id for a flow with matching serialized data, we could also
# use the `flow_id` directly here but this means we could run the flow without calling
# flow.register() every time
# --- This example no longer works as written because `create_flow_run` was dropped
flow_run = create_flow_run(flow)


# We can inspect the flow run with a rich object
print(f"Created flow run: {flow_run}")
assert flow_run.state.is_scheduled()
assert isinstance(flow_run.flow.storage, Local)

# We can execute the flow run here instead of in an agent
execute_flow_run(flow_run_id=flow_run.flow_run_id)

# `execute_flow_run` will return an updated `flow_run` object, but we can also get the
# newest data ourselves
flow_run = flow_run.get_latest()
assert flow_run.state.is_finished()

# We can get all the task run ids
print(f"Ran {len(flow_run.task_run_ids)} tasks!")

# We can get task run data for specific tasks, with lookup using the task object or slug
task_run = flow_run.get_task_run(return_task)
print(f"Got task run: {task_run}")

# Once we've retrieved a task once, it's cached so we aren't actually querying again
assert flow_run.get_task_run(task=return_task) is flow_run.get_task_run(
    task_slug="returns-1"
)

# We can load the task return value from the `Result` location
print(f"The {mapped_return_task.slug} returned: {task_run.result!r}")
# this value is also cached so we don't have to read from the result location repeatedly

# We can also load mapped task runs which will load all child task run results
print(
    f"The mapped return task returned: {flow_run.get_task_run(mapped_return_task).result}"
)

# If our mapping is very large, we can lazily iterate over the results to avoid querying
# for them all at once or holding them all in memory
for task_run in flow_run.iter_mapped_task_runs(mapped_return_task):
    print(f"Mapping at index {task_run.map_index} returned {task_run.result!r}")

```


src/prefect/backend/flow.py
```diff
@@ -0,0 +1,283 @@
...
+            many (optional): Are many results expected? If `False`, a single record will
+                be returned and if many are found by the `where` clause an exception
+                will be thrown. If `True` a list of records will be returned.
+            order_by (optional): An optional Hasura `order_by` clause to order results
+                by. Only applicable when `many` is `True`
+            error_on_empty (optional): If `True` and no tasks are found, a `ValueError`
+                will be raised. If `False`, an empty list or dict will be returned
+                based on the value of `many`.
+            jsonb_variables (optional): Dictionary variables to inject into the query
+                as jsonb GraphQL types.
```
madkinsz, line 192:
```suggestion
            jsonb_variables (optional): Dict-typed variables to inject into the query
                as jsonb GraphQL types. Keys must be consumed in the query i.e.
                in the passed `where` clause as `EnumValue("$key")`
```

src/prefect/backend/flow.py
```diff
@@ -0,0 +1,283 @@
...
+        self.serialized_flow = serialized_flow
+        self.archived = archived
+        self.project_name = project_name
+        self.core_version = core_version
+        self.storage = storage
+        self.name = name
+
+    @classmethod
+    def from_flow_data(cls, flow_data: dict) -> "FlowMetadata":
+        """ "
```
zangell44, line 59:
```suggestion
        """
```

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+    logger.info(
+        f"Beginning execution of flow run {flow_run.name!r} from {flow_run.flow.name!r} "
+        f"with {runner_cls.__name__!r}"
+    )
+    with prefect.context(flow_run_id=flow_run_id):
+        with fail_flow_run_on_exception(
+            flow_run_id=flow_run_id,
+            message="Failed to execute flow: {exc}",
+        ):
+            if flow_run.flow.run_config is not None:
```
zangell44, line 136: I believe `run_config` can be None in cases in which we don't want to use `flow.environment`

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+                flow_state = runner_cls(flow=flow).run(**run_kwargs)
+
+            # Support for deprecated `flow.environment` use
```
zangell44, line 139: at what point do we we want to stop supporting this?

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+            else:
+                environment = flow.environment
+                environment.setup(flow)
+                environment.execute(flow)
+
+    logger.info(f"Run finished with final state {flow_state}")
```
zangell44, line 145: `flow_state` would be undefined here if using `environment.execute`

src/prefect/backend/flow_run.py
madkinsz: This logic is straight from `prefect execute flow-run` so I don't think this is true

src/prefect/backend/flow_run.py
madkinsz: I'm not sure when that deprecation cycle ends

madkinsz:
@dylanbhughes While we intend to introduce a new `StartFlowRun`-like task that makes result passing a better experience, this does allow users to pass results between flows with a relatively simple task definition

```python
import time
from prefect.backend import FlowRun
from prefect.tasks.prefect import StartFlowRun


@task
def get_task_result(flow_run_id: str, task_slug: str):

    task_run = FlowRun.from_flow_run_id(flow_run_id).get(task_slug=task_slug)

    while not task_run.state.is_finished():
        time.sleep(5)
        task_run = task_run.update()

    return task_run.result


with Flow("Example: Subflow Result Passing") as parent_flow:

    sub_flow_run_id = StartFlowRun(flow_name=flow.name, project_name="default")

    say(
        # Get the task result from the sub flow
        get_task_result(flow_run_id=sub_flow_run_id, task_slug="returns-1")
    )

    # Get a mapped task results
    mapped_result = get_task_result(flow_run_id=sub_flow_run_id, task_slug="returns-2")

    # Show all all the task results
    say(mapped_result)

    # Access an indexed item
    say(mapped_result[1])
```

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+
+class FlowRun:
+    """
+    Object for getting information about a Flow Run from the backend API
+
+    Attributes:
+        flow_run_id: The uuid of the flow run
+        name: The name of the flow run
+        flow_id: The uuid of the flow this run is associated with
+        state: The state of the flow run
```
zangell44, line 211: what's the use case for `state` here? would `state` be better as a property?

if I'm accessing `FlowRun.state` I would expect the latest state to be queried for and returned

madkinsz: `FlowRun` is intended to be an immutable view. We want to limit hitting the API as much as possible. If you want the current state you'd grab the latest data with `flow_run.update()` which creates a new instance.

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+        self._task_runs[task_run.task_run_id] = task_run
+        self._task_slug_to_task_run_ids[task_run.task_slug].add(task_run.task_run_id)
+
+    def update(self) -> "FlowRun":
+        """
+        Get the a new copy of this object with the newest data from the API
+        """
+        return self.from_flow_run_id(
+            flow_run_id=self.flow_run_id,
+            load_static_tasks=False,
```
zangell44, line 257: would we want to set this to `True` here? and maybe not pass in `_task_runs`?

Could be misinterpreting here, but seems like I'd want `update` to automatically get all the newest data

madkinsz: I think at the very least I should update this to only pass task runs that are in a finished state. I might even update the cache to only store task runs that are in a finished state so `get(task)` will always return new data unless the task is done.

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+
+        return cls(
+            flow_run_id=flow_run_id,
+            flow_id=flow_run.flow_id,
+            task_runs=task_runs,
+            state=State.deserialize(flow_run.serialized_state),
+            name=flow_run.name,
+        )
+
+    def get(
```
madkinsz, line 357: Should this be `get_task_run` ?

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,596 @@
...
+                # We did not find the base mapped task in the cache so we'll
+                # drop through to query for it
+
+            result = TaskRun.from_task_slug(
+                task_slug=task_slug, flow_run_id=self.flow_run_id
+            )
+            self._add_task_run(result)
+            return result
+
+        raise ValueError(
```
zangell44, line 439: minor preference to put this at the beginning of the method

src/prefect/backend/flow_run.py
zangell44, line 357: `get_task_run` and `get_all_task_runs` might make things clearer

src/prefect/backend/flow_run.py
madkinsz: Hmm that might make more sense now that I added checks for empty `task.slug` up there. It made more sense to fall through at first because I was inferring some values that could still be empty

src/prefect/backend/flow_run.py
zangell44: yes!

src/prefect/backend/flow_run.py
zangell44: I'm probably full of it then

src/prefect/backend/flow_run.py
zangell44: Got it, "immutable view" helped things click for me, thanks!

I'm not sure there's a good pattern to replace it, but I think the mutable nature of `task_runs` may cause issues, especially if this is intended for external use.

I'm thinking of doing something (deliberately stupid for illustrative purposes) like
```python
flow_run = create_flow_run(flow)
execute_flow_run(flow_run_id=flow_run.flow_run_id)

# okay now I want to check on my flow
flow_run.get(task=my_failed_task) # this task failed
# hmmm now I can see my task failed, did my flow run fail?
print (flow_run.state) # my flow run is Scheduled, wtf?
```

src/prefect/backend/task_run.py
```diff
@@ -0,0 +1,233 @@
...
+    ) -> None:
+        self.task_run_id = task_run_id
+        self.name = name
+        self.task_id = task_id
+        self.task_slug = task_slug
+        self.state = state
+        self.map_index = map_index
+        self.flow_run_id = flow_run_id
+
+        # Uses NotLoaded so to separate from return values of `None`
```
zangell44, line 49: niiiiiiiiiiice

src/prefect/backend/task_run.py
```diff
@@ -0,0 +1,233 @@
...
+                will be thrown. If `True` a list of records will be returned.
+            order_by (optional): An optional Hasura `order_by` clause to order results
+                by. Only applicable when `many` is `True`
+            error_on_empty (optional): If `True` and no tasks are found, a `ValueError`
+                will be raised. If `False`, an empty list or dict will be returned
+                based on the value of `many`.
+
+        Returns:
+            A dict of task run information (or a list of dicts if `many` is `True`)
+        """
```
zangell44, line 160:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow_run.py [hunk 12]
```diff
@@ -0,0 +1,596 @@
...
+            map_index += 1
+
+    def get_all(self) -> List["TaskRun"]:
+        """
+        Get all task runs for this flow run in a single query. The results are cached
+        so future lookups do not query the backend.
+
+        Returns:
+            A list of TaskRun objects
+        """
```
zangell44, line 516:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow_run.py [hunk 13]
```diff
@@ -0,0 +1,596 @@
...
+            task: A `prefect.Task` object to use for the lookup. The slug will be
+                pulled from the task to actually perform the query
+            task_slug: A task slug string to use for the lookup
+            cache_results: By default, task run data is cached for future lookups.
+                However, since we lazily generate the mapped results, caching can be
+                disabled to reduce memory consumption for large mapped tasks.
+
+        Yields:
+            A TaskRun object for each mapped item
+        """
```
zangell44, line 465:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow_run.py [hunk 14]
```diff
@@ -0,0 +1,596 @@
...
+
+        Args:
+            task: A `prefect.Task` object to use for the lookup. The slug will be
+                pulled from the task to actually perform the query
+            task_slug: A task slug string to use for the lookup
+            task_run_id: A task run uuid to use for the lookup
+
+        Returns:
+            A cached or newly constructed TaskRun instance
+        """
```
zangell44, line 377:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow_run.py [hunk 15]
```diff
@@ -0,0 +1,596 @@
...
+        Args:
+            flow_run_id: the flow run id to lookup
+            load_static_tasks: Pre-populate the task runs with results from flow tasks
+                that are unmapped.
+            _task_runs: Pre-populate the task runs with an existing iterable of task
+                runs
+
+        Returns:
+            A populated `FlowRun` instance
+        """
```
zangell44, line 307:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow_run.py [hunk 16]
```diff
@@ -0,0 +1,596 @@
...
+
+    Args:
+        flow: A flow object to lookup a flow id with
+        flow_id: A flow id to use for lookup
+        **kwargs: Additional kwargs are passed to `client.create_flow_run` and can
+            be used to configure the flow run
+
+    Returns:
+        An instance of "FlowRun"
+    """
```
zangell44, line 37:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow.py [hunk 17]
```diff
@@ -0,0 +1,284 @@
...
+                in the passed `where` clause as `EnumValue("$key")`
+
+
+        Only `jsonb` variables are exposed because GraphQL queries will fail with where
+        clauses containing jsonb directly but succeed when they are a sent as query
+        variables because they are unescaped.
+
+        Returns:
+            A dict of task run information (or a list of dicts if `many` is `True`)
+        """
```
zangell44, line 202:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow.py [hunk 18]
```diff
@@ -0,0 +1,284 @@
...
+    def from_flow_id(cls, flow_id: str) -> "FlowMetadata":
+        """
+        Get an instance of this class given a `flow_id` to lookup
+
+        Args:
+            flow_id: The uuid of the flow
+
+        Returns:
+            A new instance of FlowMetadata
+        """
```
zangell44, line 90:
```python
Raises
  - ...
  ```

File: src/prefect/backend/flow.py [hunk 19]
```diff
@@ -0,0 +1,284 @@
...
+        run_config: A dict representation of the flow's run configuration
+        serialized_flow: A serialized copy of the flow
+        archived: A bool indicating if this flow is archived or not
+        project_name: The name of the project the flow is registered to
+        core_version: The core version that was used to register the flow
+        storage: The deserialized Storage object used to store this flow
+        name: The name of the flow
+
+
+    TODO: Consider changing this class name. We don't want to to overlap with
```
zangell44, line 29: +1, this naming could be confusing

src/prefect/backend/flow_run.py
zangell44: love that idea

src/prefect/backend/flow.py
madkinsz: Do you have a better name? :D

src/prefect/backend/flow_run.py
madkinsz: Well that's why `task_runs` isn't actually mutable and it's just a cache. So yes you got the newest data for that task run but you don't really know that it's being stored in `flow_run`. I think the big difference here is you're not calling `flow.state()` or `flow.get_state()` you're calling `flow.state` which is presented as a _property_ which hints that it is static.

src/prefect/backend/flow.py
```diff
@@ -0,0 +1,284 @@
...
+        searching for matches using the serialized flow
+
+        Args:
+            flow: The flow object to use
+            allow_archived: By default, archived flows are not included in the query
+                because it is possible that more than one flow will be found. If `True`
+                an archived flow can be returned.
+
+        Returns:
+            A new instance of FlowMetadata`
```
zangell44, line 113:
```suggestion
            A new instance of FlowMetadata
```

src/prefect/backend/flow.py
zangell44: Not yet! But I'm deliberately commenting and then not thinking about it, with hopes a better name will arrive in my brain later 😄

src/prefect/backend/task_run.py
madkinsz: I'm not going to add these until things are more finalized, but that's a good note

src/prefect/backend/flow_run.py
madkinsz: The names should be clearer now and the behavior has been changed so only finished tasks are cached

src/prefect/backend/task_run.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+
+        if not task_runs and error_on_empty:
+            raise ValueError(
+                f"No task runs found while querying for task runs where {where}"
+            )
+
+        return task_runs
+
+    def get_latest(self) -> "TaskRunView":
```
jcrist, line 223: Same here.

src/prefect/backend/flow.py
```diff
@@ -0,0 +1,355 @@
...
+                f"Found multiple flows matching {where}. "
+                "Provide a `project_name` as well or toggle `last_updated` "
+                "to use the flow that was most recently updated"
+            )
+
+        flow = flows[0]
+        return cls.from_flow_data(flow)
+
+    @staticmethod
+    def query_for_flow(where: dict, **kwargs: Any) -> dict:
```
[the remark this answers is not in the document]
madkinsz: Happy to make this change throughout

src/prefect/backend/flow_run.py
```diff
@@ -0,0 +1,643 @@
...
+        Args:
+            task_run: The task run to add
+        """
+        if task_run.state.is_finished():
+            self._cached_task_runs[task_run.task_run_id] = task_run
+            self._task_slug_to_task_run_ids[task_run.task_slug].add(
+                task_run.task_run_id
+            )
+
+    def get_latest(self, load_static_tasks: bool = False) -> "FlowRunView":
```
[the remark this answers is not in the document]
madkinsz: So my thinking here was

```python
flow_run  = FlowRunView.from_flow_run_id("...")
...
flow_run = flow_run.get_latest()
```

feels nicer than

```python
flow_run = FlowRunView.from_flow_run_id("...")
....
flow_run = FlowRunView.from_flow_run_id(flow_run.flow_run_id)   # eek
```

especially when the `FlowRunView` is returned from some other function and you don't necessarily have the `flow_run_id` sitting around in a variable.

I think this was further justified by the passing of cached task objects. When using `FlowRunView.get_latest()` we're passing all of the cached `TaskRunView`s along. This means that you can play with finished tasks while updating the flow's state without incurring additional network IO. This also skips loading the static tasks at initialization by default which can be a significant speed-up.

I think `TaskRunView.get_latest()` could also pass the cached `result` if it has been loaded from disk, but you probably won't have loaded that until the task run is done and `get_latest` is less relevant.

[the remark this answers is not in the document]
jcrist: Forwarding the cache of `TaskRunView` objects is one of the things I find weird about this. Calling `get_latest` will ensure the `FlowRunView` object is updated, but will leave the cache of `TaskRunView` objects outdated. I could see myself getting confused by this, so users will also likely find this confusing.

PR #4433, opened 2021-04-23
Run Flow API: Refactor `Client` and `Secrets` to live in `prefect.backend`

madkinsz:
See "Run Flow API" summary at #4434

Requires #4426 because it refactors references to the `Client` class used there.

## Summary

Now that there is a `prefect.backend` module (introduced in #4426) -- we should move the `Client` and `Secret` objects from `prefect.client.client.Client` and `prefect.client.secrets.Secret` respectively to `prefect.backend`.

## Changes

- Adds deprecation warnings to the old class locations for `Client` and `Secret`
- Moves `Client` to `prefect.backend.client.Client` 
- Moves `Secret` to `prefect.backend.secrets.Secret`

## Importance

The client lives in a weird place right now and this refactor begins to move the backend API to a first class position

## Checklist

This PR:


docs/core/concepts/secrets.md
```diff
@@ -33,10 +33,11 @@ For example, given an environment with the environment variable `PREFECT__CONTEX
...
 :::
 ::: tab Secret API
+
 ```python
->>> from prefect.client.secrets import Secret
->>> s = Secret("FOO")
->>> s.get()
+>> > from prefect.backend.secrets import Secret
+>> > s = Secret("FOO")
+>> > s.get()
```
madkinsz, line 40: Note to self, fixme

madkinsz:
Postponing this work as we will likely deprecate the existing `Client`/`Secrets` and reassess all the API access methods.

PR #4435, opened 2021-04-23
Fixed bug from Flow.copy() not copying the slugs dictionary

benfogelson:
## Summary

Fixes https://github.com/PrefectHQ/prefect/issues/4437, which caused a flow and its copies to refer to the same `slugs` dictionary.

## Changes
Copies the `slugs` dictionary as part of calling `Flow.copy()`.

## Importance
Not copying the `slugs` dictionary means that the dictionary can be modified by adding tasks to either the original flow or its copy, leading to `slugs` that are not in sync with the tasks in the flow.

## Checklist

This PR:

### DIFF
--- a/changes/pr4435.yaml
+++ b/changes/pr4435.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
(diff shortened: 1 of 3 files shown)


madkinsz (approved):
Thanks!

PR #4434, opened 2021-04-23
Run Flow API: Improve the API for running flows with the backend

madkinsz:
See the new docs at https://deploy-preview-4434--prefect-docs.netlify.app/orchestration/flow-runs/overview.html

## Summary

This PR is composed of several smaller PRs (for reviewability) that aim to improve the user experience for running flows that connect to the Cloud or Server backend APIs

- [x] Create backend module - #4426
- [x] Create objects for interacting with flow and task runs - #4426 
- [x] Update `prefect run <flow>` CLI - #4463 
- [x] Add first-class flow result passing - https://github.com/PrefectHQ/prefect/pull/4563
- [x] Update `Flow Run` documentation - https://github.com/PrefectHQ/prefect/pull/4492
- [ ] Add agentless flow run execution - #4589 (might be released after the rest)

## Future work

- Move `Client`/`Secrets` to backend module - #4433 (closed for now)
- Refactor `Client` to reuse connections and reduce auth collisions

PR #4439, opened 2021-04-26
Move `Client` methods to `prefect.backend` modules

madkinsz:
This draft outlines an example change that will be expanded to include all of the backend methods currently included in the `Client` class. This requires (but will not be included in) #4434.

## Summary

Currently, the `Client` contains all functions for interacting with the backend. Following the introduction of a dedicated backend module, it makes sense to move methods from within the client into properly scoped locations. This will make the `Client` a lot easier to reason about as it will have the singular task of managing an authenticated session with a Prefect backend (ie Cloud).

All methods requiring a `Client` will create one themselves. There will be required work to share sessions across clients so this is not unperformant and to determine if we need to allow every method to receive a custom `Client` instance.

## Changes

- Relocations
    - `Client.register` -> `prefect.backend.flow.register` 
- `prefect.backend.flow.register`
    - returns a `FlowView` instead of just a `flow_id`
    - drops support for displaying the url of the registered flow

## Checklist

This PR:

### DIFF
--- a/changes/pr4439.yaml
+++ b/changes/pr4439.yaml
@@ -1,0 +1,3 @@
+
+enhancement:
+  - "Move `Client` methods to `prefect.backend` modules - [#4439](https://github.com/PrefectHQ/prefect/pull/4439)"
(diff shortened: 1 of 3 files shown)


madkinsz:
Postponing this work as we will likely deprecate the existing `Client` and reassess all the API access methods.

PR #4441, opened 2021-04-26
Run black on files failing with newest black release

madkinsz:
e.g. failure at #4440


madkinsz:
Rolling into #4440 because neither will pass CI without the other fix.

PR #4445, opened 2021-04-27
Update KubernetetsRun run config docs

zschumacher:
## Summary

Update the `KubernetesRun` run config docs

## Changes

Adds suggestions for handling custom image tags other than `latest` on custom docker images in the `KubernetetsRun` run config

## Importance

It is likely going to be a common gotcha for anyone running the k8s agent

## Checklist

This PR:

### DIFF
--- a/docs/orchestration/flow_config/run_configs.md
+++ b/docs/orchestration/flow_config/run_configs.md
@@ -189,6 +189,39 @@ Use a custom Kubernetes Job spec for this flow, stored in S3:
 flow.run_config = KubernetesRun(job_template_path="s3://bucket/path/to/spec.yaml")
 ```
 
+It is important to note, that by default, kubernetes sets `imagePullPolicy` to `IfNotPresent`.  
+This means that if you pass in a custom image such as `image="example/image-name:dev"`, kubernetes will only pull
+the tag once, and never refresh it on subsequent updates.  
+  
+The behavior you want is likely having your kubernetes job always pull the latest image for the tag.  There are two ways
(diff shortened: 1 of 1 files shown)


zschumacher:
Not sure if I need any changes added for this, but if I missed something please let me know!

kvnkho:
+1 that this is needed. 3 people in support over the last month have had similar issues.

zangell44 (approved):
Looks good to me, thanks for putting this together @zschumacher !!

madkinsz:
Hey @zschumacher -- I appreciate that you took the time to do this! After taking a look at the example template you added, I think we should probably expose an image pull policy option as a first-class setting on the `KubernetesRun` to help users get past this without having to construct an entire template. I think after we do that we can revisit this doc.

zschumacher:
> Hey @zschumacher -- I appreciate that you took the time to do this! After taking a look at the example template you added, I think we should probably expose an image pull policy option as a first-class setting on the `KubernetesRun` to help users get past this without having to construct an entire template. I think after we do that we can revisit this doc.

@madkinsz agreed, that seems like the better solution.  Do you want me to take a stab at that in this PR?

madkinsz:
@zschumacher go for it, you'll probably get to it before us. We might be picky about the end result since RunConfig changes have to be supported for a long time though.

zschumacher:
@madkinsz @zangell44 @kvnkho new pr here https://github.com/PrefectHQ/prefect/pull/4462.  Feel free to close this one.

PR #4449, opened 2021-04-28
deprecate tokens, use API Keys

TylerWanner:
## Summary

Update docs and deprecate use of tokens in the client in favor of API Keys

## Changes

Update docs to reflect use of API Keys in place of tokens
Add list-keys, create-key, and revoke-key CLI commands
Deprecate list-tokens create-token and revoke-token CLI commands
Phase out token language from API for key
List-tenants and switch-tenants no longer blocked by check_override_auth_token

## Importance

Clear up usage of API Keys
Deprecate Token usage and language

## Checklist

This PR:


docs/orchestration/concepts/api.md
```diff
@@ -22,15 +22,7 @@ mutation {
...
-Prefect Cloud can generate two different types of API keys.
-
-#### User
-
-User-based API keys function as personal access keys. These API keys have very few permissions on their own, but can be used to authenticate with the Cloud API. Once authenticated, user-based API keys can be used to generate short-lived JWT auth tokens for any tenant the user belongs to. These API keys inherit any permissions the user has in that tenant, allowing full API access. The Client manages the process of provisioning and refreshing these tokens.
-
-#### Service Account
-
-Service Account-based API keys are used for processes like the Prefect Agent, which require the ability to execute flows on behalf of a tenant. Service Account keys are attached to their cross-tenant account, as opposed to an individual user. Consequently, they can only be generated by tenant admins.
+Prefect Cloud users can generate API Keys for their user, or create Service Accounts and API Keys associated with them.  For more information on hwo to manage these keys, go [here](api_keys.html)
```
madkinsz, line 25:
```suggestion
To authenticate with Prefect Cloud, an API Key is required. Users can generate a key associated with their user, or they can generate keys associated with a Service Account User. See [the API keys page](api_keys.html) for more details.
```

docs/orchestration/concepts/api_keys.md
```diff
@@ -1,20 +1,25 @@
 # API Keys <Badge text="Cloud"/>
 
-API keys are a central component for interacting with the Prefect Cloud API.  They encapsulate who you are as a user of the system along with your memberships to your various tenants.  API keys can also be generated for Service Accounts, allowing for keys with customizable permission sets.  Ultimately, all clients that interact with the Prefect Cloud API must provide an API key as a Bearer Token included in the request header.
+API Keys are how clients authenticate with the Prefect Cloud API.  They encapsulate the identity of a User or a Service Account.  Ultimately, all clients that interact with the Prefect Cloud API must provide an API Key as a Bearer Token included in the request header.
 
+For more information on how to use these keys to interact with the GraphQL API, go [here](api.html).
```
madkinsz, line 5:
```suggestion
See our [API documentation](api.html) more information on how to use these keys to interact with the GraphQL API.

```

docs/orchestration/concepts/api_keys.md
```diff
@@ -1,20 +1,25 @@
...
 ### User API Keys
 
-User-based API keys function as personal access keys.  When creating an API key, you can configure the following attributes:
+Users can generate API Keys to serve as personal access keys.  When creating an API key, you can configure the following attributes:
```
madkinsz, line 8: Introducing the personal access key here is confusing

```suggestion
Users can generate API Keys to interact with the API with their personal permissions.  When creating an API key, you can configure the following attributes:
```

docs/orchestration/concepts/api_keys.md
```diff
@@ -1,20 +1,25 @@
...
 
-To generate an API key for yourself, navigate to User > API Keys within the UI and click "Create an API Key".
+::: tip Key Management
+Best Practice: When tempted to create a long-lived token for CI, a Prefect Agent, or any use beyond local development, use a Service Account API Key.  Your User API Key is effectively your identity and should be treated as such.
+:::
 
 ### Service Account API Keys
 
-Service Account-based API keys are used for processes like the Prefect Agent or CI, which require the ability to execute or register flows on behalf of a tenant.  To create and manage both your tenant's Service Accounts and Service Account API keys, navigate to Team > Service Accounts.
+Create API Keys restricted to a tenant by creating API Keys associated with Service Accounts in that tenant.  These API Tokens have the same permissions as their Service Account.  To create and manage your tenant's Service Accounts and their associated API keys, navigate to Team > Service Accounts.
```
madkinsz, line 22:
```suggestion
Users can generate API Keys for Service Accounts, which have permissions restricted to the tenant the Service Account belongs to. These keys will not inherit the full permissions of the user creating them. To create and manage your tenant's Service Accounts and their associated API keys, navigate to Team > Service Accounts.  
```

docs/orchestration/concepts/api_keys.md
```diff
@@ -69,21 +75,22 @@ mutation {
 
 ## Querying for Key Information
 
-To query for information about a specific API key with GraphQL, execute the following query against `https://api.prefect.io`. This will allow you to query for API key information, however it never returns the value of the key itself.
+To query for information about API Keys with GraphQL, execute the following query against `https://api.prefect.io`. This will return API Key information, but never the value of the key.  All User API Keys for the active User will be returned, along with any Service Account API Keys which the API Key is privileged to see in the the API Key's default tenant.  Only and all Administrators are privileged to see all Service Accounts--because all Service Accounts have the Administrator role, they can also see all API Keys for any Service Account in the tenant.
```
madkinsz, line 78:
```suggestion
To query for information about API Keys with GraphQL, execute the following query against `https://api.prefect.io`. This will return API Key metadata, but the value of the key cannot be retrieved.
```
This next portion makes sense to me but is _very_ confusing. I'm not sure how to rewrite it.
`All User API Keys for the active user will be returned, along with any Service Account API Keys which the API Key is privileged to see in the the API Key's default tenant.  Only and all Administrators are privileged to see all Service Accounts--because all Service Accounts have the Administrator role, they can also see all API Keys for any Service Account in the tenant.`

docs/orchestration/concepts/api_keys.md
```diff
@@ -106,4 +113,6 @@ $ export PREFECT__CLOUD__AGENT__AUTH_TOKEN=SERVICE_ACCOUNT_API_KEY
 auth_token = SERVICE_ACCOUNT_API_KEY
 ```
 
-For information on the use of your user-scoped API keys visit the [Prefect Cloud API](api.html) page.
+## Deprecation of User Access Tokens and API Tokens
+
+API Keys replace the deprecated User Access Tokens and API Tokens, which previously followed a different paradigm.  In effect, User API Keys can be used in place of Personal Access Tokens, and Service Account API Keys should replace API Tokens (which can no longer be created in the UI), though they function slightly differently.
```
madkinsz, line 118:
```suggestion
API Keys replace the deprecated User Access Tokens and API Tokens, which used a different authentication paradigm. In effect, User API Keys can be used in place of Personal Access Tokens, and Service Account API Keys should replace API Tokens.
```

src/prefect/agent/agent.py
```diff
@@ -163,7 +163,10 @@ def __init__(
 
     def _verify_token(self, token: str) -> None:
         """
-        Checks whether a token with a `RUNNER` scope was provided
+        Checks whether a token with a `RUNNER` scope was provided (DEPRECATED)
+        Visit https://docs.prefect.io/orchestration/concepts/api_keys.html
+        for information on how to use API Keys.
```
madkinsz, line 168: This is a private function and is going to be confusing to a developer because this is still validating the key, right?

src/prefect/cli/auth.py
```diff
@@ -12,6 +13,16 @@ def check_override_auth_token():
         raise Abort
 
 
+def warn_deprecated_api_tokens():
+    warnings.warn(
+        """
+        API Tokens are DEPRECATED.
```
madkinsz, line 19:
```suggestion
        API Tokens have been deprecated in favor of API Keys.
```

src/prefect/cli/auth.py
```diff
@@ -195,19 +208,20 @@ def switch_tenants(id, slug):
 
 @auth.command(hidden=True)
 @click.option("--name", "-n", required=True, help="A token name.", hidden=True)
-@click.option("--scope", "-s", required=True, help="A token scopre.", hidden=True)
-def create_token(name, scope):
+@click.option("--user-id", "-u", required=True, help="a user id.", hidden=True)
+def create_key(name, user_id):
```
madkinsz, line 212: We'll have to maintain the old CLI command with a deprecation warning I think. At the very least, we'll need to keep the old command and throw an error saying what to use instead.

src/prefect/cli/auth.py
```diff
@@ -195,19 +208,20 @@ def switch_tenants(id, slug):
...
     """
-    Create a Prefect Cloud API token.
-
-    For more info on API tokens visit https://docs.prefect.io/orchestration/concepts/api.html
+    Create a Prefect Cloud API Key for a user (must be active user or,
+    if active user is an Administrator, a Service Account in the active tenant).
```
madkinsz, line 215: This active user blurb is confusing to me

src/prefect/cli/auth.py
```diff
@@ -228,46 +242,118 @@ def create_token(name, scope):
 
 
 @auth.command(hidden=True)
-def list_tokens():
```
madkinsz, line 231: We'll have to maintain the old CLI command with a deprecation warning I think. At the very least, we'll need to keep the old command and throw an error saying what to use instead.

src/prefect/cli/auth.py
madkinsz: Ah git diff is bad -- I see this still exists. I'll have to review it outside of GitHub.

src/prefect/cli/auth.py
madkinsz: Ah git diff is bad -- I see this still exists. I'll have to review it outside of GitHub.

src/prefect/client/client.py
```diff
@@ -99,15 +106,21 @@ def __init__(self, api_server: str = None, api_token: str = None):
 
         # store api server
         self.api_server = api_server or prefect.context.config.cloud.get("graphql")
-
-        self._api_token = api_token or prefect.context.config.cloud.get(
-            "auth_token", None
+        if api_token:
+            warnings.warn("api_token DEPRECATED -- use api_key")
+            if api_key:
+                warnings.warn("found both api_token and api_key--using api_key")
```
madkinsz, line 112: Throw an error in this case

src/prefect/client/client.py
```diff
@@ -99,15 +106,21 @@ def __init__(self, api_server: str = None, api_token: str = None):
 
         # store api server
         self.api_server = api_server or prefect.context.config.cloud.get("graphql")
-
-        self._api_token = api_token or prefect.context.config.cloud.get(
-            "auth_token", None
+        if api_token:
+            warnings.warn("api_token DEPRECATED -- use api_key")
```
madkinsz, line 110: See previous comments about warnings

src/prefect/client/client.py
```diff
@@ -203,20 +217,24 @@ def post(
...
+            - api_key (str): an API Key
+            - token (str): an auth token (DEPRECATED)
             - retry_on_api_error (bool): whether the operation should be retried if the API returns
                 an API_ERROR code
 
         Returns:
             - dict: Dictionary representation of the request made
         """
+        if token and not api_key:
+            warnings.warn("token is DEPRECATED--use api_key")
```
madkinsz, line 229: See previous comments about warnings

src/prefect/client/client.py
```diff
@@ -203,20 +217,24 @@ def post(
...
-            - token (str): an auth token. If not supplied, the `client.access_token` is used.
+            - api_key (str): an API Key
+            - token (str): an auth token (DEPRECATED)
             - retry_on_api_error (bool): whether the operation should be retried if the API returns
                 an API_ERROR code
 
         Returns:
             - dict: Dictionary representation of the request made
         """
+        if token and not api_key:
```
madkinsz, line 228: See previous comment about the `if` logic here

src/prefect/client/client.py
```diff
@@ -285,22 +304,26 @@ def graphql(
...
+            - token (str): an auth token. (DEPRECATED--use api_key)
 
         Returns:
             - dict: Data returned from the GraphQL query
 
         Raises:
             - ClientError if there are errors raised by the GraphQL mutation
         """
+        if token and not api_key:
+            warnings.warn("token is DEPRECATED--use api_key")
```
madkinsz, line 319: Same pattern that I've commented on previously.

src/prefect/client/client.py
```diff
@@ -618,7 +637,7 @@ def login_to_tenant(self, tenant_slug: str = None, tenant_id: str = None) -> boo
             },
             variables=dict(slug=tenant_slug, id=tenant_id),
             # use the API token to query the tenant
```
madkinsz, line 639: These comments should be updated (occurs a few times)

src/prefect/client/client.py
```diff
@@ -567,25 +590,21 @@ def get_available_tenants(self) -> List[Dict]:
...
+            token=self._api_key,
         )  # type: ignore
         return result.data.tenant  # type: ignore
 
     def login_to_tenant(self, tenant_slug: str = None, tenant_id: str = None) -> bool:
         """
         Log in to a specific tenant
 
-        NOTE: this should only be called by users who have provided a USER-scoped API token.
-
```
madkinsz, line 588: May be worth noting that this cannot be called if the API Key is from a Service Account

madkinsz (changes requested):
Thanks so much for getting at this!

docs/orchestration/concepts/api_keys.md
TylerWanner: I was avoiding changing that particular language but I agree

src/prefect/client/client.py
TylerWanner: it should work with an SA API Key, it will just only return the SA's tenant

docs/orchestration/concepts/api_keys.md
TylerWanner: Completely agree--but the behaviour is confusing if we don't note SOMETHING here imho.  Leaving this suggestion open

src/prefect/agent/agent.py
TylerWanner: Replaced this with _verify_key which does not check for RUNNER scope obviously, but may need some similar check.  Left _verify_token for backwards compatability but since it's private, I'm not sure we have to

src/prefect/cli/auth.py
TylerWanner: we use "active tenant" as vocabulary in client settings so I went with this "active user" terminology, but open to suggestions here for sure.  This docstring makes very little sense, even though it's technically accurate.  Will try to come back to it

src/prefect/agent/kubernetes/agent.py
```diff
@@ -766,7 +769,13 @@ def generate_deployment_yaml(
         """
 
         # Use defaults if not provided
-        token = token or ""
+        if token:
```
cicdw, line 772: These changes seem unnecessary and overly risky -- these are all internal methods that still work correctly regardless of API Key or Token distinction

src/prefect/cli/agent.py
```diff
@@ -1,11 +1,18 @@
...
 from prefect.utilities.configuration import set_temporary_config
 from prefect.utilities.serialization import from_qualified_name
 from prefect.utilities.cli import add_options
 
 COMMON_START_OPTIONS = [
+    click.option(
+        "--api-key",
+        "-k",
+        required=False,
+        help="A Prefect Cloud API Key with RUNNER scope. (DEPRECATED)",
```
cicdw, line 14: This help string is mixing the two concepts together in an inaccurate way; an API key should be provided here, there is no concept of "RUNNER" scoping, and API keys are not deprecated as this string suggests

cicdw (changes requested):
The docs changes are great; I'm less confident in all the code changes as many of the changes are for private methods that we don't need to take the risk with right now.  

I think we should break this up into stages:
- docs
- public APIs 
- internal APIs

src/prefect/agent/kubernetes/agent.py
TylerWanner: > The docs changes are great; I'm less confident in all the code changes as many of the changes are for private methods that we don't need to take the risk with right now.
> 
> I think we should break this up into stages:
> 
> * docs
> * public APIs
> * internal APIs

Cool--I can do that.  Will push a PR for just docs.

PR #4453, opened 2021-04-28
clarify docs on API Keys

TylerWanner:
## Summary

Update docs to reflect use of API Keys in place of tokens
Docs changes from https://github.com/PrefectHQ/prefect/pull/4449

## Changes

clarify use of API Keys
Remove reference to prefect auth revoke-token

## Importance

Clear up usage of API Keys

## Checklist

This PR:


docs/orchestration/concepts/api_keys.md
```diff
@@ -1,20 +1,32 @@
...
-- **API Key Name**: The name of this key; this is useful for organizational and bookkeeping purposes
+- **API Key Name**: The name of this key
 - **API Key Expiration**: An optional expiration date for the key - if no expiration is provided, the key will never expire
-- **Tenant**: Which tenant to associate the key with - this key will then have all of the permissions that you have as a member of that tenant.  For example, suppose you are a member of Tenant A as a Tenant Administrator, and Tenant B as a Read-Only User.  When you create an API key associated with Tenant A, it will have the ability to perform actions at the Tenant Admin level.  On the other hand, when you create an API key associated with Tenant B, it will only be able to perform queries and no mutations.
+- **Tenant**: The tenant to associate with the API Key - clients using this key will interact with this tenant.  This key's 
+permissions are the user's in that tenant.
+
+::: warning Mutable Tenancy
+
+Clients using this token can still list the User's available tenants and can modify the token's tenant association.
```
cicdw, line 18: There isn't currently an exposed route for this so I think we should leave this warning out until we have hardened the recommended pattern here.

docs/orchestration/concepts/api_keys.md
```diff
@@ -1,20 +1,32 @@
...
 
-User-based API keys function as personal access keys.  When creating an API key, you can configure the following attributes:
+Users can generate API Keys to interact with the API with their personal permissions.  When creating an API key, you can configure the following attributes:
 
-- **API Key Name**: The name of this key; this is useful for organizational and bookkeeping purposes
+- **API Key Name**: The name of this key
 - **API Key Expiration**: An optional expiration date for the key - if no expiration is provided, the key will never expire
-- **Tenant**: Which tenant to associate the key with - this key will then have all of the permissions that you have as a member of that tenant.  For example, suppose you are a member of Tenant A as a Tenant Administrator, and Tenant B as a Read-Only User.  When you create an API key associated with Tenant A, it will have the ability to perform actions at the Tenant Admin level.  On the other hand, when you create an API key associated with Tenant B, it will only be able to perform queries and no mutations.
+- **Tenant**: The tenant to associate with the API Key - clients using this key will interact with this tenant.  This key's
```
cicdw, line 13: suggestion for stronger wording:
```suggestion
- **Tenant**: The tenant to associate with the API Key - clients using this key can interact with this tenant only.  This key's 
```

cicdw (changes requested):
Awesome this looks solid - I left two minor suggestions, otherwise LGTM

docs/orchestration/concepts/api_keys.md
TylerWanner: I've removed the warning.  Thanks for the feedback.

FYI I seem to be able to change the associated tenant using the client via an existing method that was meant for User-scoped API tokens
`client.login_from_tenant(tenant_slug="foo")`, which uses the `switch_tenant` mutation
Is that how it should work?  I can expose a new CLI command for that in the followup, if desired

cicdw: Ah great question; that `switch_tenant` mutation actually uses the refresh / short-lived token pattern.  What I mean is it uses the current token / key to generate _new_ tokens that are scoped to the other tenant.  The underlying API key that you set on the Client isn't mutated at all.

TylerWanner: weird, even after removing client settings and auth_token from config.toml, clients started with that token default to the switched tenant, not the original.  Was trying to understand how that is persisted, because it all appears to be specific to the old paradigm, which I am not as familiar with.  

I'll leave all the login/ tenant switching stuff alone and stick to create, revoke, and list keys in the followup.

cicdw: Yea sounds good, we can also do some additional QA when we move to the code changes.

cicdw (approved):

PR #4457, opened 2021-04-29
feat(email_task): sending attachments with emails

thomasheyenbrock:
## Summary

This PR adds a new option to the `EmailTask`, allowing to send attachments with the email.

## Changes

The email task now always creates a multipart-message. The user of the task can optionally pass a list of files that should be sent as attachment with the email. The files are looked up in the local filesystem and attached to the multipart-message.

## Importance

Sending attachments with emails is a very common thing to do. A concrete use-case would be to extract and transform some data and then send the resulting data (as csv, Excel, etc.) via email.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/notifications/email_task.py
+++ b/src/prefect/tasks/notifications/email_task.py
@@ -83,6 +83,7 @@ class EmailTask(Task):
         "msg_plain",
         "email_to_cc",
         "email_to_bcc",
+        "attachments",
     )
     def run(
         self,
@@ -96,7 +97,7 @@ class EmailTask(Task):
         msg_plain: str = None,
         email_to_cc: str = None,
         email_to_bcc: str = None,
-        attachments: List[str] = [],
+        attachments: List[str] = None,
     ) -> None:
         """
         Run method which sends an email.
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/notifications/email_task.py
```diff
@@ -50,6 +56,7 @@ def __init__(
         msg_plain: str = None,
         email_to_cc: str = None,
         email_to_bcc: str = None,
+        attachments: List[str] = [],
```
zangell44, line 59: We want to avoid using any [mutable default arguments](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). I would default this to `None`. Then, if nothing is passed, set `self.attachments` to an empty list

src/prefect/tasks/notifications/email_task.py
```diff
@@ -88,6 +96,7 @@ def run(
...
```
zangell44, line 99: This should default to `None`. And we should add "attachments" to the `@defaults_from_attrs` decorator above.

zangell44 (changes requested):
Thanks for contributing to Prefect @thomasheyenbrock!
Looks good overall, a few minor feedback points

thomasheyenbrock:
@zangell44 thanks for the good feedback 🙂

PR #4458, opened 2021-04-29
Print link to flow run in StartFlowRun

bnaul:
The link artifact produced by StartFlowRun is very handy for navigating from parent to child flow, but a log message (like what `prefect run flow` produces) would still be nice: you can see it right away when following the logs from the command line, and when the parent task is running outside cloud there's no artifact at all. cc @madkinsz @cicdw

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -184,11 +184,11 @@ class StartFlowRun(Task):
             scheduled_start_time=scheduled_start_time,
         )
 
-        self.logger.debug(f"Flow Run {flow_run_id} created.")
 
         self.logger.debug(f"Creating link artifact for Flow Run {flow_run_id}.")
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -184,11 +184,10 @@ def run(
             scheduled_start_time=scheduled_start_time,
         )
 
-        self.logger.debug(f"Flow Run {flow_run_id} created.")
-
```
madkinsz, line 188: I don't think we should remove this line

madkinsz (commented):
This makes sense to me! Just the one requested change. Did you want to add yourself as a contributor and add a changelog entry? (not required for something small like this, but feel free)

bnaul:
Thanks @madkinsz, updated

madkinsz (approved):

PR #4462, opened 2021-04-29
Add `image_pull_policy` option to `KubernetesRun` config

zschumacher:
## Summary

Adds an `image_pull_policy` to `KubernetesRun` config object

## Changes

- Adds an `image_pull_policy` kwarg to `KubernetesRun` config object
- Adds imagePullPolicy to job template if `image_pull_policy` is `Always`, `Never` or `IfNotPresent`

## Importance

This is a common gotcha for Kubernetes jobs, and is a result of the discussion on #4445 

Closes #4445 

## Checklist

This PR:

### DIFF
--- a/docs/orchestration/flow_config/run_configs.md
+++ b/docs/orchestration/flow_config/run_configs.md
@@ -194,7 +194,7 @@ for the Kubernetes job:
 
 
 ```python
-flow.run_config = KubernetesRun(imagePullPolicy="Always")
+flow.run_config = KubernetesRun(image_pull_policy="Always")
 ````
 
 ### ECSRun
--- a/docs/orchestration/flow_config/run_configs.md
+++ b/docs/orchestration/flow_config/run_configs.md
@@ -189,12 +189,12 @@ Use a custom Kubernetes Job spec for this flow, stored in S3:
 flow.run_config = KubernetesRun(job_template_path="s3://bucket/path/to/spec.yaml")
 ```
 
-Always pull the newest image for the job run.  This should be used when using a custom docker image 
-with a tag other than `:latest`.  Doing so will set the kubernetes [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images)
-to `Always` on the job run.
+Specify an [imagePullPolicy](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) 
+for the Kubernetes job:
+
 
 ```python
-flow.run_config = KubernetesRun(always_pull_latest_image=True)
+flow.run_config = KubernetesRun(imagePullPolicy="Always")
 ````
 
 ### ECSRun
--- a/docs/orchestration/flow_config/run_configs.md
+++ b/docs/orchestration/flow_config/run_configs.md
@@ -189,6 +189,14 @@ Use a custom Kubernetes Job spec for this flow, stored in S3:
 flow.run_config = KubernetesRun(job_template_path="s3://bucket/path/to/spec.yaml")
 ```
 
+Always pull the newest image for the job run.  This should be used when using a custom docker image 
+with a tag other than `:latest`.  Doing so will set the kubernetes [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images)
+to `Always` on the job run.
+
+```python
+flow.run_config = KubernetesRun(always_pull_latest_image=True)
+````
+
 ### ECSRun
 
 [ECSRun](/api/latest/run_configs.md#ecsrun) configures flow runs
--- a/src/prefect/run_configs/kubernetes.py
+++ b/src/prefect/run_configs/kubernetes.py
@@ -127,7 +127,8 @@ class KubernetesRun(RunConfig):
             and image_pull_policy not in image_pull_policies
         ):
             raise ValueError(
-                f"Invalid image_pull_policy {image_pull_policy}.  Expected `Always`, `IfNotPresent`, or `Never`"
+                f"Invalid image_pull_policy {image_pull_policy}.  "
+                "Expected `Always`, `IfNotPresent`, or `Never`"
             )
 
         self.job_template_path = job_template_path
--- a/src/prefect/run_configs/kubernetes.py
+++ b/src/prefect/run_configs/kubernetes.py
@@ -39,10 +39,8 @@ class KubernetesRun(RunConfig):
         - labels (Iterable[str], optional): an iterable of labels to apply to this
             run config. Labels are string identifiers used by Prefect Agents
             for selecting valid flow runs when polling for work
-        - always_pull_new_image (bool): a boolean flag for telling kubernetes to
-            always pull new images.  This is useful when you are using a tag on
-            and image other than :latest
-            https://kubernetes.io/docs/concepts/containers/images/#updating-images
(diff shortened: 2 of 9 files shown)


slightstone (dismissed):

src/prefect/serialization/run_config.py
```diff
@@ -32,6 +32,7 @@ class Meta:
     memory_request = fields.String(allow_none=True)
     service_account_name = fields.String(allow_none=True)
     image_pull_secrets = fields.List(fields.String(), allow_none=True)
+    always_pull_new_image = fields.String(allow_none=False)
```
zangell44, line 35: This should be `fields.Boolean`, I think this is why "False == False" wasn't working in your test case

tests/serialization/test_run_configs.py
```diff
@@ -62,9 +63,14 @@ def test_serialize_kubernetes_run(config):
         "memory_request",
         "service_account_name",
         "image_pull_secrets",
+        "always_pull_new_image",
     ]
     for field in fields:
-        assert getattr(config, field) == getattr(config2, field)
```
zangell44, line 67: I think the serialization field type is causing issues here, we should not have to change the test logic here

tests/agent/test_k8s_agent.py
```diff
@@ -1262,10 +1262,19 @@ def test_generate_job_spec_uses_job_template_provided_in_run_config(self):
...
         labels["TEST"] = "VALUE"
-        flow_run = self.build_flow_run(KubernetesRun(job_template=template))
+        config = KubernetesRun(job_template=template, always_pull_new_image=True)
+
+        flow_run = self.build_flow_run(config)
         job = self.agent.generate_job_spec(flow_run)
         assert job["metadata"]["labels"]["TEST"] == "VALUE"
 
+        # test always pull new image kwarg works
+        print(type(job))
```
zangell44, line 1272: We can take this debugging statement out

tests/agent/test_k8s_agent.py
```diff
@@ -1262,10 +1262,19 @@ def test_generate_job_spec_uses_job_template_provided_in_run_config(self):
         template = self.read_default_template()
         labels = template.setdefault("metadata", {}).setdefault("labels", {})
         labels["TEST"] = "VALUE"
-        flow_run = self.build_flow_run(KubernetesRun(job_template=template))
```
zangell44, line 1265: I think this belongs in a separate test case, something like

```python
@pytest.mark.parameterize(
    ('always_pull', 'expected_pull_policy')
    [
        (True, 'Always'),
        (False, '<default pull policy here>')
    ]
def test_generate_job_spec_sets_image_pull_policy_from_run_config(self, always_pull, expected_pull_policy):
    template = self.read_default_template()
    config = KubernetesRun(job_template=template, always_pull_new_image=always_pull)
    flow_run = self.build_flow_run(config)
    job = self.agent.generate_job_spec(flow_run)
    assert (
            job["spec"]["template"]["spec"]["containers"][0]["imagePullPolicy"]
            == expected_pull_policy
        )
```

src/prefect/run_configs/kubernetes.py
```diff
@@ -38,6 +39,10 @@ class KubernetesRun(RunConfig):
         - labels (Iterable[str], optional): an iterable of labels to apply to this
             run config. Labels are string identifiers used by Prefect Agents
             for selecting valid flow runs when polling for work
+        - always_pull_new_image (bool): a boolean flag for telling kubernetes to
```
zangell44, line 42: nit: we should add "Defaults to False" somewhere in the docstring

src/prefect/serialization/__init__.py
```diff
@@ -6,3 +6,4 @@
 import prefect.serialization.state
 import prefect.serialization.environment
 import prefect.serialization.storage
+import prefect.serialization.run_config
```
zangell44, line 9: do we need to add this import?

docs/orchestration/flow_config/run_configs.md
```diff
@@ -189,6 +189,14 @@ Use a custom Kubernetes Job spec for this flow, stored in S3:
 flow.run_config = KubernetesRun(job_template_path="s3://bucket/path/to/spec.yaml")
 ```
 
+Always pull the newest image for the job run.  This should be used when using a custom docker image 
+with a tag other than `:latest`.  Doing so will set the kubernetes [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images)
+to `Always` on the job run.
+
+```python
+flow.run_config = KubernetesRun(always_pull_latest_image=True)
```
zangell44, line 197: The argument in the code is `always_pull_new_image`. I actually like `always_pull_latest_image` better. I would vote to update to `always_pull_latest_image` everywhere.

zangell44 (changes requested):
Thanks for putting this together @zschumacher, this will be super valuable! I left a few comments with requests for changes

docs/orchestration/flow_config/run_configs.md
jcrist: Why not just expose `image_pull_policy` directly? Having a nonstandard kwarg will confuse those that are used to k8s, and we already make use of k8s kwarg names elsewhere.

zangell44: yeah fair point I like that better, +1 for exposing `image_pull_policy` directly

zschumacher: yep, makes sense to me.  Do you have a preference on how that is exposed based on the rest of the repo? e.g.

1. type hint `Literal[Always, w/e]`
2. enum
3. just a string

jcrist: Just a string makes sense to me. We can check in the `KubernetesRun` constructor that the value is valid.

zangell44 (dismissed):
This looks great! Thanks @zschumacher!

src/prefect/run_configs/kubernetes.py
```diff
@@ -108,6 +121,16 @@ def __init__(
         if image_pull_secrets is not None:
             image_pull_secrets = list(image_pull_secrets)
 
+        image_pull_policies = {"Always", "IfNotPresent", "Never"}
+        if (
+            image_pull_policy is not None
+            and image_pull_policy not in image_pull_policies
+        ):
+            raise ValueError(
+                f"Invalid image_pull_policy {image_pull_policy}.  "
```
madkinsz, line 130:
```suggestion
                f"Invalid image_pull_policy {image_pull_policy!r}.  "
```

For displaying string values like this, it's nice to get the repr which will put quotes so

> Invalid image_pull_policy Foo bar

vs

> Invalid image_pull_policy 'Foo bar'

src/prefect/run_configs/kubernetes.py
```diff
@@ -108,6 +121,16 @@ def __init__(
...
+                "Expected `Always`, `IfNotPresent`, or `Never`"
```
madkinsz, line 131:
```suggestion
                "Expected 'Always', 'IfNotPresent', or 'Never'"
```
We expect string values whereas ` implies an object to me

madkinsz (changes requested):
This looks great! Just a really minor change on the error message

PR #4468, opened 2021-04-30
Allow `prefect build/register --module` to accept full import path to flow

madkinsz:
## Summary

Currently `prefect register --module my_module` does not allow you to pass a full import path to a single flow. This PR modifies the logic slightly to allow `prefect register --module my_module.flow` to import a single flow from a module.

## Changes

- Allows a full import path to a flow within a module to be passed
- Fixes bug in `import_object` when passed a module with no `.`

## Importance

This allows me to:
- Create an example like `prefect register --project default --module prefect.hello_world`
- Disambiguate flows by full path rather than `--module` and `--name` pairing (which is still valid)
- Run a single flow given an import path e.g. `prefect run --module prefect.hello_world.flow --watch` (I use these functions in #4463)

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/importtools.py
+++ b/src/prefect/utilities/importtools.py
@@ -25,4 +25,7 @@ def import_object(name: str) -> Any:
         mod = importlib.import_module(mod_name)
         return getattr(mod, attr_name)
     except ValueError:
-        return importlib.import_module(name)
+        pass
+
+    # When we can't split the string we'll just import the module
+    return importlib.import_module(name)
--- a/changes/pr4468.yaml
+++ b/changes/pr4468.yaml
@@ -1,0 +1,3 @@
+
+enhancement:
+  - "Allow `prefect build/register --module` to accept full import path to flow - [#4468](https://github.com/PrefectHQ/prefect/pull/4468)"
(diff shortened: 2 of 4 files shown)


src/prefect/utilities/importtools.py
```diff
@@ -25,4 +25,7 @@ def import_object(name: str) -> Any:
         mod = importlib.import_module(mod_name)
         return getattr(mod, attr_name)
     except ValueError:
-        return importlib.import_module(name)
+        pass
```
madkinsz, line 28: This raised an error instead of returning as intended 🤦

zangell44: maybe not worth doing in this PR, but a little refactor of this code to make it more clear what we're doing would be nice

it's not clear at first glance that `name.rsplit(".",1)` is the line expected to throw a `ValueError`

zangell44 (approved):

src/prefect/utilities/importtools.py
madkinsz: Yeah I was confused as well.. I was hoping my comment was helpful. Pass this one to the fuuuture

PR #4485, opened 2021-05-06
Set flow docker container names to flow run names

madkinsz:
## Summary

Currently, `DockerAgent` container names are UUIDs (the default). It'd be nice to set the container names to the flow run name. However, container names must be unique so we must check for collisions to avoid blocking flow run execution. It seems like a rare case that there will be a collision (since we auto-remove containers) so we add a simple index `-{i}` to the end of the name if we encounter one.

## Changes

- Sets the container name during creation by a `DockerAgent`

## Importance

Improves usability / interaction with flow run containers.

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -442,15 +442,28 @@ class DockerAgent(Agent):
             "io.prefect.flow-id": flow_run.flow.id,
             "io.prefect.flow-run-id": flow_run.id,
         }
+
+        # Generate a container name to match the flow run name, ensuring it is unique
+        container_name = flow_run.name
+        containers = self.docker_client.containers()
+        existing_names = {c["name"] for c in containers}
+        index = 0
+        while container_name in existing_names:
+            index += 1
+            container_name = f"{flow_run.name}-{index}"
+
+        # Create the container
         container = self.docker_client.create_container(
             image,
             command=get_flow_run_command(flow_run),
             environment=env_vars,
+            name=container_name,
             volumes=container_mount_paths,
             host_config=self.docker_client.create_host_config(**host_config),
             networking_config=networking_config,
             labels=labels,
         )
+
         # Connect the rest of the networks
         if self.networks:
             for network in self.networks[1:]:
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -1,11 +1,13 @@
-import multiprocessing
 import ntpath
 import posixpath
+
+import multiprocessing
 import re
 import sys
+import warnings
+from slugify import slugify
 from sys import platform
 from typing import TYPE_CHECKING, Dict, Iterable, List, Tuple
-import warnings
 
 from prefect import config, context
 from prefect.agent import Agent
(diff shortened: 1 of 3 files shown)


src/prefect/agent/docker/agent.py
```diff
@@ -442,15 +442,28 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
         }
+
+        # Generate a container name to match the flow run name, ensuring it is unique
+        container_name = flow_run.name
+        containers = self.docker_client.containers()
+        existing_names = {c["name"] for c in containers}
+        index = 0
+        while container_name in existing_names:
+            index += 1
+            container_name = f"{flow_run.name}-{index}"
```
jcrist, line 453: While this seems fine to me, a different way would be to try to create the container, and increment the name on error. This avoids the need to query the containers beforehand (provided the error returned on collision is easily identifiable).

Also, what restrictions are there on container names? Will the flow run name always be valid?

jcrist (commented):
The intent here seems fine to me, just a couple questions on implementation.

src/prefect/agent/docker/agent.py
madkinsz: Re a try/except pattern: That's okay with me but then I need to catch and parse an exception and listing containers seems relatively fast. I'm not sure optimizing for that amount of latency is going to make a big difference.

Great point on valid container names. Our flow names can be any string value but it looks like their are limitations on the docker end.
https://stackoverflow.com/questions/42642561/docker-restrictions-regarding-naming-container/42642992

Should we just fall back to a UUID if the regex fails and write a little warning log?

jcrist: > That's okay with me but then I need to catch and parse an exception

I was hoping it would be a specific exception type. If string matching on the exception is required then it's not worth it.

> Should we just fall back to a UUID if the regex fails and write a little warning log?

Perhaps we should slugify and trim the name always? Makes it easier to match container to flow run, but doesn't have to be a 1:1 match.

madkinsz: > I was hoping it would be a specific exception type. If string matching on the exception is required then it's not worth it.
I think it'd just be a `APIError` -- but I can check by trying

> Perhaps we should slugify and trim the name always? Makes it easier to match container to flow run, but doesn't have to be a 1:1 match.
I'm down if it's easy!

madkinsz:
@jcrist / @zach-angell

This turned into kind of a wild validation. Hopefully it's easy to follow. Let me know what you think.

src/prefect/agent/docker/agent.py
```diff
@@ -442,15 +444,49 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
+                    # limit the length for safety
+                    max_length=250,
+                    # Docker allows these characters for container names
+                    regex_pattern=r"[^a-zA-Z0-9_.-]+",
+                ),
+            )
+            # Docker does not allow 0 character names so use the flow run id if name
+            # would be empty
+            or flow_run.id
+        )
```
jcrist, line 469: This re.sub call could be replaced by a `str.lstrip` call.

```python
slugify(...).lstrip("_.-")
```

jcrist (commented):
Overall looks good to me.

src/prefect/agent/docker/agent.py
madkinsz: Ah I considered `lstrip` but thought it took a ordered leading string to remove. I think I'll change that.

jcrist (approved):

PR #4486, opened 2021-05-06
feat(KubernetesAgent): provide path to custom kube config

joelluijmes:
## Summary

This PR adds feature to Kubernetes Agent to use a custom Kubernetes cluster to spawn / monitor jobs. It does so by allowing an additional parameter `kube_config_file` to specify a path to kube config instead of assuming the default. Alternatively, it is possible to supply this with the `KUBE_CONFIG_FILE` environment variable.

## Changes

Add parameter `kube_config_file` to specify a path to kube config instead of assuming the default. Alternatively, it is possible to supply this with the `KUBE_CONFIG_FILE` environment variable.

## Importance

This allows to run a Kubernetes agent in cluster `foo`, while running actual jobs in cluster `bar`. Making the agent more flexible :) 

## Checklist

This PR:

### DIFF
--- a/changes/pr4486.yaml
+++ b/changes/pr4486.yaml
@@ -1,0 +1,5 @@
+feature:
+  - "KubernetesAgent: allow to customize cluster by providing path to kube config - [#4486](https://github.com/PrefectHQ/prefect/pull/4486)"
+
+contributor:
+  - "[Joël Luijmes](https://github.com/joelluijmes)"
--- a/docs/orchestration/agents/kubernetes.md
+++ b/docs/orchestration/agents/kubernetes.md
@@ -155,7 +155,18 @@ prefect agent kubernetes start --job-template /path/to/my_template.yaml
 prefect agent kubernetes start --job-template s3://bucket/path/to/my_template.yaml
 ```
 
-## Running In-Cluster
+### Custom Kube Config
+
+By default the Kubernetes Jobs created by the agent are created in the cluster
+where the agent is running. It is also possible to customize the cluster the
+agent is using by supplying a custom kube config file with the `--kube-config-file`
+flag:
+
+```bash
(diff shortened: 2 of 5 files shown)


joelluijmes:
Note: I haven't actually tested this yet, and am unsure how to proceed to do so. My guess is too create docker image with prefect from this rev?

joelluijmes:
Okay, managed to test this:

1. Created docker image
2. Added secret for kubeconfig in cluster `foo`
3. Modified agent deployment yaml to supply path as environment variable
4. Deployed the agent in cluster `foo`
5. Created a flow
6. Boom job is created in cluster `bar` 🚀 
7. Error during flow run because it doesn't resolve to right apollo, but I suspect its just another config I have to supply in my job_template (he said while diving in...)

madkinsz:
This LGTM! I think we may want to give it a test on our cluster as well. @TylerWanner ? I feel like you've expressed interest in something like this.

madkinsz:
Hey @joelluijmes -- we're hesitant to add an explicit option and environment variable when Kubernetes already provides the `KUBECONFIG` variable to determine the file path. I think this PR could be simplified to include a kwarg `load_incluster_config: bool = True` that can be toggled to skip the incluster load. Thoughts?

joelluijmes:
Aha.. I went through the code of kubernetes lib but I mistook `KUBE_CONFIG_DEFAULT_LOCATION` for a constant. But yes agreed, this PR can be easily simplified then. 
I hope to do this later this week if I have some time, but feel free to take over ;)

madkinsz:
@joelluijmes want to take a swing at this still? Otherwise I'm doing some cleanup and will close

joelluijmes:
Its okay to close this for now 👍 

Currently don't have time to properly finish this. For where I need this I'm currently running my own build.

PR #4487, opened 2021-05-06
Replace `logging.log_to_cloud` with `cloud.send_flow_run_logs`

madkinsz:
## Summary

Previously, `logging.log_to_cloud` was set in the config with a default value of `false` and it was the agent's job to toggle this to `true` via environment variable when a flow run was executed. Since this is not applicable unless using cloud, this key has been moved to `cloud.send_flow_run_logs` for clarity. Additionally, the default value has been changed to `true` and flow run execution will respect the config value unless the flag has been passed to the agent. This allows the user to toggle the setting globally via environment variable or config.toml change or per agent with the existing `--no-cloud-logs` flag.

## Changes

- Moves `logging.log_to_cloud` to `cloud.send_flow_run_logs`
- Changes default value for logging to cloud to `True` -- this does not affect current behavior since the agent defaults to `True`
- Updates the log manager to check for `context.running_with_backend` before emitting logs so that logs are not sent to cloud during local operation i.e. `flow.run()` 

## Importance

Clarifies the option for logging to cloud and makes a user-facing setting actually user togglable.

## Checklist

This PR:


docs/api/0.11.5/environments/execution.md
```diff
@@ -17,7 +17,7 @@ If pulling from a private docker registry, `setup` will ensure the appropriate k
...
 
-- `PREFECT__CLOUD__GRAPHQL` - `PREFECT__CLOUD__AUTH_TOKEN` - `PREFECT__CONTEXT__FLOW_RUN_ID` - `PREFECT__CONTEXT__NAMESPACE` - `PREFECT__CONTEXT__IMAGE` - `PREFECT__CONTEXT__FLOW_FILE_PATH` - `PREFECT__CLOUD__USE_LOCAL_SECRETS` - `PREFECT__ENGINE__FLOW_RUNNER__DEFAULT_CLASS` - `PREFECT__ENGINE__TASK_RUNNER__DEFAULT_CLASS` - `PREFECT__ENGINE__EXECUTOR__DEFAULT_CLASS` - `PREFECT__LOGGING__LOG_TO_CLOUD` - `PREFECT__LOGGING__EXTRA_LOGGERS`
+- `PREFECT__CLOUD__GRAPHQL` - `PREFECT__CLOUD__AUTH_TOKEN` - `PREFECT__CONTEXT__FLOW_RUN_ID` - `PREFECT__CONTEXT__NAMESPACE` - `PREFECT__CONTEXT__IMAGE` - `PREFECT__CONTEXT__FLOW_FILE_PATH` - `PREFECT__CLOUD__USE_LOCAL_SECRETS` - `PREFECT__ENGINE__FLOW_RUNNER__DEFAULT_CLASS` - `PREFECT__ENGINE__TASK_RUNNER__DEFAULT_CLASS` - `PREFECT__ENGINE__EXECUTOR__DEFAULT_CLASS` - `PREFECT__CLOUD__SEND_FLOW_RUN_LOGS` - `PREFECT__LOGGING__EXTRA_LOGGERS`
```
cicdw, line 20: we prob shouldn't update old archived docs

cicdw (changes requested):
We need to be super careful with this change - it could result in new agents submitting old runs that then will not ship their logs to Cloud.  There needs to be some backwards compatibility considerations / respecting the old key logic somewhere

docs/api/0.11.5/environments/execution.md
madkinsz: Oops! Will undo those changes.

madkinsz:
As Chris pointed out: If a containerized run using Prefect before this change is used, it will not log to cloud because it will not be toggled to `True`. I've pushed a couple commits to those agents / environments / specs to pass a redundant compatibility variable in.

zangell44:
Is `cloud.send_flow_run_logs` only scoped to flow run logs? I'm not sure if we send or plan to send other types of logs

jcrist:
What about `cloud.logging_enabled` instead? This flag is used for enabling/disabling the cloud log handler, so it's generic across any log source we use.

jcrist:
Ah, you're right. I could see a future where we change this though (where agent logs are also sent to cloud), so a more generic-sounding flag would be my preference. Happy either way though.

zangell44 (dismissed):

madkinsz:
> Ah, you're right. I could see a future where we change this though (where agent logs are also sent to cloud), so a more generic-sounding flag would be my preference. Happy either way though.

Hm. I could see that world too. I can imagine wanting agent logs (which are just useful for tracking flow deployment issues) but not flow run logs (which have my personal data in them). I like that the key name I have makes it immediately clear that the logs that are sent to cloud are for your flow runs which is nice because a generic key makes me feel more suspicious as I wonder "What data are they logging?" 

cc @cicdw

Did you guys take a reasonably close look at my backwards compatibility variables?

changes/pr4487.yaml
```diff
@@ -0,0 +1,3 @@
+
+enhancement:
+  - "Replace `logging.log_to_cloud` with `cloud.send_flow_run_logs` and allow user to toggle in config - [#4487](https://github.com/PrefectHQ/prefect/pull/4487)"
```
cicdw, line 3: Isn't this a breaking change for anyone who has set this config?

src/prefect/utilities/logging.py
```diff
@@ -139,7 +139,11 @@ class CloudHandler(logging.Handler):
     def emit(self, record: logging.LogRecord) -> None:  # type: ignore
         """Emit a new log"""
         # if we shouldn't log to cloud, don't emit
-        if not context.config.logging.log_to_cloud:
+        if not context.config.cloud.send_flow_run_logs:
+            return
+
+        # if its not during a backend flow run, don't emit
+        if not context.get("running_with_backend"):
```
cicdw, line 146: Because you are also relying on this context variable, you are going to miss at least two logs that I can see, and any errors that occur prior to this being set

cicdw (changes requested):
FWIW I still don't think the risk / reward tradeoff for this change is worth it, but independently of that there are still some edge cases to think through if you'd like to keep going with this

changes/pr4487.yaml
madkinsz: No, because the existing config value (confusingly) is not usable by the user. It's `False` by default and if set to `True` it will break `flow.run()` calls. It's only set to `True` in the _environment_ of flow execution by us. This is the major impetus for this change.

src/prefect/utilities/logging.py
madkinsz: Could you point out some of the logs you're talking about so I can assess the best place to address them? I think the one line change at https://github.com/PrefectHQ/prefect/blob/cd115113b6962a7107b92fcd692b2131191feba1/src/prefect/cli/execute.py#L56-L57 should cover most everything.

changes/pr4487.yaml
madkinsz: As far as I could tell this was only togglable on the agent and https://github.com/PrefectHQ/prefect/blob/log-to-cloud/src/prefect/agent/agent.py#L146-L149 respects the existing CLI flag as well as the config now

cicdw: Ah I see, yea I thought the agent still referenced the config value but I see that it doesn't 👍

src/prefect/utilities/logging.py
cicdw: Yea sorry about that:
- https://github.com/PrefectHQ/prefect/blob/master/src/prefect/engine/flow_runner.py#L245
- https://github.com/PrefectHQ/prefect/blob/master/src/prefect/engine/flow_runner.py#L261

I didn't realize this value was set so early on (I thought it was only applied once the flow runner began), so maybe these logs are safe

madkinsz:
I've added a deprecation warning if the user has set this key manually as well as a backwards compatible check in the logger so if they've set this to `False` explicitly in their cloud runs it should still prevent sending their logs to us.

zangell44 (approved):

zangell44:
Will this setting play nicely with additional loggers?

madkinsz:
> Will this setting play nicely with additional loggers?

@zangell44 in any particular way? I didn't change any behavior for additional loggers.

zangell44:
> @zangell44 in any particular way? I didn't change any behavior for additional loggers.
I'm not sure how additional loggers work or how we plan to build them out, so this is more of a curiosity.

One situation that would seem reasonable to support is not sending logs to Prefect, but sending them using additional loggers. I could see the name `cloud.send_flow_run_logs` and functionality being confusing if additional loggers are added

madkinsz:
> > @zangell44 in any particular way? I didn't change any behavior for additional loggers.
> 
> I'm not sure how additional loggers work or how we plan to build them out, so this is more of a curiosity.
> 
> One situation that would seem reasonable to support is not sending logs to Prefect, but sending them using additional loggers. I could see the name `cloud.send_flow_run_logs` and functionality being confusing if additional loggers are added

Yeah I think the way to go is something like #4550 which would allow for very granular control of logging handlers. However, this config variable would still just be a global toggle which I think makes a lot of sense. 🤷 I'm happy to change the name but as mentioned before I don't really know of logs other than agent logs that we'd want to send and having granulated toggles that case doesn't seem to complex e.g.
- `cloud.send_flow_run_logs`
- `cloud.send_agent_logs`

PR #4499, opened 2021-05-08
[DRAFT] key value store

zangell44:
## Summary

Draft PR for feedback on key value store.

## Checklist

This PR:

## Open questions for draft
- When listing keys, should we support some sort of prefixing for search?
- Where should the `kv` functions live? Moving them to the new `prefect.backend` has been proposed.
- Related: should the client include kv methods at all?
- How should local development work for kv store? Should we implement something like secrets where we have `use_local_kv = true/false` and read/write from json file? Should we only support Cloud key values until getting further user feedback?

### DIFF
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -13,6 +13,7 @@ import prefect.environments.storage  # deprecated
 import prefect.executors
 import prefect.engine.executors  # deprecated
 import prefect.artifacts
+import prefect.utilities.key_value as kv
 
 from prefect.core import Task, Flow, Parameter
 import prefect.engine
--- a/src/prefect/cli/key_value.py
+++ b/src/prefect/cli/key_value.py
@@ -1,0 +1,109 @@
+import click
+
+from prefect import Client, config
+
+
+@click.group(hidden=True)
+def kv():
+    """
+    Handle Prefect Cloud authorization.
+
+    \b
+    Usage:
+        $ prefect kv [COMMAND]
+
+    \b
+    Arguments:
+        set             Set a key value pair
+        get             Get the value associated with a key
+        delete          Delete a key value pair
+        list            List keys
+
+    \b
+    Examples:
+        $ prefect kv set foo bar
+
+    \b
+        $ prefect kv get my_key
+        bar
+
+    \b
+        $ prefect kv list
+        foo
+
+    \b
+        $ prefect kv delete foo
+        Key foo has been deleted
+    """
+    if config.backend == "server":
+        raise click.UsageError(
+            "Key value commands with server are not currently supported."
+        )
+
+
+@kv.command(name="set", hidden=True)
+@click.argument("key")
+@click.argument("value")
+def _set(key, value):
+    """
+    Set a key value pair, overriding existing values if key exists
+
+    \b
+    Arguments:
+        key         TEXT    Key to set
+        value       TEXT    Value associated with key to set
+    """
+    client = Client()
+    result = client.set_key_value(key=key, value=value)
+
+    if result is not None:
+        click.secho("Key set successfully", fg="green")
+    else:
+        click.echo("An error occurred setting the key value pair", fg="red")
+
+
+@kv.command(hidden=True)
+@click.argument("key")
+def get(key):
+    """
+    Get the value of a key
+
+    \b
+    Arguments:
+        key         TEXT    Key to get
+    """
+    client = Client()
+    result = client.get_key_value(key=key)
+    click.secho(f"Key {key} has value {result}", fg="green")
+
+
+@kv.command(hidden=True)
+@click.argument("key")
+def delete(key):
+    """
+    Delete a key value pair
+
+    \b
+    Arguments:
+        key         TEXT    Key to delete
+    """
+    client = Client()
+    result = client.delete_key_value(key=key)
+    if result:
+        click.secho(f"Key {key} has been deleted", fg="green")
+    else:
+        click.secho("An error occurred deleting the key", fg="red")
+
+
+@kv.command(name="list", hidden=True)
+def _list():
+    """
+    List all key value pairs
+    """
+    client = Client()
+    result = client.list_keys()
+
+    if result:
+        click.secho("\n".join(result), fg="green")
+    else:
+        click.secho("No keys found", fg="red")
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -1816,3 +1816,91 @@ class Client:
             mutation,
             variables=dict(input=dict(task_run_artifact_id=task_run_artifact_id)),
         )
+
+    def set_key_value(self, key: str, value: Any) -> str:
+        """
+        Set key value pair, overwriting values for existing key
+
+        Args:
+            - key (str): the name of the key
+            - value (Any): A json compatible value
+
+        Returns:
+            - id (str): the id of the key value pair
+        """
+
+        if prefect.config.backend != "cloud":
+            raise ClientError("Key Value operations are Cloud only")
+
+        mutation = {
+            "mutation($input: set_key_value_input!)": {
+                "set_key_value(input: $input)": {"id"}
+            }
+        }
+
+        result = self.graphql(
+            query=mutation, variables=dict(input=dict(key=key, value=value))
+        )
+
+        return result.data.set_key_value.id
+
+    def get_key_value(self, key: str) -> Any:
+        """
+        Get the value for a key
+
+        Args:
+            - key (str): the name of the key
+
+        Returns:
+            - value (Any): A json compatible value
+        """
+        if prefect.config.backend != "cloud":
+            raise ClientError("Key Value operations are Cloud only")
+
+        query = {
+            "query": {
+                with_args("key_value", {"where": {"key": {"_eq": key}}}): {"value"}
+            }
+        }
+        result = self.graphql(query)  # type: Any
+        if len(result.data.key_value) == 0:
+            raise ValueError(f"No value found for key: {key}")
+        return result.data.key_value[0].value
+
+    def delete_key_value(self, key: str) -> bool:
+        """
+        Delete a key value pair
+
+        Args:
+            - key (str): the name of the key
+
+        Returns:
+            - success (bool): Whether or not deleting the key succeeded
+        """
+        if prefect.config.backend != "cloud":
+            raise ClientError("Key Value operations are Cloud only")
+
+        mutation = {
+            "mutation($input: delete_key_value_input!)": {
+                "delete_key_value(input: $input)": {"success"}
+            }
+        }
+
+        result = self.graphql(query=mutation, variables=dict(input=dict(key=key)))
+
+        return result.data.delete_key_value.success
+
+    # TODO - maybe support prefix searching here?
+    def list_keys(self) -> List[str]:
+        """
+        List all keys
+
+        Returns:
+            - keys (list): A list of all keys
+        """
+        if prefect.config.backend != "cloud":
+            raise ClientError("Key Value operations are Cloud only")
+        result = self.graphql(
+            {"query": {"key_value(order_by: {key: asc})": {"key"}}}
+        )  # type: ignore
+        return [res["key"] for res in result.data.key_value]
--- a/src/prefect/cli/__init__.py
+++ b/src/prefect/cli/__init__.py
@@ -17,6 +17,7 @@ from .run import run as _run
 from .server import server as _server
 from .heartbeat import heartbeat as _heartbeat
 from .build_register import register as _register, build as _build
+from .key_value import kv as _kv
 
 
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
@@ -73,6 +74,7 @@ cli.add_command(_server)
 cli.add_command(_heartbeat)
 cli.add_command(_register)
 cli.add_command(_build)
+cli.add_command(_kv)
 
 
 # Miscellaneous Commands
--- a/src/prefect/utilities/__init__.py
+++ b/src/prefect/utilities/__init__.py
@@ -11,3 +11,4 @@ import prefect.utilities.serialization
 import prefect.utilities.tasks
 import prefect.utilities.agent
 import prefect.utilities.storage
+import prefect.utilities.key_value
--- a/src/prefect/utilities/key_value.py
+++ b/src/prefect/utilities/key_value.py
@@ -1,0 +1,56 @@
+from typing import Any, List
+from prefect.client import Client
+
+
+def set_key_value(key: str, value: str) -> str:
+    """
+    Set key value pair, overwriting values for existing key
+
+    Args:
+        - key (str): the name of the key
+        - value (Any): A json compatible value
+
+    Returns:
+        - id (str): the id of the key value pair
+    """
+    client = Client()
+    return client.set_key_value(key=key, value=value)
+
+
+def get_key_value(key: str) -> Any:
+    """
+    Get the value for a key
+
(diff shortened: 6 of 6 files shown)


src/prefect/__init__.py
```diff
@@ -13,6 +13,7 @@
 import prefect.executors
 import prefect.engine.executors  # deprecated
 import prefect.artifacts
+import prefect.utilities.key_value as kv
```
madkinsz, line 16: I think we want to stop importing everything into the top-level as a general practice

src/prefect/cli/key_value.py
```diff
@@ -0,0 +1,109 @@
+import click
+
+from prefect import Client, config
+
+
+@click.group(hidden=True)
```
madkinsz, line 6: I'm not sure I understand why we so often hide then reproduce the default click output. I'd use Jim's work at `build_register` as best-practice instead.

src/prefect/cli/key_value.py
```diff
@@ -0,0 +1,109 @@
...
+    if config.backend == "server":
+        raise click.UsageError(
+            "Key value commands with server are not currently supported."
+        )
+
+
+@kv.command(name="set", hidden=True)
+@click.argument("key")
+@click.argument("value")
+def _set(key, value):
```
madkinsz, line 47: Minor preference for something like `set_command` instead

src/prefect/client/client.py
```diff
@@ -1816,3 +1816,91 @@ def delete_task_run_artifact(self, task_run_artifact_id: str) -> None:
             mutation,
             variables=dict(input=dict(task_run_artifact_id=task_run_artifact_id)),
         )
+
+    def set_key_value(self, key: str, value: Any) -> str:
```
madkinsz, line 1820: Could you move this and your other API methods to `prefect.backend.kv_store` and have them instantiate their own Client? Otherwise we'll have to deprecate and move it as in #4439

src/prefect/client/client.py
```diff
@@ -1816,3 +1816,91 @@ def delete_task_run_artifact(self, task_run_artifact_id: str) -> None:
...
+            "mutation($input: delete_key_value_input!)": {
+                "delete_key_value(input: $input)": {"success"}
+            }
+        }
+
+        result = self.graphql(query=mutation, variables=dict(input=dict(key=key)))
+
+        return result.data.delete_key_value.success
+
+    # TODO - maybe support prefix searching here?
```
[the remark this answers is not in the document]
zangell44: I'm gonna hold off on indexing, if there's a use case it's easy enough to add

src/prefect/cli/kv_store.py
```diff
@@ -0,0 +1,96 @@
+import sys
```
zangell44, line 1: I know there's a lot of boilerplate in these commands for logging and exiting. I'd like to refactor them when we create proper abstractions out of the `build_register` decorator / `TerminalError` work.

src/prefect/backend/kv_store.py
```diff
@@ -0,0 +1,123 @@
...
+        - key (str): the name of the key
+        - value (Any): A json compatible value
+
+    Returns:
+        - id (str): the id of the key value pair
+
+    Raises:
+        - ClientError: if using Prefect Server instead of Cloud
+    """
+    if prefect.config.backend != "cloud":
```
zangell44, line 23: again a bit of boilerplate here for all kv functions, would a decorator be overkill?

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,123 @@
...
+## UI
+
+You can view, update, and delete key value pairs on the [KV Store page](https://cloud.prefect.io/team/kv) of the UI.  
+
+## Setting Key Value Pairs
+
+Setting a key value pair will overwrite the existing value if the key exists.
+
+:::: tabs
+::: tab Core
```
madkinsz, line 20: I did this in my docs and it makes more sense than our `Core` lingo I think
```suggestion
::: tab Prefect library
```

src/prefect/backend/kv_store.py
madkinsz: Let's save the decorator for later when there are more backend functions that are Cloud only

src/prefect/backend/kv_store.py
```diff
@@ -0,0 +1,123 @@
...
+        - keys (list): A list of keys
+
+    Raises:
+        - ClientError: if using Prefect Server instead of Cloud
+    """
+    if prefect.config.backend != "cloud":
+        raise ClientError("Key Value operations are Cloud only")
+    client = Client()
+    result = client.graphql(
+        {"query": {"key_value(order_by: {key: asc})": {"key"}}}
```
madkinsz, line 121: I'm not sure we should order this, or if we do we should consider doing it locally. It seems like an unnecessary load on the database. @cicdw ?

src/prefect/cli/kv_store.py
```diff
@@ -0,0 +1,96 @@
...
+    Arguments:
+        key         TEXT    Key to get
+    """
+    try:
+        result = kv_store.get_key_value(key=key)
+        click.secho(f"Key {key} has value {result}", fg="green")
+    except Exception as exc:
+        log_exception(exc)
+        click.secho(f"Error retrieving value for key {key}", fg="red")
+        sys.exit(1)
```
madkinsz, line 60: Let's import `TerminalError` and `handle_terminal_error` and use that instead since it's used in `build_register` and `run` it'll make it easier to refactor later

tests/backend/test_kv_store.py
```diff
@@ -0,0 +1,177 @@
...
+    ):
+        gql_return = MagicMock(
+            return_value=MagicMock(
+                data=GraphQLResult(
+                    set_key_value=GraphQLResult({"id": "123"}),
+                )
+            )
+        )
+        client = MagicMock()
+        client.return_value.graphql = gql_return
```
madkinsz, line 25: You don't have to do this; it's frustrating that it's done so often in our tests

See an example of a nicer way at https://github.com/PrefectHQ/prefect/blob/3d8c4c943627c2c41ce8ffc2d90c720c330c7fc3/tests/cli/test_run.py#L626

I would write this as something like...
```python
        Client = MagicMock()
        Client().graphql.return_value = GraphQLResult(
            data={"set_key_value": {"id": "123"}}
        ) 
        monkeypatch.setattr("prefect.backend.kv_store.Client", Client)

        key_value_id = set_key_value(key="foo", value="bar")
        Client().graphql.assert_called_with(
            query={
                "mutation($input: set_key_value_input!)": {
                    "set_key_value(input: $input)": {"id"}
                }
            },
            variables={"input": {"key": "foo", "value": "bar"}},
        )
        assert key_value_id == "123"
```

`GraphQLResult` is a `Box` subclass so it'll roll all of the subdictionaries into the `Box` as well 🤷

src/prefect/backend/__init__.py
```diff
@@ -0,0 +1 @@
+import prefect.backend.kv_store
```
madkinsz, line 1: It seems reasonable to expose these top-level here as

`from prefect.backend import set_key_value` 

but I'm not sure if I feel strongly

madkinsz (dismissed):
Looking good!

docs/orchestration/concepts/kv_store.md
madkinsz: (later too)

src/prefect/backend/kv_store.py
cicdw: Given that there is no limit provided so the data being returned is the same either way, I agree with you @madkinsz - we should order it locally.

docs/orchestration/concepts/kv_store.md
madkinsz: I also did `GraphQL API` instead of `GraphQL` -- thoughts?

zangell44: works for me, updated

src/prefect/backend/__init__.py
zangell44: done

src/prefect/backend/kv_store.py
zangell44: makes sense, updated

tests/backend/test_kv_store.py
zangell44: good point, I copied one test case and re-used the logic without really examining it here, updated

src/prefect/cli/kv_store.py
zangell44: done

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,123 @@
+# KV Store
+
+Key Value Store is your managed metadata database within Prefect Cloud.
```
zhen0, line 3:
```suggestion
Key Value Store is a managed metadata database within Prefect Cloud.
```

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,123 @@
+# KV Store
```
zhen0, line 1: Am I correct that this will be a cloud only feature?  If so, can we add the Cloud badge?

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,123 @@
...
+
+Key Value Store is your managed metadata database within Prefect Cloud.
+
+**Keys** are strings. **Values** are json blobs.
+
+The number of key value pairs allowed is limited by license, starting with 50 pairs on the Free tier. Values are limited to 1 MB in size.
+
+Key value pairs can be configured via the Prefect CLI, Python client, API, and UI.
+
```
zhen0, line 10: Could we add a small example of a simple flow using kv? (Or would that maybe be better in a blog post?)

zhen0 (commented):
Some small suggestions for the docs.  I also wonder if we could add an example of using KV in a flow - maybe in the KV store docs or an advanced section of the tutorial or in a blog post?

docs/orchestration/concepts/kv_store.md
madkinsz: Yeah that sounds proper

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,123 @@
+# KV Store
+
+Key Value Store is your managed metadata database within Prefect Cloud.
+
+**Keys** are strings. **Values** are json blobs.
+
+The number of key value pairs allowed is limited by license, starting with 50 pairs on the Free tier. Values are limited to 1 MB in size.
```
ThatGalNatalie, line 7: Should we will list out the number of pairs for each license here?

zangell44: Hmm I stayed away from listing out every tier because it could change in the future. And it would make for a frustrating Sales process if our Core docs + backend + Sales team aren't all in sync.

docs/orchestration/concepts/kv_store.md
zangell44: good call, updated

docs/orchestration/concepts/kv_store.md
zangell44: updated

docs/orchestration/concepts/kv_store.md
ThatGalNatalie: Ahh okay. That makes sense.

zangell44:
Good call @zhen0! I added a simple example based on previous community feedback. Let me know what you think

tests/backend/test_kv_store.py
```diff
@@ -0,0 +1,156 @@
...
+class TestListKeyValue:
+    def test_list_key_value_raises_on_server_backend(self, server_api):
+        with pytest.raises(ClientError):
+            list_keys()
+
+    def test_list_key_value_calls_client_mutation_correctly(
+        self, monkeypatch, cloud_api
+    ):
+        client = MagicMock()
+        client().graphql.return_value = GraphQLResult(
```
madkinsz, line 142: Note that I capitalized this intentionally so it mirrors the `Client` you are mocking. Lowercase makes it less clear (imo) that you are mocking the class not the instance. Not a big deal though.

madkinsz (changes requested):
Just some minor details

src/prefect/cli/kv_store.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Get the value of a key
+
+    \b
+    Arguments:
+        key         TEXT    Key to get
+    """
+    try:
+        result = kv_store.get_key_value(key=key)
+        click.secho(f"Key {key} has value {result}", fg="green")
```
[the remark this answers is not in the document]
zangell44: ahhhhhh so that's what `!r` does

docs/orchestration/concepts/kv_store.md
```diff
@@ -0,0 +1,161 @@
...
+mutation {
+  set_key_value(input: { key : "foo", value: "\"bar\"" }) {
+    id
+  }
+}
+```
+:::
+::::
+
+## Getting the Value of a Key
```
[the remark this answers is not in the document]
zangell44: updated wherever it seemed relevant

src/prefect/backend/kv_store.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Returns:
+        - success (bool): Whether or not deleting the key succeeded
+
+    Raises:
+        - ValueError: if the specified key does not exist
+        - ClientError: if using Prefect Server instead of Cloud
+    """
+    if prefect.config.backend != "cloud":
+        raise ClientError("Key Value operations are Cloud only")
```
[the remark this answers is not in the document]
zangell44: updated everywhere

src/prefect/cli/kv_store.py
```diff
@@ -0,0 +1,99 @@
...
+    Arguments:
+        key         TEXT    Key to set
+        value       TEXT    Value associated with key to set
+    """
+    try:
+        kv_store.set_key_value(key=key, value=value)
+        click.secho("Key value pair set successfully", fg="green")
+    except Exception as exc:
+        raise TerminalError(
+            f"An error occurred setting the key value pair: {str(exc)}"
```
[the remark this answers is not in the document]
zangell44: haha this is why I originally skipped the decorator, it ended up being the same number of lines of code to just log the exception and exit

I'll just `log_exception`

src/prefect/cli/kv_store.py
```diff
@@ -0,0 +1,99 @@
...
+        ) from exc
```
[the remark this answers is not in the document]
zangell44: good call, updated

tests/backend/test_kv_store.py
zangell44: fair point, updated

changes/pr4499.yaml
```diff
@@ -0,0 +1,2 @@
+feature:
+  - "Key Value Store - [#4499](https://github.com/PrefectHQ/prefect/pull/4499)"
```
madkinsz, line 2: Could this be more user-facing?

zangell44: updated, I'm not a very good marketer though 🤷‍♂️

madkinsz (dismissed):

changes/pr4499.yaml
madkinsz: Maybe "Add interface for backend key-value metadata store" @zangell44

zangell44: fine by me, i like it

madkinsz (dismissed):

madkinsz (approved):

PR #4501, opened 2021-05-10
Fix server startup check for custom server port

madkinsz:
## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -621,7 +621,10 @@ def start(
         ):
             while not started:
                 try:
-                    client = prefect.Client()
+                    # Get a client with the correct server port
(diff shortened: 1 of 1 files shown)


src/prefect/cli/server.py
```diff
@@ -621,7 +621,10 @@ def start(
         ):
             while not started:
                 try:
-                    client = prefect.Client()
+                    # Get a client with the correct server port
+                    client = prefect.Client(
+                        api_server=f"{config.server.host}:{server_port}"
```
madkinsz, line 626: I would have liked to reuse `config.server.endpoint` here but because we fill interpolated config values at load-time rather than access-time we cannot set a port in a temporary config :(

zangell44 (approved):

PR #4514, opened 2021-05-12
Fix conflict handling for docker container naming

zanieb:
## Summary

The objects returned were not in the format expected from the documentation at https://docker-py.readthedocs.io/en/stable/containers.html#container-objects and have multiple names with leading `/` and such. This makes a try/except model much more appealing.

## Importance

Fixes a docker agent on name conflicts / any existing containers

### DIFF
--- a/tests/agent/test_docker_agent.py
+++ b/tests/agent/test_docker_agent.py
@@ -315,14 +315,20 @@ def test_docker_agent_deploy_flow_sets_container_name_with_index(api, collision_
 
     if collision_count:
         # Add the basic name first
-        existing_containers = [{"name": "flow-run-name"}]
+        existing_names = ["flow-run-name"]
         for i in range(1, collision_count):
-            existing_containers.append({"name": f"flow-run-name-{i}"})
+            existing_names.append(f"flow-run-name-{i}")
     else:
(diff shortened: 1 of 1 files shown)


src/prefect/agent/docker/agent.py
```diff
@@ -458,25 +458,29 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
+                    environment=env_vars,
+                    name=container_name,
+                    volumes=container_mount_paths,
+                    host_config=self.docker_client.create_host_config(**host_config),
+                    networking_config=networking_config,
+                    labels=labels,
+                )
+            except docker.errors.APIError as exc:
+                if "Conflict" in str(exc) and "container name" in str(exc):
+                    index += 1
```
cicdw, line 478: Is this necessary? Seems like it could be simpler to just append a bit of a UUID if a conflict is found?

madkinsz: I guess this is guaranteed to work eventually whereas a bit of UUID isn't actually universally unique and we'd still have to loop. Easier to test this way too. We can revisit the implementation in the future, this has been a nightmare to try to fix quickly :(

PR #4519, opened 2021-05-12
Remove the netlify pin flow from CI

madkinsz:
## Summary

This job just breaks because the token is out of date and it's contrived to use a flow for it.


jcrist (approved):
🔥

PR #4521, opened 2021-05-12
Add upper version click pin

zangell44:
## Summary

Pins `click` < 8.0. There's a dependency issue with `dask-distributed` here https://github.com/dask/distributed/issues/4812.

## Importance

Dask executors using the Prefect image will not work if the newest version of `click` is installed.

## Checklist

This PR:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-click >= 7.0
+click >= 7.0, <8.0
 cloudpickle >=1.3.0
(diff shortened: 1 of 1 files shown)


madkinsz:
Also a compatibility issue with Server since it has an older click installed?

zangell44:
> Also a compatibility issue with Server since it has an older click installed?
click 7.x will work with both though right?

madkinsz (approved):

PR #4530, opened 2021-05-14
Move mapped task interation from `FlowRunView` to `TaskRunView`

madkinsz:
## Summary

Previously, `FlowRunView` supplied a `iter_mapped_task_runs` but it doesn't really make sense to have it there since the whole point of iterating is that your mapped results are too big to cache. The new workflow looks like...

```python
flow_run = FlowRunView.from_flow_run_id("...")

task_run = flow_run.get_task_run(my_mapped_task)

# Access all child results at once
task_run.result  # [1,2,3,4,5]

# Lazily iterate over child tasks
for child in task_run.iter_mapped():
    print(child.result)
# 1
# 2
# 3
# 4
# 5
```

## Changes

- Removes `FlowRunView` mapped task iteration
- Adds `TaskRunView.iter_mapped()`
- Fixes `FlowView` repr missing newline
- Adds `map_index` kwarg to `TaskRunView.from_task_slug`

## Importance

A better API

## Checklist

This PR:

### DIFF
--- a/src/prefect/backend/flow.py
+++ b/src/prefect/backend/flow.py
@@ -296,7 +296,7 @@ class FlowView:
                 [
                     f"flow_id={self.flow_id!r}",
                     f"name={self.name!r}",
-                    f"project_name={self.project_name!r}"
+                    f"project_name={self.project_name!r}",
                     f"storage_type={type(self.storage).__name__}",
                 ]
             )
--- a/src/prefect/backend/flow_run.py
+++ b/src/prefect/backend/flow_run.py
@@ -10,7 +10,6 @@ from typing import (
     Dict,
     Set,
     Any,
-    Iterator,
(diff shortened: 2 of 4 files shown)


madkinsz:
@jcrist @zach-angell since the `result` attribute could be really slow for mapped tasks where we need to query for all the child states then pull all the results from their disk locations, should I change it to `get_result()` ?

jcrist:
I think so? Attribute access that can do some expensive action can be unintuitive for the user, while a method better indicates that some work is happening. Also, some interactive environments may accidentally compute on tab-completion (ipython used to have this bug), resulting in unexpectedly expensive tab-completion.

madkinsz:
I'll do that in a separate PR, I think it's the right move.

Issue #3454, opened 2020-10-07
Task Proposal: Add Kafka batch consume and produce tasks


tchoedak:
# Use Case(s)

Data pipelines in many orgs are required to interact with Kafka as both a data source and as a data sink. These data pipelines that in my experience often tend to be batch based - they spin up, read a set of message from queue, perform some action, spin down.

Batch Kafka Consume -> Snowflake Flow
- Consume all messages from a Kafka topic
- Load messages into a Snowflake table

Snowflake -> Transform -> Batch Kafka Produce Flow
- Query Snowflake for a dataset
- Apply business logic transformations via pandas/spark/data-processing tool to fit the end data model on an application
- Convert data to messages
- Produce to a Kafka topic

# Task API
All kafka related tasks would live under `src/prefect/tasks/kafka/`

### __init__()
- bootstrap_servers: string of all kafka brokers to connect to. Examples: `'localhost:9092'`, multiple brokers: `'localhost:9092,localhost:9093'`
- group_id: string Group id assigned to consumers from this consume task
- kwargs: all other task kwargs

### run()

- topic: string topic to consume from
- timeout: float timeout for polling operations
- auto_offset_reset: string offset strategy. several options available from kafka configs, but `earliest/smallest` are the most common for data pipeline tasks while `latest/largest` is common for application consumers.
- message_consume_limit: int number of messages consumed to limit to for this task. Optional and only needed in data pipelines designed to handle a fixed input of messages.
- -> Returns List of messages

### __init__()
- bootstrap_servers: string of all kafka brokers to connect to.

### run()
- topic: string topic to produce messages to
- messages: list of messages to produce.
  - If a message is a dictionary, check that key/value are keys.
    - produce with key=message['key'] and value=message['value]
  - Otherwise produce all messages with value=message (no key specified)
- threshold: int count of messages produce that trigger a producer.flush() - flushing regularly is a best practice
- callback: Callable callback upon delivery of a message

# Backend Client
I would recommend [confluent-kafka-python](https://github.com/confluentinc/confluent-kafka-python) as the backend client that actually does the hard work of producing messages and consuming messages. This lib in my experience, while not as mature as [kafka-python](https://pypi.org/project/kafka-python/), is maintained by the organization backing kafka, `confluent`, and is more performant.

# Design Choices

[...]

PR #4533, opened 2021-05-15
Add (batch) Kafka tasks to the tasks library

tchoedak:
## Summary

The motivation and use cases are described in this #3454. This PR provides a new kafka module under the tasks lib that come with two task classes: a high level KafkaBatchConsume task and a high level KafkaBatchProduce task.

## Changes

- This PR adds a new directory under the `tasks` lib for `kafka` related tasks and introduces a `kafka` module.
- The kafka module provides two tasks `KafkaBatchConsume` and `KafkaBatchProduce`
- Tests and examples are included
- An extra dependency is available in `setup.py` with name `kafka` in order to install `confluent-kafka` when using these new tasks

## Importance

- It provides a useful task that will allow prefect users to build flows around a common data engineering operation: consuming or producing a match of messages from/to Kafka.

## Checklist

This PR:


alexkreidler:
This is an important feature for me, so it would be amazing if this PR could be reviewed and merged. ping @cicdw @joshmeek @jlowin @jcrist\

Thanks!

src/prefect/tasks/kafka/kafka.py
```diff
@@ -0,0 +1,211 @@
...
+            consumer.close()
+
+        return messages
+
+
+class KafkaBatchProduce(Task):
+    """
+    Task for producing a batch of messages to a Kafka topic.
+
+    Args:
```
zangell44, line 134: args like `topic` are required either at init or runtime, could we clarify this in the docstring?

examples/kafka/kafka_flow.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    kafka_produce.run(
+        bootstrap_servers=BOOTSTRAP_SERVER,
+        topic=TOPIC,
+        messages=messages[20000:],
+        flush_threshold=1000,
+    )
+
+    time.sleep(5)
```
zangell44, line 38: instead of `time.sleep`, can you set the produce upstream of the consume tasks? the docs here may be helpful https://docs.prefect.io/core/concepts/flows.html#imperative-api

examples/kafka/kafka_flow.py
```diff
@@ -0,0 +1,56 @@
...
+    print(f"Total messages: {len(x)}")
+
+
+kafka_consume = KafkaBatchConsume(BOOTSTRAP_SERVER, GROUP_ID)
+kafka_produce = KafkaBatchProduce(BOOTSTRAP_SERVER)
+
+
+with Flow("Kafka Example") as flow:
+
+    messages = [{"key": str(i), "value": str(i)} for i in range(30000)]
```
zangell44, line 23: could make `messages` a parameter? https://docs.prefect.io/core/concepts/parameters.html#parameters

src/prefect/tasks/kafka/kafka.py
```diff
@@ -0,0 +1,211 @@
...
+        **consumer_options,
+    ) -> List[bytes]:
+        """
+        Run method for this Task. Invoked by calling this Task after initialization within a
+        Flow context, or by using `Task.bind`.
+        Args:
+        - bootstrap_servers (str:, required): comma separated host and port pairs that are the
+            addresses of kafka brokers
+        - group_id (str:, required): name of the consumer group the consumer will belong to
+        - topic (List[str], required): list of topic names to consume messages from
```
zangell44, line 69: This parameter name should be plural -`topics`

zangell44 (changes requested):
Looking really good, thanks for contributing @tchoedak! Left a couple comments

src/prefect/tasks/kafka/kafka.py
```diff
@@ -0,0 +1,211 @@
...
+import confluent_kafka
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class KafkaBatchConsume(Task):
+    """
+    Task for consuming a batch of messages from a Kafka topic.
+
+    Args:
```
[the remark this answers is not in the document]
tchoedak: any recommendations on how to note that? I see in some other tasks that they're actually documented as `optional` and that a runtime check is performed to ensure they're available: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/fivetran/fivetran.py#L65-L70

examples/kafka/kafka_flow.py
tchoedak: hmm, i'd suggest not since messages should actually be built in a separate task instead of something that is configured as a parameter. Parameters seems like its more reserved for optionally changing configurations at runtime/via UI, is that right?

madkinsz: I agree it makes sense not to do a Parameter. It might make sense to have task that generates these messages though. Also this seems like a lot of messages for a toy example.

src/prefect/tasks/kafka/kafka.py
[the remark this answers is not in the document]
madkinsz: Yeah this is a little confusing and will be resolved at some point in the future as we establish better patterns for the task library. I think your suggestion is quite reasonable for now.

src/prefect/tasks/kafka/kafka.py
```diff
@@ -0,0 +1,227 @@
...
+        - bootstrap_servers (str, required): comma separated host and port pairs that are the
+            addresses of kafka brokers
+        - group_id (str, required): name of the consumer group the consumer will belong to
+        - topics (List[str], required): list of topic names to consume messages from
+        - request_timeout (float, optional): Maximum time to block waiting for message, event or callback
+        - auto_offset_reset (str, optional): configurable offset reset policy
+        - message_consume_limit (int, optional): max number of messages to consume before closing
+            the consumer
+        - kafka_configs (dict, optional): a dict of kafka client configuration properties used to
+            construct the consumer.
```
madkinsz, line 81: Can you indent these?

madkinsz:
@tchoedak thanks for contributing! If you address the last couple of little changes we can get this in the release today otherwise we'll slot it for the next release.

src/prefect/tasks/kafka/kafka.py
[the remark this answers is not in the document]
tchoedak: Feel free to tag me in an issue for when you guys have established a better pattern that requires updating task docstrings!

madkinsz (approved):
Thanks a bunch!

Trymzet:
Hi guys, looks like the task wasn't added to the [task library](https://docs.prefect.io/core/task_library/overview.html#overview)

PR #4532, opened 2021-05-14
Add per-agent flow concurrency limits

madkinsz:
## Summary

Adds flow concurrency limits at the agent level. This allows the user to limit the number of running or submitted flows spawned by a single agent.

## Changes

- Adds `--concurrency-limit/-L <int>` to the agent CLI
- Adds `max_concurrent_runs: Optional[int]` to the `Agent` base class
- Uses `**kwargs` to pass `Agent` subclass settings to the base class. Having duplicated default settings and documentation seems like it'll expose issues and introduces maintenance pains as the base class is extended.
- Adds a query to check for concurrency limits before submitting ready flow runs for deployment

## Importance

Closes https://github.com/PrefectHQ/server/issues/213 although there are a few things listed in that ticket that are unaddressed:
- The agent screen in the UI. This is in progress and should be updated in the future to display this limit.
- Storing the limit in the backend. This PR sets the limit locally via CLI flag and does not use the API for storing the limit. In the future, we'd likely want to be able to set the limit from the backend as well.
- Agent ids can still be shared by multiple agent processes. Enforcing unique names in the future seems ideal, but for now it's on the user if they run into concurrency issues with agents sharing an id.

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [x] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #4541, opened 2021-05-17
Cleanup `Agent._get_flow_run_metadata`

madkinsz:
## Summary

Adds some comments explaining the query that is used during metadata retrieval and clarifies the filtering used for retries and duplicate submission prevention.  

## Changes

- `_get_ready_flow_runs` returns a `Set[str]` of flow run ids instead of calling `_get_flow_run_metadata` directly
- `_get_flow_run_metadata` is called separately after `_get_ready_flow_runs` which will be useful for #4532 in which we may want to skip metadata queries

## Importance

Clarifies confusing behavior; refactors a little to make #4532 simpler

## Checklist

This PR:

- ~[ ] adds a change file in the `changes/` directory (if appropriate)~ not user facing

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -588,7 +588,7 @@ class Agent:
         flow_run_ids: Iterable[str],
     ) -> List["GraphQLResult"]:
         """
-        Get metadata about a collection of flow run ids to that the agent is preparing
+        Get metadata about a collection of flow run ids that the agent is preparing
         to submit
 
         This function will filter the flow runs to a collection where:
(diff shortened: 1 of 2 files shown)


src/prefect/agent/agent.py
```diff
@@ -593,29 +613,34 @@ def _get_flow_run_metadata(
         flow_run_ids = list(flow_run_ids)
         self.logger.debug(f"Retrieving metadata for {len(flow_run_ids)} flow run(s)...")
 
+        # This buffer allows flow runs to retry immediately in their own deployment
+        # without the agent creating a second deployment
+        retry_start_time_buffer = pendulum.now("UTC").subtract(seconds=3).isoformat()
+
+        where = {
```
zangell44, line 620: should we also be doing filtering based on agent labels here?

zangell44 (approved):
One question but LGTM

src/prefect/agent/agent.py
madkinsz: `get_runs_in_queue` takes care of that already

PR #4544, opened 2021-05-17
Handle events with missing timestamps in K8s agent

madkinsz:
## Summary

It looks like the event timestamp can be null per https://prefect-community.slack.com/archives/CL09KU1K7/p1621293779234300 while streaming event logs from K8s

## Changes

- Events with null timestamps are ignored

## Importance

Fixes bug

## Checklist

This PR:

### DIFF
--- a/changes/pr4544.yaml
+++ b/changes/pr4544.yaml
@@ -1,0 +1,3 @@
+
+fix:
+  - "Handle events with missing timestamps in K8s agent - [#4544](https://github.com/PrefectHQ/prefect/pull/4544)"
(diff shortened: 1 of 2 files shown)


madkinsz:
I think this is a sufficiently weird case to justify leaving it out of tests

cicdw (approved):

ciaranevans:
@madkinsz you might still be getting a similar error, but on this line: https://github.com/PrefectHQ/prefect/blob/master/src/prefect/agent/kubernetes/agent.py#L229-L231

As per the conversation here: https://prefect-community.slack.com/archives/CL09KU1K7/p1624442574450100

It doesn't appear to affect running jobs, just fails on the first attempt.

madkinsz:
Thanks @ciaranevans -- pretty confused that this behavior would change on our end but https://github.com/PrefectHQ/prefect/pull/4693 should address this attribute missing. If there are malformed events we'll log them for debugging now.

Issue #4377, opened 2021-04-08
Ensure Task Library Tasks can be completely configured from the `run` method


dylanbhughes:
## Current behavior

Some tasks in the Task Library require a user to provide certain pieces of configuration as part of the initialization method. Whenever we can, we should try to avoid this because it means that those variables are locked it at the time the Flow is built, instead of at runtime, which prevents the use of Parameters or Secrets to configure the task.

## Proposed behavior

Let's audit our Task library and ensure that these types of arguments can be provided at either initialization or at runtime whenever possible.

## Example

Our [MySQL Task](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/mysql/mysql.py#L9) is one example. Even though the Task initializes the MySQL connector as part of the `run` method, the `hostname` must be provided at initialization.

RajdeepRao:
Hey @dylanbhughes I'm new around here, but would love to contribute to prefect! I don't quite know the best way to claim this, but leaving a comment here for now?

dylanbhughes:
Welcome, @RajdeepRao! Go ahead and assign this to yourself 💯 

If you have any questions, [join us in Slack ](https://www.prefect.io/slack) in the #prefect-contributors channel 😄

StefanBRas:
In the general case, should eveything that can go in `run` always go in `run` or does it make sense to have some variables be configured in either `init` or `run`?

I have some cases where a task has variables that are almost always static, but they could also be assigned at run time if needed.

You might want to run this task several times in a flow, so it can save you from giving the same inputs several times. Kind of like using `partial` from `functool` to fix some variables.

Example:
```python
from prefect import Task, Flow

class DoSomething(Task): 
    def __init__(self, some_variable = None, **kwargs): 
        self.some_variable = some_variable 
        super().__init__(**kwargs) 
    def run(self, some_variable=None): 
        run_time_variable = some_variable or self.some_variable 
        return run_time_variable
        
with Flow('test') as flow: 
    init_task = DoSomething(some_variable='init') 
    run_task = DoSomething() 
    init_task() 
    run_task(some_variable='run')
```

dylanbhughes:
@StefanBRas the advantage of having them in `run` is that they're configurable at runtime by `Parameters` or `Secrets`, which aren't available at initialization time. In general, I try to put all args in both `init` and `run` whenever possible to give users the option

Tostino:
[...]

Issue #3845, opened 2020-12-15
MySQL Tasks: support SSL connection


jacob-lester:
Since the MySQL task uses pymysql under the hood, is there a way you all can include the "ssl" named parameter in order to query/execute using secure connection? https://github.com/PrefectHQ/prefect/blob/076de76c4b7ba2c5c870230e1bfff0526e70aa84/src/prefect/tasks/mysql/mysql.py#L68-L75

joshmeek:
Hey @jacob-lester I believe the `ssl` keyword argument could be added to this community contributed task and we welcome a PR to add it 🙂

tchoedak:
@joshmeek @jacob-lester hey guys, is this what you're looking for? https://github.com/PrefectHQ/prefect/pull/4545

PR #4545, opened 2021-05-18
Support all arguments in MySQL Tasks' run methods and support SSL configuration

tchoedak:
## Summary

This PR introduces changes that address two issues #3845 and #4377. This should enable the mysql tasks to be configured completely from just the `run` method via secrets/params and also allows `ssl` options to be configured when creating a mysql connection.

## Changes
- update all init args to be keyword args
- update all run args to have all the same keyword args as __init__ and overrided via the `defaults_from_attrs` decorator
- support passing a `ssl` keyword argument with a dict to configure ssl when connecting

## Importance
- Allows Parameters and Secrets to be used for this task, which makes this task better support prefect features
- Allows ssl options to be configured so MySQL DB's with TLS >= 1.2 can be connected to with these tasks.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/mysql/mysql.py
+++ b/src/prefect/tasks/mysql/mysql.py
@@ -216,24 +216,24 @@ class MySQLFetch(Task):
         Task run method. Executes a query against MySQL database and fetches results.
 
         Args:
-        - db_name (str): name of MySQL database
-        - user (str): user name used to authenticate
-        - password (str): password used to authenticate
-        - host (str): database host address
-        - port (int, optional): port used to connect to MySQL database, defaults to 3307 if not
-            provided
-        - fetch (str, optional): one of "one" "many" or "all", used to determine how many
-            results to fetch from executed query
-        - fetch_count (int, optional): if fetch = 'many', determines the number of results to
-            fetch, defaults to 10
-        - query (str, optional): query to execute against database
-        - commit (bool, optional): set to True to commit transaction, defaults to false
-        - charset (str, optional): charset of the query, defaults to "utf8mb4"
-        - cursor_type (Union[str, Callable], optional): The cursor type to use.
-            Can be `'cursor'` (the default), `'dictcursor'`, `'sscursor'`, `'ssdictcursor'`,
-            or a full cursor class.
-        - ssl (dict, optional): A dict of arguments similar to mysql_ssl_set()’s
-                parameters used for establishing encrypted connections using SSL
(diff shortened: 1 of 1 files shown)


src/prefect/tasks/mysql/mysql.py
```diff
@@ -44,17 +47,47 @@ def __init__(
...
+    def run(
+        self,
+        db_name: str = None,
+        user: str = None,
+        password: str = None,
+        host: str = None,
+        port: int = 3306,
+        query: str = None,
+        commit: bool = False,
+        charset: str = "utf8mb4",
```
madkinsz, line 73: These should be `None` since the defaults come from the `__init__` values

src/prefect/tasks/mysql/mysql.py
```diff
@@ -141,34 +178,60 @@ def __init__(
...
+        user: str = None,
+        password: str = None,
+        host: str = None,
+        port: int = 3306,
         fetch: str = "one",
         fetch_count: int = 10,
+        query: str = None,
         commit: bool = False,
         charset: str = "utf8mb4",
         cursor_type: Union[str, Callable] = "cursor",
```
madkinsz, line 210: Same as above, can you change these defaults to `None` since the default value is defined at init time?

madkinsz (commented):
Thanks! This looks great. Did you test the SSL connection? It looks like `.connect` takes additional ssl variables `ssl_ca, ssl_cert, ssl_disabled, ssl_key, ssl_verify_cert, ssl_verify_identity` but it kind of looks like those can all be passed with `ssl`? https://dev.mysql.com/doc/c-api/5.7/en/mysql-ssl-set.html

tchoedak:
@madkinsz no the `ssl` param hasn't been tested, i'll try and get a mysql instance with ssl enabled to test this out

tchoedak:
@madkinsz I was able to connect to a MySQL DB with SSL enabled/required by passing in `ssl_ca`, `ssl_cert`, and `ssl_key`. I tested that the other options are _accepted_ but not that they have the intended effect (for example, these 3 options + ssl_disabled=True still results in an SSL connection created, whether that's intended or not seems to be controlled either by MySQL or a layer below).

Screeny for proof that it works!
Screen Shot 2021-05-31 at 9 06 35 PM

madkinsz (approved):
Sweet! Thanks @tchoedak

PR #4559, opened 2021-05-20
Enable parallel pytest runs with xdist

madkinsz:
## Summary

Improves test times in CI from 4m 20s to 1m 30s for a ~3x speedup by using `pytest-xdist` to parallelize execution.

## Changes

- Identifies and reduces brittleness of some tests that are not robust to running in parallel
- Upgrades pytest/xdist
- Adds multiprocesss test execution to CI
- Fixes (hopefully) frequently brittle timeout test

## Importance

Faster CI = Faster feedback

## Future work

- Do we want a makefile or something instead of repeating the pytest args repeatedly? It'd be nice for local development too
- Appveyor is still slow, ought to replace it soon anyway
- Netlify builds block merge; is that necessary? We're not publishing `master` docs anyway so can we just make sure it passes before a release manually?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
 
       - run:
           name: Run tests
-          command: pytest tests -vvrfEsx
+          command: pytest tests -vvrfEsx --numprocesses 4 --dist=loadfile
 
   # test a standard install of prefect
   # this ensures we correctly capture all ImportError sitautions
@@ -158,7 +158,7 @@ jobs:
 
       - run:
           name: Run tests
-          command: pytest tests -vvrfEsx
+          command: pytest tests -vvrfEsx --numprocesses 4 --dist=loadfile
 
   # test task library explicitly
   test_task_library:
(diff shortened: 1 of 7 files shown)


tests/utilities/test_edges.py
```diff
@@ -3,7 +3,7 @@
 from prefect import Flow, Task, task
 from prefect.utilities import edges, tasks
 
-ALL_ANNOTATIONS = {edges.unmapped, edges.mapped, edges.flatten}
+ALL_ANNOTATIONS = [edges.unmapped, edges.mapped, edges.flatten]
```
zangell44, line 6: why did we change this?

madkinsz: The commit message explains it a little :P but a `set` is unordered so when tests are loaded in parallel pytest will throw an error saying the workers detected different tests

.circleci/config.yml
```diff
@@ -128,7 +128,7 @@ jobs:
 
       - run:
           name: Run tests
-          command: pytest tests -vvrfEsx
+          command: pytest tests -vvrfEsx --numprocesses 4 --dist=loadfile
```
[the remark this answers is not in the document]
madkinsz: I ran one locally and it was roughly the same. I think `loadfile` will do more for us in reducing parallel brittleness than speedup

zangell44 (approved):

jcrist (approved):

PR #4568, opened 2021-05-23
Fix bug where agents use scheduled_start_time instead of state.start_time

zangell44:
## Summary

Fixes bug where agents use `FlowRun.scheduled_start_time` instead of `FlowRun.state/serialized_state.start_time`. 

## Importance

For Scheduled Runs, the "Start Now" functionality does not work because agents are waiting until the FlowRun's originally scheduled start time to execute. See issue originally filed in Cloud here https://github.com/PrefectHQ/cloud/issues/3410

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -24,7 +24,7 @@ from prefect.serialization.state import StateSchema
 from prefect.serialization.run_config import RunConfigSchema
 from prefect.utilities.context import context
 from prefect.utilities.exceptions import AuthorizationError
-from prefect.utilities.graphql import GraphQLResult, with_args, EnumValue
+from prefect.utilities.graphql import GraphQLResult, with_args
 
 ascii_name = r"""
  ____            __           _        _                    _
@@ -643,13 +643,7 @@ class Agent:
 
         query = {
             "query": {
-                with_args(
-                    "flow_run",
-                    {
-                        "where": where,
-                        "order_by": {"scheduled_start_time": EnumValue("asc")},
-                    },
-                ): {
+                with_args("flow_run", {"where": where,}): {
                     "id": True,
                     "version": True,
                     "state": True,
@@ -680,7 +674,14 @@ class Agent:
             }
         }
         result = self.client.graphql(query)
-        return result.data.flow_run
+        return sorted(
+            result.data.flow_run,
+            key=lambda flow_run: getattr(
+                StateSchema().load(flow_run.serialized_state),
+                "start_time",
+                pendulum.now("utc"),
+            ),
+        )
 
     def _mark_flow_as_submitted(self, flow_run: GraphQLResult) -> None:
         """
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -363,7 +363,7 @@ class Agent:
             #
             # There are two possible states the flow run could be in at this point
             # - Scheduled - in this case the flow run state will have a start time
-            # - Running - in this case the flow run state may not have a start time and we default to now
+            # - Running - in this case the flow run state will not have a start time so we default to now
             flow_run_state = StateSchema().load(flow_run.serialized_state)
             start_time = getattr(flow_run_state, "start_time", pendulum.now())
             delay_seconds = max(0, (start_time - pendulum.now()).total_seconds())
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -643,7 +643,7 @@ class Agent:
 
         query = {
             "query": {
-                with_args("flow_run", {"where": where,}): {
+                with_args("flow_run", {"where": where}): {
                     "id": True,
                     "version": True,
                     "state": True,
--- a/src/prefect/agent/agent.py
+++ b/src/prefect/agent/agent.py
@@ -20,7 +20,7 @@ from prefect import config
 from prefect.client import Client
 from prefect.engine.state import Failed, Submitted
 from prefect.run_configs import RunConfig, UniversalRun
-from prefect.serialization import state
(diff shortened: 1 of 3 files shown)


tests/agent/test_agent.py
```diff
@@ -456,15 +456,19 @@ def test_deploy_flow_run_sleeps_until_start_time(monkeypatch, cloud_api):
         flow_run=GraphQLResult(
             {
                 "id": "id",
-                "serialized_state": Scheduled().serialize(),
+                "serialized_state": Scheduled(
+                    start_time=dt.add(seconds=10)
+                ).serialize(),
                 "scheduled_start_time": str(dt.add(seconds=10)),
```
jlowin, line 462: The fact that this test passed is actually the issue, it must have been looking at the scheduled start time. I recommend changing this line to make it clear that the state, not the scheduled start time, is the key.
```suggestion
                "scheduled_start_time": str(dt),
```

jlowin (changes requested):

tests/agent/test_agent.py
zangell44: yep good catch, updated

src/prefect/agent/agent.py
```diff
@@ -679,7 +674,14 @@ def _get_flow_run_metadata(
...
         }
         result = self.client.graphql(query)
-        return result.data.flow_run
+        return sorted(
+            result.data.flow_run,
+            key=lambda flow_run: getattr(
+                StateSchema().load(flow_run.serialized_state),
+                "start_time",
+                pendulum.now("utc"),
+            ),
```
madkinsz, line 683: Could this be `key=lambda flow_run: flow_run.serialized_state.get("start_time", pendulum.now("utc").isoformat())` ? I feel like we are de-serializing the state a bunch of times.

src/prefect/agent/agent.py
```diff
@@ -356,7 +356,8 @@ def _deploy_flow_run(
             # Wait for the flow run's start time. The agent pre-fetches runs that may
             # not need to start until up to 10 seconds later so we need to wait to
             # prevent the flow from starting early
-            start_time = pendulum.parse(flow_run.scheduled_start_time)
+            flow_run_state = StateSchema().load(flow_run.serialized_state)
+            start_time = getattr(flow_run_state, "start_time", pendulum.now())
```
madkinsz, line 360: There should be a comment explaining why we are using the state start time and not the `scheduled_start_time`.

Also a comment explaining that we may be working with both `Running` and `Scheduled` states and that's why we need to use `getattr`

madkinsz (changes requested):
I get that this is fixing an issue, but I find this super confusing. What is `FlowRun.scheduled_start_time` if the state has the start time that matters? I think I would rather make it so UI changes the `scheduled_start_time` instead. Do we use state start times elsewhere?

zangell44:
> I get that this is fixing an issue, but I find this super confusing. What is `FlowRun.scheduled_start_time` if the state has the start time that matters? I think I would rather make it so UI changes the `scheduled_start_time` instead. Do we use state start times elsewhere?


I agree this is confusing, I had to talk this through with @jlowin to figure out what was going on. `state.start_time` is what we should be using for execution. `scheduled_start_time` is JUST the originally scheduled start time.

So if we have a run scheduled for 9 pm tonight and I run the flow at 10 am
- `scheduled_start_time` is always 9 pm (and we can view this in the UI)
- when I run the flow, a new `Scheduled` `state` is inserted into the database and `state.start_time` is 10 am

jlowin:
@madkinsz your confusion stems from a conflation of a property of the `FlowRun` and a property of the `FlowRunState`, which is one of the reasons having both on this data model is a suboptimal (though useful) idea.

The scheduled start time of a run is the time it was originally intended to run. For example, it might be your 10am run, or your 10am run for a backfilled date of January 1st. That's a property of the run. The state is how the system communicates intent and actions. So you might want your 10am run to start early at 9:50. It's still your 10am run (`scheduled_start_time`) but you've given it a state with an instruction to start at 9:50 (`state_start_time`). We always introduce new states to reflect new instructions, we do not update mutable fields that lose history.

src/prefect/agent/agent.py
```diff
@@ -356,7 +356,16 @@ def _deploy_flow_run(
...
             # prevent the flow from starting early
-            start_time = pendulum.parse(flow_run.scheduled_start_time)
+            #
+            # `state.start_time` should be used instead of `scheduled_start_time` for
+            # execution, `scheduled_start_time` is only to record the originally scheduled
+            # start time of the flow run
+            #
+            # There are two possible states the flow run could be in at this point
+            # - Scheduled - in this case the flow run state will have a start time
+            # - Running - in this case the flow run state may not have a start time and we default to now
```
madkinsz, line 366:
```suggestion
            # - Running - in this case the flow run state will not have a start time so we default to now
```

src/prefect/agent/agent.py
```diff
@@ -356,7 +356,16 @@ def _deploy_flow_run(
             # Wait for the flow run's start time. The agent pre-fetches runs that may
             # not need to start until up to 10 seconds later so we need to wait to
             # prevent the flow from starting early
-            start_time = pendulum.parse(flow_run.scheduled_start_time)
+            #
+            # `state.start_time` should be used instead of `scheduled_start_time` for
+            # execution, `scheduled_start_time` is only to record the originally scheduled
```
madkinsz, line 361:
```suggestion
            # `state.start_time` is used instead of `flow_run.scheduled_start_time` for
            # execution; `scheduled_start_time` is only to record the originally scheduled
```

madkinsz (approved):

PR #4578, opened 2021-05-25
`prefect run`: Add automatic quoting for JSON inputs

madkinsz:
## Summary

`prefect run ... --param name=hello` would fail before since strings need quotes. A simple try/repeat pattern gives us a automatic quoting for inputs like this without adding it to things like integers.

## Importance

Usability of `--param` and `--context`

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/run.py
+++ b/src/prefect/cli/run.py
@@ -256,6 +256,18 @@ def load_json_key_values(
         A mapping of keys -> parsed values
     """
     parsed = {}
+
+    def cast_value(value) -> Any:
+        """Cast the value from a string to a valid JSON type; add quotes for the user
+        if necessary
+        """
+        try:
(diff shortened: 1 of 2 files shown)


src/prefect/cli/run.py
```diff
@@ -256,27 +256,34 @@ def load_json_key_values(
         A mapping of keys -> parsed values
     """
     parsed = {}
+
+    def cast_value(value) -> Any:
```
zangell44, line 260: `value: str`

zangell44 (approved):

src/prefect/cli/run.py
madkinsz:
```suggestion
    def cast_value(value: str) -> Any:
```

PR #4567, opened 2021-05-21
Add Github Actions CI test on Windows

madkinsz:
## Summary

We want to move away from AppVeyor and this sets up a comprehensive test using GithubActions

Will require AppVeyor to be removed from the repo after merge

## Changes

- Replaces AppVeyor with testing on Windows via Github Actions
- Tests additional python versions on Windows
- Some tests using `parse_path` were failing and I do not see how they would ever succeed on Windows; skips have been added to the relevant cases.
- Some tests are brittle on the Github Actions Windows runner and have been slightly modified

## Future work

- We could simplify our CI by rolling the Circle CI Unix tests into this workflow with a simple matrix

PR #4584, opened 2021-05-26
Remove forcerm=True from DockerStorage

orf:
## Summary

`forcerm=True` effectively disables the docker caching mechanism, which is quite useful when building images locally using `prefect register --watch`. With this change docker builds are now cached, massively reducing build and push times.

## Changes

Remove `forcerm=True` from the docker build command.

## Importance

Slow builds suck.

## Checklist

This PR:

### DIFF
--- a/changes/pr4584.yaml
+++ b/changes/pr4584.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Keep intermediate docker layers when using Docker Storage to improve caching - [#4584](https://github.com/PrefectHQ/prefect/pull/4584)"
+
+contributor:
+  - "[Tom Forbes](https://github.com/orf)"
--- a/src/prefect/storage/docker.py
+++ b/src/prefect/storage/docker.py
@@ -361,7 +361,6 @@ class Docker(Storage):
                 path="." if self.dockerfile else tempdir,
                 dockerfile=dockerfile_path,
                 tag="{}:{}".format(full_name, self.image_tag),
-                forcerm=True,
                 **self.build_kwargs,
             )
             self._parse_generator_output(output)


orf:
Appveyor fails with:

```
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

🤔

madkinsz:
Hey @orf -- this makes sense to me although I'll have to talk to the team to see why we were doing that in the first place.

Feel free to ignore appveyor; I'm in the middle of replacing it with Github Actions (which confusingly did not run here?) https://github.com/PrefectHQ/prefect/pull/4567

orf:
Thank you! For context I'm evaluating Prefect, but right now a simple flow that uses DockerStorage takes ~5 minutes to build and push. With this change it's down to ~10 seconds which is ideal for rapid iteration.

An alternative (and better) fix is to use buildkit, but unfortunately the Python Docker library doesn't support this yet.

madkinsz:
That makes sense @orf -- I think the historical concern here is that your flow may not be copied into the image if it has changed but from what I understand about layer caching, the pickle blob should be different and invalidate the cache.

This all said, generally we recommend not using `DockerStorage` when you can get away with a file based storage like `GitHub` or `S3` and a `DockerRun` run config which tends to alleviate any slow build concerns.

orf:
We're still early on our Prefect journey, but it seemed that `DockerStorage` was a perfect mix as it allowed us to version our project dependencies alongside the exact flows. If we went with a separate storage (i.e Github + DockerRun) we'd need to wait for an external process to potentially build any changes to the docker image that's required to run the flow. Failure to do so would mean that the flow either launches with an old container (if you're using the `latest` tag) or that your flow can't start because the image hasn't built yet.

It also seemed perfect for local experimentation. But I'll keep your advice in mind and explore other options!

madkinsz:
> We're still early on our Prefect journey, but it seemed that `DockerStorage` was a perfect mix as it allowed us to version our project dependencies alongside the exact flows that run in them. If we went with a separate storage (i.e Github + DockerRun) we'd need to wait for an external process to potentially build any changes to the docker image that's required to run the flow. Failure to do so would mean that the flow either launches with an old container (if you're using the `latest` tag) or that your flow can't start because the image hasn't built yet.
> 
> It also seemed perfect for a local experimentation loop that can be trivially pushed to a k8s executor, and a simple way to segment production flows from experimental ones (different ECR repositories with different permissions).
> 
> But I'll keep your advice in mind and explore other options!

That makes sense! I'm curious where you'll land. You might find https://github.com/PrefectHQ/prefect/discussions/4042 interesting.

Regarding this PR, we're chatting about it internally and I think we'll have to give it a try to make sure it handles edge cases before accepting it.

madkinsz (changes requested):
Can you add a changelog entry? https://github.com/PrefectHQ/prefect/tree/master/changes

orf:
Done!

madkinsz (approved):
Thanks! Looking forward to seeing you around more

PR #4585, opened 2021-05-26
Run Flow API: Update `watch_flow_run` to be an iterator

madkinsz:
## Summary

It seems clunky to pass a callable so instead I'm yielding log objects and the consumer can do as they wish with them without a callback

## Changes

- `watch_flow_run` has a 12hr timeout; makes sense and prevents a test from hanging forever
- Adds a test for `watch_flow_run`
- `watch_flow_run` is of type `Iterator[FlowRunLog]` instead of taking a `Callable[[FlowRunLog], None]` and returning a `FlowRunView`

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/local/agent.py
+++ b/src/prefect/agent/local/agent.py
@@ -154,6 +154,7 @@ class LocalAgent(Agent):
         # dictate the lifecycle of the flow run. However, if the user has elected to
         # show flow logs, these log entries will continue to stream to the users terminal
         # until these child processes exit, even if the agent has already exited.
+        print("HERE!")
         p = Popen(
             [sys.executable, "-m", "prefect", "execute", "flow-run"],
             stdout=stdout,
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -224,16 +224,16 @@ def wait_for_flow_run(
 
     flow_run = FlowRunView.from_flow_run_id(flow_run_id)
 
-    def log_with_flow_run_prefix(log: FlowRunLog):
-        message = f"Flow {flow_run.name!r}: {log.message}"
-        prefect.context.logger.log(log.level, message)
-
-    output_fn = log_with_flow_run_prefix if stream_state else lambda *_, **__: None
-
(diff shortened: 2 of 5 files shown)


src/prefect/agent/local/agent.py
```diff
@@ -154,6 +154,7 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
         # dictate the lifecycle of the flow run. However, if the user has elected to
         # show flow logs, these log entries will continue to stream to the users terminal
         # until these child processes exit, even if the agent has already exited.
+        print("HERE!")
```
madkinsz, line 157:
```suggestion
```

src/prefect/tasks/prefect/flow_run.py
```diff
@@ -224,16 +224,16 @@ def wait_for_flow_run(
...
-        prefect.context.logger.log(log.level, message)
-
-    output_fn = log_with_flow_run_prefix if stream_state else lambda *_, **__: None
-
     if not stream_state and stream_logs:
         warnings.warn("`stream_logs` will be ignored since `stream_state` is `False`")
 
-    return watch_flow_run(flow_run_id, stream_logs=stream_logs, output_fn=output_fn)
+    for log in watch_flow_run(flow_run_id, stream_logs=stream_logs):
+        if stream_state:
```
zangell44, line 231: is it worth passings `stream_state` as an arg to `watch_flow_run`? feels confusing to effectively use `stream_state` and `stream_logs` to restrict logs in different interfaces 

`wait_for_flow_run` is responsible for handling `stream_state` while `watch_flow_run` is PARTIALLY responsible for handling `stream_logs` because `wait_for_flow_run.stream_state` will affect the output

madkinsz: I guess my presumption was `watch_flow_run` will always stream states because otherwise it's silly to watch it but I could be convinced to make state streaming toggle there as well.

madkinsz: I agree this feels a little clunky as is. I could also just change the names of the variables?

PR #4586, opened 2021-05-26
Add the option for setting S3 upload ExtraArgs from outside to allow …

dotannn:
…configuring ACL on the uploaded object and other capabilities

## Summary

Add the option for setting S3 upload ExtraArgs from outside to allow configuring ACL on the uploaded object and other capabilities

## Changes

In S3 storage we added the option of passing extra args for the upload_file() and upload_fileobj() functions. this will allow advanced upload configurations like ACL and others.

## Importance

In our use case, the process that registers the flows in prefect-server is running from our CI/CD env which is in a separate AWS account from the S3 bucket. this leads to an acl issue where the flow objects are locked and owned by the uploader without access for the prefect server account.

## Checklist

This PR:

- [ x] adds new tests (if appropriate) - N/A
- [x ] adds a change file in the `changes/` directory (if appropriate) N/A
- [ x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate) N/A

### DIFF
--- a/src/prefect/storage/s3.py
+++ b/src/prefect/storage/s3.py
@@ -48,10 +48,12 @@ class S3(Storage):
         stored_as_script: bool = False,
         local_script_path: str = None,
         client_options: dict = None,
+        upload_options: dict = None,
         **kwargs: Any,
     ) -> None:
         self.bucket = bucket
         self.key = key
+        self.upload_options = upload_options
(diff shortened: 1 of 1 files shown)


madkinsz (approved):

madkinsz:
This LGTM but needs to pass CI

madkinsz (approved):

PR #4591, opened 2021-05-27
docs: update `prefect run` reference

alexkreidler:
Updates the documentation to replace an out of date command `prefect run server/cloud` with the new `prefect run flow`.


madkinsz:
Hey @alexkreidler ! Thanks for contributing. I'm actually rewriting this entirely at #4434 which should be released very soon so I'm not sure it's worth changing here. If we cut a release before then, I'll make sure this change gets in.

Issue #4537, opened 2021-05-17
LocalDaskExecutor multiprocessing regression due to dask-core==2021.4.0


kpweiler:
## Description
When using `LocalDaskExecutor`, tasks that can be run in parallel are not run in parallel if the installed `dask-core` library is `2021.4.0` or later.

## Expected Behavior
Parallel tasks should be executed in parallel

## Reproduction
```
from prefect import task, Flow
from prefect.executors import LocalDaskExecutor
import time

@task
def run_thing1():
    time.sleep(5)
    return True

@task
def run_thing2():
    time.sleep(5)
    return True

if __name__ == "__main__":
    with Flow("test_flow", executor=LocalDaskExecutor(scheduler="processes")) as flow:
        run_thing1()
        run_thing2()

    flow.run()
```

When this is executed with `dask-core<=2021.3.1`, the output is as follows:
```
[2021-05-16 20:27:11-0500] INFO - prefect.FlowRunner | Beginning Flow run for 'test_flow'
[2021-05-16 20:27:11-0500] INFO - prefect.TaskRunner | Task 'run_thing1': Starting task run...
[2021-05-16 20:27:11-0500] INFO - prefect.TaskRunner | Task 'run_thing2': Starting task run...
[2021-05-16 20:27:16-0500] INFO - prefect.TaskRunner | Task 'run_thing1': Finished task run for task with final state: 'Success'
[2021-05-16 20:27:16-0500] INFO - prefect.TaskRunner | Task 'run_thing2': Finished task run for task with final state: 'Success'
[2021-05-16 20:27:16-0500] INFO - prefect.FlowRunner | Flow run SUCCESS: all reference tasks succeeded
```

When this is run with `dask-core>=2021.4.0`, however, the output is as follows:
```
[2021-05-16 20:26:21-0500] INFO - prefect.FlowRunner | Beginning Flow run for 'test_flow'
[2021-05-16 20:26:22-0500] INFO - prefect.TaskRunner | Task 'run_thing1': Starting task run...
[2021-05-16 20:26:22-0500] INFO - prefect.TaskRunner | Task 'run_thing2': Starting task run...
[2021-05-16 20:26:27-0500] INFO - prefect.TaskRunner | Task 'run_thing1': Finished task run for task with final state: 'Success'
[2021-05-16 20:26:27-0500] INFO - prefect.TaskRunner | Task 'run_thing2': Finished task run for task with final state: 'Success'
[...]

PR #4602, opened 2021-06-01
Fix small-flow parallelism issues with multiprocess `LocalDaskExecutor`

madkinsz:
## Summary

dask >= 2021.04.0 adds task submission batching in the multiprocessing case
to offset performance concerns from switching to a concurrent.Futures based
process pool. Since we are using our own pool to handle cancellation robustly,
we do not gain anything from the batched submission and the default task
submission batch size of 6 makes users think their flows will not run in
parallel. When/if we switch to using the new futures based multiprocessing in
dask, we should remove this to retain performance

## Changes

- Sets `chunksize=1` for the process based `LocalDaskExecutor`

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/4537

## Checklist

This PR:

### DIFF
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -4,7 +4,7 @@ import uuid
 import sys
 import weakref
 from contextlib import contextmanager
-from typing import Any, Callable, Iterator, TYPE_CHECKING, Union, Optional
+from typing import Any, Callable, Iterator, TYPE_CHECKING, Union, Optional, Dict
 
 from prefect import context
 from prefect.executors.base import Executor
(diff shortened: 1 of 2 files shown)


madkinsz:
I've tested this with an older version of dask and noted there is no issue setting this key.

madkinsz:
Note that this dask version is not yet supported but users are using it... we can still patch support for it and consider bumping our supported dask version in #4604

src/prefect/executors/dask.py
```diff
@@ -614,14 +616,24 @@ def cull2(dsk, keys):  # type: ignore
...
-            config = {"optimization.fuse.active": False}
+            config["optimization.fuse.active"] = False
         else:
-            config = {}
 
             @contextmanager
             def patch() -> Iterator[None]:
                 yield
 
+        # Patch around https://github.com/PrefectHQ/prefect/issues/4537
```
zangell44, line 626: any reason not to slide this up into the previous `if` block?

madkinsz: I want to maintain these "patches" separately since they are focusing on fixing very different upstream issues. This will make the diff easier to parse when we remove one.

zangell44 (approved):

PR #4605, opened 2021-06-01
Cleanly handle unpicklable exceptions in tasks

madkinsz:
## Summary

While using the `DaskExecutor`, if a task fails with an exception that cannot be pickled by `cloudpickle`, Dask will fail to send the failure back to the flow runner as the `Failed` state containing the exception as a result cannot be pickled. This PR adds more robust handling to `State` pickling to handle exception pickle failures.

## Changes

- Adds `State.__getstate__` which will cast unpickable exception results to a string with helpful information

## Importance

Prefect should be able to report task failure cleanly

## Checklist

This PR:

### DIFF
--- a/src/prefect/engine/state.py
+++ b/src/prefect/engine/state.py
@@ -86,21 +86,19 @@ class State:
     def __getstate__(self) -> dict:
         data = self.__dict__
 
-        try:
-            cloudpickle.dumps(self.result)
-        except TypeError as exc:
-            if "cannot pickle" not in str(exc):
-                # Exit immediately if not a pickle failure, this error will likely occur
-                # downstream when this is pickled
-                return data
-
-            # Exception results will be cast to the repr so we can still fail cleanly
-            if isinstance(self.result, Exception):
-                data["_result"].value = (
-                    f"The following exception could not be pickled due to {exc!r}: "
-                    + repr(self.result)
-                )
-            # All other types will fall through to the downstream pickler
+        # For 'Exception' results, we will check if it can be pickled successfully and
+        # on failure convert the exception to a string instead. This allows the engine
+        # to handle tasks with unpickable exceptions successfully
+        if isinstance(self.result, Exception):
+            try:
+                cloudpickle.dumps(self.result)
+            except TypeError as exc:
+                if "cannot pickle" in str(exc):
+                    self.result = (
+                        f"The following exception could not be pickled due to {exc!r}: "
+                        + repr(self.result)
+                    )
+                    return self.__getstate__()
 
         return data
 
--- a/src/prefect/engine/state.py
+++ b/src/prefect/engine/state.py
@@ -12,6 +12,7 @@ execution. During execution a run will enter a `Running` state. Finally, runs be
 """
 import datetime
 from typing import Any, Dict, List, Optional, Type, Mapping
+from pickle import PicklingError
 
 import pendulum
 import cloudpickle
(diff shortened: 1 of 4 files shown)


src/prefect/engine/state.py
```diff
@@ -82,6 +84,40 @@ def __eq__(self, other: object) -> bool:
...
+            # Check for an exception during pickling of the 'Exception' type result
+            pickle_exc: Optional[Exception] = None
+            try:
+                cloudpickle.dumps(self.result)
+            except TypeError as exc:
+                if "cannot pickle" in str(exc):
+                    pickle_exc = exc
+            except PicklingError as exc:
+                pickle_exc = exc
+
```
madkinsz, line 108: We may want to do...
```suggestion
            except PicklingError as exc:
                pickle_exc = exc
            except:
                pass
                # Any other exception will be ignored and we'll let the pickler fail
                # downstream
```

madkinsz: Otherwise we're raising an exception early / if they're not using cloudpickle they could still succeed

madkinsz: Nevermind... since the exception message changes across python versions and the exception types can vary we'll just catch any exception here as the pickle exception in https://github.com/PrefectHQ/prefect/pull/4605/commits/e44997652fbbe580a661b559c299284913b46dd6

tests/engine/test_state.py
```diff
@@ -676,3 +679,68 @@ def test_n_map_states():
...
+        foo: int = 1
+
+    state = State(message="message", result=Data(bar="bar"))
+    new_state = cloudpickle.loads(cloudpickle.dumps(state))
+
+    assert state.message == new_state.message  # message not included in __eq__
+    assert state == new_state
+
+
+def test_state_pickle_with_unpicklable_result_raises():
```
cicdw, line 697: test name here is misleading - i think you mean 
```suggestion
def test_state_pickle_with_unpicklable_attribute_raises():
```

src/prefect/engine/state.py
```diff
@@ -82,6 +83,37 @@ def __eq__(self, other: object) -> bool:
...
+        # on failure convert the exception to a string instead. This allows the engine
+        # to handle tasks with unpickable exceptions successfully. We must explicitly
+        # ignore state signals here or we will get stuck in a recursive loop because
+        # signals are exceptions that contain a state with `_result` set to the signal
+        if isinstance(self.result, Exception) and not isinstance(
+            self.result, prefect.engine.signals.PrefectStateSignal
+        ):
+
+            # Check for an exception during pickling of the 'Exception' type result
+            pickle_exc: Optional[Exception] = None
```
cicdw, line 99: The need for this variable + all the logic below can be brought inside the `Exception` catch

cicdw (changes requested):
I think this makes a lot of sense!  Left a few superficial comments

src/prefect/engine/state.py
madkinsz: Ah yeah now that I've refactored it to handle all exceptions that make sense. Will do.

tests/engine/test_state.py
madkinsz: Hm. I'm specifically referring to the `State.result` here though not any ole attribute

cicdw: But your PR prevents the raise when the unpickleable thing is `State.result` and this tests that the raise does occur for `State.x`?

madkinsz: I guess here I could actually just set the result to `RLock()` instead of using a dataclass which simplifies the test

madkinsz: I'll update the test to be simpler; not sure why I thought I needed to make it a class.

This test covers `State.result.x` being unpickable; really just needs to cover `State.result` being unpickable and not an `Exception` type

madkinsz: Good catch :)

madkinsz: https://github.com/PrefectHQ/prefect/pull/4605/commits/478cc1abeeb1933e83f1d7571b4d84644a1779c5

zangell44 (dismissed):

cicdw (approved):

Issue #4049, opened 2021-02-02
Some CLI commands do not work if the server is running on a non-default port


sm-Fifteen:
## Description
A number of prefect components behave in usual ways if the local core server isn't running on `localhost:4200`. Aside from the UI not being configured automatically, this is a problem for all the CLI commands that do not let you set an API address, such as `prefect server create-tenant` and `prefect register flow`. 

## Expected Behavior
I would assume `prefect backend` would let you specify a port that other commands are expected to use to connect to the local prefect server instead of only the default 4200, or at least that the base `prefect` command would have a global `--api` option that applies to all sub-commands. At any rate it's rather strange that `prefect server start` even lets you set a different port as a CLI option in the first place if the official server management tools have no way to be informed of what port to use.

## Reproduction

In one window, start the server using a non-standard port configuration.
```
prefect server start --ui-port=8888 --server-port=8080
```

In another window, try running a command that needs to communicate with the server.
```
`prefect server create-tenant --name default --slug default`
```

## Environment
```json
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Windows-10-10.0.17134-SP0",
    "prefect_backend": "server",
    "prefect_version": "0.14.5",
    "python_version": "3.8.2"
  }
}
```

joshmeek:
@sm-Fifteen after you run `prefect backend server` you can set the host of your api by setting the config via:

```
[server]
host = "http://localhost"
port = "4200"
endpoint = "${server.host}:${server.port}"
```

or through the environment variable `PREFECT__SERVER__ENDPOINT=your_host:port`

sm-Fifteen:
> @sm-Fifteen after you run `prefect backend server` you can set the endpoint of your server api by setting the config [...]

I'm assuming you mean in `~/.prefect/backend.toml`, the file created by `prefect backend server`? That seems to work, so it at least solves by issue, though it would still be convenient if `prefect backend` could take parameters to let it fill out that part of the configuration automatically.

It would also make it easy to [introduce early on in the core "Getting started" tutorial](https://docs.prefect.io/core/getting_started/installation.html#running-the-local-server-and-ui) rather than [relegating it to the orchestration FAQ](https://docs.prefect.io/orchestration/faq/config.html#connecting-to-a-different-api-endpoint) where that part of the configuration seems to currently be documented.

> Before running the server for the first time, run `prefect backend server`
>  
[...]

PR #4614, opened 2021-06-02
Correct Client initialization when backend = server

mwheeler01:
## Summary

This PR corrects an issue [mentioned here](https://github.com/PrefectHQ/prefect/issues/4049#issuecomment-847407701) where the `Client` class failed to initialize correctly if the backend was set to "server"

## Changes

- Reads the configuration for `backend` in order to initialize the `Client`'s attributes properly
- Provides a more descriptive name for the "endpoint" attribute in the `Client` class
- Raises an error if `backend` is set to neither cloud nor server
- Warns if the base endpoint URL contains a path that does not end in a trailing slash, to ensure proper behavior when passed into the `Client`'s request methods

## Importance

Among other things, this PR allows the `create-tenant` command to behave correctly when Prefect is configured to work with a non-local deployment of Prefect Server.

## Checklist

This PR:


madkinsz:
Following an offline discussion,

The config uses
```toml
[cloud]
api = "${${backend}.endpoint}"
endpoint = "https://api.prefect.io"
graphql = "${cloud.api}/graphql"
```

Notably, the `cloud.api` setting references `backend` so the agent actually does reference the `server.endpoint` correctly in most cases.

I'll address any other issues in #4643. Accepting this would create merge conflicts with that work.

Thanks for the effort though.

PR #4623, opened 2021-06-07
Add new task for Prometheus pushgateway

davzucky:
## Summary

This is a new task group for prometheus. Prometheus as a job end point named pushgateway. This can allow you to send inside metrics about your flow like data size and monitor it

## Changes

This is a new set of task related to prometheus

## Importance

This PR add a new group of task that can be used by other consumers to the core prefect

## Checklist

This PR:


src/prefect/tasks/prometheus/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+Tasks for interacting with Prometheus. The main task are using pushgateway 
+"""
+try:
+    from prefect.tasks.prometheus.pushgateway import PushGaugeToGateway, PushMode
+except ImportError as err:
+    print(err)
```
zangell44, line 7: We can remove this

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,137 @@
...
+    Gauge,
+    pushadd_to_gateway,
+    push_to_gateway,
+)
+from typing import Dict, List, Optional
+from prefect import Task
+from prefect.utilities.tasks import defaults_from_attrs
+
+
+class PushMode(Enum):
```
zangell44, line 13: are there more than two push modes?
the `Enum` seems a bit heavy, since users will have to import and configure `PushMode`

if it's just `Push` and `PushAdd`, can we replace this with a boolean flag?

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,137 @@
...
+    @defaults_from_attrs(
+        "pushgateway_url",
+        "counter_name",
+        "counter_description",
+        "job_name",
+        "grouping_key",
+    )
+    def run(
+        self,
+        values: List[float] = [],
```
zangell44, line 59: we want to avoid mutable default arguments here https://docs.python-guide.org/writing/gotchas/
I don't see a need for default args for `values` and `labels`, they're both required args

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if not counter_name:
+            raise ValueError("You need to provide the counter_name parameter.")
+
+        if not counter_description:
+            counter_description = counter_name
+
+        if len(values) != len(labels):
+            raise ValueError(
+                f"Values and labels need to have the same lenght. nb Values {len(values)}, "
```
zangell44, line 102: what does `nb` mean?

can we update the comment to clarify that we're logging the length of values and labels?

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,137 @@
...
+        "counter_name",
+        "counter_description",
+        "job_name",
+        "grouping_key",
+    )
+    def run(
+        self,
+        values: List[float] = [],
+        labels: List[Dict[str, str]] = [],
+        action: PushMode = PushMode.Push,
```
zangell44, line 61: does it make sense to have the option to provide `action` at `init` or `run` time too?

zangell44 (changes requested):
Looking good, thanks for contributing @davzucky! I left a few comments.

src/prefect/tasks/prometheus/pushgateway.py
davzucky: Sure, I can. I was thinking as well as having to task, one for each mode as it is a different call as well in the API. What do you think ?

src/prefect/tasks/prometheus/pushgateway.py
davzucky: make sense, will remove them

src/prefect/tasks/prometheus/pushgateway.py
davzucky: it could true. I think having two tasks may be better as well. What do you think?

src/prefect/tasks/prometheus/__init__.py
davzucky: sure will clean it

src/prefect/tasks/prometheus/pushgateway.py
zangell44: I think using the default `action` from `init` would make the most sense

src/prefect/tasks/prometheus/pushgateway.py
zangell44: I think one task is okay for both, but I'm not as familiar with Prometheus,

davzucky:
I updated the code and split the task into two. one for push and one for pushadd to make it clear what they are doing

docs/outline.toml
```diff
@@ -499,6 +499,11 @@ title = "Postgres Tasks"
 module = "prefect.tasks.postgres"
 classes = ["PostgresExecute", "PostgresExecuteMany", "PostgresFetch"]
 
+[pages.tasks.prometheus]
+title = "prometheus Tasks"
+module = "prefect.tasks.prometheus"
+classes = ["PushGaugeToGateway"]
```
madkinsz, line 505:
```suggestion
classes = ["PushGaugeToGateway", "PushAddGaugeToGateway"]
```

src/prefect/tasks/prometheus/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+Tasks for interacting with Prometheus. The main task are using pushgateway 
+"""
+try:
+    from prefect.tasks.prometheus.pushgateway import (
+        PushGaugeToGateway,
+        PushAddGaugeToGateway,
+    )
+except ImportError as err:
+    print(err)
```
madkinsz, line 10:
```suggestion
```

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,203 @@
...
+        if not counter_name:
+            raise ValueError("You need to provide the counter_name parameter.")
+
+        if not counter_description:
+            counter_description = counter_name
+
+        if len(values) != len(labels):
+            raise ValueError(
+                f"Values and labels need to have the same lenght. Values lenght is {len(values)}, "
+                f"labels lenght is {len(labels)}."
```
madkinsz, line 84:
```suggestion
                f"Values and labels need to have the same length. Values length is {len(values)}, "
                f"labels length is {len(labels)}."
```

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,203 @@
...
+        values: List[float],
+        labels: List[Dict[str, str]],
+        pushgateway_url: str = None,
+        counter_name: str = None,
+        counter_description: str = None,
+        grouping_key: Optional[List[str]] = None,
+        job_name: str = None,
+    ) -> None:
+        """
+        Push a Gauge to prometheus [PushGateway](ttps://prometheus.io/docs/practices/pushing/).
```
madkinsz, line 52:
```suggestion
        Push a Gauge to prometheus [PushGateway](https://prometheus.io/docs/practices/pushing/).
```

changes/pr4623.yaml
```diff
@@ -0,0 +1,23 @@
...
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+task:
+  - "Add new prometheus task to push to gateway - [#4623](https://github.com/PrefectHQ/prefect/pull/4623)"
```
madkinsz, line 20: The Prometheus docs seem to recommend against this pattern

> Usually, the only valid use case for the Pushgateway is for capturing the outcome of a service-level batch job. A "service-level" batch job is one which is not semantically related to a specific machine or job instance (for example, a batch job that deletes a number of users for an entire service).

But here we have a flow/task run which is a specific job instance. Thoughts? @mashun4ek

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,203 @@
...
+        This will unpack the values and labels to create all the a collector that will be pushed
+        to pushgateway.
+
+        Args:
+            - values (List[float]): List of the values to push
+            - labels (List[Dict[str, str]]): List of the labels to push attached to the values
```
madkinsz, line 58: Aren't labels optional? Is there a reason to make this required? https://github.com/prometheus/client_python#gauge

src/prefect/tasks/prometheus/pushgateway.py
```diff
@@ -0,0 +1,203 @@
...
+                f"Values and labels need to have the same lenght. Values lenght is {len(values)}, "
+                f"labels lenght is {len(labels)}."
+            )
+
+        if len(values) == 0 or len(labels) == 0:
+            return
+
+        grouping_key_values = (
+            {key: labels[0][key] for key in grouping_key} if grouping_key else {}
+        )
```
madkinsz, line 92: Grouping keys pull from the first label? This isn't clear from the docstring. Is this mirroring prometheus behavior?

changes/pr4623.yaml
madkinsz: I'm also a bit confused about how this is supposed to be used after looking at the examples in  https://github.com/prometheus/client_python#gauge

Can you show how this would be used in a flow to push metrics of interest?

davzucky: We are using that today to push metrics about the quality of our input data.

Imagine you have a workflow with a task that extracts the quality of the data, like distribution, expected values, missing values. We are pushing this result for each job to prometheus to monitor our data into the flow.

We are using that as well to send data about our ML model performance during training. So we are able to quantify how good it is.

This task are not mean to monitor prefect. 
let me add an example into the code.

mashun4ek: > The Prometheus docs seem to recommend against this pattern
> 
> > Usually, the only valid use case for the Pushgateway is for capturing the outcome of a service-level batch job. A "service-level" batch job is one which is not semantically related to a specific machine or job instance (for example, a batch job that deletes a number of users for an entire service).
> 
> But here we have a flow/task run which is a specific job instance. Thoughts? @mashun4ek

Yes, you're right about the `push` pattern which is against Prometheus core, but in some cases, we can't use the pull approach and have to use Pushgateway. In this case, short-lived flows can't expose metrics for Prometheus to scrape, so instead, those flows will push metrics for Pushgateway. Thank you for contribution @davzucky! And I agree, a small example would be great!

davzucky: I updated the code and added a simple code that push to pushgateway the percentage of null value for a dataframe. This is something we are already using to provide data quality metric to our validation workflow

src/prefect/tasks/prometheus/pushgateway.py
davzucky: Indeed. This is true. I updated the function to support multiple grouping keys and send all of them.

src/prefect/tasks/prometheus/pushgateway.py
davzucky: yes they are, however you need to match the values. so you would need to set empty dict in this case

src/prefect/tasks/prometheus/pushgateway.py
davzucky: fixed

docs/outline.toml
davzucky: Thank you for that

davzucky:
Any update about this PR ?

madkinsz:
Hey @davzucky -- we'll try to get this a final round of review before the week is over.

madkinsz (approved):
Wonderful! Thanks for your patience @davzucky. We'll get this in the next release on Tuesday.

davzucky:
Thank you @madkinsz for merging the change

PR #4641, opened 2021-06-09
Add waiting_interval parameter to StartFlowRun

peterroelants:
## Summary

Add `waiting_interval` parameter to `StartFlowRun` to define the polling interval when waiting for the flow run to finish.

## Changes

- `waiting_interval` parameter added to `StartFlowRun` initializer.
- Adds a unit test for `StartFlowRun(wait=True).run()`

## Importance

The [default (and fixed) 10 seconds wait](https://github.com/PrefectHQ/prefect/blob/1867f205405e1f68b828ce9dcf00bbd4969903bc/src/prefect/tasks/prefect/flow_run.py#L198) has been slowing down my flow unit tests with flows depending on `StartFlowRun` Tasks. By making this a variable parameter I can speed up my unit tests significantly. There might be other cases where we want to speed up the waiting polling frequency.

## Checklist

This PR:

### DIFF
--- a/changes/startflowrun_wait_interval_parameter.yaml
+++ b/changes/startflowrun_wait_interval_parameter.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Add `poll_interval` parameter to `StartFlowRun` to define the polling interval when waiting for the flow run to finish - [#4641](https://github.com/PrefectHQ/prefect/pull/4641)"
+
+contributor:
+  - "[Peter Roelants](https://github.com/peterroelants)"
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -1,5 +1,6 @@
 import time
 import datetime
+from datetime import timedelta
 import warnings
 from typing import Any
 from urllib.parse import urlparse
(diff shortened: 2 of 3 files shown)


peterroelants:
Some design choices:
* The new parameter name is now `waiting_interval`. Alternative names:
  - `waiting_period`
  - `wait_for_finish_interval`
  - Any variations or other suggestions can also be applied.
* The parameter is added to `StartFlowRun.__init__`, because this is also where the `wait` parameter is. Alternatively this parameter could also be added to `.run()`.

Btw, I think there is something wrong with the code style checks, it is complaining about "Library stubs not installed". Please review.

madkinsz:
Hey @peterroelants, thanks for being so active lately :)

The type stubs are being addressed - mypy released a breaking change yesterday.

Are your unit tests actually starting sub-flows or have you mocked the Prefect API? If you're not actually hitting the API, the best practice here may be to just mock `time.sleep` to be a no-op which will speed-up your tests without introducing a setting that is not generally needed. I'm hesitant to accept this because it's likely to result in people lowering the interval unnecessarily, spamming our API. 10 seconds is already pretty short in the lifetime of a flow.

peterroelants:
> Are your unit tests actually starting sub-flows or have you mocked the Prefect API?

They are actually starting sub-flows and run against Prefect Server as part of integration testing.

peterroelants:
> I'm hesitant to accept this because it's likely to result in people lowering the interval unnecessarily, spamming our API.

Would you prefer adding a lower limit to the interval parameter?

madkinsz:
> I'm going to check if there is some way for me to mock time.sleep

You can definitely mock `time.sleep` to sleep for a fixed interval and just ignore the passed argument.

> Would you prefer adding a lower limit to the interval parameter?

I chatted with the team and this seems reasonable. I'd prefer

- Changing the name to `poll_interval`
- Setting a lower limit of 3 seconds

peterroelants:
> I chatted with the team and this seems reasonable. I'd prefer
> 
> * Changing the name to `poll_interval` or `poll_seconds`
> * Setting a lower limit of 3 seconds

I added the suggested changes, squashed everything in a new commit, and pushed the changes to this branch.

madkinsz (approved):

Issue #4600, opened 2021-05-28
Simplify core auth model with new API keys


jcrist:
Prefect cloud recently updated its auth model to use "API keys" instead of "API tokens". Besides a rename, there's a few key differences:

- API keys are specific to a user or service account
- API keys have no concept of "scope" (no more differentiating between RUNNER and TENANT scope tokens)
- API keys can have a default tenant, but if the user they're associated with has multiple tenants then the token can work cross-tenant

Currently we're using API keys in places where API tokens are used (no real changes in prefect core), but the above changes would let us simplify our auth model with some work.

### Current core auth model

Currently users are asked to create 2 API tokens and configure them as:

- `prefect auth login --token <TOKEN>`: for a user's personal token. This stores the token and current tenant in `~/.prefect/client/*/settings.toml`.
- `cloud.agent.auth_token`: a token to use for the agent. Previously this was a RUNNER scope token, now the docs say to create a service account "API key"

There's also `cloud.auth_token` which takes a TENANT scoped token. Usually a user never configures this, and it's forwarded by an agent during a flow run. But sometimes a user does want to configure this when e.g. making API requests from CI.

The user api tokens were a bit confusing - since they were associated with the user, on init the `prefect.Client` would convert the token + current tenant into a `TENANT` scoped token to use for all operations.

### Proposed auth model

Asking users to create and configure two tokens (or "keys" now) when running locally shouldn't be necessary. Since all API keys have the same capabilities we should be able to configure one and have the client and local agent work the same. Further, cloud now supports sending tenant info as part of the header, so there should be no reason to convert a personal multi-tenant API key into one scoped for a specific tenant. The trick is doing this in a backwards compatible way.

The two bits of information we care about for all API requests are `api_key` and current tenant. I propose determining these from a few different sources, with the following priority:

- `prefect.config.cloud.api_key` and `prefect.config.cloud.tenant_id`: two new config fields holding the api key and current tenant id respectively. If the api key is present but the tenant id is not, then `tenant_id` is `None` and the default tenant associated with the api key will be used by cloud's backend. This works the same as the old `prefect.config.cloud.auth_token` field in this respect, but supports optionally configuring the tenant to use as well.
[...]

PR #4643, opened 2021-06-09
Update `Client` and `prefect auth` CLI for API keys

madkinsz:
Docs at https://deploy-preview-4643--prefect-docs.netlify.app/orchestration/concepts/api_keys.html
## Summary

This PR adds first-class handling for API keys as described fully in #4600 

## Changes

- Deprecates all existing API token handling
- Adds new CLI commands for API keys
- Adds backward compatible API key handling to existing `prefect auth login/logout` commands

Previously, logging out with an API token would just reset your tenant and access token but leave the API token on disk. Now, we retain that behavior the first time the command is called but if you call it a _second_ time we will delete the API token. This allows users to easily remove their API token in favor of an API key.

- Agents will use an available API key as detected by the `Client`

This will not detect tokens because they are stored at a different path. However, this is new behavior for agents. This means that if you have logged in locally you may start any kind of agent and the agent will use your API key to query for flow runs then pass it _to_ the flow run for execution as well. 

- Agents will pass the tenant id alongside the API key to flow run jobs

When using the non-default tenant associated with an API key, we need to set the tenant on the flow run job or the flow run will instantiate a client with the wrong tenant id to report to Cloud. If the default tenant or a token is being used, this environment variable is just left blank.

- `prefect agent <type> start/install` take `--key` and `--tenant-id` options

Running the install command for an agent will not pull an API key from your local cache to match previous behavior (and you should be using a SA for installs anyway). 

- __Warning__: This PR does _not_ add support for API tokens to deprecated objects, notably flow "Environments".

## Importance

Since we've moved to API keys in the UI/backend, we need to get Core up to date. 

## Checklist

This PR:


src/prefect/client/client.py
```diff
@@ -80,34 +80,158 @@ class Client:
...
-        self._active_tenant_id = None
+        self._tenant_id = None
         self._attached_headers = {}  # type: Dict[str, str]
         self.logger = create_diagnostic_logger("Diagnostics")
 
-        # store api server
-        self.api_server = api_server or prefect.context.config.cloud.get("graphql")
+        # Note the default is `cloud.api` which is `cloud.endpoint` or `server.endpoint`
+        # depending on the value of the `backend` key
+        self.api_server = api_server or prefect.context.config.cloud.api
```
madkinsz, line 115: I made this change so that Agent clients and default clients would use the same api server but this breaks existing behavior and loading existing credentials -- I'll have to sort out a better way to do this :(

madkinsz: (all the failing tests are from this change)

madkinsz: I've fixed this by returning to the original default if an API key is not set

src/prefect/cli/auth.py
```diff
@@ -285,3 +444,154 @@ def revoke_token(id):
         return
 
     click.secho("Token successfully revoked", fg="green")
+
+
+@auth.command(hidden=True)
+@click.option("--name", "-n", required=True, help="A name to associated with the key")
```
mashun4ek, line 450:
```suggestion
@click.option("--name", "-n", required=True, help="A name to associate with the key")
```

src/prefect/client/client.py
```diff
@@ -80,34 +80,187 @@ class Client:
...
-        - api_server (str, optional): the URL to send all GraphQL requests
-            to; if not provided, will be pulled from `cloud.graphql` config var
+        - api_server (str, optional): the URL to send all GraphQL requests to; if not
+            provided, will be pulled from the current backend's `api` config
+            variable
+        - api_key (str, optional): a Prefect Cloud API key. If not provided, loaded
+            from `config.cloud.api_key` or from the on disk cache from the
+            `prefect auth` CLI
+        - tenant_id (str, optional): the Prefect tenant to use. If not provided, loaded
+            from `config.cloud.tenant_id` or the the on disk cache from the
```
mashun4ek, line 90:
```suggestion
            from `config.cloud.tenant_id` or the on disk cache from the
```

src/prefect/cli/auth.py
```diff
@@ -22,117 +34,234 @@ def auth():
...
+        )
+
+        # Clear the key and tenant id then write to the cache
+        client.api_key = ""
+        client._tenant_id = ""
+        client.save_auth_to_disk()
+
+        click.secho("Logged out of Prefect Cloud", fg="green")
+
+    elif client._api_token:
```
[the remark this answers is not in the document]
madkinsz: Naw if the client doesn't have an API token then they're not logged in with one at all and we can just fall through to the `else`.

docs/orchestration/agents/overview.md
```diff
@@ -54,42 +54,17 @@ LocalAgent().start()
...
-```bash
-prefect agent <AGENT TYPE> start --token <SERVICE_ACCOUNT_API_KEY>
-```
-
-:::
-
-::: tab "Prefect Config"
-
-```toml
-# ~/.prefect/config.toml
```
zangell44, line 77: do environment variables or `config.toml` work for agents anymore? I don't see them anywhere in the updated docs

docs/orchestration/concepts/api.md
```diff
@@ -32,14 +32,14 @@ Prefect Core includes a Python client for interacting with the API. The Python c
...
-If using Prefect Core's server, no authentication is required. This means the Python client can be used immediately without any extra configuration:
+If using Prefect Server, no authentication is required. This means the Python Client can be used immediately without any extra configuration:
 
 ```python
 import prefect
 client = prefect.Client()
 
 client.graphql(
-    {
+    {`
```
zangell44, line 42: is this deliberate?

docs/orchestration/concepts/api.md
```diff
@@ -51,40 +51,51 @@ client.graphql(
...
 
-Once provisioned with an API key, the Cloud Client can query for available tenants and login to those tenants. In order to query for tenants, call:
+```python
+client.get_default_tenant()
+```
+
+The tenant id can be changed on an existing client, but requests will fail if the API key is not valid for that tenant:
 
 ```python
-client.get_available_tenants()
```
zangell44, line 74: is there an easy way to get available tenants in the new system? as a user it may be very difficult to find my tenant ids

docs/orchestration/agents/overview.md
madkinsz: They do -- they're just the same as normal auth so `PREFECT__CLOUD__API_KEY` and it's covered in the API key docs. I provide a link in the agent auth section

https://deploy-preview-4643--prefect-docs.netlify.app/orchestration/agents/overview.html#authentication

zangell44: (specifically for agents)

docs/orchestration/concepts/api.md
madkinsz:
```suggestion
    {
```

Probably not lol

docs/orchestration/concepts/api.md
madkinsz: `client.get_available_tenants()` and `prefect auth list-tenants` both still work wonderfully. This was removed from this section as this was previously a _requirement_ to the login process e.g.

- Set auth token
- List tenants
- Login to tenant

Now it's just

- Set API key

- (Optional) Also set a tenant

docs/orchestration/faq/debug.md
```diff
@@ -12,16 +12,12 @@ The most likely culprit when a flow is stuck in a Scheduled state is agent misco
 Here are some tips for debugging your agent:
 
 1. Verify that your agent is running on your desired platform
-2. If using Prefect Cloud as your API, confirm that the API token given to the agent is scoped to the same tenant as your flow
+2. If using Prefect Cloud as your backend API, confirm that the API key given to the agent is set to the same tenant as the flow
 
+```bash
+$ PREFECT__CLOUD__API_KEY="YOUR-KEY" prefect get flows
+# If you do not see your flow then there is a tenant mismatch
```
zangell44, line 19: what do I do if there is a tenant mismatch?

docs/orchestration/agents/overview.md
madkinsz: You should set a separate auth key for agents via the environment anymore. The old environment variable will be passed through as a token to the client if set but if your client has an API key configured as well there will be a warning and the token will be ignored. If you do not have an API key configured globally and you set the old environment variable with either a token or API key, it will authenticate as you'd expect.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -802,6 +817,8 @@ def generate_deployment_yaml(
         agent_env[3]["value"] = image_pull_secrets or ""
         agent_env[4]["value"] = str(labels)
         agent_env[11]["value"] = backend
+        agent_env[13]["value"] = key
```
zangell44, line 820: why are these indexed at 13 and 14?

docs/orchestration/agents/overview.md
zangell44: ahhhh okay i see now we removed that logic

src/prefect/agent/agent.py
```diff
@@ -131,8 +128,9 @@ def __init__(
         agent_address: str = None,
         no_cloud_logs: bool = None,
     ) -> None:
-        # Load token and initialize client
+        # Load token for backwards compatibility
```
zangell44, line 131: what about backwards compatibility for config.cloud.agent.api_key?

tests/cli/test_agent.py
```diff
@@ -209,26 +212,48 @@ def test_agent_start_fails(monkeypatch, cloud_api):
     assert "TEST is not a valid agent" in result.output
 
 
+@pytest.mark.parametrize("use_token", [False, True])
```
zangell44, line 215: makes total sense that you can stack these decorators, had just never seen it done, neat!

src/prefect/agent/kubernetes/agent.py
madkinsz: I put them at the end to make as few changes as I could and there are some that are not set.

src/prefect/agent/agent.py
madkinsz: `config.cloud.agent.api_key` was never a thing

tests/cli/test_agent.py
madkinsz: Ah you can stack them _or_ put them in one which leads to pretty different behavior. When stacked, they are "multiplied" so all of the different combinations are easily covered.

docs/outline.toml
```diff
@@ -41,7 +41,7 @@ commands = ["local", "docker", "kubernetes", "ecs", "fargate"]
 [pages.cli.auth]
 title = "auth"
 module = "prefect.cli.auth"
-commands = ["login", "logout", "list_tenants", "switch_tenants", "create_token", "list_tokens", "revoke_token"]
+commands = ["login", "logout", "list_tenants", "switch_tenants", "create_token", "list_tokens", "revoke_token", "create_key", "list_keys", "revoke_key", "status"]
```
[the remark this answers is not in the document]
madkinsz:
```suggestion
commands = ["login", "logout", "list_tenants", "switch_tenants", "create_key", "list_keys", "revoke_key", "status", "create_token", "list_tokens", "revoke_token"]
```

Makes sense, I wasn't sure if this determined the order in the docs

src/prefect/cli/auth.py
```diff
@@ -22,135 +38,265 @@ def auth():
...
+    """
+    if not key and not token:
+        raise TerminalError("You must supply an API key or token!")
+
+    if key and token:
+        raise TerminalError("You cannot supply both an API key and token")
 
-        if not success_login:
-            raise AuthorizationError
+    abort_on_config_api_key(
```
zangell44, line 116: could we just auto remove the config?

docs/orchestration/faq/debug.md
madkinsz: This is from the old docs and I'm not really sure how often this would happen anyway. Not sure what to recommend here since if they're exporting this variable they may not be using the CLI

src/prefect/cli/auth.py
madkinsz: Naw we don't want to modify the `config.toml` programatically, we can't preserve comments easily and it's not in-line with our previous statements about it

src/prefect/agent/agent.py
zangell44: ahhh gotcha, got confused since you can set api keys under config.cloud.agent.auth_token

zangell44 (approved):

PR #4612, opened 2021-06-02
Remove deprecated items for 0.15.0

madkinsz:
## Summary

Anything that was deprecated in 0.14.0 or earlier is slated for removal in 0.15.0 as it has been deprecated for a full release cycle.

A changelog entry and more thorough summary is pending.

PR #4652, opened 2021-06-11
Remove deprecated items for 0.15.0 but retain environments

madkinsz:
## Summary

Rebase of #4612 to drop all commits removing `prefect.environment` support and the `FargateAgent` (which is needed for the fargate environment). We will likely drop these in 0.16.0 instead. We might be able to re-use the commits from #4612 then which would be nice because it was a lot of work.

## Changes

The commits should be relatively clear, see the changelog for a summary.
This drops everything deprecated since 0.14.0 except:
- The `prefect.engine.executors` module (moved to `prefect.executors` in 0.14.0)
- Flow "Environments" (deprecated in 0.14.0 replaced by "Run Configs")

## Checklist

This PR:


src/prefect/serialization/schedule.py
```diff
@@ -133,19 +131,3 @@ class Meta:
...
-class ScheduleSchema(OneOfSchema):
-    """
-    Field that chooses between several nested schemas. This class is preserved for pre-0.6.1
-    compatibility, and is deprecated in favor of NewScheduleSchema.
-    """
-
-    # map class name to schema
-    type_schemas = {
-        "Schedule": NewScheduleSchema,
-        "IntervalSchedule": schedule_compat.IntervalScheduleSchema,
```
zangell44, line 147: do we need any tests for serializing and deserializing `Intervalschedule` and `CronSchedule`?

it seems like since both are just functions returning `Schedule` classes we're okay

tests/tasks/dropbox/test_dropbox.py
```diff
@@ -25,29 +25,12 @@ def test_download_initializes_attr_from_kwargs(self, attr):
 
 
 class TestCredentials:
-    def test_creds_are_pulled_from_secret_at_runtime(self, monkeypatch):
-        task = DropboxDownload(path="test", access_token_secret="DROPBOX_ACCESS_TOKEN")
+    def uses_access_token(self, monkeypatch):
```
zangell44, line 28:
```suggestion
    def test_uses_access_token(self, monkeypatch):
```

zangell44 (approved):
One minor change, one question - looks good to me

src/prefect/serialization/schedule.py
madkinsz: I think we will just let them fail to deserialize since they're for pre 0.6.1 support

zangell44 (dismissed):

zangell44 (approved):

zangell44 (approved):

Issue #4648, opened 2021-06-11
Prefect flow secrets from Vault


datwiz:
## Current behavior
Prefect supports sourcing secrets using either the base PrefectSecret task or the EnvVarSecret task.  Secret values are then either stored in Prefect Cloud or sourced from the flow or agent config.  Setting secrets at the config level potentially provides wider exposure of sensitive information to all flows.  Similarly, the Prefectd Cloud secret store is outside of the control of the private environment.
When using Hashicorp Vault for secrets management, it would be desirable to use that privately controlled Vault service for credentials and secrets.

## Proposed behavior
Add a VaultSecret task as a sub-class of PrefectSecret that sources secrets content from a privately controlled vault service.

## Example
Given a set of key/value pairs stored in a secret in a Hashicorp Vault service at the path "secrets/fake/secret"
When I invoke VaultSecret.run() 
Then the key/value content of the secret is returned as a dictionary.

```
secret_path = "secrets/fake/secret"
secret_val = VaultSecret(secret_path).run()
```

A reference implementation that will be used as the basis for PR can be found here: https://github.com/vcdi/prefect-vault-secrets

PR #4656, opened 2021-06-13
4648 prefect flow secrets from vault

datwiz:
## Summary
Added new optional task VaultTask for accessing flow run secrets from a private Hashicorp Vault service. 

Items needing review (at least):
* correct packaging to enable optional installation.
* correct test case configuration, especially with some test cases expecting location execution and some using server.
* any additional documentation requirements to describe how to use the VaultSecret task.

## Changes
Added VaultSecret task as a subclass of SecretBase.
Test case coverage added to the secrets task directory.

The VaultSecret task is intended to be an optional module installed via pip install prefect[vault], or similar. 

## Importance
Enables use of a private secrets service for an improved security posture. 

## Checklist
This PR:


datwiz:
Thanks,
I think the majority of the failures are due to the dependency on the Hashicorp vac python module but not sure of the correct approach to adding in a new optional module dependency.
I've added the module in `setup.py` but it a) may not be added correctly, and b) may need additional changes elsewhere - possible in the circle-ci definition?

tests/tasks/secrets/test_vault_secret.py
```diff
@@ -0,0 +1,147 @@
+import os
+import hvac
```
madkinsz, line 2: You'll want to use `pytest.importorskip` here so these tests are skipped if this module is not available

src/prefect/tasks/secrets/__init__.py
```diff
@@ -7,3 +7,4 @@ class for interacting with other secret providers. Secrets always use a special
 """
 from .base import SecretBase, PrefectSecret, Secret
 from .env_var import EnvVarSecret
+from .vault_secret import VaultSecret
```
madkinsz, line 10: I think you'll have to remove this so the user will have to import from the full path if they want vault secrets since it uses an extra requirement.

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,143 @@
+import hvac
```
madkinsz, line 1: You'll want to warn if they try to import this without the extra requirement

See https://github.com/PrefectHQ/prefect/blob/8ab1543e9761ac591a3ecd6ac7d3fc87d3ae44d8/src/prefect/tasks/dbt/__init__.py

madkinsz (commented):
Made a few comments that should help you get going :)

datwiz:
That's awesome.  Thanks for the detail.  I'll get these changes rolled in and re-push.

tests/tasks/secrets/test_vault_secret.py
datwiz: added purest.importorskip("hvac") to __init__.py

src/prefect/tasks/secrets/__init__.py
datwiz: vault_secret import removed from __init__.py

src/prefect/tasks/secrets/vault_secret.py
datwiz: added the import warning wrapper

datwiz:
Resolves issue #4648

datwiz:
Going to update the implementation to support first attempting to lookup a secret at the supplied path and return all the key/value entries found, else attempt to use the provided path as a `secret/key` path and return the content of a specific key.

datwiz:
Second thought - staying with the more consistent single secret lookup approach returning the key/value pairs as dict items.

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+        vault_path = m.groupdict()
+        value = ""
+        try:
+            vault_secret = client.secrets.kv.v2.read_secret_version(
+                path=vault_path["path"], mount_point=vault_path["mount_point"]
+            )
+            value = vault_secret["data"]["data"]
+        except hvac.exceptions.InvalidPath as exc:
+            raise KeyError(f"Secret not found: {vault_path['path']}") from exc
```
TylerWanner, line 113: is it confusing to only see the path and not the mount_point/path here since the full thing is supplied to the task?

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+        client = hvac.Client()
+        # get vault address url
+        vault_url = os.getenv("VAULT_ADDR", default=None)
+        if vault_url is None:
+            vault_url = os.getenv("vault_addr", default=None)
+            if vault_url is None:
+                raise ClientError(
+                    "VAULT_ADDR url var not found. "
+                    'Either "VAULT_ADDR" or "vault_addr" env var required.'
+                )
```
madkinsz, line 73:
```suggestion
        vault_url = os.getenv("VAULT_ADDR") or os.getenv("vault_addr")
            raise ValueError(
                "VAULT_ADDR url var not found. "
                'Either "VAULT_ADDR" or "vault_addr" env var required.'
            )
```
- `os.getenv` defaults to `None` by default, I don't think it's worth being more explicit there
- `ClientError` is reserved for errors in the `prefect.Client`
- Is there precedence for checking for the lowercase variable as well? If that's not something vault generally does, just the upper-case should suffice

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+            client.auth_approle(
+                vault_creds["VAULT_ROLE_ID"], vault_creds["VAULT_SECRET_ID"]
+            )
+        else:
+            raise ClientError(
+                "Unable to authenticate with vault service.  "
+                "Supported methods: token, appRole"
+            )
+        if not client.is_authenticated():
+            raise PermissionError(
```
madkinsz, line 94: I believe `PermissionError` is reserved for file-based permission issues. a `RuntimeError` should suffice here.

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+        vault_url = os.getenv("VAULT_ADDR", default=None)
+        if vault_url is None:
+            vault_url = os.getenv("vault_addr", default=None)
+            if vault_url is None:
+                raise ClientError(
+                    "VAULT_ADDR url var not found. "
+                    'Either "VAULT_ADDR" or "vault_addr" env var required.'
+                )
+        client.url = vault_url
+        self.logger.debug(f"vault addr set to: {client.url}")
```
madkinsz, line 75:
```suggestion
        self.logger.debug(f"Vault addr set to: {client.url}")
```
Generally, I'd capitalize the first word of each of these debug logs to match the rest of the logs from Prefect

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+        if not client.is_authenticated():
+            raise PermissionError(
+                "Unable to autheticate with vault using supplied credentials"
+            )
+        self.logger.debug("passed vault authentication check")
+        # regex to parse path into 2 named parts: <mount_point>/<path>
+        secret_path_re = r"^(?P<mount_point>[^/]+)/(?P<path>.+)$"
+        m = re.fullmatch(secret_path_re, name)
+        if m is None:
+            raise KeyError(
```
madkinsz, line 102:
```suggestion
            raise ValueError(
```

src/prefect/tasks/secrets/vault_secret.py
```diff
@@ -0,0 +1,161 @@
...
+        secrets = prefect.context.get("secrets", {})
+        value = None
+        try:
+            value = secrets[name]
+        except KeyError:
+            if prefect.context.config.cloud.use_local_secrets is False:
+                # lookup the secret value in Vault
+                value = self._get_vault_secret(name)
+            else:
+                raise ValueError(f'Local secret "{name}" not found.') from None
```
madkinsz, line 156: I'm not sure this should re-use the Prefect local secret logic but I see the use-case for it. @cicdw your opinion would be helpful here.

tests/tasks/secrets/test_vault_secret.py
```diff
@@ -0,0 +1,184 @@
...
+    vs = VaultSecret("fake-local-secret")
+    with prefect.context(secrets=fake_vault_secrets):
+        val = vs.run()
+        assert val == {"fake-local-secret-key": "fake-local-secret-value"}
+
+
+def test_vault_addr_env_var_missing(server_api):
+    prefect.context.config["cloud"]["use_local_secrets"] = False
+    os.unsetenv("vault_addr")
+    os.unsetenv("VAULT_ADDR")
```
madkinsz, line 88: I believe you should use `monkeypatch.setenv` here instead of modifying the os directly

tests/tasks/secrets/test_vault_secret.py
```diff
@@ -0,0 +1,184 @@
...
+        task = VaultSecretTestTask("fake-no-vault-addr-secret")
+        task.run()
+
+
+@pytest.mark.parametrize("vault_var", ["vault_addr", "VAULT_ADDR"])
+def test_vault_addr_from_env_var(vault_var, server_api):
+    """
+    Verify accepting either upper case or lower case vault addr env vars
+    """
+    os.environ[vault_var] = "http://localhost:8200"
```
madkinsz, line 99: As above, I'd use `monkeypatch.setenv`

tests/tasks/secrets/test_vault_secret.py
```diff
@@ -0,0 +1,184 @@
...
+    prefect.context.config["cloud"]["use_local_secrets"] = False
```
madkinsz, line 100: This should use the `set_temporary_config` context manager so it does not affect other tests

tests/tasks/secrets/test_vault_secret.py
madkinsz: https://stackoverflow.com/questions/38410878/why-use-monkeypatch-setenv-instead-of-os-putenv

tests/tasks/secrets/test_vault_secret.py
```diff
@@ -0,0 +1,184 @@
...
+    Verify secret class instance can be pickeled
+    """
+    vs = VaultSecret("fake-pickeled-vault-secret")
+    new = cloudpickle.loads(cloudpickle.dumps(vs))
+    assert new.name == "fake-pickeled-vault-secret"
+    assert isinstance(new.result, SecretResult)
+
+
+# --- Test Harness Task ---
+class VaultSecretTestTask(Task):
```
madkinsz, line 39: I think we would recommend most users do something like


```python
@task
def identity(obj):
    return obj

with Flow(...) as flow:
    my_secret = VaultSecret("key")
    identity(my_secret)
```

Where their secret is resolved into a value by Prefect instead of calling `.run()` themselves. I don't think a task with a secret.run() call in it like this helps with the clarity of your tests. I'd put it in a flow if you're trying to ensure it is passed along correctly while creating an example of best practice.

madkinsz (changes requested):
Thanks for putting some time into this! It's really coming along.

TylerWanner:
This looks solid.  Should be a very clean enhancement for anyone running Vault and wanting to reduce secret surface area.  Would love to see it in use

src/prefect/tasks/secrets/vault_secret.py
TylerWanner: I agree but believe you're missing a line in there before the raise? @madkinsz

madkinsz: Thanks Tyler -- edited.

src/prefect/tasks/secrets/vault_secret.py
cicdw: Yea I don't think we should reuse the local secrets logic here honestly; that flag is intended for Prefect Cloud secrets specifically and no other custom Secret implementation uses it so that would introduce non-uniform behavior

src/prefect/tasks/secrets/vault_secret.py
datwiz: Code simplified and `ClientError` exceptions and docs updated to use `ValueError`.

The check for both upper case and lower case versions enables support for 2 different methods for configuring env vars in the job execution environment.  Over time, we've used  both scenarios.
The Vault utilities expect an upper case `VAULT_ADDR` value.  If using a custom base image, the `ENV VAULT_ADDR=<vault-add>` env var can be set in the image definition and the value used by both prefect jobs and other scripts that use the same base image.
We also use a helm chart to configure the agent and job configuration, setting the vault address there, in case we have a flow that doesn't use our base image.  We use env vars in the chart to set value, which then get [interpolated to lower case keys](https://docs.prefect.io/core/concepts/configuration.html#environment-variables), e.g. from the deployment spec:
```yaml
...
    env:
    - name: PREFECT__CLOUD__AGENT__ENV_VARS__vault_addr
      value: "<vault-addr>"
...
```

madkinsz: Thanks!

Hmm. I see your point regarding the lower-casing of environment variables. We explicitly don't do that with secrets and I think we may want to replicate that logic for the `AGENT__ENV_VARS__{VAR}` setting -- that seems like an oversight.

src/prefect/tasks/secrets/vault_secret.py
datwiz: I'll refactor to remove the use of local secrets.
I'd originally taken the approach that a secret was a secret and it could be accessed in different remote services.  The local storage of a secret then took precedence and could be used to support isolated/local development.
In practice, we seldom use the local secrets option over accessing a running Vault service, especially given the fact that secrets are usually needed to interact with other remote services anyway.

src/prefect/tasks/secrets/vault_secret.py
datwiz: changed `KeyError` to `ValueError`

src/prefect/tasks/secrets/vault_secret.py
datwiz: debug messages updated to use consistent style.

src/prefect/tasks/secrets/vault_secret.py
datwiz: `PermissionError` exceptions changed to `RuntimeError` exceptions with corresponding docs update.

tests/tasks/secrets/test_vault_secret.py
datwiz: `os.setenv()` and `os.unsetenv()`  calls replaced with monkey patch `setenv()` and `delenv()` calls.

tests/tasks/secrets/test_vault_secret.py
datwiz: now using `monkeypatch.setenv()`

tests/tasks/secrets/test_vault_secret.py
datwiz: updated tests to use `set_temporary_config()` context manager

tests/tasks/secrets/test_vault_secret.py
datwiz: Got it.  We had been writing our custom tasks with `secret.run()` calls contained in the task as opposed to passing the output of a flow level `secret()` definition in as a task argument.
Thinking about it, the nested secret in task approach does add implicit coupling.  The task argument approach does seem cleaner.  Will refactor the test cases to reflect the approach above.

src/prefect/tasks/secrets/vault_secret.py
datwiz: local secret lookup removed.  VaultSecret now only retrieves secret values from the vault service.

tests/tasks/secrets/test_vault_secret.py
datwiz: Removed `VaultSecretTestTask` as a wrapper class as it's not required.
Also added a new test case to verify use of a VaultSecret in a flow definition.

madkinsz (approved):
This looks great! I'll get it in 0.15.2. Thanks :)

th0ger:
@madkinsz @datwiz You forgot to add `VaultSecret` to the [API docs](https://docs.prefect.io/api/latest)?

PR #4669, opened 2021-06-17
SendGrid.SendEmail task - use Secret value

taljaards:
## Summary
Change SendGrid SendEmail task to just use the given secret _value_, instead of manually using the secret name to get the secret value.

## Changes
Remove use of hardcoded `SENDGRID_API_KEY` secret.
This change then outsources the `get`ting of the secret's value.

## Importance
This change allows more flexibility with how the key is given to the Task.
Previously, "Prefect-native" secrets had to be used.
This change allows getting the secret value manually in a flow, e.g. a secret saved in GCP Secret Manager, or AWS Secrets Manger. It also allows more flexibility, since the user can now save the actual secret value as a JSON blob (they just need to convert the value to a string before using this Task).

## Checklist

This PR:

- [x] adds new tests

### DIFF
--- a/src/prefect/tasks/sendgrid/sendgrid.py
+++ b/src/prefect/tasks/sendgrid/sendgrid.py
@@ -1,7 +1,6 @@
 from pathlib import Path
 from typing import Any, List, Tuple, Union
 
-from prefect.client import Secret
 from prefect.core import Task
 from prefect.utilities.tasks import defaults_from_attrs
 from python_http_client.client import Response
--- a/src/prefect/tasks/sendgrid/sendgrid.py
+++ b/src/prefect/tasks/sendgrid/sendgrid.py
@@ -22,7 +22,7 @@ class SendEmail(Task):
         - attachment_file_path (Union[str, Path], optional): The file path of the email attachment;
             can also be provided at runtime
         - sendgrid_api_key (str): The SendGrid API key used for authentication;
-            can also be provided at runtime (it is preferred, actually!)
+            can also be provided at runtime, which is preferred since a secret can be used
         - **kwargs (optional): additional kwargs to pass to the `Task` constructor
     """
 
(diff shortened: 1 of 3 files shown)


src/prefect/tasks/sendgrid/sendgrid.py
```diff
@@ -24,8 +21,8 @@ class SendEmail(Task):
             can also be provided at runtime
         - attachment_file_path (Union[str, Path], optional): The file path of the email attachment;
             can also be provided at runtime
-        - sendgrid_secret (str, optional): the name of the Prefect Secret which stores your
-            SendGrid API key; defaults to `"SENDGRID_API_KEY"`
+        - sendgrid_api_key (str): The SendGrid API key used for authentication;
+            can also be provided at runtime (it is preferred, actually!)
```
madkinsz, line 25:
```suggestion
            can also be provided at runtime, which is preferred since a secret can be used
```

src/prefect/tasks/sendgrid/sendgrid.py
```diff
@@ -37,7 +34,7 @@ def __init__(
         html_content: str = None,
         category: Union[str, List[str]] = None,
         attachment_file_path: Union[str, Path] = None,
-        sendgrid_secret: str = "SENDGRID_API_KEY",
+        sendgrid_api_key: str = None,
```
madkinsz, line 37: Generally, we would have to support this old argument with a deprecation cycle. Since this is a very new task I'm tempted to just make the switch. Thoughts @zangell44 ?

madkinsz (commented):
Thanks for following up on this!

src/prefect/tasks/sendgrid/sendgrid.py
zangell44: Yeah this is new enough I have no problem making the switch

src/prefect/tasks/sendgrid/sendgrid.py
taljaards: That's better, yes 🙂

src/prefect/tasks/sendgrid/sendgrid.py
taljaards: Haha, I was taking a chance.
Thanks!

madkinsz (approved):

PR #4676, opened 2021-06-18
Fix some issues with new auth CLI

madkinsz:
## Summary

If you had set a tenant previously you could "login" with a key that is invalid and the command would succeed. Future calls to the API would still fail of course.

## Changes

- Ensures that old tenant ids are not used during login
- Ensures that login is actually successful
- Breaks out `get_default_tenant` into Cloud/Server specific functions
- Adds `TenantView` and displays the tenant name/slug on login
- Adds some missing tests for the auth login CLI

## Importance

Bugfix; improved test coverage; more user-friendly output

## Checklist

Changes file does not include fixes to CLI because it's unreleased yet

This PR:


src/prefect/backend/tenant.py
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+        }
+
+        result = client.graphql(tenant_query)
+        tenants = result.get("data", {}).get("tenant", None)
+
+        if tenants is None:
+            raise ValueError(
+                f"Received bad result while querying for tenants where {where}: "
```
zangell44, line 102:
```suggestion
                f"Received bad result while querying for tenants where {where!r}: "
```

src/prefect/cli/auth.py
```diff
@@ -1,20 +1,29 @@
...
+        if os.environ.get("PREFECT__CLOUD__AUTH_TOKEN"):
+            click.secho(
+                "Auth token has been set in the environment. The CLI cannot be used to "
+                "manage your auth. Unset the key with `unset PREFECT__CLOUD__AUTH_TOKEN`",
+                fg="red",
+            )
+        else:
+            click.secho(
+                "Auth token has been set in the config. The CLI cannot be used to "
+                "manage your auth. Remove the key from your configuration file.",
```
zangell44, line 24: can we specify the location of the configuration file here?

zangell44 (approved):
a few minor comments, lgtm

src/prefect/backend/tenant.py
madkinsz: It's just a dict, the repr isn't particularly more meaningful I think

src/prefect/cli/auth.py
madkinsz: Probably, but its path can be changed and I don't know if it's worth the extra effort -- they had to modify it to set this value. We also don't know if they've modified a config file that they shouldn't to change this value since the config is actually a stack of files combined.

PR #4678, opened 2021-06-18
Increase opensky request timeout to 5m

gabrielvieira37:
Change opensky api request timeout to 300 seconds.

I tried to run the tutorial code, but the timeout window of only 10 seconds was far too low.

I tried to run the request manually and got the following request time in seconds:

image

Setting this timeout to 300 will help to not break the code when facing low quality internet problems.


madkinsz (approved):
Thanks!

PR #4685, opened 2021-06-21
Split the auth save/load functions out of the `Client

madkinsz:
## Summary

Previously I provided asymmetric functions in the `Client`.

`Client.save_auth_to_disk() -> None` -- uses values from the client

and

`Client.load_auth_from_disk() -> dict` -- does not mutate client

I found this asymmetry confusing. It also led to cases where you'd do

```python
client = Client()
client.tenant_id = "foo"
client.save_auth_to_disk()
```

or where you'd instantiate a client just to see what auth would be used

```python
client = Client()
api_key = client.api_key
```

To resolve this, I split out these methods into

`load_auth_from_disk(api_server: str) -> dict`
`save_auth_to_disk(api_server: str, api_key: str = None, tenant_id: str = None) -> None`

and retained

`Client.save_auth() -> None`

I'm not sure this is better. I think what might make the most sense is

`get_default_auth(api_server: str = prefect.config.cloud.api, token_only: bool = False) -> dict`

which will return an API key that has been properly resolved from disk / config _or_ a token if there's no API key and a token exists _or_ just do token resolving for places where we need that backwards compatible. This would be helpful for lowering the boilerplate in https://github.com/PrefectHQ/prefect/pull/4683

### DIFF
--- a/docs/orchestration/concepts/api.md
+++ b/docs/orchestration/concepts/api.md
@@ -84,14 +84,15 @@ client.tenant_id = "<new-id>"
 Authentication can be saved to disk. This will save the current API key and tenant id to `~/.prefect/auth.toml` and future clients instantiated without an API key or tenant id will load these values as defaults.
 
 ```python
-client.save_auth_to_disk()
+client.save_auth()
 ```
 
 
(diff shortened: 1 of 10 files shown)


madkinsz:
Not sure I want to address this yet

PR #4686, opened 2021-06-21
Always retrieve the tenant id in the Client

madkinsz:
## Summary

Previously, if using an API key `Client.tenant_id` would return `None` if a tenant had not been explicitly set. The backend will infer the default tenant for us and there's no reason to query for it. However, `Client.tenant_id` is used by the agent during registration (why it's not just inferred from authentication, I'm not sure -- cc @cicdw). Unfortunately this means that #4676 broke launching agents with API keys. This behavior was generally confusing anyway as it required `client.tenant_id or client.get_auth_tenant()` if you wanted to _for sure_ get the current tenant id without always querying the backend so I think this is probably worth doing.

## Changes

- The `tenant_id` property now retrieves the default tenant for the key and sets `_tenant_id` for API keys on first use
- Saving auth to disk uses `_tenant_id` to prevent this pull from happening unexpectedly
- `test_register_agent` now actually covers the graphql query
- The `client.tenant_id or client.get_auth_tenant()` pattern can and has been replaced with just `client.tenant_id`
- Default tenant retrieval helper methods are now private
- When logging in, the tenant id is written to disk instead of being left blank to use the key default. This means if the default for a key changes after login, the tenant will not suddenly change.

## Other considerations

We could move the `tenant_id` property to a `get_active_tenant()` method which would stop it from hiding a query to the API but change the existing interface.

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -118,7 +118,6 @@ class Client:
         self._access_token = None
         self._refresh_token = None
         self._access_token_expires_at = pendulum.now()
-        self._tenant_id = None
         self._attached_headers = {}  # type: Dict[str, str]
         self.logger = create_diagnostic_logger("Diagnostics")
 
@@ -133,18 +132,19 @@ class Client:
 
         # Load the API key
         cached_auth = self.load_auth_from_disk()
-        self.api_key = (
+        self.api_key: Optional[str] = (
             api_key
             or prefect.context.config.cloud.get("api_key")
             or cached_auth.get("api_key")
         )
 
         # Load the tenant id
-        self._tenant_id = (
+        self._tenant_id: Optional[str] = (
             tenant_id
             or prefect.context.config.cloud.get("tenant_id")
             or cached_auth.get("tenant_id")
         )
+
         # If not set at this point, when `Client.tenant_id` is accessed the default
         # tenant will be loaded and used for future requests.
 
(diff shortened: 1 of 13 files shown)


madkinsz:
There are a few routes that require explicit tenant ids e.g. `create_project` and `register_agent` which is likely related to Server compatibility. I think this is a reasonable way to provide the tenant id without diverging from the old interface too much.

src/prefect/client/client.py
```diff
@@ -188,35 +190,37 @@ def __init__(
...
-                )
+        response = self.graphql({"query": {"auth_info": "tenant_id"}})
+        tenant_id = response.get("data", {}).get("auth_info", {}).get("tenant_id", "")
 
-            return tenant_id
-        else:
+        # If the backend returns a `None` value tenant id, it indicates that an API
+        # token was passed in as an API key
+
+        if tenant_id == "":
```
zangell44, line 215: will `None` be returned as the tenant_id or will the tenant_id field be missing from the response entirely?

src/prefect/client/client.py
```diff
@@ -188,35 +190,37 @@ def __init__(
 
     # API key authentication -----------------------------------------------------------
 
-    def get_auth_tenant(self) -> str:
+    def _get_auth_tenant(self) -> str:
         """
         Get the current tenant associated with the API key being used. If the client has
         a specific tenant id set, this will verify that the given tenant id is
```
zangell44, line 196: how does this method actually do the verification? it because we're passing "x-prefect-tenant-id" as a header?
I think this comment has already tripped me up once before, maybe worth adding to the docstring

madkinsz: Since we're attaching the tenant id the backend raise an unauthenticated error if it's not a valid tenant for the key. I've added a bit more to the comment.

src/prefect/client/client.py
madkinsz: I've updated this to be clearer; but

Missing from the response -> malformed response from backend
Response is `None` -> API token was used as API key

zangell44 (approved):

PR #4693, opened 2021-06-23
Fix missing event timestamp attribute errors in K8s agent

madkinsz:
## Summary

Follow-up to https://github.com/PrefectHQ/prefect/pull/4544 which dropped events with null timestamps which now accounts for events missing timestamp attributes _entirely_

## Changes

- Uses `getattr` to check for the attribute
- Logs these events so we can see what's going on

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -7,6 +7,7 @@ import uuid
 from typing import Optional, Iterable, List, Any
 
 import json
+import pendulum
 import yaml
 
 import prefect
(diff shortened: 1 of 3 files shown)


zangell44 (approved):

src/prefect/agent/kubernetes/agent.py
```diff
@@ -227,12 +227,20 @@ def manage_jobs(self) -> None:
...
-                                    pod_events.items, key=lambda x: x.last_timestamp
+                                    pod_events.items,
+                                    key=lambda e: getattr(e, "last_timestamp", None),
                                 ):
-                                    # Skip old events or events without timestamps
+                                    # Skip events without timestamps
+                                    if not getattr(event, "last_timestamp", None):
+                                        self.logger.debug(
+                                            f"Encountered K8s event on pod {pod_name!r}"
+                                            " with no timestamp: {event!r}"
```
madkinsz, line 237:
```suggestion
                                            f" with no timestamp: {event!r}"
```

zangell44 (approved):

PR #4694, opened 2021-06-23
Add `prefect run` example to `hello_world` module docstring

madkinsz:
Now that the `prefect run` changes are in `master`

Includes a fix to a flakey docker test

### DIFF
--- a/src/prefect/hello_world.py
+++ b/src/prefect/hello_world.py
@@ -18,6 +18,11 @@ Examples:
     ------------------------------------------
     $ prefect create project 'default'
     $ prefect register --project default -m prefect.hello_world
+
+
+    Run this flow with the Prefect backend and agent
+    ------------------------------------------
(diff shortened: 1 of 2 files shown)


src/prefect/hello_world.py
```diff
@@ -18,6 +18,11 @@
     ------------------------------------------
     $ prefect create project 'default'
     $ prefect register --project default -m prefect.hello_world
+
+
+    Run this flow with the Prefect backend and agent
```
zangell44, line 23: is there a way to make it clearer that an agent must be running for this to work?

madkinsz: Well, `--watch` is included which means after 15 seconds it'll tell you if your agent is not running

zangell44: that's how I found out 😄

zangell44 (approved):

Issue #4663, opened 2021-06-16
Single connection timeout causes Zombie Run


novotl:
## Description

We're running Prefect Server on our Kubernetes cluster and schedule Flows with KubernetesAgent. Occasionally we observe strange flow run failures and upon investigating flow run logs it's clearly visible that there are 2 instances of that flow run running in parallel (one Zombie run that lost heartbeat process and other created by Lazarus). 

Also after some time, this log appears (delay is caused by 10 minute heartbeat cutoff)
```
No heartbeat detected from the remote task; marking the run as failed.
```
along with that, we sometimes also get this warning log
```
Error getting flow run info
Traceback (most recent call last):
  File "/opt/venv/lib/python3.6/site-packages/urllib3/connection.py", line 160, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw
  File "/opt/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 84, in create_connection
    raise err
  File "/opt/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 74, in create_connection
    sock.connect(sa)
socket.timeout: timed out
...
requests.exceptions.ConnectTimeout: HTTPConnectionPool(host='prefect-apollo', port=4200): Max retries exceeded with url: /graphql (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7f28fb632cc0>, 'Connection to prefect-apollo timed out. (connect timeout=15)'))
```
Clearly, there is a period in which the Apollo service was not available, but it was restored after some time (because no more warning logs).

The second instance of the task is because of the lost heartbeat of the first instance. The heartbeat process started here
https://github.com/PrefectHQ/prefect/blob/ea306ffa41bd520ebe56036f9d1ae48f5725280a/src/prefect/engine/cloud/flow_runner.py#L77-L85

There is no error handling in the heartbeat code, so a single timeout causes the whole heartbeat process to go down. When the connection becomes available again the heartbeat process is dead. This is unfortunate, as with the default heartbeat interval = 30 sec and heartbeat cutoff = 10 minutes we still got plenty of heartbeat tries to keep the flow run alive.
https://github.com/PrefectHQ/prefect/blob/ea306ffa41bd520ebe56036f9d1ae48f5725280a/src/prefect/cli/heartbeat.py#L91-L98

[...]

PR #4736, opened 2021-07-02
Update flow run heartbeats to be robust to exceptions

madkinsz:
## Summary

Currently, a connection error during heartbeats relies on `Client` logic for retries and after the retries expire, the heartbeat process will fail. As long as the heartbeat's parent is running we should attempt to send a signal that the flow run is being executed. This PR adds exception handling to the heartbeat.

## Changes

- Adds an error level log when an exception is encountered
- Ignores exceptions during client calls
- Logs base exceptions before exiting the process (ie sigterm)
- Heartbeat command logs are sent to Cloud

## Importance

Robust heartbeats! Hopefully this addresses #3058 as well as #4663

## Checklist

This PR:


madkinsz:
I've noticed that we're piping stdout/err to DEVNULL when we call this command so we'll want to make sure these logs are written to the backend or change the output handling in the flow runner before merging this

zangell44 (dismissed):

madkinsz:
Instead of dealing with consuming data from a subprocess; we will drop the debug level log and send all error logs to Cloud

madkinsz:
In the future, we may want to avoid sending the same error a bunch of times :)

src/prefect/cli/heartbeat.py
```diff
@@ -89,10 +91,32 @@ def flow_run(id, num):
...
+
+        try:  # Log signal-like exceptions that cannot be ignored
+
+            while iter_count < (num or 1):
+
+                try:  # Ignore (but log) client exceptions
+                    client.update_flow_run_heartbeat(id)
+                except Exception as exc:
+                    logger.error(
+                        f"Failed to send heartbeat with exception: {exc!r}",
```
cicdw, line 109: Does repr'ing the exception cause the log message to be twice as long?

cicdw (approved):
Non-critical question but this LGTM

zangell44 (approved):

src/prefect/cli/heartbeat.py
madkinsz: Hm; it depends on the exception but most of what I'm looking for with `exc_info` is the traceback. It looks like

```
└── 15:32:43 | ERROR   | Failed to send heartbeat with exception: ValueError('Foo')
  File "/Users/mz/prefect/core/src/prefect/cli/heartbeat.py", line 107, in flow_run
    raise ValueError("Foo")
ValueError: Foo
```

Might be a little overkill but 🤷 I'll keep that in mind for future error logs.

PR #4744, opened 2021-07-06
Checkpoint task results when `SUCCESS` is raised

madkinsz:
## Summary

If a user wants to customize the state message but still end successfully, they can do

```python
raise prefect.engine.signals.SUCCESS("My custom message", result=<data>)
```

but because state signals bypass checkpointing logic, this task result is not stored in the provided `Result` location.

## Changes

- Removes a signal re-raise that was a workaround while waiting for #4664 
- `SUCCESS` signals are treated like a successful task run instead of falling through

## Importance

- Allows setting custom state messages while returning data
- Makes task run result checkpointing more consistent

## Checklist

This PR:

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -879,14 +879,11 @@ class TaskRunner(Runner):
                 new_state.loop_count
             )
 
-        except signals.PrefectStateSignal:
-            # Other state signals will be handled by the `call_state_handlers`
-            # decorator. Once Prefect signal exceptions are modified to inherit
-            # from `BaseException` instead of `Exception`, this can be removed
-            raise
-        except signals.ENDRUN:
-            # As above, the `ENDRUN` is a special exception
-            raise
+        except signals.SUCCESS as exc:
+            # Success signals can be treated like a normal result
+            new_state = exc.state
+            assert isinstance(new_state, Success)
+            value = new_state.result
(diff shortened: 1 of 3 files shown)


madkinsz:
Not sure where the best place to test this is yet.

madkinsz:
@cicdw just want to check there are no objections to this

src/prefect/engine/task_runner.py
```diff
@@ -879,14 +879,11 @@ def get_task_run_state(self, state: State, inputs: Dict[str, Result]) -> State:
...
 
-        except signals.PrefectStateSignal:
-            # Other state signals will be handled by the `call_state_handlers`
-            # decorator. Once Prefect signal exceptions are modified to inherit
-            # from `BaseException` instead of `Exception`, this can be removed
-            raise
-        except signals.ENDRUN:
-            # As above, the `ENDRUN` is a special exception
-            raise
+        except signals.SUCCESS as exc:
```
cicdw, line 882: I think you still need the other special cases to prevent the catch-all handling on line 889, but otherwise this special case definitely makes sense to me

madkinsz: Sweet. Since #4664 changes our internal signals to `BaseException` we don't.

zangell44 (approved):
lgtm

PR #4747, opened 2021-07-06
Check truthiness instead of identity in checkpointing logic

bnaul:
Curious if I'm missing something here since the current style is pretty unorthodox which maybe suggests it's like that for a reason? But in addition to violating [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations), the current logic behaves incorrectly for the default `Task.checkpoint` value of `None` https://github.com/PrefectHQ/prefect/blob/master/src/prefect/core/task.py#L324, which I've run into locally with checkpointing enabled.

Doesn't look like this has been touched lately but @cicdw might be the most recent.

### DIFF
--- a/src/prefect/engine/task_runner.py
+++ b/src/prefect/engine/task_runner.py
@@ -901,8 +901,8 @@ class TaskRunner(Runner):
         # checkpoint tasks if a result is present, except for when the user has opted out by
         # disabling checkpointing
         if (
-            prefect.context.get("checkpointing") is True
-            and self.task.checkpoint is not False
+            prefect.context.get("checkpointing")
+            and self.task.checkpoint
             and value is not None
         ):
             try:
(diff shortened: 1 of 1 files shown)


cicdw:
Purely from memory, the `None` distinction was due to old backwards compatibility reasons where `None` meant "use the configured default" (see https://github.com/PrefectHQ/prefect/pull/1933).  Now that the configuration setting is gone, `None` means "use whatever Prefect would like to use" which defaults to having checkpointing enabled per the `is not False` logic.

While I fully understand the proposed change here, it actually represents a breaking change unless you also update the task default to `True` (for the reasons above).  

Side question for my own understanding: why do you say this violates PEP 8?

bnaul:
Makes sense; I see how this usage of None does achieve the desired result, even if it's a bit counter-intuitive to look at. If `True` and `None` have the same affect here, then would it make sense to change the default for `Task.checkpoint` to `True`?

Re: PEP8, here is the relevant snippet from the link I posted
image

cicdw:
In this case we did want to actually check identity with `False` but I see what you mean.  

> would it make sense to change the default for Task.checkpoint to True?

Yea it looks like that should be fine, I can't think of any reason that would break, especially if behavior tests pass.

madkinsz:
Closing as stale. 

There are lots of places we could give this a second look, but it's not always just style. There are some weird gotchas, for example:

```python
if x:
   ...
```

makes sense until `x` is a dataframe which will throw exceptions on truthiness checks. Since we're moving towards stability here, I'm hesitant to go around changing these comparisons.

PR #4749, opened 2021-07-06
Enable overwrite for Azure Blob Storage upload task

jnoynaert:
## Summary

The current Azure blob task doesn't allow using [the overwrite parameter](https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobclient?view=azure-python#upload-blob-data--blob-type--blobtype-blockblob---blockblob----length-none--metadata-none----kwargs-)

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/azure/blobstorage.py
+++ b/src/prefect/tasks/azure/blobstorage.py
@@ -75,6 +75,12 @@ class BlobStorageUpload(Task):
         - azure_credentials_secret (str, optional): the name of the Prefect Secret
             that stores your Azure credentials; this Secret must be an Azure connection string
         - container (str, optional): the name of the Azure Blob Storage to upload to
+        - overwrite_blob (bool, optional): Whether the blob to be uploaded should overwrite
+                    the current data. If True, upload_blob will overwrite the existing data.
+                    Note for Append blob types: if set to False and the data already exists,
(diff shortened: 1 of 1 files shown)


src/prefect/tasks/azure/blobstorage.py
```diff
@@ -75,6 +75,12 @@ class BlobStorageUpload(Task):
         - azure_credentials_secret (str, optional): the name of the Prefect Secret
             that stores your Azure credentials; this Secret must be an Azure connection string
         - container (str, optional): the name of the Azure Blob Storage to upload to
+        - overwrite_blob (bool, optional): Whether the blob to be uploaded should overwrite
+                    the current data. If True, upload_blob will overwrite the existing data.
+                    Note for Append blob types: if set to False and the data already exists,
+                    an error will not be raised and the data will be appended to the existing
+                    blob; if set overwrite=True, then the existing append blob will be deleted,
+                    and a new one created.
```
madkinsz, line 83:
```suggestion
                    If the blob is an 'append' type blob and this is `True`, it will be
                    deleted instead of appended.
```
And the same below. I don't think that we need to explain that it will append as normal if this is not set.

madkinsz (changes requested):
Hey @jnoynaert -- thanks for contributing. Can you add a test and changes entry as well?

PR #4752, opened 2021-07-09
Expose schedule activation from `prefect register` CLI

gmontanola:
## Summary
Enables toggling schedule activation upon CLI flow registering. Useful for CI and development

Related issue: https://github.com/PrefectHQ/prefect/issues/4531

## Changes
1. New CLI option `--schedule/--no-schedule` for the `register` command. Defaults to `--schedule`.
2. This option is passed down util it reaches `register_serialized_flow` and adds a new `input` called `et_schedule_active` to GraphQL request.

## Importance
This will help with some CI use cases.

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/build_register.py
+++ b/src/prefect/cli/build_register.py
@@ -746,8 +746,10 @@ REGISTER_EPILOG = """
 @click.option(
     "--schedule/--no-schedule",
     help=(
-        "Toggles the flow schedule activation upon registering. Default "
-        "behavior is activated. Useful for CI and development."
+        "Toggles the flow schedule upon registering. By default, the "
+        "flow's schedule will be activated and future runs will be created. "
+        "If disabled, the schedule will still be attached to the flow but "
+        "no runs will be created until it is activated."
     ),
     default=True,
 )
--- a/src/prefect/cli/build_register.py
+++ b/src/prefect/cli/build_register.py
@@ -349,6 +349,8 @@ def register_serialized_flow(
         - force (bool, optional): If `False` (default), an idempotency key will
             be generated to avoid unnecessary re-registration. Set to `True` to
             force re-registration.
+        - schedule (bool, optional): If `True` (default) activates the flow schedule
+            upon registering.
 
     Returns:
         - flow_id (str): the flow id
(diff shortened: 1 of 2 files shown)


gmontanola:
Hi there folks!

I'm not sure of how to implement tests for this feature. If someone happens to know a nice way, feel free to share it here!

src/prefect/cli/build_register.py
```diff
@@ -338,6 +338,7 @@ def register_serialized_flow(
     serialized_flow: dict,
     project_id: str,
     force: bool = False,
+    schedule: bool = True,
```
madkinsz, line 341: I would prefer this be `set_schedule_active` for clarity

madkinsz: In the `register` CLI code just `schedule` is fine

src/prefect/cli/build_register.py
```diff
@@ -649,6 +660,10 @@ def watch_for_changes(
 \b  Watch a directory of flows for changes, and re-register flows upon change.
 
 \b    $ prefect register --project my-project -p myflows/ --watch
+
+\b  Register a flow found in `flow.py` and disables its schedule.
```
madkinsz, line 664:
```suggestion
\b  Register a flow found in `flow.py` and disable its schedule.
```

src/prefect/cli/build_register.py
```diff
@@ -728,9 +743,19 @@ def watch_for_changes(
     default=False,
     is_flag=True,
 )
+@click.option(
+    "--schedule/--no-schedule",
+    help=(
+        "Toggles the flow schedule activation upon registering. Default "
+        "behavior is activated. Useful for CI and development."
```
madkinsz, line 750:
```suggestion
        "Toggles the flow schedule upon registering. By default, the "
        "flow's schedule will be activated and future runs will be created. "
        "If disabled, the schedule will still be attached to the flow but "
        "no runs will be created until it is activated."
```

madkinsz (changes requested):
Thanks for doing this! Been on my list for a bit now.

There should be two tests for this, I think:

- One test where `prefect register` is called with `--schedule` and `--no-schedule` and the `Client` is mocked. Assert that `Client.register_serialized_flow` is called with the correct `set_schedule_active` value
- One test where `Client.graphql` is mocked and `Client.register_serialized_flow` is called with each value for `set_schedule_active`. Assert that `Client.graphql` is called with the correct flag value.

src/prefect/cli/build_register.py
madkinsz: Disregard, I forgot this was specific to `build_register`

madkinsz:
@gmontanola Whoops, I forgot that Jim created a non-Client method to register the flow. I'll edit my comment above to reflect the correct tests to add.

PR #4753, opened 2021-07-09
Fix get cloud url bug

zangell44:
## Summary

This PR addresses edge cases in using `Client.get_cloud_url` when using legacy API tokens.

## Changes

Simplifies the logic for `get_cloud_url` by
- eliminating the `as_user` parameter
- adding a check for edge case where `Client.tenant_id` is `None`

I think we can do without the `as_user`. For legacy API tokens, the token will only have permission to view one tenant and the correct tenant slug will be returned. For API Keys, `tenant_id` should be correctly populated and we can determine correct tenant slug based on the query.

## Importance

Makes current code backwards compatible with API Tokens.

## Checklist

This PR:


zangell44:
resolved in #4758

PR #4760, opened 2021-07-12
Install flaky to auto-rerun known/marked flaky tests

madkinsz:
It is now consuming an unacceptable amount of time to re-run CI due to a few tests that always succeed on re-run. I'd rather fix the tests than auto-rerun them but that isn't entirely feasible at the moment and some tests are flaky by nature. At the very least, this will indicate to us which tests need additional attention when we have more throughput.


madkinsz:
Re-running CI a couple of times here in hopes of catching some more obvious flakes.

zangell44 (approved):

Issue #4726, opened 2021-07-01
-r flag on prefect register CLI command


bennnym:
## Current behavior
Today I am unable to look in a nested folder structure for flows. I need to specify multiple folders and run multiple commands so that I can register these flows in CI.

This is frustrating because I then need to rebuild and push the same Docker image that I am using for Storage multiple times. It is inefficient.

My folder structure is like so

```
.
├── flows
│   ├── common
│   └── delete_markets
│       ├── flow.py
│       └── tasks.py
│   └── update_markets
│       ├── flow.py
│       └── tasks.py
 ```

## Proposed behavior
can we add a -pr flag or something to tell the CLI a path to recursively search for flows and only build the one Docker storage image?

madkinsz:
I'm interested in this but I'm hesitant because of the implications of recursion when using the `--watch` flag. Is there a reason that `-p flows/delete_markets -p flows/update_markets` is infeasible for you?

bennnym:
The reason is from a CI perspective.

If I'm registering multiple flows I don't want to just automate a huge register command with 10 x -p foo. 

I have got around it with this for now, but I see it being a common need potentially for others.

PR #4761, opened 2021-07-12
Allow `prefect build/register` paths to contain globs for recursion

madkinsz:
## Summary

Allows `prefect build/register --path` to contain globs like `*` and `**` to recursively select files to load flows from. Will require users to wrap the it in quotes like `--path "**/flow.py"` so their shell does not expand it. If the shell expands it, click gives a useless error about the command not being found. https://github.com/PrefectHQ/prefect/pull/4761/commits/25ae0c5ed0b36ef4148fc04368ced1d957d63c3b implements this differently by allowing paths to be passed as arguments rather than "options" which allows you to just use native shell expansion to pass the files to check. I could be convinced either implementation is proper but I think globbing using Python and keeping the interface consistent might make the most sense.

## Changes

- Adds glob support to path expansion in build/register
- Expands paths before `--watch` functionality is called (I'm not sure why this was not done before and it may have some scary side-effect)

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/4726

## Checklist

This PR:


madkinsz:
@zangell44 thoughts? Holding off on tests until someone says it makes sense

zangell44:
I think it makes sense 👍

zangell44 (approved):

PR #4766, opened 2021-07-13
Fix  broken cron schedule link in Schedule Docs

fakela:
## Summary

the cron link in this section of the docs seems to be broken  #4660

## Changes

Fixes the broken link

## Importance

It is important as it fixes 404 link

## Checklist

This PR:


madkinsz (approved):
Hm, I'm not sure we should be linking to the old examples but this seems fine for now.

PR #4767, opened 2021-07-13
Git storage extra files

### LINKED ISSUE #4754: document loading external files with `Git` storage
## Current behavior

No docs

## Proposed behavior

Create an example doc illustration how to use `Git` storage to load external files (non-python files) from a repo with `Git` storage

zangell44:
## Summary

This PR adds support and documentation for loading external files using `Git` storage.

## Changes

- When extracting a flow from a file, pass the file path as `__file__`
- Document `Git` storage use case where additional files are loaded

## Importance

Closes #4754 

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/storage.py
+++ b/src/prefect/utilities/storage.py
@@ -82,22 +82,24 @@ def extract_flow_from_file(
         raise ValueError("Provide either `file_path` or `file_contents`.")
 
     # Load objects from file into dict
-    exec_vals = {}  # type: ignore
+    # if a file_path has been provided, provide __file__ as a global variable
+    # so it resolves correctly during extraction
+    exec_vals = {"__file__": file_path} if file_path else {}  # type: ignore
     exec(contents, exec_vals)
 
     # Grab flow name from values loaded via exec
-    flows = {o.name: o for o in exec_vals.values() if isinstance(o, prefect.Flow)}
+    flows = {o.name: o for o in exec_vals.values() if isinstance(o, prefect.Flow)}  # type: ignore
     if flows:
         if flow_name:
             if flow_name in flows:
-                return flows[flow_name]
+                return flows[flow_name]  # type: ignore
             else:
                 flows_list = "\n".join("- %r" % n for n in sorted(flows))
                 raise ValueError(
                     f"Flow {flow_name!r} not found in file. Found flows:\n{flows_list}"
                 )
         else:
-            return list(flows.values())[0]
+            return list(flows.values())[0]  # type: ignore
     else:
         raise ValueError("No flows found in file.")
 
--- a/src/prefect/utilities/storage.py
+++ b/src/prefect/utilities/storage.py
@@ -84,7 +84,7 @@ def extract_flow_from_file(
     # Load objects from file into dict
     # if a file_path has been provided, provide __file__ as a global variable
     # so it resolves correctly during extraction
-    exec_vals = {"__file__": file_path} if file_path else {}  # type: Dict[str, Any]
+    exec_vals: Dict[str, Any] = {"__file__": file_path} if file_path else {}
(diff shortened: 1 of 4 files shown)


tests/utilities/test_storage.py
```diff
@@ -61,6 +61,20 @@ def flow_path(self, tmpdir):
 
         return full_path
 
+    @pytest.fixture
+    def flow_path_with_additional_file(self, tmpdir):
+        contents = """from prefect import Flow\nfrom pathlib import Path\nwith open(str(Path(__file__).resolve().parent)+"/test.txt", "r") as f:\n\tname=f.read()\nf2=Flow(name)"""
```
madkinsz, line 66: Can you change this to a multiliner with `textwrap.dedent?` so it's easy to read?

src/prefect/utilities/storage.py
```diff
@@ -82,22 +82,24 @@ def extract_flow_from_file(
...
     # Load objects from file into dict
-    exec_vals = {}  # type: ignore
+    # if a file_path has been provided, provide __file__ as a global variable
+    # so it resolves correctly during extraction
+    exec_vals = {"__file__": file_path} if file_path else {}  # type: ignore
     exec(contents, exec_vals)
 
     # Grab flow name from values loaded via exec
-    flows = {o.name: o for o in exec_vals.values() if isinstance(o, prefect.Flow)}
+    flows = {o.name: o for o in exec_vals.values() if isinstance(o, prefect.Flow)}  # type: ignore
```
madkinsz, line 91: Can we type-cast instead of ignoring?

src/prefect/utilities/storage.py
```diff
@@ -82,22 +82,24 @@ def extract_flow_from_file(
         raise ValueError("Provide either `file_path` or `file_contents`.")
 
     # Load objects from file into dict
-    exec_vals = {}  # type: ignore
+    # if a file_path has been provided, provide __file__ as a global variable
+    # so it resolves correctly during extraction
+    exec_vals = {"__file__": file_path} if file_path else {}  # type: ignore
```
madkinsz, line 87: Unrelated to your PR: Can we audit if `file_contents` is used and perhaps just simplify by dropping it? I feel like composing something with `extract_flow_from_source(code: str, variables: dict = None)` would make more sense than rolling these into a single method as has been done here.

zangell44: `file_contents` is used quite frequently for other storage, I'm happy to refactor if you'd like but I'd want to do so in another PR

madkinsz: Alas. This actually makes sense for remote file based storage so I don't see refactoring it happening anytime soon.

madkinsz (approved):

chicago-joe:
Will this work with AWS CodeCommit storage as well?

Trymzet:
This will break on a Jupyter notebook with a `NameError` because Jupyter does not populate `__file__`.

zangell44:
@chicago-joe this will not work with CodeCommit, only `Git` storage at the moment. Happy to chat a bit further about CodeCommit enhancements or workarounds if you'd like to open an issue

zangell44:
@Trymzet you're correct that Jupyter notebooks (and other execution enironments) do not populate `__file__`. However, I'm not sure I understand the Prefect specific use case you're describing. 

We set `__file__` when the Flow is loaded from storage to allow users to access `__file__` for loading additional files https://github.com/PrefectHQ/prefect/pull/4767/files#diff-8b9309d6d1ca7b8dd179ef8df5c1e02aa0011a639f2d1753f8eee09ca08cc133R87.

For any code executed outside of loading the flow from storage (which should happen when an agent picks up the flow run), `__file__` is not used

PR #4774, opened 2021-07-14
fix query bug when getting all task runs

pjanowski:
## Summary
There is a bug in the new backend function to get all task runs for a FlowRunView object. This is an attempt to fix it. The resulting query is successful but it returns two instances of every task. Would appreciate someone taking a look (@madkinsz) to see if that's intentional (or if the cached filter is not working as intended).

I'm not writing a test just yet until one of the core devs takes a look to confirm this is going in the right direction.

## Changes
The current code has bad syntax in the dictionary and uses an uncrecognized keyword for the GraphQL query.

## Importance
Currently FlowRunView.get_all_task_runs() as documented [here](https://docs.prefect.io/orchestration/flow-runs/inspection.html#python) fails.

## Checklist

This PR:

### DIFF
--- a/src/prefect/backend/flow_run.py
+++ b/src/prefect/backend/flow_run.py
@@ -722,7 +722,7 @@ class FlowRunView:
         task_run_data = TaskRunView._query_for_task_runs(
             where={
                 "flow_run_id": {"_eq": self.flow_run_id},
-                "task_run_id": {"_not", {"_in": list(self._cached_task_runs.keys())}},
+                "task_id": {"_nin": list(self._cached_task_runs.keys())},
             }
         )
         task_runs = [TaskRunView._from_task_run_data(data) for data in task_run_data]


src/prefect/backend/flow_run.py
```diff
@@ -722,7 +722,7 @@ def get_all_task_runs(self) -> List["TaskRunView"]:
         task_run_data = TaskRunView._query_for_task_runs(
             where={
                 "flow_run_id": {"_eq": self.flow_run_id},
-                "task_run_id": {"_not", {"_in": list(self._cached_task_runs.keys())}},
+                "task_id": {"_nin": list(self._cached_task_runs.keys())},
```
madkinsz, line 725: Thanks for catching this! I believe this should be

```suggestion
                "id": {"_nin": list(self._cached_task_runs.keys())},
```

since we're querying the task run table and excluding task run ids not task ids.

This looks right; I can confirm this works as intended locally as well since all of our tests for this mock the backend API

pjanowski: Thanks Michael, a couple of things:
1) changed to `task_id` to `id`, that makes sense.
2) I'm still getting two instances for every task. eg for a 4-node flow:
```
0:TaskRunView(task_run_id='8c320aee-dc45-4345-8d7f-9473cb334317', task_id='ba55c7f3-bc7c-4a2a-911b-280a309b6567', task_slug="noop['a']-1", state=<Success: "Task run succeeded.">, result=1)
1:TaskRunView(task_run_id='de059738-656e-4dfe-8cbb-37c89ac24337', task_id='5edf0ea5-92fa-4a62-ad0f-34edde717ab1', task_slug='noop-1', state=<Success: "Task run succeeded.">, result={'a': 1})
2:TaskRunView(task_run_id='78ee8ff2-ebc5-4e7e-a4ec-5c5754a032fe', task_id='92e1bfaa-99b5-49fb-a4ba-3aa918488782', task_slug="x['a']-1", state=<Success: "Task run succeeded.">, result=1)
3:TaskRunView(task_run_id='29505fad-afb2-46d2-acca-5df47aa11ef5', task_id='593edfb8-3bbb-456c-ae05-c74503bab709', task_slug='x', state=<Success: "Task run succeeded.">, result={'a': 1})
4:TaskRunView(task_run_id='8c320aee-dc45-4345-8d7f-9473cb334317', task_id='ba55c7f3-bc7c-4a2a-911b-280a309b6567', task_slug="noop['a']-1", state=<Success: "Task run succeeded.">, result=1)
5:TaskRunView(task_run_id='de059738-656e-4dfe-8cbb-37c89ac24337', task_id='5edf0ea5-92fa-4a62-ad0f-34edde717ab1', task_slug='noop-1', state=<Success: "Task run succeeded.">, result={'a': 1})
6:TaskRunView(task_run_id='78ee8ff2-ebc5-4e7e-a4ec-5c5754a032fe', task_id='92e1bfaa-99b5-49fb-a4ba-3aa918488782', task_slug="x['a']-1", state=<Success: "Task run succeeded.">, result=1)
7:TaskRunView(task_run_id='29505fad-afb2-46d2-acca-5df47aa11ef5', task_id='593edfb8-3bbb-456c-ae05-c74503bab709', task_slug='x', state=<Success: "Task run succeeded.">, result={'a': 1})
```

3) why did the windows unit tests fail below?
4) `I can confirm this works as intended locally as well since all of our tests for this mock the backend API`: does that mean there are already tests, i.e. I don't need to add a test for this?

madkinsz: 2. It looks like this is because we add all the newly returned tasks to the cache then return the new tasks plus the cache. The return value should be determined before the cache is updated.
3. There's something wrong with CI. We're looking into it, you can ignore it for now.
4. The tests for this class are at `tests/backend/test_flow_run.py` but this is actually not covered by unit tests. You can look at how the other tests work there and try to write a test or I can help do so.

pjanowski: `It looks like this is because we add all the newly returned tasks to the cache then return the new tasks plus the cache. The return value should be determined before the cache is updated.` 
Updated, let me know if that looks good to you. I think we should simply return the cache which at that point should be complete (with both old cached and updated tasks).

src/prefect/backend/flow_run.py
```diff
@@ -730,7 +730,7 @@ def get_all_task_runs(self) -> List["TaskRunView"]:
         for task_run in task_runs:
             self._cache_task_run_if_finished(task_run)
 
-        return task_runs + list(self._cached_task_runs.values())
+        return list(self._cached_task_runs.values())
```
madkinsz, line 733: This won't suffice because some of the task runs may not be finished

pjanowski: See now. Also, (apologies if this suggestion should go somewhere else, let me know), I think what's still missing here is that child flows are treated differently then the parent flow when you want to get the results of tasks. For example if I have

```
with Flow("parent-flow") as flow2:
    res = create_flow_run(flow_name="mysubflow", project_name="myproject")
    res2 = mytask()

res3 = flow2.run() #prefect.engine.State ()

#to get result of task in parent flow I can use the name of the task in the flow
res3.result[res2].result

# but to get the result of a task in the child flow
subflow_id = res3.result[res].result
frv=FlowRunView.from_flow_run_id(subflow_id)
frv.get_task_run(task_slug="x").get_result()
```

So in one case to get at the result I need to know the assigned variable names in the flow declaration but in the other case I need to know the slugs or task_id's in the registered subflow.  Am I missing something?

src/prefect/backend/flow_run.py
```diff
@@ -728,9 +733,10 @@ def get_all_task_runs(self) -> List["TaskRunView"]:
         task_runs = [TaskRunView._from_task_run_data(data) for data in task_run_data]
         # Add to cache
         for task_run in task_runs:
-            self._cache_task_run_if_finished(task_run)
+            if self._cache_task_run_if_finished(task_run):
+                task_runs.remove(task_run)
 
-        return list(self._cached_task_runs.values())
+        return task_runs + list(self._cached_task_runs.values())
```
madkinsz: I would prefer something like this instead of modifying `_cache_task_run_if_finished` just for this purpose. I think it's clearer as well

```python
        new_task_runs = [TaskRunView._from_task_run_data(data) for data in task_run_data]
        
        # Combine the retrieved task runs with the cached ones to create a return value
        # before caching the new ones to avoid returning duplicates
        task_runs = new_task_runs + list(self._cached_task_runs.values())
        
        # Add new runs to cache
        for task_run in new_task_runs:
            self._cache_task_run_if_finished(task_run)

        return task_runs
```
        
Thanks again for working on this!  It's get to see someone get involved in some of the lower-level parts.

pjanowski: Michael, I made a commit following your suggestion, dropped the comments as the code is self-explanatory. 

Currently we're blocked on our Prefect instance, so for adding the test instance you can wait until we're unblocked (might be a while) or add it yourself. You'd probably have an easier time mocking the task_info body for the Client.

Also, wondering what you think about the comment I made above regarding how the child flow task results are retrieved vs the parent flow.

Lastly, I haven't been able to successfully map `create_child_flow` (using `.map(params)` or `applied_map()`) Just curious if you know about that scenario and if it's supported directly. If not I'll ask in Slack.

madkinsz: That all makes sense.

Mapping should work but I haven't tried it. If you can open an issue with the details / errors that'd be great!

src/prefect/backend/flow_run.py
madkinsz: Yeah unfortunately this is because of how/when task slugs are generated. I originally implemented something that would do the lookup by task object but it doesn't work consistently. This is something that we want to polish in the future but there are some lower-level change necessary first.

madkinsz (commented):
Let's drop the changes to `_cache_task_run_if_finished` then it looks good.

I can write a test for this after it's merged. Do you want to add a changelog entry for this fix?

src/prefect/backend/flow_run.py
pjanowski: We figured mapping out now, it's working!

pjanowski:
Ok, I added the test. Note that the original suggestion wouldn't work because it would still double count any tasks that had already been cached. I think that's why originally chose to modify `_cache_task_if_finished`. Now I added logic to parse the `new_task_runs` for those that have not been cached yet.

madkinsz:
> Note that the original suggestion wouldn't work because it would still double count any tasks that had already been cached.

The `_nin` clause should exclude us from receiving any tasks that have already been cached. You'd have to account for this in the test data that's returned. I'll try to give this a test today as well.

pjanowski:
You're right. See now. Apologies for all the iterations on this.

pjanowski:
@madkinsz kind ping on this, we'd like to take a production dependency on this code. Thanks!

madkinsz:
@pjanowski  Can you merge `master` into this branch so it passes CI?

madkinsz (approved):

madkinsz:
Thanks!

PR #4777, opened 2021-07-14
Docker run host config

ngriffiths13:
## Summary

This PR allows to pass runtime commands to the Docker Agent via a host_config dictionary in the DockerRun config.

## Changes

The DockerRun config allows for an additional argument which is the host_config. The Docker agent
adds the provided host_config to the default host_config, overwriting any defaults.

## Importance

This PR will allow for more flexibility in how Prefect Flows are deployed with Docker.

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -403,7 +403,7 @@ class DockerAgent(Agent):
         container_mount_paths = self.container_mount_paths
         if container_mount_paths:
             host_config.update(binds=self.host_spec)
-        if run_config and run_config.host_config:
+        if run_config is not None and run_config.host_config:
             # The host_config passed from the run_config will overwrite defaults
             host_config.update(run_config.host_config)
 
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -403,7 +403,7 @@ class DockerAgent(Agent):
         container_mount_paths = self.container_mount_paths
         if container_mount_paths:
             host_config.update(binds=self.host_spec)
-        if run_config is not None:
+        if run_config is not None and run_config.host_config is not None:
             # The host_config passed from the run_config will overwrite defaults
             host_config.update(run_config.host_config)
 
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -403,6 +403,9 @@ class DockerAgent(Agent):
         container_mount_paths = self.container_mount_paths
         if container_mount_paths:
             host_config.update(binds=self.host_spec)
+        if run_config is not None:
+            # The host_config passed from the run_config will overwrite defaults
+            host_config.update(run_config.host_config)
 
         networking_config = None
         # At the time of creation, you can only connect a container to a single network,
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -403,7 +403,7 @@ class DockerAgent(Agent):
         container_mount_paths = self.container_mount_paths
         if container_mount_paths:
             host_config.update(binds=self.host_spec)
-        if run_config is not None and run_config.host_config is not None:
+        if run_config and run_config.host_config:
             # The host_config passed from the run_config will overwrite defaults
             host_config.update(run_config.host_config)
 
--- a/src/prefect/run_configs/docker.py
+++ b/src/prefect/run_configs/docker.py
@@ -13,6 +13,9 @@ class DockerRun(RunConfig):
         - labels (Iterable[str], optional): an iterable of labels to apply to this
             run config. Labels are string identifiers used by Prefect Agents
             for selecting valid flow runs when polling for work
+        - host_config (dict, optional): A dictionary or runtime arguments to pass to
+            the Docker Agent. See link below for options.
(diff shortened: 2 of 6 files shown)


src/prefect/run_configs/docker.py
```diff
@@ -30,7 +33,13 @@ class DockerRun(RunConfig):
     """
 
     def __init__(
-        self, *, image: str = None, env: dict = None, labels: Iterable[str] = None
+        self,
+        *,
+        image: str = None,
+        env: dict = None,
+        labels: Iterable[str] = None,
+        host_config: dict = {}
```
madkinsz, line 41:
```suggestion
        host_config: dict = None
```
This will raise complaints in our static analysis; you'll want to set the default in the __init__ method body. I'd be consistent with whatever is being done with `env`

src/prefect/agent/docker/agent.py
```diff
@@ -403,6 +403,9 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
         container_mount_paths = self.container_mount_paths
         if container_mount_paths:
             host_config.update(binds=self.host_spec)
+        if run_config is not None and run_config.host_config is not None:
+            # The host_config passed from the run_config will overwrite defaults
+            host_config.update(run_config.host_config)
```
madkinsz, line 408: No need for an explicit comparison to `None` (which is generally frowned upon in PEP recommendations, although you'll see mixed styles in this repo)
```suggestion
        if run_config is not None and run_config.host_config:
            # The host_config passed from the run_config will overwrite defaults
            host_config.update(run_config.host_config)
```

You could also do this, but I think the above is better
```python
        if run_config is not None:
            # The host_config passed from the run_config will overwrite defaults
            host_config.update(run_config.host_config or {})
```

madkinsz (commented):
Does it look there's a reasonable place to assert that this is merged and called correctly in the agent? I know some of the docker-related tests are missing coverage.

src/prefect/agent/docker/agent.py
ngriffiths13: Would it be preferred to remove both of the comparisons to None? I was originally just trying to replicate how it worked with the env_vars. But I could also do something like this:
```suggestion
        if run_config and run_config.host_config:
            # The host_config passed from the run_config will overwrite defaults
            host_config.update(run_config.host_config)
```

ngriffiths13: I will also take a closer look at the docker agent tests and see if there is a good place to check this. There wasn't an obvious one at a quick glance. Would you like me to write a new one if there aren't any good places to check the merged dictionary worked?

madkinsz: The opinions here have varied on this one, hence the different messages :) I'm fine with either choice for the `host_config` value check, I'd rather not touch the `run_config` value check.

I think you can test this in `tests/agent/test_docker_agent.py::test_docker_agent_deploy_flow_run_config` or a copy of it extended for this case. `api.create_container.call_args[1]["host_config"]` should contain the settings of interest.

ngriffiths13: @madkinsz So with the testing I am in a tricky spot. The api.create_container is a mock that returns a set value. Seems like the Docker SDK has a lot of overhead that isn't wanted in the tests. Which means it doesn't actually check if the dictionary merged into the host_config. The host config is also not returned anywhere else so also not possible to test. 

One option is to pull out the ~10 lines of code in the `deploy_flow` method of `DockerAgent` and create a new method called `create_host_config` that can be tested. What are your thoughts on that? These lines would get pulled out into their own function:

```
host_config = {
            "auto_remove": True,
            # Compatibility for linux -- https://github.com/docker/cli/issues/2290
            "extra_hosts": {"host.docker.internal": "host-gateway"},
        }  # type: dict
container_mount_paths = self.container_mount_paths
if container_mount_paths:
  host_config.update(binds=self.host_spec)
if run_config is not None and run_config.host_config:
  # The host_config passed from the run_config will overwrite defaults
  host_config.update(run_config.host_config)```

Comment (madkinsz) [reply src/prefect/agent/docker/agent.py]: I'm not opposed to moving it out if it's not feasible to check the mock, but it seems like it should be. . I see now that `api.create_container.call_args[1]["host_config"]` will have a mocked, fixed value but `api.create_host_config.call_args` should have the merged host config for inspection still.

File: tests/agent/test_docker_agent.py [hunk 3]
```diff
@@ -476,9 +476,20 @@ def test_docker_agent_deploy_flow_run_config(api, run_kind, has_docker_storage):
 
     if run_kind == "docker":
         env = {"TESTING": "VALUE"}
-        run = DockerRun(image=image, env=env)
+        host_config = {"shm_size": "128m"}
```
madkinsz, line 479: Can you include `auto_remove=False` here so we can confirm it overrides correctly as well?

ngriffiths13: Yes I can. Should I add it in with `pytest.mark.parametrize` so that we  can test both of them fairly easily in the same test?

madkinsz: Since we also set `extra_hosts` we can cover both overriding an existing value and leaving one untouched without any parameterization here

madkinsz (approved):
Great, thanks!

ngriffiths13:
@madkinsz Is there anything else I need to do for this?

madkinsz:
Nope. I'll merge once it passes CI

Issue #4770, opened 2021-07-14
Running dependent flows doc has the same example twice


pjanowski:
The [Running Dependent Flows](https://docs.prefect.io/core/idioms/flow-to-flow.html) doc presents the same code example for both the imperative and functional syntax.

Bojne:
Great flag! I had fixed the change and submitted a PR for this. (update 15 files all at once, details in the PR)

Bojne:
@madkinsz merged PR #4781 already took care of this, this issue is ready to close 👍🏻

PR #4781, opened 2021-07-15
fix: reformat tabs title to fix code block syntax in doc 

Bojne:
## Summary

As @pjanowski reported in https://github.com/PrefectHQ/prefect/issues/4770, in our docs, the code block syntax breaks in multiple documents. This was due to the vuepress tab plugin doesn't take tab titles with a `" "` mark around it. I removed the `" "` mark and the doc syntax got fixed. 

## Changes

For each tab title under `::::tab`, remove the quote notation from the title string. 

Before Fix 
Screen Shot 2021-07-16 at 2 31 31 AM

After Fix 
Screen Shot 2021-07-16 at 2 41 01 AM

## Importance

Remove the confusion of code block example in our documents. 

## Checklist

This PR:


changes/issue4770.yaml
```diff
@@ -0,0 +1,5 @@
+fix:
+  - "Doc code block tab formating - [#4781](https://github.com/PrefectHQ/prefect/pull/4781)"
```
madkinsz, line 2: We actually don't do changelog entries for docs since they aren't real bugfixes

madkinsz (changes requested):
Thanks! Just drop the changelog entry and this is good. You can push with `[ci skip]` in the commit to skip running tests since this is just a docs change.

PR #4800, opened 2021-07-21
Include `host_spec` bind for win32 docker volumes

scoussens:
## Summary

This change was to allow for windows builds to use internal volumes (created outside prefect), and resolve a bug that was not properly incorporating named volumes as the external target

## Changes

allows for the external target to be a named volume

## Importance

windows users were previously unable to use named volumes

## Checklist

This PR:

- [x ] adds new tests (if appropriate)
- [x ] adds a change file in the `changes/` directory (if appropriate)

### DIFF
--- a/changes/pr4800.yaml
+++ b/changes/pr4800.yaml
@@ -1,0 +1,3 @@
+
+fix:
+  - "Fix binding of named volumes to flow containers with Docker agent - [#4800](https://github.com/PrefectHQ/prefect/pull/4800)"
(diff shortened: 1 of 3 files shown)


madkinsz:
Glad this worked for you! This will require some tests changes and a changelog entry. We also ought to check out the unix implementation to make sure it's doing the right thing as well. Let me know if you're going to work on this, otherwise it'll just be on my stack.

scoussens:
I'm not familiar with your testing layout, but I can take a look.

The unix implementation looks to be working the same as the previous
windows implementation, at least regarding this fix, so would need to be
adjusted as well.

Seth

PR #4808, opened 2021-07-22
Attempt to lower the OOM score for hearbeat processes on startup

madkinsz:
## Summary

Users running memory intensive tasks find that their heartbeat process is killed while their flow is running. This means that if the task is long-running, we'll kill it think it's a zombie flow run. In an attempt to reduce the chances of the heartbeat being killed before the flow run process, we adjust the OOM score. This will only be available on linux.

## Changes

- `prefect heartbeat flow-run` sets a low OOM score

## Importance

Less false-positive zombie runs

## Checklist

This PR:

### DIFF
--- a/src/prefect/cli/heartbeat.py
+++ b/src/prefect/cli/heartbeat.py
@@ -1,5 +1,5 @@
 import time
-import os
+import sys
 
 import click
 
@@ -98,13 +98,9 @@ def flow_run(id, num):
     # Lower the OOM score for the hearbeat process so it is not killed before the flow
     # run
     try:
-        if os.path.exists("/proc/self/oom_score_adj"):
+        if sys.platform == "linux":
             with open("/proc/self/oom_score_adj", "wb") as f:
                 f.write("-500".encode("ascii"))
-        else:
-            logger.debug(
-                "Cannot set OOM score for heartbeat thread, `oom_score_adj` not available."
-            )
     except Exception:
         logger.debug("Failed to lower OOM score for hearbeat thread.", exc_info=True)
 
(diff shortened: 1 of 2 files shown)


madkinsz:
@zangell44 thoughts?

src/prefect/cli/heartbeat.py
```diff
@@ -94,6 +95,15 @@ def flow_run(id, num):
     logger = get_logger("heartbeat")
     iter_count = 0
 
+    # Lower the OOM score for the hearbeat process so it is not killed before the flow
+    # run
+    try:
+        if sys.platform == "linux":
+            with open("/proc/self/oom_score_adj", "wb") as f:
+                f.write("0".encode("ascii"))
```
madkinsz, line 103: The score ranges from -1000 to 1000. I initially set it to -500, but a brief internet search makes it look like setting the score below 0 requires admin permissions but I kind of assume it starts at 0? That means that to make this work we may need to set the flow run process `prefect execute flow-run` to a higher value e.g. `100` then set this one back to `0` since it would inherit the parent's setting? Hmm. Not as simple as I hoped.

madkinsz: See a decent description at https://askubuntu.com/a/1349313

madkinsz:
This is feeling a bit complicated and hopefully #4736 catches any intermittent issues. If a flow is memory starved, we can't guarantee good behavior (and really, no program can). Heartbeats that exit with a SIGKILL / -9 will likely be fixed by increasing memory. Perhaps this is something to revisit in a future design.

PR #4809, opened 2021-07-23
Only set `host-gateway` if docker version is compatible

madkinsz:
Quick fix that will need some real testing on linux systems; we may want to warn if we cannot set the value. Alternatively, we could just expose the host-config on the docker agent CLI and allow a user to override this field.

### DIFF
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -6,7 +6,7 @@ import re
 import warnings
 from slugify import slugify
 from sys import platform
-from typing import TYPE_CHECKING, Dict, Iterable, List, Tuple
+from typing import TYPE_CHECKING, Dict, Iterable, List, Tuple, Any
 
 from prefect import config, context
 from prefect.agent import Agent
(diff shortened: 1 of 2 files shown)


madkinsz:
@cicdw thoughts on doing something like this? It feels weird but lots of people are not on our "required" minimum version.

src/prefect/agent/docker/agent.py
```diff
@@ -398,11 +398,18 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
 
-        host_config = {
-            "auto_remove": True,
+        # By default, auto-remove containers
+        host_config: Dict[str, Any] = {"auto_remove": True}
+
+        docker_engine_version = tuple(
+            int(i) for i in self.docker_client.version()["Version"].split(".")
+        )
+        if docker_engine_version >= (20, 10, 0):
```
cicdw, line 407: For this version check I recommend using 
```python
import packaging

if packaging.version.parse(self.docker_client.version()["Version"]) < packaging.version.parse("20.10.0"):
```
and emphasizing / conditioning on the "bad version" instead of the good one.

cicdw (commented):
@madkinsz this seems OK / safe in spirit to me; I left one comment and I also think we should raise a warning if the version is less than our recommended version

src/prefect/agent/docker/agent.py
madkinsz: Whoa packaging.version.parse 😮 I was mostly concerned about my bad parser so this looks great.

tests/agent/test_docker_agent.py
```diff
@@ -488,28 +494,40 @@ def test_docker_agent_deploy_flow_run_config(api, run_kind, has_docker_storage):
...
             "auto_remove": True,
-            "extra_hosts": {"host.docker.internal": "host-gateway"},
         }
         run = None if run_kind == "missing" else UniversalRun()
 
+    if docker_engine_version == "20.10.0":
+        exp_host_config["extra_hosts"] = {"host.docker.internal": "host-gateway"}
+        warns_on_old_version = nullcontext()
+    else:
+        warns_on_old_version = pytest.warns(
```
cicdw, line 504: Conditional logic within a test that actually splits into different asserts is in general brittle; it's very easy to update the test / a fixture and not actually test both paths without ever realizing it.  In this case, the solution would be to preserve `test_docker_agent_deploy_flow_run_config` as the happy path, and introduce a new test `test_docker_agent_warns_on_old_docker_versions` that tests for the warning.

cicdw (changes requested):
request to split the test into two tests, otherwise LGTM 👍

tests/agent/test_docker_agent.py
madkinsz: Sounds good to me

madkinsz:
@anticorrelator review me please

anticorrelator (approved):
lgtm

PR #4815, opened 2021-07-25
Allow custom cursor types in SnowflakeQuery task

austen-bouza:
Adds a 'cursor_type' keyword argument to the run method of the
SnowflakeQuery task. This change allows other cursor types such as
DictCursor to be used, enabling queries to be returned using different
types of data structures.

## Summary

Enables different cursor types in the SnowflakeQuery task.

## Changes

Explicitly defines the cursor type when constructing a Snowflake cursor, and allows that value to be overridden using a keyword argument defined in the task's `run` method.

## Importance

This change allows data from a Snowflake query to be returned in data structures other than a list of tuples. This is useful when working with large datasets that need to be mutable, as it saves the extra processing time and memory consumption required to copy the returned data into a different data structure.

## Checklist

This PR:

- [ X] adds new tests (if appropriate)
- [ X] adds a change file in the `changes/` directory (if appropriate)
- [ X] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)


austen-bouza:
I wanted to add a unit test for this change, but found it difficult to introspect the cursor type when calling the `run` method. Mocking a cursor directly defeats the purpose of the test, since this value should be derived from the new `cursor_type` keyword argument. The return value of `fetchall` would also need to be dependent on the cursor type, so I couldn't directly hardcode that value in advance of calling `run`.

madkinsz:
Hey @austen-bouza -- I think mocking `conn` and assuring `conn.cursor` is called with the given cursor type should be sufficient for a unit test. We just want to make sure that there's coverage for correct use of the passed in type.

madkinsz (approved):
Thanks!

PR #4817, opened 2021-07-26
Ignore schedules by default for CLI flow runs and add flag to run based on schedule

tchoedak:
## Summary
This PR introduces a new flag in `cli.run.run`, `--schedule` or `-s` that allows a cli run against a path or module for a flow to use the schedule within it. By making it a flag, by default these schedules will be ignored which is the expectation from this #4763 .

## Changes
- Adds new flag in `cli.run.run`: `--schedule/-s`
- updates the cli run command's `flow.run` call to not run on schedule by default.

## Importance
Based on the issue's conversation, this could be needed because the CLI is intended to be used to trigger actions immediately. Currently that's not the behavior if a user wants to run a flow via the CLI if it has a defined schedule.

## Checklist

This PR:


src/prefect/cli/run.py
```diff
@@ -438,6 +438,12 @@ def cast_value(value: str) -> Any:
     ),
     is_flag=True,
 )
+@click.option(
+    "--schedule",
+    "-s",
+    help="Execute the flow run according to the schedule defined with the flow.",
```
madkinsz, line 444:
```suggestion
    help="Execute the flow run according to the schedule attached to the flow.",
```

And perhaps something like..

> When this flag is set, this command will wait between scheduled flow runs.
> If the flow has no schedule, this flag will be ignored.
> If used with a non-local run, an exception will be thrown.

madkinsz: We should also throw an exception if someone passes this flag for a non-local run.

madkinsz (commented):
This is a good start, I'll review again when the tests are done. Ping me if you need help :)

madkinsz:
Hey @tchoedak if this is working by Tuesday we can get it in the next release :) let me know if you need anything

src/prefect/cli/run.py
```diff
@@ -557,6 +569,12 @@ def run(
         )
     params_dict = {**file_params, **cli_params}
 
+    if schedule and flow_or_group_id:
```
tchoedak, line 572: Can I assume that checking `if flow_or_group_id` is sufficient to identify that it's intended to be a cloud run?

madkinsz: Cloud runs can be done with just `name` as well so it's not sufficient. I would move this to right under the `if path or module` check which is when we are certain we are doing a local run.

madkinsz:
Just that minor point. May be worth covering passing `--name` instead of `--id` in your test but I do not feel strongly about it. Thanks again for working on this

src/prefect/cli/run.py
tchoedak: Gotcha, in this case i've moved the check and updated it to check for name as well:
```python
if schedule and (flow_or_group_id or name):
```

src/prefect/cli/run.py
```diff
@@ -623,6 +617,12 @@ def run(
 
     # Backend flow run -----------------------------------------------------------------
 
+    if schedule and (flow_or_group_id or name):
+        raise ClickException(
+            "Received a local only flag for a non-local flow; "
+            "cannot specify both `--schedule` and `--id`"
+        )
+
```
madkinsz, line 625: This error message is a bit confusing if they've just specified name. I'd just do
```suggestion
    if schedule:
        raise ClickException("`--schedule` can only be specified for local flow runs")
```

There's a `return` above this so we're definitely not in a local-run anymore and you don't need to check the other variables either.

madkinsz:
Sorry my change suggestion broke your test message assertion

madkinsz (approved):

PR #4821, opened 2021-07-29
[cli] Prevent server access from the outside

HenriTEL:
## Summary
By default docker-compose services listen to all host interfaces, meaning all machines in the connected networks have direct access to services exposing a port.
This commit sets the binding IP to localhost by default to mitigate this security issue.

## Changes
* Limited services exposing ports to localhost in `prefect server` cli
* Add `--expose` flag to `prefect server` cli to make the Core server and UI listen to all interfaces

## Importance
This is quite important as it fixes a security issue when working with a local environment.

### DIFF
--- a/changes/pr4821.yaml
+++ b/changes/pr4821.yaml
@@ -1,0 +1,8 @@
+breaking:
+  - Services run by `prefect server` cli are now local by default (listen to localhost instead of 0.0.0.0)
+
+enhancement:
+  - Add `--expose` flag to `prefect server` cli to make the Core server and UI listen to all interfaces
+
+contributor:
+  - "[Henri Hannetel](https://github.com/HenriTEL)"
--- a/src/prefect/cli/docker-compose.yml
+++ b/src/prefect/cli/docker-compose.yml
@@ -9,7 +9,7 @@ services:
   postgres:
     image: "postgres:11"
     ports:
-      - "${POSTGRES_HOST_PORT:-5432}:5432"
+      - "${POSTGRES_HOST_IP:-127.0.0.1}:${POSTGRES_HOST_PORT:-5432}:5432"
     environment:
       POSTGRES_USER: ${POSTGRES_USER}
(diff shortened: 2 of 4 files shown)


HenriTEL:
@madkinsz can you review that please ?

madkinsz:
Hey @HenriTEL - there are a few things to address with this PR.

- This appears to be a breaking change; users use this command to deploy Server to external hosts and expect to be able to connect. It's possible I'm misunderstanding so please correct me if I'm wrong.
- You've added environment variables that can be used to configure the host which is nice because now someone can do `UI_HOST_IP=0.0.0.0 prefect server config`. However, users are not expected to know the environment variables that are available in the template `docker-compose.yml`. Most of the environment variables used there are populated from the config or CLI arguments.
- There is no changelog entry

It seems like the best path forward is:
- Retain the existing default of 0.0.0.0 so it is backwards compatible
- Add a flag to the CLI like `--local-only` (not in love with that) that will set those environment variables in the `docker-compose` call to `127.0.0.1` if they have not been set in the environment already. I don't think each service is worth exposing individual in the config/CLI and users that need to override a single service can do so by setting environment variables as you've exposed already.

madkinsz:
Thanks for contributing by the way! We're very busy lately so I'm slow to get to reviewing, but I appreciate it a lot :)

madkinsz:
Hey @HenriTEL, I understand your concern here. We know that people use this command for running non-local servers frequently and we do not want to introduce an unexpected break in their experience without clear documentation. We can introduce a breaking change that

- Sets the default to 127.0.0.1 as you've already implemented
- Adds an `--expose` flag to `prefect server start` so this can be easily toggled to 0.0.0.0

We will then want to update the documentation to make this behavior clear to minimize the confusion for new users that may not understand these concepts.

HenriTEL:
@madkinsz I just updated my PR, can you review the changes ?

madkinsz (approved):
LGTM; thanks!

PR #4844, opened 2021-08-04
Enable threaded heartbeats

anticorrelator:
## Summary

This PR builds an alternate heartbeat implementation, threads instead of a subprocess. The working theory is that some flow-runs will get killed because the heartbeat process is terminated prematurely (https://github.com/PrefectHQ/prefect/pull/4808), using threads might be able to avoid this failure mode.

## Changes

- allows the configuration of the heartbeat implementation
  - subprocess heartbeats can run independent of the GIL, but are more vulnurable
  - threaded heartbeats can't be targeted independently
- Runner classes implement a general interface for enabling heartbeats, but now that heartbeats are strictly flow-only, I've started to undo this
- Removed the `heartbeat task-run` CLI command, it's no longer used

## Notes

- in the PR, I've implemented an event-based thread termination mechanism
  - we might be able to get rid of this and rely on the `daemonization` property of Python threads only, if we know that `flow_runs` always run in their own Python process
- the scope of these changes might be a bit big, as I was picking between multiple ideas of how to scope this change. If it's better, we should narrow the scope to avoid touching the `Runner` classes and how they interface with heartbeat processes

## Checklist

This PR:


src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,48 @@
...
+        iter_count = 0
+        with prefect.context({"flow_run_id": id, "running_with_backend": True}):
+            with log_heartbeat_failure(logger):
+                while iter_count < (self.num or 1) and not self.stop_event.is_set():
+                    self.stop_event.wait(timeout=config.cloud.heartbeat_interval)
+                    send_heartbeat(self.id, client, logger)
+                    iter_count += 1 if self.num else 0
+
+
+def send_heartbeat(id, client, logger):
```
madkinsz, line 29: I'd prefer `flow_run_id` so we don't collide with the python builtin

src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,48 @@
...
+        threading.Thread.__init__(self)
+        self.num = num
+        self.daemon = True  # 'daemonizes' the thread, so Python will terminate it when all non-daemonized threads have finished
+        self.stop_event = stop_event
+
+    def run(self):
+        logger = get_logger('heartbeat')
+        client = Client()
+        iter_count = 0
+        with prefect.context({"flow_run_id": id, "running_with_backend": True}):
```
madkinsz, line 21: id does not have your intended value here

src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,48 @@
...
+            with log_heartbeat_failure(logger):
+                while iter_count < (self.num or 1) and not self.stop_event.is_set():
+                    self.stop_event.wait(timeout=config.cloud.heartbeat_interval)
+                    send_heartbeat(self.id, client, logger)
```
madkinsz, line 25: `self.id` is never set

src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,48 @@
+import threading
+from contextlib import contextmanager
+
+import prefect
+from prefect import config
+from prefect.client import Client
+from prefect.utilities.logging import get_logger
+
+
+class HeartbeatThread(threading.Thread):
```
madkinsz, line 10: I think this might be easier to read as a function inlined into your `threaded_heartbeat` function. Keeping the implementation together seems nice and subclassing creates confusion about which attrs are for your code and which are for `threading.Thread`


```
def threaded_heartbeat(...):

    def send_heartbeat_thread(stop_event, flow_run_id):
         # contents of run here
    
    thread = threading.Thread(send_heartbeat_thread)
    
```

src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,48 @@
...
+        client.update_flow_run_heartbeat(id)
+    except Exception as exc:
+        logger.error(
+            f"Failed to send heartbeat with exception: {exc!r}",
+            exc_info=True,
+        )
+
+
+@contextmanager
+def log_heartbeat_failure(logger):
```
madkinsz, line 40: I think you can take this and `send_heartbeat` and move them into the executor utils. We should use these contexts from the flow run heartbeat CLI as well.

madkinsz (changes requested):
This looks like a good start. Mostly organizational comments.

src/prefect/utilities/executors.py
```diff
@@ -41,50 +43,78 @@ def run_with_heartbeat(
...
-                        stdout=subprocess.DEVNULL,
-                        stderr=subprocess.DEVNULL,
-                    )
-            except Exception:
-                self.logger.exception(
-                    "Heartbeat failed to start.  This could result in a zombie run."
-                )
+
+        if not self._heartbeat():
+            configured_heartbeat = contextlib.nullcontext
```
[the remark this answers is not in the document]
anticorrelator: I think because we need to support 3.6 I'm just going to define a nullcontext by hand and name it

src/prefect/engine/cloud/flow_runner.py
```diff
@@ -72,21 +72,21 @@ def __init__(self, flow: Flow, state_handlers: Iterable[Callable] = None) -> Non
     def _heartbeat(self) -> bool:
         try:
             # use empty string for testing purposes
-            flow_run_id = prefect.context.get("flow_run_id", "")  # type: str
-            self.client.update_flow_run_heartbeat(flow_run_id)
+            self.flow_run_id = prefect.context.get("flow_run_id", "")  # type: str
```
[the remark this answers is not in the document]
cicdw: +1

src/prefect/utilities/executors.py
```diff
@@ -39,52 +51,94 @@ def run_with_heartbeat(
...
-                        stderr=subprocess.DEVNULL,
-                    )
-            except Exception:
-                self.logger.exception(
-                    "Heartbeat failed to start.  This could result in a zombie run."
-                )
+            use_heartbeat = self._heartbeat()
+        except Exception:
+            use_heartbeat = False
+            logger = get_logger()
```
cicdw, line 61: Can you use `self.logger` here (where `self` is the provided runner class)? That will ensure this log gets exposed as a part of the run in the backend.

src/prefect/utilities/executors.py
```diff
@@ -39,52 +51,94 @@ def run_with_heartbeat(
...
+            logger.exception(
+                "Heartbeat process is misconfigured.  This could result in a zombie run."
+            )
```
cicdw, line 64:
```suggestion
            exc_info=True)
```

src/prefect/utilities/executors.py
```diff
@@ -39,52 +51,94 @@ def run_with_heartbeat(
...
+        if not use_heartbeat or (prefect.context.config.heartbeat_mode == "off"):
+            configured_heartbeat = no_heartbeat()
+        elif prefect.context.config.heartbeat_mode == "thread":
+            configured_heartbeat = threaded_heartbeat(self.flow_run_id)
+        elif prefect.context.config.heartbeat_mode == "process":
+            configured_heartbeat = subprocess_heartbeat(self.heartbeat_cmd, self.logger)
+        # because the threaded heartbeat mode is experimental the future, let's not catch configuration
+        # error with `else` -- stale configuration should break tests
+
+        with configured_heartbeat:
```
cicdw, line 75: I like the context manager implementation over the decorator 👏

src/prefect/utilities/executors.py
```diff
@@ -12,11 +13,22 @@
...
+    Union,
+    Sequence,
+    Mapping,
+    Iterator,
+)
 
 import prefect
 from prefect.exceptions import TaskTimeoutSignal
 from prefect.utilities.logging import get_logger
+from prefect.utilities.threaded_heartbeat import HeartbeatThread
```
cicdw, line 31: Is there a reason you put this into its own file?  Seems potentially unnecessary

src/prefect/utilities/threaded_heartbeat.py
```diff
@@ -0,0 +1,65 @@
...
+        num: int = None,
+    ) -> None:
+        threading.Thread.__init__(self)
+        # 'daemonizes' the thread, so it will terminate when all non-daemonized threads have finished
+        self.daemon = True
+        self.flow_run_id = flow_run_id
+        self.num = num
+        self.stop_event = stop_event
+
+    def run(self: "HeartbeatThread") -> None:
```
cicdw, line 26: Minor note that `mypy` doesn't care if you don't annotate `self` on a class method

src/prefect/utilities/executors.py
```diff
@@ -39,52 +51,94 @@ def run_with_heartbeat(
...
+            use_heartbeat = False
+            logger = get_logger()
+            logger.exception(
+                "Heartbeat process is misconfigured.  This could result in a zombie run."
+            )
+
+        if not use_heartbeat or (prefect.context.config.heartbeat_mode == "off"):
+            configured_heartbeat = no_heartbeat()
+        elif prefect.context.config.heartbeat_mode == "thread":
+            configured_heartbeat = threaded_heartbeat(self.flow_run_id)
```
cicdw, line 69: Ah I see, is this why you set the run ID as an attribute?

src/prefect/engine/cloud/flow_runner.py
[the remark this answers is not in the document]
anticorrelator: I wanted to use this interface to get everyone's input on where we want the boundaries for this PR to be. The heartbeat machinery already mutates this class state to present information that contains the `flow_run_id`, previously this took the form of exposing `self.heartbeat_cmd`. I'm not attached to this pattern at all, but in stopping here I had two thoughts: trying to completely remove all of the old heartbeat idiom touches quite a bit of code, as it's tested for several places, and for this stab at drawing a reasonable stopping point, it was to use a similar mechanism to pass information to the new heartbeat (mutating the state once `_heartbeat` is called)--which I do not love. I was going to make another PR against this one just to sketch out what it would look like to go all the way, but if everyone dislikes this pattern as much as I do maybe we should just, at the very least, take this piece out.

src/prefect/utilities/executors.py
anticorrelator: There's no deep reason at all--without knowledge about how to organize things I tried superficially mimic existing structure. The current CLI-heartbeat is in its own file so I put this one in its own too, is there another place that makes more sense?

src/prefect/utilities/executors.py
anticorrelator: yeah, this is the analogy to `self.heartbeat_cmd`, I'd like to remove at least one of them but I want to get input on how deep we want to go in ripping out old idiom now that heartbeats are cloud-flow-run only, as I mentioned above

src/prefect/utilities/threaded_heartbeat.py
anticorrelator: In my recent cleanup I played with writing it out with an inlined `Thread` and a `target` with a handful of arguments, but it ended up just looking confusing to my eyes. `Thread` was designed to be subclassed at times, according to the docs, so if it doesn't feel too wrong for you, would you mind if I kept it this way?

src/prefect/utilities/executors.py
cicdw: I was thinking alongside the subprocess implementation in `utilities/executor.py` so it's colocated with its use

src/prefect/utilities/executors.py
cicdw: Yea I see how this could get complicated; ultimately `flow_run_id` _is_ provided in context so you could extract it from there in this line, but I'm not sure that's any better than the mild class mutation you already have

cicdw:
@anticorrelator this isn't particularly important but since you're cleaning up the old task run heartbeat code could you also remove the `update_task_run_heartbeat` method on the `Client` in `client.py`?

src/prefect/utilities/executors.py
```diff
@@ -39,52 +54,151 @@ def run_with_heartbeat(
...
+        #   submit tasks to run within daemonic subprocesses
+        current_env = dict(os.environ).copy()
+        auth_token = prefect.context.config.cloud.get("auth_token")
+        api_key = prefect.context.config.cloud.get("api_key")
+        tenant_id = prefect.context.config.cloud.get("tenant_id")
+        api_url = prefect.context.config.cloud.get("api")
+        current_env.setdefault("PREFECT__CLOUD__AUTH_TOKEN", auth_token)
+        current_env.setdefault("PREFECT__CLOUD__API_KEY", api_key)
+        current_env.setdefault("PREFECT__CLOUD__TENANT_ID", tenant_id)
+        current_env.setdefault("PREFECT__CLOUD__API", api_url)
```
madkinsz, line 121: There is a utility that make this cleaner at `prefect.utilities.configuration.to_environment_variables`

madkinsz: e.g.
```python
    env.update(
        to_environment_variables(
            prefect.context.config,
            include={
                "logging.level",
                "logging.format",
                "logging.datefmt",
                "cloud.send_flow_run_logs",
            },
        )
    )
```

src/prefect/utilities/executors.py
```diff
@@ -39,52 +54,151 @@ def run_with_heartbeat(
...
+        logger.exception(
+            "Heartbeat process failed to start.  This could result in a zombie run."
+        )
+
+    finally:
+        if p is not None:
+            exit_code = p.poll()
+            if exit_code is not None:
+                msg = "Heartbeat process died with exit code {}".format(exit_code)
+                logger.error(msg)
```
madkinsz, line 141:
```suggestion
                logger.error(f"Heartbeat process died with exit code {exit_code}")
```

src/prefect/config.toml
```diff
@@ -56,6 +56,7 @@ endpoint = "https://api.prefect.io"
 graphql = "${cloud.api}/graphql"
 use_local_secrets = true
 heartbeat_interval = 30.0
+heartbeat_mode = "process" # ['process', 'thread', 'off']
```
madkinsz, line 59: If a user toggles a heartbeat to 'off' isn't it just going to cause all their long running flows to be killed? I guess they could toggle heartbeats _and_ the checking service. Sounds like a complicated toggle for them but 🤷

madkinsz (commented):
This mostly LGTM. Are you going to update any docs?

src/prefect/config.toml
cicdw: I think in general this config is a power user config -- if we hear back positive results from our affected users, the idea would be to migrate entirely to threads

cicdw (commented):
LGTM - not approving since @madkinsz has some outstanding questions / suggestions

madkinsz (dismissed):

docs/orchestration/concepts/services.md
```diff
@@ -35,7 +35,19 @@ Where necessary, flow runs without submitted or running task runs will be resche
 
 ## Zombie Killer
 
-The `Zombie Killer` service is responsible for handling zombies, which Prefect defines as tasks that claim to be running but haven't updated their heartbeat in the past 2 minutes.
+The `Zombie Killer` service is responsible for handling zombies, which Prefect defines as tasks that claim to be running but haven't updated their heartbeat in the past 2 minutes. It's possible for the system kernel to terminate the heartbeat process prematurely and trigger the `Zombie Killer`. To circumvent this, try configuring `Prefect` to run the heartbeat in a thread.
```
madkinsz, line 38: This should note the Server heartbeat timeout as well

madkinsz: I'd also put this note on circumventing in a `### Heartbeat modes` section below the description of the actual service. It feels weird to tell them that the system may misbehave while we're still explaining how the zombie killer service is useful

madkinsz (approved):

PR #4852, opened 2021-08-10
Better rate limit retries

cicdw:
## Summary

During certain large burst events coming from a highly parallelized workflow, API rate limits can be encountered (represented as 429 status codes).  The default retry logic for bad status codes is not well suited for the longer durations over which rate limits are computed, so this PR separates out 429 client handling logic into its own separate custom code path.

## Changes

Adds generous sleeps and retries when the client encounters a 429 status code, maximizing the chances for success when this (rare) situation is encountered.

## Importance

Better UX and workflow stability. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -723,7 +723,7 @@ class Client:
         while rate_limited:
             jitter = random.random() * 10 * (2 ** iter_count)
             naptime = 3 * 60 + jitter  # 180 second sleep + increasing jitter
-            self.logger.debug("Rate limit encountered; sleeping...")
+            self.logger.debug(f"Rate limit encountered; sleeping for {naptime}s...")
             time.sleep(naptime)
             response = self._send_request(
                 session=session, method=method, url=url, params=params, headers=headers
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -1543,6 +1543,24 @@ class Client:
             headers={"X-PREFECT-HEARTBEAT-ID": flow_run_id},
         )
 
+    def update_task_run_heartbeat(self, task_run_id: str) -> None:
+        """
+        Convenience method for heartbeating a task run.
+
+        Does NOT raise an error if the update fails.
+
+        Args:
+            - task_run_id (str): the task run ID to heartbeat
+        """
+        mutation = {
+            "mutation": {
+                with_args(
+                    "update_task_run_heartbeat", {"input": {"task_run_id": task_run_id}}
+                ): {"success"}
+            }
+        }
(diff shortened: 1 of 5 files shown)


src/prefect/client/client.py
```diff
@@ -703,14 +704,35 @@ def _request(
...
+            jitter = random.random() * 10 * (2 ** iter_count)
+            naptime = 3 * 60 + jitter  # 3 + e min sleep
+            self.logger.debug("Rate limit encountered; sleeping...")
+            time.sleep(naptime)
+            response = self._send_request(
+                session=session, method=method, url=url, params=params, headers=headers
+            )
+            rate_limited = response.status_code == 429
+            iter_count += 1
+            if iter_count > 6:
```
madkinsz, line 733: Did you mean for this to be `>=`?

src/prefect/client/client.py
```diff
@@ -703,14 +704,35 @@ def _request(
...
+        # each time we encounter a rate limit, we sleep for
+        # 3 minutes + random amount, where the random amount
+        # is uniformly sampled from (0, 10 * 2 ** rate_limit_counter)
+        # up to (0, 640), at which point an error is raised if the limit
+        # is still being hit
+        rate_limited = response.status_code == 429
+        iter_count = 1
+        while rate_limited:
+            jitter = random.random() * 10 * (2 ** iter_count)
+            naptime = 3 * 60 + jitter  # 3 + e min sleep
```
madkinsz, line 725: The jitter is in seconds right? This comment makes it feel like minutes

src/prefect/client/client.py
```diff
@@ -703,14 +704,35 @@ def _request(
...
+            self.logger.debug("Rate limit encountered; sleeping...")
```
madkinsz, line 726: It might be nice to include the seconds in the log; no strong feelings here
```suggestion
            self.logger.debug(f"Rate limit encountered; sleeping for {naptime}s...")
```

madkinsz (commented):
Makes sense! We could try to release this today

src/prefect/client/client.py
cicdw: Yea the jitter is in seconds, I see what you mean and can update 👍

src/prefect/client/client.py
cicdw: Nope I meant for the strict inequality check; ultimately the question is what's the longest sleep we want to allow for, which I recently realized has other side effects that we need to consider :/

madkinsz (approved):

dwildes-prefect (approved):
LGTM

PR #4869, opened 2021-08-13
Fix binding of named volumes to flow containers with Docker agent

madkinsz:
Supersedes #4800 

## Summary

A `bind` was not being generated in the `host_spec` for named volumes. Consequently, the volumes were attached but not available from inside the container as expected. This PR adds proper `bind` specs for named volumes.

## Changes

- `bind` clauses are generated for named volumes

## Importance

Fixes a bug on win32/unix where named volumes were not available in the flow container

## Example

`prefect agent docker start --volume my_named_volume:/tmp`

## Checklist

This PR:

### DIFF
--- a/tests/agent/test_docker_agent.py
+++ b/tests/agent/test_docker_agent.py
@@ -898,7 +898,7 @@ def test_docker_agent_is_named_volume_win32(monkeypatch, api, path, result):
             ["some-name:/ctr/path"],
             ["some-name"],
             ["/ctr/path"],
-            {},
+            {"some-name": {"bind": "/ctr/path", "mode": "rw"}},
         ),
         (
             # multiple volumes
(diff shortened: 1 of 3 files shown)


madkinsz:
@anticorrelator review me please :)

tests/agent/test_docker_agent.py
```diff
@@ -898,7 +898,7 @@ def test_docker_agent_is_named_volume_win32(monkeypatch, api, path, result):
             ["some-name:/ctr/path"],
             ["some-name"],
             ["/ctr/path"],
-            {},
+            {"some-name": {"bind": "/ctr/path", "mode": "rw"}},
```
anticorrelator, line 901: is this change in coverage intentional vs adding another full case?

madkinsz: The old code was actually just wrong and did not work, so this change is correcting coverage of bad behavior. A bind must be generated alongside a named volume.

anticorrelator (approved):

PR #4870, opened 2021-08-13
Allow custom clone urls with `Git` storage

### LINKED ISSUE #4850: Git Storage is Currently Incompatible with Azure DevOps

**adam.everington**: Hey all, i'm struggling to connect to our Azure Devops Repo using the GIT storage class. I keep getting an HTTP Unauthorized error. I've tried connecting with git credentials and with a PAT and neither work. Has anyone else managed to connect to AZ Devops?

**kevin701**: Hey <@U02A67U1UUA>, I have not seen Azure DevOps come up here before. I did see <https://prefect-community.slack.com/archives/CL09KU1K7/p1628171528402100|this> which was connections to Azure Storage and Azure Container Registry. I don’t know if it’s a similar format. I can try this and get back to you by Monday.

**adam.everington**: Hey Kevin, thanks that would be great, i'll continue to try in the mean time, I know that the connection to Devops can be awkward at the best of times and in other c# apps we have we've had to request a token and then encode and embed that into the request header.

**kevin701**: Hey man, I have figured out the issue after digging a bit. Basically we form the URL is the params provided and the URL takes this form:
`url = f"https://{self.git_token_secret}@{self.repo_host}/{self.repo}.git"`

This works for Github, Bitbucket, and Gitlab where the Git repo cloning ends in `.git` . For Azure DevOps though, the URL looks something like this:
`working_url = "<https://kdykho@dev.azure.com/kdykho/prefect/_git/prefect>"`

That ending `.git` causes the URL to fail, but azure devops has a different URL.

I’ll chat with an engineer about how to go about this.

**kevin701**: <@ULVA73B9P> open “Git Storage is Currently Incompatible with Azure DevOps”

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1628238958463000?thread_ts=1628238958.463000&cid=CL09KU1K7).

zangell44:
## Summary

This PR allows users to pass a secret name containing the exact clone url for `Git` storage.

## Importance

Resolves #4850. I suspect there will be other cases where providers don't use standard syntax for cloning and this logic should cover those cases.

## Checklist

This PR:

### DIFF
--- a/changes/pr4870.yaml
+++ b/changes/pr4870.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Allow custom git clone url for `Git` storage - [#4870](https://github.com/PrefectHQ/cloud/pull/4870)"
--- a/docs/orchestration/flow_config/storage.md
+++ b/docs/orchestration/flow_config/storage.md
@@ -284,6 +284,20 @@ storage = Git(
 )
 ```
 
+`Git` storage will attempt to build the correct git clone url based on the parameters provided. Users can override this logic and provide their git clone url directly.
+
+To use a custom git clone url, first create a Secret containing the url. Next, specify the name of the secret when creating your `Git` storage class.
+
+```python
+# example using Azure devops url
+# using a secret named 'MY_REPO_CLONE_URL' with value 'https://<username>:<personal_access_token>@dev.azure.com/<organization>/<project>/_git/<repo>'
+
+storage = Git(
+    path="flows/my_flow.py",
+    git_clone_url_secret_name="MY_REPO_CLONE_URL" # use the value of this secret to clone the repository
+)
+```
+
 :::tip Git Deploy Keys
 To use `Git` storage with Deploy Keys, ensure your environment is configured to use Deploy Keys. Then, create a `Git` storage class with `use_ssh=True`.
 
--- a/src/prefect/serialization/storage.py
+++ b/src/prefect/serialization/storage.py
@@ -161,6 +161,7 @@ class GitSchema(BaseStorageSchema):
     flow_name = fields.String(allow_none=True)
     git_token_secret_name = fields.String(allow_none=True)
     git_token_username = fields.String(allow_none=True)
+    git_clone_url_secret_name = fields.String(allow_none=True)
     branch_name = fields.String(allow_none=True)
     tag = fields.String(allow_none=True)
     commit = fields.String(allow_none=True)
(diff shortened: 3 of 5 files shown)


madkinsz (commented):
@zangell44 Anything this is waiting for other than a changes entry?

zangell44:
@madkinsz waiting on my changelog + documents.

Kevin was doing some testing with Azure DevOps on this. When he gives the green light I'll open a full PR

sm-Fifteen:
Why does the URL need to be secret? Can't the username and password be specified separately to the git library when pulling, so that the URL can just be a plain parameter to the storage? Having to register a new separate secret per flow coming from the git storage seems like it would get cumbersome pretty quickly.

kvnkho:
Tested and can confirm this worked for me after I made a Cloud secret. Got all of the warnings too. This was the form of my secret:
```
https://kdykho:<personal_access_token>@dev.azure.com/kdykho/prefect/_git/prefect
```
Just some explanation why the URL is like that:

organization - dev.azure.com/kdykho
project - prefect
repo - prefect

kdykho - user attached to personal access token
<personal_access_token> - personal access token created

zangell44:
@sm-Fifteen all fair points!

I agree creating new secrets is cumbersome, but you only have to create one secret per git repository, not per flow. 

The point of `Git` storage is to be agnostic to the underlying git type. I think the flexibility of this solution will benefit users enough to make it worthwhile. As the feature continues to be used, we're happy to hear feedback on how we can further standardize how we construct urls for different hosts and eliminate any excessive user burdens.

madkinsz (approved):

sm-Fifteen:
> I agree creating new secrets is cumbersome, but you only have to create one secret per git repository, not per flow. 

Ah, you're right, my bad.

> The point of `Git` storage is to be agnostic to the underlying git type.

I get that, but from what I can tell, the current PR makes it so specifying the URL at all prevents you from passing the username and password/access token separately; you have to pass them through the URL, which is generally considered a bad practice because URLs often get logged (not to mention how it's extra work for the user to put the URL together by themselves).

The proposed iteration of git storage really only works with HTTP, given how the password secret, `git_token_secret_name`, is ignored when using ssh. That means ssh can only be used for unauthenticated pulling, which I've never personally seen before.

The solution that would make sense to me would be to have the repo URL be a constant, non-secret parameter, maybe with the username component (for urls like `ssh://git@github.com` or simply `git@github.com` and to use that URL as-is with the credentials passed separately, to be included as headers (for HTTP) or however the server requests authentication to work (for SSH, where dulwich can use `password` to either login with a username and password or to decrypt `key_filename` if it is specified).

That really seems to me like the only way to be properly "agnostic to the underlying git type".

PR #4871, opened 2021-08-14
Updating ENDRUN Docstring

kvnkho:
## Summary

I am just changing two lines with the ENDRUN docstring. Multiple users have gotten the expectation that it would shut down the entire Flow because of the way it is currently written. I also just added a use case to use this to avoid task retries.

## Checklist

This PR:

### DIFF
--- a/changes/pr4871.yml
+++ b/changes/pr4871.yml
@@ -1,2 +1,0 @@
-enhancement:
-  - "Clarifying docs that ENDRUN does not end flows - [#4871](https://github.com/PrefectHQ/prefect/pull/4871)"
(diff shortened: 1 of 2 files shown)


madkinsz (approved):

madkinsz:
Although honestly, perhaps we should fix the behavior of `ENDRUN` ? cc @cicdw

cicdw:
`ENDRUN` was never intended to be something that halted all flow run execution (if you consider that when running on Dask, the flow runner asynchronously submits tasks for execution and then waits for their completion, this would be a non trivial communication pattern to implement).  

Given that we have some changes to execution that will allow for much easier short-circuiting logic for flow runs, I don't think we should try to build anything new on top of `ENDRUN`.

PR #4872, opened 2021-08-16
Dremio task now uses arrow.flight.auth2 logic

strojank:
Dremio task `DremioFetch` now `uses arrow.flight.auth2` logic instead of `legacy.arrow.flight.auth`

## Summary

Dremio is deprecating legacy Arrow Flight authentication. This PR updates the logic to be in accordance with default Dremio settings.

## Changes

`DremioFetch` task now uses different middleware from `pyarrow`. Legacy mode is not supported. This introduces breaking change for existing users of `DremioFetch`. 

## Importance

It updates Dremio tasks to work with latest Dremio versions.

## Checklist

This PR:


src/prefect/tasks/dremio/dremio.py
```diff
@@ -84,14 +127,80 @@ def run(self, user: str, password: str, host: str, port: int, query: str) -> Dic
...
-        client = flight.FlightClient(f"grpc+tcp://{host}:{port}")
-        client.authenticate(HttpDremioClientAuthHandler(user, password))
-
-        info = client.get_flight_info(
-            flight.FlightDescriptor.for_command(query + "--arrow flight")
+        # Authenticate with the server endpoint.
+        bearer_token = client.authenticate_basic_token(
+            username, password, initial_options
+        )
+        print("[INFO] Authentication was successful. Token is valid for 30 hours.")
```
madkinsz, line 173: Can you use `self.logger` instead so these will show up in the UI?

src/prefect/tasks/dremio/dremio.py
```diff
@@ -84,14 +127,80 @@ def run(self, user: str, password: str, host: str, port: int, query: str) -> Dic
...
+
+        flight_desc = flight.FlightDescriptor.for_command(query)
+        print("[INFO] Query: ", query)
+
+        # In addition to the bearer token, a query context can also
+        # be provided as an entry of FlightCallOptions.
+        # options = flight.FlightCallOptions(headers=[
+        #     bearer_token,
+        #     (b'schema', b'test.schema')
+        # ])
```
madkinsz, line 183: What's your plan for this?

src/prefect/tasks/dremio/dremio.py
madkinsz: A bunch of this seems DEBUG level

tests/tasks/dremio/test_dremio.py
```diff
@@ -1,20 +1,47 @@
...
 from pyarrow.flight import FlightUnavailableError
 
 from prefect.tasks.dremio.dremio import DremioFetch
 
 
 class TestDremioFetch:
     def test_construction(self) -> None:
-        task = DremioFetch(user="test", password="test", host="test")
-        assert (task.user == "test") and (task.host == "test")
+        task = DremioFetch(username="test", password="test", hostname="test")
```
madkinsz, line 10: I prefer to pass different values for each of these so your test ensures they are set correctly

madkinsz (commented):
Thanks for tackling this. A few comments.

src/prefect/tasks/dremio/dremio.py
strojank: I would leave this out for now.

src/prefect/tasks/dremio/dremio.py
strojank: Thanks for your feedback @madkinsz. I used `self.logger` where appropriate and removed redundant print statements.

tests/tasks/dremio/test_dremio.py
strojank: Good call. Done.

madkinsz (approved):

madkinsz:
Looking forward to getting this released!

PR #4875, opened 2021-08-16
Add a _on_scheduler_message callback to the DaskExecutor. Closes #4874

### LINKED ISSUE #4874: Allow subclasses of DaskExecutor to receive worker events
## Current behavior

Currently the DaskScheduler has a [`_watch_dask_events()`](https://github.com/PrefectHQ/prefect/blob/0dfbe5b3d176a6f72ee998c349747b34328618af/src/prefect/executors/dask.py#L241) method. It just logs a message when a worker is added or removed.

## Proposed behavior

It would be handy for subclasses of the `DaskScheduler` to be able to handle these worker added/removed messages itself, without needing to re-implement the worker status subscriptions (there is a fair bit of code around closing the communication channel and handling an apparent Dask bug).

## Example

Imagine you where running a Dask cluster on Kubernetes and you wanted to implement something that would tail worker logs using the k8s Python client. You could override the `on_scheduler_message` callback to implement this quite easily.

```python
class MyDaskScheduler(DaskScheduler):
    async def on_scheduler_message(self, op, message):
        if op == "add":
            tail_kubernetes_logs(message['worker'])
```

orf:
## Summary
Described in #4874 - the tl;dr is that it would be handy to allow subclasses of the `DaskExecutor` to handle scheduler messages in their own way.

## Changes

This refactors the current `_watch_dask_events` loop to call `_on_scheduler_message`, where it can be overriden

## Importance

Watching scheduler events seems non-trivial, making it easily extensible is very handy to some use-cases. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -238,6 +238,13 @@ class DaskExecutor(Executor):
         finally:
             self.client = None
 
+    async def _on_scheduler_message(self, op, message):
+        if op == "add":
+            for worker in message.get("workers", ()):
+                self.logger.debug("Worker %s added", worker)
+        elif op == "remove":
+            self.logger.debug("Worker %s removed", message)
+
     async def _watch_dask_events(self) -> None:
         scheduler_comm = None
         comm = None
@@ -260,11 +267,7 @@ class DaskExecutor(Executor):
                 except OSError:
                     break
                 for op, msg in msgs:
-                    if op == "add":
-                        for worker in msg.get("workers", ()):
-                            self.logger.debug("Worker %s added", worker)
-                    elif op == "remove":
-                        self.logger.debug("Worker %s removed", msg)
+                    await self._on_scheduler_message(op, msg)
         except asyncio.CancelledError:
             pass
         except Exception:
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -238,7 +238,7 @@ class DaskExecutor(Executor):
         finally:
             self.client = None
 
-    async def _on_scheduler_message(self, op, message):
+    async def _on_scheduler_message(self, op: str, message: dict) -> None:
         if op == "add":
             for worker in message.get("workers", ()):
                 self.logger.debug("Worker %s added", worker)
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -238,7 +238,14 @@ class DaskExecutor(Executor):
         finally:
             self.client = None
 
-    async def _on_scheduler_message(self, op: str, message: dict) -> None:
+    async def on_worker_status_changed(self, op: str, message: dict) -> None:
+        """
+        This method is triggered when a worker is added or removed from the cluster.
+
+        Args:
+            - op (str): Either "add" or "remove"
+            - message (dict): Information about the event that the scheduler has sent
+        """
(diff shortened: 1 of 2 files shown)


src/prefect/executors/dask.py
```diff
@@ -238,6 +238,13 @@ def start(self) -> Iterator[None]:
         finally:
             self.client = None
 
+    async def _on_scheduler_message(self, op: str, message: dict) -> None:
```
madkinsz, line 241: If we intend this to be overridden by users, should we remove the private `_` marking? Can you add a docstring to `on_scheduler_message`?

madkinsz (commented):
Hi! This seems like a reasonable change. I'd like to hear more about your use-case and potential use-cases though.

orf:
> edit: I see the logging case in the linked issue now. Still interested in hearing some more if you have them.

Sure! Dask doesn’t seem to offer much in the way of logging facilities. If you run a flow with a local Dask cluster you see the standard output from the workers in your console, as you would expect.

However if you then launch the flow via a remote Dask Gateway you don’t get any output. This isn’t great. Our plan is to keep track of workers as they are added or removed, and simply use the Python kubernetes client to stream the worker (+ scheduler) stdout to the client where it can be displayed to the user.

src/prefect/executors/dask.py
orf: Sure! Will do 👌

fjetter:
FWIW we (i.e. `dask.distributed`) are currently investigating the possibility for users to subscribe to all of our internal events, see https://github.com/dask/distributed/pull/5217 that's not logging but should cover almost all user cases. We're still in a scoping phase but this might interest you.

I just want to note that overloading the `on_scheduler_message` will not do much good since the events which are coming over this comm are very limited and defined in the below plugin

https://github.com/dask/distributed/blob/db0b2352bed85306459d17643ba5bb79f7109754/distributed/scheduler.py#L8066-L8079

src/prefect/executors/dask.py
orf: Done! I changed it to be `on_worker_status_changed` to be more explicit than just `on_scheduler_message`

orf:
> FWIW we (i.e. `dask.distributed`) are currently investigating the possibility for users to subscribe to all of our internal events, see [dask/distributed#5217](https://github.com/dask/distributed/pull/5217) that's not logging but should cover most use cases since for most stdlib log messages we also log an internal event (not all but most). We're still in a scoping phase but this might interest you.

I think there are a few different use cases here, and the changes proposed there would be quite interesting for some. But I like my logging to be fairly simple and reliable, so I'd always lean towards reading the actual process stdout via k8s (or some other method) than a more magical event-based approach. This obviously doesn't work if you've got workers that are shared between users however.

> I just want to note that overloading the `on_scheduler_message` will not do much good since the events which are coming over this comm are very limited and defined in the below plugin

If you can correlate the worker name to an entity which you can then poll for logs (i.e k8s) I think it's very useful. I've renamed the method to be `on_worker_status_changed` to be a bit more explicit and future proof!

src/prefect/executors/dask.py
madkinsz: I think that's a reasonable rename considering fjetter's comment

madkinsz (approved):

PR #4899, opened 2021-08-22
Make k8s flow run container args consistent 

g-clef:
## Summary
When the K8s agent generates the job spec for a flow run, it does so differently depending on whether the spec is generated inside `generate_job_spec_from_environment` or `generate_job_spec_from_run_config`. 

## Changes
This PR changes agent/kubernetes/agent.py -> generate_job_spec_from_run_config to not split the args string in the job spec.

## Importance
At least in my environment, k8s flow jobs created by `generate_job_spec_from_run_config` are not executing properly. When looking at the pod logs, the pod logs contain the prefect usage help text. 

Before this change, the k8s job spec generated by the `generate_job_spec_from_run_config` method would have args that looked like: 
```
spec:
  containers:
  - args:
    - prefect
    - execute
    - flow-run
    command:
    - /bin/sh
    - -c
```
This caused jobs to not run, as the container only ran "/bin/sh -c prefect", and just returned the prefect usage text. If I modify the job spec to be:
```
      containers:
      - args:
        - prefect execute flow-run
        command:
        - /bin/sh
        - -c

```
Then the job runs. So, I think the `get_flow_run_command(flow_run).split()` is the wrong way to do it.

## Checklist

This PR:

### DIFF
--- a/changes/pr4899.yaml
+++ b/changes/pr4899.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix k8s job spec generation when using run config - [#4899](https://github.com/PrefectHQ/prefect/pull/4899)"
+
+contributor:
+  - "[Aaron Gee-Clough](https://github.com/g-clef)"
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -689,7 +689,7 @@ class KubernetesAgent(Agent):
             container["imagePullPolicy"] = run_config.image_pull_policy
 
         # Set flow run command
-        container["args"] = get_flow_run_command(flow_run).split()
+        container["args"] = get_flow_run_command(flow_run)
 
         # Populate environment variables from the following sources,
         # with precedence:


madkinsz:
Hm, this was intentionally changed to a list in https://github.com/PrefectHQ/prefect/pull/3805 and has been in since 0.14.0. I know people have successfully been using Kubernetes runs. I'm curious why this isn't working for you & what best practice is here.

g-clef:
My problem may be because I'm using k8s + a custom job template. The template I'm using is here: https://github.com/g-clef/malwarETL-prefect/blob/fixes/template_container/prefect-job-template.yaml

What I'm struck by is the method of generating the args is different between the two methods. I feel like either both methods should be splitting the args, or neither should.

madkinsz:
We won't change the behavior for environments because they are deprecated and exist purely for backwards compatibility. They will be removed in the next major release. It looks like we decided to split the args in 3805 because it works with more entrypoints. It seems like it may be the ideal implementation going forward. I'm curious why it breaks with your template, we can definitely update documentation around best practices once we understand what's going on there. cc @TylerWanner

g-clef:
'k. I'm not sure why it's only working in my environment with the non-split version of `args`. If it helps, here's how the agent is run in my context: https://github.com/g-clef/malwarETL-prefect/blob/fixes/kubernetes/prefect-agent.yaml

g-clef:
On reflection, actually, I think I do know why it's happening:

When args is a list, the `command: ['/bin/sh/', '-c']` and `args: ['prefect', 'execute', 'flow-run']` translates to `/bin/sh -c prefect execute flow-run`. However, `/bin/sh -c` is looking for a single expression to run, and in that formulation there are 3 arguments to /bin/sh, only one of which is grabbed by `-c`. That means the /bin/sh command line will treat `execute` and `flow-run` as arguments to pass to `/bin/sh` (which it discards), rather than passing them in to the `-c` option for execution.

That's why the default result was the prefect help text, and why it works with one a single value in args: the single-value entry (with spaces) gets quoted, and therefore appears to be a single expression to `-c`, and runs `prefect execute flow-run`, rather than just `prefect`.

So I guess this boils down to: given that the default template doesn't include `command`, if the array version persists, what `command` should I use if I just want it to run the default `prefect execute flow-run`?

madkinsz:
@g-clef thanks for clarifying, I agree that looks like what's happening. If you don't set `command` at all, does it work? Defining a `command` overrides your container's `entrypoint` which with the Prefect image should be compatible with our generated args.

g-clef:
That seems to have worked (I mean, the flow crashed, but that's because I made a mistake, not because the job failed to launch). I take it the recommended way, therefore, is to not include `args` or `command` in a custom job template?

(thanks for taking the time to chase this down with me, by the way.)

g-clef:
(Aside: thanks for diving into this with me.) If I don't set `command` at all I get `prefect: command not found`

madkinsz:
Yeah I can find a way to make that more clear in the documentation. I can note this here for others as well.

`command` overrides container `ENTRYPOINT` values which we generally would like to avoid. If we do not override this, users can define entrypoints that wrap their flow runs. For example, in the Prefect base image, we use `tini` to ensure zombies are cleaned up have a small script that install runtime pip packages for users. I think that passing the `args` as a list supports more typical entrypoints but notably will not work with `/bin/sh -c` which expects a single string argument.

`args` will always be overridden by prefect to insert the flow run command.

PR #4901, opened 2021-08-23
Support for RRule (iCal style recurrence rule) clocks / schedules.

nitay:
## Summary
Adds support for RRule (iCal style recurrence rule) clocks / schedules.

## Changes
Adds `RRuleClock` and `RRuleSchedule`

## Importance
RRules are much more powerful than the existing clock impls. This makes it easy for Prefect to work with organizations that are already leveraging iCal style event logic.

## Checklist

This PR:

### DIFF
--- a/src/prefect/schedules/clocks.py
+++ b/src/prefect/schedules/clocks.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from dateutil import rrule
 from typing import Any, Iterable, List, Set, Union
 
 import pendulum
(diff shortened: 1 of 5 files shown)


src/prefect/serialization/schedule.py
```diff
@@ -89,6 +90,65 @@ class Meta:
     labels = fields.List(fields.Str(), allow_none=True)
 
 
+class RRuleSchema(ObjectSchema):
+    class Meta:
+        object_class = rrule.rrule
+
+    _freq = fields.Integer(required=True)
+    _dtstart = DateTimeTZ(allow_none=True)
```
nitay, line 98: there's more fields I need to add am not done here yet

cicdw:
Hey @nitay -- overall the direction looks good; I'm impressed you've been able to grok the full schedule code paths!  (I think this may be the first community contributed clock, too).  I have a minor concern that the serialization attributes are a little complex, and because you're relying on private attributes they could be brittle across dateutil versions so I'd like to confirm that these are the most efficient way to represent rrules. 

I definitely want to play around with the code a little bit to understand rrules better but overall this looks supportable. 

I can give a more thorough code review whenever you feel it's ready for that and in the meantime I'll see if we can simplify the serializers at all.

nitay:
thx @cicdw - I agree with you the private fields felt hacky. happy to do sth cleaner, do you have any recommendations? 
In terms of thorough review all that's left is just adding more fields and some more tests so the overall structure is not gonna change assuming it's acceptable. So I think what makes sense here is you lmk if this current mechanism is good then I'll add everything else. if there's another way that's preferred (to point above) then I'll switch to that and impl everything that way

cicdw:
@nitay I poked around and it seems like all the initialization arguments to an `rrule` are stored privately so your implementation does seem to be the most efficient.  Once you've added all the fields to the serializers I can do a thorough review to see if we can streamline any but otherwise I feel comfortable moving forward with this 👍

cicdw:
Hey @nitay -- apologies for this new piece of info, but hopefully you agree that it's great news: it seems there is a much easier way to serialize `rrule`s!  

If you use `str(rrule)` a standardized format is output that can be reparsed back into an `rrule` object using `rrule.rrulestr`.  See here for some examples: https://dateutil.readthedocs.io/en/stable/rrule.html#rrulestr-examples

This format is also supposed in javascript and therefore can be used to provide a UI interface to these schedule objects which would be a great followup project.

nitay:
@cicdw interesting, good find - I had seen the str formatting before but didn't see its usage for parsing thought it was just debug only and not standard. let me try that approach

src/prefect/schedules/clocks.py
```diff
@@ -379,3 +383,89 @@ def events(self, after: datetime = None) -> Iterable[ClockEvent]:
...
+            include_after = True
+
+        # RRule throws if trying to compare timezone-aware datetimes to timezone-naive ones.
+        # To protect from this happening we grab the first element in the RRule to see what
+        # timezone awareness it has, and use that to alter our `after` variable accordingly.
+        first_item = next(iter(self.rrule_obj), after)
+        if after.tzinfo is not None and first_item.tzinfo is None:
+            after = after.naive()
+        if after.tzinfo is None and first_item.tzinfo is not None:
+            after = after.in_tz(first_item.tz)
```
nitay, line 458: For some reason on ser/deser RRules lose the timezone. I'm not sure why and my debugging didn't see why or how to fix it. If you have a way lmk
I actually made the comparing timezone aware to naive issue come up a bunch in testing even without serialization - e.g if you just use regular datetime instead of pendulum it's easy to accidentally cause it.
Given both of those things, seems like it's good to be defensive here and work no matter what gets passed in
Given both of these things, seemed like it'd

src/prefect/schedules/clocks.py
```diff
@@ -379,3 +383,89 @@ def events(self, after: datetime = None) -> Iterable[ClockEvent]:
...
+            include_after = True
+
+        # RRule throws if trying to compare timezone-aware datetimes to timezone-naive ones.
+        # To protect from this happening we grab the first element in the RRule to see what
+        # timezone awareness it has, and use that to alter our `after` variable accordingly.
+        first_item: datetime = next(iter(self.rrule_obj), after)
+        if after.tzinfo is not None and first_item.tzinfo is None:
+            after = after.replace(tzinfo=None)
+        if after.tzinfo is None and first_item.tzinfo is not None:
+            after = after.replace(tzinfo=first_item.tzinfo)
```
nitay, line 458: For some reason on ser/deser RRules lose the timezone. I'm not sure why and my debugging didn't see why or how to fix it. If you have a way lmk
I actually made the comparing timezone aware to naive issue come up a bunch in testing even without serialization - e.g if you just use regular datetime instead of pendulum it's easy to accidentally cause it.
Given both of those things, seems like it's good to be defensive here and work no matter what gets passed in

cicdw: @nitay you could grab the timezone off the initializing start date / rrule and (de)serialize it as a separate attribute

nitay: @cicdw I don't think that would help. I presume you mean to then pass `dtstart` to `rrulestr()`. I actually tried hardcoding a UTC `dtstart` initially -  rrulestr doesn't actually grab the timezone from it

nitay: some more details - afaict `rrulestr()`'s only way to get a timezone is through parsing it from the str representation. in particular it looks for a `TZID=` parameter. however, `str(rrule)` doesn't seem to actually write the `TZID=` anywhere in its code. something doesn't add up either I'm missing something silly in my understanding, or there may be some bug in `rrule` 🤷‍♂️ ?

cicdw: Ah not quite -- I mean add an additional attribute on the schedule called "timezone" that you serialize as a string independently of the rrule representation; you could pull this off the `dtstart` of the rrule obj.  Then to deserialize it back into an rrule, you first convert the string to an rrule then manually set the timezone back on the `dtstart` attribute.

PR #4902, opened 2021-08-24
[RFC] Test+fix potential log upload race condition

dbast:
Two potentials solutions:
* Extend https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/logging.py#L111 to do an exact measurement of the serialised (json.dumps) log dict or add some additional 200 bytes of assumed overhead (the while condition also needs to check that those 200 bytes are free) or
* Extend https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/logging.py#L167 to truncate until the whole serialised log dict size <= MAX_LOG_LENGTH (not only the message).

Implemented solution 1 with json.dumps.

## Summary

Avoid log upload race condition by better batch size measurement and parameter setup.

## Changes

* Lowers the MAX_BATCH_LOG_LENGTH to leave some room for the graphql serialisation overhead.
* Do more accurate pending_length measurement for the batched log entries.
* Require some extra free 200 bytes (the overhead of the log dict) before adding more log messages to the batch to be uploaded. 

## Importance

This is important to avoid log upload deadlocks, where the client can not recover from via retries.

## Checklist

This PR:

### DIFF
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -36,9 +36,9 @@ PREFECT_LOG_RECORD_ATTRIBUTES = (
 
 MAX_LOG_LENGTH = 1_000_000  # 1 MB - max length of a single log message
 MAX_BATCH_LOG_LENGTH = 3_999_000  # max total batch size for log messages
-                                  # 4 MB = 5 MB (backend limit)
-                                  #      - 1 MB (MAX_LOG_LENGTH)
-                                  #      - 1 kB (serialisation overhead)
+# 4 MB = 5 MB (backend limit)
+#      - 1 MB (MAX_LOG_LENGTH)
+#      - 1 kB (serialisation overhead)
 
 
 class LogManager:
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -10,6 +10,7 @@ When running locally, log levels and message formatting are set via your Prefect
 """
 import atexit
 import logging
+import json
 import sys
 import threading
 import time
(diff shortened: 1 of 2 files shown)


src/prefect/utilities/logging.py
```diff
@@ -106,9 +110,11 @@ def _write_logs(self) -> None:
         cont = True
         while cont:
             try:
-                while self.pending_length < MAX_BATCH_LOG_LENGTH:
+                # ensuring extra 200 bytes of free batch size for the
+                # log dict properties beside the message with MAX_LOG_LENGTH.
+                while self.pending_length + 200 < MAX_BATCH_LOG_LENGTH:
                     log = self.queue.get_nowait()
-                    self.pending_length += len(log.get("message", ""))
+                    self.pending_length += len(json.dumps(log))
```
dbast, line 117: json.dumps is maybe too expensive here... could be replaced by e.g.
```suggestion
                    self.pending_length += len(log.get("message", "")) + 200
```

dbast:
Happy to hear about ideas to simplify this.

dbast:
Closing this for now...

madkinsz:
Hey @dbast -- I have been keeping an eye on this and intend to weigh in when I have a bit more time. Have our recent fixes helped or are you just closing this from lack of response?

dbast:
The batch sizes have been increased and aligned in front- and backend:
* https://github.com/PrefectHQ/server/pull/273
* https://github.com/PrefectHQ/prefect/pull/4900

I still believe that this topic is not yet fully resolved as there potentially is a race condition in the batch size measuring under high log pressure, which this PR is addressing.

@madkinsz Would be nice to get some feedback/review for the PR.

madkinsz:
I'll definitely take a look this week. Thanks!

madkinsz:
I'm continuing to think about this after reading it through a few times. I think we'll need to have some internal discussions about the best way to handle the cases you've presented.

dbast:
We running is in production since some month already... and it fixes the issue for us without any other side effects so far.

madkinsz:
Thanks for the additional information! I'm also working on the log handler for Orion right now and have ensured that it does not exceed the max size. I'm not sure what the best way to backport that implementation or use this implementation is yet, but it's on my mind.

madkinsz:
Hey @dbast — I'd appreciate your review of https://github.com/PrefectHQ/prefect/pull/5316

Would you like to be listed as a contributor?

madkinsz:
Closing in favor of #5316

PR #4903, opened 2021-08-24
Fix marshmallow metadata kwarg deprecation warning

carlos-alberto:
https://github.com/PrefectHQ/prefect/issues/4540

`serialization` was erroneously using the kwarg `key` to
marshmallow.fields.Dict, this should be `keys`. Marshmallow was seeing
this as `additional_metadata`, so showing a deprecation warning.

## Summary

Fixes uses of `marshmallow.fields.Dict` to use `keys` correctly as a kwarg
rather than `key`. 

(See https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Dict).

## Changes

Fixes uses of `marshmallow.fields.Dict` to use `keys` correctly as a kwarg
rather than `key`.

## Importance

Fixes a warning raised by prefect. Our test flow has errors raised on warnings, so we had to
explicitly ignore this error throughout our test runs.

## Checklist

This PR:

### DIFF
--- a/changes/pr4903.yaml
+++ b/changes/pr4903.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fixes uses of marshmallow.fields.Dict to use keys correctly as a kwarg rather than key. - [#4903](https://github.com/PrefectHQ/prefect/pull/4903)"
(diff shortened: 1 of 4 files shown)


madkinsz (approved):
Thanks!

Issue #4885, opened 2021-08-19
Passing a SecretTask to MySQLTask for the password fails to connect


shearer12345:
## Description

Passing a SecretTask to MySQLTask for the password fails to connect.

This is because the MySQL Task allows passing a password through both `init` and `run`, but the password passed through `run` is never used (neither assign to `self.password`, nor used directly in the connection).

In contrast, Postgress Task only allows passing a password through `run`, and uses that in the connection.

I'm happy to make a PR for this.

```
# https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/mysql/mysql.py#L105
conn = pymysql.connect(
    host=self.host,
    user=self.user,
    password=self.password,
    db=self.db_name,
    charset=self.charset,
    port=self.port,
    ssl=ssl,
)
```

```
# https://github.com/PrefectHQ/prefect/blob/master/src/prefect/tasks/postgres/postgres.py#L175
conn = pg.connect(
    dbname=self.db_name,
    user=self.user,
    password=password,
    host=self.host,
    port=self.port,
)
```

## Expected Behavior

Passing a SecretTask to MySQLTask for the password connects

## Reproduction
```
from prefect import Flow
from prefect.tasks.mysql import MySQLFetch
from prefect.tasks.secrets import PrefectSecret

database_name = 'database'
sql_fetch = MySQLFetch(db_name=database_name, user='user',
                       host='db_server,com', fetch='all'
                       )

with Flow("mysql_connection_failure_flow") as mysql_connection_failure_flow:
    password = PrefectSecret("password")
    query_result = sql_fetch(password=password, query='SELECT * FROM table')
```

## Environment
```
{
  "config_overrides": {
    "context": {
      "secrets": false
    }
  },
  "env_vars": [],
  "system_information": {
    "platform": "Windows-10-10.0.19042-SP0",
    "prefect_backend": "server",
    "prefect_version": "0.15.4",
    "python_version": "3.8.10"
  }
}

**Comment by shearer12345** [comment]
Work around by getting the secret outside of a task - using the [secrets API directly](https://docs.prefect.io/core/concepts/secrets.html#local-context).
```
from prefect import Flow
from prefect.client import Secret
from prefect.tasks.mysql import MySQLFetch
from prefect.tasks.secrets import PrefectSecret

database_name = 'database'

s = Secret("password")
sql_fetch = MySQLFetch(db_name=database_name, user='user',
                       host='db_server,com', fetch='all',
[...]

PR #4907, opened 2021-08-24
Fix MySQL tasks that never use runtime arguments

tchoedak:
## Summary
This PR introduces a fix to a bug introduced in #4545 that is highlighted by issue #4885: Runtime arguments were ignored over task init arguments.

A task such as
```python
password_from_secret = "MySecretPassword"

task = MySQLFetch()
task.run(user="admin", host="localhost", port="3302", password=password_from_secret)
```

would fail because `pmysql.connect` would be called with `self.password` which would be `None` in this scenario.

A task such as
```python
password_from_secret = "MySecretPassword"

task = MySQLExecute(user="admin", host="localhost", port="3302", password="password")
task.run(password=password_from_secret)
```
would also fail if only `password_from_secret` contains the correct password because `pymysql.connect` would be called with `self.password` which would be `"password"` in this scenario.

## Changes
Make sure runtime args are used first over init args. Init args are still used thanks to @default_from_attrs decorator if runtime args aren't available.

## Importance
Ensures MySQL tasks work properly with runtime arguments.

## Checklist

This PR:

### DIFF
--- a/changes/pr4907.yaml
+++ b/changes/pr4907.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Use runtime arguments over init arguments instead of ignoring them for MySQL Tasks - [#4907](https://github.com/PrefectHQ/prefect/pull/4907)"
+
+contributor:
+  - "[Tenzin Choedak](https://github.com/tchoedak)"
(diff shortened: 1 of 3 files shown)


madkinsz (dismissed):
LGTM; want to add a change entry?

madkinsz (approved):

PR #4914, opened 2021-08-27
Patch client.py to properly retrieve the api_server from the prefect context.

xuevin:
## Summary
Patch `client.py` to properly retrieve the `api_server` from the prefect context. 

## Changes
Add a missing `self` for api_server

## Importance

This bug caused downstream issues when spawning dask workers on a kubernetes cluster. The api_server was reset to the cloud api instead of the one provided in the context. 

## Checklist

This PR:

### DIFF
--- a/changes/pr4914.yaml
+++ b/changes/pr4914.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix client instantiation to properly use the api server defined in the prefect context."
+
+contributor:
+  - "[Vincent Xue](https://github.com/xuevin)"
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -153,7 +153,7 @@ class Client:
 
         self._api_token = api_token or prefect.context.config.cloud.get("auth_token")
 
-        if not self.api_key and not api_server:
+        if not self.api_key and not self.api_server:
             # The default value for the `api_server` changed for API keys but we want
             # to load API tokens from the correct backwards-compatible location on disk
             self.api_server = prefect.config.cloud.graphql
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -153,7 +153,7 @@ class Client:
 
         self._api_token = api_token or prefect.context.config.cloud.get("auth_token")
 
-        if not self.api_key and not self.api_server and self._api_token:
+        if not self.api_key and not api_server and self._api_token:
             # The default value for the `api_server` changed for API keys but we want
             # to load API tokens from the correct backwards-compatible location on disk
             self.api_server = prefect.config.cloud.graphql
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -153,7 +153,7 @@ class Client:
 
         self._api_token = api_token or prefect.context.config.cloud.get("auth_token")
 
-        if not self.api_key and not self.api_server:
+        if not self.api_key and not self.api_server and self._api_token:
             # The default value for the `api_server` changed for API keys but we want
             # to load API tokens from the correct backwards-compatible location on disk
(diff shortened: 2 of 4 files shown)


xuevin:
@madkinsz I updated the tests to point to `cloud.api`.

madkinsz:
Shouldn't it be loading `cloud.graphql` if `cloud.api` is not defined? Just want to make sure we're not breaking old behavior by updating these tests.

xuevin:
This is also a change which I am not very familiar with but I assumed that the the `client.py` handled the fallback api via the following snippet.  

https://github.com/PrefectHQ/prefect/blob/d61fa6aac9330c5817cc8e8b8f8cca2d634ea7e1/src/prefect/client/client.py#L155-L158

madkinsz:
The backwards compat refactor is tricky to reason about. I'll make sure it's working as intended today then merge. Thanks again for your work on this :)

madkinsz:
This will indeed not be backwards compatible. The `if` statement there was written correctly and as intended. We want to change the _default_ value for `self.api_server` if using an API token. Otherwise, we may fail to load the token from the correct place on disk as the location is generated from `self.api_server`. We may be able to add `and not self._api_token` to the statement you modified which would only resort to the old `api_server` default if the token is not already available (as it would be on a dask worker). 

Are you using API tokens or API keys for auth when encountering this dask worker issue?

xuevin:
Hi @madkinsz . I changed the patch to check for `prefect.context.config.backend==cloud`. This seems to resolve the issues, but I do not feel that this not a very elegant solution. Suggestions for change are welcome.

Just as a summary: 

Dask-workers currently are unable to connect to a prefect-server because the `client.api_server` attribute is initialized incorrectly.  When the dask-worker initializes the client, `api_server=None` and there exists no api_keys  anywhere in the  flow (client init,  context, or cached_auth) .  This causes the `self.api_server` to default to `prefect.config.cloud.graphql` (See image below) This is an issue when a dask-worker seeks credentials as it is a clean slate and only has what is provided in the context.

image

The proposed fix is to set the `client.api_server` to `prefect.config.cloud.graphql` only in the case when `prefect.context.config.backend == "cloud"`. 

With this patch, the dask-worker will receive context information that the backend is `server` (with no tokens and keys) and use the `prefect.context.config.cloud.api` as desired. 



=====

Also I am not sure why `tokens` or `keys` would play a role in this issue, given that the dask-worker  (to my knowledge) does not get passed when using `backend=server`

madkinsz:
I think this is actually a reasonable solution. We don't need to fallback to the deprecated `api_server` default when using server.

madkinsz (approved):
Thanks for digging into this!

MrJack91:
@madkinsz: We use the prefect server, set up with reverse proxy to use the same ssl-cert for graphql and the gui.
* https://url.com => gui
* https://url.com/graphql => graphql

Before this change we could set this env's and the graphql were set correctly. 
PREFECT__SERVER__HOST=https://url.com
PREFECT__SERVER__PORT=443

The suffix was added. because this lines were executed:
https://github.com/PrefectHQ/prefect/blob/d61fa6aac9330c5817cc8e8b8f8cca2d634ea7e1/src/prefect/client/client.py#L155-L158

Now we miss this part. It works if i add the suffix in the port
PREFECT__SERVER__PORT=443/graphql

Or how should i setup my environment....?

madkinsz:
Hey @MrJack91 -- sorry I missed this.

This looks like an unintended breaking change. I've opened an issue at https://github.com/PrefectHQ/prefect/issues/4969. Unfortunately I'm not sure what the best way to fix this in the long run is.

MrJack91:
@madkinsz Thanks for your response!

PR #4917, opened 2021-08-30
Pass existing environment variables to agentless flow runs

madkinsz:
## Summary

Includes the local environment in agentless flow runs by default. This was previously not available as I did not want to expose the entire local environment to flow runs--it seemed like a security risk. 

Alternatively, we could allow opt-in for a list of environment variables or encourage users to pass thing via prefect local secrets.

## Importance

Users frequently store environment variables that their flows needs on the machine their flow runs on

## Checklist

This PR:

### DIFF
--- a/tests/backend/test_execution.py
+++ b/tests/backend/test_execution.py
@@ -1,5 +1,6 @@
 import pendulum
 import pytest
+import os
 import sys
 from unittest.mock import MagicMock, call
 from subprocess import CalledProcessError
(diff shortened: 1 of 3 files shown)


madkinsz:
@TylerWanner @anticorrelator review me please

TylerWanner:
LGTM!  Feels like a usability lift, though I wouldn't suggest anyone to inject env vars from execution environment into a flow run in production, if avoidable.  It is good practice to make the execution environment as agnostic as possible, and it is a bad idea to throw some undefined sack of env vars around like this in production, but this can be very useful in development.

TylerWanner:
Sorry for waffling here.  Given that agentless execution is a subprocess and not directly equivalent to running via agent, it does make sense to include the host env vars by default.  In fact, this expectation should be clear enough that I do not think it needs a toggle.  Anybody using --execute is effectively taking ownership of their execution environment (away from Prefect), and should expect it to work this way.

tests/backend/test_execution.py
```diff
@@ -58,34 +58,42 @@ class Mocks:
...
-        execute_flow_run_in_subprocess("flow-run-id")
+        execute_flow_run_in_subprocess(
+            "flow-run-id", include_local_env=include_local_env
+        )
 
         # Should pass the correct flow run id to wait for
         mocks.wait_for_flow_run_start_time.assert_called_once_with("flow-run-id")
 
+        # Merge the starting env and the env generated for a flow run
+        base_env = os.environ.copy() if include_local_env else {}
```
anticorrelator, line 76: I've always had an idiosyncratic preference for `dict()` instead of `{}` because of the visual similarity to set syntax

anticorrelator (approved):
lgtm

tests/backend/test_execution.py
madkinsz: That makes sense, I may do that more in the future.

PR #4924, opened 2021-08-31
Catch BaseException in multiprocessing_safe_run_and_retrieve

zangell44:
## Checklist

This PR:


madkinsz:
Makes sense to me if we want the subprocess signals to propagate to the main process.

madkinsz:
@anticorrelator can you take this over? We need to decide if we want to catch _all_ `BaseException` types or just `PrefectSignal`. I think that it might make sense to pass all base exceptions up but `PrefectSignal` is a safer option 🤷

anticorrelator:
After thinking this through I think we should only be catching `PrefectSignal` to resolve this issue. If we need later catch `BaseException` we can handle them separately

madkinsz (commented):
I agree with that assessment, can you add a changelog entry?

madkinsz (approved):

PR #4927, opened 2021-08-31
Add output filtering example to Mapping docs

sgasse:
Prefect already provides a built-in functionality to filter the output
of mapped tasks that should be used in downstream tasks. This commit
adds a paragraph with an example.

## Summary
Update Mapping documentation with an example of filtering mapped tasks output.

## Changes
Update `docs/core/concepts/mapping.md

## Importance
This could save someone else some time when building flows with mapped tasks.

## Checklist
This PR:

- [x] adds new tests (not required since only the documentation is amended)
- [x] adds a change file in the `changes/` directory (not required)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (not required)


madkinsz (approved):
Sweet! Thanks for contributing.

PR #4943, opened 2021-09-04
Adding Census Task to Library

dflynn20:
@crw I based a lot of the structure off your Fivetran task library, I hope you don't mind! This is for triggering reverse ETL jobs with getcensus.com through the api trigger as a Prefect Task.

## Summary

Adds a task for triggering a Census Sync via a Prefect task.

## Changes

This adds a Census task to the Prefect Task library.

## Importance

Adding Reverse ETL orchetration into Prefect with getcensus.com completes the data pipeline circle of life.

## Checklist

This PR:


tests/tasks/census/test_census.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+# from prefect import Flow
```
zangell44, line 3:
```suggestion
```

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+            status_response = requests.get(sr_url)
+            response_dict = status_response.json()
+            if status_response.status_code != 200 or "data" not in response_dict.keys():
+                raise ValueError(
+                    f"Getting status of sync failed, please visit Census Logs at {log_url} to see more."
+                )
+            result = response_dict["data"]
+            status = result["status"]
+            if status == "working":
+                self.logger.info(
```
zangell44, line 104:
```suggestion
                self.logger.debug(
```
seems more like a debug level log but I'm not very familiar with the integration - do users want to see this every run?

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+                    f"Sync {sync_id} still running after {round(time.time()-start_time, 2)} seconds."
+                )
+                continue
+            break
+
+        self.logger.info(
```
zangell44, line 110:
```suggestion
        self.logger.debug(
```
same reasoning as above

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+        secret, sync_id = confirmed_pattern.groups()
+        response = requests.post(api_trigger)
+
+        if response.status_code != 200:
+            error_string = f"Sent POST, failed with status code {response.status_code}: {response.text}."
+            raise ValueError(error_string)
+
+        sleep_time = max(MIN_WAIT_TIME, poll_status_every_n_seconds)
+
+        self.logger.info(
```
zangell44, line 82:
```suggestion
        self.logger.debug(
```

src/prefect/tasks/census/census.py
dflynn20: Makes sense to me!

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+            - dict: dictionary of statistics returned by Census on the specified sync
+        """
+
+        if not api_trigger:
+            raise ValueError(
+                """Value for parameter `api_trigger` must be provided. See Census sync
+                                configuration page."""
+            )
+
+        pattern = r"https:\/\/bearer:secret-token:(.*)@app.getcensus.com\/api\/v1\/syncs\/(\d*)\/trigger"
```
zangell44, line 61: it would be nice if the pattern checking was separated into it's own function

if `api_trigger` is provided at `__init__`, we should check the pattern is correct at that time instead of waiting until run time

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+        if not confirmed_pattern:
+            raise ValueError(
+                """Invalid parameter for `api_trigger` please paste directly from the Census
+                                sync configuration page. This is CaSe SenSITiVe."""
+            )
+
+        secret, sync_id = confirmed_pattern.groups()
+        response = requests.post(api_trigger)
+
+        if response.status_code != 200:
```
zangell44, line 76: `response.raise_for_status()` might do everything you need here

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+                self.logger.debug(
+                    f"Sync {sync_id} still running after {round(time.time()-start_time, 2)} seconds."
+                )
+                continue
+            break
+
+        self.logger.debug(
+            f"Sync {sync_id} has finished running after {round(time.time()-start_time, 2)} seconds."
+        )
+        self.logger.info(f"View details here: {log_url}.")
```
zangell44, line 113: should this be debug level too?

dflynn20: I think this should be info actually because this that link takes the user to the UI to be able to drill into the invalid or failed records on the syncs. 

If in general though: info-level logs are not encouraged, I'm happy to turn that into a debug level here and then return that log url in the response dictionary as well so it's up to the data eng.

src/prefect/tasks/census/census.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Returns a dictionary of:
```
zangell44, line 115: Love the detail here! Can we put this in the docstring instead? It's very hard to track down in the source code

dflynn20: For sure, let me check on formatting and I will add that in!

src/prefect/tasks/census/census.py
dflynn20: Ooh okay, let me look into that!

src/prefect/tasks/census/census.py
dflynn20: A miss here is worth a 
```if not confirmed_pattern:
            raise ValueError(
                """Invalid parameter for `api_trigger` please paste directly from the Census
                                sync configuration page. This is CaSe SenSITiVe."""
            )``` 
right?

File: tests/tasks/census/test_census.py [hunk 9]
```diff
@@ -0,0 +1,22 @@
...
+        assert task.api_trigger == "random_url.com"
+        with pytest.raises(ValueError, match="paste"):
+            task.run()
+
+    def test_improper_secret_failed(self):
+        task = CensusSyncTask()
+        with pytest.raises(ValueError, match="failed"):
+            task.run(
+                api_trigger="https://bearer:secret-token:BLAH@app.getcensus.com/api/v1/syncs/12/trigger"
+            )
```
zangell44, line 22: (not essential) - it would be nice to have a test where the task runs successfully and we mock out the necessary API calls

src/prefect/tasks/census/census.py
zangell44: got it, `info` sound appropriate then, thanks for the detail here!

tests/tasks/census/test_census.py
dflynn20: I was thinking about that, but I didn't want to expose the API key for an actual account (even if it's a demo account) on the github repo for security reasons

src/prefect/tasks/census/census.py
zangell44: for sure! I'm envisioning something like the following (irrelevant code snippets omitted)

```python
class CensusSyncTask(Task):
    ...
    def __init__(self, api_trigger=None, **kwargs):
        if api_trigger:
            self.confirm_url_pattern(api_trigger)
        self.api_trigger = api_trigger
        super().__init__(**kwargs)

    def run(...):
        ...
        confirmed_pattern = self.get_confirmed_url_pattern(api_trigger)
        secret, sync_id = confirmed_pattern.groups()
        ...

    @staticmethod
    def get_confirmed_url_pattern(api_trigger: str):
        pattern = r"https:\/\/bearer:secret-token:(.*)@app.getcensus.com\/api\/v1\/syncs\/(\d*)\/trigger"
        url_pattern = re.compile(pattern)
        # Making sure it is a valid api trigger.
        confirmed_pattern = url_pattern.match(api_trigger)

        if not confirmed_pattern:
            raise ValueError(
                """Invalid parameter for `api_trigger` please paste directly from the Census
                                sync configuration page. This is CaSe SenSITiVe."""
            )
      return confirmed_pattern

```

zangell44 (changes requested):
Couple comments, looking good overall!

src/prefect/tasks/census/census.py
dflynn20: Thanks for putting me onto this method. It's awesome

src/prefect/tasks/census/census.py
dflynn20: I didn't see this until I had already pushed the almost identical code (but in the wrong place). Marking this as resolved now.

zangell44 (approved):

PR #4949, opened 2021-09-08
Remove $ from code examples for copy and paste

jboxman:
## Summary

I'm not sure which style applies, but the `$` in some of the CLI commands means that you cannot paste as-is. This PR removes the `$`. The other approach is to modify the copy and paste script to strip any leading `^$\s?`.

## Checklist

This PR:


madkinsz (approved):
Hey @jboxman -- this is a good point, those `$` should just be styled and not a part of the copyable content. We'll likely want to revisit this to be consistent across all our docs.

PR #4959, opened 2021-09-11
Remove duplicate descriptions in API docs for triggers

ryoasu:
## Summary
Duplicate description of `all_finished` in [API docs for Triggers](https://docs.prefect.io/api/latest/triggers.html#functions)

## Changes
Removed duplicate description of `all_finished` in the API documentation of triggers.
The reason seems to be that `always_run`, an alias of `all_finished` is included in `pages.triggers.functions` in outline.toml.
Therefore, we removed `always_run` from `pages.triggers.functions`.


madkinsz (approved):
Thanks!

PR #4968, opened 2021-09-15
Fix PostgresFetch with headers for one row

noppaz:
## Summary
This fixes a bug in PostgresFetch where TypeError is raised when fetch = "one" (default) and col_names = True.

## Changes

As the psycopg cursor.fetchone() method returns a Tuple[Any] and not a List[Tuple[Any]] as cursor.fetchall and cursor.fetchmany the concatenation of tuple and list fails. This puts the single tuple row into a list when col_names is true.

## Importance

Fixes a bug, "one" is the default fetch argument but I don't think it's used that often and less so in conjunction with headers and therefore not discovered yet.

## Checklist

This PR:

### DIFF
--- a/changes/pr4968.yaml
+++ b/changes/pr4968.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix PostgresFetch with headers for one row - [#4968](https://github.com/PrefectHQ/prefect/pull/4968)"
+
+contributor:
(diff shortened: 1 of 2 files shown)


madkinsz (approved):

madkinsz:
Thanks!

PR #4971, opened 2021-09-15
Allow max_retries=0 override at task level

jakekaplan:
## Summary
Allow `max_retries` to be set at the task level to 0, overriding config setting

## Changes
This PR resolves https://github.com/PrefectHQ/prefect/issues/4880

The below code now executes correctly when `max_retries` and `retry_delay` is already specified at the config level.

```
@task(max_retries=0, retry_delay=None)
def my_task():
    ...
```

## Importance
This PR is important because it can be useful to have a single specific task not retry and it matches what the parameters intended.

## Checklist

This PR:

### DIFF
--- a/changes/pr4971.yaml
+++ b/changes/pr4971.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Allow max_retries to be set to 0 when overriding the config level parameter - [#4971](https://github.com/PrefectHQ/prefect/pull/4971)"
--- a/changes/pr4971.yaml
+++ b/changes/pr4971.yaml
@@ -1,2 +1,2 @@
-fix:
-  - "Allow max_retries to be set to 0 when overriding the config level parameter - [#4971](https://github.com/PrefectHQ/prefect/pull/4971)"
+enhancement:
+  - "Allow disabling retries for a task with `max_retries=0` when retries are globally configured - [#4971](https://github.com/PrefectHQ/prefect/pull/4971)"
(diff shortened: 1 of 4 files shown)


madkinsz:
Closes #4880 

Can you add a test ensuring that a task that runs with `max_retries=0` does not perform any of the retry logic? There are lots of places in the code where values are compared explicitly to `None` instead of relying on truthiness.

jakekaplan:
> Can you add a test ensuring that a task that runs with `max_retries=0` does not perform any of the retry logic? There are lots of places in the code where values are compared explicitly to `None` instead of relying on truthiness.

I can definitely do that. Would the correct place to add that test be in `test_task_runner.py`?

changes/pr4971.yaml
```diff
@@ -0,0 +1,2 @@
+fix:
+  - "Allow max_retries to be set to 0 when overriding the config level parameter - [#4971](https://github.com/PrefectHQ/prefect/pull/4971)"
```
madkinsz, line 2:
```suggestion
  - "Allow disabling retries for a task with `max_retries=0` when retries are globally configured - [#4971](https://github.com/PrefectHQ/prefect/pull/4971)"
```
I try to make these messages address user stories since they end up in the public changelog

changes/pr4971.yaml
```diff
@@ -0,0 +1,2 @@
+fix:
```
madkinsz, line 1:
```suggestion
enhancement:
```
This isn't really a bug and we tend to lean towards enhancements when improving usability

madkinsz (commented):
LGTM, just some comments on the changelog entry.

jakekaplan:
> LGTM, just some comments on the changelog entry.

Made the corrections. Thank you for the feedback! I'll try and keep those in mind

madkinsz (approved):

Issue #4824, opened 2021-07-30
Returning tuple does not work with mapped tasks


pauleikis:
## Description

Hello,

I have a problem, that when running a mapped task which returns a tuple, it does not work as expected (or does not work at all). It works fine if the task is not mapped and either "nout" task annotation parameter is used, or type annotation Tuple is used.

## Expected Behavior

I would expect the behaviour for mapped and non-mapped tasks is the same.

## Reproduction

I could not make it run using "nout":

```python
from prefect import Flow, task


@task
def extract():
    return [1, 2, 3, 4, 5]


@task(nout=2)
def transform(num: int):
    return num, num + 10


@task
def load(x: int, y: int):
    print(x, y)


with Flow("flow") as flow:
    data = extract()
    x, y = transform.map(data)
    load.map(x=x, y=y)


if __name__ == '__main__':
    flow.run()
```

```
TypeError: Task is not iterable. If your task returns multiple results, pass `nout` to the task decorator/constructor, or provide a `Tuple` return-type annotation to your task.
```

If instead I was passing "nout" to `.map` call, I made it run, but it produced incorrect results, I expected `load` task to be run 5 times, instead it was run only twice with incorrect parameters

```python
from prefect import Flow, task


@task
def extract():
    return [1, 2, 3, 4, 5]


@task
def transform(num: int):
    return num, num + 10


@task
def load(x: int, y: int):
    print(x, y)


with Flow("flow") as flow:
    data = extract()
    x, y = transform.map(data, task_args={"nout": 2})
    load.map(x=x, y=y)


if __name__ == '__main__':
    flow.run()
```
```
[2021-07-30 09:56:14+0300] INFO - prefect.FlowRunner | Beginning Flow run for 'flow'
[2021-07-30 09:56:14+0300] INFO - prefect.TaskRunner | Task 'extract': Starting task run...
[2021-07-30 09:56:14+0300] INFO - prefect.TaskRunner | Task 'extract': Finished task run for task with final state: 'Success'
[2021-07-30 09:56:14+0300] INFO - prefect.TaskRunner | Task 'transform': Starting task run...
[2021-07-30 09:56:14+0300] INFO - prefect.TaskRunner | Task 'transform': Finished task run for task with final state: 'Mapped'
[2021-07-30 09:56:14+0300] INFO - prefect.TaskRunner | Task 'transform[0]': Starting task run...
[...]

Issue #4839, opened 2021-08-04
Handling of Multiple Return Values in Mapped Tasks


marcm-ml:
## Current behaviour
When a task that returns multiple values is mapped, the argument "nout" in the task declaration or the type annotation has no effect. When providing task_args={"nout": 2} in the task.map() call, the task can be unpacked. However, it has -in my opinion- not the desired effect compared to using "nout" in non-mapped tasks.

The result of a mapped task with task_args={"nout": 2} are just the ascending entries in the mapped list (here: list[0] and list[1]). This behaviour is unintuitive and quite different to the behaviour in non-mapped tasks when using "nout".

In cases where complex mapped tasks return multiple arguments which are also processed differently later in the flow, this behaviour leads to unnecessary handling of this mapped list in follow up tasks.

## Proposed behaviour
The .map() function should have its own "nout" argument such that multiple return values can be handled similar to non-mapped tasks. This can be achieved by using an intermediat task which reduces the result of the mapped task and uses `tuple(zip(*mapped_result))` to unpack the return values. See examples below. However, in complex parallelized flows it is desirable to not reduce each mapped task due to performance reasons. Therefore, the proposed behaviour should work on a deeper level without reducing to keep the benefit of consecutive mapped tasks. 

### Current behaviour:

    from prefect import task
    from typing import Tuple

    @task
    def task_nout(x: int):
        return x-1, x+1
    
    with Flow("nout") as flow:
        dec, inc = task_nout.map(x=[1, 2, 3, 4, 5], task_args={"nout": 2})
    
    result = flow.run()
    print(result.result[dec].result)  # This returns (0, 2)
    print(result.result[inc].result)  # This returns (1, 3)

### Proposed behaviour:
With intermediate task:

    @task
    def task_nout(x: int):
        return x-1, x+1
    
    @task(nout=2)
    def zip_task(lt: list) -> Tuple[int, int]:
        return tuple(zip(*lt))
    
    with Flow("flow_fit") as flow:
        mapped_task = task_nout.map(x=[1, 2, 3, 4, 5])
        dec, inc = zip_task(mapped_task)
    
    result = flow.run()
    print(result.result[dec].result)  # This returns [0, 1, 2, 3, 4]
    print(result.result[inc].result)  # This returns [2, 3, 4, 5, 6]

With proposed "nout" argument in .map()

    @task
    def task_nout(x: int):
        return x-1, x+1
    
    with Flow("nout") as flow:
[...]

PR #4974, opened 2021-09-16
Mapping support for tasks with nout defined

Fraznist:
## Summary

Tuple expansion of __mapped__ tasks, where each mapped task returns a tuple, now created mapped instances of the GetItem task.

## Changes

- Adds a new helper function: Flow.is_mapped(Task)
  - Checks upstream edges of a given task to determine whether the task will be executed as a mapped task.
- Augment Task.__iter__ to support mapped tasks with nout specified:
  - Attempts to infer an active flow context, and whether a task is a mapped task in the active flow context.
  - If the task is mapped and has nout set, generates `nout` amount of GetItem tasks, each mapping over the original tasks output (a list of size `nout` tuples)

## Importance

Tuple expansion improves code readability in a pythonic way. Currently, unmapped parts of workflows can use tuple expansions to generate GetItem tasks, which is great. This PR enables the same readability to mapped workflows.

## Example
```
with Flow() as f:
    list_of_tuples = orig_and_x2_task.map([1,2,3]) # [(1, 2), (2, 4), (3, 6)]
    orig, mult = list_of_tuples
    # orig: [1, 2, 3]
    # mult: [2, 4, 6]
```

## Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Released on September 2, 2021.
 - Add `--idempotency-key` to `prefect run` - [#4928](https://github.com/PrefectHQ/prefect/pull/4928)
 - Add support for larger flow registration calls - [#4930](https://github.com/PrefectHQ/prefect/pull/4930)
 - Ignore schedules by default for CLI flow runs and add flag to run based on schedule for local only runs [#4817](https://github.com/PrefectHQ/prefect/pull/4817)
+- Add mapping support to tasks with nout specified for tuple expansion [#4974](https://github.com/PrefectHQ/prefect/pull/4974)
 
 ### Task Library
 
@@ -22,6 +23,7 @@ Released on September 2, 2021.
 - Enhancement: Log boto exceptions encountered in the in AWS `BatchSubmit` task - [#4771](https://github.com/PrefectHQ/prefect/pull/4771)
 - Breaking: Legacy Dremio authentication has been updated to the new pattern in `DremioFetch` - [#4872](https://github.com/PrefectHQ/prefect/pull/4872)
 - Fix: Use runtime arguments over init arguments instead of ignoring them for MySQL Tasks - [#4907](https://github.com/PrefectHQ/prefect/pull/4907)
+
 ### Fixes
 
 - Adjust log limits to match backend logic for better UX - [#4900](https://github.com/PrefectHQ/prefect/pull/4900)
@@ -29,6 +31,7 @@ Released on September 2, 2021.
 - API server settings are passed correctly to task workers when using Prefect Server - [#4914](https://github.com/PrefectHQ/prefect/pull/4914)
 - Do not attempt attempt to set `host_gateway` if using an unsupported Docker Engine version - [#4809](https://github.com/PrefectHQ/prefect/pull/4809)
 - Ignore jobs without a `flow_run_id` label in `KubernetesAgent.manage_jobs` - [#4934](https://github.com/PrefectHQ/prefect/pull/4934)
+
 ### Breaking Changes
 
 - Services run by `prefect server` cli are now local by default (listen to localhost instead of 0.0.0.0); use `--expose` if you want to connect from a remote location - [#4821](https://github.com/PrefectHQ/prefect/pull/4821)
@@ -42,6 +45,7 @@ Released on September 2, 2021.
 - [Kathryn Klarich](https://github.com/klarich)
 - [Tenzin Choedak](https://github.com/tchoedak)
 - [Vincent Xue](https://github.com/xuevin)
+- [Emre Akgün](https://github.com/Fraznist)
 
 ## 0.15.4 <Badge text="beta" type="success" />
 
@@ -50,6 +54,7 @@ Released on August 17, 2021.
 ### Docs
 
 - Add a getting started section with a quick start guide for both core and orchestration sections - [#4734](https://github.com/PrefectHQ/prefect/pull/4734)
+
 ### Enhancements
 
 - Expose Snowflake cursor type to SnowflakeQuery task arguments [#4786](https://github.com/PrefectHQ/prefect/issues/4786)
@@ -386,7 +391,7 @@ Released on April 5, 2021.
 
 ### Enhancements
 
-- Add terminal flow state handler override  - [#4198](https://github.com/PrefectHQ/prefect/issues/4198)
+- Add terminal flow state handler override - [#4198](https://github.com/PrefectHQ/prefect/issues/4198)
 - When manually set, `prefect.context.date` will be used to determine dependent values - [#4295](https://github.com/PrefectHQ/prefect/pull/4295)
 - `prefect.context.date` will be cast to a `DateTime` object if given a parsable string - [#4295](https://github.com/PrefectHQ/prefect/pull/4295)
 - Expand logging for `DaskExecutor`, including the cluster dashboard address (if available) - [#4321](https://github.com/PrefectHQ/prefect/pull/4321)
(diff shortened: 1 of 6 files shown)


Fraznist:
Will add tests, currently can't run pytest on my machine

src/prefect/core/flow.py
```diff
@@ -805,6 +806,24 @@ def edges_from(self, task: Task) -> Set[Edge]:
...
+        Returns:
+            - bool: `True` if the Task is mapped in this flow, `False` otherwise
+
+        Raises:
+            - ValueError: if `task` is not found in this flow
+        """
+        for edge in filter(lambda e: e.mapped, self.edges_to(task)):
+            return True
+
+        return False
```
madkinsz, line 825: I prefer something like rather than a `filter`/`lambda`
```suggestion
        return any(edge.mapped for edge in self.edges_to(task))
```

madkinsz (commented):
Sweet, this doesn't look to complicated. Thanks for contributing! I've got a few administrative requests. We'll likely have another reviewer for this as well.

CHANGELOG.md
```diff
@@ -1701,7 +1705,7 @@ Released on June 25, 2020.
 ### Fixes
 
 - Fix `all_extras` tag not being set during CI job to build image - [#2801](https://github.com/PrefectHQ/prefect/issues/2801)
-- Quiet *no candidate Cached states were valid* debug logging - [#2815](https://github.com/PrefectHQ/prefect/issues/2815)
+- Quiet _no candidate Cached states were valid_ debug logging - [#2815](https://github.com/PrefectHQ/prefect/issues/2815)
```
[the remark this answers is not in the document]
Fraznist: my bad didn't realize i was formatting on save:(

src/prefect/core/flow.py
```diff
@@ -805,6 +806,21 @@ def edges_from(self, task: Task) -> Set[Edge]:
             )
         return self.all_downstream_edges()[task]
 
+    def is_mapped(self, task: Task) -> bool:
```
cicdw, line 809: This may suggest that it's checking whether the flow is mapped - to be more explicit, can this be `is_mapped_task`?

src/prefect/core/task.py
```diff
@@ -755,7 +755,6 @@ def map(
                     )
                 )
         task_args = task_args.copy() if task_args else {}
-        task_args.setdefault("nout", None)
```
cicdw, line 758: For my own understanding, why did this need to be removed?

src/prefect/core/task.py
```diff
@@ -984,6 +983,20 @@ def __iter__(self) -> Iterator:
                 "pass `nout` to the task decorator/constructor, or provide a "
                 "`Tuple` return-type annotation to your task."
             )
+        flow: Flow = prefect.context.get("flow", None)
+        if flow and flow.is_mapped(self):
```
cicdw, line 987: If this runs outside of a flow context, then it will trigger the Line 1000 incorrectly; maybe a better way to handle this is to raise a `ValueError` if a flow is not found in context with a message like "Tasks cannot be unpacked outside of a Flow context."

cicdw (changes requested):
Thanks for the PR @Fraznist! I left some comments.

I will admit I'm quite nervous about the suggested changes here; the fact that using some syntactic sugar also restructures the underlying data is very surprising behavior to me.  Here's an example of why I find this surprising:

```python
@task(nout=2)
def my_tuple_task(x):
    return x - 1, x + 1

@task
def printer(obj):
    print(obj)

with Flow("example") as flow:
    mapped = my_tuple_task.map([1, 2, 3, 4])

    # mapped tasks return lists, so this prints a list of tuples
    printer(mapped)
    printer(mapped[0]) # grabs first element of list and prints the tuple
    printer(mapped[1]) # grabs second element of list and prints the tuple
    
    # for consistency with other python methods / behavior
    # I would expect `a` to be the first element of the list
    a, b = mapped
    printer(a) # prints new data structure 
```

To paraphrase, mapped tasks are analogous to lists, and the proposed changes break my expectation about how list unpacking works and results in what I view as inconsistent behavior between indexing and unpacking.

src/prefect/core/flow.py
Fraznist: sure, I was trying to be consistent with `flow.edges_from(task)`, which allows it to be read like a sentence.

src/prefect/core/task.py
Fraznist: If that line is kept, `__iter__` therefore task unpacking cannot be called on mapped tasks. `__iter__` throws a value error for tasks without nout specified. 
This used to be correct since we didn't allow task unpacking. However this PR aims to change that, i.e. it aims to bring nout support for mapped tasks.

src/prefect/core/task.py
Fraznist: The old implementation executes what currently is the line 1000 as well. Further down the call stack, `task.bind` raises an exception that instructs to use an active flow context. That behavior is consistent between both master and this PR.

Fraznist:
I see, this certainly breaks the consistency of behavior between python and prefect. I never thought about such a case. 
Personally, I think this is a field where prefect can afford to stray away from python behavior. Unpacking is a fixed size operation, analogous to tuples (and `nout`). When I see a task that returns a tuple, unpacking feels like a natural option, regardless if it is mapped or not. Even if `tuple_2_task.map()` returns a list of 2-tuples, when I see `a, b = tuple_2_task.map()` it is quite clear to me that a will be a list of first tuple elements, and b will be a list of 2nd. Because mapped tasks and lists usually aren't fixed size, people rarely use unpacking on them. 
I think this is a user friendly addition. Rather than strictly following the hierarchy list of tuples, prefect smartly matches the fixed size operation and the fixed size data structure. Again, I don't remember the last time I saw someone do unpacking on a list in python, that's what tuples are for.
Overall, it is your decision. The similarity between coding in python and developing prefect flows is what got me excited for prefect in the first place, so I am kind of doubting myself here. On one hand, it feels natural. On the other it breaks behavioral consistency with python.

src/prefect/core/task.py
cicdw: Right, but the catch is that line 1000 is _only_ valid for the old implementation, so it shouldn't be run or relied upon in this new implementation - raising an explicit error earlier on avoids depending on that deeper behavior which may change unexpectedly.

cicdw:
Yea, I do understand that for your use case this may be a valuable convenience, but because this doesn't match Python's behavior it is effectively new feature behavior that we'd want to have very strong evidence for implementing and maintaining (there are _always_ edge cases with mapping haha).

Let's leave this open for a bit to see if we can design an alternative way of achieving this convenience

Fraznist:
Something like a wrapper would be simpler to implement than embedding this functionality in `Task.__iter__`. Something like:
```
@task(nout=2)
def my_tuple_task(x):
    return x - 1, x + 1
    
with Flow("example") as flow:
    list_of_tuples = my_tuple_task.map([1, 2, 3]) # [(0, 2), (1, 3), (2, 4)]
    a, b = transpose(list_of_tuples) 
    # transpose internally generates 
    # GetItem().map(list_of_tuples, unmapped(0)), GetItem().map(list_of_tuples, unmapped(1))
```
Could also be augmented by kwargs to only extract certain tuple indexes. Also wouldn't need the upstream task to be mapped, It could be an authentic list of tuples geenrated by a single task in this case.

cicdw:
Hey @Fraznist I think that's a great idea -- we could put it in [`utilities/tasks.py`](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/utilities/tasks.py)

madkinsz:
Linking to issues this would close #4824 and #4839

@pauleikis @sm-Fifteen @Nesqulck does this `to_apply` workaround example apply to your use-cases?

sm-Fifteen:
@madkinsz: Looking at the given examples and the included tests, yes, that would work with the use cases I needed.

Nesqulck:
It appears that this method using apply_map works around the reduce issue compared to using the zip method described in #4839. This would work in consecutive mapped tasked quite well. However, it is considerably slower if the next task is not mapped (aka reduce task). This effect is even more pronounced if we have many more return values since we effectively iterate nout times over the result of the mapped task to get the different return values. This also applies to the transpose warpper proposed above. Hence, a simple reduce task and the much more optimized zip function is more desirable in this case.

Point is: I guess I would also want a more concise and obvious way to unpack a task into its return values regardless of mapping or no mapping. Furthermore, I would also vote for keeping python's behaviour regarding earlier discussion. So, either a container such as transpose or another argument similar to nout should be used.

Im pretty sure such a behaviour can be achieved somewhere in FlowRunner or TaskRunner. I tried but admittingly this was a little over my head.

madkinsz:
Thanks for your thoughts! We'll need to look into this a bit further for high performance cases, it may be worth getting a functional interface in and addressing a performant implementation separately.

Nesqulck:
Any updates on this?

madkinsz:
I'm happy to review an implementation using the pattern from https://github.com/PrefectHQ/prefect/pull/4974#issuecomment-924061147 -- I'm busy stabilizing existing interfaces though and can't tackle it myself right now.

madkinsz:
Closing as stale, thanks for the discussion everyone! My above comment about reviewing a simpler implementation still stands.

PR #4980, opened 2021-09-17
Specify dockerignore for docker storage

### LINKED ISSUE #4931: Add option to `Docker` storage to respect `.dockerignore` file if using `files` argument

**michael.ball**: Hey there. I’ve been playing around with Docker storage today, trying to get all source code packaged together with the flows each time they are registered, and am using the `files` and `env_vars`  attributes as outlined in the Docs. But it seems that my `.dockerignore` file (in the directory from which I am registering the flow) is ignored by this build process. I have massive data directories nested at different levels inside the codebase so it’s essential there be a way to include this somehow. Anyone come across this / have any ideas? I’m curious about playing around with a custom Dockerfile, but the code that copies flows to a tmp Dockerfile in `create_docker_file` is dynamic so it seems tough that I could write a standalone Dockerfile to handle this case.

**kevin701**: Hey <@U02CELMJ5EJ>, this is the first I’ve seen this. I’ll look into it.

**zach555**: Hey <@U02CELMJ5EJ>, can you try including `.dockerignore` in the `files` arg for the storage?

**michael.ball**: Interesting… I’ll check it out :slightly_smiling_face:

**michael.ball**: Actually wait, the `.dockerignore` is already there because it falls within the directory I’m trying to copy. I just tried to explicitly copy it as a separate `files` arg, but same deal.

**michael.ball**: It’s actually a bit strange… the `create_docker_file` outputted script just has a standard COPY command in it, which I thought was supposed to obey `.dockerignore` … Perhaps the build context is not where I think it is

**zach555**: &gt; Perhaps the build context is not where I think it is
Yes! (unless you've looked through the internals of Docker storage)

When building Docker storage, Prefect creates a temporary directory that contains a new Dockerfile (created by Prefect) and also copies over all your files. For the `.dockerignore` to be respected, it will need to be copied into the root of this temporary directory <https://github.com/PrefectHQ/prefect/blob/master/src/prefect/storage/docker.py#L435>

**zach555**: But I'm not sure how exactly to do that :thinking_face:

**michael.ball**: Ah well, there’s the answer! But yes I’m also not immediately sure how to handle that nicely. I think that a .dockerignore path to go with the dockerfile one would be cool :shrug: but it looks unfixable without changing the Prefect source

**zach555**: Yeah I can open an issue for this regardless of workarounds
will it work if you specify `files` like so?

```
files = {
    ...
    "/path/to/.dockerignore": ".dockerignore"
}
```

**michael.ball**: Unfortunately not, because that just creates a COPY statement inside the generated `Dockerfile` that copies the `.dockerignore` inside the image, whereas the docker build process needs to read the `.dockerignore` file before starting to build (at least, this is how I understand why it doesn’t work)

**michael.ball**: I might be able to work around it! I saw that if I specify a Dockerfile then it uses the current directory as build context, and in that case Docker would find the `.dockerignore` file automatically

**michael.ball**: Will return to it tomorrow and lyk! Thanks for your help

**zach555**: <@ULVA73B9P> open "Add option to `Docker` storage to respect `.dockerignore` file if using `files` argument"

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1630600807008500?thread_ts=1630600807.008500&cid=CL09KU1K7).

anticorrelator:
Resolves #4931
When writing a `Dockerfile` using the `files` argument, any `dockerignore` files won't be copied to the new context, 
this change provides an option to specify a `dockerignore` file when using Docker storage.

This PR:


madkinsz:
@anticorrelator where did we land on this implementation?

madkinsz (approved):

PR #4981, opened 2021-09-18
Add data parameter to SQLiteQuery

dsaxton:
## Summary

This allows for query parameterization when running queries against a SQLite database using the `SQLiteQuery` task. Currently if users want to include variables in a query string they'll have to do manual string formatting / escaping which is error-prone and in some cases dangerous. Adding this capability also aligns the API with that of other database tasks such as `PostgresExecute`.

Side comment, but since other database engines are in their own directories (e.g., `postgres`, `mysql`), should this directory be called `sqlite` instead of `database`?

## Changes

Adds an optional `data` argument to the `SQLiteQuery` task.

## Importance

More API consistency and helps end users avoid certain types of errors.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/database/sqlite.py
+++ b/src/prefect/tasks/database/sqlite.py
@@ -16,17 +16,22 @@ class SQLiteQuery(Task):
         - query (str, optional): the optional _default_ query to execute at runtime;
             can also be provided as a keyword to `run`, which takes precedence over this default.
             Note that a query should consist of a _single SQL statement_.
+        - data (tuple, optional): values to use when `query` is a parametrized string. See
+            https://docs.python.org/3/library/sqlite3.html for more details.
         - **kwargs (optional): additional keyword arguments to pass to the
             standard Task initalization
     """
 
-    def __init__(self, db: str = None, query: str = None, **kwargs: Any):
+    def __init__(
+        self, db: str = None, query: str = None, data: tuple = None, **kwargs: Any
+    ):
         self.db = db
         self.query = query
+        self.data = data
         super().__init__(**kwargs)
 
     @defaults_from_attrs("db", "query")
-    def run(self, db: str = None, query: str = None):
+    def run(self, db: str = None, query: str = None, data: tuple = None):
         """
         Args:
             - db (str, optional): the location of the database (.db) file;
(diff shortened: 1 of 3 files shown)


dsaxton:
CI failure looks unrelated to the change

src/prefect/tasks/database/sqlite.py
```diff
@@ -16,35 +16,45 @@ class SQLiteQuery(Task):
...
-        with closing(sql.connect(db)) as conn:
-            with closing(conn.cursor()) as cursor:
+        data = cast(tuple, data)
+        with closing(sql.connect(db)) as conn, closing(conn.cursor()) as cursor:
+            if data is None:
                 cursor.execute(query)
-                out = cursor.fetchall()
-                conn.commit()
+            else:
+                cursor.execute(query, data)
```
madkinsz, line 55: Is there a way to pass `data` in both cases and leave out this branch?

madkinsz (commented):
Hi! Thanks for contributing. This makes sense, one minor comment.

Don't worry about the CI failure, that one is flaky and we can rerun.

> Side comment, but since other database engines are in their own directories (e.g., postgres, mysql), should this directory be called sqlite instead of database?

Probably but we can't move it easily now. To be addressed at the next major version change.

src/prefect/tasks/database/sqlite.py
dsaxton: We can't pass `None` data directly without getting an error (I think it wants there to be an empty tuple for an unparametrized query; I added another test just to more explicitly hit that code path). We could maybe inline it as something like `cursor.execute(query, () if data is None else data)` if that's a bit cleaner?

src/prefect/tasks/database/sqlite.py
```diff
@@ -16,35 +16,42 @@ class SQLiteQuery(Task):
...
+        self, db: str = None, query: str = None, data: tuple = (), **kwargs: Any
+    ):
         self.db = db
         self.query = query
+        self.data = data
         super().__init__(**kwargs)
 
     @defaults_from_attrs("db", "query")
-    def run(self, db: str = None, query: str = None):
+    def run(self, db: str = None, query: str = None, data: tuple = ()):
```
madkinsz, line 34: @dsaxton did you mean to add `data` to `defautls_from_attrs` ? That might break with the `()` vs `None`

dsaxton: Thanks, I missed that

dsaxton:
@madkinsz Gentle ping, the CI failure looks unrelated if you think this might be ready

madkinsz (commented):
Sorry about the delay. I meant to ask for one more test covering setting `data` on `__init__` instead of `run` since I think `defaults_from_attrs` may misbehave there.

dsaxton:
> Sorry about the delay. I meant to ask for one more test covering setting `data` on `__init__` instead of `run` since I think `defaults_from_attrs` may misbehave there.

Added a new test, let me know if it's what you had in mind or if you'd like anything cleaned up

madkinsz (approved):
LGTM; thanks! Did you want to add yourself as a contributor in the changelog entry?

dsaxton:
> LGTM; thanks! Did you want to add yourself as a contributor in the changelog entry?

Sure, thanks for the review :+1:

madkinsz (approved):

PR #4983, opened 2021-09-18
Fix broken doc links

dsaxton:
## Summary

It looks like the links for "Task Looping" and "Signalling" are broken here: https://docs.prefect.io/core/tutorial/07-next-steps.html#advanced-control-structures. I made some semi-educated guesses about where these should point, but please let me know if this should be changed. Thanks!


madkinsz (approved):

madkinsz:
Thanks!

PR #4996, opened 2021-09-23
Add flatten support to apply_map

Fraznist:
## Summary

Adds `flatten` support to `apply_map`.

## Changes

`flatten` edge annotations were missing in the implementation of `apply_map` . Now, an edges flatten property is derived via the same way its mapped property is derived, and all instances of `flow.add_edge` in `apply_map` use this flattened edge indicator. 

## Importance

Flattening nested structures before mapping over is a widespread use case. Complex mapping pipelines that necessitate `apply_map` should be able to benefit from `flatten` as well. 
Old behavior silently disregards the flatten edge annotation, therefore I think this is strictly an enhancement, and hope it doesn't make any breaking changes.

## Side Note
- I will add tests soon, and then make this an actual PR.
- Admittedly, I don't fully understand the `apply_map` implementation. I just put flatten bool values to where they are allowed to go :sweat_smile: 

## Checklist

This PR:

### DIFF
--- a/changes/pr4996.yaml
+++ b/changes/pr4996.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - 'Add flatten support to apply_map [#4996](https://github.com/PrefectHQ/prefect/pull/4996)'
+
+contributor:
+  - '[Emre Akgün](https://github.com/Fraznist)'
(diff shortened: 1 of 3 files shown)


madkinsz:
It's a complicated bit of code to grasp the implementation :) thanks for contributing! I think this makes sense at a glance, looking forward to the tests.

madkinsz (approved):

PR #5001, opened 2021-09-27
Kubernetes agent job YAML allowing imagePullSecrets to be unset as code intended

AndrewFarley:
## Summary
Allowing imagePullSecrets in Kubernetes agent to be unset in job YAML when it is unused.  This patch fixes compatibility issues with releases of Kubernetes before (v1.19.13, v1.20.9, v1.21.3) because it triggers this bug https://github.com/kubernetes/kubernetes/pull/103133 which manifests issues such as this https://github.com/kubernetes/autoscaler/issues/4353 

## Changes
Very minor tweak that allows what appears to be intended to be possible which is to unset the imagePullSecrets key entirely if this value is unset or empty.  The supporting code in this file has certain code paths to `del` this key which would never be possible without this adjustment (lines 581 and 896 of `agent/kubernetes/agent.py`)

## Importance
Fixes an issue with cluster-autoscaler and Kubernetes releases prior to v1.19.13, v1.20.9, v1.21.3 by removing an optional and empty-set that doesn't need to be set if unused.  A google on this topic shows that this usage of an empty map has caused issues in a few other libraries and frameworks.

**Note:** This change has zero impact for functionality it merely adds what I believe was intended in the original code and thus should probably not require additional work such as tests or references.

## Checklist
- [X] (N/A) adds new tests
- [X] (N/A) adds a change file in the `changes/` directory
- [X] (N/A) updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs

### DIFF
--- a/changes/pr5001.yaml
+++ b/changes/pr5001.yaml
@@ -1,3 +1,6 @@
 
 fix:
   - "Allow unsetting imagePullSecrets with an empty string - [#5001](https://github.com/PrefectHQ/prefect/pull/5001)"
+
+contributor:
+  - "[Andrew Farley](https://github.com/AndrewFarley)"
--- a/changes/pr5001.yaml
+++ b/changes/pr5001.yaml
@@ -1,0 +1,3 @@
+
+fix:
+  - "Allow unsetting imagePullSecrets with an empty string - [#5001](https://github.com/PrefectHQ/prefect/pull/5001)"
--- a/changes/pr5001.yaml
+++ b/changes/pr5001.yaml
@@ -3,4 +3,4 @@ fix:
   - "Allow unsetting imagePullSecrets with an empty string - [#5001](https://github.com/PrefectHQ/prefect/pull/5001)"
 
 contributor:
-  - "[Andrew Farley](https://github.com/AndrewFarley)"
+  - "[Farley Farley](https://github.com/AndrewFarley)"
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -113,7 +113,7 @@ class KubernetesAgent(Agent):
         )
         if image_pull_secrets is None:
             image_pull_secrets_env = os.getenv("IMAGE_PULL_SECRETS")
-            if image_pull_secrets_env is not None and image_pull_secrets_env != '':
+            if image_pull_secrets_env is not None and image_pull_secrets_env != "":
                 image_pull_secrets = (
                     [s.strip() for s in image_pull_secrets_env.split(",")]
                     if image_pull_secrets_env is not None
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -113,11 +113,14 @@ class KubernetesAgent(Agent):
         )
         if image_pull_secrets is None:
             image_pull_secrets_env = os.getenv("IMAGE_PULL_SECRETS")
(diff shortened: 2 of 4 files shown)


madkinsz:
Thanks for the contribution! Is it feasible to modify a test to ensure this is working as intended? Seems like the old test didn't cover this correctly if it was passing.

madkinsz:
Hey @AndrewFarley -- gentle ping to see if you're still planning on working on this. If not, we can take it over.

AndrewFarley:
> Hey @AndrewFarley -- gentle ping to see if you're still planning on working on this. If not, we can take it over.

@madkinsz I haven't had time to work on tests for this.  But this is absolutely a breaking issue for us and I'm sure anyone else using a modern Kubernetes cluster.  If you want to take over this please do, I started looking into this issue and I couldn't figure out how to add tests to confirm this working, and I realized that I need to update the "deployment" not just the "job" (I use the Job runner)

AndrewFarley:
Yeah, doesn't look like I have the time for this, too much work.  Please if possible someone carry this to the finish line.  It needs...

- Adding this modification/improvement to `Deployment` not just the `Job` interface
- Adding tests for both Deployment and Job to verify the Kubernetes YAML generated does not include image_pull_secrets if it is not set

AndrewFarley (commented):
Nice, thanks, LGTM!  I'd say we can either...
- Merge this as-is, and make another Issue/PR for "Deployment" support of this bugfix
- Wait for someone to add code to update Deployment and tests for fixing this

Any preferences?

abrookins:
Hey @AndrewFarley, thanks for the PR and I'm glad I can help! I'll test this on a k8s cluster tomorrow and update the ticket.

I wasn't able to reproduce the same issue with the deployment yaml under test. An empty string passed in for `image_pull_secrets` results in us unsetting the `imagePullSecrets` key as expected, rather than rendering the key with an empty value. So I think the deployment case works, but I may have misunderstood the issue, so let me know if that's the case!

AndrewFarley:
> Hey @AndrewFarley, thanks for the PR and I'm glad I can help! I'll test this on a k8s cluster tomorrow and update the ticket.
> 
> I wasn't able to reproduce the same issue with the deployment yaml under test. An empty string passed in for `image_pull_secrets` results in us unsetting the `imagePullSecrets` key as expected, rather than rendering the key with an empty value. So I think the deployment case works, but I may have misunderstood the issue, so let me know if that's the case!

Ahh, then I just didn't test that path since I'm not using it.  I assumed further changes were necessary but if that works then we're golden!

AndrewFarley:
Ready for merge!

madkinsz:
Let's add a fix changelog entry for this and then it's good to go.

abrookins:
@madkinsz This is ready for a final pass. I found an additional related issue with `KubernetesRun` while testing and fixed it. I also simplified the logic a bit.

abrookins:
Hey @AndrewFarley, I added a changelog entry to this PR that lists you as the contributor for this fix. Just wanted to double check though - is that OK with you?

AndrewFarley:
@abrookins No complaints about that, but changed your naming of my name since my legal name changed some years back.  And yeah, that's my legal name now.  :P  Good to merge!

abrookins:
@AndrewFarley Oh, I missed that, apologies and thank you! 🙌 

@madkinsz Changelog is in!

madkinsz (approved):

Issue #3966, opened 2021-01-15
Possible memory leak when using DaskExecutor in Kubernetes


joelluijmes:
## Description

I’m running Prefect in Kubernetes. What I have is a flow which spawns 15 Kubernetes high CPU intense jobs. In order to do some parallelism, I have in the flow a DaskExecutor configured (6 workers, 1 thread).

What I see is the prefect-job which is created by the Kubernetes agent, uses quite some resources which grows over time. See attached screenshot.
![image](https://user-images.githubusercontent.com/4273556/104746448-97751e00-574f-11eb-8d25-a25d9346dde1.png)

Note: I’m talking about the job created by the prefect agent. The actual job executing code is only using 188MiB.

It seems like there is a possible memory leak in either Prefect or Dask. Is there a better alternative to deploy this? Which uses less resources? 

## Expected Behavior

Well first of all, I wouldn't expect this high memory usage. But most of all, not that it seems to be growing indefinitely. In this instance, the flow wasn't killed (yet) by Kubernetes due high memory usage, but that is something I was running into. 

## Reproduction

Reproduction is bit tricky as this flow is generated from a config file on running the script. I think I got all the relevant bits below:

```python
executor = DaskExecutor(cluster_kwargs={"n_workers": 6, "threads_per_worker": 1, "memory_limit": "1GB"})
schedule = CronSchedule(cron, start_date=datetime.utcnow())
handler = slack_notifier(only_states=[state.Failed])
storage_gcr = Docker(
    base_image="prefecthq/prefect:0.14.1-python3.8",
    registry_url="...",
    python_dependencies=["pandas"],
)
with Flow(
    f"CloudSQL to BigQuery ({cron})",
    storage=storage,
    executor=executor,
    schedule=schedule,
    run_config=KubernetesRun(
        # prefect_job_template contains a job spec describing resource limits, secrets, nodeselector..
        job_template=prefect_job_template
    ),
) as flow:
    # Class with @resource_manager which spawns off a pod and service to run CloudSQL Proxy
    with CloudSQLProxyManager(...) as service_name:
        # Job which calls RunNamespacedJob.run() on run
        job = CreateSyncJob(
            job_spec_template,
            state_handlers=[cleanup_on_failure, handler],
        )
        # Link the dependency of task to the CloudSQLProxyManager
        task = job(service_name)
```

Visualizing one of the flows show:
[...]

PR #5004, opened 2021-09-28
[WIP] Fix For DaskExecutor Bloating Memory

kvnkho:
##  The Problem

A lot of users have reported seeing Prefect consuming a lot of memory when using the DaskExecutor. Sample issues can be seen [here](https://github.com/PrefectHQ/prefect/issues/3966) and [here](https://github.com/pangeo-forge/pangeo-forge-recipes/issues/208). It was also hinted [here](https://github.com/dask/distributed/issues/2757). 

In the example of @sharkinsspatial in the `pangeo-forge` link above, a simple code snippet such as the one below causes the Dask scheduler to perform poorly (memory issues over a 8GB scheduler). The task does not even have a return.

```python
from time import sleep

@task
def map_fn(x):
    sleep(0.25)
task_range = list(range(1, 400000))
with Flow('map_testing') as flow:
    map_fn.map(task_range)
```

Yet, an equivalent snippet with pure Dask is able to run in a reasonable amount of time. Prefect is adding a significant amount of overhead to these submit calls. 

```python
client = Client(cluster)

def map_fn(x):
    sleep(0.25)

value_range = list(range(1, 400000))

for value in value_range:
    future = client.submit(map_fn, value)
    futures.add(future)

client.gather(futures)
```

##  Hypothesis

In order to understand where the overhead comes from, we need to look at the following Dask snippets. Snippet 1  is basically the pure Dask one from above. The size of `x` below is roughly 20 KB and the `sum_the_list()` function sums `x` with `i`. 

```python
x = list(range(1,2500))
n = 100000

_futures = []
for i in list(range(n)):
    _futures.append(client.submit(sum_the_list, x, i))
result = client.gather(_futures)
```

Compare this with a second snippet where we send a future ahead of time for the value of `x`. This way, the workers will already take care of getting a copy of `x` and we reduce data transfer. This is done by submitting the `dummy()` function ahead of time.

```python
def dummy():
    return x

_futures = []
data_fut = client.submit(dummy)
for i in list(range(n)):
    _futures.append(client.submit(sum_the_list, data_fut, i))
result = client.gather(_futures)
```
##  Profiling

We can then profile these two setups using the `performance_report` context manager of Dask. For each of these tests, I spun up a new cluster to avoid any unmanaged memory from affecting the numbers. 

```
from dask.distributed import performance_report

with performance_report(filename="dask-report-efficient.html"):
    ....
```

### Scenario 1 - Inefficient Execution

[This](https://gistcdn.githack.com/kvnkho/966b10bc83c7fefc9b6a02cf661e316d/raw/4be22be43912512602fa7659d2df6fb029b1f2cb/dask-report-inefficient.html) is the report for the inefficient case. Mean memory was 900 MB and total executor time was 640 seconds

### Scenario 2 - Efficient Execution

[This](https://gistcdn.githack.com/kvnkho/966b10bc83c7fefc9b6a02cf661e316d/raw/4be22be43912512602fa7659d2df6fb029b1f2cb/dask-report-efficient.html) is the report for the efficient case. Mean memory was around 400 MB and total execution time was 200 seconds. 

## The Problem in Prefect Terms

In Prefect, the FlowRunner currently submits things to the executor with the following code.

```python
# this is where each child is submitted for actual work
for idx, states in enumerate(list_of_upstream_states):
    .....
    submitted_states.append(
        executor.submit(
            run_task,
            task=task,
            state=current_state,
            upstream_states=states,
            context=dict(
                prefect.context,
                **task_contexts.get(task, {}),
                map_index=idx,
            ),
            flow_result=self.flow.result,
            task_runner_cls=self.task_runner_cls,
            task_runner_state_handlers=task_runner_state_handlers,
            upstream_mapped_states=upstream_mapped_states,
            extra_context=extra_context(task, task_index=idx),
        )
    )
```
The `context` alone here is around 9.8 KB, but we are also repeatedly sending the `task`, `flow_result`, `task_runner_cls`, `task_runner_state_handlers`, and `upstream_mapped_states`. It's hard to say what these add up to, but we can definitely ease the burden of the scheduler by moving relevant data to workers ahead of time.

### client.scatter()

The Dask client has a method called `client.scatter()`. This takes a piece of data and moves it across the workers instantly. The instant part is the main problem. If workers eventually get added due to autoscaling, they don't have the pieces of data that were sent previously.

### Using a future

This leads us to the second option, which is to send these variables as futures. When Dask needs them for an operation, it will take care of getting them to the relevant worker if not already present. This is what was shown in one of the earlier code snippets (posted again for quick viewing).:
```python
def dummy():
    return x

_futures = []
data_fut = client.submit(dummy)
for i in list(range(n)):
    _futures.append(client.submit(sum_the_list, data_fut, i))
result = client.gather(_futures)
```

## Potential Solutions

There are pretty much only two places the code can be edited to send the task data as futures. The first is [the FlowRunner](https://github.com/PrefectHQ/prefect/blob/f014529c66b3939279f1eaea10edffd735d81da8/src/prefect/engine/flow_runner.py#L595-L615) and the second is the [DaskExecutor](https://github.com/PrefectHQ/prefect/blob/f014529c66b3939279f1eaea10edffd735d81da8/src/prefect/executors/dask.py#L382-L409) code.

Note that the FlowRunner makes 100 submit calls for a mapped task with 100 elements. So at the executor level, all that is seen is the individual element of the mapped tasks.

### Editing `DaskExecutor`

I wanted to purely edit `DaskExecutor`, but I don't think it's reliable. The process would be editing the `submit` call to be responsible for passing the futures. The problem is, the `DaskExecutor` would need to check if the futures already exist to know whether to do this submission or if the submission was already done. Checking for future existence on the schedule is wonky. 

You can try it with
```python
print(client.futures)
```
which returns:
`{'dummy-7c9a8c022791a7950f23fcb1179657a2': <FutureState: finished>}`

This means that your functions need to be properly named in order to distinguish the Futures. You would need to parse this string, which could easily lead to false positives. But then even from there, it's unclear how to get a reference to this `dummy()` and then pass it to my `client.submit()` call even if I know the future is there.

### Editing `FlowRunner`

This made me positive I needed to edit the flow runner. Before the entry of the loop that submits mapped tasks individually, we can already submit a future to the workers, and then feed that future into `executor.submit()`. That is the basis on the PR, and then the rest of the work is adding the `scatter` method across all Executors. 

## To-Do

This standalone PR should reduce the memory footprint by a bit already. The only problem is that context is a bit harder to untangle because it is like 90% the same across mapped tasks but 10% different. We need to split it and `submit()` the 90% and then combine it with the 10% inside the task. This involved editing the TaskRunner.

This PR:

### DIFF
--- a/src/prefect/engine/flow_runner.py
+++ b/src/prefect/engine/flow_runner.py
@@ -576,6 +576,20 @@ class FlowRunner(Runner):
 
                     submitted_states = []
 
+                    # the next loop submits tasks by repeatedly using executor.submit()
+                    # this causes a lot of duplication unless we use futures to send
+                    # repeatedly used data to the workers ahead of time.
+                    def _no_op(x):
+                        return x
+
+                    # TO-DO: context needs to be submitted too but it is more intertwined
+                    task_fut = executor.submit(_no_op, task, _scatter=True,)
+                    # context_fut = executor.submit(lambda *args: context)
+                    flow_result_fut = executor.submit(_no_op, self.flow.result, _scatter=True)
+                    task_runner_class_fut = executor.submit(_no_op, self.task_runner_cls, _scatter=True)
+                    task_runner_state_handlers_fut = executor.submit(_no_op, task_runner_state_handlers, _scatter=True)
+                    upstream_mapped_states_fut = executor.submit(_no_op, upstream_mapped_states, _scatter=True)
+
                     for idx, states in enumerate(list_of_upstream_states):
                         # if we are on a future rerun of a partially complete flow run,
                         # there might be mapped children in a retrying state; this check
@@ -596,7 +610,7 @@ class FlowRunner(Runner):
                         submitted_states.append(
                             executor.submit(
                                 run_task,
-                                task=task,
+                                task=task_fut,
                                 state=current_state,
                                 upstream_states=states,
                                 context=dict(
@@ -604,10 +618,10 @@ class FlowRunner(Runner):
                                     **task_contexts.get(task, {}),
                                     map_index=idx,
                                 ),
-                                flow_result=self.flow.result,
-                                task_runner_cls=self.task_runner_cls,
-                                task_runner_state_handlers=task_runner_state_handlers,
-                                upstream_mapped_states=upstream_mapped_states,
+                                flow_result=flow_result_fut,
+                                task_runner_cls=task_runner_class_fut,
+                                task_runner_state_handlers=task_runner_state_handlers_fut,
+                                upstream_mapped_states=upstream_mapped_states_fut,
                                 extra_context=extra_context(task, task_index=idx),
                             )
                         )
--- a/src/prefect/executors/base.py
+++ b/src/prefect/executors/base.py
@@ -41,6 +41,7 @@ class Executor:
         Returns:
             - Any: a future-like object
         """
+        # _scatter is a reserved keyword to send futures distributedly
         raise NotImplementedError()
 
     def wait(self, futures: Any) -> Any:
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -414,6 +414,15 @@ class DaskExecutor(Executor):
         if self.client is None:
             raise ValueError("This executor has not been started.")
 
+        # _scatter == True will send the futures to Dask workers ahead of time to
+        # avoid resubmission and save memory. `client.submit()` handles autoscaling
+        # while `client.scatter()` does not
+        _scatter = kwargs.pop("_scatter", False)
+
+        if _scatter:
+            fut = self.client.submit(fn, *args, **kwargs)
+            return fut
+
         kwargs.update(self._prep_dask_kwargs(extra_context))
         if self._should_run_event is None:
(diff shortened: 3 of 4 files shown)


BitTheByte:
@kvnkho can we get an update for this?

kvnkho:
Yeah I gave it a shot fairly recently (last 2 weeks of 2021) and replaced the task submission with futures (everything except context because that is harder). I found no improvement from doing that. It might be that context is the meat of the bloat, but context is quite intertwined in a few places so it's harder to create a future for the context.

If I am unable to get anything working soon, then the focus will probably be getting this right from Orion.

BitTheByte:
As I understood Orion doesn't have this problem currently?  I've got a full ecosystem built on top of Prefect and was surprised to find such a problem it's really hard to replace it at this stage

madkinsz:
@BitTheByte — Orion is naively submitting tasks to Dask right now so it likely has a similar problem. I'll be looking into solving this in Orion in the next few weeks. Hopefully, I'll learn enough to backport a solution to here.

@kvnkho — I'm going to close this for now. Thanks for investigating.

Issue #3836, opened 2020-12-15
Dask Client.profile artifact


bnaul:
Just a general proposal for now: what would folks think of an artifact that calls Client.profile and uploads the output? Might have to just persist the HTML version to some storage location and link to that, probably the full flame graph info would not really translate to Markdown very easily.

Still getting a handle on what artifacts are good for so if this makes no sense let me know and it won’t hurt my feelings 😄. But if it does make sense I’d be happy to implement it.

jcrist:
Hmmm, I can see this being useful, but I'm not sure how to fit it into the current system. A few notes:

- Currently artifacts are only at the task level, while you'd want this to be at the flow level (unless you wanted to add a task to do it at the end of a flow run). This would require some backend changes, but nothing terribly complicated.
- Dask profiles contain potentially proprietary info about your code, we wouldn't want to store this in prefect's system. As such, the most we could do is store a link to the profile stored elsewhere (but then this is further complicated, since user's would want to configure the storage location/backend). They can also be quite large, so we wouldn't want to store them in our DB anyway.

You could do this fully in userspace though, with a custom task added to the end of a flow run. This could be done with a `resource_manager` to ensure it always runs. Something like (untested pseuducode):

```python
import prefect
from distributed import worker_client

@prefect.resource_manager
class dask_profile:
    def __init__(self, file_path):
        # could also have the upload step create the file name, so no explicit path needed
        self.file_path = file_path

    def setup(self):
        pass

    def cleanup(self, _):
        with worker_client() as client:
            profile = client.profile()
        upload_to_path(profile, self.file_path)
        prefect.artifacts.create_link(self.file_path)

with Flow("your-flow"):
    with dask_profile("s3://bucket/path/to/profile.json"):
        # put all your tasks in here
        # the profile will run after all contained tasks, regardless of success/failure
        pass
```

Something like that might suit your needs, and wouldn't need to be contained in prefect core at all. It also gives you full flexibility with what you want to upload and where, something we'd likely have to make configurable if we put it in the main library.

bnaul:
Thanks @jcrist! Something like the above was what I had in mind for now; we had talked in https://prefect-community.slack.com/archives/CL09KU1K7/p1602269846235100 about how I'm currently persisting these w/o any artifact integration. I thought it might make sense to move upstream to Core since it feels pretty broadly applicable, but I agree it seems hard to support as just a drop-in since there's a few user configuration steps and it doesn't quite fit the standard pattern.

PR #5032, opened 2021-10-08
Add dask performance report

### LINKED ISSUE #5031: Suport for Dask Performance Report on Task/Flow Runs
## Current behavior

It's very hard to get a Dask performance report for a Flow. These performance reports are very useful for memory profiling and checking if tasks cause memory issues. More on the performance report can be found [here](https://distributed.dask.org/en/latest/api.html#distributed.performance_report). It generates an HTML doc. This would also help debug flows that end up in a stuck RUNNING state.

Without Prefect, using it looks like:

```python
with performance_report(filename="dask-report.html"):
    return dd.read_csv(filepath)
```

But with Prefect, it becomes very hard to use it. Trying 

```python
@task(checkpoint=False)
def using_checkpoint_false(filepath):
    with worker_client():
        with performance_report(filename="dask-report.html"):
            return dd.read_csv(filepath)
```

only gives the report for one task. This means that for a mapped task with 10 elements, you get 10 HTML files.

The other attempt

```python
with Flow("example") as flow:
    with performance_report(...):
        some_task()
        other_task()
```

just throws an error.

## Proposed behavior

I may be wrong but this needs to be done at the `executor.submit()` level.

```python
    with performance_report(...):
        executor.submit()
```

which means that we need changes in the FlowRunner to support this. The FlowRunner takes a mapped task and calls `executor.submit()` for each map iterator. This means that to get the report for a mapped task, it needs to be done at the FlowRunner level.

## Open Questions
Do we support this at the task level or Flow level?
How can it be done that doesn't make the solution Executor specific? (the FlowRunner is currently executor agnostic)

zangell44:
## Summary

Adds the ability to generate a performance report for a flow run using a DaskExecutor. The performance report is an html file that is viewable in a web browser.

The performance report is saved in a user specified path and available as a property, `flow.executor.performance_report` in a flow's terminal state handler.

I also considered displaying the performance report in the UI. However, the performance report contains lots of javascript that we can't execute in our UI. Certainly worth further investigation if anyone has ideas.

As a bonus I've gotten rid of the 5000 lines of ReferenceError's in the netlify build.

## Importance

Closes #5031 

## Checklist

This PR:

### DIFF
--- a/docs/orchestration/flow_config/executors.md
+++ b/docs/orchestration/flow_config/executors.md
@@ -278,43 +278,18 @@ flow.executor = DaskExecutor(address="tcp://...")
 
 To generate a [performance report](https://distributed.dask.org/en/latest/api.html#distributed.performance_report)
 for a flow run, specify a `performance_report_path` for the `DaskExecutor`.
-
-#### How can I access the performance report?
+#### Saving the report to a local file
 
 The performance report will saved as a `.html` file where the flow run is executed based on the `performance_report_path`.
 To view the report, open the html file in a web browser.
 
 For local execution or flows executed using a Local Agent, the file will be accessible on your local machine.
 
-For other agent types, the report file location is not guaranteed to be easily accessible after execution.
-When using a Kubernetes Agent, for example, the report will be saved on the Kubernetes pod responsible for
-executing the flow run.
-
-For cases in which the performance report location is not easily accessile after flow execution, the report 
-is also available as a string in the flow's terminal state handler, which can be used to write the report to a convenient location
-by accessing `flow.executor.performance_report`.
-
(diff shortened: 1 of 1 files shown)


bobcolner:
Will it be possible to access the dask performance report in stateless/standalone prefect core/dev mode? Would love to see and example of that 'flow'.

Thanks

bnaul:
> I also considered displaying the performance report in the UI. However, the performance report contains lots of javascript that we can't execute in our UI. Certainly worth further investigation if anyone has ideas.

In https://github.com/PrefectHQ/prefect/issues/3836 we had discussed making an artifact that just links to the report, maybe that’s an option here?

Issue #4402, opened 2021-04-14
Avoid rate limiting while using ECS Agent


madkinsz:
There are very low rate limits in task definition registration with the same name and now that we are creating task definitions for every flow run and using the flow slug as the name users can easily encounter this limit. We need to

- Change task definition names to avoid frequent collisions
- Look at setting up boto3 retries

We can change the task definition name to include a short unique id or we can consider (and I think this may be the best option) using the flow run id instead of the slug. i.e. `prefect-flow-run-{id}`

We should be able to setup a retry via the boto3 client e.g.

```python
import boto3
from botocore.config import Config

config = Config(
   retries = {
      'max_attempts': 10,
      'mode': 'adaptive'
   }
)

ec2 = boto3.client('ec2', config=config)
```

I think that change may require a look from someone on our team so we can ensure that the config is handled properly across our uses of the boto client.

Originally worked on by @joe1981al at https://github.com/PrefectHQ/prefect/issues/4380

joe1981al:
@madkinsz Just an update, even with unique IDs we are still hitting rate limits for register / deregister.  Not sure if boto3 is using the environment variables or if it's overwritten by [L182-184](https://github.com/PrefectHQ/prefect/blob/bbb6c49bb3bfefd88e420a1b9e50565278dbbc62/src/prefect/agent/ecs/agent.py#L182) 

I'm wondering if there is a way to have the agent code generate the task definition then have that first task definition's ARN written back to the run_config for the flow...  Either default or "reuse-task-definition" option...

madkinsz:
@joe1981al can you show what errors you're encountering? I thought the rate limits were for the task definition family names and a unique id would remove that bottleneck?

joe1981al:
@madkinsz having a unique ID did pass the rate limit on the family names one as it has a lower rate limit than overall register / deregister.  Amazon won't publish rate limits for ECS as they say they are dynamic...  

`An error occurred (ThrottlingException) when calling the DeregisterTaskDefinition operation: Rate exceeded`

Rongrs:
Hi, any progress on that? 
This issue is preventing us from using ECS agents right now, and it looked like there's a valid solution, no?

joe1981al:
@Rongrs 

You could try setting these env variables, it does help:

```
- Name: AWS_RETRY_MODE
  Value: adaptive
- Name: AWS_MAX_ATTEMPTS
  Value: 100
```

ryan-cf:
[...]

PR #5059, opened 2021-10-14
Avoid ecs rate limiting

anticorrelator:
## Summary
Attempts to address #4402. As discussed with @jakekaplan, we've decided to try and minimize the ECS agent's contribution to hitting the AWS rate limit by disambiguating the flow-run slug with the flow-run id.

- Because people might be using the slug name being associated with the flow in the AWS console, we've only appended the run id
- The adaptive retry method is currently [experimental](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes-adaptive), and users can easily configure their aws cli settings, we think we should leave it alone until we learn whether or not using an adaptive retry mode will resolve this issue for all users.


madkinsz:
Also, the original issue added a new truncated UUID to each definition whereas this reuses the full flow run id.

PR #5066, opened 2021-10-19
Fixed ReadTimeoutErrors for kubernetes client after connection reset by keep-alive connections

jonasmiederer:
## Summary

In some environments, the [kubernetes agent fails to deploy flows (and also on other tasks involving the kubernetes API server) due to the server closing the connection after some time being idle.

[This has been observed when running the kubernetes cluster within the Azure Kubernetes Service](https://prefect-community.slack.com/archives/C014Z8DPDSR/p1631240218210500) (but possibly others too). The firewall / loadbalancer closes idle connections after 4 minutes. The urllib3 connection pool used by the kubernetes client library then tries to connect to the server which results in a ReadTimeoutError. Thus the flow deployment is stuck in the submitted state and is only rescued by the Lazarus agent after 15 minutes.

This PR introduces keep-alive messages so that the connection will be kept open for a longer time.
It is disabled by default to avoid problems on existing configurations which are running fine.

If the `cloud.agent.kubernetes_keep_alive` is set to `true` the connection pool will use keep-alive messages.

Since neither the kubernetes nor the urllib3 offer keep-alive functionality out of the box, the solution is a little hacky by setting the flags directly within the socket options.

Moreover the `KubernetesAgent` constructor was slightly refactored to leverage the `prefect.utilities.kubernetes.get_kubernetes_client` function for  `batch_client` and `core_client` in order to make the creation more consistent across the project.

## Changes

- Adds `keep-alive` functionality for kubernetes clients
- Use existing factory for client creation in `KubernetesAgent`

## Importance

- Flow deployments are failing
- ReadTimeoutErrors are occurring
- Communication with Kubernetes API is unstable 

## Checklist

This PR:

### DIFF
--- a/changes/pr5066.yaml
+++ b/changes/pr5066.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Introduced keep-alive connections for kubernetes client API connections - [#5066](https://github.com/PrefectHQ/prefect/pull/5066)"
+
+contributor:
+  - "[Jonas Miederer](https://github.com/jonasmiederer)"
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -126,6 +126,7 @@ class KubernetesAgent(Agent):
             os.getenv("DELETE_FINISHED_JOBS", "True") == "True"
         )
 
+        from prefect.utilities.kubernetes import get_kubernetes_client
         from kubernetes import client, config
 
         try:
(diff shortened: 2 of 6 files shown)


jonasmiederer:
Hi @madkinsz, thanks a lot for your feedback!

I thought having a long-running connection which is kept-alive and ready to use would be better than waiting for the timeout and recreating the connection pool or even recreating the whole client object every few minutes on the timeout.

The urllib3 library lists that approach in their [docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connection.html#:~:text=For%20example%2C%20if%20you%20wish%20to%20enable%20TCP%20Keep%20Alive%20in%20addition%20to%20the%20defaults%2C%20you%20might%20pass%3A) and similar problem cases use that fix ([here](https://carsonip.me/posts/azure-tcp-idle-timeout-tcp-keepalive-python/), [here](https://stackoverflow.com/a/35278688/2948849), [here](https://www.finbourne.com/blog/the-mysterious-hanging-client-tcp-keep-alives), [here](https://github.com/Azure/AKS/issues/1052#issuecomment-501424152) or [here](https://github.com/apache/airflow/pull/11406))

But clearing up the connection and then do a retry would definitely work as well, so if you think that approach is better then I will modify the PR.

madkinsz (approved):

madkinsz:
Thanks!

Issue #4675, opened 2021-06-18
Task RunNamespacedJob doesn't output all logs from the pod


mashun4ek:
## Description

[RunNamespacedJob](https://github.com/PrefectHQ/prefect/blob/6e9170134b59ed5b66afa752bb275e7e44330ed4/src/prefect/tasks/kubernetes/job.py#L580) outputs Prefect logs from CloudFlowRunner and CloudTaskRunner, and logs related to job being created and completed, but it doesn't show the logs from the pod itself, although this task uses [ReadNamespacedPodLogs](https://github.com/PrefectHQ/prefect/blob/6e9170134b59ed5b66afa752bb275e7e44330ed4/src/prefect/tasks/kubernetes/job.py#L753) task.

## Expected Behavior

Output all logs from pod created by RunNamespacedJob task.

## Reproduction

Run the following flow, compare logs from the pod `prefect-job-XXXX` and logs in UI (the line `test logs` is present in pod logs, but not in UI).
```
from prefect import Flow, task
from prefect.tasks.kubernetes.job import RunNamespacedJob
from prefect.run_configs import KubernetesRun
from prefect.storage import GitHub

body = {
    "apiVersion": "batch/v1",
    "kind": "Job",
    "metadata": {"name": "masha"},
    "spec": {
        "template": {
            "spec": {
                "containers": [
                    {
                        "name": "test",
                        "image": "busybox",
                        "command": [
                            "echo",
                            "test logs",
                        ]
                    }
                ],
                "restartPolicy": "Never",
            },
        },
        "backoffLimit": 0,
    }
}

execute_job = RunNamespacedJob(body=body, namespace="mariia-testing", log_level="debug", delete_job_after_completion=False, kubernetes_api_key_secret=None)

with Flow("Test K8s Pod Logs", run_config=KubernetesRun()) as flow:
    execute_job(log_level="debug")

if __name__ == "__main__":
    flow.register(project_name="", labels=[""])
```

## Environment

"prefect_backend": "cloud",
"prefect_version": "0.14.22",
"python_version": "3.8.5"

kiksekage:
I am also missing logs in `RunNamespacedJob` tasks. `prefect diagnostics` output:
```
{
  "config_overrides": {
    "backend": true,
    "server": {
      "database": {
        "host": true,
        "password": true,
        "username": true
      },
      "ui": {
        "apollo_url": true,
        "host_port": true,
        "port": true
      }
    }
  },
  "env_vars": [
[...]

PR #5074, opened 2021-10-21
Fix RunNamespacedJob Logging Issue

tyun08:
## Summary
The current `RunNamespacedJob` logging function on UI is broken as reported in issue: #4675 .

As described from [user @arodus's analysis](https://github.com/PrefectHQ/prefect/issues/4675#issuecomment-869378484), the issue is because of the missing context for the `ReadNamespacedPodLogs` task running from the Flow. 

I removed the `with ThreadPoolExecutor() as pool:` method from it and tested in our project, the logging function is working as expected now:

image

## Importance
Blocking users from using `RunNamespacedJob` to properly log the k8s pods output. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/kubernetes/job.py
+++ b/src/prefect/tasks/kubernetes/job.py
@@ -724,56 +724,56 @@ class RunNamespacedJob(Task):
 
         pod_log_streams = {}
 
-        with ThreadPoolExecutor() as pool:
-            completed = False
-            while not completed:
-                job = api_client_job.read_namespaced_job_status(
-                    name=job_name, namespace=namespace
+        completed = False
+        while not completed:
+            job = api_client_job.read_namespaced_job_status(
+                name=job_name, namespace=namespace
+            )
+
+            if log_level is not None:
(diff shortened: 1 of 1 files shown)


madkinsz:
Hi @tyun08 ! Thanks for contributing. I hadn't looked at this because it's not passing CI, do you need some help?

The reason the context is missing (per analysis in #4675) is that the context is not shared across threads. Hence, removing the thread pool will resolve the issue. I'm not sure what the original intention with the thread pool was here, it looks like it was being used to stream logs from multiple pods at once. I'm not sure we should lose that functionality. If there are multiple pods, I believe this implementation will only collect logs from one pod at a time (until pod completion).

tyun08:
@madkinsz totally makes sense to keep rather than remove. I'll close out my PR as it's not really a fix of the issue. The issue needs to revisited still though. One thing I don't understand is, why you need a new thread to watch logs instead of just running it on the same thread as of k8s job run (which just sent request to k8s api). To me, it worked locally well without creating a new thread for it.

PR #5085, opened 2021-10-26
Add dbt cloud tasks

AlessandroLollo:
## Summary
This PR adds a new task to trigger a dbt Cloud job run.

## Importance
This new task enable users to orchestrate dbt Cloud job runs from within Prefect.

## Checklist
This PR:


src/prefect/tasks/dbt/dbt_cloud_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        - wait_interval: the number of seconds to wait between API calls
+
+    Returns:
+        - The job run result, namely the "data" key in the API response
+
+    Raises:
+        - prefect.engine.signals.FAIL: if "finished_at" is not None and the result status != 10
+    """
+    attempts = 1
+    while not max_attempts or attempts <= max_attempts:
```
madkinsz, line 71: At first, I read this as retry attempts but this is actually the maximum amount of time you'll wait for the job to be finished. I'd clarify that this function will fail immediately if any of the API requests _fails_. Perhaps instead of `max_attempts` you should have `max_wait_time` which is more meaningful to the user instead of considering the combination of both `max_attempts` and `wait_interval`.

Note this would reflect into your task-level settings as well.

test-requirements.txt
```diff
@@ -3,3 +3,4 @@ testfixtures >= 6.10.3
 pytest-env >= 0.6.0
 pytest-xdist >= 2.0
 flaky >= 3.0
+requests-mock >= 1.9.3
```
madkinsz, line 6: Generally, I'm quite hesitant to add testing requirements for contributions to the task library. I'm not sure this gives us much more than a simple mock, but I'm willing to leave it in.

src/prefect/tasks/dbt/dbt_cloud_utils.py
madkinsz: I think this function also ought to be renamed to reflect that it will wait for job completion.

madkinsz (changes requested):
A few small comments on `get_job_run` here. Nice work!

test-requirements.txt
madkinsz: Interestingly, another user wrote a task library PR and is using `responses` for mocking. https://github.com/getsentry/responses

At a glance, I think that project looks more interesting although it's pre-1.0. Curious if you have any thoughts.

madkinsz (approved):
Thanks for working through it with us!

Issue #4957, opened 2021-09-09
v0.15.5 breaks `FlowView` api


schmidt-jake:
## Description
A regression in prefect v0.15.5 causes results returned by `prefect.backend.flow_run.FlowView.from_id` to be erroneous. The resulting `view.flow` object's `parameters` function returns an empty set even though parameters are part of the flow. This also seems to happen for the `tasks` and `edges` attributes.

## Expected Behavior
When `prefect.backend.flow_run.FlowView.from_id` is called with a valid flow id, a flow view should be returned with an attribute called `flow` which has the tasks, edges, parameters, etc of the flow.

## Reproduction
```python
import prefect as pf
from prefect.backend.flow_run import FlowView

with pf.Flow("jake_test") as flow:
    x = pf.Parameter("x")
    x * x

if __name__ == "__main__":
    flow_id = flow.register(project_name="recursion-mlops-poc")
    flow_view = FlowView.from_id(flow_id)
    assert len(flow_view.flow.parameters()) > 0
```
With `prefect==0.15.5`, this code fails at the `assert`. Switch to `prefect==0.15.4` and it'll work.

Curiously, with v0.15.5, `flow_view.serialized_flow` contains the missing parameter:
```python
{
    "name": "jake_test",
    "type": "prefect.core.flow.Flow",
    "storage": {
        "path": None,
        "type": "Local",
        "flows": {
            "jake_test": "/Users/jacob.schmidt/.prefect/flows/jake-test/2021-09-09t19-56-53-341697-00-00"
        },
        "secrets": [],
        "directory": "/Users/jacob.schmidt/.prefect/flows",
        "__version__": "0.15.5",
        "stored_as_script": False
    },
    "schedule": None,
    "parameters": [
        {
            "name": "x",
            "slug": "x",
            "tags": [],
            "type": "prefect.core.parameter.Parameter",
            "default": None,
            "outputs": "typing.Any",
            "required": True,
            "__version__": "0.15.5"
        }
    ],
    "run_config": {
        "env": None,
        "type": "UniversalRun",
        "labels": [
            "JSCHMIDT-MBP.local"
        ],
        "__version__": "0.15.5"
    },
    "__version__": "0.15.5",
    "environment": None,
    "reference_tasks": []
}
```

## Environment
`python 3.7.11`
`prefect==0.15.5` / `prefect==0.15.4`

`prefect diagnostics`:
```json
{
  "config_overrides": {
    "cloud": {
      "auth_token": true
[...]

PR #5106, opened 2021-11-03
Fix load of serialized flows with bad edge task schemas

madkinsz:
Changes in #4930 resulted in dropping tasks/edges from the serialized flow stored in the backend. This caused the `FlowView` to break in some cases (https://github.com/PrefectHQ/prefect/issues/4957). The serialized flow was reconstructed in the backend in https://github.com/PrefectHQ/server/pull/306 but somehow the duplicate edges have bad data in the `upstream_task` and `downstream_task` fields, just containing a `slug` string and not a `{"slug": slug}` dict.

This attempts to fix this by updating task schemas that are just strings within edge schemas to be in a slug dictionary format. This may be better resolved in the backend.

### DIFF
--- a/src/prefect/serialization/edge.py
+++ b/src/prefect/serialization/edge.py
@@ -1,4 +1,4 @@
-from marshmallow import fields
+from marshmallow import fields, pre_load
 
 import prefect
 from prefect.serialization.task import TaskSchema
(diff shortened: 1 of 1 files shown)


madkinsz:
Should be resolved by https://github.com/PrefectHQ/prefect/pull/5116 and https://github.com/PrefectHQ/server/pull/309

PR #5117, opened 2021-11-05
Move artifacts functions to `prefect.backend.artifacts`

madkinsz:
## Summary

With the introduction of `prefect.backend` in `0.15.0`, there was an intention to move the existing `artifacts` API there. This matches the location of other backend-only specific features like the kv-store.

## Changes

- Adds warnings to old artifact functions
- Moves artifact functions to new module
- Adds `_artifact` suffix where needed to artifact create/update functions

## Importance

The old artifacts location can be removed in #5113 

## Checklist

This PR:

### DIFF
--- a/src/prefect/artifacts.py
+++ b/src/prefect/artifacts.py
@@ -1,40 +1,13 @@
-import time
+import warnings
 from typing import Optional
 
-from prefect import context, Client
-from prefect.exceptions import ClientError
-
-
-def _running_with_backend() -> bool:
-    """
-    Determine if running in context of a backend. This is always true when running
-    using the `CloudTaskRunner`.
-
-    Returns:
-        - bool: if `_running_with_backend` is set in context
-    """
-    return bool(context.get("running_with_backend"))
-
-
-def _create_task_run_artifact(kind: str, data: dict) -> str:
-    client = Client()
-    task_run_id = context.get("task_run_id")
-    # XXX: there's a race condition in the cloud backend for mapped tasks where
-    # the task run lookup might fail temporarily. This should last a few
-    # seconds max, for now we retry a few times.
-    retries = 5
-    while True:
-        try:
-            return client.create_task_run_artifact(
-                task_run_id=task_run_id, kind=kind, data=data
-            )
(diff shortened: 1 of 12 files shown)


src/prefect/backend/artifacts.py
```diff
@@ -0,0 +1,117 @@
...
+from prefect.exceptions import ClientError
+
+
+def _running_with_backend() -> bool:
+    """
+    Determine if running in context of a backend. This is always true when running
+    using the `CloudTaskRunner`.
+
+    Returns:
+        - bool: if `_running_with_backend` is set in context
```
zangell44, line 14:
```suggestion
        - bool: if `running_with_backend` is set in context
```

zangell44, line 14:
```suggestion
        - bool: if `running_with_backend` is set in context
```

zangell44: gh suggested twice 🤔

zangell44 (approved):
Tiny comment, lgtm

zangell44 (approved):

src/prefect/artifacts.py
```diff
@@ -47,9 +24,11 @@ def create_link(link: str) -> Optional[str]:
     Returns:
         - str: the task run artifact ID
     """
-    if not _running_with_backend():
-        return None
-    return _create_task_run_artifact("link", {"link": link})
+    warnings.warn(
+        "`prefect.artifacts.create_link` has been moved to `prefect.backend.create_link_artifact`. "
+        "Please update your imports. The old function will be removed in 1.0.0."
```
anticorrelator, line 29: should we use the phrasing "This function will be removed" vs. "The old function"?

anticorrelator (approved):
This LGTM, I had a minor thought about specific wording

src/prefect/artifacts.py
madkinsz: How about
```suggestion
        "Please update your imports. This import path will be removed in 1.0.0."
```

anticorrelator (approved):

Issue #5046, opened 2021-10-11
LocalDaskExecutor fails for > python 3.7 on AWS Lambda


jacques-:
## Description

Not sure if this is a bug or should be a feature request - but basically the `LocalDaskExecutor` does not work in AWS Lambda if python version is > 3.7.

We get an error msg like the following:
```
[Errno 38] Function not implemented: OSError

    sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 38] Function not implemented
```

This is due to `multiprocessing.pool` relying on shared memory (`/dev/shm`) which is not available in Lambda. See https://forums.aws.amazon.com/thread.jspa?threadID=219962 for ref.

## Expected Behavior

Expected the `LocalDaskExecutor` work in python 3.8/3.9 like is does in 3.7

## Reproduction

Reproduction is fairly trivial if you run any flow using `LocalDaskExecutor` in Lambda with a 3.8 python runtime.

## Work around

Looked at the comments here: https://github.com/PrefectHQ/prefect/blob/a6988f64e0538a5e35270004b39658a757d9d01e/src/prefect/executors/dask.py#L642 and based on that we changed https://github.com/PrefectHQ/prefect/blob/a6988f64e0538a5e35270004b39658a757d9d01e/src/prefect/executors/dask.py#L540 from:

```
                    from multiprocessing.pool import ThreadPool

                    self._pool = ThreadPool(num_workers)
```

to:

```
                    from concurrent.futures import ThreadPoolExecutor

                    self._pool = ThreadPoolExecutor(num_workers)
                    self._pool.close = self._pool.shutdown
                    self._pool.join = self._pool.shutdown
```

This seems to work, but no idea if this is the proper way to do it of if it's going to subtly break things elsewhere?

madkinsz:
Hi! We'll need to investigate if we can switch to the `futures` implementation. Thanks for bringing this to our attention!

jacques-:
Cool, we've been testing this against a fair number of slightly complex flows for the last few hours in our dev environment and no errors jumping out yet, so that is at least some kind of datapoint.

andywaugh:
Hey, small update but just an FYI that we've been using the change detailed above in our prod env for a couple weeks now and it seems to work without issue. Are you open to us cleaning it up a bit and submitting a PR?

madkinsz:
Thanks @andywaugh — I'd be happy to review a PR.

jacques-:
[...]

PR #5128, opened 2021-11-10
Issue 5046 futures implementation for threading

jacques-:
## Summary

This is a PR that implements a futures based implementation for threading `LocalDaskExectutor` as discussed in #5046

## Changes

Swap out `multiprocessing.pool.ThreadPool` for `concurrent.futures.ThreadPoolExecutor`

## Importance

In python>=3.8 ThreadPool uses a /dev/shm ramdisk that isn't available in AWS Lambda, this fixes that.

## Checklist

This PR:

### DIFF
--- a/changes/issue5046.yaml
+++ b/changes/issue5046.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Switched to a futures based implementation for LocalDaskExecutor - [#5046](https://github.com/PrefectHQ/prefect/issues/5046)"
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -462,7 +462,7 @@ class DaskExecutor(Executor):
 
 
 def _multiprocessing_pool_initializer() -> None:
-    """Initialize a process used in a `multiprocssing.Pool`.
+    """Initialize a process used in a `concurrent.futures.ProcessPoolExecutor`.
 
     Ensures the standard atexit handlers are run."""
     import signal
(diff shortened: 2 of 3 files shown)


jacques-:
Looking at what's failing, it seems the biggest issue is that this futures approach doesn't work with python 3.6. I'm happy to work on a version that uses the existing multiprocessing for 3.6, and only uses the new bits for >3.6 if that is something you'd consider?

madkinsz:
I think it may be simplest to implement it in the 3.7+ compatible way and change the base branch to the `1.0.0-rc1`, we can drop 3.6 support in the 1.0 release.

madkinsz:
I wonder if we can take this opportunity to use a `ProcessPoolExecutor` as well to reduce the complexity of two separate pool abstractions.

jacques-:
> I wonder if we can take this opportunity to use a `ProcessPoolExecutor` as well to reduce the complexity of two separate pool abstractions.

Haha, yes just after the PR I thought you might ask that - it does seem the sensible thing to do.

Out of curiosity, do you have a sense of when you guys are looking at a 1.0 release?

jacques-:
Seemed easier to just make a new PR to change the base branch: https://github.com/PrefectHQ/prefect/pull/5134

madkinsz:
> > I wonder if we can take this opportunity to use a `ProcessPoolExecutor` as well to reduce the complexity of two separate pool abstractions.
> 
> Haha, yes just after the PR I thought you might ask that - it does seem the sensible thing to do.
> 
> Out of curiosity, do you have a sense of when you guys are looking at a 1.0 release?

@jacques- I can't really commit to a timeframe :) the release candidate should be ready this month though!

jacques-:
> > > I wonder if we can take this opportunity to use a `ProcessPoolExecutor` as well to reduce the complexity of two separate pool abstractions.
> > 
> > 
> > Haha, yes just after the PR I thought you might ask that - it does seem the sensible thing to do.
> > Out of curiosity, do you have a sense of when you guys are looking at a 1.0 release?
> 
> @jacques- I can't really commit to a timeframe :) the release candidate should be ready this month though!

oooh, ok nice - will prob need to start looking at what a migration to 1.0 looks like soon then!

PR #5095, opened 2021-11-01
Fix/flow to flow docs

AlessandroLollo:
## Summary

This PR is an attempt to fix issue https://github.com/PrefectHQ/prefect/issues/5092

Replaces StartFlowRun with create_flow_run in https://docs.prefect.io/core/idioms/flow-to-flow.html

## Importance
Keep docs up to date.
This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #5129, opened 2021-11-10
Add `raise_final_state` bool to `wait_for_flow_run`

madkinsz:
## Summary

Artifact creation was added to `create_flow_run` in #5125 but when users map over this task it will create _many_ artifacts which is not particularly helpful. I'd rather encourage users to create artifacts in a subsequent task. We could consider adding this later with a toggle, but I'd rather lean minimal.

`wait_for_flow_run` does not reflect the final state of the child flow run like `StartFlowRun(wait=True)` does. Here, we add this functionality with an opt-in `raise_final_state` bool. This was noted in #5095 and will unblock that work.

## Changes

- Remove artifact creation from `create_flow_run`
- Add `raise_final_state` bool to `wait_for_flow_run`

## Importance

Additional feature parity for new child flow run tasks

## Checklist

This PR:

- ![ ] adds a change file in the `changes/` directory (if appropriate)~ Will add directly in release

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -225,7 +225,10 @@ def get_task_run_result(
 
 @task
 def wait_for_flow_run(
-    flow_run_id: str, stream_states: bool = True, stream_logs: bool = False
+    flow_run_id: str,
+    stream_states: bool = True,
+    stream_logs: bool = False,
+    raise_final_state: bool = False,
 ) -> "FlowRunView":
     """
     Task to wait for a flow run to finish executing, streaming state and log information
@@ -235,6 +238,8 @@ def wait_for_flow_run(
         - stream_states: Stream information about the flow run state changes
         - stream_logs: Stream flow run logs; if `stream_state` is `False` this will be
             ignored
+        - raise_final_state: If set, the state of this task will be set to the final
+            state of the child flow run on completion.
 
     Returns:
         FlowRunView: A view of the flow run after completion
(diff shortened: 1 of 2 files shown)


zangell44 (approved):

kvnkho (approved):

bnaul:
> when users map over this task it will create many artifacts which is not particularly helpful. I'd rather encourage users to create artifacts in a subsequent task

Personally I don't agree, this is exactly our use case and we use the artifacts from `StartFlowRun` extensively to monitor the progress of subflows. For me creating that artifact after the fact would be pointless, as at that point everything has already concluded.

Obviously it's straightforward to just wrap `create_flow_run` in something that also makes the artifact, but given that this functionality was already present in `StartFlowRun` and I doubt anyone was clamoring to remove it for the sake of "minimalism", it feels not very user-friendly to just remove it with no new alternative.

madkinsz:
Unfortunately we didn't have a chance to discuss this in the original PR. I'm not seeking minimalism here, but rather a consistently user friendly experience. There was indeed confusion about artifact creation in the original task and there were complaints of many artifacts being created while mapping. I'd like to allow the creation of, for example, a markdown artifact with all of the child flow runs in it. I agree that for people not spawning a bunch of child flow runs, an artifact is useful. I don't yet know what the default behavior should be for `create_flow_run` though, and am not willing to commit to it yet.

> creating that artifact after the fact would be pointless, as at that point everything has already concluded.

`create_flow_run` returns immediately after the flow run is created. Generally, the flow run shouldn't have concluded yet.

bnaul:
fair enough, i liked the old behavior but if i'm outnumbered then 🤷 

> create_flow_run returns immediately after the flow run is created. Generally, the flow run shouldn't have concluded yet.

we rate limit our flow runs to a few at a time, so by the time the last one is created all but the last couple have in fact finished.

PR #5140, opened 2021-11-11
[1.0.0] Drop client handling of authentication tokens

madkinsz:
## Summary

Removes nearly all handling for and references to authentication tokens. Authentication token code remains for:

- Deleting existing tokens from the user's machine.
- Detecting the existence of tokens on the user's machine.
- Passing through API keys as if they were tokens for containerized flow run jobs, providing rough backwards compatibility for flows running in containers with old versions of Prefect.
- Detecting if a token has been passed as an API key into the client (for a nicer error message).

### Client changes

- `Client(api_token=...)` has been removed.
- All client token loading has been removed and existing tokens will be ignored without warning.
- `Client.login_to_tenant` has been renamed to `switch_tenant`
- `Client.tenant_id = ...` will verify that a UUID has been passed.
- `Client.request(token=...)` has been refactored to `Client.request(api_key=...)` (this applies to all request methods on the client).
- `Client.active_tenant_id` has been removed.
- `Client.get_auth_token()` has been removed.
- `Client.logout_from_tenant()` has been removed.
- `Client.get_cloud_url` no longer accepts the `as_user` argument.
- `Client.get_default_tenant_slug` no longer accepts the `as_user` argument.

### CLI changes

- `prefect auth login` no longer accepts authentication tokens.
- `prefect auth purge-tokens` has been added to delete the Prefect-managed tokens directory.
- `prefect auth status` will warn about lingering authentication tokens and advise on removal.

### Agent changes

- `Agent._verify_token` has been removed.
- Agents will no longer check for authentication at the `prefect.cloud.agent.auth_token` config key.
- `PREFECT__CLOUD__AUTH_TOKEN` is no longer passed through to non-container flow run jobs.
- `PREFECT__CLOUD__AUTH_TOKEN` is set to the value of `PREFECT__CLOUD__API_KEY` for containerized flow run jobs which will allow containers running old versions of Prefect to work with 1.0.0 agents with the caveat that the API key must be used with the default tenant.

## Importance

Authentication tokens have been replaced and deprecated since 0.15.0.


src/prefect/cli/auth.py
```diff
@@ -477,17 +478,32 @@ def status():
...
+            "The authentication tokens settings path still exists. These have been "
+            "removed in favor of API keys. We recommend purging old tokens with "
+            "`prefect auth purge-tokens`",
             fg="yellow",
         )
 
-    if not client._api_token and not client.api_key:
-        click.secho("You are not logged in!", fg="yellow")
+    if config.cloud.get("auth_token"):
+        if os.environ.get("PREFECT__CLOUD__AUTH_TOKEN"):
```
zangell44, line 490: hmm will this always warn when we pass `PREFECT__CLOUD__AUTH_TOKEN` to flow run containers?

https://github.com/PrefectHQ/prefect/pull/5140/files#diff-3803ec5e942824e9426b9fd46beb5f7cbd4b1d626fbe1bd803f2bdc06f68df2fR580

zangell44: ahhh nvm okay i guess this is just the auth cli

src/prefect/utilities/executors.py
```diff
@@ -116,7 +116,6 @@ def subprocess_heartbeat(heartbeat_cmd: List[str], logger: Logger) -> Iterator[N
             to_environment_variables(
                 prefect.context.config,
                 include={
-                    "cloud.auth_token",
```
zangell44, line 119: do we need this for pre 1.0 containers?

src/prefect/client/client.py
```diff
@@ -139,60 +128,20 @@ def __init__(
...
         )
 
+        # mypy struggles with this attribute type if not created here
+        self._tenant_id: Optional[str] = None
+
         # Load the tenant id
-        self._tenant_id: Optional[str] = (
-            tenant_id
+        # This assignment is validated by the `Client.tenant_id` setter
+        self.tenant_id: Optional[str] = (
```
zangell44, line 136: how do we infer a tenant_id from an api key again?
does the user have to specify a tenant?

zangell44: answering my own question again nvm I see the logic in `_get_auth_tenant` now

zangell44 (approved):
lgtm

My only comment is it feels weird to be implicitly switching using `tenant_id` but explicitly switching using `tenant_slug`. Would it be worth changing the interface a bit to allow explicitly switching by `id` rather than the caller having to know about internals of `Client`?

zangell44 (approved):

PR #5158, opened 2021-11-19
Docs: add Concepts page for Artifacts API

tpdorsey:
## Summary
Adds a new Orchestration > Concepts > Artifacts page providing usage notes for the Artifacts API.

## Changes
Adds new page: docs/orchestration/concepts/artifacts.md
Modifies TOC to include new page: docs/.vuepress/config.js

## Importance
Adds documentation of how artifacts appear in Prefect UI and examples for adding link and markdown artifacts to task definitions. 

## Checklist

This PR:


docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,88 @@
...
+
+After the task runs, navigate to the UI’s Artifacts tab to see the output.
+
+## Creating Markdown Artifacts
+
+To create link artifacts, just import `create_markdown` from `prefect.artifacts`, then add a call to `create_markdown()` to any task orchestrated through a Prefect API. 
+
+Pass `create_markdown()` a string that will be rendered as an artifact. The string can contain any [Github-flavored Markdown](https://github.github.com/gfm/) markup including images, links, and tables. 
+
+Note that any images referenced in your markdown must be linked by absolute URL and served by an external web server. Linking to local files or by relative URL is not supported.
```
anna-geller, line 65: instead of: "by absolute URL and served by an external web server."
I would phrase it something like: "by absolute URL of a publicly available image"
The reason: when I use a URL of an S3 object which is a publicly available image, then it's technically served from object storage rather than a web server, and this would also provide the info that it must be a publicly available image rather than e.g. served from a webserver in a company's VPN.

This is just a minor comment - you can keep the current formulation

docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,88 @@
+# Artifacts <Badge text="Beta"/>
+
+The [Artifacts API](/api/latest/backend/artifacts.html) enables you to publish data from task runs that is rendered natively in the Prefect UI, both Prefect Cloud and Prefect Server. 
+
+Using the Artifacts API, you can easily publish information directly to the Prefect UI. These published artifacts are linked to specific task runs and flow runs, and artifacts can render more sophisticated information than you'd typically log, such as reports, tables, charts, images, and links to external data.
+
+Currently, the Artifacts API enables you to render the following artifact types:
+
+- Links
+- Markdown strings
```
anna-geller, line 10: I would say just Markdown without the "strings"  - technically, both the link and markdown are a string data type in Python

anna-geller (approved):
left just minor comments. Great work and thanks for adding it to the docs so quickly!

docs/orchestration/concepts/artifacts.md
tpdorsey: Good call

docs/orchestration/concepts/artifacts.md
tpdorsey: Good call. Changing this instance. In other places I want to maintain the "string containing markdown" specificity because, technically, users can pass any string content, (including properly escaped HTML). In the specific case of markdown we do render the styling automatically.

docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,88 @@
...
+
+::: tip Artifacts render individually
+Note that each artifact you create in a task renders as an individual artifact in the Prefect UI. In other words, each call to `create_link()` or `create_markdown()` creates its own, separate artifact.
+
+Within a task, you may use these commands as many times as necessary, but they do not operate in the same manner as a `print()` command where you might string together multiple calls to add additional items to a report. 
+
+As a best practice, such as when using `create_markdown()` to create artifacts like reports or summaries, compile your message string separately, then pass to `create_markdown()` to create the full artifact.
+:::
+
+For more background on the design goals for the Artifacts API, see the [Introducing: The Artifacts API](https://medium.com/the-prefect-blog/introducing-the-artifacts-api-b9e5972db043) blog post and the [Great Expectations task](/api/latest/tasks/great_expectations.html).
```
madkinsz, line 24: Do we want to create a copy of this blog on the website instead of medium?

docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,88 @@
...
+@task
+def send_artifact():
+    create_link("https://www.prefect.io/")
+```
+
+After the task runs, navigate to the UI’s Artifacts tab to see the output.
+
+## Creating Markdown Artifacts
+
+To create link artifacts, just import `create_markdown` from `prefect.artifacts`, then add a call to `create_markdown()` to any task orchestrated through a Prefect API.
```
madkinsz, line 61: These imports are moving in 1.0, see #5117

docs/orchestration/concepts/artifacts.md
tpdorsey: I will check with Alyssa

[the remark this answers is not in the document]
tpdorsey: @madkinsz it was on the blog already, so I updated the link here.

docs/orchestration/concepts/artifacts.md
tpdorsey: What do you think is the best way to handle that? We can push this for now, and I can stage an update for 1.0 that includes the revisions to the imports?

madkinsz: They're available at the new path on 0.15.8 and will continue to be available at the old path until 1.0. I'd recommend using the new paths in the examples and noting that if you're on a version before 0.15.8 that they are available at a different name? A later change is alright too, but users will get warnings if they use your examples on our latest version.

docs/orchestration/concepts/artifacts.md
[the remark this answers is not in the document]
tpdorsey: Covered by an updated push. Resolving.

docs/orchestration/concepts/artifacts.md
tpdorsey: Updated in a new push. Let me know if you think this covers the imports adequately.

docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,96 @@
...
+- Markdown
+
+Link artifacts render a clickable hyperlink as an artifact.
+
+Markdown artifacts render strings that can include [Github-flavored Markdown](https://github.github.com/gfm/) markup for formatting. 
+
+::: tip Artifacts render individually
+Note that each artifact you create in a task renders as an individual artifact in the Prefect UI. In other words, each call to `create_link_artifact()` or `create_markdown_artifact()` creates its own, separate artifact.
+
+Within a task, you may use these commands as many times as necessary, but they do not operate in the same manner as a `print()` command where you might string together multiple calls to add additional items to a report.
```
madkinsz, line 19: Do you think it's worth explaining the ID return value and the `update_markdown_artifact` method? It's a bit tricky to use.

docs/orchestration/concepts/artifacts.md
```diff
@@ -0,0 +1,96 @@
...
+    create_markdown_artifact("# Heading\nText with [link]("https://www.prefect.io/").")
+```
+
+`create_markdown_artifact()` accepts any valid Python string and variable interpolation including, for example, f-strings.
+
+```python
+@task
+def send_artifact(heading, link_text, link):
+    create_markdown_artifact(f"# {heading}\nLink to [{link_text}]({link}).")
+```
```
madkinsz, line 90: This is a little confusing. It accepts a string. The formatting here happens before it is sent to the function. I'm worried this statement will lead to confusion where people think `create_markdown_artifact("{foo} {bar}", foo=1, bar=2)` will work.

tpdorsey: I can change this to a simpler example.

docs/orchestration/concepts/artifacts.md
tpdorsey: Yes, can you send me an example to ponder, and I'll let you know if I have questions.

tpdorsey: Added examples based on the code you provided. Also added a section covering `delete_artifact`.

madkinsz:
Moved to #5180 for clean merging.

PR #5006, opened 2021-09-29
Fix urllib3 deprecation warning

bphillips-exos:
## Summary

```
../../usr/local/lib/python3.9/dist-packages/prefect/storage/bitbucket.py:17
  /usr/local/lib/python3.9/dist-packages/prefect/storage/bitbucket.py:17: DeprecationWarning: 
Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. 
Use 'allowed_methods' instead
    BITBUCKET_CLOUD_RETRY = Retry(
```

PR #5167, opened 2021-11-23
Use allowed_methods in Retry as method_whitelist is depreciated 

KipCrossing:
## Summary

This is a simple PR to stop a `DeprecationWarning`

```
Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
``` 

## Changes
Use `allowed_methods` instead of `method_whitelist` in the Retry object

## Importance
This will ensure compatibility with future versions 

## Checklist

These points are all pretty much n/a

This PR:

### DIFF
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -18,7 +18,7 @@ BITBUCKET_CLOUD_RETRY = Retry(
     total=10,
     backoff_factor=0.5,
     status_forcelist=[429],
-    method_whitelist=["HEAD", "GET", "OPTIONS"],
+    allowed_methods=["HEAD", "GET", "OPTIONS"],
 )
 
 if TYPE_CHECKING:
--- a/src/prefect/storage/bitbucket.py
+++ b/src/prefect/storage/bitbucket.py
@@ -6,7 +6,7 @@ from urllib.error import HTTPError
 import requests
 from requests.exceptions import HTTPError as RequestsHTTPError
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry
+from urllib3.util.retry import Retry
 
 import prefect
 from prefect.client import Secret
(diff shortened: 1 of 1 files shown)


KipCrossing:
I find it strange that `Retry` is being imported from `requests` like:

```python
from requests.packages.urllib3.util.retry import Retry
```

Instead of directly from `urllib3`, like:

```python
from urllib3.util.retry import Retry
```

Perhaps this was an auto import. Unless there is a special reason for this, I can make this change too. 

---

Re the failed linting:

Here are the versions of `requests` and `urllib3` I tested with:

```
Name: requests
Version: 2.26.0
```

```
Name: urllib3
Version: 1.26.7
```

It looks like the tests/mypy is running with older versions.

madkinsz:
Note that there's an open PR for this at https://github.com/PrefectHQ/prefect/pull/5006, but I'm happy to accept this one instead if you're around.

The static analysis check is using `requests-2.26.0` and `urllib3-1.26.7` (from CI logs). It's pulling the type definition from a typeshed file though

```
src/prefect/storage/bitbucket.py:17: error: Unexpected keyword argument "allowed_methods" for "Retry"
/usr/local/lib/python3.8/site-packages/mypy/typeshed/third_party/2and3/requests/packages/urllib3/util/retry.pyi:24: note: "Retry" defined here
```

Fixing this type check might require moving to a newer version of mypy with separate type shed packages (see #4642, #4638) which I don't want to do. I think a `# type: ignore` with a comment explaining this is fine.

`requests.packages.urllib3` is around for legacy reasons. Feel free to update those imports, that'd be great.

KipCrossing:
@madkinsz ah, that makes sense. I made those changes, so hopefully, it's good to go.

src/prefect/storage/bitbucket.py
```diff
@@ -18,7 +18,8 @@
     total=10,
     backoff_factor=0.5,
     status_forcelist=[429],
-    method_whitelist=["HEAD", "GET", "OPTIONS"],
+    # ignoring new allowed_methods param until mypy update
```
madkinsz, line 21: Just a minor change here. We're quite unlikely to upgrade mypy for this repository.
```suggestion
    # typeshed is out of date with urllib3 and missing `allowed_methods`
```

KipCrossing: Done!

madkinsz (approved):
Thanks!

Issue #4963, opened 2021-09-13
Docker agent with server tasks stuck submitted: host.docker.internal connection error


cbrown1234:
## Description

The docker agent deployed locally on the same machine as the docker server fails to run tasks, they get stuck in the submitted state on the server. From the logs, it seems to relate to an inability to connect to `host.docker.internal`.
 
```
[2021-09-13 15:59:17,700] INFO - agent | Deploying flow run 2d52c82a-e890-4f66-a845-7f5b0664ef0d to execution environment...
[2021-09-13 15:59:18,038] INFO - agent | Completed deployment of flow run 2d52c82a-e890-4f66-a845-7f5b0664ef0d
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/urllib3/connection.py", line 169, in _new_conn
    conn = connection.create_connection(
  File "/usr/local/lib/python3.8/site-packages/urllib3/util/connection.py", line 96, in create_connection
    raise err
  File "/usr/local/lib/python3.8/site-packages/urllib3/util/connection.py", line 86, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 394, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/local/lib/python3.8/site-packages/urllib3/connection.py", line 234, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/local/lib/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/local/lib/python3.8/http/client.py", line 1302, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.8/http/client.py", line 1251, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.8/http/client.py", line 1011, in _send_output
    self.send(msg)
  File "/usr/local/lib/python3.8/http/client.py", line 951, in send
    self.connect()
  File "/usr/local/lib/python3.8/site-packages/urllib3/connection.py", line 200, in connect
[...]

PR #5182, opened 2021-11-30
Allow override of Prefect API url in docker runs and improve inference

madkinsz:
## Summary

The API URL used by the agent can be unconnectable from within a spawned flow run container when using Prefect Server. Currently, we replace `localhost` with `host.docker.internal` to fix this. However, this is brittle in some cases which we've been patching (e.g. #4714, #4809, #2433)

With the change to limit exposure of the API by default in #4821, when `--expose` is not used on server startup, it is not connectable even with `host.docker.internal`. In #5156, I documented adding the `prefect-server` network to the run containers to allow them to connect, but this does not work on Linux. This PR replaces `host.docker.internal` with `apollo` if the `prefect-server` network is detected so that the internal network is used. This allows running an agent and server side-by-side without the `--expose` flag.

Additionally, this allows the API URL to be overridden entirely by users so that, if there is another brittle edge-case, they can just provide the proper URL.

Closes https://github.com/PrefectHQ/prefect/issues/4963

## Changes

- Allows `--env PREFECT__CLOUD__API=...` and `DockerRun(env=...)` to override the API the flow run will connect to once spawned.
- Updates inference to "just work" when the `prefect-server` network is provided.

## Importance

Fixes major connection issues with `DockerAgent` spawned containers while using Prefect Server

## Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,16 @@ Released on November 30, 2021.
 - Clarify `ECSRun` documentation, especially the ambiguities in setting IAM roles - [#5110](https://github.com/PrefectHQ/prefect/issues/5110)
 - Fix deprecated usage of `marshmallow.fields.Dict` in RRule schedules - [#4540](https://github.com/PrefectHQ/prefect/issues/4540), [#4903](https://github.com/PrefectHQ/prefect/pull/4903)
 
+### Fixes
+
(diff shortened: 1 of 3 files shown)


madkinsz:
I will add a changes entry directly into #5181

madkinsz:
Rebased onto `release-0.15.0` so I get the sodaspark fix and can add the changelog entry now.

zangell44 (approved):
lgtm

Issue #5189, opened 2021-12-02
k8 agent doesn't cleanup after upgrading to 0.15.10. 


thomasfrederikhoeck:
## Description
After upgrading to 0.15.10 and enable `kubernetes_keep_alive` I start getting the following error message and the completed jobs don't get cleaned up. I might suspect it doesn't do the management of the jobs in general but I haven't confirmed this. I suspect it is related to #5066 as it started when upgrading from 0.15.7 to 0.15.10

```
[2021-12-02 15:45:36,640] ERROR - agent | Error while managing existing k8s jobs
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 406, in heartbeat
    self.manage_jobs()
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 354, in manage_jobs
    and not self.client.get_flow_run_state(
  File "/usr/local/lib/python3.8/site-packages/prefect/client/client.py", line 1661, in get_flow_run_state
    return prefect.engine.state.State.deserialize(flow_run.serialized_state)
AttributeError: 'NoneType' object has no attribute 'serialized_state'
ERROR:agent:Error while managing existing k8s jobs
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 406, in heartbeat
    self.manage_jobs()
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 354, in manage_jobs
    and not self.client.get_flow_run_state(
  File "/usr/local/lib/python3.8/site-packages/prefect/client/client.py", line 1661, in get_flow_run_state
    return prefect.engine.state.State.deserialize(flow_run.serialized_state)
AttributeError: 'NoneType' object has no attribute 'serialized_state'
[2021-12-02 15:46:38,459] ERROR - agent | Error while managing existing k8s jobs
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 406, in heartbeat
    self.manage_jobs()
  File "/usr/local/lib/python3.8/site-packages/prefect/agent/kubernetes/agent.py", line 354, in manage_jobs
    and not self.client.get_flow_run_state(
  File "/usr/local/lib/python3.8/site-packages/prefect/client/client.py", line 1661, in get_flow_run_state
[...]

PR #5190, opened 2021-12-02
Fix error while managing K8s jobs when the flow run is deleted

madkinsz:
## Summary

Solves issues where the kubernetes agent will fail to manage jobs if one of the jobs references a flow run id that no longer exists.

## Changes

- We _always_ check the flow run state when we manage K8s jobs, previously this would only be checked if there was a failed pod.
- We skip managing jobs where the flow run id no longer exists
- `Client.get_flow_run-state` raises an `ObjectNotFoundError` if the flow run does not exist instead of an `AttributeError`

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/5189

## Checklist

This PR:


src/prefect/client/client.py
```diff
@@ -1658,6 +1659,9 @@ def get_flow_run_state(self, flow_run_id: str) -> "prefect.engine.state.State":
 
         flow_run = self.graphql(query).data.flow_run_by_pk
 
+        if not flow_run:
```
anticorrelator, line 1662: just a nit: do we intend this to fail if the flow_run is explicitly `None`?

madkinsz: I'm intentionally covering any form empty data here as the return line will fail with an `AttributeError`

tests/agent/test_k8s_agent.py
```diff
@@ -811,7 +811,56 @@ def test_k8s_agent_manage_jobs_pass(monkeypatch, cloud_api):
     agent.heartbeat()
 
 
+def test_k8s_agent_manage_jobs_handles_missing_flow_runs(
+    monkeypatch, cloud_api, caplog
+):
+    Client = MagicMock()
+    Client().get_flow_run_state.side_effect = ObjectNotFoundError()
```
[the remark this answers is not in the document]
madkinsz: 👍 done

anticorrelator (approved):
lgtm

PR #5191, opened 2021-12-02
Use `allowed_methods` instead of `methods_whitelist` which is deprecated by urllib3

madkinsz:
## Summary

Follow-up to implement changes as in https://github.com/PrefectHQ/prefect/pull/5167 to the `Client.request` method.

Users will see this warning always now that urllib3 is upgraded, see https://github.com/PrefectHQ/prefect/pull/5179

## Changes

Uses `allowed_methods` when specifying retries in `Client.request`

## Importance

Deprecation warnings from upstreams are concerning to users

## Checklist

This PR:


zangell44 (dismissed):

madkinsz:
I do not believe this merits a changelog entry. Interestingly, this was entirely untested so I added a simple test.

zangell44 (approved):

PR #5194, opened 2021-12-03
beta1 is deprecated in Kubernetes v1.17+

anna-geller:
beta1 is [no longer needed](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and throws a warning: 
```
Warning: rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
```

Also, defaulting to a higher Python version than 3.6 would be nice.


src/prefect/agent/kubernetes/agent.py
```diff
@@ -836,7 +836,7 @@ def generate_deployment_yaml(
 
         version = prefect.__version__.split("+")
         image_version = (
-            "latest" if len(version) > 1 or latest else (version[0] + "-python3.6")
+            "latest" if len(version) > 1 or latest else (version[0] + "-python3.9")
```
madkinsz, line 839: This is a relatively serious breaking change — we can bump to 3.8 in 1.0, but we can't just bump the default python version here.

```suggestion
            "latest" if len(version) > 1 or latest else (version[0] + "-python3.6")
```

anna-geller:
@madkinsz thanks a lot, I thought it's only relevant for the K8s template. It was probably the reason why tests were failing?

madkinsz:
@anna-geller the failing tests are asserting that the result handler warning is raised. Since you changed its message, it's the tests fail.

Yeah it just changes the template, but for users suddenly their flows could be running on 3.9 instead of 3.6 which is a breaking change.

tpdorsey (approved):
String changes look fine to me.

anna-geller:
> 

@madkinsz do you want me to revert back to "No result handler was specified" or adjust some test? Or are we good to merge this PR?

madkinsz:
> > 
> 
> @madkinsz do you want me to revert back to "No result handler was specified" or adjust some test? Or are we good to merge this PR?

@anna-geller you can do either. You can see the failing tests in CircleCI (or just search for the old message in the repo).

anna-geller:
> > > 
> > 
> > 
> > @madkinsz do you want me to revert back to "No result handler was specified" or adjust some test? Or are we good to merge this PR?
> 
> @anna-geller you can do either. You can see the failing tests in CircleCI (or just search for the old message in the repo).

@madkinsz I removed the change, but the CI tests fail on black. What am I doing wrong? How can it be that modifying one YAML file causes formatting issues in totally unrelated files? I need help here.

madkinsz:
@anna-geller a new black release is out and must have a small change. Upgrade black locally and format the given file and we'll just fix it for everyone.

PR #5208, opened 2021-12-07
Correct import for KV Store

tpdorsey:
## Summary
Corrects the import statement in examples for KV Store

## Changes
Change import statements to `from prefect.backend.kv_store import ...`.

https://docs.prefect.io/orchestration/concepts/kv_store.html

Confirmed that correct imports are shown in Artifacts examples.

## Importance
Previous examples were incorrect and would fail.

## Checklist

This PR:

~~- [ ] adds new tests (if appropriate)~~
~~- [ ] adds a change file in the `changes/` directory (if appropriate)~~
~~- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)~~

### DIFF
--- a/docs/orchestration/concepts/kv_store.md
+++ b/docs/orchestration/concepts/kv_store.md
@@ -6,7 +6,7 @@ Key Value Store is a managed metadata database for Prefect Cloud.
 
 The number of key value pairs allowed is limited by license, starting with 10 pairs on the Free tier. Values are limited to 10 KB in size.
 
-Key value pairs can be configured via the Prefect CLI, Python library, API, and UI.
+Key value pairs can be configured via the Prefect CLI, Python library, API, and UI. See the [KV Store API](/api/latest/backend/kv_store.html) documentation for further usage details.
 
 ## UI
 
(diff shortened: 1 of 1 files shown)


madkinsz:
The current imports are fine though

```
❯ ipython
from Python 3.8.12 (default, Oct 12 2021, 06:23:56) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.29.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from prefect.backend import get_key_value, set_key_value
```

tpdorsey:
Interesting. Looks like you can import from either prefect.backend or prefect.backend.kv_store? I haven't tested yet, but both look like they import. I would think the later is more intuitive and matches the API docs. 
https://docs.prefect.io/api/latest/backend/kv_store.html#functions

madkinsz:
We expose the full API of backend functions at `prefect.backend`. The submodule use is optional. I don't have a strong preference for which we use in documentation, but generally I'd choose the simpler import path i.e. we don't tell users to import `Flow` from `prefect.flows`, we tell them to import it from `prefect`.

tpdorsey:
Imports work as originally documented. There is no compelling reason to update the docs at this point. In future, it would be better to update all locations in the docs where shortcut imports work.

PR #5213, opened 2021-12-08
Improve multiprocess based timeout handler

madkinsz:
A user was running into an indeterminate issue with the multiprocess based handler. While investigating, I added some useful logs and discovered what `queue.empty()` is not reliable and the stdlib documentation recommends not using it at all. Additionally, joining a process before we've retrieved data from a queue can cause deadlocks. Previously, we even sent a SIGTERM to the process before retrieving the data.

## Changes
- Adds pickling/unpickling debug logs
- Uses `queue.get(block=True, timeout=timeout)` instead of `process.join(timeout)`

### DIFF
--- a/src/prefect/utilities/executors.py
+++ b/src/prefect/utilities/executors.py
@@ -411,6 +411,7 @@ def run_with_multiprocess_timeout(
             raise result
         return result
     except Empty:
+        logger.debug(f"{name}: No result returned within the timeout period!")
         raise TaskTimeoutSignal(f"Execution timed out for {name}.")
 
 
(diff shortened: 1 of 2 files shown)


abrookins (approved):
This looks right to me. Nice! Sounds like in addition to `empty()` being unreliable, simply joining a process before we've pulled the work from the queue could deadlock, and your PR should resolve that issue. https://docs.python.org/3/library/multiprocessing.html#all-start-methods ("Joining processes that use queues").

madkinsz:
> 



> This looks right to me. Nice! Sounds like in addition to `empty()` being unreliable, simply joining a process before we've pulled the work from the queue could deadlock, and your PR should resolve that issue. https://docs.python.org/3/library/multiprocessing.html#all-start-methods ("Joining processes that use queues").

Ooh good find. This work was indeed instigated by a report of deadlock while pulling the result from a queue.

jakekaplan (approved):
LGTM! Moving the timeout from `join` to the `get` makes a lot of sense

PR #5220, opened 2021-12-10
Fixing Serialized Flow Examples in Docs

kvnkho:
## Summary

The examples we have in [this docs](https://docs.prefect.io/orchestration/tutorial/next-steps.html#examples) page don't work as reported in [this discussion](https://github.com/PrefectHQ/prefect/discussions/5215). 

Trying to run these vie the UI raises:

`Failed to load and execute Flow's environment: ValueError('Flow is not contained in this Storage')`

There are three Flows there and all of their storage look like the following. The important part is that `flows` is empty. When the Flow Run is triggered, there is no Flow to search for in the storage.
```
"storage": {"access_token_secret": null, "path": "examples/parameters.py", "ref": "d44b72a950ebda9f7bc6a9712fc71e2e9c680d25", "flows": {}, "repo": "PrefectHQ/prefect", "base_url": null, "secrets": [], "__version__": "0.15.6+74.gd44b72a95", "type": "GitHub"}}
```

Normally, this isn't an issue because registering a flow serializes it. However, this is never registered and the JSON is just output.

## Changes

We serialize the Flow and add the appropriate check to ensure that there is a Flow in storage.

## Importance
Users need this working for example Flows in the UI

## Checklist

This PR:

### DIFF
--- a/changes/pr5220.yaml
+++ b/changes/pr5220.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fixing serialized example in docs - [#5220](https://github.com/PrefectHQ/prefect/pull/5220)"
--- a/docs/core/examples/overview.md
+++ b/docs/core/examples/overview.md
@@ -25,8 +25,12 @@ $ prefect register --json https://docs.prefect.io/examples.json \
     --name "Example: Mapping"
 ```
 
-These can then be run using any agent with a ``prefect-examples`` label. For
-example, to start a local agent for running the examples:
+These can then be run using any agent with a ``prefect-examples`` label. Before starting the agent, 
+make sure the Github extra is installed so the agent can pull the flows by doing:
+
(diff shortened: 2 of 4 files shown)


anna-geller:
@kvnkho Nice work! The only thing I would change would be modifying [this page](https://docs.prefect.io/core/examples/overview.html#running-with-prefect-cloud-or-server): 
- add a hint that even though a local agent is used here, those examples should work on any agent (I think this was the purpose of this example)
- ideally, before starting the agent, add: `pip install "prefect[github]"` because if the GitHub extra is not installed before the local agent is started, the flow will fail due to missing package

kvnkho:
Will add in a bit

tpdorsey (approved):

madkinsz:
Thanks!

PR #5223, opened 2021-12-11
Fix bigquery insert task

TilakMaddy:
## Summary

Added support to stream insert a *large* number of rows to the BigQuery Table 

## Changes

It now streams chunks of rows instead of all at once. Before this we run into an error when inserting like 100,000 records. Now, we don't.

## Importance

insert_rows_json - Legacy Streaming API

https://cloud.google.com/bigquery/quotas#streaming_inserts
Maximum allowed rows per request - 50,000 rows

    - A maximum of 500 rows is recommended.
    - Batching can increase performance and throughput to a point, but at the cost of per-request latency.
    - Too few rows per request and the overhead of each request can make ingestion inefficient.
    - Too many rows per request and the throughput can drop. Experiment with representative data (schema and data sizes) to determine the ideal batch size for your data.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/gcp/bigquery.py
+++ b/src/prefect/tasks/gcp/bigquery.py
@@ -267,20 +267,41 @@ class BigQueryStreamingInsert(Task):
 
         # get table reference
         table_ref = client.dataset(dataset_id).table(table)
+    
+        '''
+        insert_rows_json - Legacy Streaming API
 
-        # stream data in
-        response = client.insert_rows_json(table=table_ref, json_rows=records, **kwargs)
+        https://cloud.google.com/bigquery/quotas#streaming_inserts
+
+        Maximum rows per request - 50,000 rows
+            - A maximum of 500 rows is recommended.
+            - Batching can increase performance and throughput to a point, but at the cost of per-request latency.
+            - Too few rows per request and the overhead of each request can make ingestion inefficient.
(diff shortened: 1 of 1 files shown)


src/prefect/tasks/gcp/bigquery.py
```diff
@@ -267,20 +268,41 @@ def run(
...
+        Maximum rows per request - 50,000 rows
+            - A maximum of 500 rows is recommended.
+            - Batching can increase performance and throughput to a point, but at the cost of per-request latency.
+            - Too few rows per request and the overhead of each request can make ingestion inefficient.
+            - Too many rows per request and the throughput can drop. Experiment with representative data (schema and data sizes) to determine the ideal batch size for your data.
+        '''
+
+        def batch(iterable, size):
+            it = iter(iterable)
+            while item := list(itertools.islice(it, size)):
```
madkinsz, line 286: We still support Python 3.6/7 so you can't use the walrus operator :)

madkinsz:
This needs to support Python 3.7 and needs tests. Closing as stale, feel free to reopen if you resume work.

davidschrooten:
StreamingInsert is useless for large amount of data without batching.

TilakMaddy:
Absolutely ! I agree :)

PR #5232, opened 2021-12-14
Multiple local agents with the same label

anna-geller:
## Summary
Adding a section to clarify that local agents are not intended to be used as a queue.

### DIFF
--- a/changes/pr5232.yaml
+++ b/changes/pr5232.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Adding a section to the local agent documentation - [#5232](https://github.com/PrefectHQ/prefect/pull/5232)"
--- a/docs/orchestration/agents/local.md
+++ b/docs/orchestration/agents/local.md
@@ -121,11 +121,39 @@ and using supervisor.
 
 ## Multiple local agents with the same label
 
-In general, assigning a unique label to each agent is recommended. 
-While Prefect Cloud has a mechanism to ensure that each flow run gets executed 
-only once, race conditions may happen if you have multiple agents with the same 
-label. At the time of writing, Prefect has no notion of a task queue that would 
-allow round-robin flow runs based on available resources on a local agent. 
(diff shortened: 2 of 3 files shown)


tpdorsey:
This valuable. I have a few suggestions for filling this out with some additional information, so just providing a rough edit here:

## Local agents and labels

We recommend assigning a unique label to each agent, particularly when running multiple local agents. While Prefect Cloud has a mechanism to ensure that each flow run gets executed 
only once, race conditions may occur if you have multiple agents with the same 
label. 

[Can you give an example here with an actual label, and link up further in the page to the discussion of [Labels](#labels) for details. That section does not show a "working example."]

Currently, Prefect has no notion of a task queue that would 
allow round-robin flow runs based on available resources on a local agent. 
If you need such functionality, you would need to ensure on your own that you 
assign flows to agents that have enough resources by leveraging labels. 
To help with scale, we recommend using one of the other agents. [I'd expand this by specifying which agents and linking to them]

anna-geller:
> This valuable. I have a few suggestions for filling this out with some additional information, so just providing a rough edit here:
> 
> ## Local agents and labels
> We recommend assigning a unique label to each agent, particularly when running multiple local agents. While Prefect Cloud has a mechanism to ensure that each flow run gets executed only once, race conditions may occur if you have multiple agents with the same label.
> 
> [Can you give an example here with an actual label, and link up further in the page to the discussion of [Labels](#labels) for details. That section does not show a "working example."]
> 
> Currently, Prefect has no notion of a task queue that would allow round-robin flow runs based on available resources on a local agent. If you need such functionality, you would need to ensure on your own that you assign flows to agents that have enough resources by leveraging labels. To help with scale, we recommend using one of the other agents. [I'd expand this by specifying which agents and linking to them]

I took all your recommendations into account and added an example to make it clear. Thanks a lot!

tpdorsey (approved):
Looks good. Thank you for adding these clarifications.

PR #5237, opened 2021-12-15
Add proxies argument to slack_notifier

### LINKED ISSUE #5236: Add proxies to slack_notifier
## Current behavior
`prefect.utilities.notifications.slack_notifier` sends requests to the Slack webhook using `requests.post` without allowing  `proxies={"http": "0.123.456.789"}` to be passed, which is problematic when Internet access must be done through a proxy.

 It is possible to go around this by setting HTTP_PROXY / HTTPS_PROXY environment variables, but this is not ideal / possible when running on private infrastructure which by default does not access the Internet and should not have these variables set.

## Proposed behavior
Add an optional `proxies` argument to `prefect.utilities.notifications.slack_notifier` which gets passed to `requests.post`.

## Example
To use Slack notifications from Prefect running on private infrastructure.

VincentAntoine:
## Summary

Add `proxies` argument to `slack_notifier`.

## Changes

`prefect.utilities.notifications.slack_notifier` now passes a new optional `proxies` argument to `requests.port` to send requests to the Slack webhook through proxies.

## Importance

- Allows to send Slack notifications through proxies when required.
- Resolve #5236 

## Checklist

This PR:

### DIFF
--- a/changes/pr5237.yaml
+++ b/changes/pr5237.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Allow passing of proxies argument to slack_notifier - [#5237](https://github.com/PrefectHQ/prefect/pull/5237)"
+
+contributor:
+  - "[Vincent Chéry](https://github.com/VincentAntoine)"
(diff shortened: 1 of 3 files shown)


madkinsz (approved):
Thanks!

PR #5253, opened 2021-12-20
Add build of release candidate docker image

madkinsz:
Adds build and publish of docker images for the `1.0.0rc` branch (not yet created). A copy of the publishing of development artifacts for `master`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,6 +462,49 @@ jobs:
                 command: |
                   docker login --username  $DOCKER_HUB_USER --password $DOCKER_HUB_PW
                   docker push prefecthq/prefect:core
+  build_release_candidate_docker_image:
+    docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -547,6 +590,13 @@ workflows:
             branches:
               only: master
 
+      - build_release_candidate_docker_image:
+          python_version: '3.7'
```
cicdw, line 594: any reason to restrict this?

madkinsz: No intention here. This is just the test image for our round of internal tests; we can add other versions now or when we add the CI job to publish to PyPi.

abrookins (approved):
This looks good. We can keep it at 3.7 and just make sure our instructions for setting up the test environment specify to use that version.

PR #5258, opened 2021-12-21
Schedule_delay parameter for delaying flow execution

### LINKED ISSUE #5171: Adding a timedelta to the `scheduled_start_time` in the `create_flow_run` task
## Current behavior

If the user creates a child flow run from a parent flow, but wants this child flow run to be scheduled further away in the future, currently it wouldn't work, since the value of the `scheduled_start_time` would be frozen at registration. This has been requested as part of this thread in the community Slack: https://prefect-community.slack.com/archives/CL09KU1K7/p1637778182289200 The user wanted that after the first task has completed, a child flow run of another flow would be triggered but with a 2 minutes delay in between. 

### Workaround

There are two options:

1. The recommended workaround atm is to add a task in between that sleeps for 2 minutes. 
2. Adding a task that determines the timedelta, as shown below:

```python
from prefect.tasks.prefect import create_flow_run, wait_for_flow_run
import pendulum

@task
def in_minutes(minutes: int):
    return pendulum.now().add(minutes=minutes)

with Flow("parent-flow") as flow:
    a = upstream_task()
    child_run_id = create_flow_run(project_name="hello world",
                                   flow_name="hello-flow",
                                   labels=["test-label"],
                                   scheduled_start_time=in_minutes(2, upstream_tasks=[a]),
                                   upstream_tasks=[a])
    wait_for_flow_run(child_run_id)
```

## Proposed behavior

Adding a `timedelta` to the `create_flow_run` task.

```python
from prefect.tasks.prefect import create_flow_run, wait_for_flow_run
import pendulum

with Flow("parent-flow") as flow:
    a = upstream_task()
    child_run_id = create_flow_run(project_name="hello world",
                                   flow_name="hello-flow",
                                   labels=["test-label"],
                                   scheduled_start_time=pendulum.now(),
				   schedule_delay_seconds=120,
                                   upstream_tasks=[a])
    wait_for_flow_run(child_run_id)
```

exotikh3:
## Summary

Functionality to delay execution of flow 

## Importance

Closes #5171 

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -1458,6 +1458,7 @@ class Client:
         run_config: RunConfig = None,
         labels: List[str] = None,
         scheduled_start_time: datetime.datetime = None,
+        schedule_delay: datetime.timedelta = None,
         idempotency_key: str = None,
         run_name: str = None,
         version_group_id: str = None,
@@ -1476,6 +1477,8 @@ class Client:
             - labels (List[str], optional): a list of labels to apply to the flow run
             - scheduled_start_time (datetime, optional): the time to schedule the execution
                 for; if not provided, defaults to now
+            - schedule_delay (timedelta, optional): period which will be waited for before executing
+                task; if not provided, defaults as 0
             - idempotency_key (str, optional): an idempotency key; if provided, this run will
                 be cached for 24 hours. Any subsequent attempts to create a run with the same
                 idempotency key will return the ID of the originally created run (no new run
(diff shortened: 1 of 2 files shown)


anna-geller:
Closes #5171

madkinsz:
Hi! Thanks for contributing. The intent here was actually for `scheduled_start_time` to take a `Union[datetime, timedelta]` instead of just a `datetime`. If a `timedelta` is given we will set the `scheduled_start_time` to `pendulum.now("utc") + the timedelta.

This will allow users to use kwarg as intended when defining the input to their task statically.

exotikh3:
> Hi! Thanks for contributing. The intent here was actually for `scheduled_start_time` to take a `Union[datetime, timedelta]` instead of just a `datetime`. If a `timedelta` is given we will set the `scheduled_start_time` to `pendulum.now("utc") + the timedelta.
> 
> This will allow users to use kwarg as intended when defining the input to their task statically.

ok, but just to clarify: is the place for setting this up was chosen correctly? I believe that there is a space to do so when the task is being pick up from db. if yes I will modify this to take a single 'scheduled_start_time' argument.

madkinsz:
Nope this shouldn't require changes to the `Client`. This logic should be happening in the task itself.

i.e. at https://github.com/PrefectHQ/prefect/pull/5258/files#diff-33e69166fa725dfcba50547f47708109ab2462a79e6f7cd376a5a9f7887ceaffL147 you should be passing a new scheduled start time calculated as described above.

exotikh3:
@madkinsz so should I move the functionality from the client to the "create_flow_run"?

madkinsz:
> @madkinsz so should I move the functionality from the client to the "create_flow_run"?

Yep!

madkinsz:
Closing in favor of #5303

PR #5271, opened 2021-12-22
Add release candidate support to `Docker` storage

madkinsz:
`Docker` storage needs special handling to detect release candidate versioning and use the proper base image tag


cicdw (approved):
good catch

PR #5134, opened 2021-11-11
switched to futures based LocalDaskExecutor

jacques-:
## Summary

This is a PR that implements a futures based implementation `LocalDaskExectutor` as discussed in #5046

## Changes

Swap out `multiprocessing.pool.Pool` for `concurrent.futures.Executor`

## Importance

In python>=3.8 ThreadPool uses a /dev/shm ramdisk that isn't available in AWS Lambda, this fixes that.

## Checklist

This PR:

- [X] adds new tests (if appropriate)
- [X] adds a change file in the `changes/` directory (if appropriate)
- [X] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #5279, opened 2021-12-27
Issue 5046 futures implementation

jacques-:
## Summary

This is a PR that implements a futures based implementation `LocalDaskExectutor` as discussed in #5046.

I'm reopening this PR to continue working on it after https://github.com/PrefectHQ/prefect/pull/5134 was closed.

## Changes

Swap out `multiprocessing.pool.Pool` for `concurrent.futures.Executor`

## Importance

In python>=3.8 ThreadPool uses a /dev/shm ramdisk that isn't available in AWS Lambda, this fixes that.

## Checklist

This PR:


madkinsz:
Sorry this got auto-closed when I deleted the base branch. Looks like there's the same issue as before https://github.com/PrefectHQ/prefect/pull/5134#issuecomment-967213703

jacques-:
Ok, so with a recent dask (2021.9.1) I get the following: 

```ImportError while loading conftest '/root/project/tests/conftest.py'.
tests/conftest.py:11: in <module>
    from distributed import Client
/usr/local/lib/python3.7/site-packages/distributed/__init__.py:3: in <module>
    from .actor import Actor, ActorFuture
/usr/local/lib/python3.7/site-packages/distributed/actor.py:6: in <module>
    from .client import Future, default_client
/usr/local/lib/python3.7/site-packages/distributed/client.py:31: in <module>
    from dask.compatibility import apply
E   ImportError: cannot import name 'apply' from 'dask.compatibility' (/usr/local/lib/python3.7/site-packages/dask/compatibility.py)

Exited with code exit status 4
```

Going to see when that was dropped from dask and try that version.

jacques-:
Ok @madkinsz - bumping dask to 2021.06.0 fixes this, seems all good now?

jacques-:
Ok, got it working using dask `2021.06.0`, which fixes everything but creates a new error on windows:

```
================================== FAILURES ===================================
_ TestFlowRunMethod.test_flow_dot_run_updates_the_scheduled_start_time_of_each_scheduled_run _
[gw0] win32 -- Python 3.8.10 c:\hostedtoolcache\windows\python\3.8.10\x64\python.exe

self = <tests.core.test_flow.TestFlowRunMethod object at 0x000002C03D9A8B80>

    def test_flow_dot_run_updates_the_scheduled_start_time_of_each_scheduled_run(self):
    
        start_times = [pendulum.now().add(seconds=i * 0.2) for i in range(1, 4)]
        REPORTED_START_TIMES = []
    
        @task
        def record_start_time():
            REPORTED_START_TIMES.append(prefect.context.scheduled_start_time)
    
        f = Flow(
            name="test",
            tasks=[record_start_time],
            schedule=prefect.schedules.Schedule(
                clocks=[prefect.schedules.clocks.DatesClock(dates=start_times)]
            ),
        )
        f.run()
>       assert REPORTED_START_TIMES == start_times
E       AssertionError: assert [DateTime(202...e('Etc/GMT'))] == [DateTime(202...e('Etc/GMT'))]
E         Right contains one more item: DateTime(2022, 1, 3, 15, 15, 38, 718143, tzinfo=Timezone('Etc/GMT'))
E         Full diff:
E           [
E            DateTime(2022, 1, 3, 15, 15, 38, 318143, tzinfo=Timezone('Etc/GMT')),
E            DateTime(2022, 1, 3, 15, 15, 38, 518143, tzinfo=Timezone('Etc/GMT')),
E         -  DateTime(2022, 1, 3, 15, 15, 38, 718143, tzinfo=Timezone('Etc/GMT')),
E           ]
```

Any ideas @madkinsz?

madkinsz:
@jacques- don't worry about `test_flow_dot_run_updates_the_scheduled_start_time_of_each_scheduled_run` — that test has some weird behaviors.

src/prefect/executors/dask.py
```diff
@@ -595,8 +593,7 @@ def _posttask(self, key, value, dsk, state, id):  # type: ignore
                     if exiting_early:
                         self._interrupt_pool()
                     else:
-                        self._pool.close()
-                    self._pool.join()
+                        self._pool.shutdown(wait=False)
```
madkinsz, line 596: Shouldn't this be `wait=True` since it's not an early exit?

jacques-: I think that's right yes, fixed now.

madkinsz:
I believe that the `ThreadPoolExecutor` does not have a `terminate` function so the `_interrupt_pool` method will fail. Can you check if there's a test covering this — I'm surprised there was not a failure here. I think we may want to change it to

`pool.shutdown(wait=False, cancel_futures=True)`

src/prefect/executors/dask.py
```diff
@@ -512,11 +511,11 @@ def _interrupt_pool(self) -> None:
         if self._pool is None:
             return
 
-        # Terminate the pool
-        self._pool.terminate()
+        # Shutdown the pool
+        self._pool.shutdown(wait=False)
```
madkinsz, line 515: Let's cancel any unstarted futures as well
```suggestion
        self._pool.shutdown(wait=False, cancel_futures=True)
```

madkinsz: @jacques- my apologies, looks likes this is a 3.9+ only feature. I'm okay with excluding it entirely instead of toggling by version.

jacques-:
`cancel_futures` is only for > 3.8, should I add a version check and do if we are running on 3.9+?

This test looks like it should be covering it by the sounds of things, but I'm guessing those timers will need some experimentation to see if they are still correct etc to be absolutely sure. This test good enough or needs more testing? https://github.com/PrefectHQ/prefect/blob/d31622c981bd5666d4e07dff331537766f58d067/tests/executors/test_executors.py#L164

madkinsz:
We can just not use that flag at all. Sorry about that!

jacques-:
@madkinsz is there anything else that can/needs to be done here?

madkinsz (approved):
Thank you! I appreciate it.

PR #5292, opened 2021-12-31
Robust communication on server or cloud instances

BitTheByte:
# Summary
This allow more network errors to happen on unstable  networks or when the graphql service is temporarily unavailable

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -727,7 +727,7 @@ class Client:
             headers.update(self._attached_headers)
 
         session = requests.Session()
-        retry_total = 12
+        retry_total = 6
         retries = requests.packages.urllib3.util.retry.Retry(
             total=retry_total,
             backoff_factor=1,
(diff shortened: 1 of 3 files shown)


src/prefect/client/client.py
```diff
@@ -727,7 +727,7 @@ def _request(
             headers.update(self._attached_headers)
 
         session = requests.Session()
-        retry_total = 6 if prefect.config.backend == "cloud" else 1
+        retry_total = 12
```
madkinsz, line 730: Hi! I think I'm okay with removing the conditional check here but is there a particular reason you bumped this to 12? With the exponential backoff, the last sleep would be ~34 minutes. Can you move back to 6?

BitTheByte: Sure

madkinsz (approved):

madkinsz:
Thanks!

BitTheByte:
@madkinsz can we get this merged today? I'm having a CI based on the master branch :)

madkinsz (approved):

Issue #5230, opened 2021-12-14
Improve dask-scheduler memory management of Prefect `State` objects  


xuevin:
## Description
The dask-scheduler uses `dask.sizeof` to manage memory for scaling and work distribution. Tasks which return a large payload can cause scheduling to stall, or may cause workers to exit. This can be better managed by giving the scheduler awareness of the payload sizes.

## Expected Behavior
`State.__sizeof__()`  should be defined to return the number of bytes stored in the object.

## Reproduction
Define a  task which returns a large numpy array. The dask-scheduler will report the managed-memory at 48 B. 
```
@task
def memory_task():
    x = np.random.randint(9, size=2*10**6)
    return x
```
Proposed resolution: Currently State objects do not have `__sizeof__` defined, which causes scheduler to think the payload is 48 B.  Ideally, this `__sizeof__` function should be implemented to capture the size of the result produced /stored in `State.result`.

madkinsz:
This seems reasonable, I'd review a PR that adds a basic sizeof implementation.

PR #5304, opened 2022-01-07
State __sizeof__ impl

cjmartian:
Signed-off-by: Connor Martin <connormartin7@gmail.com>

## Summary

Find size of `State` object and it's children as requested in https://github.com/PrefectHQ/prefect/issues/5230

## Changes

Adds a `__sizeof__` implementation to `State`

## Importance

Was requested in https://github.com/PrefectHQ/prefect/issues/5230 as to correctly report the size of state.

## Checklist

This PR:


madkinsz:
Hey @cjmartian — sorry for the delayed review here!

This is a sweet implementation :) but I think collecting the deep size is probably overkill. We should able to just include the size of `State.result` without recursively descending. Is there a reason you felt a recursive implementation was necessary?

cjmartian:
Hey @madkinsz,

Whoops, I just figured the `State` class should get the complete size of the `State` object, instead of just result. Maybe another method that just returns size of the result?

```
def get_result_size(self):
  return sys.getsizeof(self.result)
```

madkinsz:
Generally `__sizeof__` implementations in Python do not retrieve deep sizes. Here, we want to include the size of the `result` so that Dask can optimizing passing states over the network so it'd have to be implemented in the dunder. I'm actually not sure if the `result` size will be sufficient for Dask but it seems like a good first step. Note that we want to include the overhead of the `State` object still so we'll need to use something like `super().__sizeof__() + getsizeof(self.result)`

src/prefect/engine/state.py
```diff
@@ -110,6 +111,9 @@ def __getstate__(self) -> dict:
 
         return data
 
+    def __sizeof__(self) -> int:
+        return super().__sizeof__() + sys.getsizeof(self.restult)
```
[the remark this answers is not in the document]
cjmartian: Oops, just saw this. Thanks for catching it too!

madkinsz:
@cjmartian thanks! Can you add a test case covering this and update the changes entry?

cjmartian:
@madkinsz Let me know about that test, want to make sure it's actually useful.

tests/engine/test_state.py
```diff
@@ -749,3 +752,14 @@ def __init__(self, *args) -> None:
...
+
+@patch("prefect.engine.state.super")
+def test_state_sizeof(mock_super):
+    state = State(result=10)
+    mock_super().__sizeof__.return_value = 55
+    # Don't use sys.getsizeof(state) here because
+    # getsizeof() includes overhead that is hard
+    # to account for.
+    # Instead call __sizeof__ directly
+    assert state.__sizeof__() == 55 + sys.getsizeof(state.result)
```
madkinsz, line 765: This LGTM but I'll have @abrookins take a look too.

Perhaps one additional assertion that `sys.getsizeof(state) >= state.__sizeof__()` so we can be sure it's being applied in practice?

cjmartian: Sure thing, I can add that to the test

madkinsz (approved):

madkinsz:
Thanks!

tests/engine/test_state.py
abrookins: I declare retroactively that this looks good to me. 😄

PR #5314, opened 2022-01-11
Fix response key in airbyte task health check

mcfuhrt:
## Summary
Currently an exception is thrown within the health check function when trying to trigger a connection in Airbyte.

## Changes
The non-existing key from the response used for the health check is replaced by a valid, existing key.

## Importance
Without the change it is not possible to execute an AirbyteTask successfully, because this will always lead to a KeyError: 'db'.

## Checklist
Bildschirmfoto 2022-01-11 um 14 40 34

This PR:

### DIFF
--- a/src/prefect/tasks/airbyte/airbyte.py
+++ b/src/prefect/tasks/airbyte/airbyte.py
@@ -76,7 +76,7 @@ class AirbyteConnectionTask(Task):
         try:
             response = session.get(get_connection_url)
             self.logger.info(response.json())
-            health_status = response.json()["db"]
+            health_status = response.json()["available"]
             if not health_status:
                 raise AirbyteServerNotHealthyException(
                     f"Airbyte Server health status: {health_status}"
--- a/src/prefect/tasks/airbyte/airbyte.py
+++ b/src/prefect/tasks/airbyte/airbyte.py
@@ -76,9 +76,7 @@ class AirbyteConnectionTask(Task):
         try:
             response = session.get(get_connection_url)
             self.logger.info(response.json())
-            key = "db"
-            if "available" in response.json():
-                key = "available"
+            key = "available" if "available" in response.json() else "db"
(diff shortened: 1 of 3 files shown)


desertaxle (changes requested):
Thanks for your contribution! It looks like this was a change that Airbyte made recently (in v0.35.4-alpha according to their [release notes](https://github.com/airbytehq/airbyte/releases)). Since we most likely will have users running older version of Airbyte where the health check contains the `db` key, I propose that we check for the presence of both the `db` flag and the `available` key to account for different Airbyte server versions. I think it will be a quick change, but let me know if you have any questions!

mcfuhrt:
> Thanks for your contribution! It looks like Airbyte recently made a change to their health check endpoint(in v0.35.4-alpha according to their [release notes](https://github.com/airbytehq/airbyte/releases)). Since we most likely will have users running older version of Airbyte where the health check contains the `db` key, I propose that we check for the presence of both the `db` flag and the `available` key to account for different Airbyte server versions. I think it will be a quick change, but let me know if you have any questions!

I have also thought about this and would consider it a good idea if both are offered. I have adjusted the PR accordingly

desertaxle (dismissed):
LGTM!

src/prefect/tasks/airbyte/airbyte.py
```diff
@@ -76,7 +76,10 @@ def _check_health_status(self, session, airbyte_base_url):
         try:
             response = session.get(get_connection_url)
             self.logger.info(response.json())
-            health_status = response.json()["db"]
+            key = "db"
+            if "available" in response.json():
+                key = "available"
```
madkinsz, line 81: Preference for the more "pythonic" variable assignment here

```suggestion
            key = "available" if "available" in response.json() else "db"
```

madkinsz (approved):
Nice!

PR #5316, opened 2022-01-12
Enforce logging limits by size instead of message length

madkinsz:
## Summary

This builds on the work in https://github.com/PrefectHQ/prefect/pull/4902 and logging in Orion to adjust the calculations for logging payload limits.

## Changes

Since we remove an item from the queue to place it in the batch, the pending limit could exceed the batch limit by the size of a single log. When batching logs, we enforce a limit as `MAX_BATCH_LOG_SIZE - MAX_LOG_SIZE` to ensure we never exceed the `MAX_BATCH_LOG_SIZE`. If `MAX_LOG_SIZE` is greater than `MAX_BATCH_LOG_SIZE`, we will batch with `MAX_LOG_SIZE` instead.

Here, we also use `sys.getsizeof(log)` to check sizes instead of `len(log.message)`. This accounts for the overhead of the additional log fields without special calculations. `sys.getsizeof(dict(...))` should always exceed `len(json.dumps(dict(...)))` due to the overhead of a Python dictionary so we should only see a reduction in size on serialization. This means we're overestimating log sizes, which results in more batches but we shouldn't exceed the maximums.

## Importance

Maximums are _enforced_ server side, so the client should _never_ exceed the maximum or the logs will not be sent. If any log batches exceed the maximum, no future logs will be sent because batches are never dropped.

## Checklist

This PR:

### DIFF
--- a/tests/utilities/test_logging.py
+++ b/tests/utilities/test_logging.py
@@ -123,7 +123,7 @@ def test_cloud_handler_emit_warns_and_truncates_long_messages(
 
     with pytest.warns(
         UserWarning,
-        match="Received a log of size 360 bytes, exceeding the limit of 200",
+        match="Received a log of size 3[0-9][0-9] bytes, exceeding the limit of 200",
     ):
         logger.info("h" * 220)
 
--- a/tests/utilities/test_logging.py
+++ b/tests/utilities/test_logging.py
@@ -2,6 +2,7 @@ import contextlib
 import datetime
 import logging
 import time
+import json
 from unittest.mock import MagicMock
 
 import click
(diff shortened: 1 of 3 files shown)


tests/utilities/test_logging.py
```diff
@@ -223,10 +221,13 @@ def test_log_manager_startup_and_shutdown(logger, log_manager):
 
 
 def test_log_manager_batches_logs(logger, log_manager, monkeypatch):
-    monkeypatch.setattr(prefect.utilities.logging, "MAX_BATCH_LOG_LENGTH", 100)
+    monkeypatch.setattr(prefect.utilities.logging, "MAX_BATCH_LOG_SIZE", 1000)
+    monkeypatch.setattr(prefect.utilities.logging, "MAX_LOG_SIZE", 500)
+
     # Fill up log queue with multiple logs exceeding the total batch length
```
dbast, line 227:
```suggestion
    # Fill up log queue with multiple logs exceeding the MAX_BATCH_LOG_SIZE
```

10 * 50 = 500, which imho doesn't anymore exceed the max_batch_log_size

tests/utilities/test_logging.py
```diff
@@ -223,10 +221,13 @@ def test_log_manager_startup_and_shutdown(logger, log_manager):
 
 
 def test_log_manager_batches_logs(logger, log_manager, monkeypatch):
-    monkeypatch.setattr(prefect.utilities.logging, "MAX_BATCH_LOG_LENGTH", 100)
+    monkeypatch.setattr(prefect.utilities.logging, "MAX_BATCH_LOG_SIZE", 1000)
+    monkeypatch.setattr(prefect.utilities.logging, "MAX_LOG_SIZE", 500)
```
dbast, line 225:
```suggestion
    monkeypatch.setattr(prefect.utilities.logging, "MAX_BATCH_LOG_SIZE", 400)
    monkeypatch.setattr(prefect.utilities.logging, "MAX_LOG_SIZE", 100)
```

tests/utilities/test_logging.py
```diff
@@ -238,6 +239,11 @@ def test_log_manager_batches_logs(logger, log_manager, monkeypatch):
     assert messages == [f"{i}" * 50 for i in range(10)]
     assert log_manager.client.write_run_logs.call_count == 5
 
+    for upload in log_manager.client.write_run_logs.call_args_list:
+        log_entries = [log_entry for log_entry in upload[0][0]]
+        payload = json.dumps(log_entries)
+        assert len(payload) <= 1000, payload
```
dbast, line 245:
```suggestion
        assert len(payload) <= 400, payload
```

dbast:
Nice! That is a bit more elegant than https://github.com/PrefectHQ/prefect/pull/4902 and also should cover all the corner cases. I like the usage of sys.getsizeof which overestimates compared to json.dumps.

Just some inline comments regarding the test case.

tests/utilities/test_logging.py
madkinsz: Each log is ~350 bytes — we can be sure this test is working as intended because we assert logs are sent 5x

```
assert log_manager.client.write_run_logs.call_count == 5
```

madkinsz:
@dbast Happy to add you as a contributor on this changelog entry if you're interested, just lmk!

abrookins (approved):
LGTM! 👍

dbast:
@madkinsz Thanks for fixing this and also the new release!

munjalpatel:
@madkinsz this is not working anymore for `PyPy` due to `sys.getsizeof(log)`. I logged issue #5389 for it.

Issue #5299, opened 2022-01-04
Unexpected error when flattened task fails


andywaugh:
## Description
Something is going wrong when a flattened task fails - this can be caused by a signal FAIL or simply an unhandled exception. Specifically, the `len` call on this line https://github.com/PrefectHQ/prefect/blob/17b885bbae4aa01a9859ff28f9d9cd3623dedc1a/src/prefect/utilities/executors.py#L728 appears to be the issue.

This behaviour only seems to happen when `flatten` is used. If a mapped task fails without the use of flatten the flow works as expected.

Apologies if this is raised in error; this seems like it would be core to a lot of flows so I'm surprised it hasn't already been raised if it's a legitimate bug.

## Expected Behavior
I expect the flow to complete with a status of FAILED.

## Reproduction
PoC code:
```
from prefect import Task, Flow, mapped, flatten
from prefect.engine.signals import FAIL


class SuccessTask(Task):
    def run(self, some_input):
        print(some_input)
        return [False, True, False]


class RaiseErrorTask(Task):
    def run(self, do_raise):
        if do_raise:
            raise TypeError("Example error")

        return True


class RaiseFailTask(Task):
    def run(self, do_raise):
        if do_raise:
            raise FAIL("fail...")

        return True


class TestFlow(Flow):
    def __init__(self, **kwargs) -> None:
        super().__init__(name="test_flow", **kwargs)
        with self:
            res = SuccessTask()(some_input=1)
            # Same behaviour seen when unhandled exception is thrown - swap RaiseFailTask for RaiseErrorTask below
            res2 = RaiseFailTask()(do_raise=mapped(res))
            SuccessTask()(some_input=flatten(res2))


TestFlow().run()
```

Output:
```
[2022-01-04 12:43:41+0000] INFO - prefect.FlowRunner | Beginning Flow run for 'test_flow'
[2022-01-04 12:43:41+0000] INFO - prefect.TaskRunner | Task 'SuccessTask': Starting task run...
1
[2022-01-04 12:43:41+0000] INFO - prefect.TaskRunner | Task 'SuccessTask': Finished task run for task with final state: 'Success'
[2022-01-04 12:43:41+0000] INFO - prefect.TaskRunner | Task 'RaiseFailTask': Starting task run...
[2022-01-04 12:43:41+0000] INFO - prefect.TaskRunner | Task 'RaiseFailTask': Finished task run for task with final state: 'Mapped'
[...]

Issue #4566, opened 2021-05-21
flatten should handle SKIP signals


dylanbhughes:
```python
from prefect import task, Flow, flatten
from prefect.engine.signals import SKIP


@task
def return_a_list():
    return [2, 4, 6, 8]


@task
def make_list_of_lists_or_skip(input_item):
    if input_item % 2 == 0:
        raise SKIP("even thing")

    return [(input_item, input_item * 2), (input_item, input_item * 2)]


@task
def do_additional_thing(input_item):
    return input_item[0] + input_item[1]


with Flow("my flow") as flow:
    my_list = return_a_list()
    transformed_list = make_list_of_lists_or_skip.map(my_list)
    result = do_additional_thing.map(flatten(transformed_list))
```

## Current behavior

The flow above fails with:
```
Python 3.8.10 (default, May 19 2021, 11:01:55)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.23.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: flow.run()
[2021-05-21 10:43:13-0400] INFO - prefect.FlowRunner | Beginning Flow run for 'my flow'
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'return_a_list': Starting task run...
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'return_a_list': Finished task run for task with final state: 'Success'
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'make_list_of_lists_or_skip': Starting task run...
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'make_list_of_lists_or_skip': Finished task run for task with final state: 'Mapped'
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'make_list_of_lists_or_skip[0]': Starting task run...
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | SKIP signal raised: SKIP('even thing')
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'make_list_of_lists_or_skip[0]': Finished task run for task with final state: 'Skipped'
[2021-05-21 10:43:13-0400] INFO - prefect.TaskRunner | Task 'make_list_of_lists_or_skip[1]': Starting task run...
[...]

Issue #4455, opened 2021-04-29
Restarting mapped task which uses flatten from prefect cloud throws TypeError("object of type 'NoneType' has no len()")


jeremytee97:
## Description

When restarting a flow from Prefect Cloud, it fails when attempting to restart for my mapped task. I realized it happens when I am using `flatten` - in the `invoke_lambda` method.  I have all of my previous inputs, all cached in s3Result. 
![image](https://user-images.githubusercontent.com/37123816/116499492-d97d3680-a8de-11eb-96ef-e7a6f90f5517.png)

Here are the error logs from the cloud
![image](https://user-images.githubusercontent.com/37123816/116499169-139a0880-a8de-11eb-974d-7b3bd55b27c3.png)

It is weird because if I were to create a new flow run, it would be able to run without any issues. I have posted it on the slack channel and Kevin Kho was able to reproduce it!

link to our slack conversation - https://prefect-community.slack.com/archives/CL09KU1K7/p1619596538209400
## Expected Behavior

Since I have cached all the previous steps in s3Results, it is supposed to be able to rerun from where it failed. 

## Reproduction

```import json
import boto3
from prefect import Flow, flatten, task
from prefect.engine import signals
from prefect.engine.results.s3_result import S3Result
from prefect.executors import LocalExecutor
from prefect.run_configs import LocalRun
from prefect.storage import S3

@task(
    name="get_names",
    target="result/{flow_name}/{flow_run_name}/{task_full_name}",
    checkpoint=True,
    result=S3Result("prefect-flow"),
)
def get_names():
    return ["Peter", "James", "Gee"]

@task(
    name="get_skills",
    target="result/{flow_name}/{flow_run_name}/{task_full_name}",
    checkpoint=True,
    result=S3Result("prefect-flow"),
)
def get_skills(name):
    skills_mapping = {
        "Peter": ["IT", "Accounting"],
        "James": ["Mathematics", "English"],
        "Gee": ["Python", "SQL"],
    }
    return skills_mapping.get(name, [])

@task(
    name="invoke_lambda",
    target="result/{flow_name}/{flow_run_name}/{task_full_name}",
    checkpoint=True,
    result=S3Result("prefect-flow"),
)
def invoke_lambda(skill):
    print(f"invoking lambda with skill: {skill}")
    lambda_client = boto3.client("lambda", region_name="ap-southeast-1")
    response = lambda_client.invoke(
        FunctionName="prefect-test", Payload=json.dumps(skill)
[...]

Issue #4084, opened 2021-02-06
prefect.flatten fails when processing results that contain Exceptions


rousik:
If a mapped task is passed to `prefect.flatten` and any of the mapped tasks fail, this will crash `flatten_mapped_children` method on this line:
https://github.com/PrefectHQ/prefect/blob/8ce99d9752b175eeaf8a0b15ba3d0a80c221b69e/src/prefect/utilities/executors.py#L559

Here's a simplified flow that triggers this problem:
```python
from prefect import task, Flow, flatten


@task
def crash_if_odd(i):
  if i % 2:
    raise KeyError("Fails")
  return i


@task
def summarize(nums):
  return sum(nums)


with Flow("Test flow") as f:
  summarize(flatten(crash_if_odd.map([1,2])))
f.run()
```

And the exception:
```
...
[2021-02-06 16:47:51-0700] INFO - prefect.TaskRunner | Task 'crash_if_odd[0]': Finished task run for task with final state: 'Failed'
[2021-02-06 16:47:51-0700] INFO - prefect.TaskRunner | Task 'crash_if_odd[1]': Starting task run...
[2021-02-06 16:47:51-0700] INFO - prefect.TaskRunner | Task 'crash_if_odd[1]': Finished task run for task with final state: 'Success'
[2021-02-06 16:47:51-0700] ERROR - prefect.FlowRunner | Unexpected error: TypeError("object of type 'KeyError' has no len()")
Traceback (most recent call last):
  File "/home/jaro/data/anaconda3/envs/pudl-dev/lib/python3.8/site-packages/prefect/engine/runner.py", line 48, in inner
    new_state = method(self, state, *args, **kwargs)
  File "/home/jaro/data/anaconda3/envs/pudl-dev/lib/python3.8/site-packages/prefect/engine/flow_runner.py", line 495, in get_flow_run_state
    children = executors.flatten_mapped_children(
  File "/home/jaro/data/anaconda3/envs/pudl-dev/lib/python3.8/site-packages/prefect/utilities/executors.py", line 559, in flatten_mapped_children
    [executor.submit(lambda c: len(c._result.value), c) for c in mapped_children]
  File "/home/jaro/data/anaconda3/envs/pudl-dev/lib/python3.8/site-packages/prefect/utilities/executors.py", line 559, in <listcomp>
    [executor.submit(lambda c: len(c._result.value), c) for c in mapped_children]
[...]

PR #5337, opened 2022-01-20
Add support for flatten() when processing results that contain no flattenable value

gebing:
## Summary

Add support for flatten() when processing results that contain no flattenable value.

## Checklist

This PR:

### DIFF
--- a/changes/issue4084.yaml
+++ b/changes/issue4084.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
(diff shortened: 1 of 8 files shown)


madkinsz:
Hey @gebing, thanks for contributing! Is there an existing issue this is closing? I can follow the implementation here, but I'm not sure I fully understand the test case. Could you share a couple more cases demonstrating how this changes the behavior of `flatten`?

gebing:
it fixed for Issue #4084, #3631, #5299, #4455, #4566, #4277.

madkinsz:
Thanks for linking those! I see they're in the commit message too now :)

I'd appreciate if we added some more test coverage:
- The simplest flow that demonstrates the changed behavior
- Handling of non-signal exceptions

gebing:
i already made a test case for demonstrate the behavior for flow with flatten() to process the results that contain no flattenable value in function `test_flatmap_one_unflattenable_input()` of `tests/core/test_task_map.py`, which would throw an exception before. Can you give me more specific information for what kind of test cases do you need me to add?

bphillips-exos:
+1 on getting this merged, I ran into this issue while trying to use a mapped task as an upstream task.

cicdw (approved):
LGTM

madkinsz:
I've updated this to have test coverage for:
- Tasks that raise exceptions
- Tasks that return non-iterable values that are not exceptions

I've added a warning for the second case since this would previously have resulted in a `TypeError` .

Issue #5204, opened 2021-12-07
Lack of a Descriptive Error When Logged In with Another API Key


kvnkho:
## Current behavior

If a user is current logged in with an API Key, and they try to do:

`prefect auth login --key <MY_KEY>`

They get an error

`Unauthorized. Invalid Prefect Cloud API key.`.

They need to do `prefect auth logout` before the new API key can be used.

## Proposed behavior

The error could suggest that maybe they need to use `prefect auth logout`.

meehawk:
@kvnkho would love to take this up. Seems manageable (while I brainstorm over the #4386)

kvnkho:
Actually sorry @meehawk, I tried to replicate this myself just now but could not. Let me know if you can.

You can try:
1. Creating an account with Prefect Cloud at cloud.prefect.io , we have a very generous free tier so you shouldn't have to put a credit card
2. Go to service accounts under Team in the navigation bar
3. Create a service account
4. Create an API Key
5. Login with `prefect auth login --key API_KEY`

After that, I tried the following to replicate but couldn't:
1. Logging in with a new key.
2. Logging in with a new key in a new service account while deleting the old one.

And these didn't seem to replicate it. I tried on 0.15.7 and 0.15.10 and saw nothing. Will ping the person to reported to chime in and add details.

meehawk:
Ahh @kvnkho I'll try to follow the steps you've put 🤓

kvnkho:
@meehawk ,

I talked to the person who reported it @TilakMaddy and I was able to replicate it. You need two different accounts.

1. Log-in with API key in account 1
2. Create API key with account 2
3. Try to log-in with account 2 API key

I don't know if you're familiar with `click` the Python library to create CLIs but the place to start would be [here](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/auth.py#L133-L154). You can see the error message.

I think you can check if the error caused has any difference. If it does, you can catch it and have a different error message. If it doesn't I get just add details to the current error that says "Consider trying prefect auth logout first" or something like that.

madkinsz:
I think that it probably ought to just log them in with their new key. There shouldn't need to be a logout step here.

Is this reproducible on the `1.0.0-rc1` branch? It might have something to do with the token support.

meehawk:
@kvnkho I was just going to ping you saying that I had difficulties replicating this. Totally missed your new comment on the issue

TilakMaddy:
[...]

PR #5355, opened 2022-01-21
Explicitly unset the tenant id during `prefect auth login`

madkinsz:
This fixes a bug where the cached tenant id could be loaded when a new key is used which would result in unauthorized errors if the new key was not compatible with the old tenant


madkinsz:
A regression test for this is a pain as the entire `Client` is mocked in these CLI tests. I think fixing this bug is reasonable for now with manual QA and we can explore more robust testing patterns in 1.0 where the auth pattern is simplified.

madkinsz:
Closes #5204

kvnkho (approved):

kvnkho:
I did a manual QA on this and this fixes the issue!

PR #5362, opened 2022-01-24
Raise an error when Airbyte sync job fails to mark the task run as Failed

anna-geller:
## Summary
Airbyte task currently checks for the job status here:
https://github.com/PrefectHQ/prefect/blob/01705d38b9819a8e38d2696ee56999bc7a3faf89/src/prefect/tasks/airbyte/airbyte.py#L271

However, when the job status is failed, it only logs the error rather than raising an exception:
https://github.com/PrefectHQ/prefect/blob/01705d38b9819a8e38d2696ee56999bc7a3faf89/src/prefect/tasks/airbyte/airbyte.py#L280

## Changes
Raising an exception to mark the task run as Failed.

## Importance
Fixes a bug flagged by a community user in [this Slack thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1643021910053100).

## Checklist

This PR:

- [] adds new tests (if appropriate)
- [] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

### DIFF
--- a/changes/pr5362.yml
+++ b/changes/pr5362.yml
@@ -1,0 +1,2 @@
+fix:
+  - "Raise an error when Airbyte sync job fails to mark the task run as Failed - [#5362](https://github.com/PrefectHQ/prefect/pull/5362)"
--- a/src/prefect/tasks/airbyte/airbyte.py
+++ b/src/prefect/tasks/airbyte/airbyte.py
@@ -21,6 +21,10 @@ class JobNotFoundException(Exception):
     pass
 
 
+class AirbyteSyncJobFailed(Exception):
+    pass
+
+
 class AirbyteConnectionTask(Task):
     """
     Task for triggering Airbyte Connections, where "A connection is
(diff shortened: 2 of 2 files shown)


madkinsz:
Let's add a changelog entry for this, it's a significant change in behavior.

anna-geller:
> Let's add a changelog entry for this, it's a significant change in behavior.

I added `changes/pr5362.yml` already - or do you mean something else?

madkinsz (approved):

anna-geller:
> Don't we want the same behaviour for cancelled jobs? Otherwise we're stuck in an infinite loop.

I see. If you would be willing to contribute a PR for it, we would take time to review and QA this behavior together. Is this common that someone cancels Airbyte sync jobs? I imagined it as a data replication mechanism you wouldn't want to cancel to avoid some inconsistent database state, but I'm not that familiar with Airbyte

b4stien:
> Is this common that someone cancels Airbyte sync jobs?

There is a button to cancel one in Airbyte's interface, so it can happen. Not that frequently I'd agree, but disproportionately during test/setup to have a fast feedback loop.

> I imagined it as a data replication mechanism you wouldn't want to cancel to avoid some inconsistent database state

Depending on the mode you chose the next synchronization will overwrite previous data, so there won't be any inconsistency.

I'd see if I have time to draft a PR, thanks.

PR #5366, opened 2022-01-25
Allow ECS task definition role ARNs to override ECSAgent defaults.

youcandanch:
## Summary
Allow the ECSAgent to retrieve task role and execution role ARNs from an explicitly passed in ECS task definition.

## Changes
- During ECSAgent's deploy_flow function, if the user has provided a task_definition_arn for their RunConfig, retrieve that task definition from ECS using the Boto client.
- Pass the task definition for the flow into `ECSAgent.get_run_task_kwargs`, so we can determine if the task definition already has either task or execution roles associated with it.

## Importance
As it stands right now, if one explicitly sets task_role_arn or execution_role_arn on the ECSAgent, those values will be used as a default unless an explicit ARN is passed within the flow's RunConfig. This introduces a bit of an edge case, where a user can create their own task definition in advance _with_ a task role or execution role as _part of that task definition, but end up having the task be run with the agent's defaults. This can result in behaviour where a task can come up with an unexpected task or execution role, and thus either have too many or too few IAM permissions.

## Checklist

This PR:


youcandanch:
@anna-geller for review! I have a standup in a few minutes but am going to come back to add a few tests on this after.

anna-geller:
Thanks for contributing! This is awesome. I tested this behavior and you're right that in the current Prefect version the agent overrides the task_role_arn even when a custom one is set within the task definition. Your PR fixes that problem.

Here is how I tested it: 
- started an agent with a task role that only has access to S3 (no glue and Athena access): 
`prefect agent ecs start --label dev --task-role-arn arn:aws:iam::xxx:role/prefectTaskRole --execution-role-arn arn:aws:iam::xxx:role/prefectECSAgentTaskExecutionRole`
- built this flow that needs access to Glue and Athena (available on my custom IAM role called: prefectTaskRoleETL)

```python
from datetime import datetime
import requests
import pandas as pd
import awswrangler as wr
from prefect.storage import S3
from prefect.run_configs import ECSRun
from prefect.client.secrets import Secret


FLOW_NAME = "ecs_demo_ecr_etl"
ACCOUNT_ID = Secret("AWS_ACCOUNT_ID").get()
STORAGE = S3(
    bucket="prefect-datasets",
    key=f"flows/{FLOW_NAME}.py",
    stored_as_script=True,
    local_script_path=f"{FLOW_NAME}.py",
)

RUN_CONFIG = ECSRun(
    labels=["dev"],
    task_definition=dict(
        family=FLOW_NAME,
        requiresCompatibilities=["FARGATE"],
        networkMode="awsvpc",
        cpu=1024,
        memory=2048,
        taskRoleArn=f"arn:aws:iam::{ACCOUNT_ID}:role/prefectTaskRoleETL",
        executionRoleArn=f"arn:aws:iam::{ACCOUNT_ID}:role/prefectECSAgentTaskExecutionRole",
        containerDefinitions=[
            dict(
                name="flow",
                image="annageller/community:latest",  # public image, you can use it too
            )
        ],
    ),
    run_task_kwargs=dict(cluster="prefectEcsCluster"),
)


@task
def extract_current_prices():
    url = "https://min-api.cryptocompare.com/data/pricemulti?fsyms=BTC,ETH,REP,DASH&tsyms=USD"
    r = requests.get(url)
    return r.json()


@task
def transform_current_prices(json_data: dict) -> pd.DataFrame:
    df = pd.DataFrame(json_data)
    df["TIME"] = datetime.utcnow()
    return df.reset_index(drop=True)


@task
def load_current_prices(df: pd.DataFrame):
    wr.s3.to_parquet(
        df=df,
        path="s3://prefectdata/crypto/",
        dataset=True,
        mode="append",
        database="default",
        table="crypto",
    )
    print("Data loaded to a data lake")


with Flow(FLOW_NAME, storage=STORAGE, run_config=RUN_CONFIG,) as flow:
    raw_data = extract_current_prices()
    data = transform_current_prices(raw_data)
    load_current_prices(data)
```

This flow running on the latest release Prefect version throws an error:
`botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the GetTable operation: User: arn:aws:sts::xxx:assumed-role/prefectTaskRole/4bbe8f38a56b493fb0af73cd6bd5af00 is not authorized to perform: glue:GetTable on resource: arn:aws:glue:us-east-1:xxx:catalog because no identity-based policy allows the glue:GetTable action`

this error shows that the agent's task role (prefectTaskRole) has been used even though in the ECSRun's task definition I set the role prefectTaskRoleETL. 

Running the same flow with your fix runs as expected. 

From a functionality perspective, this is good to be merged, but if you could add tests, this would be great. Thanks for raising this and a quick fix.

youcandanch:
@anna-geller tests added and rebased off current master!

With Orion coming around the corner -- and also seeing a v1.0 release candidate crop up -- what's the plan for maintenance on the agent code in this repo? Is it still going to be used in the future? While hopping around the agent code, I saw a bunch of places I think could use some touching up and refactoring that I'd be happy to push a PR for, but don't want to necessarily either trample on y'all and/or be modifying something that's just going to go away in the future.

madkinsz:
@youcandanch there are definitely some places that could use refactoring, but since we want to have this code be as stable as possible I'd be hesitant to accept any large changes. We're committed to fixing any issues that crop up here, but the focus of our attention on new features will be in Orion.

youcandanch:
Got it! Makes total sense to me.

youcandanch:
Okay, @anna-geller tests added. I'm seeing the "test_task_library" CircleCI flow failing, but in all honesty I'm a bit stumped as to why -- seems unrelated to my changes, but if I'm missing something just let me know and happy to hop in and fix it!

anna-geller (approved):
It looks good from my side.

madkinsz (approved):

madkinsz:
Thanks!

PR #5335, opened 2022-01-19
Bump hasura image version to 2.1.1 to fix M1 support

treuherz:
## Summary

Proper Apple Silicon support was added to `hasura/graphql` engine in [v2.1.0](https://github.com/hasura/graphql-engine/releases/tag/v2.1.0). This brings it into the Server.

## Changes

Just the version of the `hasura/graphql` image in the server's composefile from 2.0.9 to 2.1.1

## Importance

This should mean the prefect server can start up on M1 Macs without them needing [8GB of RAM](https://github.com/hasura/graphql-engine/issues/6337#issuecomment-969300069)!

## Checklist

This PR:

- ~~adds new tests (if appropriate)~~
- [x] adds a change file in the `changes/` directory (if appropriate)
- ~~updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)~~

PR #5368, opened 2022-01-25
Release 0.15.13

madkinsz:
### Enhancements

- Ensure that maximum log payload sizes are not exceeded - [#5316](https://github.com/PrefectHQ/prefect/pull/5316)

### Server

- Upgrade Hasura to v2.1.1 which includes support for Apple M1 - [#5335](https://github.com/PrefectHQ/prefect/pull/5335)

### Fixes

- Fix bug where logout was required before logging in with a new key if the new key does not have access to the old tenant - [#5355](https://github.com/PrefectHQ/prefect/pull/5355)

### Task Library

- Fix bug where the Airbyte sync job failure would not be reflected in the task state - [#5362](https://github.com/PrefectHQ/prefect/pull/5362)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,17 @@
 
 Released on January 25, 2022.
 
-### Enhancements
+### Server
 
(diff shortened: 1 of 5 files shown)


abrookins (approved):
LGTM 👍

Issue #5389, opened 2022-01-31
[REGRESSION] v0.15.13 is no longer working in PyPy


munjalpatel:
## Description
Upgrade to `v0.15.13` is no longer working in `PyPy`.

This regression is caused by #5316 due to the usage of `sys.getsizeof(log)` which does not work in PyPy.
Here is a potential fix: https://stackoverflow.com/questions/42562838/is-there-any-alternative-to-sys-getsizeof-in-pypy

The actual error:

```
Exception has occurred: TypeError
getsizeof(...)
    getsizeof(object, default) -> int

    Return the size of object in bytes.

sys.getsizeof(object, default) will always return default on PyPy, and
raise a TypeError if default is not provided.

First note that the CPython documentation says that this function may
raise a TypeError, so if you are seeing it, it means that the program
you are using is not correctly handling this case.

On PyPy, though, it always raises TypeError.  Before looking for
alternatives, please take a moment to read the following explanation as
to why it is the case.  What you are looking for may not be possible.

A memory profiler using this function is most likely to give results
inconsistent with reality on PyPy.  It would be possible to have
sys.getsizeof() return a number (with enough work), but that may or
may not represent how much memory the object uses.  It doesn't even
make really sense to ask how much *one* object uses, in isolation
with the rest of the system.  For example, instances have maps,
which are often shared across many instances; in this case the maps
would probably be ignored by an implementation of sys.getsizeof(),
but their overhead is important in some cases if they are many
instances with unique maps.  Conversely, equal strings may share
their internal string data even if they are different objects---or
empty containers may share parts of their internals as long as they
are empty.  Even stranger, some lists create objects as you read
them; if you try to estimate the size in memory of range(10**6) as
the sum of all items' size, that operation will by itself create one
million integer objects that never existed in the first place.
```

madkinsz:
Thanks for the report @munjalpatel, I've drafted a fix at https://github.com/PrefectHQ/prefect/pull/5390

PR #5390, opened 2022-01-31
Fix usage of `sys.getsizeof`

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/5389 by including defaults for usage of `sys.getsizeof` which will otherwise throw a `TypeError` when using PyPy.

Fixes a related issue where `sys.getsizeof` would underestimate the size of log messages. This was a regression introduced in #5316. This improves test coverage to address that case and introduces realistic log size calculations.

### DIFF
--- a/src/prefect/utilities/logging.py
+++ b/src/prefect/utilities/logging.py
@@ -14,9 +14,11 @@ import sys
 import threading
 import time
 import warnings
+import json
 from queue import Empty, Queue
 from typing import Any, List, Optional, Generator, Union
 from contextlib import contextmanager
+from uuid import uuid4
 
 import pendulum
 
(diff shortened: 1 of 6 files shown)


src/prefect/utilities/logging.py
```diff
@@ -109,7 +109,7 @@ def _write_logs(self) -> None:
             try:
                 while self.pending_length < max_batch_length:
                     log = self.queue.get_nowait()
-                    self.pending_length += sys.getsizeof(log)
+                    self.pending_length += sys.getsizeof(log, len(log["message"] + 150))
```
munjalpatel, line 112:
```suggestion
                    self.pending_length += sys.getsizeof(log, len(log["message"]) + 150)
```

munjalpatel (changes requested):

src/prefect/utilities/logging.py
```diff
@@ -109,7 +109,9 @@ def _write_logs(self) -> None:
             try:
                 while self.pending_length < max_batch_length:
                     log = self.queue.get_nowait()
-                    self.pending_length += sys.getsizeof(log)
+                    self.pending_length += sys.getsizeof(
+                        log, len(log.get("message", "")) + 150
```
abrookins, line 113: Can you explain this line a bit more? What does the 150 represent?

madkinsz: I've actually learned that `getsizeof` doesn't include the size of string literal values and the old code is just wrong :'( the tests were passing because we didn't have any strings that were large enough to exceed the overhead. I've rewritten this.

abrookins (approved):
This makes sense! LGTM. 👍

zangell44 (approved):

PR #5402, opened 2022-02-03
SpacyNLP task v3 compatibility

makalaaneesh:
## Summary
This PR addresses https://github.com/PrefectHQ/prefect/issues/5358 by making `SpacyNLP` task compatible with spacy version >= 3.0

## Changes
- The error was thrown by `spacy.load`. As per the [docs for spacy.load](https://spacy.io/api/top-level#spacy.load), the argument name is now `config`, making `component_cfg` invalid. Therefore, `config` is now used
- To maintain backward compatibility, both spacy <3 and >=3 are supported. 
- A new `exclude` param is added to `SpacyNLP` task

## Importance
This will help remove the upper bound constraints on `spacy` in requirements

## Checklist

This PR:

### DIFF
--- a/tests/tasks/spacy/test_spacy_tasks.py
+++ b/tests/tasks/spacy/test_spacy_tasks.py
@@ -1,4 +1,4 @@
-from unittest.mock import MagicMock, Mock
+from unittest.mock import MagicMock, Mock, patch
 
 import pytest
 import spacy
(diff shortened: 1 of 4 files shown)


tests/tasks/spacy/test_spacy_tasks.py
```diff
@@ -17,6 +17,11 @@ def test_initialization(self):
         task = SpacyNLP(text="This is some text", nlp=spacy.blank("en"))
         assert task.text == "This is some text"
 
+    def test_load_nlp_model(self):
```
makalaaneesh, line 20: Ideally, I would want this to run on both spacy<3.0 and spacy>3.0! 
So, i was wondering - is there a way to achieve this with CI? Or some other way?

desertaxle (changes requested):
Thanks for your contribution! This is looking good. Could you remove the version pin for `spacy` from `setup.py`? It shouldn't be necessary after these changes.

tests/tasks/spacy/test_spacy_tasks.py
madkinsz: I don't think we can test against multiple versions simply here. I'd manually test against the old version and have CI test the latest for now.

madkinsz: You can include a test that mocks the spacy version and asserts that the correct call is made for < 3.0

makalaaneesh: > You can include a test that mocks the spacy version and asserts that the correct call is made for < 3.0

Ah, this makes sense, I'll do this, thanks!

makalaaneesh:
Thanks for the quick review @desertaxle ! I've made the changes and also added another test case for spacy versions < 3.0

desertaxle (approved):
LGTM!

Issue #5384, opened 2022-01-28
Prefect build should remove intermediate docker containers


th0ger:
## Description

When building Prefect flows for [Docker storage](https://docs.prefect.io/api/latest/storage.html#docker), all intermediate docker container/image layers are left dangling. This has resulted in our build server(s) running out of disk multiple times.

Normally, docker removes intermediate containers by default:

```console
$ docker build --help
	--rm   Remove intermediate containers after a successful build (default true)
```

But since Prefect utilizes [docker-py](https://docker-py.readthedocs.io), the default is to [NOT remove intermediates](https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.build):

> **rm** (*bool*) – Remove intermediate containers. The `docker` build command now defaults to `--rm=true`, but we have kept the old default of *False* to preserve backward compatibility.

This results in a lot of unnamed `<none>` images and a huge overlay folder `/var/lib/docker/overlay2`. If the server is running multiple docker services, it is not trivial to conclude that these anonymous `<none>` images comes from Prefect.

## Expected Behavior

The proposed fix is to let Prefect remove intermediates by default. Either by setting the default `build_kwargs={"rm": "True"}` in `prefect.storage.docker.Docker()` (I have tested that this works). Or by setting `rm=True` when calling `docker-py` `build()`, unless supplied by the user.

I think most users are expecing Prefect builds to work as `docker build`, not as the peculiar `docker-py` implementation.

There is typically no benefit of persisting all these intermediates. The build process is usually triggered by changed code or run on another build node, so the reuse of image layers seems minimal.

## Reproduction

Create a flow with storage like:

```python
flow.storage = Docker(
        dockerfile="Dockerfile",
        registry_url=DOCKER_REGISTRY,
        image_name=IMAGE_NAME,
        image_tag=VERSION,
        # build_kwargs={"rm": "True"}  # Fix
        )
```

```console
$ prefect build --path myflow.py
# Take note of the image ans container IDs in the build output.
```

```console
$ docker image ls -a
REPOSITORY                           TAG                 IMAGE ID       CREATED         SIZE
<none>                               <none>              51bb2a5722dd   5 minutes ago   1GB
<none>                               <none>              5cb129803c80   5 minutes ago   1GB
<none>                               <none>              e5ed9eed2b3d   5 minutes ago   1GB
<none>                               <none>              250419814322   5 minutes ago   1GB
<none>                               <none>              5630bb427883   5 minutes ago   1GB
[...]

PR #5419, opened 2022-02-08
Use rm=True as default while building docker images for storage

makalaaneesh:
## Summary

Hey folks! This fixes https://github.com/PrefectHQ/prefect/issues/5384
While building docker images for storage, rm=True is used as default, which deletes intermediate containers. 

## Changes

The default while building images for docker storage is changed to now use `rm=True`

## Importance

This prevents the situation where servers run out of disk space because intermediate docker containers are left behind after building images for flows.

## Checklist

This PR:

### DIFF
--- a/src/prefect/storage/docker.py
+++ b/src/prefect/storage/docker.py
@@ -380,6 +380,7 @@ class Docker(Storage):
                 path="." if self.dockerfile else tempdir,
                 dockerfile=dockerfile_path,
                 tag="{}:{}".format(full_name, self.image_tag),
+                rm=self.build_kwargs.pop("rm", True),
                 **self.build_kwargs,
             )
             self._parse_generator_output(output)
--- a/changes/issue5384.yaml
+++ b/changes/issue5384.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "While building docker images for storage, rm=True is used as default, which deletes intermediate containers - (#5384)[https://github.com/PrefectHQ/prefect/issues/5384]"
+
+contributor:
+  - "[Aneesh Makala](https://github.com/makalaaneesh)"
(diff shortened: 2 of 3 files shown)


src/prefect/storage/docker.py
```diff
@@ -380,6 +380,7 @@ def _build_image(self, push: bool = True) -> tuple:
                 path="." if self.dockerfile else tempdir,
                 dockerfile=dockerfile_path,
                 tag="{}:{}".format(full_name, self.image_tag),
+                rm=self.build_kwargs.pop("rm", True),
                 **self.build_kwargs,
```
madkinsz, line 384: If `build` is called more than once and the user sets `rm=False`, this will not behave as intended.

madkinsz (commented):
Thanks for contributing! One issue of note.

src/prefect/storage/docker.py
madkinsz: There isn't really a use-case that I know of, but people use things in weird ways and it'd be a sneaky bug to sort out later :)

madkinsz: They could be sharing `build_kwargs` across multiple flows that they are registering. We don't copy it in `__init__` so we shouldn't mutate the input value. If you move the logic to the constructor, make sure you create a copy of the input `build_kwargs`.

makalaaneesh: Thanks, that's a fair example! 
I made the changes :) I ended up using copy (and not deepcopy) because I thought it would be unlikely that the `__init__` would need to mutate the values of `build_kwargs` passed in by the user. Let me know if you think `deepcopy` makes more sense in this context!

madkinsz: Just `copy` makes sense!

PR #5441, opened 2022-02-14
Add custom serializer support to retrieving task run results

madkinsz:
## Summary

If using a custom serializer for a task result, the `get_task_run_result` / `TaskRun.get_result` methods will fail as the custom serializer is not stored in the serialized state and the client defaults to using a `PickleSerializer`. 

When writing tasks like

```
@task(slug="output", result=GCSResult("<path>", serializer=PandasSerializer(file_type="parquet")), checkpoint=True)
```

The serializer will not be used.

## Changes

This extends `get_result` to allow a custom serializer e.g. `serializer=PandasSerializer(file_type="parquet")` to be passed in for restoring the value. 

## Importance

Extends `get_result` to support our full feature-set for tasks which use custom serializers to store data in more efficient ways.

## Checklist

This PR:

### DIFF
--- a/src/prefect/tasks/prefect/flow_run.py
+++ b/src/prefect/tasks/prefect/flow_run.py
@@ -57,6 +57,7 @@ from prefect.engine.signals import signal_from_state
 from prefect.run_configs import RunConfig
 from prefect.utilities.graphql import EnumValue, with_args
 from prefect.utilities.tasks import defaults_from_attrs
+from prefect.engine.serializers import Serializer
 
 
 @task
@@ -169,7 +170,11 @@ def create_flow_run(
 
 @task
 def get_task_run_result(
-    flow_run_id: str, task_slug: str, map_index: int = -1, poll_time: int = 5
+    flow_run_id: str,
+    task_slug: str,
+    map_index: int = -1,
+    poll_time: int = 5,
+    serializer: Serializer = None,
 ) -> Any:
     """
     Task to get the result of a task from a flow run.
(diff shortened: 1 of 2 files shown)


src/prefect/tasks/prefect/flow_run.py
```diff
@@ -169,7 +170,11 @@ def create_flow_run(
 
 @task
 def get_task_run_result(
-    flow_run_id: str, task_slug: str, map_index: int = -1, poll_time: int = 5
+    flow_run_id: str,
+    task_slug: str,
+    map_index: int = -1,
+    poll_time: int = 5,
+    serializer: Serializer = None,
```
jawnsy, line 177: maybe a bad question because I'm not too familiar with type hints, but should this be `Optional[Serializer]`? or, when are we supposed to use a type `T` vs `Optional[T]`?

madkinsz: Assignment of default values to `None` makes the inferred type automatically `Optional`. There's been a recent movement to avoid this pattern (I can't find a great source covering it right now), but we use it everywhere still. Eventually, we'll make the shift to explicit `Optional` types but they've also shifted the standard pattern to `T | None` in [PEP-0604](https://peps.python.org/pep-0604/) in Python 3.10+.

madkinsz: https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling

jawnsy: Ooo, awesome. Thanks for the education & links!

madkinsz:
I'd love if anyone would take this over. Closing for now.

PR #5464, opened 2022-02-17
Add toggle to disable the Dask cancellation event

madkinsz:
## Summary

@bnaul reported that requesting worker clients during tasks may have detrimental effects on scheduler performance when many tasks are running on a Dask cluster. We request a worker client during every task in order to support better cancellation mechanisms. We can allow users that have performance issues to trade better cancellation for less scheduler load by adding a toggle to disable the `maybe_run` code path.

## Changes

- Adds `disable_cancellation_event` to `DaskExecutor` allowing the cancellation handling feature to be disabled

## Importance

Improved dask performance for users at scale.

## Checklist

This PR:

### DIFF
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -138,6 +138,7 @@ class DaskExecutor(Executor):
         client_kwargs: dict = None,
         debug: bool = None,
         performance_report_path: str = None,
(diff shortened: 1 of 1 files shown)


madkinsz:
Unrelated test failure on Windows. For investigation later preferably.

PR #5465, opened 2022-02-17
Use `schedule=` keyword

ahuang11:
The description mentions: 
"Simple schedules can be attached to Flows via the schedule keyword argument:"

But the example doesn't actually utilize the schedule keyword argument explicitly (only implicitly). This adds the explicit keyword argument `schedule=`

## Checklist

This PR:


ahuang11:
Not sure if this `** adds a change file in the changes/ directory (if appropriate)**` is necessary for this mini pr

PR #5507, opened 2022-03-03
Add `.pipe` method to `prefect.Task`

multimeric:
## Summary
Adds `task.pipe`, which allows for a pipeline-line method chain, e.g. `outer(inner(arg))` can be rewritten as `inner(arg).pipe(outer)`.

## Changes

* Adds a single method to the base `Task` class: `.pipe`

## Importance

Discussion of pipe operators exist in nearly all languages. Bash and other shells have the `|` pipe operator built-in. [The R language has recently implemented a native pipe `|>`](https://www.r-bloggers.com/2021/05/the-new-r-pipe/). [A proposal for pipes in JavaScript was highly popular](https://github.com/tc39/proposal-pipeline-operator#pipe-operator--for-javascript). In the Python world, Pandas has a [`DataFrame.pipe` method](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pipe.html) which has the exact same motivations for chaining function calls when an appropriate DataFrame method doesn't exist. Workflow languages, which belong to the same class of software as Prefect, often have such pipe operators. For example [nextflow recently added the pipe operator](https://www.nextflow.io/docs/latest/dsl2.html#pipes).

Firstly, a pipe allows for "cleaner" code when large numbers of functions are composed. If for example we wanted to call 5 different tasks in Prefect, we might do `task_5(task_4(task_3(task_2(task_1(arg)))))`. This has two issues: firstly, this needs to be read right-to-left instead of left-to-right as in most code. Secondly, it's very ugly. 

We might instead decide to put each task call on each line:
```python
res_1 = task_1(arg)
res_2 = task_2(res_1)
res_3 = task_3(res_2)
res_4 = task_4(res_3)
res_5 = task_5(res_4)
```
This fixes the right-to-left issue, but forces us to assign intermediate variables to each result. This increases the mental effort involved in writing a flow ("what is the meaning of this result?"), but also pollutes the namespace, and it is very easy to mix up the result from one task with the result from another. 

Using pipes solves all of these issues:
```python
task_1(arg).pipe(task_2).pipe(task_3).pipe(task_4).pipe(task_5)
```
There is perhaps some argument that the `.pipe` is too verbose, but it helps with clarity. In theory this could be improved by shortening the method name or instead overriding an operator like `|`.

## Checklist

This PR:


madkinsz:
Before I dive into this, I'm curious if you've tried the `|` operator? We do support piping via that already, but it looks a little different than what you have here.

multimeric:
Oh, and another difference I note is that the `|` operator only sets the task dependencies (ie the task ordering), but it doesn't actually bind the output of the LHS task to the input of the RHS task. This makes it largely useless for making pipelines. So it means if I `task_a | task_b`, where `task_b` requires an argument, the whole thing will fail. And I don't believe it's possible to perform that binding without assigning the output of `task_a` to a variable first.

multimeric:
Let me know if you need any input from me here.

src/prefect/core/task.py
```diff
@@ -907,6 +907,22 @@ def outputs(self) -> Any:
             return_annotation = Any
         return return_annotation
 
+    def pipe(self, task: "Task", *args: Any, **kwargs: Any) -> "Task":
```
madkinsz, line 910: Would you mind changing the argument name here to `__task` so it cannot collide with kwargs?

I'm tempted to ban positional args entirely (with `__task, *, **kwargs`) since the behavior may be confusing.

madkinsz (commented):
I think this makes sense to accept, I've got a few minor requests.

@tpdorsey can you identify a good place to add documentation for this?

tpdorsey:
> @tpdorsey can you identify a good place to add documentation for this?

My first thought is under "Prefect Idioms", but I don't like how this is currently constructed and would prefer to move all of it under "Advanced Tutorials". Which could maybe become "Advanced Topics"?

Not sure this fits in "Core Concepts", which I think is best reserved for higher level Prefect building blocks.

Open to other thoughts about this.

src/prefect/core/task.py
multimeric: Since prefect internally only uses kwargs (?), I'm happy to ban positional args.

multimeric: But are you happy with `task` being positional? Because otherwise the user will always have to `.pipe(task=t, arg=x)`.

multimeric: I solved this by changing the signature to `def pipe(self, task: "Task", /, **kwargs: Any) -> "Task"`, which lets us keep the neat "task" name, but forces it to be a positional argument so doesn't risk clashing with a keyword called "task". I've added a test for this too.

madkinsz:
Thanks for making the changes! `/` is only available in 3.8+ so unfortunately we cannot use it. Using `*` should accomplish what we want https://www.python.org/dev/peps/pep-3102/#specification

multimeric:
My understanding is that you want `task` to be a positional-only argument, but I don't think we can do that using only `*`. That can be used to force arguments to be keyword arguments, but using `**kwargs` already does this. Perhaps I should change `task` to `__prefect_task` as you suggested to make it much harder to collide with a kwarg?

madkinsz:
Ah I'm sorry you're running into some weird behavior from our docs tooling where the `__` name mangling is confusing it... I'll see if I can fix that.

madkinsz:
https://github.com/PrefectHQ/prefect/pull/5519 should resolve these silly doc checks.

multimeric:
Many thanks for that. Unfortunately I can't work out how to hook my new documentation page into the website.

tpdorsey:
@multimeric sorry I've been out all day, will take a look at the docs aspect now.

docs/core/idioms/piping.md
```diff
@@ -0,0 +1,63 @@
...
+* Have to come up with names that don't clash with existing variable names
+* Have to make sure we don't use the wrong variable as an input to any task
+* Have polluted the namespace with variables we aren't going to actually use more than once
+
+Functional piping offers an alternative that doesn't suffer the issues of either these approaches.
+
+## Piping in Prefect
+
+Prefect offers two types of "pipe":
+* `|`, the pipe operator which sets task dependencies but does *not* pass arguments between tasks, and
```
tpdorsey, line 48:
```suggestion
* `|`, the pipe operator, which sets task dependencies, but does *not* pass arguments between tasks
```

docs/core/idioms/piping.md
```diff
@@ -0,0 +1,63 @@
...
+* `.pipe()`, the `Task` pipe method which sets task dependencies *and* passes arguments between tasks
```
tpdorsey, line 49:
```suggestion
* `.pipe()`, the `Task` pipe method, which sets task dependencies *and* passes arguments between tasks
```

src/prefect/core/task.py
```diff
@@ -907,6 +907,23 @@ def outputs(self) -> Any:
...
         return return_annotation
 
+    def pipe(_prefect_self, _prefect_task: "Task", **kwargs: Any) -> "Task":
+        """
+        "Pipes" the result of this task through another task. `some_task().pipe(other_task)` is
+        equivalent to `other_task(some_task())`, but can result in more readable code when used in a long
+        chain of task calls.
+
+        Args:
+            - _prefect_task: The task to apply
```
tpdorsey, line 917:
```suggestion
            - _prefect_task: The task to apply.
```

src/prefect/core/task.py
```diff
@@ -907,6 +907,23 @@ def outputs(self) -> Any:
...
+            - **kwargs: Additional keyword arguments to include as task arguments
```
tpdorsey, line 918:
```suggestion
            - **kwargs: Additional keyword arguments to include as task arguments.
```

src/prefect/core/task.py
```diff
@@ -907,6 +907,23 @@ def outputs(self) -> Any:
...
+
+        Returns:
+            - Task: A new task with the new arguments bound to it
```
tpdorsey, line 921:
```suggestion
            - Task: A new task with the new arguments bound to it.
```

tpdorsey (changes requested):
To add the new doc page to Idioms, you'll have to add a link to the page in /core/idioms/idioms.md. Follow the example of the other links - it's a bullet list of markdown links.

This is a great contribution and excellent documentation, both docstrings and the new doc page. Thank you!

multimeric:
Okay the docs seem to be working. Check out the relevant pages here:
* https://deploy-preview-5507--prefect-docs.netlify.app/core/idioms/idioms.html
* https://deploy-preview-5507--prefect-docs.netlify.app/api/latest/core/task.html#task-2

multimeric:
What do you think about me adding a placeholder option so that you can:
```python
task_a().pipe(task_b, arg_b = PLACEHOLDER, arg_a = 3)
```
This would indicate that `arg_b` should be set to the output of `task_a`, rather than assuming it should be the first parameter. This is a similar convention to what we do in R with pipes:
```r
func_a %>% func_b(1, ., 3)
```
The `.` here is the same as `PLACEHOLDER`, and it indicates where to put the previous function's output in the next function's signature.

multimeric:
Actually, although I think the above suggestion would be a useful improvement in the long run, I'd rather just get this minimal version merged and think about improving it later. In the meantime this use case can be resolved by just passing in a lambda e.g. 
```python
task_a().pipe(task(lambda x: f("positional", keyword=x)))
```

madkinsz:
I think that's an interesting idea. I'd call it `PIPE` and have it importable at `prefect.task` so you'd do `task_a.pipe(task_b, 1, 2, prefect.task.PIPE, a=4)`. I agree it seems better as a follow-up.

tpdorsey (dismissed):

multimeric:
Okay, I've added a change entry. I also added a bit more content to the docs if you want to take a look.

multimeric:
Here are the updated docs: https://deploy-preview-5507--prefect-docs.netlify.app/core/idioms/piping.html

multimeric:
The test is failing but I believe that's just a random failure unrelated to my change.

docs/core/idioms/piping.md
```diff
@@ -0,0 +1,159 @@
...
+secondary arguments. Therefore the following will always fail:
+
+```python
+task_a().pipe(task_b, 1, 2, 3)
+```
+
+This isn't actually a limitation however, because most arguments can be
+expressed as keyword arguments.
+Only [positional-only arguments](https://www.python.org/dev/peps/pep-0570/)
+and `*args` cannot, but these are not supported by Prefect anyway.
```
tpdorsey, line 159:
```suggestion
and `*args` cannot. These are not supported by Prefect.
```

tpdorsey: Reads better to me just to state it. This change also makes it easy to edit in the future should this limitation be removed. But I'm fine with the original if you prefer.

tpdorsey (dismissed):
Well executed documentation. Thanks.

docs/core/idioms/piping.md
multimeric: No, that's a good change.

madkinsz (approved):

max-sixty:
This is great, thanks for adding!

For a future improvement: one way of making this terser would be to override an operator like `>>`. Or if it's confusing to have two operators for slightly different piping, pursue an idea like the `prefect.PIPE` above and allow that to work with `|`.

multimeric:
So I would be happy to use `|` but it would be a breaking change to have it pipe the actual arguments (see above). Also it's not particularly useful unless extra keyword arguments can be bound to it, but I see no way to do that with the current api. e.g. this would work:

```python
task_a() >> task_b.bind(keyword = "foo")
```

max-sixty:
Yes, agree! TBC, I was suggesting it would be possible to make these equivalent:

```python
task_a() >> task_b(kw=3)

task_a().pipe(task_b(kw=3)
```

To use `|`, we'd need the placeholder you suggested above https://github.com/PrefectHQ/prefect/pull/5507#issuecomment-1062852090 like:

```python
task_a() | task_b(x=prefect.PIPE, kw=3)
```

If I'm thinking about this correctly, it would be backward-compatible, because the default would be to not pass anything, so consistent with the current approach.

multimeric:
```
task_a() >> task_b(kw=3)
```
I think this can't work with the current task API because the moment you `task_b(kw=3)` it will say "not all arguments were provided" and crash. So in order to do this we would need a "bind but don't check for complete arguments" method which doesn't yet exist afaik (`bind` doesn't help here).

```
task_a() | task_b(x=prefect.PIPE, kw=3)
```
I think this *might* work as is, because technically all the arguments have been provided here. We could just add another code path for the `|` operator which checks for the `PIPE` constant, and if it's in the argument dict we swap it out for the real output. As you say, it would be backward-compatible because the `PIPE` constant hasn't existed before now, so it's not possible for someone to have an existing pipeline break with this change.

PR #5508, opened 2022-03-03
Add workaround for self kwarg

multimeric:
## Summary
Fix a task failure when `self` is used as a task argument

## Changes
`self` is popped from the formatting dictionary used to populate result strings. This should not be included because it means that we get `result.exists(self="foo", self="bar")`, which is an error.

## Importance

Currently there is a crash whenever any method is used directly as a function. For example, in the below code note how the task's function *is* `Path.is_absolute` which is a method that we are using like a function, and which therefore needs a `self` argument:

```python
    from pathlib import Path

    seconds_task = task(Path.is_absolute, target="{{task_slug}}_{{map_index}}", result=LocalResult())
    initial = Path("foo")

    with Flow("test") as flow:
        seconds_task(initial)
        flow.run()
```

## Checklist

This PR:

### DIFF
--- a/tests/core/test_task.py
+++ b/tests/core/test_task.py
@@ -830,10 +830,11 @@ def test_task_called_outside_flow_context_raises_helpful_error(use_function_task
 def test_task_call_with_self_succeeds():
     from pathlib import Path
 
-    seconds_task = task(Path.is_absolute, target="{{task_slug}}_{{map_index}}", result=LocalResult())
+    seconds_task = task(
+        Path.is_absolute, target="{{task_slug}}_{{map_index}}", result=LocalResult()
+    )
     initial = Path("foo")
 
     with Flow("test") as flow:
         seconds_task(initial)
         assert flow.run().is_successful()
-
--- a/tests/core/test_task.py
+++ b/tests/core/test_task.py
@@ -828,12 +828,19 @@ def test_task_called_outside_flow_context_raises_helpful_error(use_function_task
 
 
 def test_task_call_with_self_succeeds():
-    from pathlib import Path
+    import dataclasses
+
+    @dataclasses.dataclass
+    class TestClass:
+        count: int
+
+        def increment(self):
+            self.count = self.count + 1
 
     seconds_task = task(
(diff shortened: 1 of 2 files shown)


multimeric:
I have no idea why one unrelated test failed. Any ideas?

tests/core/test_task.py
```diff
@@ -825,3 +825,16 @@ def run(self, x):
...
+    from pathlib import Path
+
+    seconds_task = task(
+        Path.is_absolute, target="{{task_slug}}_{{map_index}}", result=LocalResult()
+    )
+    initial = Path("foo")
+
+    with Flow("test") as flow:
+        seconds_task(initial)
+        assert flow.run().is_successful()
```
madkinsz, line 840: It looks like you're using `Path.is_absolute` as a method here to show how `self` needs special handling, but I'm worried that won't be apparent to someone reading this in the future. Can you define a simple class with a single method instead?

madkinsz:
Thanks for contributing! That test is flaky, I've rerun the suite.

madkinsz (approved):
FWIW the real solution here is for us to use `__prefect_self__` in the implementations like pydantic does so they do not collide but the user's `self` is still accessible. I don't think a change like that is in scope though.

Thanks for the fix!

madkinsz:
@multimeric I just noticed you didn't include a changelog entry -- I'll add one at release time. Do you want to be listed as a contributor?

multimeric:
> FWIW the real solution here is for us to use __prefect_self__ in the implementations like pydantic does so they do not collide but the user's self is still accessible. I don't think a change like that is in scope though.

Damn, that's a way better solution. I would have been happy to do that instead?

> @multimeric I just noticed you didn't include a changelog entry -- I'll add one at release time. Do you want to be listed as a contributor?

Yes please! I'm happy to add a changelog entry now if you'd like

madkinsz:
> Damn, that's a way better solution. I would have been happy to do that instead?

I think we'd want to do it in quite a few places and I'm hesitant unless this patch causes issues for people.

> Yes please! I'm happy to add a changelog entry now if you'd like

Don't worry about it, I editorialize the entries when releasing anyway. I just have to remember :)

PR #4731, opened 2021-07-01
add default values to datetime parameters

wangjoshuah:
## Summary
Add default values to `DateTimeParameter` tasks

## Changes
Enable users to set default values for the datetime parameter class. 

## Importance
It would be convenient to have sane defaults for the user to fall back on.

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #5541, opened 2022-03-12
Feat/Support default value on `DatetimeParameter`

### LINKED ISSUE #5448: Default option for DateTimeParameter
## Current behavior
You can set a default value for a `Parameter`, ie:
```
my_param = Parameter('my_param', default=None)
my_other_param = Parameter('my_other_param', default='hi')
```
but not for a `DateTimeParameter`. If nothing is supplied by the user, it will default to `None` if it's optional (`required = False`), and will raise an error if `required = True`.

When a DateTimeParameter is in a flow, it's likely that there's code later in the flow which operates on this date (eg. adding days) and which expect `DateTime` objects. Some of these operations raise errors when they're provided with a `NoneType` object. This means that if a `DateTimeParameter` is optional, it is likely to result in errors downstream. This means that we need to supply a value for all DateTimeParameters each time we run the flow (this gets cumbersome if we are testing and triggering repeated runs through the UI with the same DateTimeParameters each time).

## Proposed behavior
It would be awesome if we can set default values for `DateTimeParameter`s too which mirrors the behaviour of the defaults on `Parameter`. That is, if a `DateTimeParameter` is optional but we have provided a default, the flow can still run using this default value rather than raising an error or needing to supply values each time we run a flow. Currently, we would have to handle the `None` case when it's required (eg. `if date is None then date=...`), but this is effectively the same as setting a default. It would be better and more readable if the default was attached to the `DateTimeParameter` argument upfront.

To stay consistent with the behaviour of `Parameter`, if a `DateTimeParameter` is required and a default is provided, it will still raise a 'no parameter received' error if nothing is supplied by the user.

Ideally we could pass through a `Date` or `DateTime` object, but a string would suffice too, if JSON-serializable requirements get in the way.

## Example
```
# Ideal
from pendulum import DateTime
my_date = DateTimeParameter('my_date', default=DateTime(2020, 9, 1))
# Admitting a string would be ok too
my_other_date = DateTimeParameter('my_other_date', default='2021-01-01T00:00:00+00:00')
```

sp1thas:
## Summary

The purpose of this PR is to support a default value on `DatetimeParameter`. This PR is based on #4731 

Closes #5448 

## Changes

This PR changes the `prefect.core.parameter.DateTimeParameter`

## Importance

This PR is important because a default value will be supported on `DateTimeParameter` too.

## Checklist

This PR:


madkinsz (approved):
LGTM! Thanks for contributing!

Issue #5552, opened 2022-03-15
Parent flow fails while child succeeds 


thomasfrederikhoeck:
## Description

Occasionally I get the following error when running sub-flows using `create_flow_run` and `wait_for_flow_run`
```python
Task "wait_for_flow": Exception encountered during task execution!
Traceback (most recent call last):
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/engine/task_runner.py", line 876, in get_task_run_state
    value = prefect.utilities.executors.run_task_with_timeout(
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/utilities/executors.py", line 467, in run_task_with_timeout
    return task.run(*args, **kwargs)  # type: ignore
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/tasks/prefect/flow_run.py", line 258, in wait_for_flow_run
    flow_run = FlowRunView.from_flow_run_id(flow_run_id)
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/backend/flow_run.py", line 595, in from_flow_run_id
    return cls._from_flow_run_data(flow_run_data, task_runs=task_runs)
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/backend/flow_run.py", line 528, in _from_flow_run_data
    state = State.deserialize(flow_run_data.pop("serialized_state"))
  File "/root/miniconda3/lib/python3.8/site-packages/prefect/engine/state.py", line 390, in deserialize
    state = StateSchema().load(json_blob)
  File "/root/miniconda3/lib/python3.8/site-packages/marshmallow_oneofschema/one_of_schema.py", line 153, in load
    raise exc
marshmallow.exceptions.ValidationError: {'_schema': 'Invalid data type: None'}
```

This will result in the parent flow failing. **However the child flow doesn't fail and run to success**. When looking at the time stamps of the fail it looks like the parent flow fails almost before the child flow starts. I think this suggests that error is happening  because there is a little delay between the task being send through `create_flow_run` and being available for `wait_for_flow`

## Expected Behavior
That parent flow completes with success too. 

## Reproduction
It is hard to reproduce as it only happens sometimes but you need to create a parent and child flow using `create_flow_run` and `wait_for_flow`

## Environment
Prefect Cloud
Kubernetes agent 
Azure Kubernetes Service
python 3.8.11
prefect 0.15.13

madkinsz:
[...]

PR #5554, opened 2022-03-15
Use pending state when flow run state is missing

Fraznist:
## Summary

There is a slight timeframe in which an initialized flow run does not have a serialized state. During this time a backend FlowRunView object cannot be initiated.

## Changes

With this change, `FlowRunView` instances can be initialized when `serialized_state` value is None. A default state of `Pending` is used to fill the null value instead, with a short message describing that the flow run state is not yet initialized.

## Importance
`wait_for_flow_run` task could fail, even though child flow executes correctly. See issue #5552 

## Checklist

This PR:

### DIFF
--- a/changes/pr5554.yaml
+++ b/changes/pr5554.yaml
@@ -1,0 +1,5 @@
+fix:
+  - 'Fixes bug where the FlowRunView could fail to initialize when the backend has no state data - [#5554](https://github.com/PrefectHQ/prefect/pull/5554)'
+
+contributor:
+  - '[Emre Akgün](https://github.com/Fraznist)'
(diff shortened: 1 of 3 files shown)


madkinsz (approved):
Thanks! This can be a `fix` change entry like.. "Fixes bug where the `FlowRunView` could fail to initialize when the backend has no state data"

Fraznist:
> Thanks! This can be a `fix` change entry like.. "Fixes bug where the `FlowRunView` could fail to initialize when the backend has no state data"

I'll just shamelessly copypasta this since I'm unonriginal when it comes to commits :sweat_smile:

thomasfrederikhoeck:
Will this make it into next release, @madkinsz ? :-)

madkinsz (approved):

Issue #5523, opened 2022-03-09
Docs Issue - suggestion to clarify


kdog1425:
Hello!

I've found an issue here:`parallel pipelines.`

I believe this is a bit misleading - because the work is not done in parallel, unless the relevant executor is chosen. I suggest clarifying and adding a note which emphasises that to actually get parallel execution, one needs to set LocalDaskExecutor or DaskExecutor as the executor class. 

Page reference: https://docs.prefect.io/core/concepts/mapping.html

Fraznist:
Instead of `creating independent parallel pipelines.`
Something like `creating identical, but independent, task pipelines`

I feel like introducing how to achieve actual parallelism via executors should be out of scope at that stage of the docs. Docs shouldn't make you jump around pages, especially forward jumps tend to be confusing for beginners. Removing the `parallel` keyword should eliminate the need to mention `Dask` or `executors`.

kdog1425:
Makes sense to me @Fraznist - shall I open a PR?

tpdorsey:
Resolved by https://github.com/PrefectHQ/prefect/pull/5587

PR #5587, opened 2022-03-22
Update mapping.md

kdog1425:
## Summary
This proposes a change to the docs, with reference to this issue
https://github.com/PrefectHQ/prefect/issues/5523

## Changes
Clarifying the use of the word `parallel` in the context of mapping.

## Importance
This is important because without it developers may think that mapping creates parallel execution, while in fact the execution depends on the Executor which is used.

## Checklist

This PR:

### DIFF
--- a/docs/core/concepts/mapping.md
+++ b/docs/core/concepts/mapping.md
@@ -1,6 +1,6 @@
 # Mapping
 
-Prefect introduces a flexible map/reduce model for dynamically executing parallel tasks.
+Prefect introduces a flexible map/reduce model for dynamically executing parallel tasks (parallel in the sense of independent data handling for each task, not parallelized simultaneous execution).
 
 Classic "map/reduce" is a powerful two-stage programming model that can be used to distribute and parallelize work (the "map" phase) before collecting and processing all the results (the "reduce" phase).
 
--- a/docs/core/concepts/mapping.md
+++ b/docs/core/concepts/mapping.md
@@ -1,6 +1,6 @@
 # Mapping
 
-Prefect introduces a flexible map/reduce model for dynamically executing parallel tasks (parallel in the sense of independent data handling for each task, not parallelized simultaneous execution).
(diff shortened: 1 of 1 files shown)


docs/core/concepts/mapping.md
```diff
@@ -1,6 +1,6 @@
 # Mapping
 
-Prefect introduces a flexible map/reduce model for dynamically executing parallel tasks.
+Prefect introduces a flexible map/reduce model for dynamically executing parallel tasks (parallel in the sense of independent data handling for each task, not parallelized simultaneous execution).
```
tpdorsey, line 3: This is a good catch.

Perhaps take this further and, instead of qualifying parallel (because execution could, in fact, be parallelized _or_ distributed), let's take parallel out of this introductory sentence entirely.

I'm thinking rewriting as something like: "a flexible map/reduce model for dynamically executing tasks across an iterable input. This, in turn, gives you the ability to execute mapped tasks in a distributed or parallel manner using an [executor](link) like the [DaskExecutor](link)."

I don't have a problem with linking users to the more advanced material, but you're right that we should explain clearly first.

tpdorsey (commented):
Good catch

docs/core/concepts/mapping.md
kdog1425: I like your rewrite @tpdorsey. The main reason I wanted to qualify 'parallel' instead of removing it, was because it appears multiple times later on in this page. Do you feel it is not a problem in that sense?

kdog1425: I think that can work, I'll work that in and commit in a bit.

tpdorsey:
@madkinsz do you understand why tests are failing here?

tpdorsey (approved):

madkinsz:
> @madkinsz do you understand why tests are failing here?

Something wrong with Dask right now. I'll patch separately.

Issue #5556, opened 2022-03-16
Ban calling `flow.run()` from within its own context


madkinsz:
## Current behavior

You are allowed to call a flow in its context manager

```python
with Flow(...) as flow:
    flow.run()
```

This can lead to poor behavior, e.g. https://github.com/PrefectHQ/prefect/issues/5555

## Proposed behavior

`flow.run` should not be callable from within the the flow context. If called, an error should be raised explaining that the flow must be called outside of the context block.

## Example

This prevents errors and aligns user expectations with the internal behavior of Prefect.

Fraznist:
it seems `flow.run` employs a similar check to detect if it is being called while it is being loaded to storage:
```python
if prefect.context.get("loading_flow", False):
    warnings.warn(
        "Attempting to call `flow.run` during execution of flow file will lead to "
        "unexpected results.",
        stacklevel=2,
    )
    return None
```
Similarly, being in a `with Flow()` context manager sets a key named `flow` in the `prefect.context`. See:
https://github.com/PrefectHQ/prefect/blob/96183e95ad069c927799a6a272bc3d9a9cce3197/src/prefect/core/flow.py#L357

In short, I think the following should work:
```python
if "flow" in prefect.context:
    raise ValueError("flow.run cannot be called from within a flow context.")
```
I just slapped `ValueError` in there but what would be a sensible exception @madkinsz ?

madkinsz:
Yeah that's a reasonable implementation. I'd raise a `RuntimeError` and perhaps include a minimal example in the error message indicating how run should be called e.g.

```python
with Flow(...) as flow:
   # wrong
   flow.run()

# right
flow.run()
```

tirkarthi:
I guess this can be closed with https://github.com/PrefectHQ/prefect/pull/5588 merged

madkinsz:
Yep thanks!

PR #5588, opened 2022-03-22
Raise error if flow.run called within Flow ctx

Fraznist:
## Summary
Implements #5556 -> Raise informative error when `flow.run()` is called within a flow context.

## Details
Found a check in `flow.register()` that uses `hasattr` instead of `'flow' in prefect.context`. Opted to use that for consistency. 

## Checklist

This PR:

### DIFF
--- a/changes/pr5588.yaml
+++ b/changes/pr5588.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - 'Raise RuntimeError when calling flow.run within Flow context - [#5588](https://github.com/PrefectHQ/prefect/pull/5588)'
+
+contributor:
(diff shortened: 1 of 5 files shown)


Fraznist:
Side note, the check in `flow.register` raises a `ValueError` instead of what we discussed @madkinsz (`RuntimeError`). That is left unchanged.
https://github.com/PrefectHQ/prefect/blob/96183e95ad069c927799a6a272bc3d9a9cce3197/src/prefect/core/flow.py#L1671

madkinsz (approved):

madkinsz:
Thank you!

Issue #5592, opened 2022-03-23
context.get("logger") and self.logger behave differently on mapped task runs (possibly more)


Fraznist:
## Description
For a class based task, `self.logger` is a snapshot of a prefect logger instance during task initialization time. While the logger within prefect context (`prefect.context.get("logger")`) changes throughout the lifecycle of a flow build/run/etc. This causes the two methods of logging within a task run to behave differently. 

One instance is mapped task runs, where the loggers name indicates both the task name and the map index. Keep in mind that differing behavior may not be limited to mapped tasks, this is just the example I could find.

## Expected Behavior

Using self.logger within `task.run` and deriving the logger from context to behave the same.

## Reproduction

```python
class MyTask(prefect.Task):
    def run(self):
        map_index = prefect.context.get("map_index")
        self.logger.warning("{}".format(map_index))

@prefect.task
def myrun():
    logger = prefect.context.get("logger")
    map_index = prefect.context.get("map_index")
    logger.warning("{}".format(map_index))

with prefect.Flow("Logger_Compare") as f:
    MyTask().map(upstream_tasks=[list(range(1))])
    myrun.map(upstream_tasks=[list(range(1))])
f.run()
```

Fraznist:
## Proposal
Within `prefect.Task`:
```python
@property
def logger(self):
    return prefect.context.get("logger")
```
This way `self.logger` still returns a logger, and both options boil down to `context.logger`

madkinsz:
I'm not entirely sure of the implications of this without digging in, but if you want to give it a go I'm willing to review it.

PR #5594, opened 2022-03-23
Convert task.logger to instance_property

Fraznist:
## Summary

#5592 
Use `@property` to use `context.logger`  when accessing `task.logger`

## Importance

`task.logger` and `context.logger` could produce different results even if used under the same conditions. This pr fixes that by using `context.logger` under the covers when accessing `task.logger`

## Checklist

This PR:


tests/core/test_task.py
```diff
@@ -364,13 +364,6 @@ def test_create_task_with_task_run_name(self):
 def test_task_has_logger():
     t = Task()
     assert isinstance(t.logger, logging.Logger)
-    assert t.logger.name == "prefect.Task"
```
madkinsz, line 367: What is the new logger name if you're outside a flow?

Fraznist: Just `prefect`. It seems `prefect.context.logger` isn't modified at all when outside a flow.
We can modify it to use `prefect.Task` if the current name is less informative than `prefect.Task`, like plain `prefect`. And use the context supplied logger otherwise.

madkinsz: I'm a little concerned about breaking compatibility for people who are relying on it being `prefect.Task`, let me check with the rest of the team.

Fraznist: Sure, I don't know what I was thinking deleting test to be honest :sweat_smile:. Anyways, I'm thinking something like this would be compatible with the tests I removed?
```python
    @instance_property
    def logger(self) -> logging.logging.Logger:
        # If metadata about a task exists in context, then the logger must be modified
        if prefect.context.get("task_id"):
            return prefect.context.get("logger")

        return logging.get_logger(self.name)
```

madkinsz: That seems pretty reasonable. I'll check back again once I see what other people think about it.

madkinsz:
This is currently stalled because, after chatting internally, we think the logger attached to the `Task` object is _supposed_ to be different than the logger in the context. I think the suggestion is that the logger should always be retrieved from context during the `run` method. I'm not sure we should change the behavior of the task object's logger now that we're at 1.x.

madkinsz:
Closing as stale. Feel free to continue discussion.

Issue #5380, opened 2022-01-27
Adding stuff to context


kvnkho:
## Proposed behavior

This is to track requests of things to add to context. There is a user request to add two things:

1.flow_group_id
2.tenant_id

## Example

This would let you do:
`prefect.context.flow_group_id` 
and
`prefect.context.tenant_id`

heyitskevin:
Made a PR for this issue here. I hope I followed all the contribution guidelines. Please correct me if I didn't.
https://github.com/PrefectHQ/prefect/pull/5604

kvnkho:
Thanks for the PR! The core team will review it.

cicdw:
Closing as stale, and in v2 we can add lazy-loading attributes like this via `prefect.runtime`

PR #5604, opened 2022-03-27
Issue 5380

heyitskevin:
## Summary
This PR adds additional data to the context object per this issue: https://github.com/PrefectHQ/prefect/issues/5380

## Changes
Small changes to the `get_flow_run_info()` method that hits GQL to pull down additional context data.

## Checklist
Allows for further context introspection by tenant and by group.

This PR:


kvnkho:
I am working on QA-ing this now to see if it bloats the context too much, which we have issues with for mapping at scale. Will report back.

madkinsz:
We're hesitant to extend the context with additional data due to performance concerns. We've already had quite a few instances of memory strain due to context copying operations. 

I'm going to consider this a bit further, but now that we're at 1.x I'm worried about introducing performance regressions.

discdiver:
Can we close this @madkinsz ?

madkinsz:
@cicdw had suggested that we should lean towards accepting it in the past which is why I did not close it when cleaning up old changes.

PR #5613, opened 2022-03-29
Add JOB_IMAGE environment variable

cnuss:
## Summary
This PR adds a optional `JOB_IMAGE` environment variable.

## Changes
This PR adds a new environment variable `JOB_IMAGE` to the `get_flow_image` method, with precedence above the provided default.

## Importance
Locally we run on an image `workflows:latest`, however, in the cloud, we build an image and our CI system can inject the image name and tag into the environment variables at runtime.

This change allows us to do just that!

## Checklist

This PR:


madkinsz:
While this follows the format of the other variables JOB_MEM_REQUEST, JOB_MEM_LIMIT, etc. I don't think those are actually being used? It'd be nice to support this pattern but I don't find it particularly clear that that specifying `JOB_IMAGE` would change the behavior of Prefect flow runs. Perhaps `PREFECT_FLOW_RUN_JOB_IMAGE_DEFAULT` would be better?

madkinsz (changes requested):
Marking this as changes requested so I do not get updates that my review is pending.

madkinsz:
Closing as stale. Feel free to reopen.

PR #5631, opened 2022-04-01
Add `watch_worker_events` toggle to `DaskExecutor`

### LINKED ISSUE #5549: `dask_kubernetes` Restart feature is broken
It seems like something in prefect's codebase broke `dask_kubernetes` ability to restart and respawn worker pods, if for some reason the worker pod is killed the default behavior of  `dask_kubernetes` is to restart this pod again however when used with prefect it does not, using the code below it will create dask's cluster for testing - try killing one or two of the pods -

```python
from dask_kubernetes import KubeCluster, make_pod_spec

pod_spec = make_pod_spec(image='daskdev/dask:latest')
cluster = KubeCluster(pod_spec)
cluster.adapt(minimum=4, maximum=4)
input("pause ...")
```

Using prefect dask Kubernetes run - try killing worker pods they just die -
```python
from dask_kubernetes import KubeCluster, make_pod_spec
from prefect.run_configs import KubernetesRun
from prefect.storage import Docker
import prefect
import time

storage = Docker(
    base_image="prefecthq/prefect:1.0.0-python3.9",
    python_dependencies=[
        "dask-kubernetes==2021.3.1",
    ],
)


@task()
def transform(number: int)->int:
    time.sleep(30)


with Flow(
    "Dask Kubernetes Flow",
    storage=storage,
    executor=DaskExecutor(
        cluster_class=lambda: KubeCluster(make_pod_spec(image=prefect.context.image)),
        adapt_kwargs={"minimum": 4, "maximum": 4},
    ),
    run_config=KubernetesRun(),
) as flow:
    transform.map([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])

```

## Expected Behavior
I expect to see workers restarting and respawning again

madkinsz:
Adds a `watch_worker_events` toggle to the `DaskExecutor`. When disabled, we will not subscribe to the Dask worker events which are normally logged when workers are added or removed from the connected cluster.

This toggle is disabled by default when `adapt_kwargs` is used to resolve issues where Dask clusters fail to scale correctly as only a single subcriber is allowed at once and Prefect replaces the clusters handler. Otherwise, it is enabled by default for backwards compatibility. Users may override the default behavior by setting `True` or `False` explicitly. 

Closes #5549.

### DIFF
--- a/changes/pr5631.yaml
+++ b/changes/pr5631.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix issue where adaptive Dask clusters failed to replace workers - [#5549](https://github.com/PrefectHQ/prefect/issues/5549)"
--- a/src/prefect/executors/dask.py
+++ b/src/prefect/executors/dask.py
@@ -275,6 +275,11 @@ class DaskExecutor(Executor):
         """
         This method is triggered when a worker is added or removed from the cluster.
 
+        This method will not be called if `watch_worker_status` is not set.
+
+        We recommend not relying on this method since worker status subscription is used
+        by Dask cluster implementations to manage worker state.
+
         Args:
             - op (str): Either "add" or "remove"
             - message (dict): Information about the event that the scheduler has sent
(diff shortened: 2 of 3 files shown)


desertaxle (dismissed):
LGTM!

jawnsy (dismissed):

jawnsy (approved):

Issue #5139, opened 2021-11-11
task retry functionality should have option to specify on which errors to retry


schmidt-jake:
## Current behavior
When using `Task.max_retries`, a task will be retried for *any* exception that is thrown in the task.

## Proposed behavior
There should be a new kwarg for `Task` called something like `retry_on` which should take an iterable of exception classes, which, if thrown, will trigger a retry. If an exception is raised that isn't in `retry_on`, then the task should immediately fail without retrying. If `retry_on` is None (which should be its default), it will retry on all errors (which is the current functionality). This would bring prefect's retry functionality in line with common python retry APIs such as `retrying` and `tenacity`.

## Example
```python
from prefect import task, Flow, Parameter


@task(retry_on={ValueError})
def fail_sometimes(a: int):
    if a % 2 == 0:
        raise ValueError()
    else:
        raise KeyError()

with Flow("test") as flow:
    fail_sometimes(a=Parameter("a"))

flow.run(a=0)  # hits ValueError and retries
flow.run(a=1). # hits KeyError and fails/returns immediately
```

jhemmingsson:
This would be a great feature, indeed! +1

Kazmirchuk:
I've hit this problem as well
for now, I'll try raising `prefect.engine.signals.ENDRUN` to skip task retries
I wonder, if issues like this have any chance to be fixed, given that probably most development effort is now focused on Orion?

madkinsz:
We're still working on v1 :) this feature is small enough that I think it is reasonable to add. See #5634.

PR #5634, opened 2022-04-01
Add `retry_on` to allow tasks to retry on a subset of exception types

madkinsz:
## Summary

Adds `retry_on: Optional[Iterable[Type[Exception]]]` to tasks allowing users to specify a subset of exception types that tasks should retry on.

## Changes

- Adds `retry_on` as an optional setting for `Task`
- Adds `retry_on` handling to the task runner

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/5139

## Checklist

This PR:


src/prefect/core/task.py
```diff
@@ -368,6 +373,28 @@ def __init__(
...
+                ) from exc
+            for v in retry_on:
+                if not isinstance(v, type):
+                    raise TypeError(
+                        f"Invalid `retry_on` value {v!r}. "
+                        f"Expected an exception type but got an instance of {type(v).__name__}"
+                    )
+                if not issubclass(v, Exception):
+                    raise TypeError(
+                        "Invalid `retry_on` value {v!r}. Expected an exception subclass."
```
ahuang11, line 396: Does this need f string?

src/prefect/core/task.py
```diff
@@ -368,6 +373,28 @@ def __init__(
             raise ValueError(
                 "A datetime.timedelta `retry_delay` must be provided if max_retries > 0"
             )
+        if retry_on and not max_retries > 0:
+            raise ValueError(
+                "A numer of `max_retries` must be provided if `retry_on` is set."
```
ahuang11, line 378: Number

src/prefect/core/task.py
```diff
@@ -368,6 +373,28 @@ def __init__(
...
+            )
+        if retry_on:
+            try:
+                retry_on = set(retry_on)
+            except TypeError as exc:
+                raise TypeError(
+                    "Unexpected type for `retry_on`. "
+                    "Expected an iterable of exception types."
+                ) from exc
+            for v in retry_on:
```
ahuang11, line 388: For clarity, can we rename `v` to `exc_type`?

src/prefect/core/task.py
```diff
@@ -368,6 +373,28 @@ def __init__(
...
+                if not isinstance(v, type):
```
ahuang11, line 389: Would the next if statement already catch this?
```
if not issubclass(v, Exception):  
```

Also, I'm confused about checking against `type`

ahuang11: Okay TIL you can isinstance against type, but still wondering what this catches that `issubclass` doesn't
```
>>> isinstance(ValueError, type)
True
>>> isinstance(3, type)
False
>>> isinstance(ValueError, Exception)
False
>>> issubclass(ValueError, Exception)
True
```

ahuang11 (changes requested):

src/prefect/core/task.py
madkinsz: We'll throw an exception if the first argument to `issubclass` is not a type so we need to check that first

```
In [2]: issubclass(ValueError(), Exception)
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 issubclass(ValueError(), Exception)

TypeError: issubclass() arg 1 must be a class
```

src/prefect/core/task.py
madkinsz: I could, but it's a little misleading since it's not confirmed to be an exception type yet. At this point, all we know is that it's some value. Thoughts?

src/prefect/core/task.py
madkinsz: 👍 
Do these as suggestions so we can just fix them immediately
```suggestion
                "A number of `max_retries` must be provided if `retry_on` is set."
```

src/prefect/core/task.py
madkinsz:
```suggestion
                        f"Invalid `retry_on` value {v!r}. Expected an exception subclass."
```

src/prefect/core/task.py
```diff
@@ -315,6 +319,7 @@ def __init__(
         tags: Iterable[str] = None,
         max_retries: int = None,
         retry_delay: timedelta = None,
+        retry_on: Iterable[Type[Exception]] = None,
```
ahuang11, line 322: Should we accept a non-iterable `Exception` too?

```suggestion
        retry_on: Union[Type[Exception], Iterable[Type[Exception]]] = None,
```

I imagine there'd be a lot more people checking a single exception rather than multiple.

madkinsz: Sure, want to push a commit that adds support?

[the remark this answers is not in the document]
madkinsz: Yeah, that might be nice. However, I think you can basically do that with a state handler right now e.g. add a state handler that checks for a failed state and returns a retry state if the result matches what you want. I'd rather keep this interface simple and point power users to existing hooks for state handling.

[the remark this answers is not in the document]
jawnsy: Makes sense! I definitely agree with the decision to make things simple and tailored to the most common use case

madkinsz:
@ahuang11 your changes are not passing mypy and the docstring will need to be updated as well

src/prefect/core/task.py
```diff
@@ -393,7 +393,7 @@ def __init__(
                         f"Invalid `retry_on` value {v!r}. Expected an exception subclass."
                     )
         else:
-            self.retry_on = None
+            self.retry_on = set([])
```
madkinsz, line 396: Let's leave this as `None` please and make the type just include `Optional[...]`. 

For future reference this could just be `set()`.

src/prefect/core/task.py
```diff
@@ -393,7 +396,7 @@ def __init__(
                         f"Invalid `retry_on` value {v!r}. Expected an exception subclass."
                     )
         else:
-            self.retry_on = set([])
+            self.retry_on = None
```
madkinsz, line 399: Sorry to keep nitpicking, but you can set this to `None` on L382 and remove this branch

ahuang11 (approved):

PR #5637, opened 2022-04-04
fix for secret.exists()

satoshiking:
## Summary
we should not fetch secrets from could if we work with `backend` server (not cloud one)

## Checklist

This PR:


madkinsz:
Thanks! Can you add a changes entry and update the tests?

madkinsz (approved):
Thanks again!

PR #5643, opened 2022-04-05
Handle empty string as tenant id

zangell44:
## Summary

Fixes a bug where logging in for the first time would throw an error.

See bug report here https://github.com/PrefectHQ/cloud/issues/3951

## Importance

When first logging in, several attempts are made to find a tenant id. Several of these checks can return `''` instead of `None`, which leads to the error message detailed in the bug report.
https://github.com/PrefectHQ/prefect/blob/master/src/prefect/client/client.py#L139

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -533,7 +533,7 @@ class Client:
         # is still being hit
         rate_limited = response.status_code == 429
         if rate_limited and rate_limit_counter <= 6:
-            jitter = random.random() * 10 * (2 ** rate_limit_counter)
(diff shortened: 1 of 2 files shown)


src/prefect/client/client.py
```diff
@@ -284,7 +284,7 @@ def tenant_id(self) -> str:
 
     @tenant_id.setter
     def tenant_id(self, tenant_id: Union[str, uuid.UUID, None]) -> None:
-        if tenant_id is None:
+        if tenant_id in (None, ""):
```
madkinsz, line 287: Just a style preference here
```suggestion
        if not tenant_id:
```

Issue #5656, opened 2022-04-07
`Flow.visualize` fails when there are mapped lambdas


bayersglassey-zesty:
## Description

If any edge in a flow is a lambda with .map called on it, then Flow.visualize produces an error (see below).

## Expected Behavior

See the visualized graph

## Reproduction

```
from prefect import Flow, task

inc = task(lambda x: x+1)

with Flow('test') as flow:
    result = inc.map([1,2,3])

flow.visualize()
```

...this dies with something like:
```
CalledProcessError: Command '[PosixPath('dot'), '-Kdot', '-Tpdf', '-O', 'tmp7z7ozyzt']' returned non-zero exit status 1. [stderr: b"Error: tmp7z7ozyzt: syntax error in line 2 near 'shape'\n"]
```

If I change the last line to `flow.visualize(filename='/var/tmp/myflow')`, the error still occurs but I can inspect the generated Dot file:
```
$ cat /var/tmp/myflow 
digraph {
	139939532296256 [label=<lambda> <map> shape=box]
	139939532296976 [label="Constant[list]" shape=ellipse]
	139939532296976 -> 139939532296256 [label=x style=dashed]
}
```

...so the issue is that `label=<lambda> <map>` needs to be quoted, i.e. should be `label="<lambda> <map>"`.

## Environment

```
$ prefect diagnostics
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Linux-5.16.11-76051611-generic-x86_64-with-glibc2.34",
    "prefect_backend": "cloud",
    "prefect_version": "1.1.0",
    "python_version": "3.8.9"
  }
}
```

bayersglassey-zesty:
I'm going to see if I can get a fix together. I followed this page: https://docs.prefect.io/core/development/overview.html
It doesn't mention a specific Python version, and it doesn't look like Tox is being used, and setup.py says `python_requires=">=3.7"`. So I chose Python 3.8:
```python
python3.8 -m venv ~/.venvs/prefect
. ~/.venvs/prefect/bin/activate
pip install -U pip
pip install -e '.[dev]'
```

Then I ran the tests with `pytest`, following this page: https://docs.prefect.io/core/development/tests.html#writing-tests
However, I get 14 errors:
```
12 failed, 6078 passed, 85 skipped, 2 xpassed, 239 warnings in 270.75s (0:04:30)
```

It looks like all 14 errors follow this pattern:
```
        runner = CliRunner()
[...]

PR #5662, opened 2022-04-09
Fix flow visualization when tasks are mapped lambdas (#5656)

bayersglassey-zesty:
## Summary

See: https://github.com/PrefectHQ/prefect/issues/5656
In short, this PR fixes an issue where a mapped lambda, e.g. `task(lambda x: x + 1).map([1, 2, 3])`, causes `graphviz.Digraph.render` (called by `Flow.visualize`) to generate DOT syntax with the label unquoted, e.g. `[label=<lambda> <map>]`, because it assumes that a string enclosed in angle brackets should contain valid HTML; and the resulting syntax is incorrect, so the `dot` command correctly rejects it, leading to an error.

Quick way to check whether the changes in PR fix the issue or not:
```python
with Flow('test') as flow:
    xs = task(lambda x: x + 1).map([1, 2, 3])
flow.visualize()
```

...after this PR, that does not result in an error.

## Changes

In Flow.visualize, we now wrap node labels with `graphviz.quoting.nohtml`, which tells graphviz to quote the labels even if they look vaguely HTML-like, e.g. `'<lambda> <map>'`.

## Importance

Mapped lambdas are something new users will likely play with, e.g. if they follow this page: https://docs.prefect.io/core/concepts/mapping.html
```python
numbers = [1, 2, 3]
map_fn = task(lambda x: x + 1)
reduce_fn = task(lambda x: sum(x))

with Flow('Map Reduce') as flow:
    mapped_result = map_fn.map(numbers)
    reduced_result = reduce_fn(mapped_result)
```

...in this case, without the fix in this PR, calling `flow.visualize()` (which a new user may reasonably be expected to do) will fail.

## Checklist

This PR:

Note: I assume this small fix does not require a change file or updated docstrings. Please let me know if that is not the case.

### DIFF
--- a/changes/pr5662.yaml
+++ b/changes/pr5662.yaml
@@ -1,0 +1,5 @@
+fix:
+  - 'Fixes graphviz syntax error when visualizing a flow with a task which is a mapped lambda. [#5662](https://github.com/PrefectHQ/prefect/pull/5662)'
+
+contributor:
+  - '[Ben Ayers-Glassey](https://github.com/bayersglassey-zesty/)'
--- a/src/prefect/core/flow.py
+++ b/src/prefect/core/flow.py
@@ -1363,7 +1363,12 @@ class Flow:
         for t in self.tasks:
             is_mapped = any(edge.mapped for edge in self.edges_to(t))
             shape = "box" if is_mapped else "ellipse"
-            name = "{} <map>".format(t.name) if is_mapped else t.name
+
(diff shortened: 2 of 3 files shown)


madkinsz:
Thanks for the PR @bayersglassey-zesty — this looks like a great fix. Can you add the change entry? You can get credit for contributing and this is a nice fix to list on release.

madkinsz (approved):

bayersglassey-zesty:
Thanks! :)

Issue #5578, opened 2022-03-19
Allow service account to be specified for `KubernetesFlowRunner`


tekumara:
## Description

Allow service account to be specified for `KubernetesFlowRunner`. This is currently possible in Prefect 1 with `KubernetesRun`

PR #5690, opened 2022-04-18
Adding service_account option to KuberentesFlowRunner

albarrentine:
## Summary

Hey :wave:, this PR is related to #5578, and adds the ability in Prefect 2.0 to specify a service account when creating a KubernetesFlowRunner. I'm currently running this fork of 2.0b3 on our EKS cluster and it's the last thing that was needed to start with Prefect 2.0 as our orchestrator. Feel free to close/change, just putting it in the universe.

## Changes

- adds a `service_account` attribute to `KubernetesFlowRunner`, with updated docstrings
- adds a test that it creates the correct Kubernetes job spec when specifying a service account
- while I was at it, added a test for the namespace argument, so should be a net positive in coverage

## Importance

In many/most production Kubernetes clusters, service accounts are useful in granting access to out-of-cluster resources like databases, storage, secrets, etc. (e.g. we use IRSA). This would allow a user to have, say, some service accounts with DB access, some that can access a Dask cluster for pure compute jobs, some that can do both, maybe for cost-savings there's one k8s cluster with different namespaces/service accounts to talk to staging/prod, etc. I also create different users on the database level such as those used for migrations (which have CREATE/DROP table privileges, etc.) that only run on deploy, and application users which may only have SELECT/INSERT/UPDATE or in some cases only SELECT. With service accounts, the db password can be stored in secrets manager/mounted on Kubernetes, etc. and the service account then only has access to the secret for its lower-privilege DB user. The other nice thing about using service accounts is there are only a minimal set of resources that even need to be accessible to the Prefect agents. For instance, I've taken the `prefect orion kubernetes-manifest` output, terraformed/modularized it, and separated out the Orion server to its own deployment with a service account that can talk to RDS Postgres, and there are prefect agent deployments for every work queue which only need the K8S RBAC fu and access to S3 storage for flows. With the service_account option in place on the flow runners, those agents can run many types of flows across namespaces with access to different resources.

## Checklist

This PR:

### DIFF
--- a/src/prefect/flow_runners.py
+++ b/src/prefect/flow_runners.py
@@ -767,6 +767,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
     Attributes:
         image: An optional string specifying the tag of a Docker image to use for the job.
         namespace: An optional string signifying the Kubernetes namespace to use.
+        service_account: An optional string specifying which Kubernetes service account to use.
         labels: An optional dictionary of labels to add to the job.
         image_pull_policy: The Kubernetes image pull policy to use for job containers.
         restart_policy: The Kubernetes restart policy to use for jobs.
@@ -777,6 +778,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
 
     image: str = Field(default_factory=get_prefect_image_name)
     namespace: str = "default"
+    service_account: str = "default"
     labels: Dict[str, str] = None
     image_pull_policy: KubernetesImagePullPolicy = None
     restart_policy: KubernetesRestartPolicy = KubernetesRestartPolicy.NEVER
@@ -977,6 +979,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
                                 "env": k8s_env,
                             }
                         ],
+                        "serviceAccountName": self.service_account
                     }
                 },
                 "backoff_limit": 4,
--- a/src/prefect/flow_runners.py
+++ b/src/prefect/flow_runners.py
@@ -979,7 +979,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
                                 "env": k8s_env,
                             }
                         ],
-                        "serviceAccountName": self.service_account
+                        "serviceAccountName": self.service_account,
                     }
                 },
                 "backoff_limit": 4,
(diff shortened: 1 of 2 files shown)


src/prefect/flow_runners.py
```diff
@@ -777,6 +778,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
 
     image: str = Field(default_factory=get_prefect_image_name)
     namespace: str = "default"
+    service_account: str = "default"
```
jawnsy, line 781: Please change this to `service_account_name` in order to more closely match the Kubernetes field name (`serviceAccountName`), and to avoid confusion with the deprecated `serviceAccount` field from the [PodSpec API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podspec-v1-core).

Changing the default value to "`default`" is a small behavior change over the prior behavior, so I think this should be: `service_account: str = None`, and do something similar to `image_pull_policy` where we only apply the value if it is non-null:

https://github.com/PrefectHQ/prefect/blob/a27139e2af4e72a25c0adb5fe31f796515ef7301/src/prefect/flow_runners.py#L989-L992

tests/test_flow_runners.py
```diff
@@ -1514,6 +1514,7 @@ async def test_creates_job(
                                 ],
                             }
                         ],
+                        "serviceAccountName": "default",
```
jawnsy, line 1517: With the change to make this an optional setting, the resulting job spec should be unchanged (no `serviceAccountName` by default). This helps us avoid running afoul of admission controller hooks that may want to enforce particular service accounts or use of namespace-default service accounts for whatever reason.

jawnsy (commented):
Hey there, thanks so much for your contribution! This change looks great to me, and thanks as well for adding tests. There are just a few minor things for us to resolve before we can merge.

jawnsy (approved):
Thanks for this!

madkinsz (approved):

albarrentine:
awesome, appreciate the feedback @jawnsy! In addition to being a joy to use, have found the Prefect codebase super clean and easy to navigate, propers to you and your team.

madkinsz:
Thanks! We appreciate hearing that :)

PR #5710, opened 2022-04-22
Fix collision with existing storage during deployment creation

madkinsz:
When a storage block was attached directly to a deployment specification, deployment creation would register the block with the flow and deployment name. Since block names must be unique on the server, this means that deployment specifications with a storage block can only be created once. Subsequent attempts would fail with an `ObjectAlreadyExists` error. This resolves the issue by including the flow version to reduce the chance of collision and appending a numeric suffix until a collision no longer occurs.


desertaxle (commented):
This looks good, but I wonder if there's a more efficient way that we could handle this. Could we append a timestamp or hash to the block name instead to avoid naming collisions?

Issue #5485, opened 2022-02-27
Mapped tasks trigger multiple times on GKE Autopilot


anyweez:
Hopefully I'm not conflating multiple things here, but I'm in quite a mess when I try to run one of my flows in a GKE Autopilot cluster.

## Description
Mapped tasks appear to (sometimes) trigger multiple times on GKE Autopilot. This puts flow runs into unstable states -- sometimes completing successfully, sometimes completing but reporting a failure, and sometimes the flow run itself does not complete due to failure.

In production, my tasks may run for multiple hours and I've observed 3-4 version of a child task kick off throughout the lifecycle of the original instance (including after the original instance has succeeded 🤯 ) despite no indication that previous attempts have failed.  I've [attached a log](https://github.com/PrefectHQ/prefect/files/8148045/prefect-multitrigger.txt) from a child task for the minimal example below. A few observations:

* In this instance, all `slow_run` children triggered more than once. In some cases this only appears to affect certain children, and *occasionally* I don't observe the issue (this is rare). This makes me think we're dealing with a race condition of some sort, potentially related to autoscaling in GKE.
* After being started three times, the first instance of `slow_run[0]` was actually killed by ZombieKiller and marked as a failed run for a few minutes. It then was changed to `Success` after the final instance of `slow_run[0]` completed (this one ran for 12 minutes, as shown in logs).
* Child tasks 0 and 2 of `report_results` both print out "Final value is None" which suggests the result of `slow_run` isn't passed down successfully.

## Expected Behavior
Mapped tasks should run through the expected [state transitions](https://docs.prefect.io/core/concepts/states.html#flow-state-transitions). Child tasks shouldn't re-trigger unless something happens to previous invocations, e.g. failures requiring retries.

I would expect the fact that I'm on GKE Autopilot (which includes autoscaling) to not affect the state transitions for child tasks.

## Reproduction

```python
from prefect import Flow

def share_state_transition(task, old_state, new_state):
    logger = prefect.context.get('logger')
    logger.info("\nState transition for {0}: {1} -> {2} (retrying={3})\n".format(
        task, 
        old_state, 
        new_state,
        new_state.is_retrying(),
    ))

    return new_state

@prefect.task
def get_list():
    return [1, 2, 3, 4, 5]

@prefect.task(state_handlers=[share_state_transition,],)
def slow_run(num):
    logger = prefect.context.get('logger')

    num_minutes = random.randint(4, 12)
    logger.info(
[...]

PR #5715, opened 2022-04-25
Fix: #5485 - Prevent task precomputation 

BitTheByte:
## Summary
Fix: https://github.com/PrefectHQ/prefect/issues/5485

## Changes
This is a proof of concept, I fully tested it but it needs expert to take a look too 

## Importance
Prevent task recomputation

## Checklist

This PR:

- [ x ] adds new tests (if appropriate)
- [ x ] adds a change file in the `changes/` directory (if appropriate)
- [ x ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

### DIFF
--- a/src/prefect/engine/result/base.py
+++ b/src/prefect/engine/result/base.py
@@ -21,12 +21,14 @@ _absence_ of computation / data.  This is in contrast to a `Result` whose value
 is `None`.
 """
 import copy
+import types
 import pendulum
 import uuid
 from typing import Any
 
 from prefect.engine.serializers import PickleSerializer, Serializer
 from prefect.utilities import logging
+from hashlib import md5
 
 
 # Subclass of `NotImplementedError` to make it easier to distinguish this error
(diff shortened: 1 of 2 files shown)


BitTheByte:
This fixes the problem when users don't supply their own `Result` however the logic breaks otherwise and fallback into the old broken behavior which will need some tweaks

src/prefect/engine/result/base.py
```diff
@@ -21,12 +21,14 @@
...
 import copy
+import types
 import pendulum
 import uuid
-from typing import Any
+from typing import Any, Callable
 
 from prefect.engine.serializers import PickleSerializer, Serializer
 from prefect.utilities import logging
+from hashlib import md5
```
jawnsy, line 31: Since md5 suffers from known hash collision attacks, perhaps we should consider sha256 or a different algorithm instead? I'm unsure what the support looks like, particularly for FIPS-compliant Python implementations, though.

BitTheByte: Md5 hash collisions require a carefully crafted input to achieve which I considered very unlikely to happen I also choose it because its somewhat fast and short. But surely if it would cause a problem it has to change

BitTheByte:
I testing this the last couple of days. I believe this doesn't break anything

BitTheByte:
I tested this the last couple of days. I believe this doesn't break anything

PR #5716, opened 2022-04-25
Fix error when = is present in parameter value

oscarwyatt:
## Summary
Fix error that arises when a (legitimate) equals sign is in a parameter value

## Changes
How keys and values are parsed in a CLI parameter

## Importance
In our use-case, we use [SMILES](https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system) strings 
to represent molecules. In this specification, an equals sign can represent a certain kind of bond. 
As is, passing a value with an equals sign causes an error

```
Invalid parameter option 'r-group-pattern='C=C'. Expected format 'key=value'.
```

This represents a huge problem for us and there is no easy obvious workaround.
This fixes that by only splitting on the first instance of an equals sign, where the parameter key and value are split.
I've checked that this works as expected included a test 

## Checklist

This PR:

### DIFF
--- a/changes/pr5716.yaml
+++ b/changes/pr5716.yaml
@@ -1,0 +1,5 @@
+fix:
+  - 'Fixes bug parsing a parameter that includes an equals ("=") sign [#5716](https://github.com/PrefectHQ/prefect/pull/5716)'
+
+contributor:
+  - '[oscarwyatt](https://github.com/oscarwyatt)'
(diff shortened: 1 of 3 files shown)


jawnsy (approved):
Welcome to the Prefect community, and thanks so much for contribution! This seems like a great change to me.

madkinsz (commented):
Could you also add a "fix" `changes/` entry?

Thanks for fixing this! Definitely an oversight.

oscarwyatt:
Whoops, I've added a `changes/` entry now

PR #5718, opened 2022-04-26
Add handling for null states to `get_task_run_info`

madkinsz:
Sometimes, the backend returns task run information before the
state has been populated in the database. This results in a null
state payload which throws an exception on deserialization.
Generally, it is safe to treat these null states as `Pending`.

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -6,16 +6,7 @@ import time
 import uuid
 import warnings
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    Mapping,
-    NamedTuple,
-    Optional,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, NamedTuple, Optional, Union
 from urllib.parse import urljoin, urlparse
 
 # if simplejson is installed, `requests` defaults to using it instead of json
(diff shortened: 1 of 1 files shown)


src/prefect/client/client.py
```diff
@@ -1247,17 +1238,26 @@ def get_flow_run_info(self, flow_run_id: str) -> FlowRunInfoResult:
                 version=tr.version,
                 task_id=tr.task.id,
                 task_slug=tr.task.slug,
-                state=State.deserialize(tr.serialized_state),
+                state=State.deserialize(tr.serialized_state)
+                if tr.serialized_state
+                else prefect.engine.state.Pending,
```
zangell44, line 1243: should we instantiate the state here?

madkinsz: Yep definitely!
```suggestion
                else prefect.engine.state.Pending(),
```

src/prefect/client/client.py
```diff
@@ -1247,17 +1238,26 @@ def get_flow_run_info(self, flow_run_id: str) -> FlowRunInfoResult:
...
+                if tr.serialized_state
+                else prefect.engine.state.Pending(),
             )
             for tr in result.task_runs
         ]
+
+        state = (
+            prefect.engine.state.State.deserialize(result.serialized_state)
+            if result.serialized_state
+            else prefect.engine.state.Pending
```
madkinsz, line 1251:
```suggestion
            else prefect.engine.state.Pending()
```

src/prefect/client/client.py
```diff
@@ -1483,7 +1483,12 @@ def get_task_run_info(
 
         task_run_info = result.data.get_or_create_task_run_info
 
-        state = prefect.engine.state.State.deserialize(task_run_info.serialized_state)
+        state = (
+            prefect.engine.state.State.deserialize(task_run_info.serialized_state)
+            if task_run_info.serialized_state
+            else prefect.engine.state.Pending
```
madkinsz, line 1489:
```suggestion
            else prefect.engine.state.Pending()
```

zangell44 (approved):

madkinsz:
:D I was going to write tests for this.

Issue #5570, opened 2022-03-17
RunGreatExpectationsValidation passes wrong type for run_id, causing LocalResult serialization to fail


dmcguire81:
## Description

When running a `Flow` that includes a task of type `RunGreatExpectationsValidation` in the `LocalAgent`, without using an explicit `Result` type for the `Flow`, the `Flow` registers cleanly and runs *nearly* to completion successfully. However, the serialization for the default `Result` type (`LocalResult`) at the end of an (otherwise successful) task run fails because of misuse of the `great-expectations` API in the `Task` implementation.

## Expected Behavior

The `Task` should run to completion and store the result, normally. Instead, it crashes with:

```
AttributeError: 'dict' object has no attribute 'to_json_dict'
```

## Reproduction

[Published docs](https://docs.prefect.io/api/latest/tasks/great_expectations.html#rungreatexpectationsvalidation) are sufficient to reproduce this case, copied here for convenience:

```python
from prefect import Flow, Parameter
from prefect.tasks.great_expectations import RunGreatExpectationsValidation

validation_task = RunGreatExpectationsValidation()

with Flow("ge_test") as flow:
    checkpoint_name = Parameter("checkpoint_name")
    prev_run_row_count = 100  # can be taken eg. from Prefect KV store
    validation_task(
        checkpoint_name=checkpoint_name,
        evaluation_parameters=dict(prev_run_row_count=prev_run_row_count),
    )

flow.run(parameters={"checkpoint_name": "my_checkpoint"})
```

The stacktrace implicates the `LocalResult` serialization [inside of the write method](https://github.com/PrefectHQ/prefect/blob/1.1.0/src/prefect/engine/results/local_result.py#L115):

```
  File "./env/lib/python3.8/site-packages/prefect/engine/results/local_result.py", line 115, in write
    value = self.serializer.serialize(new.value)
```

[...]

PR #5724, opened 2022-04-27
Fix defect in serialization of GE results in LocalResult

dmcguire81:
## Summary

Addresses defect documented in #5570.

## Changes

Replaces a `RunIdentifier` supplied as a `dict` with the actual class, which serializes correctly. Unfortunately, `great-expectations` does not make it easy on anyone trying to leverage their API, as many key classes serialize to *look* like instances of `dict`, e.g.:

```python
>>> from great_expectations.core import RunIdentifier
>>> run_identifier = RunIdentifier("foo")
>>> str(run_identifier)
'{\n  "run_time": "2022-04-27T23:09:25.216846+00:00",\n  "run_name": "foo"\n}'
```

## Importance
Serialization is currently broken for both `LocalResult` and `S3Result` (that I've tested). This means the following functionality *does not work*:
* Running `prefect[ge]` in the `LocalAgent`, since `LocalResult` is implied in that context, and there is no (obvious) way to override it
* Caching `prefect[ge]` results in the cloud with `S3Result` (and, likely, any other `Result` subclasses)

## Checklist

This PR:

- [ ] ~~updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)~~

### DIFF
--- a/tests/tasks/great_expectations/test_great_expectations_validation.py
+++ b/tests/tasks/great_expectations/test_great_expectations_validation.py
@@ -225,6 +225,12 @@ class TestInitialization:
         )
         results = task.run()
         assert type(results) is ValidationOperatorResult
+        serializable_results = repr(results)
+        assert "run_id" in serializable_results
+        assert "run_results" in serializable_results
+        assert "validation_operator_config" in serializable_results
+        assert "evaluation_parameters" in serializable_results
+        assert "success" in serializable_results
 
     def test_suite_name_with_batch_kwargs(self):
         task = RunGreatExpectationsValidation(
@@ -238,6 +244,12 @@ class TestInitialization:
         )
         results = task.run()
         assert type(results) is ValidationOperatorResult
+        serializable_results = repr(results)
+        assert "run_id" in serializable_results
+        assert "run_results" in serializable_results
+        assert "validation_operator_config" in serializable_results
+        assert "evaluation_parameters" in serializable_results
+        assert "success" in serializable_results
 
     def test_v2_checkpoint_api(self):
         task = RunGreatExpectationsValidation(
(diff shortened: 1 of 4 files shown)


dmcguire81:
Let me know if you want an addition to the `changes/` directory. The other steps didn't seem necessary.

dmcguire81:
@desertaxle anything else you need from me before reviewing this?

desertaxle (commented):
Thanks for your contribution @dmcguire81! 

Could you add a changes file so that we can properly attribute you for this contribution? Also, could you add a test or a check to an existing test to ensure that we continue to properly set `run_id`?

tests/tasks/great_expectations/test_great_expectations_validation.py
```diff
@@ -225,6 +225,12 @@ def test_assets_to_validate(self):
         )
         results = task.run()
         assert type(results) is ValidationOperatorResult
+        serializable_results = repr(results)
+        assert "run_id" in serializable_results
```
dmcguire81, line 229: These are pulled from the `ValidationOperatorResult` constructor.

tests/tasks/great_expectations/test_great_expectations_validation.py
```diff
@@ -254,6 +272,11 @@ def test_v3_checkpoint_api_pass(self):
         )
         results = task.run()
         assert type(results) is CheckpointResult
+        serializable_results = repr(results)
+        assert "run_id" in serializable_results
```
dmcguire81, line 276: These are pulled from the `CheckpointResult` constructor.

dmcguire81:
I'm not able to reproduce the `mypy` errors, locally.

madkinsz:
I've fixed the mypy errors in #5748 — you can ignore them :)

desertaxle (approved):

madkinsz:
Thanks for contributing!

Issue #4840, opened 2021-08-04
DockerAgent --network host doesn't work as expected


AlexanderVanEck:
## Description

When starting the `prefect agent docker start --network host` the network gets added after the container is created (just before the container is started). Then we a flow is executed through the agent the container immediately crashes because it cannot connect to `PREFECT__SERVER__HOST` when the host is set to a private prefect server. The DNS lookup of the `PREFECT__SERVER__HOST` fails with ```Max retries exceeded with url: /graphql/graphql (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f4ff56015f8>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution',))```

## Expected Behavior
 
One would expect when passing `--network host` that the container would be set to use the host network by default. However due to the use of docker-py's `connect_container_to_network` instead of adding it to host_config directly the resulting container has ` 'NetworkMode': 'default'` and `'Hostname': '1100502c3bf1'` (ID of container?) instead of NetworkMode=Host and Hostname=<Hostname>.

### What fails

1. Set up a private prefect server
2. Configure `PREFECT__SERVER__HOST` (and `PREFECT__SERVER__BACKEND`=server)
3. start `prefect agent docker start --network host`
4. Try to launch any DockerRun flow and see it fail

## Workaround

Starting the prefect DockerAgent without `--network host` but instead supply `host_config` to DockerRun configuration.

```
Flow(
    run_config=DockerRun(
        host_config={
            'network_mode': 'host',
        },
        ...
    )
)
```

This sets the host network as the default network.

## Questions

* When using `--network host` how would we instruct requests to use the host interface?
* How is `--network host` supposed to work?

## Environment

```
{
  "config_overrides": {},
  "env_vars": [
    "PREFECT__SERVER__HOST",
    "PREFECT__BACKEND"
  ],
  "system_information": {
    "platform": "Linux-5.4.0-80-generic-x86_64-with-debian-bullseye-sid",
    "prefect_backend": "server",
    "prefect_version": "0.15.3",
    "python_version": "3.6.13"
  }
}
```

AlexanderVanEck:
Just to let you know - this may also be caused when running an older version of prefect server which doesn't have DockerRun.host_config. Quite probable that this is what's causing the issue for us.

vldbnc:
[...]

Issue #2638, opened 2020-05-22
Cannot run flows if docker agent is started with the network option


ManuelMourato25:
## Description
Every time I try to run a flow using a docker agent with the network option specified, like so:
```
# Custom bridge network
prefect agent start docker -l docker-execs-test --show-flow-logs --network test-network
# Network used by Prefect services
prefect agent start docker -l docker-execs-test --show-flow-logs --network prefect-server
# Host network
prefect agent start docker -l docker-execs-test --show-flow-logs --network host
```
I get the following error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 677, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 392, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/local/lib/python3.6/http/client.py", line 1262, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/local/lib/python3.6/http/client.py", line 1308, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.6/http/client.py", line 1257, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.6/http/client.py", line 1036, in _send_outp ut
    self.send(msg)
  File "/usr/local/lib/python3.6/http/client.py", line 974, in send
    self.connect()
  File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 187, in connect
    conn = self._new_conn()
  File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 167, in _new_conn
    % (self.host, self.timeout),
urllib3.exceptions.ConnectTimeoutError: (<urllib3.connection.HTTPConnection object at 0x7fb645a37cf8>, 'Connection to host.docker.internal timed out. (connect timeout=30)')
```

## Reproduction
1. In order to reproduce, create any flow and store it in a docker storage like so:
```
# Define a flow and save it in a test_flow variable
....
task_sequence = [task1,task2]
test_flow.chain(*task_sequence)

storage = Docker(image_name="flows-storage", image_tag="v1.0",local_image=True)
storage.build()
test_flow.storage=storage
test_flow.register(build=False,labels=['docker-execs-test'])
```
The image is created with no issues.

[...]

PR #5748, opened 2022-05-03
Add inference of Docker network mode for 'host' and 'none' networks

madkinsz:
This brings our behavior in-line with the Docker CLI which will
change the network mode on created containers to 'host' or 'none'
respectively if the special network names 'host' or 'none' are used.

These networks appear to be immutably defined by Docker.

'host' networking is still not expected to work on non-Linux systems,
see the Docker documentation for more information.

Closes https://github.com/PrefectHQ/prefect/issues/4840
Closes https://github.com/PrefectHQ/prefect/issues/2638

### DIFF
--- a/src/prefect/agent/docker/agent.py
+++ b/src/prefect/agent/docker/agent.py
@@ -449,7 +449,7 @@ class DockerAgent(Agent):
 
             # Ensure that we change the network mode from bridge to host/none if using
             # the special predefined network names
-            if network == "host" or network == "none":
+            if network in ("host", "none"):
                 host_config.setdefault("network_mode", network)
 
         labels = {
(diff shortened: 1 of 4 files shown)


peytonrunyan (approved):

src/prefect/agent/docker/agent.py
```diff
@@ -442,9 +442,16 @@ def deploy_flow(self, flow_run: GraphQLResult) -> str:
...
         if self.networks:
+            network = self.networks[0]
             networking_config = self.docker_client.create_networking_config(
-                {self.networks[0]: self.docker_client.create_endpoint_config()}
+                {network: self.docker_client.create_endpoint_config()}
             )
+
+            # Ensure that we change the network mode from bridge to host/none if using
+            # the special predefined network names
+            if network == "host" or network == "none":
```
jawnsy, line 452: very nitpicky comment but writing as:

```suggestion
            if network in ("host", "none"):
```

may be a tiny bit better? this is fine though too

jawnsy (approved):

src/prefect/agent/docker/agent.py
madkinsz: Sure 🤷

Issue #5751, opened 2022-05-04
Prefect orion kuberenetes-manifest Fails to Find Image


ColeMurray:
## Description

When creating a kuberenetes manifest, the current code generates an incorrect / not publicly available docker image.

## Reproduction / Example

```prefect orion kubernetes-manifest```

Observe image 
```
    spec:
      containers:
      - name: api
        image: prefecthq/prefect:dev-python3.8
        command: ["prefect", "orion", "start", "--host", "0.0.0.0", "--log-level", "WARNING"]
        imagePullPolicy: "IfNotPresent"
        ports:
        - containerPort: 4200
      - name: agent
        image: prefecthq/prefect:dev-python3.8
        command: ["prefect", "agent", "start", "kubernetes"]
        imagePullPolicy: "IfNotPresent"
        env:
          - name: PREFECT_API_URL
            value: http://orion:4200/api

```

## Error 
The error is generated from here
https://github.com/PrefectHQ/prefect/blob/orion/src/prefect/cli/orion.py#L197
https://github.com/PrefectHQ/prefect/blob/orion/src/prefect/flow_runners.py#L92

where the default image_tag will become dev which does not exist on Dockerhub
https://hub.docker.com/r/prefecthq/prefect

## Fix

A potential fix would be to update "dev" -> "latest"

**Comment by madkinsz** [comment]
We aren't using the "latest" tag since that's being used by v1 right now. 

I presume you are on a version of the code that is not a release tag. If so, this is behaving as designed; if you are using a development version of the code the generated image tag will include "dev" which is not intended to be a published image but rather one you build yourself with `prefect dev build-image` or a manual call to Docker. For example, [we do this in CI](https://github.com/PrefectHQ/prefect/blob/orion/.github/workflows/python-tests.yaml#L68) to test flow run execution in images.

We could consider adding a warning to this command, as there is also a `prefect dev kubernetes-manifest` that is intended to be used when developing against a local cluster.

**Comment by ColeMurray** [comment]
Yes, I am using the orion branch.

Does it make sense to update the documentation on contributing to include this step?
https://orion-docs.prefect.io/contributing/overview/#kubernetes-development

After successfully building the image locally, the container appears to be erroring on a missing npm

```
Successfully installed prefect-0+unknown
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
[...]

PR #5752, opened 2022-05-04
Add No Ui Flag to Kubernetes Dev

ColeMurray:
## Summary

When creating a kubernetes manifest, the previous template did not include the no-ui option. This resulted
in the container crashing on load, as it does not have npm, which gets run as part of the UI standup process.

To overcome this, we launch orion without the UI in kube dev

## Changes

Fixes kubernetes dev manifest to not launch UI

## Importance

Containers will fail on start
https://github.com/PrefectHQ/prefect/issues/5751

## Checklist

This PR:

- [x ] adds a change file in the `changes/` directory (if appropriate)

### DIFF
--- a/src/prefect/cli/templates/kubernetes-dev.yaml
+++ b/src/prefect/cli/templates/kubernetes-dev.yaml
@@ -20,7 +20,7 @@ spec:
             - -c
             - >-
               pip install -e /opt/prefect/repo\[dev\] &&
-              prefect dev start
+              prefect dev start --no-ui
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
Thank you for contributing! Sorry I mislead you on the flag to pass :)

Just as a note, in the future we'll probably want to run the UI with hot-reloading in a separate container that does have `npm`. This would be ideal for development. However, it's very rare that UI development needs to happen with a server on K8s so I don't view this as a priority.

PR #5759, opened 2022-05-08
Add PREFECT_ORION_DATABASE_PASSWORD Setting

ColeMurray:
## Summary

Add PREFECT_ORION_DATABASE_PASSWORD Setting

To allow users to inject passwords as secrets, create setting that allows templating in the database connection url.

## Changes

This change allows users to pass the database password separately from the database url, allowing users to inject secrets from secret managers.

## Importance

Without this functionality, users will default to putting passwords in plaintext in the database connection url.

## Checklist

This PR:


ColeMurray:
Not sure what is going on here with the transient failures. It looks like something threading related, and unrelated to the actual change.

ColeMurray:
Ping on this, any changes requested?

madkinsz:
I think we generally presumed the entire database URL would be secret. Is there a reason you'd want the connection URL to be exposed but the password portion secret? The implementation makes sense to me though. We'll need to update some documentation, but that can happen separately.

@jakekaplan / @zangell44 for review.

ColeMurray:
> I think we generally presumed the entire database URL would be secret. Is there a reason you'd want the connection URL to be exposed but the password portion secret? The implementation makes sense to me though. We'll need to update some documentation, but that can happen separately.
> 

It is mostly a limitation of AWS CDK and how secrets work (Terraform likely has the same issue). Our database password is created as secret. A secret cannot be created from a Secret string within CDK (mostly to avoid leaking plain-text secrets in the underlying CFN template). Due to this, there isn't a way to construct a Secret within CDK that would be the full database URL. A potential workaround could be to use SSM parameters to store the concat of DB password and URL, but you'd end up with the secret in plaintext in the underlying CFN template, which makes it not so secret :D 

By separating out the Database Password, we're able to build the URL dynamically within the app

Context:
```
this.secretPassword = new Secret(this, 'secretPassword', {
    generateSecretString: {
        passwordLength: 30,
        excludePunctuation: true,
    },
    description: "Password for rds db"
});

this.rdsDb = new DatabaseInstance(this, 'RdsDbInstance', {
            engine: DatabaseInstanceEngine.POSTGRES,
            instanceType: props.instanceType,
            vpc: props.vpc,
            securityGroups: [props.securityGroup],
            storageEncrypted: true,
            multiAz: false,
            autoMinorVersionUpgrade: false,
            allocatedStorage: 20,
            storageType: StorageType.GP2,
            deletionProtection: true,
            iamAuthentication: true,
            port: props.port || 5432,
            publiclyAccessible: false,
            vpcSubnets: {
                subnetType: SubnetType.PUBLIC
            },
            credentials: {
                username: props.username,
                password: this.secretPassword.secretValue
            }
        });
```

```
taskDef.addContainer('ServiceContainer', {
            essential: true,
            image: ContainerImage.fromDockerImageAsset(serviceImage),
            logging: new AwsLogDriver({
                streamPrefix: 'service'
            }),
            portMappings: [{
                containerPort: this.CONTAINER_PORT
            }],
            environment: {
                DATABASE_URL: databaseUrl,
            },
            secrets: {
                DB_PASSWORD: ecsSecret.fromSecretsManager(props.dbPasswordSecret)
            },
            memoryLimitMiB: 300,
        });
```

jakekaplan (commented):
Changes look fine to me 👍 

Maybe worth noting in the docs that `PREFECT_ORION_DATABASE_CONNECTION_URL` can still contain the password and that `PREFECT_ORION_DATABASE_PASSWORD` will ONLY be used for templating the connecting setting, and not used directly

zangell44 (approved):
LGTM

madkinsz:
Thanks!

PR #5783, opened 2022-05-12
Use the local flow runner for `flow.run` by default

madkinsz:
Combined with #5782, this will resolve issues where calling `flow.run()` in a process that is executing a flow run managed by an agent will attempt to contact the API for orchestration and fail when the flow is not found. `flow.run()` should execute the flow locally here.

This is a breaking change for users that are depending on the config to change the local flow runner. I'm not sure that is ever a recommend pattern though.


madkinsz:
There's a viable workaround without this change although it is clunky. I'd rather avoid the breaking change for now.

Issue #5776, opened 2022-05-11
`prefect build` does not produce deterministic results


carpetmaker:
## Description

In some cases, `prefect build` command does not produce a deterministic output.
Consequently, it causes `prefect register` to bump up the flow version, even when **there are no flow code changes!** (as mentioned in #5777 )

## Expected Behavior

`prefect build` should produce an identical output for a particular input flow

## Actual Behavior

`prefect build` produces a different output for each of the 2 consecutive runs 

## Reproduction

Take this simple example:
```python
from prefect import task, Flow, Parameter


@task()
def task1(p1, p2):
    return p1, p2


with Flow('flow1') as flow:
    param1 = Parameter('param1')
    t1 = task1(param1, param1)
```
Save it as `flow.py`
Keep in mind that **this issue only occurs when the same parameter is passed to a particular task multiple times**.
Run this command:
```shell
prefect build -p flow.py -o flow1.json
```
And then again:
```shell
prefect build -p flow.py -o flow2.json
```

Now, compare the outputs:
```shell
diff flow1.json flow2.json
```
As you can see, the outputs differ!
The same can be reproduced with the `-m <module>` flag as well.

**NOTE**: In the above example, there is a total of 2 possible outputs, so there is a 50% chance of different output being generated.  _Nondeterminism_! Run it a few times to reproduce the results.

## Proposed solution

Output JSON object needs to be properly sorted during serialization.
It is the cause of this non-deterministic behavior.
Please take a look below in this thread to see 2 different file outputs.
One naive solution is to sort all lists in the output object before before being outputted.
But this needs more investigating...

## Environment

Versions (irrelevant for this issue):
- prefect version: 1.2.1
- python version: 3.8.x
- platform: darwin

carpetmaker:
### `flow1.json` output (formatted)
```json
{
  "flows": [
    {
      "__version__": "1.2.1",
      "edges": [
        {
          "__version__": "1.2.1",
          "downstream_task": {
            "__version__": "1.2.1",
            "slug": "task1-1"
          },
          "flattened": false,
          "key": "p1",
          "mapped": false,
          "upstream_task": {
            "__version__": "1.2.1",
            "slug": "param1"
          }
        },
        {
          "__version__": "1.2.1",
          "downstream_task": {
            "__version__": "1.2.1",
            "slug": "task1-1"
          },
          "flattened": false,
          "key": "p2",
          "mapped": false,
          "upstream_task": {
[...]

PR #5785, opened 2022-05-13
Sort by key when upstream and downstream slug are same for deterministic order.

tirkarthi:
## Summary

As the parameter is used repeatedly multiple times the upstream and downstream slug can be same. Add `key` which can be used as additional field for sorting to ensure deterministic order for serialized output and also during register.

## Changes

Add `edge.key` as additional parameter for sorting. Use empty string when `None` so that `TypeError` is not raised.

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/5776

## Checklist

The below test command for 100 runs will fail on randomly without patch.

```
# https://pypi.org/project/pytest-repeat/
pytest --count=100 -s -x tests/serialization/test_flows.py -k test_serialize_flow_sorts_by_key
```

This PR:

### DIFF
--- a/changes/pr5785.yaml
+++ b/changes/pr5785.yaml
@@ -1,0 +1,7 @@
+# An example changelog entry
+#
+fix:
+  - "Make flow build deterministic when upstream and downstream slug are same - [#5785](https://github.com/PrefectHQ/prefect/pull/5785)"
+
+contributor:
+  - "[Karthikeyan Singaravelan](https://github.com/tirkarthi)"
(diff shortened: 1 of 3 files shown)


madkinsz (commented):
Thanks for contributing! This looks like an appropriate fix to me. Can you just add a fix change entry?

tirkarthi:
Thanks @madkinsz , added a changelog entry. Please review the wording.

madkinsz (approved):

Issue #5522, opened 2022-03-09
Task `got some positional-only arguments passed as keyword arguments`


multimeric:
## Description
When we wrap a function with positional-only arguments in a task, it fails to execute with the above error

## Expected Behavior
Either the task should succeed, or if positional-only args are not supported, it should throw an error at task wrapping time (not at runtime as it does now)

## Reproduction

```python
from prefect import task, Flow

@task()
def add(a, b, /):
    return a + b

with Flow("flow") as flow:
    add(1, 2)
    flow.run()
```

Gives me:
```
[2022-03-09 17:18:28+1100] INFO - prefect.FlowRunner | Beginning Flow run for 'flow'
[2022-03-09 17:18:28+1100] INFO - prefect.TaskRunner | Task 'add': Starting task run...
[2022-03-09 17:18:28+1100] ERROR - prefect.TaskRunner | Task 'add': Exception encountered during task execution!
Traceback (most recent call last):
  File "/home/migwell/Programming/prefect/venv/lib/python3.9/site-packages/prefect/engine/task_runner.py", line 876, in get_task_run_state
    value = prefect.utilities.executors.run_task_with_timeout(
  File "/home/migwell/Programming/prefect/venv/lib/python3.9/site-packages/prefect/utilities/executors.py", line 468, in run_task_with_timeout
    return task.run(*args, **kwargs)  # type: ignore
TypeError: add() got some positional-only arguments passed as keyword arguments: 'a, b'
[2022-03-09 17:18:28+1100] INFO - prefect.TaskRunner | Task 'add': Finished task run for task with final state: 'Failed'
[2022-03-09 17:18:28+1100] INFO - prefect.FlowRunner | Flow run FAILED: some reference tasks failed.
```
## Environment

```json
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Linux-5.11.0-49-generic-x86_64-with-glibc2.33",
    "prefect_backend": "cloud",
    "prefect_version": "1.0.0+16.g93dc0fcaf",
    "python_version": "3.9.5"
  }
}
```

madkinsz:
Yeah we don't support this. Function parameters are always passed as kwargs, I think. This behavior should be resolved in v2, which has better signature handling, this is a good reminder to add tests for this new syntax there.

A definition time error sounds preferable as you've noted.

multimeric:
If this is fixed in V2, is there any need to fix the error handling?

[...]

PR #5789, opened 2022-05-14
Validate positional-only arguments in task signature.

tirkarthi:
## Summary

Validate positional-only arguments in task signature before bind to raise `TypeError` instead of exception during task run

## Changes

Provide better error message to positional only arguments

## Importance

Closes https://github.com/PrefectHQ/prefect/issues/5522 with better error message

## Checklist

This PR:

### DIFF
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -659,6 +659,18 @@ class Task(metaclass=TaskMetaclass):
             - Task: a new Task instance
         """
         new = self.copy(**(task_args or {}))
+
+        positional_args = [
+            name
+            for name, parameter in inspect.signature(new).parameters.items()
+            if parameter.kind == inspect.Parameter.POSITIONAL_ONLY
+        ]
+        if positional_args:
+            raise TypeError(
+                "Prefect passes arguments to task as keyword arguments. "
+                f"Positional-Only arguments found : {positional_args}"
+            )
+
         new.bind(
             *args, mapped=mapped, upstream_tasks=upstream_tasks, flow=flow, **kwargs
         )
--- a/src/prefect/core/task.py
+++ b/src/prefect/core/task.py
@@ -312,6 +312,7 @@ class Task(metaclass=TaskMetaclass):
     Raises:
         - TypeError: if `tags` is of type `str`
         - TypeError: if `timeout` is not of type `int`
+        - TypeError: if positional-only parameters are present in task's signature
     """
 
     def __init__(
@@ -348,6 +349,18 @@ class Task(metaclass=TaskMetaclass):
 
         self.name = name or type(self).__name__
         self.slug = slug
+
+        positional_args = [
+            name
+            for name, parameter in self.__signature__.parameters.items()
+            if parameter.kind == inspect.Parameter.POSITIONAL_ONLY
+        ]
+        if positional_args:
+            raise TypeError(
+                "Found positional-only parameters in the function signature for "
+                f"task {self.name}: {positional_args}. Prefect passes arguments "
+                "using keywords and does not support positional-only parameters."
+            )
 
         self.logger = logging.get_logger(self.name)
 
--- a/tests/core/test_flow.py
+++ b/tests/core/test_flow.py
@@ -1976,6 +1976,38 @@ class TestSerializedHash:
         assert hashes[0]  # Ensure we don't have an empty string or None
         assert len(set(hashes)) == 1
 
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8),
+        reason="Positional-Only arguments are only supported in Python 3.8+",
+    )
+    def test_task_positional_only_arguments(self, tmpdir):
+        contents = textwrap.dedent(
+            """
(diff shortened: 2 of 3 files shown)


src/prefect/core/task.py
```diff
@@ -659,6 +659,18 @@ def __call__(
...
+        positional_args = [
+            name
+            for name, parameter in inspect.signature(new).parameters.items()
+            if parameter.kind == inspect.Parameter.POSITIONAL_ONLY
+        ]
+        if positional_args:
+            raise TypeError(
+                "Prefect passes arguments to task as keyword arguments. "
+                f"Positional-Only arguments found : {positional_args}"
+            )
```
madkinsz, line 672: Could we check this at `__init__` time instead of `__call__` time so we raise the error at flow definition instead of runtime?

madkinsz: Also, cc @tpdorsey I think we'll want to compose an error message here that is consistent with our style.

I'm imagining something like

> Found positional-only argument(s) in function signature for task {task_name}: {positional_args}. 
> Prefect passes arguments using keywords and does not support positional-only arguments.

tests/core/test_flow.py
```diff
@@ -1976,6 +1976,38 @@ def dummy_task():
...
+            dummy_task()
+
+        if __name__ == "__main__":
+            flow.run()
+        """
+        )
+        script = tmpdir.join("flow.py")
+        script.write_text(contents, encoding="utf-8")
+
+        result = subprocess.run([sys.executable, str(script)], capture_output=True)
```
madkinsz, line 2002: The test above this uses a subprocess to run the flow because we're testing that a hash is consistent across Python instances. This test doesn't need a subprocess and they're at least an order of magnitude slower. Can you write the test with the flow/task definition inline and use `with pytest.raises` instead?

madkinsz (changes requested):

madkinsz:
Thanks for contributing! A few requested changes. Can you add a changelog entry? We may want to update the `Task` docstring to note that positional-only arguments are not allowed as well.

src/prefect/core/task.py
tpdorsey: My suggested refinement of your suggestion, which I think works quite well:
 
"Found positional arguments in the function signature for task {task_name}: {positional_args}.
Prefect passes arguments using keywords and does not support positional-only arguments."

madkinsz: Maybe following the PEP (https://peps.python.org/pep-0570/) we should say "positional-only" parameters. I don't think we can drop the "-only" from the first line.

tpdorsey: Agreed, follow PEP example: 

"Found positional-only arguments in the function signature for task {task_name}: {positional_args}.
Prefect passes arguments using keywords and does not support positional-only arguments."

madkinsz: The document also refers to "parameters" instead of "arguments" so we may want to use that? I don't have a strong preference though.

tpdorsey: I do like to follow the PEP example when we can. However, looking at the Python docs, "argument" and "parameter" seem to be used interchangeably.  

If there is a distinction, I perceive it as: parameter is the definition of what you can pass, argument is what you pass as a value for that parameter. Maybe.

I don't have strong feelings about it either. I think the usage pattern is to use the term that promotes readability.

tirkarthi: I have used parameters throughout. It also prefer that parameters read appropriate for signature and arguments are more appropriate for the values supplied to the function. Below is a sample error message.

```python
python
Python 3.10.4 (main, Apr  8 2022, 17:35:13) [GCC 9.4.0] on linux
>>> def foo(a, b, /): pass
... 
>>> foo(a=1)
TypeError: foo() got some positional-only arguments passed as keyword arguments: 'a'
```

```
  File "/tmp/foo.py", line 4, in <module>
    def add(a, b, /):
  File "/home/karthikeyan/stuff/python/prefect/src/prefect/utilities/tasks.py", line 406, in <lambda>
    return lambda fn: prefect.tasks.core.function.FunctionTask(
  File "/home/karthikeyan/stuff/python/prefect/src/prefect/core/task.py", line 162, in init
    old_init(self, *args, **kwargs)
  File "/home/karthikeyan/stuff/python/prefect/src/prefect/tasks/core/function.py", line 50, in __init__
    super().__init__(name=name, **kwargs)
  File "/home/karthikeyan/stuff/python/prefect/src/prefect/core/task.py", line 162, in init
    old_init(self, *args, **kwargs)
  File "/home/karthikeyan/stuff/python/prefect/src/prefect/core/task.py", line 359, in __init__
    raise TypeError(
TypeError: Found positional-only parameters in the function signature for task add: ['a', 'b']. Prefect passes arguments using keywords and does not support positional-only parameters.
```

tests/core/test_flow.py
madkinsz: Ah I see that makes a lot of sense. Can you just use `exec` instead of a subprocess then?

PR #5802, opened 2022-05-18
Allow Orion UI to run on Windows

stormljor:
## Summary
Windows defaults to the mimetype text/plain for .js files, leading to the Orion UI loading a blank page, with the following in the JavaScript console:
"Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec."

This can be mitigated by changing the mimetype in the Windows registry, but that seems a bit invasive for such a minor fix.

## Changes
This PR overrides the mime type for .js files if os.name == "nt", allowing the UI to load.

## Importance

While it's still not possible to develop the UI on Windows (all the commands in dev that run `subprocess.check_output(['npm', ...])` fail due to not checking the path by default), this at least allows the final product to run on a Windows desktop/server.

## Checklist

(I'm assuming I need to check these even if I haven't done anything, as it wasn't appropriate?)
This PR:

### DIFF
--- a/src/prefect/orion/api/server.py
+++ b/src/prefect/orion/api/server.py
@@ -3,6 +3,7 @@ Defines the Orion FastAPI app.
 """
 
 import asyncio
+import mimetypes
 import os
 import warnings
 from functools import partial
@@ -200,6 +201,11 @@ def create_orion_api(
 
 def create_ui_app(ephemeral: bool) -> FastAPI:
     ui_app = FastAPI(title=UI_TITLE)
+
+    if os.name == "nt":
+        # Windows defaults to text/plain for .js files
+        mimetypes.init()
+        mimetypes.add_type("application/javascript", ".js")
 
     @ui_app.get("/ui-settings")
     def ui_settings():


madkinsz:
Thanks for contributing!

Sorry about the confusion regarding the PR checklist, it does not apply to Orion development.

I wonder if we should be changing the content-type that the server sends for these files instead. I'll loop in some people that should know more :)

@jawnsy @cicdw @znicholasbrown

znicholasbrown (commented):
This makes sense to me - it seems like FastAPI should auto-add mime types for all static assets so we may need to look at this issue more holistically. 

I'd imagine we want to be sending the appropriate mime types regardless of the OS, however I'll open a ticket to take a look at that. I'd be comfortable merging this as-is if it unblocks Windows users 👍🏻

stormljor:
I'm not sure if it's uvicorn or FastAPI that has the "problem", but mimetypes behave strange on Windows:
```python
Python 3.9.9 (tags/v3.9.9:ccb0e6a, Nov 15 2021, 18:08:50) [MSC v.1929 64 bit (AMD64)] on win32
>>> import mimetypes
>>> mimetypes.types_map[".js"]
'application/javascript'
>>> mimetypes.init()
>>> mimetypes.types_map[".js"]
'text/plain'
```

madkinsz:
Thanks! The failing tests are resolved in our internal repo and are due to a breaking change in the latest Pydantic release.

madkinsz:
If you have the chance to determine if the issue is in the upstream or in our static files implementation that'd be great! I'd rather not include this change in the long run and we should get it fixed upstream if the issue is there.

stormljor:
Looks like it's not being fixed upstream; changing the Windows registry isn't an option in all cases: https://github.com/encode/starlette/issues/829

stormljor:
Having looked more at this (and if you'll forgive me for continuing such an old ticket), it seems that installing certain Microsoft products triggers the fault - the most common seems to be Visual Studio - which by chance was installed on all the computers I tested this on.

Maybe this is more suited for an FAQ/troubleshooting section rather than a code fix.

PR #5816, opened 2022-05-20
Make job/pod watch timeouts configurable on KubernetesFlowRunner

albarrentine:
## Summary

Hey all! This PR adds configurable timeouts for job and pod watches respectively in the KubernetesFlowRunner, which should help in environments like Fargate that may need longer timeouts. Putting this up for discussion, happy to implement it a different way or squash if this is already handled somewhere else (noticed these were marked with TODOs already).

## Changes

Adds two configuration options to the KubernetesFlowRunner:

- job_watch_timeout_seconds: replaces the hard-coded 5 second timeout in watching for Job creation in K8s with a configurable option
- pod_watch_timeout_seconds: ditto for Pod creation

## Importance

Currently we're running flows on Fargate with Orion/agents deployed on EC2 (all in EKS)~~, and while a flow may run successfully with all tasks completing, the job completes in Kubernetes, etc. the flow run still gets marked as Crashed.~~ As far as I can glean from the following log output on the agents, ~~it appears to be related to~~ there is an error caused by a hard-coded 5 second timeout in watching for pod creation (may be mistaken, could be a symptom rather than a cause if, say, there's some special RBAC needed beyond what's in `prefect orion kube-manifest`):

```
06:50:09.903 | INFO    | prefect.flow_runner.kubernetes - Flow run job 'strict-spidert6rzb' has status {'active': None,
 'completed_indexes': None,
 'completion_time': None,
 'conditions': None,
 'failed': None,
 'ready': None,
 'start_time': None,
 'succeeded': None,
 'uncounted_terminated_pods': None}
06:50:09.903 | INFO    | prefect.flow_runner.kubernetes - Starting watch for pod to start. Job: strict-spidert6rzb
06:50:14.922 | ERROR   | prefect.flow_runner.kubernetes - Pod never started. Job: strict-spidert6rzb
```

~~The job ran for several hours, all tasks show succeeded in the UI, logs are showing up fine, it stayed in the Running state, but after all tasks finished successfully, it was marked as Crashed and this was the only error I could find. The Orion logs then show this:~~

~~09:35:37 AM Crash detected! Execution was interrupted by an unexpected exception.
09:39:01 AM Using task runner 'ConcurrentTaskRunner'
09:39:01 AM Crash detected! Execution was interrupted by an unexpected exception.~~

~~It spins up another pod, presumably because the first one is marked as never having started, then that pod's logs say the same as above but with one additional line:~~

~~13:39:01.491 | INFO    | Flow run 'strict-spider' - Using task runner 'ConcurrentTaskRunner'
13:39:01.570 | ERROR   | Flow run 'strict-spider' - Crash detected! Execution was interrupted by an unexpected exception.
13:39:01.586 | INFO    | prefect.engine - Engine execution of flow run 'dd782509-5144-4b70-8cbe-a5d7157480ce' aborted by orchestrator: This run has already terminated.~~

In the case of an environment like Fargate, the job will get created within the 5 seconds, but the pod may take a while to start (it will be in a status of Pending) since the cluster spins up an independent node for each pod, which for our region seems to take 30-60 seconds. It would be useful to have the ability to configure a timeout for a longer period on the pod, maybe 5 minutes or so to make sure it's pretty much surely not going to start. I included a job wait timeout as well, for completeness and because it was also marked as a TODO, though here the default is probably reasonable (there may be some other condition to monitor there since the job can start but hits the backoff limit and then still stays as Pending in Orion - happened to us at first because Fargate pods did not have access to CoreDNS).

This was on a production workflow that I wanted to move to Prefect and Fargate which connects to a database, reads/writes to S3, uses Secrets Manager, runs in a container with internal libraries/dependencies installed, downloads from the web, runs some ETL, has a least-permission IAM role, etc. Everything's working beautifully otherwise.

## Checklist

This PR:

### DIFF
--- a/src/prefect/flow_runners.py
+++ b/src/prefect/flow_runners.py
@@ -796,6 +796,8 @@ class KubernetesFlowRunner(UniversalFlowRunner):
     service_account_name: str = None
     labels: Dict[str, str] = None
     image_pull_policy: KubernetesImagePullPolicy = None
+    job_watch_timeout_seconds: int = 5
+    pod_watch_timeout_seconds: int = 5
     restart_policy: KubernetesRestartPolicy = KubernetesRestartPolicy.NEVER
     stream_output: bool = True
 
@@ -872,7 +874,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
             func=self.client.list_namespaced_pod,
             namespace=self.namespace,
             label_selector=f"job-name={job_name}",
-            timeout_seconds=5,  # TODO: Make configurable?
+            timeout_seconds=self.pod_watch_timeout_seconds,
         ):
             if event["object"].status.phase == "Running":
                 watch.stop()
@@ -905,7 +907,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
             func=self.batch_client.list_namespaced_job,
             field_selector=f"metadata.name={job_name}",
             namespace=self.namespace,
-            timeout_seconds=5,  # TODO: Make configurable?
+            timeout_seconds=self.job_watch_timeout_seconds,
         ):
             if event["object"].status.completion_time:
                 watch.stop()
--- a/src/prefect/flow_runners.py
+++ b/src/prefect/flow_runners.py
@@ -785,6 +785,8 @@ class KubernetesFlowRunner(UniversalFlowRunner):
         service_account_name: An optional string specifying which Kubernetes service account to use.
         labels: An optional dictionary of labels to add to the job.
         image_pull_policy: The Kubernetes image pull policy to use for job containers.
+        job_watch_timeout_seconds: Number of seconds to watch for job creation before timing out (default 5).
+        pod_watch_timeout_seconds: Number of seconds to watch for pod creation before timing out (default 5).
         restart_policy: The Kubernetes restart policy to use for jobs.
         stream_output: If set, stream output from the container to local standard output.
     """


albarrentine:
Update: the Crashed status was not related to the timeout (didn't feel like it was), found something in the pod log that was not getting tracked by Prefect and was a simple permissions issue with uploading to S3 storage after the flow completed. Timeout issue still applies to running jobs on Fargate.

For a few other reasons (will need the ability to clean up job pods on completion so it doesn't leave extra instances running, as well as the ability to specify more resources above the default of 0.25 VCPUs/0.5GB of RAM), I'll hold off on Fargate for regular jobs at the moment, can just use it for ephemeral Dask clusters, which is probably a better use case anyway. Would still like to run some jobs on Fargate when those changes are in place so will leave the PR up for discussion.

chrisguidry:
Hi @albarrentine, thanks for the PR!  We've reorganized `flow_runners.py` into separate files under the `flow_runners/` directory, sorry for moving things out from under you here.

https://github.com/PrefectHQ/prefect/blob/orion/src/prefect/flow_runners/kubernetes.py#L63

I'll make the corresponding changes in our upstream repo so that they are in 2.0b7 if that works for you.  You can also update this PR and we'll make sure it gets merged before the next release.  Just let us know!

chrisguidry:
Just letting you know that I've made these changes in our upstream repo, so they'll definitely be in 2.0b7 merge, with you listed as the contributor.  Thanks again for the PR!

albarrentine:
thanks @chrisguidry that makes sense to split up the modules. Upgrading our to 2.0b6 this evening and will look out for the latest and greatest

PR #5819, opened 2022-05-23
Allow custom job name from passed job template

crflynn:
## Summary

This allows custom job name prefix using the job template passed to the KubernetesAgent.

## Changes

The current code in `KubernetesAgent.generate_job_spec` forces a job name prefix of `prefect-job`. This change  means that the job template's `metadata.name` (if it exists) will be used as the job name prefix. If not, it will fall back to `prefect-job`

## Importance

This PR allows users to name jobs so that they may be more easily identifiable.

## Checklist

This PR:

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -479,7 +479,8 @@ class KubernetesAgent(Agent):
 
         identifier = uuid.uuid4().hex[:8]
 
-        job_name = f"prefect-job-{identifier}"
+        job_name = _get_or_create(job, "metadata.name", "prefect-job")
+        job_name = f"{job_name}-{identifier}"
 
         # Populate job metadata for identification
         k8s_labels = {
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -480,6 +480,8 @@ class KubernetesAgent(Agent):
         identifier = uuid.uuid4().hex[:8]
 
         job_name = _get_or_create(job, "metadata.name", "prefect-job")
+        # 63 max length - 9 identifier - 6 pod identifier
+        job_name = job_name[:48]
         job_name = f"{job_name}-{identifier}"
 
         # Populate job metadata for identification
--- a/changes/pr5819.yaml
+++ b/changes/pr5819.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Allow custom job name from passed job template - [#5819](https://github.com/PrefectHQ/prefect/pull/5819)"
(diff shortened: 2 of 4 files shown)


madkinsz (commented):
@crflynn Thanks for contributing! Can you add a changes/ entry?

src/prefect/agent/kubernetes/agent.py
```diff
@@ -479,7 +479,8 @@ def generate_job_spec(self, flow_run: GraphQLResult) -> dict:
 
         identifier = uuid.uuid4().hex[:8]
 
-        job_name = f"prefect-job-{identifier}"
+        job_name = _get_or_create(job, "metadata.name", "prefect-job")
```
jawnsy, line 482: If a user passes a `metadata.name` for a job, would they expect that the job name is used as-is? I suppose we have (at least) a few options here:

1. If a user passes a name, use it as-is (without any suffix) and if there is no name, then use the existing auto-generated `prefect-job-xxx` name.
2. We accept a "name_prefix" parameter when creating the job spec, which defaults to `prefect-job` and can be customized by the user (we'd append the identifier in this case, similar to how this PR behaves). Having this makes it clear that the name is a prefix for the job, but has the downside that we have to deal with the case where both a `metadata.name` and `name_prefix` are supplied. We can resolve that by using one preferentially to the other in that case.

Overall, I think I would find it surprising if I provided a job spec and the name was mutated in a way where I also could not disable or override that behavior.

madkinsz: We'll check with more of our team and see what makes the most sense.

Do you happen to have examples of the other schedulers that use this behavior? That'd be helpful.

crflynn: The other ones I've used are Airflow and Argo-workflows and IIRC they both add suffixes to jobs like prefect already does here.

crflynn: One other consideration here might be some defensive logic around the length of the identifier according to k8s resource name limits.

madkinsz: Do those pull the `metadata.name` from the job specification as the prefix though?

crflynn: * [yes](https://github.com/apache/airflow/blob/1bf6dded9a5dcc22238b8943028b08741e36dfe5/airflow/kubernetes/pod_generator.py#L126) for airflow
* [yes](https://github.com/argoproj/argo-workflows/blob/7e34ac5138551f0ebe0ca13ebfb4ad1fc8553ef1/workflow/common/convert.go#L37) for argo, although argo uses a CRD for its workflows but it pulls from the workflow name.

crflynn:
Any movement on this?

madkinsz (approved):
Hey! I had some time to chat with more of the team about this and we're willing to accept it. Thank you for your patience! Can you add a brief blurb documenting how this works? Perhaps in https://docs.prefect.io/orchestration/agents/kubernetes.html#custom-job-template

crflynn:
I will add something, sure. I also think we need to add some code which trims the name if it's too long. Not sure what the limits are but I'll add that as well.

PR #5823, opened 2022-05-24
Make 429 rate limit backoff configurable

bnaul:
## Summary

When a 429 rate limit response is encountered, the client retries+backs off repeatedly until the response succeeds or some maximum number of failures is reached. The original configuration from #4852 starts out by sleeping for between [180, 190] seconds, then [180, 200], etc, up to [180, 820].

For a typical (for us 🙂) case where a 429 occurs because of a high volume of task runs all starting at once, retrying every single task in 3 minutes is not a very good idea, as it almost guarantees that the exact same rate limit will be hit again, and probably again and again until the window gets wide enough to spread out the requests over several minutes at which point they will hopefully succeed.

## Changes

Adds a `config.cloud.rate_limiting` variable where the default parameters can be overridden, to allow for longer/shorter backoffs as desired.

I think I would also advocate for changing the defaults as well, the initial [180, 190] backoff seems like a recipe for a bad time in almost any case. If the extra config parameters feel too noisy/cluttered I could maybe imagine that just tweaking the defaults could be good enough for most cases; but the reasoning behind making them configurable was so we could have our flow runners / agents retry more times and more frequently, but also have our thousands of workers retry much more gradually to avoid things getting too backed up.

cc @cicdw @madkinsz @hardisty

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -532,7 +532,7 @@ class Client:
                 * (prefect.config.cloud.rate_limiting.backoff_exponent ** rate_limit_counter)
             )
             naptime = prefect.config.cloud.rate_limiting.backoff_s + jitter
-            self.logger.debug(
+            self.logger.warning(
                 f"Rate limit encountered (attempt {rate_limit_counter}); sleeping for {naptime}s..."
             )
             time.sleep(naptime)
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -517,16 +517,24 @@ class Client:
             )
 
         # custom logic when encountering an API rate limit:
-        # each time we encounter a rate limit, we sleep for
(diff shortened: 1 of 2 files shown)


madkinsz:
My instinct is that we'd rather just hardcode an increase to jitter on the first few retries than allow full configuration of the back-off, but I'll defer to Chris here.

bnaul:
Definitely can see that, the main issue is that having an agent or flow runner back off for 10-20 minutes is quite annoying and things can start to pile up for other reasons. I'm not sure that there is a single backoff pattern that makes sense for both the workers and the general Prefect machinery.

I guess another approach might be to detect whether this is in a task run context or not, and adjust the backoff time based on that?

madkinsz (approved):
Once this passes CI we can get this merged.

PR #5788, opened 2022-05-13
Feature/orion k8s extra configs

mkarbo:
## Summary
This PR implements additional ad-hoc configuration options for kubernetes flow runner, following the pattern of dask-kubernetes (https://kubernetes.dask.org/en/latest/_modules/dask_kubernetes/common/objects.html#make_pod_spec)

For instance, we need to set imagePullSecrets variable to pull from container registry. This PR will accomodate most future configuration requirements for k8s flow runner without having to implement them 1-by-1. 

## Changes
It allows for deeper customization of k8s flow runner.

## Importance
It helps accomodate un-anticipated requirements for deployments using kubernetes flow runner, such as imagepullsecrets or other implementation/case specific requirements.

## Checklist

This PR:

- [x] adds new tests (if appropriate)
- [x] updates docstrings for any new functions or function arguments,

PR #5834, opened 2022-05-25
Add k8s secrets support to Orion KubernetesFlowRunner

codref:
## Summary
This allows the configuration of k8s secrets as environment variables in the `KubernetesFlowRunner` class.  
Example of usage would be:  

```python
DeploymentSpec(
    name="demo-flow-deployment-with-secret",
    flow=demo_flow,
    flow_runner=KubernetesFlowRunner(
        secrets={"vault": "VAULT_TOKEN"},
        env={"qwe": "bbb"},
        namespace="prefect"
    )
)
```

Inside the task the secret is read as an environment variable:

```python
@task
def log_env_variable():
    logger = get_run_logger()
    logger.info(os.getenv('VAULT_TOKEN'))    
```

`secrets` property is optional but if configured the secret must exists in the namespace.

## Changes

Adds the `secrets` property to the class `KubernetesFlowRunner` as a `Dict[str, str]`.  
Cycle the new dictionary and pushes a dict to the `k8s_env` array with the structure:  

```python
                "name": v,
                "valueFrom": {
                    "secretKeyRef": {
                        "name": k,
                        "key": v,
                        "optional": False
                    }
                }
```

The structure is copied from https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables  

Secret is configured as non-optional due to the expected dependency of the flow/tasks logic to the configured variable.  

Added method `test_use_secrets` to `tests/test_flow_runners.py` which verify that the env array contains the expected secret when the KubernetesFlowRunner is created with *secrets*.

## Importance

In an enterprise environment we can expect at least some basic credentials management. This would allow at least a simple and effective use of tokens inside the flow logic.  
First use case could be some Vault integration - one secret to fetch many.  
Without this feature, the only workaround would be to pass the token in an `env` property, compromising the security by design.  
This method is still considered effective only on a trusted environment - as the user might push some code to print out the secret on a log and read it but at least it mitigates the accidental disclosure (e.g. centralized logging accessed by a malicious user).  

## Checklist

This PR:

### DIFF
--- a/src/prefect/flow_runners.py
+++ b/src/prefect/flow_runners.py
@@ -787,6 +787,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
         image_pull_policy: The Kubernetes image pull policy to use for job containers.
         restart_policy: The Kubernetes restart policy to use for jobs.
         stream_output: If set, stream output from the container to local standard output.
+        secrets: An optional dictionary of Kubernetes secret names and the relevant keys to be mounted as environment variables.
     """
 
     typename: Literal["kubernetes"] = "kubernetes"
@@ -798,6 +799,7 @@ class KubernetesFlowRunner(UniversalFlowRunner):
     image_pull_policy: KubernetesImagePullPolicy = None
     restart_policy: KubernetesRestartPolicy = KubernetesRestartPolicy.NEVER
     stream_output: bool = True
+    secrets: Dict[str, str] = Field(default_factory=dict)
 
     _client: "CoreV1Api" = PrivateAttr(None)
     _batch_client: "BatchV1Api" = PrivateAttr(None)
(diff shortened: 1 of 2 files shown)


mkarbo:
see also reply from @madkinsz at [this pr](https://github.com/PrefectHQ/prefect/pull/5788) that they are actively on the k8s flow runner - we also needed this feature

madkinsz:
We've got a new way to pass customizations to the K8s job coming in 2.0b7. We'll want to see if that satisfies this use-case before looking into a simpler approach (as implemented here) for users unfamiliar with the details of Kubernetes.

codref:
That's a great news, looking forward to see the changes pushed on b7 release!

madkinsz:
We now have secret support. I think we'd still be interested in exposing this shortcut for environment variable secrets, but we'll want to see use-cases for a bit longer first. For this implementation, we'll want to find a name that better reflects that you're specifying environment variables to pull from secrets.

Issue #5842, opened 2022-05-27
Secrets via environmental variables - $ expansion in passwords should be skippable


taljaards:
## Description

<details>
  <summary>From Slack</summary>
  
- https://prefect-community.slack.com/archives/CL09KU1K7/p1653561935947319
- ![image](https://user-images.githubusercontent.com/6089256/170636144-f68aa430-977d-4ec9-a8cf-c09d14d753a8.png)
  
</details>

Environmental variables are expanded by this line https://github.com/PrefectHQ/prefect/blob/ad734292bc4113a52777d6ebe87cfa1c13a3eee8/src/prefect/configuration.py#L83
This causes an edge case for passwords containing `$`. The code wants to expand it, but it shouldn't be expanded since it's a password.

It would be nice to prevent the expansion somehow. From the end-user's side, they can save the password URL encoded, but that feels like a gotcha.

## Expected Behavior
See below.

## Reproduction
```
import os
from prefect.client import Secret

secret_name = 'JSON_SECRET'
# os.environ[
#     f"PREFECT__CONTEXT__SECRETS__{secret_name}"
# ] = '{"db1": {"password": "password", "user": "userA"}, "db2": {"password": "P@$$w0rd", "user": "userB"}}'

secret_value = Secret(secret_name).get()
print(secret_value['db2']['password'])
# This prints 'P@$w0rd', not 'P@$$w0rd'
```

## Environment
<details>
  <summary>prefect diagnostics</summary>
  
```sh
(.venv) PS C:\Users\staljaard\Source\data-integration> prefect diagnostics
{
  "config_overrides": {      
    "backend": true,
    "cloud": {
      "request_timeout": true
    },
    "server": {
      "database": {
        "password": true
      }
    }
  },
  "env_vars": [
    "PREFECT__CONTEXT__SECRETS__JSON_SECRET"
  ],
  "system_information": {
    "platform": "Windows-10-10.0.19041-SP0",
    "prefect_backend": "server",
    "prefect_version": "1.2.0",
    "python_version": "3.7.9"
  }
}
```
  
</details>

PR #5843, opened 2022-05-27
Allow escaping dollar signs in environmental variables to prevent expansion for Secrets

taljaards:
## Summary
In environmental variables - allows escaping `$` with `\$` to prevent interpolation.
This is an idea to address #5842 - please check; I'm unsure of the side effects of this suggestion.

## Changes
Changes the way interpolation is done. Before interpolation, `\$` is replaced with a placeholder `§` to prevent interpolation. Afterwards, the `$` is inserted back.

## Importance
This allows an easy way of saving `$` in environmental variables, for example in passwords.
The user experience should hopefully be a bit easier/straightforward. The alternative is to save a URL-encoded string inside the env var, which is more work and less straightforward.

## Checklist
This PR:


madkinsz:
I'm very hesitant to make a low-level change in v1 like this since its side-effects could be serious and we're trying to be very stable. Feel free to chime in on the issue at #5842 if you're encountering this issue. @taljaards if you're interested in documenting a workaround in the issue that'd be most helpful!

PR #5846, opened 2022-05-30
fix raising error in __init__.py for snowflake task

thesubneo:
## Summary
Importing SnowflakeQuery fails when Prefect was not installed with the "snowflake" extra but with wrong exception.

## Changes
Fix missing 'as err' so the right exceptions is raised instead of python's syntax exception

## Checklist

This PR:


anna-geller (approved):
Nice, thanks for contributing this fix!

desertaxle (approved):

PR #5872, opened 2022-06-05
fix: match DbtShellTask.run return type to ShellTask.run

tekumara:
When creating a DbtShellTask with `return_all=True` the return type of `run` is `List[str]` but this is not reflected in the return type.

This PR updates the return type to match the [return type of `run`](https://github.com/PrefectHQ/prefect/blob/de5bdcf/src/prefect/tasks/shell.py#L93) in the super class (ShellTask).

Signed-off-by: Oliver Mannion <125105+tekumara@users.noreply.github.com>

### DIFF
--- a/src/prefect/tasks/dbt/dbt.py
+++ b/src/prefect/tasks/dbt/dbt.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any
+from typing import Any, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


tekumara:
FYI - the test failures looks unrelated to this change?

desertaxle (approved):

PR #5899, opened 2022-06-13
fix: add orion exports 

tekumara:
## Summary

Declares orion's public API for export.

## Changes

Adds the `_all_` attribute to  `__init__.py` for all orion packages . See #5293 which is the same change applied to Prefect 1.

## Importance

Fixes "is not exported from module" errors in vscode/pyright/mypy strict mode in code that imports the prefect API.

eg:

```
$ pyright
  /Users/tekumara/code/orion-demo/flows/kubes_flow.py:1:21 - error: "flow" is not exported from module "prefect"
    Import from "prefect.flows" instead (reportPrivateImportUsage)
  /Users/tekumara/code/orion-demo/flows/kubes_flow.py:1:27 - error: "get_run_logger" is not exported from module "prefect"
    Import from "prefect.logging" instead (reportPrivateImportUsage)
  /Users/tekumara/code/orion-demo/flows/kubes_flow.py:4:34 - error: "KubernetesFlowRunner" is not exported from module "prefect.flow_runners"
    Import from "prefect.flow_runners.kubernetes" instead (reportPrivateImportUsage)
```

## Checklist

This PR:

- [NA] adds new tests (if appropriate)
- [NA] adds a change file in the `changes/` directory (if appropriate)
- [NA] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)


madkinsz:
Thanks for contributing! I'm confused about this though... we are using the relative import syntax specified in https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface to expose these objects.

tekumara:
I see `__init__.py` has the form:
```
from .orion.schemas.states import State
```

From the [pyright docs](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface):
> uses the form “from .A import X”, symbol “A” is not private unless the name begins with an underscore (but “X” is still private)

So I think this means State is still private.

Whereas
```
from .orion.schemas.states import State as State
```
would make `State` public, if you prefer that form.

madkinsz:
Ah gee I've misread that it seems. Thanks for pointing that out. Let me see how the rest of the team feels about the redundant alias vs maintaining an `__all__` list.

abrookins:
@madkinsz I'm in favor. @bunchesofdonald how do you feel about it?

madkinsz:
I prefer the redundant alias since it's not separate from the import. `__all__` is a more common pattern but the suggested usage in that document is to reduce the exposed symbols when using wild-card imports.

PR #5946, opened 2022-06-27
Add retries to fetching secrets

cjmartian:
## Summary

This PR adds an optional retry when attempting to fetch a secret. When retrying there is a small delay implemented, which is also a parameter available to the user.

This was motivated by flows failing after trying to fetch a secret, knowing that they are present in the secret store. The hope is that a retry will help alleviate the issue of flows failing when there are intermediate connectivity issues. 

## Changes

This will add two new parameters to `Secret.get()`; `retries` and `delay`.

`retries` is an optional `int` that specifies the number of retries to attempt before raising an issue. The default is 0, meaning there will be no retries and if it fails to fetch the secret for any reason, it will raise an exception. It's worth noting that we only retry on backend fetches of secrets.

`delay` is an optional `Union[float, int]` that specifies the amount of time, in seconds, we should wait before attempting to retry the `Secret.get()` operation. The default is `0.5` currently, but this can be set by the user. Again, this only happens when we are using the backend secret store. 

## Importance

Sometimes intermediate connection errors happen, we should be able to retry when they do. 

## Checklist

This PR:

### DIFF
--- a/src/prefect/client/secrets.py
+++ b/src/prefect/client/secrets.py
@@ -61,7 +61,8 @@ references to: `..../toml/decoder.py`.
 """
 
 import json
-from typing import Any, Optional
+from time import sleep
+from typing import Any, Optional, Union
 
 import prefect
 from prefect.client.client import Client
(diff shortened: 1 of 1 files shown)


cjmartian:
This was fixed elsewhere. Closing as requested [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1657220518433109?thread_ts=1656623472.005169&cid=CL09KU1K7).

PR #5957, opened 2022-07-03
Avoid croniter infinite loop

jlowin:
## Summary

We thought we had identified all the weird DST issues with croniter, but a user recently stumbled across a new one. A simplified demo:

```python
import pendulum
from croniter import croniter
from datetime import datetime

cron = croniter('0 0 * * 0', pendulum.datetime(2022, 8, 21, tz='America/Santiago'))
for i in range(10):
    print(cron.next(datetime))
```
Which results in
```
2022-08-28 00:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
2022-09-03 23:00:00-04:00
```

To avoid this (and others potentially lurking), this adds a check to see if we're stuck in a date loop when evaluating cron. If so, we exit the loop by creating a new cron instance that begins yielding immediately AFTER the date question. 

## Changes

Fixes a bug where certain DST conditions put schedules into an infinite loop

## Checklist

This PR:

### DIFF
--- a/changes/PR5957.yaml
+++ b/changes/PR5957.yaml
@@ -1,0 +1,21 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - task
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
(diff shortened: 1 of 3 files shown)


jawnsy (approved):
LGTM

madkinsz (approved):

PR #5958, opened 2022-07-03
Add Git Support for Databricks Submit Multi Tasks Run

edmondo1984:
## Summary

During spring, Databricks introduced support for Git source when executing jobs on Databricks. This is a particularly useful function because Databricks will automatically check out the code from the specified Repo at the specified commit, branch or tag, before executing the Notebook or the Spark job. Previously, people would need to build the automation to implement CI/CD. 

This feature is also already available in the Terraform provider for defining jobs, and it is now GA

https://docs.databricks.com/repos/jobs-remote-notebook.html

## Changes
- Adds an additional GitSource model and modify signatures of DatabricksSubmitMultitaskRun task
- Extends the signature of the relevant method
- Refactor tests by splitting into four modules: fixtures, mocks, tests for the DatabricksSubmitMultitaskRun and tests for DatabricksSubmitRun
- Introduce unit tests for the new functionality

## Importance

Because this is a very useful feature and it is not possible to intercept the API call performed by the task, people end up duplicating the Task job. This wouldn't be so bad, if not that some of the very useful methods for polling the job status are internal to prefect modules, and one would either need to violate Python best practices for visibility, or duplicate the code also of those methods (see `_handle_databricks_task_execution`)

## Checklist

This PR:

- [x] adds new tests (if appropriate)
- [x] adds a change file in the `changes/` directory (if appropriate)
- [x] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

PR #5964, opened 2022-07-05
Git support

ahuang11:
Attempting to fix tests in https://github.com/PrefectHQ/prefect/pull/5958 since I don't have write access


madkinsz:
Merged into #5958

PR #5975, opened 2022-07-07
Fixing mistake in adding a new parameter to the task

### LINKED ISSUE #5974: Databricks GitSource needs to be passed upon .run invocation but won't work if passed at task construction
## Description
Because I have been sloppy, I forgot to add a parameter to the annotation here https://github.com/PrefectHQ/prefect/blob/f3c8b8150897997372339a4690a2de055fd40717/src/prefect/tasks/databricks/databricks_submitjob.py#L1002 which result in the attribute being defaulted to None unless specified in other ways.

## Expected Behavior
If you set a Git Source at constructor time, this should just work

edmondo1984:
## Summary
Fixes #5974 which I caused myself

### DIFF
--- a/src/prefect/tasks/databricks/databricks_submitjob.py
+++ b/src/prefect/tasks/databricks/databricks_submitjob.py
@@ -1000,6 +1000,7 @@ class DatabricksSubmitMultitaskRun(Task):
         "polling_period_seconds",
         "databricks_retry_limit",
         "databricks_retry_delay",
+        "git_source"
     )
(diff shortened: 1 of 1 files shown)


madkinsz (approved):

madkinsz:
@edmondo1984 looks like this needs a formatting run

edmondo1984:
> @edmondo1984 looks like this needs a formatting run

Fixed, do I need to create a change .yml ?

PR #6004, opened 2022-07-14
use event instead of sleep

a6502:
## Summary
Make src/prefect/utilities/asyncio.py:run_sync_in_interruptible_worker_thread use a event instead of sleep

## Changes

Changes run_sync_in_interruptible_worker_thread to use a anyio.Event instead of anyio.sleep(0). A new subfunction to_thread_run_sync was added so that the was a place to ~~raise~~set the event.

## Importance
Subfunction send_interrupt_to_thread uses anyio.sleep(0) which effectively causes the main thread to run in an endless
loop hogging the python GIL and thus starving the other threads from execution time.

## Checklist

This PR:


madkinsz:
Hi! Thanks for contributing! This makes sense to me and it's simpler so I'm happy to accept it, but I'm curious if you did any benchmarking comparisons with and without this change? My understanding was that `sleep(0)` is always _supposed_ to yield.

a6502:
> but I'm curious if you did any benchmarking comparisons with and without this change?

Not real benchmarking but empirical evidence: I have a test-flow that does some reasonably heavy geo-it processing as a first step and starts a bunch of aws batch jobs for the real heavy lifting. After upgrading to 2.0b8 this first step suddenly took hours instead of a couple of minutes.
Looking at what was going on I noticed that the main thread was using nearly 100% cpu while the thread that was supposed to be doing the processing hardly got any cpu. Further digging using strace revealed that the main thread was doing a lot of epoll() calls and that the 'processing' tread was stuck on a 'aqcuire mutex' call (futex).
I initially changed the sleep(0) to a sleep(1) which immediately improved things a lot. After a bit of thinking I came up with the event solution in this PR.

> My understanding was that `sleep(0)` is always _supposed_ to yield.

A far as I understand this is a yield at the python level mainly to other coroutines in the same thread. For that yield to happen this thread has to first acquire the cpython  GIL. I don't know of the top of my head when a cpython thread releases the GIL, but  I think given the empirical evidence above that a sleep(0) is not enough.

madkinsz:
> I don't know of the top of my head when a cpython thread releases the GIL, but I think given the empirical evidence above that a sleep(0) is not enough.

I know that `sleep(0)` should release the GIL but perhaps the async version doesn't. I'd be curious to explore adding a synchronous sleep afterwards.

```python
await anyio.sleep(0)
await time.sleep(0)
```

Anyway, thanks for the explanation and the contribution. We've got some other places where we want to use this polling pattern so there may be some follow-up work.

madkinsz (approved):

a6502:
> I know that `sleep(0)` should release the GIL but perhaps the async version doesn't. 

Another possibility here it that the thread actually releases the GIL but then wins first place among the "thundering herd" of threads trying to acquire the GIL. I would need to read up on the underlying futex mechanism (on Linux) and how the kernel chooses the thread that gets the lock to understand this better.
 
> Anyway, thanks for the explanation and the contribution. We've got some other places where we want to use this polling pattern so there may be some follow-up work.

A place that I know is using sleep(0) is in the ConcurrentTaskRunner in src/prefect/task_runners.py. I I have written a patch for that already but it needs some more testing and a bit of cleanup before I turn it in a PR. Hopefully within the next couple of days.

madkinsz:
Hm, it looks like this change is consistently failing in CI because it hangs forever on `tests/test_engine.py::TestFlowRunCrashes::test_timeouts_do_not_hide_crashes`

madkinsz (changes requested):
Unfortunately this isn't working as written. I think especially when tasks throw an exception.

I don't have a great MRE to test performance changes here. Can you try just including a `time.sleep(0)` immediately after the `anyio.sleep` call?

madkinsz:
@a6502 the history has diverged since we rewrote commit messages. Should be relatively easy to reapply your commits on the new history on `main`. Let me know if you have questions.

PR #6028, opened 2022-07-19
Fix empty return in S3List task

mcfuhrt:
## Summary
If the S3List task does not return any files, the task automatically fails. Here it would be useful if instead an empty list is returned

## Changes
Check if files are return before adding them to the function return value

## Checklist

This PR:


desertaxle (approved):

madkinsz:
I lack permissions to push changes to your branch and have updated this in #6534

Issue #3375, opened 2020-09-23
ShellTask Powershell Compatibility


dylanbhughes:
## Current behavior

ShellTask doesn't seem to work with powershell by default

## Proposed behavior

Update to accept `powershell` as a valid input.

dylanbhughes:
Related: https://github.com/PrefectHQ/prefect/issues/1984

jbowlen:
Is it foreseeable when powershell scripts can be executed?

madkinsz:
I'd be happy to review a pull request adding this functionality, but cannot provide a timeline for our team looking into it.

PR #6035, opened 2022-07-21
[Enhancement] Powershell as shell input to ShellTask

chinmay-bhat:
## Summary

This PR adds Powershell as a valid input shell for ShellTask. https://github.com/PrefectHQ/prefect/issues/3375

## Changes

Powershell is unable to read the temp file while it's open in another process (`with tempfile.NamedTemporaryFile(prefix="prefect-") as tmp:`).
So the solution is to update the file with the helper script and command, add a `ps1` suffix, and close it. 
Only then we run the temp script with powershell, and finally unlink (delete) the file.
For reference, see this [stackoverflow answer](https://stackoverflow.com/a/5344603) 

## Checklist

This PR:

### DIFF
--- a/changes/pr6035.yaml
+++ b/changes/pr6035.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Powershell as shell input to ShellTask - [#6035](https://github.com/PrefectHQ/prefect/pull/6035)"
+
+contributor:
+  - "[Chinmay Bhat](https://github.com/chinmay-bhat)"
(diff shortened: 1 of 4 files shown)


chinmay-bhat:
It seems [Powershell](https://github.com/PowerShell/PowerShell) is not installed for tests, so it's throwing errors.

madkinsz:
Hey! Thanks for contributing :) Considering all of the implementation changes necessary for Powershell, should we introduce a separate `PoweshellTask`?

chinmay-bhat:
Hi @madkinsz , yeah that would be a better solution!

madkinsz (approved):
Great! Thank you :)

PR #6078, opened 2022-07-28
Head off possible `PendingRollbackError`s from the `scheduler` and `flow-run-notifications` `LoopService`s.

chrisguidry:
The comments in the code tell the story in detail, but because this loop-service
handles Exceptions internally, it may sometimes handle transient DB errors that
require a rollback before proceeding with more work.

Part of PrefectHQ/nebula#1220

### DIFF
--- a/src/prefect/orion/services/flow_run_notifications.py
+++ b/src/prefect/orion/services/flow_run_notifications.py
@@ -31,11 +31,12 @@ class FlowRunNotifications(LoopService):
         async with session:
             async with session.begin():
                 while True:
-                    # Drain the queue one entry at a time, because if a database error
-                    # happens while sending a notification, we'll need to rollback the
-                    # transaction, which effectively re-queues any notifications that
-                    # we pulled here.  If we drain in batches larger than 1, we risk
-                    # double-sending notifications when a transient error occurs.
+                    # Drain the queue one entry at a time, because if a transient
+                    # database error happens while sending a notification, the whole
+                    # transaction will be rolled back, which effectively re-queues any
+                    # notifications that we pulled here.  If we drain in batches larger
+                    # than 1, we risk double-sending earlier notifications when a
+                    # transient error occurs.
                     notifications = await db.get_flow_run_notifications_from_queue(
                         session=session,
                         limit=1,
@@ -55,12 +56,14 @@ class FlowRunNotifications(LoopService):
                     finally:
                         connection = await session.connection()
                         if connection.invalidated:
-                            # If the connection was invalidated due to an error that
-                            # we handled in _send_flow_run_notification, we'll need to
-                            # rollback the session before we can proceed with more
-                            # iterations of the loop.  This may happen due to transient
-                            # database connection errors, but will _not_ happen due to
-                            # an calling a third-party service to send a notification.
+                            # If the connection was invalidated due to an error that we
+                            # handled in _send_flow_run_notification, we'll need to
+                            # rollback the session in order to synchronize it with the
+                            # reality of the underlying connection before we can proceed
+                            # with more iterations of the loop.  This may happen due to
+                            # transient database connection errors, but will _not_
+                            # happen due to an calling a third-party service to send a
+                            # notification.
                             await session.rollback()
                             assert not connection.invalidated
 
--- a/src/prefect/orion/services/flow_run_notifications.py
+++ b/src/prefect/orion/services/flow_run_notifications.py
@@ -69,7 +69,6 @@ class FlowRunNotifications(LoopService):
         self,
         session: sa.orm.session,
         db: OrionDBInterface,
-        # notification: OrionDBInterface.Notification,
         notification,
     ):
         try:
--- a/src/prefect/orion/services/flow_run_notifications.py
+++ b/src/prefect/orion/services/flow_run_notifications.py
@@ -2,7 +2,6 @@
 A service that checks for flow run notifications and sends them.
 """
 import asyncio
-from typing import List
 
 import sqlalchemy as sa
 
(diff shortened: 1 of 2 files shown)


src/prefect/orion/services/flow_run_notifications.py
```diff
@@ -34,9 +31,14 @@ async def run_once(self, db: OrionDBInterface):
...
                 while True:
+                    # Drain the queue one entry at a time, because if a database error
+                    # happens while sending a notification, we'll need to rollback the
+                    # transaction, which effectively re-queues any notifications that
+                    # we pulled here.  If we drain in batches larger than 1, we risk
+                    # double-sending notifications when a transient error occurs.
                     notifications = await db.get_flow_run_notifications_from_queue(
                         session=session,
-                        limit=self.batch_size,
+                        limit=1,
```
zangell44, line 41: are we worried about 
- performance of doing this one by one?
- clearing "bad" notifications from the queue? does this actually delete it? 🤔 

alternatively we could
- open a transaction
- get notifications from queue (which removes them)
- commit
- open a new transaction for looking up the block document and sending the notification etc

src/prefect/orion/services/flow_run_notifications.py
```diff
@@ -46,59 +48,75 @@ async def run_once(self, db: OrionDBInterface):
...
+                            assert not connection.invalidated
 
     @inject_db
-    async def send_flow_run_notifications(
-        self, session: sa.orm.session, db: OrionDBInterface, notifications: List
+    async def send_flow_run_notification(
+        self,
+        session: sa.orm.session,
+        db: OrionDBInterface,
+        # notification: OrionDBInterface.Notification,
```
zangell44, line 72:
```suggestion
```
minor cleanup

src/prefect/orion/services/scheduler.py
```diff
@@ -134,6 +127,40 @@ def _get_select_deployments_to_schedule_query(self, db: OrionDBInterface):
         )
         return query
 
+    async def _collect_flow_runs(
+        self,
+        session: sa.orm.Session,
+        deployment_ids: List[UUID],
+        now: pendulum.DateTime,
```
zangell44, line 134: probably better to set `now = pendulum.now("UTC")` instead of passing through

src/prefect/orion/services/scheduler.py
```diff
@@ -134,6 +127,40 @@ def _get_select_deployments_to_schedule_query(self, db: OrionDBInterface):
...
+                    f"Error scheduling deployment {deployment_id!r}.",
+                )
+            finally:
+                connection = await session.connection()
+                if connection.invalidated:
+                    # If the error we handled above was a database error that caused the
+                    # transaction to rollback and become invalidated, rollback this
+                    # session, break from this loop iteration, and have the main loop
+                    # in run_once start over
+                    await session.rollback()
```
zangell44, line 160: wouldn't this rollback all work done so far including work done outside of this function?

chrisguidry: I don't think I'm explaining the situation in the comment well enough, I'll work on improving it, since this is so obscure.  The DB transaction has _already been rolled back_ and we're just detecting that situation with `connection.invalidated`.  Calling `session.rollback` helps to get the session back in sync with the underlying connection. That way, when we get to the end of the `with session.begin():` block in the outer function, the context manager won't attempt to `commit()` it.

chrisguidry: I added more words, does this make better sense?

zangell44: hmmmm okay i think im following now
would this cause issues if there's one "bad" deployment within a group that's commonly queried for together?

zangell44: (i dont think it would be any worse than it currently is, just want to guard against it)

src/prefect/orion/services/flow_run_notifications.py
chrisguidry: > are we worried about
> * performance of doing this one by one?

I am worried about that.  I'd like to add some monitoring to the flow_run_notifications queue depth to make sure we are able to keep it drained.

> * clearing "bad" notifications from the queue? does this actually delete it? 🤔

I don't think I'm worried about that one.  This kind of comes back to the subtlety of the error I'm handling here.  Any "normal" error we see here (like a problem with a bad Slack token, for example) will _not_ cause this transaction to roll back, so we won't re-enqueue those.  The kinds of errors I'm catching here are ones that _invalidate the connection_, like transient DB and environmental errors.  In those cases, the transaction is actually _already rolled back_, so we just need to get the Python session objects synced back up with that reality.

I think my comments in the code are a little inaccurate, I'll tweak them.

src/prefect/orion/services/scheduler.py
chrisguidry: In both the `scheduler` and `flow-run-notifications`, we're sensitive to that problem: one bad apple spoiling the bunch, so to speak.  That's why we have the exception handling here, to guard against that case.  The only problem is that our exception handlers are also catching a small class of errors that we can't recover from unless we check the `connection.invalidated`.

I can't quite prove it, but an error like this may be an example of something we're not properly recovering from: https://github.com/PrefectHQ/nebula/issues/1157

src/prefect/orion/services/flow_run_notifications.py
chrisguidry: > alternatively we could
> * open a transaction
> * get notifications from queue (which removes them)
> * commit
> * open a new transaction for looking up the block document and sending the notification etc

Jake suggested this in my other version of the PR, to use a separate session and transaction for the notification part.  The only thing I was worried about there was that a transient error (connection drop, db restart, etc) would end up with us not sending a user notification when we should have.  I guess we're running into the whole "queue guarantees" problem here.  I'm trying to make sure we deliver at least once, rather than at most once.

zangell44 (approved):

PR #6080, opened 2022-07-28
Ignore prefect api url in test harness

### LINKED ISSUE #6055: prefect_test_harness does not ignore PREFECT_API_URL
## Description

When PREFECT_API_URL is set and a flow is run using `prefect_test_harness` then the test tries to use PREFECT_API_URL. I was expecting the harness would ignore PREFECT_API_URL because it creates a temporary database.

## Reproduction / Example
```python
from prefect import flow
from prefect.testing.utilities import prefect_test_harness
import os


@flow
def my_favorite_flow():
    return 42


def test_my_favorite_flow():
    os.environ["PREFECT_API_URL"] = "http://foobar:4200/api"
    with prefect_test_harness():
        # run the flow against a temporary testing database
        assert my_favorite_flow() == 42
```

Fails with:

```
RuntimeError: Cannot create flow run. Failed to reach API at http://foobar:4200/api/.
```

prefect 2.0b12

zangell44:
Closes #6055

### DIFF
--- a/src/prefect/testing/utilities.py
+++ b/src/prefect/testing/utilities.py
@@ -124,7 +124,8 @@ def prefect_test_harness():
                     restore_defaults={prefect.settings.PREFECT_API_URL},
                     # Use a temporary directory for the database
                     updates={
-                        prefect.settings.PREFECT_ORION_DATABASE_CONNECTION_URL: DB_PATH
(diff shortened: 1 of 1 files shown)


chrisguidry (approved):
Great catch by the original reporter!

PR #6081, opened 2022-07-28
Coerce any floating point extrema (`NaN`, `-Infinity`, `Infinity`) in JSON objects to `None` for the DB

chrisguidry:
PostgreSQL [explicitly does not support these values](https://www.postgresql.org/docs/current/datatype-json.html#JSON-TYPE-MAPPING-TABLE), and while SQLite supports saving and reading full JSON documents containing these values, any attempt to parse them in SQLite (like to use them with JSON operators) will fail.

This change follows the Javascript convention, which is to coerce these values to `null`, which should have good standards compatibility with the frontend and other non-Python clients.  The JSON spec does not comment on what to do with these values, but [MDN says they are prohibited](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#numbers).

This was observed in our Prefect Cloud 2.0 environment: https://github.com/PrefectHQ/nebula/issues/1157 as errors like:

```
<class 'asyncpg.exceptions.InvalidTextRepresentationError'>: invalid input syntax for type json
[DETAIL] Token "NaN" is invalid
```

### DIFF
--- a/src/prefect/orion/utilities/database.py
+++ b/src/prefect/orion/utilities/database.py
@@ -54,15 +54,15 @@ def _generate_uuid_sqlite(element, compiler, **kwargs):
 
     return """
     (
-        lower(hex(randomblob(4))) 
-        || '-' 
-        || lower(hex(randomblob(2))) 
-        || '-4' 
-        || substr(lower(hex(randomblob(2))),2) 
-        || '-' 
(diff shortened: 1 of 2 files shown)


bunchesofdonald (approved):
:+1: Makes sense to me.

zangell44 (approved):

PR #6085, opened 2022-07-28
Remove deployment execute

anticorrelator:
`prefect deployment execute` is a convenience command used to try and run deployments without an active agent. This command is now very sensitive to the base directory from which which it's called, which only causes confusion and I think we should remove it.


anna-geller (approved):
Looks great, nothing to complain about 👍

PR #6088, opened 2022-07-28
Set `format` of JsonPatch schema to `rfc6902`

bunchesofdonald:
In issue #6079 we found that the `customizations` OpenAPI schema was being output as `string` and this caused the UI to display the field as a simple text input.

This PR changes the schema to be an `array` with `object` children. This makes sense because JSON Patch is an array of JSON objects. This also sets the `format` to `rfc6902` so that further validation can be performed per the RFC spec. The schema for JSON Patch is now:

```
"customizations": {
    "title": "Customizations",
    "type": "array",
    "format": "rfc6902",
    "items": {
        "type": "object",
        "additionalProperties": {
            "type": "string"
        }
    }
}
```

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -67,11 +67,11 @@ class KubernetesJob(Infrastructure):
     # shortcuts for the most common user-serviceable settings
     image: str = Field(default_factory=get_prefect_image_name)
     namespace: str = "default"
-    service_account_name: str = None
+    service_account_name: Optional[str] = None
     image_pull_policy: Optional[KubernetesImagePullPolicy] = None
 
     # connection to a cluster
-    cluster_config: KubernetesClusterConfig = None
+    cluster_config: Optional[KubernetesClusterConfig] = None
 
     # settings allowing full customization of the Job
     job: KubernetesManifest = Field(
--- a/src/prefect/utilities/pydantic.py
+++ b/src/prefect/utilities/pydantic.py
@@ -229,4 +229,13 @@ class PartialModel(Generic[M]):
 class JsonPatch(JsonPatchBase):
     @classmethod
     def __modify_schema__(cls, field_schema):
-        field_schema.update(type="string", format="json-string")
+        field_schema.update(
+            {
+                "type": "array",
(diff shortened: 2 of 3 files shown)


src/prefect/infrastructure/kubernetes.py
```diff
@@ -67,11 +67,11 @@ class KubernetesJob(Infrastructure):
     # shortcuts for the most common user-serviceable settings
     image: str = Field(default_factory=get_prefect_image_name)
     namespace: str = "default"
-    service_account_name: str = None
+    service_account_name: Optional[str] = None
     image_pull_policy: Optional[KubernetesImagePullPolicy] = None
 
     # connection to a cluster
-    cluster_config: KubernetesClusterConfig = None
+    cluster_config: Optional[KubernetesClusterConfig] = None
```
bunchesofdonald, line 74: I also noticed that these types were incorrect, they should be optional since we're assigning `None` by default.

chrisguidry (approved):
Nice!  What does the UI do with it in the near term, does it just skip field types it doesn't recognize?

bunchesofdonald:
At the moment it looks like it'll make it a Tags input (because it's an `array` type). I _think_ we just need to update [this method](https://github.com/PrefectHQ/orion-design/blob/main/src/components/BlockSchemaPropertyInput.vue#L62) and check for the `format`.

src/prefect/infrastructure/kubernetes.py
abrookins: Is it really optional though?

bunchesofdonald: It is since we're setting it to None, typeing wise at least.

abrookins: Ah yeah it is

madkinsz (approved):

madkinsz:
We could add additional annotations to the schemas for the UI as well, e.g. we could include `prefect-type: "tag"` on the schema for tag objects and `prefect-type: "json-patch-item"` on this schema.

madkinsz:
cc @zhen0

zhen0:
Thanks @pleek91 - for info

PR #6089, opened 2022-07-28
Preserve schedules set via the API in building deployments

cicdw:
This PR ensures that schedules set via API / UI on a deployment are preserved when building deployments from the CLI.  This behavior mirrors the current behavior with deployment `description`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,7 +41,7 @@ site/
 
 # UI artifacts
 src/prefect/orion/ui/*
-orion-ui/node_modules
+**/node_modules
 
 # Databases
 *.db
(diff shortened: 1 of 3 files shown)


abrookins:
Can we get a test for this behavior?

cicdw:
@abrookins I need to refactor the whole cli deployment test suite -- if you're OK with it, I'd like to work on that as a high priority project in itself; I also would like to add more tests around how storage is working.  I can get that all in by Monday at the latest but if we're going to cut a release in the near term I'd like to include this nicety.  Let me know what you think!

abrookins:
Sounds good to me! @cicdw

abrookins (approved):
LGTM 👍

PR #6090, opened 2022-07-28
Remove circleCI config

cicdw:


cicdw:
Turns out this file is necessary so long as we want tests to run on our 1.0 branches, which I think we _do_ want for a while longer.  Closing for now.

Issue #5986, opened 2022-07-10
Log when a flow run fails because of missing parameters


tekumara:
## Description

When a Deployment doesn't provide parameters to a flow that requires them, the KubernetesFlowRunner fails the flow run without providing any logs:

<img width="1197" alt="Screen Shot 2022-07-10 at 5 57 58 pm" src="https://user-images.githubusercontent.com/125105/178136386-c1614302-0ffc-4bd9-9e79-83b35baba7ad.png">

So it's not clear what went wrong. Would be nice to have logs that indicate the failure, which is the case when running the deployment locally, eg:
```
$ prefect deployment execute flow-params/no-params
Loading flow from deployed location...
Running flow...
Traceback (most recent call last):
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/prefect/utilities/callables.py", line 23, in get_call_parameters
    bound_signature = inspect.signature(fn).bind(*call_args, **call_kwargs)
  File "/Users/tekumara/.pyenv/versions/3.9.13/lib/python3.9/inspect.py", line 3045, in bind
    return self._bind(args, kwargs)
  File "/Users/tekumara/.pyenv/versions/3.9.13/lib/python3.9/inspect.py", line 2960, in _bind
    raise TypeError(msg) from None
TypeError: missing a required argument: 'i'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/prefect/cli/_utilities.py", line 44, in wrapper
    return fn(*args, **kwargs)
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/prefect/utilities/asyncio.py", line 193, in wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/prefect/utilities/asyncio.py", line 140, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/Users/tekumara/code/orion-demo/.venv/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
[...]

PR #6091, opened 2022-07-28
Improve validation error messages and logging

peytonrunyan:
This PR adds 3 new exceptions to improve errors when parameters are incorrectly supplied to flow runs. Closes https://github.com/PrefectHQ/prefect/issues/5980 and closes https://github.com/PrefectHQ/prefect/issues/5986.

These errors add information about which parameters were invalid, and why they were invalid, when a flow run fails for a validation error. It adds the information in two locations - the flow run's logs, and the flow run's state. Previously we were only logging the validation error with the engine logger, and were including no information about the parameters.

See here for more context: https://github.com/PrefectHQ/orion/pull/2319

**Signature Mistmatch Error** - if parameters are provided to a deployment run that do not match flow signature
Screen Shot 2022-08-12 at 8 45 59 AM

**Parameter Type Error** - if the signature matches, but the data types provided do not match those specified in the signature
Example with deployment
Screen Shot 2022-08-12 at 8 45 48 AM

Example with flow run
Screen Shot 2022-08-12 at 9 16 06 AM

**Parameter Bind Error** - Happens when parameters passed to a flow that is run as a script are mismatched, or when a subflow's parameters are mismatched

Example with subflow run from a deployment
Screen Shot 2022-08-12 at 9 24 28 AM

Example with flow run as a script
Screen Shot 2022-08-12 at 9 19 01 AM


src/prefect/cli/deployment.py
```diff
@@ -27,6 +27,7 @@
     DeploymentYAML,
     PackageManifest,
     load_deployments_from_yaml,
+    load_flow_from_deployment,
```
cicdw, line 30: unused import?

src/prefect/engine.py
```diff
@@ -191,14 +197,22 @@ async def create_then_begin_flow_run(
         raise RuntimeError(
             f"Cannot create flow run. Failed to reach API at {client.api_url}."
         ) from connect_error
-
     state = Pending()
     if flow.should_validate_parameters:
         try:
             parameters = flow.validate_parameters(parameters)
+        except (ParameterTypeError, SignatureMismatchError) as exc:
+            validation_failure_msg = str(exc)
```
cicdw, line 205: minor nit -- I always prefer `repr(exc)` for exceptions, as the repr preserves the exception type whereas the string representation is _only_ the message.

Also I think you meant to pass this variable in as the `message` input to the state below 👍

src/prefect/engine.py
```diff
@@ -191,14 +197,22 @@ async def create_then_begin_flow_run(
...
+            state = Failed(
+                message=str(exc),
+                data=DataDocument.encode("cloudpickle", exc),
+            )
         except Exception as exc:
+            validation_failure_msg = (
+                f"Validation of flow parameters failed with an unexpected error: {exc}"
```
cicdw, line 212:
```suggestion
                f"Validation of flow parameters failed with an unexpected error: {exc!r}"
```
(ensures repr is used)

src/prefect/engine.py
```diff
@@ -213,6 +227,7 @@ async def create_then_begin_flow_run(
     engine_logger.info(f"Created flow run {flow_run.name!r} for flow {flow.name!r}")
 
     if state.is_failed():
+        flow_run_logger(flow_run).error(validation_failure_msg)
```
cicdw, line 230: I think it could be easy for this variable name to change, and so instead of relying on a variable set within conditional code paths above I think you'll find this to be more robust if you did this instead:
```suggestion
        flow_run_logger(flow_run).error(state.message)
```

src/prefect/engine.py
```diff
@@ -253,27 +267,51 @@ async def retrieve_flow_then_begin_flow_run(
...
+        parameters=parameters,
     )
 
     if flow.should_validate_parameters:
+        failed_state = None
         try:
-            parameters = flow.validate_parameters(flow_run.parameters)
+            parameters = flow.validate_parameters(parameters)
+        except (ParameterTypeError, SignatureMismatchError) as exc:
+            flow_run_logger(flow_run).error(str(exc))
```
cicdw, line 294:
```suggestion
            flow_run_logger(flow_run).exception("Flow parameter validation failed.")
```
using `.exception` will log the traceback 👍

src/prefect/engine.py
```diff
@@ -253,27 +267,51 @@ async def retrieve_flow_then_begin_flow_run(
...
+            failed_state = Failed(
+                message=str(exc),
```
cicdw, line 296:
```suggestion
                message=repr(exc),
```

src/prefect/engine.py
```diff
@@ -253,27 +267,51 @@ async def retrieve_flow_then_begin_flow_run(
...
+                data=DataDocument.encode("cloudpickle", exc),
+            )
         except Exception as exc:
-            flow_run_logger(flow_run).exception("Flow run received invalid parameters.")
-            state = Failed(
-                message="Flow run received invalid parameters.",
+            unexpected_error = (
+                f"Validation of flow parameters failed with an unexpected error: {exc}"
```
cicdw, line 301:
```suggestion
                f"Validation of flow parameters failed with an unexpected error: {exc!r}"
```

cicdw (changes requested):
I'll want to give this another pass -- it feels like we have an opportunity to consolidate this logic into a unified place / function

peytonrunyan:
Awesome feedback, thanks! I'll get those changes up. Regarding your first message about consolidating logic - would you mind doing it on a huddle so that I can see/hear your process for reasoning around the changes? I'm trying to develop my eye for things like that.

cicdw:
@peytonrunyan absolutely - we can review both this PR and the backend logic that we discussed on Friday 👍

src/prefect/engine.py
```diff
@@ -253,27 +263,49 @@ async def retrieve_flow_then_begin_flow_run(
...
+            failed_state = Failed(
+                message=repr(exc),
+                data=DataDocument.encode("cloudpickle", exc),
+            )
         except Exception as exc:
-            flow_run_logger(flow_run).exception("Flow run received invalid parameters.")
-            state = Failed(
-                message="Flow run received invalid parameters.",
+            unexpected_error = f"Validation of flow parameters failed with an unexpected error: {exc!r}"
+            flow_run_logger(flow_run).error(unexpected_error)
```
madkinsz, line 297:
```suggestion
            flow_run_logger(flow_run).exception(unexpected_error)
```

madkinsz:
@cicdw  @peytonrunyan Is this still in progress? I'd love to get these changes in now and DRY the code in a follow-up as these are valuable improvements for user experience.

cc @bunchesofdonald

peytonrunyan:
Chris and I are about to be working on finishing this up I think.

src/prefect/engine.py
```diff
@@ -214,6 +213,7 @@ async def create_then_begin_flow_run(
     engine_logger.info(f"Created flow run {flow_run.name!r} for flow {flow.name!r}")
 
     if state.is_failed():
+        flow_run_logger(flow_run).error(state.message)
```
cicdw, line 216: I think this log chronologically makes more sense up on line 200 with `.exception` instead of `.error`

cicdw (commented):
minor request to move a log, otherwise LGTM

src/prefect/engine.py
peytonrunyan: I think it's in that order to avoid the strange situation of logging in the order:
>> Flow run <flow_run_name> received invalid parameters and is marked as failed
>> Created flow run <flow_run_name> for flow <flow.name>

Which I think could be confusing. Definitely updating to exception.

madkinsz:
@peytonrunyan let's get this across the line

peytonrunyan:
I shot Chris a message this morning. I think he's driving. Just merged his requested changes.

src/prefect/engine.py
```diff
@@ -192,14 +192,14 @@ async def create_then_begin_flow_run(
         raise RuntimeError(
             f"Cannot create flow run. Failed to reach API at {client.api_url}."
         ) from connect_error
-
     state = Pending()
     if flow.should_validate_parameters:
         try:
             parameters = flow.validate_parameters(parameters)
         except Exception as exc:
+            flow_run_logger(flow_run).exception(state.message)
```
cicdw, line 200: @peytonrunyan I think you meant 
```suggestion
            flow_run_logger(flow_run).exception("Validation of flow parameters failed.")
```
(note: no need to format the exception since `.exception` will log the full traceback)

cicdw: Also it looks like `flow_run` doesn't exist yet, so maybe my suggestion introduces too much complexity here

peytonrunyan: Yup, but also I just realized that we don't have a flow run for a flow run logger until after line 212.

peytonrunyan:
Just a sec - the reversion also removed the change to exception from error and I'm trying to get my local git back in order.

madkinsz:
Can you examples of the final user experience this provides in the PR body?

peytonrunyan:
Done. It looks like the way things currently execute, I won't ever hit the signature mismatch error. Instead I wind up with
```
prefect.exceptions.ParameterBindError: Error binding parameters for function 'dog': missing a required argument: 'puppy'.
Function 'dog' has signature 'puppy: str' but received args: () and kwargs: {'kitty': 42}.
```

test_flow.py
```diff
@@ -0,0 +1,10 @@
+from prefect import flow
```
madkinsz, line 1: Did you mean to commit this file?

madkinsz:
Shouldn't we hit a signature mismatch error when the parameters are already determined, e.g. a flow run for a deployment?

src/prefect/engine.py
```diff
@@ -254,27 +253,45 @@ async def retrieve_flow_then_begin_flow_run(
         )
         return state
 
+    try:
+        parameters = get_call_parameters(
+            flow.fn, call_args=(), call_kwargs=flow_run.parameters
+        )
```
madkinsz, line 259: This is why you're never seeing a signature mismatch error. This should probably resolve arguments and keyword arguments using the dedicated function first. Why do we do this call here?

peytonrunyan: I think in general I had a bit of a poor understanding of how things executed. In `Flow.run()` we call `parameters = get_call_parameters` before submitting to `enter_flow_run_engine_from_flow_call`. Then we submit to either `create_and_begin_subflow_run` or `create_then_begin_flow_run`. After that step we call `validate_parameters`. 

The call to `get_call_parameters` to `retrieve_flow_then_begin_flow_run` preserves that order.

peytonrunyan: Removing it does expose the error. I'll get the PR body updated

peytonrunyan:
Alright - should be fully up to date. All of the exceptions are reachable and documented.

src/prefect/engine.py
```diff
@@ -254,27 +253,34 @@ async def retrieve_flow_then_begin_flow_run(
         )
         return state
 
+    parameters = flow_run.parameters
+
     await client.update_flow_run(
         flow_run_id=flow_run_id,
         flow_version=flow.version,
+        parameters=parameters,
```
madkinsz, line 261: I think the reason we were doing the prior code was to retrieve default values for unprovided parameters. No reason to update the flow run parameters in this following call if we're not mutating them.

cicdw (approved):

madkinsz (approved):

PR #6094, opened 2022-07-28
Split up the insert of logs into smaller chunks.

chrisguidry:
When large log batches arrive at the API, they may exceed what we can INSERT in
a single statement, which is limited to 32,767 parameters (and will also have
some textual query size limit as well).  This splits the logs into chunks that
we know will be beneath those limits.

Observed in Prefect Cloud: https://github.com/PrefectHQ/nebula/issues/1247

### DIFF
--- a/src/prefect/orion/models/logs.py
+++ b/src/prefect/orion/models/logs.py
@@ -10,6 +10,16 @@ from sqlalchemy import select
 import prefect.orion.schemas as schemas
 from prefect.orion.database.dependencies import inject_db
 from prefect.orion.database.interface import OrionDBInterface
+from prefect.utilities.collections import batched_iterable
+
+# We have a limit of 32,767 parameters at a time for a single query...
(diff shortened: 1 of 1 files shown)


zangell44 (approved):

src/prefect/orion/models/logs.py
```diff
@@ -10,6 +10,16 @@
...
 from prefect.orion.database.interface import OrionDBInterface
+from prefect.utilities.collections import batched_iterable
+
+# We have a limit of 32,767 parameters at a time for a single query...
+MAXIMUM_QUERY_PARAMETERS = 32_767
+
+# ...and logs have a certain number of fields...
+NUMBER_OF_LOG_FIELDS = len(schemas.core.Log.schema()["properties"])
+
+# ...so we can only INSERT batches of a certain size at a time
```
abrookins, line 21: Love the story these comments are telling me

PR #6103, opened 2022-07-29
bump readchar to >=3.0.6

Cube707:
readchar released a new version, that finally fixes support for arrow-keys on windows. As this featrue is usefull to this project I recomend requiring that version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ pydantic >= 1.8.2
 python-slugify >= 5.0
 pytz >= 2021.1
 pyyaml >= 5.4.1
-readchar >= 3.0.5
+readchar >= 3.0.6
 rich >= 11.0
 slack-sdk>=3.15.1
 sqlalchemy[asyncio] >= 1.4.20, != 1.4.33


madkinsz:
Hi! Thanks for the contribution. We keep our minimum bounds for packages as low as we actually require to preserve maximum compatibility with other projects. Since we do not require the changes in the most recent version, we won't increase our required version. Installations of Prefect will get the latest version of `readchar` by default and upgrades can be done safely.

Cube707:
Ok, that's an unexpected reply, but your call.

Just to be sure: `v3.0.5` actually worsend windows suppored compared to `v3.0.4`, so by that logic the requiremend should be lowerd to `v3.0.4`

madkinsz:
There's nothing in our minimum version requirement preventing users from receiving the latest version. However, if we bump our minimum version and another package requires an older version, we will not be compatible with it. You can see extensive discussion of this at https://iscinumpy.dev/post/bound-version-constraints/.

Cube707:
yes I know that. I just don't get why requiring a version that has potential to break a feature of this project is acceptable, but requiring the version that interduces the fix is not.

madkinsz:
Ah, I see what you're saying. Apologies for the misunderstanding. If the `prefect cloud select-workspace` command is not working on Windows, we should bump the minimum requirement. Please link to an issue or describe the feature that's broken in the future.

Cube707:
The short of it: Windows uses two different "escape" bytes, to denote special keys that are made up of two characters. `readchar-v3.0.4` only registered one of these special bytes, meaning that the speciall keys (including arrow keys) would only work on some systems. With `v3.0.5` a PR was merged that changed *only*  the four arrow keys to the other byte, making them work on some windows systems, but breaking previously working systems. With `v3.0.6` this is fixed and both versions are are finally detected correctly.

See my explanation in this issue: https://github.com/magmax/python-readchar/issues/66
and this comment on the now reverted bad PR: https://github.com/magmax/python-readchar/issues/66

And see this commit that fixed it: https://github.com/magmax/python-readchar/commit/a7c06456f6a10363b0d9b88c521b9958dfa666b0

madkinsz (approved):
Thanks for explaining! Technically we could make this version change based on the platform but it seems simpler to just bump our requirement in the long run.

Cube707:
I would also say thats over-complicating things. 

Also `v3.0.5` broke BACKSPACE on posix platforms. Thats not relevant for this project, but still not bad to avoid it.

PR #6105, opened 2022-07-29
Update `from_path` and skip upload on same path

### LINKED ISSUE #6082: Local file system should only be exposed as anonymous block and should not be configurable from the UI
## Problem
When building a local deployment using an implicitly created local-file-system block as here:

https://github.com/anna-geller/prefect-deployment-patterns/blob/2530e57cc1769f4cdf8476ee795b762068fafb94/a_project_template/deploy_commands_local_fs.bash#L2 

everything works well and it creates an anonymous block:

image

Because the deployment and related block have been created from the same machine on which the code is stored and on which the runs will get executed, everything works seamlessly -- this way, the correct `base_path` gets automatically inferred and set on the block. 

However, when trying to create a block from the UI 
image

and passing the slug into a build command

```
prefect deployment build flows/hello.py:hello --name localfs --tag dev -sb local-file-system/localfs
```

the build fails with:

```
File "/opt/miniconda3/envs/final/lib/python3.9/site-packages/prefect/filesystems.py", line 114, in put_directory
    shutil.copytree(from_path, local_path, dirs_exist_ok=True)
NameError: name 'from_path' is not defined
```

image

Using the automatically created anonymous block and passing it as `-sb`, fails too with the same error:
```
NameError: name 'from_path' is not defined
```

image

## Solution: Disable creation of local file system blocks from the UI

From the UX perspective, the creation of local storage blocks from the UI seems confusing since you still need to run it from the machine where this code is located and you also need to start an agent from the machine on which this local code is supposed to run. 

### Suggested approach
An explicit storage block should be disabled for the local file system to avoid countless bugs users might come across. Allowing to create this type of storage ONLY from the CLI from the machine where this code is located would make everything less ambiguous and would improve the UX + avoid confusion. Already the fact that the user might manually specify the wrong `base_path` from the UI, which doesn't match where the code is actually located, has a big potential for bugs and confusion.  

cc @cicdw

anna-geller:
@cicdw closes #6082

### DIFF
--- a/src/prefect/filesystems.py
+++ b/src/prefect/filesystems.py
@@ -108,10 +108,14 @@ class LocalFileSystem(ReadableFileSystem, WritableFileSystem):
         if local_path is None:
             local_path = Path(".").absolute()
 
-        if sys.version_info < (3, 8):
-            shutil.copytree(from_path, local_path)
+        if local_path == to_path:
+            print("Skipping upload")
         else:
-            shutil.copytree(from_path, local_path, dirs_exist_ok=True)
+            if sys.version_info < (3, 8):
+                shutil.copytree(local_path, to_path)
+            else:
+                shutil.copytree(local_path, to_path, dirs_exist_ok=True)
+
 
     async def read_path(self, path: str) -> bytes:
         path: Path = self._resolve_path(path)
--- a/src/prefect/filesystems.py
+++ b/src/prefect/filesystems.py
@@ -109,7 +109,7 @@ class LocalFileSystem(ReadableFileSystem, WritableFileSystem):
             local_path = Path(".").absolute()
 
         if local_path == to_path:
-            print("Skipping upload")
+            pass
         else:
             if sys.version_info < (3, 8):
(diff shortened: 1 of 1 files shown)


src/prefect/filesystems.py
```diff
@@ -108,10 +108,13 @@ async def put_directory(self, local_path: str = None, to_path: str = None) -> No
         if local_path is None:
             local_path = Path(".").absolute()
 
-        if sys.version_info < (3, 8):
-            shutil.copytree(from_path, local_path)
+        if local_path == to_path:
+            print("Skipping upload")
```
cicdw, line 112: `print` statements are generally not best practice within application code -- I propose that for now we put a `pass` here, and in the future we can wire up some loggers to these objects.  The API only accepts logs attached to flow runs so we'll have to do logging configuration to make that happen which I don't think is worth it just yet
```suggestion
            pass
```

cicdw (changes requested):
minor line change request, otherwise great catch

cicdw (approved):

Issue #6102, opened 2022-07-29
Orion UI reveals database secret in settings page


clamydo:
When going to the 'Settings' tab in the UI, all configuration variables are displayed, *including* the PostgreSQL connection string *with* password, `PREFECT_ORION_DATABASE_CONNECTION_URL` (see attached image)

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/13063485/181708410-ff269c21-2621-4be4-b685-730230bc299f.png">

Furthermore, `PREFECT_ORION_DATABASE_PASSWORD` is displayed as well (empty in my case).

clamydo:
Duplicate I’d #5941

PR #6121, opened 2022-07-29
Hide password values

ahuang11:
### Summary

Fixes https://github.com/PrefectHQ/prefect/issues/6102

### Steps Taken to QA Changes

```
export PREFECT_ORION_DATABASE_PASSWORD='postgres_password'
export PREFECT_ORION_DATABASE_CONNECTION_URL='postgresql+asyncpg://postgres:postgres_password@localhost/orion'
prefect orion start
```
image

### Checklist

This pull request is:
- [ x ] A short code fix

**Happy engineering!**


abrookins:
There are deeper problems here @ahuang11, like the fact that we could log your connection URL unmasked. We need to think a bit holistically about how to resolve this, and it may require changes to our Settings interface.

1. We could look for this specific setting anywhere it might get printed to stdout, logged, or displayed in the UI and add special casing for it.
2. We could teach our Settings system what to do with values that the user wants "masked" for display purposes, but wants us to retrieve for its actual purpose.

abrookins:
Short term we may want to do something like this to buy us time depending on how hard changing settings would be though. 😅

ahuang11:
Actually, since I'm literally overwriting the value in the post_root_validators, not sure if this affects downstream, or if post_root_validators simply is for rendering in the UI (which I doubt)

ahuang11:
I think we need a `SecretSetting` class

PR #6124, opened 2022-07-29
Better error handling for errors in the telemetry service

cicdw:
### Summary

When the telemetry service is unavailable, the errors can look really messy and repeat consistently.  Because this doesn't affect application behavior, this PR shuts down the telemetry service when failing to ping.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


dylanbhughes (approved):

chrisguidry (approved):
That looks right to me.  We _could_ setup a timer for it to eventually try again later, but this is a good stopgap

jlowin (dismissed):
I don't think this will work -- the `stop()` method blocks until the service actually exits, so I think calling it from inside the service will infinite loop.

jlowin:
@cicdw in https://github.com/PrefectHQ/prefect/pull/6131 I added a kwarg it so you can safely call `await self.stop(block=True)` from inside the service

src/prefect/orion/services/telemetry.py
```diff
@@ -101,17 +102,21 @@ async def run_once(self):
...
+            # this looks like an impactful server error
+            self.logger.error(
+                "Failed to send telemetry.", exc_info=PREFECT_DEBUG_MODE.value()
             )
-
-            try:
-                result.raise_for_status()
-            except Exception:
-                self.logger.exception("Failed to send telemetry.")
+            await self.stop()
```
madkinsz, line 119:
```suggestion
            await self.stop(block=False)
```

madkinsz:
Can we add a test for this?

madkinsz:
@cicdw I've finished this, @dylanbhughes please review again

dylanbhughes (approved):

PR #6131, opened 2022-07-29
Clean up loop service methods

jlowin:
### Summary
The `LoopService` class has a few public methods that have ambiguous names, most notably `shutdown()` (which should not be called directly and does not shut the service down).

This PR changes `setup()` and `shutdown()` to `_on_start()` and `_on_stop()` to indicate that they are private event handlers, not public interfaces.

In addition, the `.stop()` method (which *does* stop the service) currently blocks until the service stops. Unfortunately this means it can not be called from *inside* a service, but that's not obvious. This PR adds a `block` kwarg to `stop()` (defaults to `True`) that makes it possible to disable the block behavior. This means services can safely call `.stop(block=False)` from their own `run()` methods.

Lastly, there was some confusion about the role of the `should_stop` flag, which is reset to `False` in the `_on_stop()` method. This seems odd, but is used by the `.stop()` method to see if the service has actually stopped. Therefore, this PR adds an additional behavior flag, `_is_running`, which the `.stop()` method monitors instead. Now, `_on_stop()` sets `_is_running=False` which is a much more obvious behavior. (Note: this PR also makes `_should_stop` private).

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


src/prefect/orion/services/loop_service.py
```diff
@@ -103,39 +107,47 @@ async def start(self, loops=None) -> None:
...
+            while pendulum.now("UTC") < next_run and not self._should_stop:
                 await asyncio.sleep(
                     min(1, (next_run - pendulum.now("UTC")).total_seconds())
                 )
 
-        await self.shutdown()
+        await self._on_stop()
 
-    async def stop(self) -> None:
+    async def stop(self, block=True) -> None:
```
cicdw, line 118: IMO the default should make this safe to call without understanding, but I'm not as familiar with all the places this code is used.

chrisguidry (approved):
Very nice.  I think it's clearer when you're both inside and outside the service; there's only `.start` and `.stop` for either case.

src/prefect/orion/services/loop_service.py
chrisguidry: I think the _typical_ use for this (do we even know that?) is to `await service.stop()` from _outside_ I think.

jlowin: I made this change -- I think in looking at the codebase, services are almost always run as long-running services and Prefect itself is really the only actor that needs to start / stop them. However, developers of services may want to call `.stop()` to create an early stopping behavior (as @cicdw did)

madkinsz:
Thoughts on `stop` and `stop_nowait`? The second does not need to be an async method which I think would make usage even clearer.

jlowin:
@madkinsz I like that idea

src/prefect/orion/services/loop_service.py
```diff
@@ -103,39 +107,68 @@ async def start(self, loops=None) -> None:
...
+
+                wait_time = (pendulum.now("UTC") - start_time).total_seconds()
+
+                # if we've been waiting more than 2 seconds, don't check so frequently
+                if wait_time > 2:
+                    sleep_time = 0.5
+
+                # if we've been waiting longer than `loop_seconds`, something's wrong
+                # log a warning and exit
+                if wait_time > self.loop_seconds:
```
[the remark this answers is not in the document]
madkinsz: In the long run it'd be good to wait on an event there instead of sleeping which is more efficient but this is nice and simple for now. Thanks!

PR #6134, opened 2022-07-29
add index on flow_run.infrastructure_document_id

jakekaplan:
### Summary
Adds an index on `flow_run.infrastructure_document_id`. This is helpful for increasing delete performance of blocks which reference this column via foreign key, which will resolve: https://github.com/PrefectHQ/nebula/issues/1269

### Steps Taken to QA Changes
I ran migrations locally on both sqlite and postgres

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


chrisguidry (approved):
Excellent diagnostics!

peytonrunyan (approved):

src/prefect/orion/database/migrations/versions/postgresql/2022_07_29_181713_fa985d474982_add_index_to_flow_run_infrastructure_.py
```diff
@@ -0,0 +1,33 @@
...
+# revision identifiers, used by Alembic.
+revision = "fa985d474982"
+down_revision = "add97ce1937d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_index(
```
jlowin, line 19: For Postgres I think this should be a concurrent index creation. It’s probable that no one using OSS has enough rows that this would take any time at all but for example in Cloud2 this would lock the Orion tables

jakekaplan: good call I added concurrent creation here

PR #6135, opened 2022-07-29
Makes the agent's inner loop more robust in the face of environmental failures

chrisguidry:
### Summary
This takes the approach of remembering the recent track record for a service,
so that momentary blips won't cause it to exit prematurely, but an ongoing
series of problems that might indicate a deeper issue (network, server, or cloud
outage) will.

Big thanks to @anticorrelator for help with the statistical underpinning here, it turned out very simple.  Math is real!

### Steps Taken to QA Changes

(none yet, this is a draft)

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/cli/agent.py
+++ b/src/prefect/cli/agent.py
@@ -1,13 +1,9 @@
 """
 Command line interface for working with agent services
 """
-import traceback
-from collections import deque
-from typing import Callable, Coroutine, Deque, List, Optional
+from typing import List
 from uuid import UUID
 
-import anyio
-import httpx
 import typer
 
 from prefect.agent import OrionAgent
@@ -17,7 +13,7 @@ from prefect.cli.root import app
 from prefect.client import get_client
 from prefect.exceptions import ObjectAlreadyExists
 from prefect.settings import PREFECT_AGENT_QUERY_INTERVAL, PREFECT_API_URL
-from prefect.utilities.collections import distinct
+from prefect.utilities.services import critical_service_loop
 
 agent_app = PrefectTyper(
     name="agent", help="Commands for starting and interacting with agent processes."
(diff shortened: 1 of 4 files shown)


src/prefect/cli/agent.py
```diff
@@ -97,15 +101,82 @@ async def start(
...
+                track_record.append(-1)
+                exceptions.append(exc)
+            else:
+                raise
+        except KeyboardInterrupt:
+            return
+
+        # If the last result was a failure, see if it's time to cut our losses because
+        # we're failing more often than we're succeeding
+        if track_record[-1] < 0 and sum(track_record) < 0:
```
chrisguidry, line 164: `sum(track_record)` is essentially saying "more than half", which is probably something we'd like to tune?

jlowin:
Neat idea, I wonder if we could bake it into the base `LoopService` class as well (and I'm reminded of an old urge to implement agents as loop services...)

abrookins (commented):
There is some interesting prior art in CS that we might be able to leverage (maybe as future changes) for this called Phi Accrual Failure Detection.

chrisguidry:
> I wonder if we could bake it into the base LoopService class as well (and I'm reminded of an old urge to implement agents as loop services...)

@jlowin yes!  I pulled it out as an independent function so we could use it in more places, with `LoopService` in mind for sure.  I hadn't thought of making the agent a `LoopService`, but that's a solid idea I may pursue too

> Phi Accrual Failure Detection

@abrookins Thank you!  Half the battle is knowing the names of these things, will look into that as the alternative.  One weakness of my naive approach here is it's fairly harsh when it does decide to quit: if you encounter an error and the last 50%+ tries were errors, exit.  That determinism might be a strength in that it's simpler and easier to explain, but I bet Phi Accrual's probabilistic approach is a better/smoother user experience

chrisguidry:
I took a dive this morning on the Phi Accrual method, which is very cool but doesn't quite fit the problem here.  That method is heavily dependent on the _time_ between heartbeats, with the assumption that we're trying to detect something like an unreachable or crashed process on the other end.  It assumes the latency of the responses will get longer and longer until we can confidently say the other end of the communication is gone.  That's not quite the input we have here, but it got me thinking and reading about statistical approaches to problems like we are trying to deal with here.

I believe what we're trying to do here is _predict the probability that the next loop iteration is going to fail_ based on the recent history of successes and failures.  This immediately leads us to the [Binomial Distribution](https://en.wikipedia.org/wiki/Binomial_distribution), with each loop iteration being a [Bernoulli trial](https://en.wikipedia.org/wiki/Bernoulli_trial) (a trial with one of two outcomes, success or failure).

That seems like a great fit here, but I have one big question: the binomial distribution assumes that the trials are _independent_ of one another; is that true of what we're predicting here?  If we're trying to smooth out our response to random network glitches and temporary brief outages of the Orion server or Cloud, I think we can consider those independent.  Is that a safe assumption? 

@abrookins Thanks for getting me to think about the prior art!  Even though the Phi Accrual isn't quite the droid we're looking for, I think there could well be upcoming problems where it applies, like predicting outages in third-party systems by looking at the runtimes of tasks.

peytonrunyan:
This sort of problem would violate the assumptions of a Bernoulli trial. The events are dependent, and the probability of success for a given trial is not fixed. To give a concrete example, `P(Second failed heartbeat | First failed heartbeat)` is going to be higher than `P(First failed heartbeat)`. 

There are some fun options here, like doing an epsilon-greedy algorithm, but it's non-deterministic. 

A decaying average (average weighted by distance from the present) is a probably a good option.

chrisguidry:
That's what I'm wondering, @peytonrunyan: if we're trying to smooth over intermittent network glitches, then maybe they _can_ be treated as independent events?

@bunchesofdonald Yep, a decaying average does seem like a good choice (and Peyton mentioned it too).

chrisguidry:
After a long and fruitful dive on this with @anticorrelator, he helped me see that there is a good statistical basis for just looking at the last N consecutive failures.  I hope I'm explaining it properly in the comments.

madkinsz (approved):

src/prefect/utilities/services.py
```diff
@@ -0,0 +1,94 @@
...
+        # consecutive errors from each.  In the everything-is-fine distribution, that
+        # would be a very low-probability occurrance, but in the everything-is-on-fire
+        # distribution, that is a high-probability occurrance.
+        #
+        # Remarkably, we only need to look back for a small number of consecutive
+        # errors to have reasonable confidence that this is indeed an anomaly.
+        # @anticorrelator and @chrisguidry estimated that we should only need to look
+        # back for 3 consectutive errors.
+        most_recent = list(reversed(track_record))
+        if len(most_recent) > consecutive and all(most_recent[:consecutive]):
```
anticorrelator, line 77: alternatively we could only store the most recent 3 events, this check would be simplified in this case too!

chrisguidry: Yep, this got ugly on me.  Take a look at the current version and let me know what you think?

PR #6136, opened 2022-07-30
Fail login on invalid profile name

anticorrelator:
Empty profile names will produce an invalid `profiles.toml` file, causing errors later on. This PR makes it so we'll simply fail if an empty profile name is provided.

### DIFF
--- a/src/prefect/cli/cloud.py
+++ b/src/prefect/cli/cloud.py
@@ -294,6 +294,7 @@ async def login(
         "Creating a profile for this Prefect Cloud login. Please specify a profile name: "
     )
 
+    cloud_profile_name = cloud_profile_name.strip()
     if cloud_profile_name == "":
         exit_with_error("Please provide a non-empty profile name.")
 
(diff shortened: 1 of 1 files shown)


zhen0:
Thank you!

src/prefect/cli/cloud.py
```diff
@@ -294,6 +294,9 @@ async def login(
         "Creating a profile for this Prefect Cloud login. Please specify a profile name: "
     )
 
+    if cloud_profile_name == "":
```
chrisguidry, line 297: Great catch!

This is definitely an edge case, but we might want to strip the user's input of leading/trailing spaces:

```suggestion
    cloud_profile_name = cloud_profile_name.strip()
    if cloud_profile_name == "":
```

chrisguidry (approved):

src/prefect/cli/cloud.py
anticorrelator: great idea!

PR #6138, opened 2022-07-30
Automatically add v1 label to 1.x PRs

jlowin:
### Summary
This action should automatically apply the existing `v1` label to any PR's targeting the `1.x` branch

(I tried proving this out in #6137 by changing the base branch to 1.x but the total lack of shared history terrified github and it forcibly closed the PR)

### DIFF
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,0 +1,2 @@
+# Apply label "v1" if target branch is "1.x"
+'v1': ['1.x']
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,7 +1,7 @@
 name: PR Labeler
-
-# don't only run when opened in case the branch is changed afterward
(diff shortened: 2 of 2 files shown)


.github/pr-labeler.yml
```diff
@@ -0,0 +1,2 @@
+# Apply label "v1" if target branch is "1.x"
+'v1': ['1.x']
```
cicdw, line 2: you shouldn't need these new files - we already have a labeler action, so I think appending this to https://github.com/PrefectHQ/prefect/blob/main/.github/labeler.yml will do the trick 👍

cicdw (changes requested):
I believe that some updated config is all that's needed here to achieve the same outcome

jlowin:
Oh that's a great idea -- ok I'll combine both of your suggestions

jlowin:
GitHub freaks out if you change the base branch to 1.x so going to reopen a new PR

PR #6145, opened 2022-07-31
[WIP] Add image to manifest

anna-geller:
This is a proposal to add an `--image` option to the `prefect deployment build` command since the image is often built as part of CI and tagged with GITHUB_SHA, and the YAML file can therefore not be manually adjusted with a new image tag when the build process is done automatically in CI


cicdw:
Closing as we will be revisiting the manifest file more holistically based on feedback

PR #6161, opened 2022-08-01
Remove `deployments.Deployment` and `prefect deployment preview`

bunchesofdonald:
### Summary
Now that Deployments are entirely YAML based the `Deployment` class is both broken and unused. This removes that class as well as the related `prefect deployment preview` command.

### Steps Taken to QA Changes
Ran the test suite and all tests pass locally.


cicdw (approved):
LGTM and FYI we will be re-wiring up the packager logic to deployments soon

Issue #6110, opened 2022-07-29
Renaming Active Profile should change profile config


wangjoshuah:
In prefect 2.0, if you rename your active profile, the CLI will keep a reference to the old profile name, and it will not be able to be found.

Repro Steps:

```bash
prefect profile create original_name
prefect profile rename original_name new_name
prefect cloud workspace ls
```

Current workaround is to just remember to use the new profile.

PR #6164, opened 2022-08-01
Switch the name of the active profile on rename if in use

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6110 by changing the active profile in the profiles file if it is the one that was renamed. Includes a warning if you've renamed an active profile set by environment variable.

### DIFF
--- a/src/prefect/cli/profile.py
+++ b/src/prefect/cli/profile.py
@@ -1,6 +1,7 @@
 """
 Command line interface for working with profiles.
 """
+import os
 import textwrap
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


src/prefect/cli/profile.py
```diff
@@ -251,6 +252,17 @@ def rename(name: str, new_name: str):
     profiles.add_profile(profiles[name].copy(update={"name": new_name}))
     profiles.remove_profile(name)
 
+    # Switch the active profile to the new name if provided
```
bunchesofdonald, line 255: I find this comment a bit confusing, maybe:

```suggestion
    # If the active profile was renamed switch the active profile to the new name.
```

?

bunchesofdonald (approved):

src/prefect/cli/profile.py
madkinsz: Yeah idk what I was typing there, thanks

PR #6172, opened 2022-08-02
Deployment version

cicdw:
### Summary

Wires up deployment versions to the client on the heels of #6170

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -1,51 +1,5 @@
 """
 Objects for specifying deployments and utilities for loading flows from deployments.
-
-Deployments can be defined with the `Deployment` object.
-
-To use your deployment, it must be registered with the API. The `Deployment.create()`
-method can be used, or the `prefect deployment create` CLI.
-
-Examples:
-    Define a flow
-    >>> from prefect import flow
-    >>> @flow
-    >>> def hello_world(name="world"):
-    >>>     print(f"Hello, {name}!")
-
-    Write a deployment that sets a new parameter default
-    >>> from prefect.deployments import Deployment
-    >>> Deployment(
-    >>>     flow=hello_world,
-    >>>     name="my-first-deployment",
-    >>>     parameters={"name": "Earth"},
-    >>>     tags=["foo", "bar"],
-    >>> )
-
-    Add a schedule to the deployment to run the flow hourly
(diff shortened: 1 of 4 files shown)


anna-geller:
Works like a charm!

image

image

anna-geller:
@cicdw should we loop in Nicholas to add this field to the UI as well? would be really nice for troubleshooting - imagine you can immediately check from the UI the version indicating your Git commit and then check the corresponding commit in your repo when something goes wrong

anna-geller (approved):

src/prefect/deployments.py
```diff
@@ -1,51 +1,5 @@
 """
 Objects for specifying deployments and utilities for loading flows from deployments.
-
```
chrisguidry, line 3: Yup

chrisguidry (approved):
Looks good to me if we are only considering `version` as an additional piece of user-provided metadata on the `Flow`, and not changing the cardinality of `Flows`.

I do wonder a bit about the user expectation.  Would people expect there to be multiple `Flows` at a time, with different versions?  That definitely blows up the scope of this change, so we probably shouldn't go there yet.  Just something to think about, we may want to have some explanation in the docs about `version`.

I think we could make `version` more valuable without a lot of effort if we add `flow_version` to `FlowRun`, so when you look back at past runs, you see which version of the flow was running at that time.

src/prefect/deployments.py
anna-geller: @chrisguidry I would expect that after adding a version on a deployment, users might leverage that for versioning their deployed flows (e.g. with the version being GITHUB_SHA) and will use the version set on the flow decorator only to version flows triggered without deployments (e.g., triggered locally or e.g. from Amazon Lambda function)

afaik, we currently don't expose this version anywhere in the UI or CLI, so currently, we only use this attribute internally without exposing it in any way to users

cicdw:
@chrisguidry the `version` provided on the flow decorator is already being used to populate a `version` field on the flow run table 👍

madkinsz (approved):

chrisguidry:
> @chrisguidry the `version` provided on the flow decorator is already being used to populate a `version` field on the flow run table 👍 

Ohhhh, thank you, that sounds good

PR #6191, opened 2022-08-02
Create a local 1.x branch before we check if the commit exists in it

### LINKED ISSUE #6190: CircleCI is skipping builds for commits to our 1.x branch
### Describe the bug

When we merge a commit into the 1.x branch, CircleCI should build docker images. However, it now skips these builds because the branch doesn't have associated PRs.

### To Reproduce

```python
An example build: https://app.circleci.com/pipelines/github/PrefectHQ/prefect/12364/workflows/55e6ac7c-e177-42df-849c-72dfe85c3ab6/jobs/150758

You can also run:
git checkout prefect
git branch --contains aa60df895e2d0cc86d8c6792dd0184d6b92a90e0 --points-at 1.x | grep 1.x | wc -l
error: malformed object name '1.x'
       0
```

### Error

```
git branch --contains aa60df895e2d0cc86d8c6792dd0184d6b92a90e0 --points-at 1.x | grep 1.x | wc -l
error: malformed object name '1.x'
       0
``` 

### Versions

- OS: Alpine Linux
- Python: NA
- Prefect: 1.x

### Additional context

_No response_

abrookins:
### Summary
Attempts to fix our CircleCI commit check now that the default branch is "main." Previously, we cloned the repo and tested if a commit was in the "master" branch. Now, we need to check if the commit is in the "1.x" branch, but because that isn't the default branch, we need to create it locally before we check the commit.

Fixes #6190 

### Steps Taken to QA Changes

I manually ran the git commands in our CircleCI config locally until I could reproduce the error (I had to clone a fresh copy of `prefect`, which finally reproduced the issue). Then I ran the command `git branch 1.x origin/1.x` and verified that the error did not occur.

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,7 +345,7 @@ jobs:
           name: 1.x branch check
           command: |
             apk add git
-            git checkout 1.x
+            git branch 1.x origin/1.x
             if [[ $(git branch --contains $CIRCLE_SHA1 --points-at 1.x | grep 1.x | wc -l) -ne 1 ]]; then
               echo "commit $CIRCLE_SHA1 is not a member of the 1.x branch"
               exit 1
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
Thanks for digging into this !!

PR #6201, opened 2022-08-03
Add Flow Run Name to k8s job spec

### LINKED ISSUE #5992: Include flow run name in the metadata on the kubernetes job spec
## Current behavior
Currently the Kubernetes Job Spec generated by the agent only includes the following labels:

        k8s_labels = {
            "prefect.io/identifier": identifier,
            "prefect.io/flow_run_id": flow_run.id,  # type: ignore
            "prefect.io/flow_id": flow_run.flow.id,  # type: ignore
        }

## Proposed behavior
Include the Flow run name in the K8s labels for the Job Spec.

masonmenges:
Adds the Flow Run Name to the k8s job spec for logging purposes

### Summary
Some users want to be able to retain their logs outside of prefect 1.0, currently the K8s job spec includes the Flow Id, Flow Run Id and Flow Identifier. This just introduces the Flow run name into the job spec as well and shouldn't affect most users.

Closes #5992 

### Checklist

This pull request is:

- [] A documentation / typographical error fix
	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -486,6 +486,7 @@ class KubernetesAgent(Agent):
             "prefect.io/identifier": identifier,
             "prefect.io/flow_run_id": flow_run.id,  # type: ignore
             "prefect.io/flow_id": flow_run.flow.id,  # type: ignore
+            "prefect.io/flow_run_name": flow_run.name,  # type: ignore
         }
         _get_or_create(job, "metadata.labels")
         _get_or_create(job, "spec.template.metadata.labels")


madkinsz:
I'm hesitant to accept this since the flow run name is mutable. We even commonly recommend customizing the name by updating the flow run name after the job has started. I worry this will lead to mismatched expectations about the value of this field, especially for auto-scheduled flow runs.

On a separate note, if we included this you may need to slugify the flow run name to ensure it meets Kubernetes character limits. See https://github.com/PrefectHQ/prefect/blob/main/src/prefect/infrastructure/kubernetes.py#L460-L465 for example.

masonmenges:
@madkinsz That definitely makes sense, I can look at this a bit more in detail later, this was requested by a client but if it doesn't make sense to include then we can go a different direction.

PR #6205, opened 2022-08-03
Use the correct type annotations for PrefectBaseModel.subclass()

chrisguidry:
### Summary
The `PrefectBaseModel.subclass` method (and the underlying `pydantic_subclass` function) say they take _instances_ of `BaseModel` but they actually take _types_ derived from `BaseModel`.

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/orion/utilities/schemas.py
+++ b/src/prefect/orion/utilities/schemas.py
@@ -31,12 +31,16 @@ class DateTimeTZ(pendulum.DateTime):
             raise ValueError("Unrecognized datetime.")
 
 
+B = TypeVar("B", bound=BaseModel)
+S = TypeVar("S", bound=BaseModel)
+
+
 def pydantic_subclass(
-    base: Type[BaseModel],
+    base: Type[B],
     name: str = None,
     include_fields: List[str] = None,
     exclude_fields: List[str] = None,
-) -> Type[BaseModel]:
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
You could even improve this with a type variable with a bound of `BaseModel` allowing the created subclasses to retain their parent type.

src/prefect/orion/utilities/schemas.py
```diff
@@ -31,12 +31,16 @@ def validate(cls, v):
...
+
+
 def pydantic_subclass(
-    base: Type[BaseModel],
+    base: Type[B],
     name: str = None,
     include_fields: List[str] = None,
     exclude_fields: List[str] = None,
-) -> Type[BaseModel]:
+) -> Type[S]:
```
madkinsz, line 43: This actually won't retain the type information that we care about. This is basically equivalent to what you had before. Since we're returning a _subtype_ of `B` you can just use `-> Type[B]`. It's a bit of a lie since we're modifying the fields but we do inherit from the base type still.

chrisguidry: Good catch, how's this version look?

madkinsz (approved):

PR #6206, opened 2022-08-03
 Improve sanitization of labels for Kubernetes Jobs

madkinsz:
Previously, we were not sanitizing the key at all and there was no test coverage for this.

Here, I separate handling of keys and values and sanitize both. Additionally, there's now a fallback if the slugified content is empty to the original content which will let Kubernetes throw an informative error about the bad label.

## Future work

- Consider throwing a local error when slugification results in empty content
- Add checks/enforcement of starting and ending with alphanumeric characters


chrisguidry (approved):
Excellent!

PR #6211, opened 2022-08-03
Chore: Remove @prefecthq/miter-design

pleek91:
# Description
We haven't used this deprecated library in a while. Finally getting around to removing it.


znicholasbrown (approved):
RIP

PR #5394, opened 2022-01-31
[WIP] Add multi platform support for prefect docker image

neumann-nico:
## Summary

Currently only x64 prefect images are build and distributed in dockerhub. This PR adds support for arm64.

## Changes

Changes the Travis CI to build both x64 and arm64 images.

## TODO
Currently the x64 image is build, tagged, tested and only then pushed.
Using `docker buildx` for multi platform support does not allow to `--load` images to test and push images later. 

Multi platform builds are required to use the `--push` flag which directly pushes to the docker registry.
There should be a workaround to test the image before pushing it. Issue to the problem: https://github.com/docker/buildx/issues/59

## Importance

More and more arm systems are used both in desktop and server environment. Running native images on those systems improves the performance significantly.

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)
- [ ] updates docstrings for any new functions or function arguments, including `docs/outline.toml` for API reference docs (if appropriate)

Issue #5388, opened 2022-01-30
prefecthq/prefect docker image: arm64 support


neumann-nico:
## Current behavior
The current prefecthq/prefect docker images are only build for linux/amd64.

## Proposed behavior
It would be great to have a multi plattform image supporting arm64.

## Example
Systems running arm64 need to emulate x86_64 images which makes it slower to run. So native support would be awesome!

neumann-nico:
I created a PR which is unfortunately not working yet: https://github.com/PrefectHQ/prefect/pull/5394

ddelange:
Just noting that spinning up the prefect helm chart on an ARM cluster has more blockers than just `prefecthq/prefect`, none of these have `linux/arm64` images:

- docker.io/hasura/graphql-engine:v2.0.9
- docker.io/prefecthq/apollo:2022.01.25
- docker.io/prefecthq/prefect:1.0rc0
- docker.io/prefecthq/server:2022.01.25
- docker.io/prefecthq/ui:2022-01-14

neumann-nico:
@ddelange Yes you are right, there are more images that need to be updated to arm64.
* `ui` and `appollo` use `node` images which are arm64 compatible
* `prefect` and `server` use `python` images which are arm64 compatible
* newer `graphql-engine` images are arm64 compatible

I would be happy to help with the arm64 support

ddelange:
From the helm side, it's also cool if we can set (global/per-service) `nodeSelector` and `tolerations`.

In our case, we're running a mixed amd/arm k8s cluster with hard taints on the arm nodes, to avoid scheduling incompatible services on the arm nodes.

ddelange:
For multi-arch docker manifests, you cant use the `docker pull; docker tag; docker push` pattern that this repo [currently](https://github.com/PrefectHQ/prefect/blob/1.2.4/.circleci/config.yml#L10) uses to re-tag the `ui, apollo, server` images.

The multi-arch way is using [`regctl`](https://github.com/regclient/regclient):
```
curl -L https://github.com/regclient/regclient/releases/latest/download/regctl-linux-amd64 >regctl
chmod 755 regctl
regctl image copy prefect/server:master prefect/server:42
```

ddelange:
@madkinsz now that orion is GA, do you have an update on when this will be uncommented? :)

https://github.com/PrefectHQ/prefect/blob/2.3.1/.github/workflows/release.yaml#L127-L129

ddelange:
[...]

PR #6216, opened 2022-08-03
Add multiplatform Docker image builds

### LINKED ISSUE #5732: Consider adding arm64 base images for Prefect 2.0

**tom115**: Hello! I’ve got a load of `invalid duration format` errors showing up in my Orion API, just checking if this is a bug or misconfig on my part?

**tom115**: 
```
Version:             2.0b3
API version:         0.3.0
Python version:      3.9.12
Git commit:          58a401bc
Built:               Wed, Apr 13, 2022 11:21 AM
Profile:             default
Server:
  Database:          postgresql
```

**anna**: to check if this is a misconfiguration or a bug, when exactly did you see this error showing up in your API logs - when you triggered a specific run, you said it might be when viewing flow runs from the UI, could you confirm that? could help troubleshoot

if this happens when running a flow, could you share flow code or deployment spec?

**tom115**: It’s definitely not when running a flow, I can confirm that part

**anna**: so far I don't have enough info to reproduce or say anything helpful to fix the issue

**tom115**: One moment, I’ll see if I can reliably get it to happen.

**anna**: that would be helpful, thanks a lot! If I can't reproduce or help further, I'll open an issue and ask the team

**tom115**: Right, I just refreshed the UI three times with the flow runs tab selected and had it appear each time.

**tom115**: I’m not sure how to delete flow runs, if I can do that I can nuke them all and see if it still happens

**anna**: so it looks like if I start ephemeral Orion with Postgres and view the UI anywhere, I should be able to reproduce?

**anna**: sorry, but I think the only reliable way to nuke old flow runs would be to reset the DB entirely

**tom115**: Okay. I can do that. Let me gather more info from the current state first

**tom115**: Dammit, hadn’t realised nuking would wipe the different storage options

**tom115**: As in the fixtures, rather than the user configured ones

**tom115**: Well, refreshing now it’s totally nuked no longer causes the error, so that’s good info.

**anna**: nice work!

**tom115**: I want to recreate my flows and have some that aren’t pending to see if that’s the cause but I don’t know how to set storage up again without that seed data. I’d have thought DB recreation would have included those entries but apparently not.

**anna**: sorry to hear that you lost some of your configurations. we tried to make the storage setup easy with the CLI

**tom115**: Oh no, it’s not my storage config I’m worried about, that’s dead easy

**anna**: some engineers are working on adding storage to the DeploymentSpec so if that's of any consolidation, you won't need to rely on a global storage config in the near future

**tom115**: it’s the storage configs that come baked into the DB that allow you to set storage up

**tom115**: Without those the create storage command won’t work, it just tells me to select an option from an empty list

**anna**: Can you make sure that all your browser windows with Orion are closed before doing that?

**anna**: I had a similar issue - closing all browser tabs, then resetting DB and starting orion/creating storage should work. We have an open issue for that

**tom115**: &gt;  some engineers are working on adding storage to the DeploymentSpec so if that’s of any consolidation, you won’t need to rely on a global storage config in the near future
This will be awesome. Storage has been by far the hardest part to fully automate deployment for.

**anna**: 100% agreed

**tom115**: Hmmm, something isn’t right here. I closed all UI windows and reset the DB again but via the CLI

**tom115**: it confirmed reset against local SQLite, but I’m using a kubernetes cluster and PostgreSQL

**anna**: I actually meant that via CLI reset command

**tom115**: Checked config and the API URL is still set to the proper location, so it should be looking at/interacting with the deployed version.

**anna**: in that case you would need to recreate a Kubernetes deployment?

**tom115**: I’ll try doing that, it seemed to still be running happily post reset, but can’t have been quite as happy as it seemed

**tom115**: Well I didn’t cover myself in glory there. Was bouncing the wrong pod over and over wondering why nothing was happening.

**anna**: still, looks like everything is working now?

**anna**: I'll be away for a bit now, LMK if you still have any issues, I can check later

**tom115**: Okay, everything is restored and I have deployments back in. Refreshing isn’t causing the error. Let’s test my theory

**tom115**: No worries. I’ll leave a note so people can debug. It’s not urgent on my side really, I just wanted cleaner logs to debug other stuff.

**tom115**: • Tried refreshing UI with one failed flow run in place. Exception didn’t occur
• Tried refreshing UI with failed and pending flow run in place. Exception didn’t occur
• Error has eventually returned after a period of letting the system run (schedule a few flow runs, normal UI usage etc)
• I can now trigger the error reliably again by refreshing the UI
Sorry, this isn’t a great starting point for debugging but I can’t seem to pin down a reliable set of steps to get into this state. Only config I added post reset was storage. I also have `PREFECT_API_URL` and `PREFECT_AGENT_QUERY_INTERVAL` env vars set.

**anna**: I could try to replicate with Kubernetes setup but just to let you know, if this doesn't work for you, you always can try those two alternatives:
• running Orion without Kubernetes - e.g. on a local instance you could set it up <https://discourse.prefect.io/t/getting-started-with-cloud-2-0-using-ec2-instance-as-the-execution-layer/551|on EC2>
• using <https://discourse.prefect.io/t/how-to-get-started-with-prefect-cloud-2-0/539|Cloud 2.0>
How did you set up your Orion instance on Kubernetes? Did you follow <https://orion-docs.prefect.io/tutorials/kubernetes-flow-runner/|this tutorial here>?

**tom115**: I did follow that tutorial. Everything seems to work, it’s just there’s stacks of errors in the logs that make it hard to find errors I care about.

**tom115**: Mostly timeouts and then the ones I describe above. I figure the timeouts are fairly self explanatory, but the invalid duration seemed like it might be a bug

**tom115**: I wouldn’t want to use a different deployment mechanism like EC2 as that’s going to bring down a lot of additional maintenance burden. I’d be okay with using cloud (I’m considering it now), but I’d still need to know I could run it successfully on my own stack first as a safety net since it’s such a core part of our system.

**tom115**: That safety net isn’t just hypothetical either. There’s a decent chance our product will need to be self-hostable for an enterprise version in future. We’d be able to use the cloud offering for the main SaaS solution, but couldn’t use it for those self-hosted instances.

**anna**: I can 100% understand that, and I'm sure this is a solvable problem. Hard to say what exactly went wrong there, but I can reassure you that we build Prefect 2.0 in a way that you can use the OSS product for your production deployments.

**tom115**: Oh yeah for sure, I don’t mind there being some teething problems. You’re pretty up front about this being a beta and the product has lots of features that offset that

**tom115**: for a chunk of the problems I’ve faced I’m hoping to wrap them up in a kubernetes operator and make it open source so others can pick it up, I just want to make sure I’ve got the right solutions internally before I do that.

**tom115**: I’d also need to wait on the new storage interface as I wouldn’t want to hack around the current one and then immediately replace it. Everything else is pretty workable as-is.

**anna**: Nice! If you want a sneak peek of the storage interface, it's based on the `fsspec` interface which provides a lot of flexibility in that regard - if you want to check this: <https://github.com/PrefectHQ/prefect/blob/orion/src/prefect/blocks/storage.py>

**tom115**: <@U02H1A95XDW> I might have just found the issue, mea culpa. I’m on an M1 Mac locally, so it’s emulating. I only just noticed prefect doesn’t have an arm64 image. My experience of emulation with docker is that it’s pretty ropey, and the errors you can get are… weird. Are there any plans for an arm build of the image?

**tom115**: I can build a local version in the meantime, but would be awesome if the official one had both.

**anna**: Nice to hear you found out the root cause of the issue, good work!

Let me open an issue for that.

<@ULVA73B9P> open "Consider adding arm64 base images for Prefect 2.0"

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1651178497293389?thread_ts=1651178497.293389&cid=CL09KU1K7).

madkinsz:
### Summary

Adds Docker image builds for the following platforms
- linux/amd64
- linux/arm/v7
- linux/386
- windows/amd64

For example of additional available platforms, see https://hub.docker.com/_/python?tab=tags

Related to https://github.com/PrefectHQ/prefect/issues/5388 https://github.com/PrefectHQ/prefect/pull/5394 https://github.com/PrefectHQ/prefect/pull/5408

Closes #5732

### Steps Taken to QA Changes

Temporarily running these builds during normal Python test runs.

**Happy engineering!**

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -75,6 +75,9 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
       - name: Set up Docker Buildx
         if: ${{ matrix.build-docker-images }}
         uses: docker/setup-buildx-action@v2
@@ -119,6 +122,8 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
+          # TODO: Remove after testing is complete; only releases need multiplatform builds
+          platforms: linux/amd64,linux/arm/v7,linux/386,windows/amd64
           build-args: |
             PYTHON_VERSION=${{ matrix.python-version }}
             BASE_IMAGE=prefect-conda
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,9 +122,8 @@ jobs:
           - "3.10"
 
     steps:
-      # Not needed until we want multiarchecture builds
-      # - name: Set up QEMU
-      #   uses: docker/setup-qemu-action@v1
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
@@ -152,6 +151,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
+          platforms: linux/amd64,linux/arm/v7,linux/386,windows/amd64
           build-args: |
             PYTHON_VERSION=${{ matrix.python-version }}
             BASE_IMAGE=prefect-conda


madkinsz:
Oof, since our builds and pushes are separated this is a huge can of worms.

See https://github.com/docker/buildx/issues/59 and related. Closing for now as this is going to be a nontrivial project.

ddelange:
hi @madkinsz 👋

it looks like this PR only targets the `-conda` suffix images: where would the changes be needed for the 'normal' (smaller) python3x images?

im looking for a minimally intrusive way to get this done within your release flow

Issue #6214, opened 2022-08-03
Create_flow_run missing flow version error


rjhawar:
### Describe the bug

The create_flow_run takes in a flow object however since a flow object doesn't contain a "flow_version" field the method will always error out.

### To Reproduce

```python
from prefect.client import get_client
import asyncio

async def main():
    async with get_client() as client:
            flow_model = await client.read_flow(<FLOW_ID>)
            flow_run_response = await client.create_flow_run(flow=flow_model)

loop = asyncio.get_event_loop()
loop.run_until_complete(main())
```

### Error

```
# Copy complete stack trace and error message here, including log output if applicable.

line 592, in create_flow_run
flow_version=flow.version,
AttributeError: 'Flow' object has no attribute 'version'
``` 

### Versions

- OS: 11.6
- Python: 3.9.13
- Prefect: 2.0.2

madkinsz:
This method consumes a `prefect.flows.Flow` not a `prefect.orion.schemas.core.Flow`. I believe you're looking for `OrionClient.create_flow_run_from_deployment`.

rjhawar:
@madkinsz 
Ah ok my apologies there that makes a lot more sense now. 

The reason i wanted to use that method over `OrionClient.create_flow_run_from_deployment` was because it allows you to assign unique name/tags per flow run. The deployment method only allows you change the parameters.

 I can always use `OrionClient.create_flow_run_from_deployment` and then use `update_flow_run` method to change the name and tags, but that is a bit more work. Any chance we can make this easier by adding the ability to add names/tags via the `OrionClient.create_flow_run_from_deployment` method?

madkinsz:
Yep! Just not something we'd needed internally yet.

PR #6217, opened 2022-08-03
Add flow run name and tag support to `create_flow_run_from_deployment`

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6214

These are already supported by the API but were excluded from the client implementation.


bunchesofdonald (approved):
That should do it.

cicdw (approved):

PR #6226, opened 2022-08-04
[PROPOSAL] Add remote file system dependencies

anna-geller:
### Summary
Currently, both users and engineers doing QA have to build custom images to build deployments using any remote file system blocks because dependencies are not included in the base image.

Adding this as PREFECT_EXTRAS might be a better place to add that, but I didn't know how.

This pull request is:


madkinsz:
I think we'll need to have a discussion about why these implementations are in the core library now. The clients for interacting with these services have a lot of dependencies and are quite heavy. I definitely do not want to include them in the base image if we can avoid it. 

You do not have to build a custom image to install them at runtime, you can use https://docs.prefect.io/concepts/infrastructure/#installing-extra-dependencies-at-runtime

anna-geller:
> I think we'll need to have a discussion about why these implementations are in the core library now. The clients for interacting with these services have a lot of dependencies and are quite heavy. I definitely do not want to include them in the base image if we can avoid it.
> 
> You do not have to build a custom image to install them at runtime, you can use https://docs.prefect.io/concepts/infrastructure/#installing-extra-dependencies-at-runtime

True, but it feels weird to be installing things at runtime, it roughly adds 35-seconds latency to each flow run. But I agree with you that this is the best from all tradeoffs -- let's close the issue in favor of adding docs on how users can add those dependencies to their images https://github.com/PrefectHQ/prefect/pull/6245

PR #6227, opened 2022-08-04
Allow for multiple deployment YAML paths in apply CLI

cicdw:
### Summary

Allows for multiple paths to be specified and removes the non-working default (**cc:** @zangell44 )

### DIFF
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -258,59 +258,57 @@ def _load_deployments(path: Path, quietly=False) -> PrefectObjectRegistry:
 
 @deployment_app.command()
 async def apply(
-    path: Path = typer.Argument(
-        None,
-        help="The path to a deployment YAML file.",
-        show_default=False,
-    )
+    paths: List[str] = typer.Argument(
+        ...,
+        help="One or more paths to deployment YAML files.",
+    ),
 ):
     """
     Create or update a deployment from a YAML file.
     """
-    if path is None:
-        path = "deployment.yaml"
+    for path in paths:
 
-    # load the file
-    with open(str(path), "r") as f:
-        data = yaml.safe_load(f)
+        # load the file
+        with open(str(path), "r") as f:
+            data = yaml.safe_load(f)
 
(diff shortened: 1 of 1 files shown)


src/prefect/cli/deployment.py
```diff
@@ -258,60 +258,58 @@ def _load_deployments(path: Path, quietly=False) -> PrefectObjectRegistry:
...
+            deployment = DeploymentYAML(**data)
+            app.console.print(f"Successfully loaded {deployment.name!r}", style="green")
+        except Exception as exc:
+            exit_with_error(f"'{path!s}' did not conform to deployment spec: {exc!r}")
+
+        async with get_client() as client:
+            # prep IDs
+            flow_id = await client.create_flow_from_name(deployment.flow_name)
+
+            deployment.infrastructure = deployment.infrastructure.copy()
```
zangell44, line 286: can someone link deployment infrastructure to a stable block?
im confused here
(no change in logic im just curious)

zangell44 (approved):

src/prefect/cli/deployment.py
madkinsz: I had planned for that to be the case eventually, but needed to set the run command and image tag (for container based runs) at deployment creation time which required a copy. If we have the manifest attached to the deployment now separately, we could probably load those fields from there at runtime instead and just override whatever is on the block to support a stable block. It gets a little confusing though.

madkinsz:
Note if creation of one fails it'll just stop at that failure. I think we should separate the load and apply steps to minimize the chance of this occurring e.g.

instead of

- Load deployment
- Register deployment
... repeat

do

- Load all deployments given, which performs validation on the input YAML definitions
- Register all deployments given

cicdw:
@madkinsz would you like me to address that in this PR or a follow up?

madkinsz:
It sounds like this fixes a broken experience so I'm fine with a follow-up!

anna-geller (approved):
Works!

image

bunchesofdonald (approved):

PR #6239, opened 2022-08-04
Use event instead of sleep, second attempt

a6502:
Replace the use of anyio.sleep(0) with wait()ing on anyio.Event()s

### Summary
In a multithreaded environment the use of polling with anyio.sleep(0) in the main thread causes that thread to claim the GIL on every possible opportunity, thus taking time away from the other threads.
Even when only the main thread is running the use of polling is just a waste of CPU-resources and energy.

This PR is a suggestion as I believe that this PR uncovered some bugs in either the tests of Prefect itself that would need fixing first. (see below)

(See #6004 for the original PR)

### Steps Taken to QA Changes
The tests in tests/test_task_runners.py and tests/utilities/test_asyncutils.py all pass, the latter with a bit of help.

Manual tests with some of my production use cases saw more than 95% reduction in user CPU time.

However, some tests in tests/test_tasks.py fail. After quite a bit of digging it seems that in those failing test the wait() method on the ConcurrentTaskRunner instance is called from a different thread ("ThreadPoolExecutor-%d_%d") than the thread that called
submit() ("MainThread"). This causes an error "Non-thread-safe operation invoked on an event loop other than the current one" from python3.10/asyncio/base_events.py.
As the docstring of _run_and_store_result mentions that it expects to be called from the main thread I believe this to be an error in either pytest or Prefect itself. I hope someone more knowledgeable can explain what is going on here.

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -328,6 +328,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         """
         Block until the run result has been populated.
         """
+        result = None  # result in case of timeout
         with anyio.move_on_after(timeout):
             await self._events[run_key].wait()
             # I would like to use pop to avoid a memory leak here but wait() seems to
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -250,6 +250,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
 
         # Runtime attributes
         self._task_group: TaskGroup = None
+        self._events: Dict[UUID, anyio.Event] = {}
         self._results: Dict[UUID, Any] = {}
         self._run_keys: Set[UUID] = set()
         super().__init__()
@@ -284,6 +285,9 @@ class ConcurrentTaskRunner(BaseTaskRunner):
 
         # Track the keys so we can ensure to gather them later
         self._run_keys.add(run_key)
+        # Going for a minimally invasive change for now but probably
+        # the _run_keys set can go away now: just use _events.keys()
+        self._events[run_key] = anyio.Event()
 
         return PrefectFuture(
             task_run=task_run,
@@ -312,20 +316,26 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         Since this run is occuring on the main thread, we capture exceptions to prevent
         task crashes from crashing the flow run.
         """
+
         try:
             self._results[run_key] = await run_fn(**run_kwargs)
         except BaseException as exc:
             self._results[run_key] = exception_to_crashed_state(exc)
+        finally:
+            self._events[run_key].set() # raise event
 
     async def _get_run_result(self, run_key: str, timeout: float = None):
         """
         Block until the run result has been populated.
         """
         with anyio.move_on_after(timeout):
-            result = self._results.get(run_key)
-            while not result:
-                await anyio.sleep(0)  # yield to other tasks
-                result = self._results.get(run_key)
+            await self._events[run_key].wait()
+            # I would like to use pop to avoid a memory leak here but wait() seems to
+            # be called twice sometimes? Not cleaning up task_run results is going to be
+            # a problem for flows with lots of tasks or large results
+            #self._events.pop(run_key)
+            #result = self._results.pop(run_key)
+            result = self._results[run_key]
         return result
 
     async def _start(self, exit_stack: AsyncExitStack):
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -322,7 +322,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         except BaseException as exc:
             self._results[run_key] = exception_to_crashed_state(exc)
         finally:
-            self._events[run_key].set() # raise event
+            self._events[run_key].set()  # raise event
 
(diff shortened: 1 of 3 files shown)


madkinsz:
Thanks for the follow-up! I'll investigate that thread issue, I knew something weird was lurking in there as I've seen some errors in the past.

madkinsz:
@a6502 it's going to be challenging to ensure that the event belongs to the correct loop here. Can you share a minimal example that is improved by this implementation so I can reproduce?

a6502:
@madkinsz : I'll try to come up with an example early next week, I need to think for a bit what the best example is.

madkinsz:
If you could try changing the concurrent task runner implementation to just call `time.sleep(0)` after a `anyio.sleep(0)` and let me know if that improves performance in your workflow as well, it'd be helpful.

a6502:
Ok, I had an idea on how the illustrate the problem with anyio.sleep(0): a script that does some heavy computations in python in a worker thread:

`
#!/usr/bin/env python3

# std python
from functools import partial
#import hashlib
import random
import string
import time

# pip install
import anyio
import pyaes

#password = "Hello World!"
# A 256 bit (32 byte) key
#key = hashlib.sha256(password.encode()).digest()


plaintext = """
Lorem ipsum dolor sit amet, consectetur adipiscing elit.  In semper purus id
congue dignissim.  Praesent vel varius nisi.  Mauris ultricies ligula sit
amet mi blandit, sed gravida magna vehicula.  Pellentesque efficitur ante
arcu, a aliquet dolor blandit vitae.  Donec a porttitor est, ac euismod est. 
Maecenas sodales justo vitae nulla malesuada elementum.  Donec in elit enim. 
Aenean rutrum fermentum felis non tempor.  Mauris enim enim, pharetra id
maximus sed, luctus eu justo.  Nunc eu nibh interdum, pulvinar nunc eu,
auctor lectus.  Etiam fermentum mi auctor dui faucibus ultricies.  Vivamus
blandit laoreet lacus quis tristique.
"""


def burnsome(key):
    aes = pyaes.AESModeOfOperationCTR(key)
    ciphertext = aes.encrypt(plaintext.encode())

    # The counter mode of operation maintains state, so decryption requires
    # a new instance be created
    aes = pyaes.AESModeOfOperationCTR(key)

    decrypted = aes.decrypt(ciphertext)
    decrypted = decrypted.decode()

    assert(decrypted == plaintext)

def burnmore(name):
    start_time_ns = time.process_time_ns()
    start_counter_ns = time.perf_counter_ns()
   
    for i in range(1000):
        key = ''.join(random.choices(
                string.ascii_uppercase + string.digits,
                k=32
            )).encode()
        burnsome(key)

    end_counter_ns = time.perf_counter_ns()
    end_time_ns = time.process_time_ns()

    elapsed_ns = end_counter_ns - start_counter_ns
    process_time_ns = end_time_ns - start_time_ns

    print(f'test "{name}" elapsed: {elapsed_ns/1E9}s cpu: {process_time_ns/1E9}s')


async def main():
    result = None
    
    async def tosyncburn(name):
        await anyio.to_thread.run_sync(partial(burnmore, name), cancellable=True)
        nonlocal result
        result = True

    async def zerosleeper():
        while not result:
            await anyio.sleep(0)
            
    async def doublezerosleeper():
        while not result:
            await anyio.sleep(0)
            time.sleep(0)

    name = "without anyio.sleep(0)"
    print(f'start test "{name}"')
        tg.start_soon(tosyncburn, name)

    result = None # reset
    name = "with anyio.sleep(0)"
    print(f'start test "{name}"')
        tg.start_soon(zerosleeper)
        tg.start_soon(tosyncburn, name)

    result = None # reset
    name = "with anyio.sleep(0) and time.sleep(0)"
    print(f'start test "{name}"')
        tg.start_soon(doublezerosleeper)
        tg.start_soon(tosyncburn, name)


if __name__ == "__main__":
    anyio.run(main)
`

On my Linux machine this gives:

> start test "without anyio.sleep(0)"
> test "without anyio.sleep(0)" elapsed: 4.652968012s cpu: 4.625242448s
> start test "with anyio.sleep(0)"
> test "with anyio.sleep(0)" elapsed: 11.798294407s cpu: 14.419711149s
> start test "with anyio.sleep(0) and time.sleep(0)"
> test "with anyio.sleep(0) and time.sleep(0)" elapsed: 7.274337091s cpu: 7.916279085s

So adding a `time.sleep(0)` definitely improves thing but avoiding polling in the first place is the superior solution.

I'll give `time.sleep(0) a go on my real-world testcase once I manage to reassemble my test environment.

madkinsz:
Great thanks! That's really helpful.

The primary issue here is that there's multithreaded access to the task runner but an `asyncio.Event` is not thread-safe. We require multithreaded access to the task runner because we support calling a synchronous task from an asynchronous workflow without it blocking the primary event loop. I may be able to create an event primitive that is async _and_ thread safe or find a way to send task submission back to the main thread, but it's pretty tricky.

madkinsz:
Thanks again for your work on this. Hopefully https://github.com/PrefectHQ/prefect/pull/6948 addresses your issue. I hope to solve the locking issue for multithreaded loops more broadly in the future as well.

PR #6132, opened 2022-07-29
Add `dev qa` command to automatically run flows and deployments for QA purposes

peytonrunyan:
### Summary

This PR allows us to use a single command: `prefect dev qa` to:
1. (optionally) create a QA work queue and QA agent
2. run a series of QA flows in a qa directory
3. register deployments in a qa deployments directory
4. run submit those deployments to be run

The command expects that all deployment will be under the `qa/deployments` directory and that the flow name is the same as the deployment directory that it is located under. So the deployment directory `qa/deployments/parameters` will have a main flow called `parameters`.

### Steps Taken to QA Changes

I've run the command under each of the following
- Empty `.prefect` folder
- A `.prefect` folder that already has multiple deployments and flow runs 
- When connected to cloud

It successfully runs all of the flows and deployments under each condition. 

### Checklist

This pull request is:

- [ ] A documentation / typographical error fix
	- No tests or issue needed
- [ ] A short code fix
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a bug report issue 
	- Please include tests. One-line fixes without tests will not be accepted unless it's related to the documentation only.
- [x] A new feature implementation
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a feature enhancement issue 
	- Please include tests
    - Please make sure that your QA steps are both thorough and easy to reproduce by someobdy with limited knowledge of the feature that you are submitting

**Happy engineering!**

PR #6240, opened 2022-08-04
Chore: Integration tests

pleek91:
# Description
Adds a workflow for running integration tests. These tests are tests run on the orion ui and designed to test both the UI itself and its integration with the api.


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          node-version-file: '.nvmrc'
+          cache-dependency-path: '**/package-lock.json'
+        
+      - name: Cache Playwright Browsers
+        id: cache-playwright-browsers
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-browsers
```
madkinsz, line 45: Nice! Looks like this is working well. You may want to retrieve the playwright version in a prior step for the `key`.

pleek91: Yes I gotta figure that piece out. Although I did find this in the playwright repository which makes me wonder if this is sufficient. https://github.com/microsoft/playwright-github-action/actions/runs/91812666/workflow

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,60 @@
+name: Integration Tests
```
madkinsz, line 1: It seems like this should just replace the existing `UI tests` at https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/ui-tests.yml

It'd be nice to continue to include the "UI" prefix on the file and workflow name for clarity.

pleek91: Yes we probably don't need that ui-tests workflow once this is in main as this workflow will fail if it cannot build the ui. 

As far as using the "UI" prefix I'm not sure. This does use the actual api. Which means a test could be broken because something is wrong with the API.

.github/workflows/integration-tests.yml
madkinsz: The [npm example](https://github.com/actions/cache/blob/main/examples.md#java---maven) shows how to use the result of a previous command downstream. Interesting that the docs say you should use the playright version. FWIW the cache in the workflow you linked is a couple years old.

.github/workflows/integration-tests.yml
madkinsz: It's not actually going to run any tests defined outside of the UI code though. I'd expect integration tests to cover the full workflow from Python -> API -> UI 🤷 Perhaps we'll add additional integration-style tests to this workflow in the future though.

pleek91: That's my plan. I'm hoping to leverage @peytonrunyan 's work [here](https://github.com/PrefectHQ/automated-qa) so that when the tests run there is data from the python client visible in the UI and we can assert on that.

.github/workflows/integration-tests.yml
pleek91: Think I've got that working now

.github/workflows/integration-tests.yml
peytonrunyan: This is way more up-to-date fwiw: https://github.com/PrefectHQ/prefect/pull/6132

I'm getting my windows VM setup to test it, and after that it just needs a review.

znicholasbrown (approved):

Issue #5847, opened 2022-05-30
Add Flask support - running a flow from a Flask app currently results in `"MissingContextError: No settings context found"`


anna-geller:
Slack discussion transcript [here in Linen](https://linen.prefect.io/t/427947/I-m-trying-to-start-a-prefect-orion-flow-from-a-flask-route-)

# MRE

## Using Flask
```python
from flask import Flask
from prefect import flow, get_run_logger

app = Flask(__name__)


@flow(name="valentin_flask")
def process_request():
    logger = get_run_logger()
    logger.info("Hello from Flask - received  input: %s")


@app.route("/event", methods=["GET"])
def push_event():
    process_request()
    return {"ok": True}


if __name__ == "__main__":
    app.run(host="0.0.0.0", debug=True, port=8080)
```
returns 
```
  File "/home/val/.pyenv/versions/3.8.12/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/val/.pyenv/versions/3.8.12/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/val/dev/prefect-poc/venv/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/home/val/dev/prefect-poc/venv/lib/python3.8/site-packages/prefect/client.py", line 91, in with_injected_client
    client_context = get_client()
  File "/home/val/dev/prefect-poc/venv/lib/python3.8/site-packages/prefect/client.py", line 101, in get_client
    ctx = prefect.context.get_settings_context()
  File "/home/val/dev/prefect-poc/venv/lib/python3.8/site-packages/prefect/context.py", line 230, in get_settings_context
    raise MissingContextError("No settings context found.")
prefect.exceptions.MissingContextError: No settings context found.
```

## Using FastAPI

The same works fine using FastAPI:
```python
from fastapi import FastAPI
from prefect import flow, get_run_logger

app = FastAPI(title="Valentin App")


@flow(name="valentin_flask")
def process_request():
    logger = get_run_logger()
    logger.info("Hello from Flask - received  input: %s")


@app.get("/event")
def get_event():
    process_request()
    return {"ok": True}
```

b-long:
I've asked [this question in the Prefect Slack community](https://prefect-community.slack.com/archives/CL09KU1K7/p1653047985779569) as well, re-posting here for reach:

[...]

PR #6260, opened 2022-08-05
Fix missing global settings by caching object instead of relying on contextvars

madkinsz:
### Summary

If a flow is called from a different thread than Prefect is imported from, the root settings context is missing. This is because the root settings context is accessed via context variable which is by definition thread independent. Additionally, if the flow was called from a different event loop than the one the module was imported from, the settings context would be missing. To address this, we no longer rely on the context variable. If the context variable does not detect any settings, we return a global variable where we have stashed a `SettingsContext` instance. We initialize this once on module import and it is accessible by all threads and event loops.

Initially, I thought we should just cache the settings context per thread or event loop and reinitialize it where necessary. See [this branch](https://github.com/PrefectHQ/prefect/compare/main...fix-global-settings) for implementation. This also solves the issue but I do not think it is clean or robust as we would need to inject this call everywhere settings are needed.

Closes https://github.com/PrefectHQ/prefect/issues/5847

### Steps Taken to QA Changes

The example in the linked issue was used to confirm functionality.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -78,7 +78,7 @@ if not hasattr(sys, "frozen"):
     sys.meta_path = [Prefect1ImportInterceptor()] + sys.meta_path
 
 
-prefect.context.enter_root_settings_context()
+prefect.context.root_settings_context()
 prefect.context.initialize_object_registry()
 
 # The context needs updated references for flows and tasks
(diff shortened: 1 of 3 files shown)


bunchesofdonald (approved):
Makes sense to me and I agree this is going to be less fragile than trying to cache per thread.

PR #6264, opened 2022-08-05
Change log level to DEBUG for starting task runner

madkinsz:
This log is confusing, particularly for the default task runner. Startup of the task runner is only significant if it takes a long time, e.g. the Dask task runner creating a cluster. However, those task runners should provide additional logs indicating the slow operation. The utility of this log is retained for users that enable a debug level log allowing us to determine where an issue occurs.


anna-geller (approved):
Nice!

discdiver:
This will be a great improvement to reduce confusion and unnecessary output. Thank you @madkinsz!

PR #6267, opened 2022-08-05
Skip showing lines in cov report, migrations, tests

### LINKED ISSUE #6266: Reduce coverage output
### First check

### Prefect Version

2.x

Coverage outputs should not include test files, migrations, or line-level reporting. Coverage reports should include % of coverage per file and only show relevant (non-test, non-migration) files.

The coverage report is currently spammy because so many lines in our codebase lack coverage. The report is also including test files and database migrations, which we likely won't add test coverage for.

### Example Use

_No response_

### Additional context

_No response_

abrookins:
### Summary

Reduce the information printed in our coverage report:
- Do not report on tests
- Do not report on database migrations
- Do not show line-level reporting -- this isn't helpful yet because too many lines lack coverage

Fixes #6266

### Steps Taken to QA Changes

I ran coverage with this set of options, and I liked the result.

### Checklist

This pull request is:

	- No tests or issue needed
- [ x A short code fix
	- Please include tests

**Happy engineering!**


cicdw (approved):

madkinsz:
@abrookins looks like this might require a change to the dockerignore file too? The tree in the built docker images is dirty.

.coveragerc
```diff
@@ -0,0 +1,16 @@
+[run]
+branch = True
+
+[report]
+exclude_lines =
+    # Don't complain about abstract methods, they aren't run:
+    @(abc\.)?abstractmethod
```
chrisguidry, line 7: Whoa

.coveragerc
```diff
@@ -0,0 +1,16 @@
...
+
+omit =
+    src/prefect/orion/database/migrations/versions/*
+    tests/*
```
chrisguidry, line 11: Can we move this to the `pytest` invocation just for CI?  Reporting coverage information about tests helps to identify poor tests and those that are being "masked" like this:

```python
def test_this_super_awesome_thing():
  assert 1 == 2  # this test never runs

def test_this_super_awesome_thing():
  assert 1 == 1
```

I agree we wouldn't be that worried about this in CI, but it would be nice locally and I don't know if you can un-omit something on the CLI.

chrisguidry (approved):

.coveragerc
abrookins: Lifted from the Coverage docs. 😄

.coveragerc
abrookins: Yeah, I'll work on that!

abrookins:
@madkinsz Good catch!

.coveragerc
abrookins: @chrisguidry I couldn't find a way to do this cleanly, so for now I kept reporting on for tests. That wasn't my main concern here anyway!

Issue #6230, opened 2022-08-04
AWS ECR images still don't work with DockerContainer infrastructure blocks


anna-geller:
### Describe the bug

Note: I replaced my account ID with 123456789

```shell
prefect deployment build flows/hello.py:hello --output deployments/hello-docker.yaml --name docker --tag dev -sb s3/dev --infra docker-container
```

### To Reproduce

```python
###
### A complete description of a Prefect Deployment for flow 'hello'
###
name: aws-docker
description: null
version: be1cc83bab9418ead74cdfcd664e746b
tags:
- dev
parameters: {}
schedule: null
infrastructure:
  type: docker-container
  env: {}
  labels: {}
  name: null
  command:
  - python
  - -m
  - prefect.engine
  image: 123456789.dkr.ecr.us-east-1.amazonaws.com/prefect-dataops:latest
  image_pull_policy: null
  networks: []
  network_mode: null
  auto_remove: false
  volumes: []
  stream_output: true
###
### DO NOT EDIT BELOW THIS LINE
###
flow_name: hello
manifest_path: hello-manifest.json
storage:
  bucket_path: prefect-orion/aaa
  aws_access_key_id: '**********'
  aws_secret_access_key: '**********'
  _is_anonymous: true
  _block_document_name: anonymous-3135a42f-2796-4929-92ca-454f4e6d6011
  _block_document_id: e1fab4da-003d-415e-a753-4679a3541483
  _block_type_slug: s3
parameter_openapi_schema:
  title: Parameters
  type: object
  properties:
    user:
      title: user
      default: Marvin
      type: string
  required: null
  definitions: null
```

### Error

```
Agent started! Looking for work from queue 'Agent queue dev'...
08:17:43.786 | INFO    | prefect.agent - Submitting flow run '46e14f5b-94c4-4406-8a39-73b685497c87'
08:17:44.235 | INFO    | prefect.infrastructure.docker-container - Pulling image '38306982838.dkr.ecr.us-east-1.amazonaws.com/prefect-dataops:latest'...
08:17:44.781 | ERROR   | prefect.agent - Flow runner failed to submit flow run '46e14f5b-94c4-4406-8a39-73b685497c87'
Traceback (most recent call last):
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
[...]

PR #6303, opened 2022-08-08
Add `DockerRegistry` block for `DockerContainer` to pull images from registries with authentication

madkinsz:
- Adds a `DockerRegistry` block with a `login` capability.
- Adds an `image_registry` setting to the `DockerContainer` infrastructure block allowing login to a registry before image pull.

Addresses https://github.com/PrefectHQ/prefect/issues/6230 — I do not think it is fair to close this since a separate implementation with ECR token refresh support will be required.

### DIFF
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -54,7 +54,7 @@ class DockerRegistry(Block):
             omitted.
         reauth: If already logged into the registry, should login be performed again?
             This setting defaults to `True` to support common token authentication
-            patterns such as ECS.
+            patterns such as ECR.
     """
 
     _block_type_name = "Docker Registry"
@@ -72,7 +72,9 @@ class DockerRegistry(Block):
             username=self.username,
             password=self.password.get_secret_value(),
             registry=self.registry_url,
-            reauth=True,
+            # See https://github.com/docker/docker-py/issues/2256 for information on
+            # the default value for reauth.
+            reauth=self.reauth,
         )
 
     def _get_client(self):
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -58,7 +58,7 @@ class DockerRegistry(Block):
     """
 
     _block_type_name = "Docker Registry"
-    _block_schema_capabilities = ["login"]
+    _block_schema_capabilities = ["docker-login"]
 
     username: str
     password: SecretStr
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -56,7 +56,8 @@ class DockerRegistry(Block):
             This setting defaults to `True` to support common token authentication
             patterns such as ECR.
     """
-_logo_url = "https://images.ctfassets.net/gm98wzqotmnx/2IfXXfMq66mrzJBDFFCHTp/344dda583986d2d0db361c92dd650693/Moby-logo.webp?h=250"
+
+    _logo_url = "https://images.ctfassets.net/gm98wzqotmnx/2IfXXfMq66mrzJBDFFCHTp/344dda583986d2d0db361c92dd650693/Moby-logo.webp?h=250"
     _block_type_name = "Docker Registry"
     _block_schema_capabilities = ["docker-login"]
 
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -12,6 +12,7 @@ from slugify import slugify
 from typing_extensions import Literal
 
 import prefect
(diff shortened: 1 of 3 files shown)


chrisguidry (approved):
I'm still surprised we have to do this at all, but reading through the related `docker-py` issue, it makes sense.  It also makes a lot of sense to be a block.

What would the username and password be for ECR?  The result of `aws ecr get-login-password ...`?

madkinsz:
I'm not 100% sure on ECR, you can see some examples in the upstream issue https://github.com/docker/docker-py/issues/2256#issuecomment-887175827

anna-geller:
This is awesome, thanks so much Michael! 

@chrisguidry yes correct, you would use this command to get the token which can be used as password, the username is always AWS:`docker login -u AWS -p your_token`

In theory, AWS considers this as insecure and recommends using --password-stdin:
```
aws ecr get-login-password --region region | docker login --username AWS --password-stdin aws_account_id.dkr.ecr.region.amazonaws.com
```

but either way, the token is only valid for 12 hours, so not sure whether after this initial login, dockerpy can consider just using permissions attached via IAM role to the EC2 instance...

madkinsz:
We can add an `ElasticContainerRegistry` block that pulls the ECR token at runtime to the `prefect-aws` collection. cc @desertaxle

src/prefect/infrastructure/docker.py
```diff
@@ -39,6 +40,63 @@ class ImagePullPolicy(AutoEnum):
...
+        password: The password to log into the registry with.
+        registry_url: The URL to the registry. Generally, "http" or "https" can be
+            omitted.
+        reauth: If already logged into the registry, should login be performed again?
+            This setting defaults to `True` to support common token authentication
+            patterns such as ECR.
+    """
+
+    _block_type_name = "Docker Registry"
+    _block_schema_capabilities = ["login"]
```
desertaxle, line 61: I think the capability should be a little more specific
```suggestion
    _block_schema_capabilities = ["docker-login"]
```

src/prefect/infrastructure/docker.py
```diff
@@ -39,6 +40,63 @@ class ImagePullPolicy(AutoEnum):
...
+    Attributes:
+        username: The username to log into the registry with.
+        password: The password to log into the registry with.
+        registry_url: The URL to the registry. Generally, "http" or "https" can be
+            omitted.
+        reauth: If already logged into the registry, should login be performed again?
+            This setting defaults to `True` to support common token authentication
+            patterns such as ECR.
+    """
+
```
desertaxle, line 59: Here's a fun little Moby logo for this block
```suggestion
_logo_url = "https://images.ctfassets.net/gm98wzqotmnx/2IfXXfMq66mrzJBDFFCHTp/344dda583986d2d0db361c92dd650693/Moby-logo.webp?h=250"
```

src/prefect/infrastructure/docker.py
madkinsz: Agreed.

src/prefect/infrastructure/docker.py
```diff
@@ -39,6 +40,63 @@ class ImagePullPolicy(AutoEnum):
...
+_logo_url = "https://images.ctfassets.net/gm98wzqotmnx/2IfXXfMq66mrzJBDFFCHTp/344dda583986d2d0db361c92dd650693/Moby-logo.webp?h=250"
+    _block_type_name = "Docker Registry"
+    _block_schema_capabilities = ["docker-login"]
+
+    username: str
+    password: SecretStr
+    registry_url: str
+    reauth: bool = True
+
+    def login(self):
```
madkinsz, line 68: Might want this to be async first?

src/prefect/infrastructure/docker.py
anna-geller: @madkinsz the current logo doesn't render well, we can take the one from Alex for both this and DockerContainer Block 
image

anna-geller:
This works 🎉 

image

image
 
QAed with S3 storage block and Dockerfile that does nothing but adding s3fs dependency:

```
FROM prefecthq/prefect:dev-python3.9
RUN pip install s3fs
```

anna-geller (approved):

madkinsz:
@anna-geller This was waiting for a few things still :) I'll open a follow-up.

feliperazeek:
@anna-geller I am going over your docker-based [sample project](https://github.com/anna-geller/prefect-docker-deployment/), how would I modify it to use ECR? thank you very much!

zanieb:
@feliperazeek I'd recommend opening a new issue instead of bumping this old pull request.

PR #6279, opened 2022-08-06
Deployment updates

cicdw:
### Summary

This PR is the first in a series of changes that update deployments in a few different ways based on feedback; the current PR:
- removes the `manifest.json` files
- updates deployment schemas to include 3 new fields, and expands `build` CLI to support these fields
- stops saving local file system blocks automatically
- exposes a way to provide targeted infrastructure overrides via dot-delimited paths

And follow-up PRs will expand upon this with:
- Python-based deployment creation
- better Docker support
- and last but not least improved documentation around recipes and best practices

### Steps Taken to QA Changes

- updated tests, wrote new tests
- tested using some QA projects
- @anna-geller is supporting

PR #6317, opened 2022-08-09
Update work queues to be based on name, not tags

jlowin:
### Summary
Depends on #6279 

Implements the changes described in the [raft design doc](https://www.notion.so/prefect/Improving-Run-Work-Queue-Affinity-5a61f9ef96e6495d8605a43327591311), namely:
- work queues operate strictly by name, not by matching tags. Deployments (and flow runs) are explicitly linked to a single work queue
- as a convenience, work queues are automatically created whenever deployments or agents reference them. This means users do not need to interact with work queues explicitly unless they want to use work-queue specific features like pausing and concurrency
- the "correct" way to start an agent is now `prefect agent start <work queue name>`
- backwards-compatibility is maintained. If `work_queue.filter is not None`, it is considered a "legacy" work queue and uses tag-based matching. Conversely, if `work_queue.filter is None` then it is a "new" work queue and tag matching is not used.
- deployment.yaml uses a default value of `work_queue_name="global"`

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**


anna-geller:
I did a short QA - so far, it works for both filtered and global queue

image

we would probably need to add a deprecation warning for the --tag and add a --work-queue-name (-wqn) flag instead? or take the tag provided on "build" as a work-queue name
image


I was confused that setting tag "dev" wasn't enough for it to be picked up by the dev queue, instead it got a global queue unless YAML would be modified and changed to "dev"
image

---
Backwards compatibility works partially atm, in the sense that even though I can start an agent with a work-queue tag "dev" (old work queue):
image

The run will still be directed by default to the global work queue rather than the "dev" queue

jlowin:
@anna-geller thanks for the feedback. I've added deprecation warnings to the CLI (and exposed work queues in the build command). 

I've also ensured that if you provide tags, the work queue that is created uses the deprecated behavior (which I think was the situation you hit, in which the newly-created work queue only matched on name)

src/prefect/client.py
```diff
@@ -1270,10 +1283,14 @@ async def create_deployment(
         schedule: schemas.schedules.SCHEDULE_TYPES = None,
         parameters: Dict[str, Any] = None,
         description: str = None,
+        work_queue_name: str = None,
```
anna-geller, line 1286:
```suggestion
        work_queue_name: str ="global",
```
should this be "global" by default?

src/prefect/deployments.py
```diff
@@ -125,17 +137,36 @@ def immutable_fields_dict(self):
     tags: List[str] = Field(default_factory=list)
     schedule: schemas.schedules.SCHEDULE_TYPES = None
     flow_name: str = Field(..., description="The name of the flow.")
+    work_queue_name: Optional[str] = Field(
+        None,
```
anna-geller, line 141:
```suggestion
       "global",
```

anna-geller:
I QAed just now 

It works well with the legacy tags (so we definitely don't break the legacy tags) 

The deprecation hints when using the old method with tags are really nice:
image
image


But the new way doesn't work atm with CLI (it only works when manually setting the `work_queue_name` in YAML)

If I don't explicitly provide the `--work-queue global`, it ends up with the value `null` instead of `global` in the deployment YAML file (`work_queue_name: null`)

And if I do set it explicitly, it still ends up being null atm:
image

it could be that syncing with main can solve it, but I didn't want to break anything

So the agent/queue functionality works and we don't break anyone, but the `build` CLI doesn't work yet

jlowin:
@anna-geller after speaking with @cicdw, decided to remove the default because it moves the failure mode of "why isn't my work running" to users not having started agents against the right work queue, which is a lot harder for us to diagnose. Having a `null` default with a comment (see https://github.com/PrefectHQ/prefect/pull/6339) means that we can actually detect this failure mode in many places, most importantly including the UI ("Hey, this deployment has no work queue!").

I do like the idea of a default if it removes frictions but I'm still hesitant that a queue called "global" is a useful place for every deployment to start out pointing at (it might even reintroduce the problem this PR is meant to solve, where users unintentionally overlap work and agents). If we discover a useful default value - or take one from infrastructure - then I'm all for adding it, since it would be non-breaking, but adding it now and finding it's not helpful is difficult to undo.

anna-geller:
> "Hey, this deployment has no work queue!"

Nice!

> the idea of a default

Totally understandable! I didn't mean to introduce it as a default, I only thought this was your intention but you just forgot to add it :) this was because the CLI still tells us "global" is the default - let's fix the CLI info and we're good then 👍 

image

src/prefect/cli/deployment.py
```diff
@@ -364,11 +365,14 @@ async def build(
...
+    work_queue_name: str = typer.Option(
         None,
-        "-t",
-        "--tag",
-        help="One or more optional tags to apply to the deployment.",
+        "-q",
+        "--work-queue",
+        help=(
+            "The work queue that will handle this deployment's runs. "
+            "It will be created if it doesn't already exist. Defaults to 'global'."
```
anna-geller, line 374:
```suggestion
            "It will be created if it doesn't already exist. Defaults to 'null'."
```

anna-geller (approved):
QAed with various deployment configurations incl. local code + local process, local process with S3+Azure+GCS and with dev Docker container and everything works well now 👍

jlowin:
@anna-geller sorry for all your hard work but I'm going to split this PR into #6342 and #6343 to try to minimize the number of changes!

PR #6319, opened 2022-08-09
Python deployment builds

cicdw:
### Summary
This PR refactors our deployment CLI logic into the `prefect.deployments.Deployment` class with the ultimate goal of exposing this class to users; we received significant feedback that defining deployments and building them from within Python was desirable, and this PR addresses that feedback.

In particular, all CLI logic is moved into the base Python class; the only thing remaining to do here is:
- possibly write some additional tests now that the logic is isolated
- ensure there is a well-established public interface that mirrors `build` / `apply`
- QA and polish

**NOTE** that this PR is currently pointed to #6279 for ease of reading and will be updated to `main` at the appropriate time

### Steps Taken to QA Changes

Have been using this locally against the Cloud API while developing.


anna-geller:
@cicdw I destroyed my DB and checked 3 times to be sure: it doesn't work with the CLI commands - any "build" CLI command throws the same `ObjectNotFound` error:

```shell
(dev) ➜  prefect deployment build flows/healthcheck.py:healthcheck --name process --tag dev -o deployments/process.yaml -ib process/dev -v my_github_sha

Found flow 'healthcheck'
  File "/Users/anna/repos/prefect/src/prefect/client.py", line 1379, in read_deployment_by_name
    response = await self._client.get(f"/deployments/name/{name}")
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/httpx/_client.py", line 1751, in get
    return await self.request(
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/httpx/_client.py", line 1527, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
  File "/Users/anna/repos/prefect/src/prefect/client.py", line 280, in send
    response.raise_for_status()
  File "/Users/anna/repos/prefect/src/prefect/client.py", line 226, in raise_for_status
    raise PrefectHTTPStatusError.from_httpx_error(exc) from exc.__cause__
prefect.exceptions.PrefectHTTPStatusError: Client error '404 Not Found' for url 'http://127.0.0.1:4200/api/deployments/name/healthcheck/process'
Response: {'detail': 'Deployment not found'}
For more information check: https://httpstatuses.com/404

The above exception was the direct cause of the following exception:

  File "/Users/anna/repos/prefect/src/prefect/deployments.py", line 195, in load
    deployment = await client.read_deployment_by_name(
  File "/Users/anna/repos/prefect/src/prefect/client.py", line 1382, in read_deployment_by_name
    raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
prefect.exceptions.ObjectNotFound

  File "/Users/anna/repos/prefect/src/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/Users/anna/repos/prefect/src/prefect/utilities/asyncutils.py", line 193, in wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/Users/anna/repos/prefect/src/prefect/utilities/asyncutils.py", line 140, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/opt/miniconda3/envs/dev/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/opt/miniconda3/envs/dev/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/opt/miniconda3/envs/dev/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/anna/repos/prefect/src/prefect/cli/deployment.py", line 423, in build
    await deployment.load()  # load server-side settings, if any
  File "/Users/anna/repos/prefect/src/prefect/deployments.py", line 217, in load
    except ObjectNotFound:
NameError: name 'ObjectNotFound' is not defined
An exception occurred.
```

Btw, shouldn't it be `flow_function_name` instead of `flow_name`?

src/prefect/deployments.py
```diff
@@ -172,3 +240,195 @@ def cast_storage_to_block_type(cls, value):
...
+            yaml.dump(self._editable_fields_dict(), f, sort_keys=False)
+            do_not_edit_msg = " DO NOT EDIT BELOW THIS LINE "
+            msg_length = len(do_not_edit_msg)
+            f.write(
+                f"###{' ' * msg_length}###\n###{do_not_edit_msg}###\n###{' ' * msg_length}###\n"
+            )
+            yaml.dump(self._immutable_fields_dict(), f, sort_keys=False)
+
+    @sync_compatible
+    async def apply(self) -> UUID:
```
zangell44, line 355: general developer experience question - how careful should we be when using the `sync_compatible` decorator for common user interfaces?

IIRC the snag is anything decorated with `sync_compatible` will need to be awaited ONLY IF it is called in an async context

This might cause some weirdness for users. I think this might have already caused a bug during internal development as well?

zangell44 (approved):
LGTM

src/prefect/deployments.py
```diff
@@ -2,59 +2,69 @@
...
 from prefect.utilities.importtools import import_object
 
 
 @inject_client
 async def load_flow_from_flow_run(
-    flow_run: schemas.core.FlowRun, client: OrionClient
+    flow_run: schemas.core.FlowRun, client: OrionClient, ignore_storage: bool = False
 ) -> Flow:
     """
     Load a flow from the location/script provided in a deployment's storage document.
```
bunchesofdonald, line 36: It'd be nice to have a docstring for `ignore_storage`.

src/prefect/deployments.py
```diff
@@ -172,3 +240,195 @@ def cast_storage_to_block_type(cls, value):
...
+                            )
+                        )
+            except ObjectNotFound:
+                return False
+        return True
+
+    @sync_compatible
+    async def update(self, ignore_none: bool = False, **kwargs):
+        """
+        Performs an in-place update with the provided settings.
```
bunchesofdonald, line 297: A docstring for `ignore_none` would be helpful too.

src/prefect/deployments.py
```diff
@@ -172,3 +240,195 @@ def cast_storage_to_block_type(cls, value):
...
+                # if not building off a block, will create an anonymous block
+                self.infrastructure = self.infrastructure.copy()
+                infrastructure_document_id = await self.infrastructure._save(
+                    is_anonymous=True,
+                )
+
+            # we assume storage was already saved
+            storage_document_id = getattr(self.storage, "_block_document_id", None)
+
+            deployment_id = await client.create_deployment(
```
bunchesofdonald, line 376: If I create a Deployment and then apply it in a script, what will happen when I run the script more than once?

bunchesofdonald (approved):

src/prefect/deployments.py
cicdw: Oh boy I have a lot of thoughts on this -- the way I thought about it in this case was that this class and its methods will generally be called within a standalone python script, and most likely in a `if __name__ == "__main__":` block.  Starting up your own event loop is a whole thing in this situation, so most likely users will want to call it synchronously.  

I also had a bug where I called a sync compatible method within an async function and the issue was subtle - I wonder if we could find a way to raise if these aren't awaited or something

src/prefect/deployments.py
cicdw: Two things:
- the project directory structure will be re-uploaded to remote storage (if remote storage is being used)
- the `create_deployment` REST route performs an upsert, so IDs, etc. remain stable

bunchesofdonald: Awesome, that's what I was hoping :)

cicdw:
@bunchesofdonald merge conflicts resolved and docstrings added 👍

bunchesofdonald (approved):
Thanks for adding those docstrings!

PR #6321, opened 2022-08-09
Add test for list of secrets

jlowin:
### Summary

Adds a unit test for blocks with a field typed as `List[SecretStr]`

### DIFF
--- a/tests/orion/models/test_block_documents.py
+++ b/tests/orion/models/test_block_documents.py
@@ -1,4 +1,5 @@
 import string
+from typing import List
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


chrisguidry (commented):
Your tests are always so nice, thank you!

chrisguidry (approved):

jlowin:
Note: it was pointed out in Slack that the UI won't render this properly because it looks for secret fields, not nested fields... we may actually want to discourage this practice

cicdw:
@jlowin that feels like something we should solve in the UI

PR #6340, opened 2022-08-09
Load flows for deployments in a worker thread

madkinsz:
### Summary

If not run in a worker thread, the user's code is executed in an "async" context which means that any methods decorated by `sync_compatible` will return a coroutine. However, top-level code in a script is always intended to be run in a "sync" context, so the user can never `await` the coroutines and code that works without a deployment will break while using a deployment.

### Steps Taken to QA Changes

_Needs QA and unit test_

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -18,6 +18,7 @@ from prefect.flows import Flow
 from prefect.infrastructure import DockerContainer, KubernetesJob, Process
 from prefect.logging.loggers import flow_run_logger
 from prefect.orion import schemas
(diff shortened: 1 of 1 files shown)


cicdw (approved):
interesting - very good detective work

madkinsz:
Thanks! Yeah we need to be careful about performing synchronous IO from asynchronous contexts in general, but it's even more important when it could execute code.

It looks like there are not any isolated tests for `load_flow_from_flow_run` yet. Are those in progress somewhere?

PR #6341, opened 2022-08-09
Pluralize CLI commands

anticorrelator:
### Summary
In this PR we further extend our `Typer` wrappers to add the ability to add pluralized versions of our subcommands to our CLI. This implementation creates hidden (e.g. our documentation and help menus will only list the "singular" subcommands) alternate commands that we can specify when adding a subcomand to a `Typer` object via the `add_typer` method.

@cicdw suggested that we might want to consider a simple default pluralization scheme for each subcommand. This would reduce the maintenance overhead. One benefit to explicitly specifying pluralizations: in this PR I've purposefully omitted pluralizations for specific subcommands that make no sense, though I think that decision should ultimately be a product one.

Another interesting point is we could use the ability to explicitly specify alternate spellings for subcommands to handle various common typos if we wish (cc @chrisguidry).

Pluralized subcommands:
- `block`
- `concurrency-limit`
- `deployment`
- `flow`
- `flow-run`
- `profile`
- `work-queue`

Subcommands I've left unpluralized:
- `agent`
- `cloud`
- `config`
- `kubernetes`
- `orion`

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

### DIFF
--- a/src/prefect/cli/_types.py
+++ b/src/prefect/cli/_types.py
@@ -1,6 +1,8 @@
 """
 Custom Prefect CLI types
 """
+from typing import List
+
 import typer
 import typer.core
 
@@ -47,21 +49,22 @@ class PrefectTyper(typer.Typer):
         typer_instance: "PrefectTyper",
         *args,
         no_args_is_help: bool = True,
-        pluralization_string: str = None,
+        aliases: List[str] = None,
         **kwargs,
     ) -> None:
         """
         This will cause help to be default command for all sub apps unless specifically stated otherwise, opposite of before.
         """
-        if pluralization_string is not None:
-            super().add_typer(
-                typer_instance,
-                *args,
-                name=pluralization_string,
-                no_args_is_help=no_args_is_help,
-                hidden=True,
-                **kwargs,
-            )
+        if aliases:
+            for alias in aliases:
+                super().add_typer(
+                    typer_instance,
+                    *args,
+                    name=alias,
+                    no_args_is_help=no_args_is_help,
+                    hidden=True,
+                    **kwargs,
+                )
 
         return super().add_typer(
             typer_instance, *args, no_args_is_help=no_args_is_help, **kwargs
--- a/src/prefect/cli/_types.py
+++ b/src/prefect/cli/_types.py
@@ -47,11 +47,22 @@ class PrefectTyper(typer.Typer):
         typer_instance: "PrefectTyper",
         *args,
         no_args_is_help: bool = True,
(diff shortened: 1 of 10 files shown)


src/prefect/cli/_types.py
```diff
@@ -47,11 +47,22 @@ def add_typer(
...
         """
         This will cause help to be default command for all sub apps unless specifically stated otherwise, opposite of before.
         """
+        if pluralization_string is not None:
+            super().add_typer(
+                typer_instance,
+                *args,
+                name=pluralization_string,
+                no_args_is_help=no_args_is_help,
+                hidden=True,
```
chrisguidry, line 62: That's cool, I didn't know you could hide commands like that.

src/prefect/cli/concurrency_limit.py
```diff
@@ -25,7 +25,7 @@
     name="concurrency-limit",
     help="Commands for managing task-level concurrency limits.",
 )
-app.add_typer(concurrency_limit_app)
+app.add_typer(concurrency_limit_app, pluralization_string="concurrency_limits")
```
chrisguidry, line 28: Should this be `concurrency-limits`, or does `typer` make that change for us?  In either case, could we get an additional test that shows what happens with `snake_case` names (i.e. they are used as-is, they are converted to `kebab-case`, etc)

chrisguidry (approved):
Pretty neat!  Just that one Q about the `snake_case` command

src/prefect/cli/concurrency_limit.py
anticorrelator: whoops, this is a mistake on my part

src/prefect/cli/_types.py
```diff
@@ -47,11 +47,22 @@ def add_typer(
         typer_instance: "PrefectTyper",
         *args,
         no_args_is_help: bool = True,
+        pluralization_string: str = None,
```
madkinsz, line 50: Can we just implement this as `alias`? It seems nice to have it refer to its purpose rather than the content we are providing.

anticorrelator: Yeah I think that's a good idea, I was also thinking to allow it to be a collection so we can have multiple aliases for typos etc

billpalombi:
Approved from a product perspective. This is a pareto UX improvement- I don't see any downsides. Thank you for doing this!

billpalombi (approved):

madkinsz (approved):

PR #6342, opened 2022-08-09
Backend changes for work queues operating on name, not tags

jlowin:
### Summary

Implements the backend changes described in the [raft design doc](https://www.notion.so/prefect/Improving-Run-Work-Queue-Affinity-5a61f9ef96e6495d8605a43327591311), namely:
- work queues operate strictly by name, not by matching tags. Deployments (and flow runs) are explicitly linked to a single work queue
- as a convenience, work queues are automatically created whenever deployments (or agents) reference them. This means users do not need to interact with work queues explicitly unless they want to use work-queue specific features like pausing and concurrency
- backwards-compatibility is maintained. If `work_queue.filter is not None`, it is considered a "legacy" work queue and uses tag-based matching. Conversely, if `work_queue.filter is None` then it is a "new" work queue and tag matching is not used.
- deployment.yaml uses a default value of `work_queue_name=null`

Note: this work was originally contained in #6317 but I'm trying to split the PR into smaller sections

### Steps Taken to QA Changes

See @anna-geller work in #6317

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/orion/schemas/actions.py
+++ b/src/prefect/orion/schemas/actions.py
@@ -61,6 +61,7 @@ class DeploymentCreate(
             "flow_id",
             "schedule",
             "is_schedule_active",
+            "work_queue_name",
             "description",
             "tags",
             "parameters",
@@ -88,6 +89,7 @@ class DeploymentUpdate(
             "schedule",
             "is_schedule_active",
             "description",
+            "work_queue_name",
             "tags",
             "manifest_path",
             "path",
@@ -352,11 +354,12 @@ class WorkQueueCreate(
     schemas.core.WorkQueue.subclass(
         "WorkQueueCreate",
         include_fields=[
-            "filter",
             "name",
             "description",
             "is_paused",
             "concurrency_limit",
+            # filters are deprecated
+            "filter",
         ],
     )
 ):
@@ -370,11 +373,11 @@ class WorkQueueUpdate(
     schemas.core.WorkQueue.subclass(
         "WorkQueueUpdate",
         include_fields=[
-            "filter",
-            "name",
             "description",
             "is_paused",
             "concurrency_limit",
+            # filters are deprecated
+            "filter",
         ],
     )
 ):
@@ -382,8 +385,6 @@ class WorkQueueUpdate(
 
     class Config:
         extra = "forbid"
-
-    name: Optional[str] = Field(None, description="The name of the work queue.")
 
 
 class FlowRunNotificationPolicyCreate(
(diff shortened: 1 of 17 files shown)


src/prefect/orion/schemas/actions.py
```diff
@@ -370,11 +373,11 @@ class WorkQueueUpdate(
     schemas.core.WorkQueue.subclass(
         "WorkQueueUpdate",
         include_fields=[
-            "filter",
-            "name",
```
zangell44, line 374: is this an acceptable "breaking" change? actions schemas forbid extra fields, so old clients will likely send `name` in the payload and receive 422 responses

zangell44:
One comment on removing `name` from the action schemas.

Would it also make sense to add `work_queue_name` to the `FlowRunUpdate` schema? It would be pretty sweet to just "put" a stalled flow run into work queue in the UI.

tests/orion/api/test_deployments.py
```diff
@@ -101,6 +101,15 @@ async def test_create_deployment(
...
         assert deployment.storage_document_id == storage_document_id
 
+    async def test_default_work_queue_name_is_none(self, session, client, flow):
+
+        data = DeploymentCreate(
+            name="My Deployment", manifest_path="", flow_id=flow.id
+        ).dict(json_compatible=True)
+        response = await client.post("/deployments/", json=data)
+        assert response.status_code == status.HTTP_201_CREATED
+        assert response.json()["work_queue_name"] == None
```
zangell44, line 111: Hmmm I thought the default value for work queue name was `"global"`

jlowin: See discussion [here](https://github.com/PrefectHQ/prefect/pull/6317#issuecomment-1209911886) -- I need to update the description of the PR!

src/prefect/orion/schemas/actions.py
jlowin: Ah, good point.

jlowin:
> Would it also make sense to add work_queue_name to the FlowRunUpdate schema? It would be pretty sweet to just "put" a stalled flow run into work queue in the UI.

I considered something like this when creating a flow run but I was worried that this would require checking for work queue existence every time a run was created (in order to see if we need to create the queue). However I think we don't need to add the guarantee of creating work queues to flow runs -- if you put a flow run in a nonexistent work queue, that's on you (and we can be helpful in showing that in the UI). Only deployments and agents need the auto-create.

In order to keep PR's manageable, I would like to take this up as a separate enhancement.

zangell44 (approved):

PR #6347, opened 2022-08-10
Agents can pull from multiple work queues

jlowin:
### Summary
Depends on #6343 

This PR allows agents to support multiple work queues. Aside from convenience, this introduces a couple of interesting recipes when paired with the changes in #6342 and #6343. For example, users could create a low-priority work queue with a low concurrency limit and a high-priority work queue with a high or unlimited concurrency limit and have one agent pull from both.

This PR does not modify the public interface of `prefect agent start`, though it does print a suggestion that users should prefer `prefect agent start -q hello` over `prefect agent start hello`:

image

This is partially to move away from the ambiguous first argument that can be either a string or an ID and into an explicit world of work queue names. In the (far) future, we intend to make work queues configurable while agents are running, which requires a more mutable structure.

Multiple queues can be passed to the agent:

image


src/prefect/agent.py
```diff
@@ -66,19 +58,24 @@ def __init__(
...
+        work_queues = []
+        for name in self.work_queues:
+            try:
+                # support IDs and names
+                if isinstance(name, UUID):
+                    work_queue = await self.client.read_work_queue(id=name)
+                else:
+                    work_queue = await self.client.read_work_queue_by_name(name)
+            except ObjectNotFound:
+                work_queue = await self.client.create_work_queue(
```
madkinsz, line 74: Might be nice to have a log here

madkinsz: We may also need to handle idempotency here in the eventually. What if I scale my agents horizontally and there's a race condition where multiple agents create the same queue? The slower one would crash. Perhaps that's okay with pod restarts.

src/prefect/agent.py
```diff
@@ -94,31 +91,31 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
-            else:
-                raise
+        # load runs from each work queue
+        for work_queue in await self.get_work_queues():
 
-        # Check for a paused work queue for display purposes
-        if not submittable_runs:
-            work_queue = await self.client.read_work_queue(work_queue_id)
+            # print a nice message if the work queue is paused
             if work_queue.is_paused:
```
madkinsz, line 100: Since we're inspecting the queue here we may want to `yield` each work queue from `get_work_queues()` to ensure that data is fresh.

src/prefect/agent.py
```diff
@@ -94,31 +91,31 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+                        id=work_queue.id, limit=10, scheduled_before=before
+                    )
+                    submittable_runs.extend(queue_runs)
+                except httpx.HTTPStatusError as exc:
+                    if exc.response.status_code == status.HTTP_404_NOT_FOUND:
+                        self.logger.error(
+                            f"Work queue {work_queue.name!r} ({work_queue.id}) not found."
+                        )
+                    else:
+                        self.logger.error(exc)
```
madkinsz, line 117: Do we want a traceback?
```suggestion
                        self.logger.exception(exc)
```

madkinsz: Perhaps not, perhaps the traceback is only necessary for unexpected exception types.

src/prefect/agent.py
```diff
@@ -94,31 +91,31 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
-            )
-        except httpx.HTTPStatusError as exc:
-            if exc.response.status_code == status.HTTP_404_NOT_FOUND:
-                raise ValueError(
-                    f"No work queue found with id '{work_queue_id}'"
-                ) from None
-            else:
-                raise
+        # load runs from each work queue
+        for work_queue in await self.get_work_queues():
```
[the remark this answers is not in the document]
madkinsz: The previous overhead when there were no runs to submit is a bit moot since it has nothing better to do. When there are runs to submit is when we need to be fast.

I'm totally fine with leaving this as is for now. We can add caching in the future if necessary. I just wanted to make sure it was on our minds.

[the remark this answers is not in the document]
jlowin: I think the prefetch behavior should mitigate the need to be fast when there are runs -- we're loading them into the agent with up to 10 seconds of buffer time. I thought you were more concerned with the overhead on the server of the additional calls. In either case, we can examine and optimize in the future but I'm not overly concerned.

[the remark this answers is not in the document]
madkinsz: Fair! 👍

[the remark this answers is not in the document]
jlowin: @madkinsz I added a 30 second cache

src/prefect/agent.py
jlowin: Added a protection against this. Rather than re-entering a read/create try/except pattern, if the create fails we just continue without that work queue. We'll pick it up on the next go-round (and most likely, some other agent is running it anyway)

pleek91:
@jlowin I think this branch needs to be rebased or something. Showing changes that are already in main as new changes

jlowin:
@pleek91 thanks, weird — I think it has something to do with these dependent PRs. Once #6343 is merged I’ll make sure it’s addressed

tpdorsey (approved):
LGTM

src/prefect/agent.py
```diff
@@ -66,19 +57,42 @@ def __init__(
...
+                try:
+                    work_queue = await self.client.create_work_queue(name=name)
+                    self.logger.info(f"Created work queue '{name}'.")
+
+                # if creating it raises an exception, it was probably just
+                # created by some other agent; rather than entering a re-read
+                # loop with new error handling, we log the exception and
+                # continue.
+                except Exception as exc:
+                    self.logger.exception(exc)
```
madkinsz, line 91: We'll already display the full traceback here so let's provide context as the string instead
```suggestion
                    self.logger.exception(f"Failed to create work queue {name!r}.")
```

src/prefect/agent.py
```diff
@@ -66,19 +57,42 @@ def __init__(
...
-            raise ValueError("No work queue name provided.")
-        try:
-            work_queue = await self.client.read_work_queue_by_name(self.work_queue_name)
-            return work_queue.id
-        except ObjectNotFound:
-            return await self.client.create_work_queue(name=self.work_queue_name)
+
+        # if queues were cached less than 30 seconds ago, yield the cached values
+        now = pendulum.now("UTC")
+        if self._cache.get("ts", now.subtract(seconds=31)) > now.subtract(seconds=30):
```
madkinsz, line 68: If only we could walrus. This kind of awkward, can we just handle the case where the cache is empty explicitly? e.g.
```suggestion
        if self._cache and self._cache["ts"] > now.subtract(seconds=30):
```

Somewhat related, I have a minor preference for splitting the cache into two separate variables for clarity

```python
self._work_queue_cache_expiration: datetime = None
self._work_queue_cache: List[WorkQueue] = []
```

jlowin: Thanks, great suggestion

madkinsz (approved):

PR #6348, opened 2022-08-10
Fix tracing bug for unpacking and integer singletons

peytonrunyan:
### Summary

This PR fixes a bug pointed out by @khuyentran1401 where unpacked values were not being correctly traced. This was because we were associated the parent task with the tuple of results that it returned, rather than any individual result. In the process of fixing the unpacking bug, I found out that we were also associating singleton integer values with tasks, which was causing tracing issues as well. 

We now recursively map the return value and its components to the task, to a depth of 1. 
Given the result `[1, ["a","b"], ("c",)]`, the following elements will be mapped to the state:
        - `[1, ["a","b"], ("c",)]`
        - `["a","b"]`
        - `("c",)`

The int `1` will not be mapped to the state because it is a singleton.

### Steps Taken to QA Changes

I ran the breaking flow and several iterations of other flows that involve unpacking to make sure that the tasks trace correctly. I also added extra tests in `test_tasks` to cover unpacking scenarios. 
Here's an image from a recent flow run showing that the trace is in the correct order (task 3 received a trackable result from task 1, and task 2 received an untrackable result from task 1
Screen Shot 2022-08-09 at 8 51 10 PM

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


src/prefect/engine.py
```diff
@@ -1281,14 +1282,26 @@ def link_state_to_result(state: State, result: Any) -> None:
     - The field can be preserved on copy.
     - We cannot set this attribute on Python built-ins.
     """
-    # We cannot track some Python built-ins since they are singletons and could create
-    # confusing relationships, e.g. `None`
-    if type(result) in UNTRACKABLE_TYPES:
-        return
+
+    def link_if_not_untrackable(obj: Any, state: State, ctx: FlowRunContext) -> None:
```
jlowin, line 1286: This function name is confusing because of the double negative - `link_if_trackable` would be preferable. In addition, this function is being redefined on every call of `link_state_to_result` but I don't believe it's making use of any variables in the closure; it could be defined once as a private function in `engine.py`

src/prefect/engine.py
```diff
@@ -1281,14 +1282,26 @@ def link_state_to_result(state: State, result: Any) -> None:
...
+        """Track connection between a task run result and its associated state if it has a unique ID.
+
+        We cannot track booleans, Ellipsis, None, NotImplemented, or the integers from -5 to 256
+        because they are singletons.
+        """
+        if (type(obj) in UNTRACKABLE_TYPES) or (
+            type(obj) == int and obj in UNTRACKABLE_INTS
```
jlowin, line 1293: Why define `UNTRACKABLE_INTS` explicitly rather than do the range check here? 
```suggestion
            isinstance(obj, int) and (-5 <= obj <= 256)
```

jlowin (commented):
Does this issue exist for some but not all strings as well? I believe strings with alphanumerics reuse IDs.

src/prefect/engine.py
jlowin: I have a minor preference for the range check because creating a set of ~250 objects feels a little heavy for checking if the rule is satisfied (I say that only because the rule exists, if it was a bunch of strings we'd have no choice but to create the set). But not a strong opinion, the set check itself is lightweight.

jlowin:
I am not sure... turning to our old friend [stackoverflow](https://stackoverflow.com/questions/67596036/simplest-way-to-defeat-string-interning-in-python) we find a couple inventive suggestions... not sure how to evaluate these though.

peytonrunyan:
Will, Michael, Dustin and I had a chat about it and the consensus was "let's try just accepting all strings for now and keep an ear out for feedback that this is causing confusion". Unless you feel strongly about it, it looks like we're going to try to trace all strings.

src/prefect/utilities/collections.py
```diff
@@ -346,6 +346,74 @@ def visit_nested(expr):
     return result
 
 
+def apply_to_collection(expr, visit_fn: Callable[[Any], Any], max_depth: int = -1):
```
madkinsz, line 349: #6371 should land very soon

peytonrunyan:
@madkinsz is there anything that you think is missing here?

src/prefect/engine.py
```diff
@@ -1267,16 +1267,40 @@ def get_state_for_result(obj: Any) -> Optional[State]:
...
+    """Track connection between a task run result and its associated state if it has a unique ID.
+
+    We cannot track booleans, Ellipsis, None, NotImplemented, or the integers from -5 to 256
+    because they are singletons.
+    """
+    if (type(obj) in UNTRACKABLE_TYPES) or (
+        isinstance(obj, int) and (-5 <= obj <= 256)
+    ):
+        return
+    ctx.task_run_results[id(obj)] = state
```
madkinsz, line 1280: Curious why this is separated out? It makes it so you need to use `partial` and have it consume multiple additional arguments that would otherwise be available if it was written as a closure. Our current standard for defining visit functions is to do so in a closure where it's being used.

tests/test_engine.py
```diff
@@ -1034,3 +1038,167 @@ def my_flow():
...
+    @pytest.mark.parametrize(
+        "test_input", [True, False, -5, 0, 1, 256, ..., None, NotImplemented]
+    )
+    def test_link_state_to_result_with_untrackables(self, monkeypatch, test_input):
+        ctx = self.MockFlowRunContext()
+
+        def get():
+            return ctx
+
+        monkeypatch.setattr("prefect.engine.FlowRunContext.get", get)
```
madkinsz, line 1067: This feels like a hack. I'd try to avoid mocks unless you're making assertions about calls. Let's introduce a `flow_run_context` fixture that builds and enters a fake flow run context for the `flow_run` fixture.

tests/test_tasks.py
```diff
@@ -1453,6 +1448,137 @@ async def test_task_inputs_not_populated_with_singleton_results_upstream(
...
+
+        task_2_run = await orion_client.read_task_run(
+            t2_state.state_details.task_run_id
+        )
+
+        assert task_2_run.task_inputs == dict(
+            task_2_input=[TaskRunResult(id=t1_state.state_details.task_run_id)],
+        )
+
+    async def test_task_inputs_populated_with_result_upstream_from_state_with_unpacking_mixed_tackability(
```
madkinsz, line 1495:
```suggestion
    async def test_task_inputs_populated_with_result_upstream_from_state_with_unpacking_mixed_trackability(
```

madkinsz: or..
```suggestion
    async def test_task_inputs_populated_with_result_upstream_from_state_with_unpacking_mixed_untrackable_types(
```

tests/test_tasks.py
```diff
@@ -1453,6 +1448,137 @@ async def test_task_inputs_not_populated_with_singleton_results_upstream(
...
+
+        task_2_run = await orion_client.read_task_run(
+            t2_state.state_details.task_run_id
+        )
+
+        assert task_2_run.task_inputs == dict(
+            task_2_input=[],
+        )
+
+    async def test_task_inputs_populated_with_result_upstream_from_state_with_unpacking_no_trackables(
```
madkinsz, line 1539:
```suggestion
    async def test_task_inputs_populated_with_result_upstream_from_state_with_unpacking_no_trackable_types(
```

src/prefect/engine.py
peytonrunyan: Jeremiah suggested doing it a few comments above. I'm flexible either way.

madkinsz: He noted it's not making use of variables in the closure so presumably you were using `partial` then as well? It's making use of the flow run context and state variables from the parent and it's simpler to just not use `partial` and keep the function inline. Sorry for the whiplash but I'd go back since there's not a use-case for this function as a standalone.

jlowin: I agree here -- earlier I don't think we were using `visit_collection` at all (so no partial) but now it makes sense to redefine rather than use partial just to mimic the same.

peytonrunyan: Yeah, when I originally made the function I figured I would pass it the parameters explicitly just for the sake of clarity, and in case we ever did need to move it somewhere else. I figured the partial gave me the portability to move it elsewhere while the closure kept the pattern. But if it's never going to be used anywhere else, I can toss it back inside and just use the closure without the parameters.

madkinsz: Yeah I'd avoid optimizing for the case where we might move it somewhere else before we have a need to. Following the pattern we use for `visit_fn` elsewhere should be the priority unless there's additional justification.

src/prefect/engine.py
```diff
@@ -1289,14 +1300,23 @@ def link_state_to_result(state: State, result: Any) -> None:
...
+        We cannot track booleans, Ellipsis, None, NotImplemented, or the integers from -5 to 256
+        because they are singletons.
+        """
+        if (type(obj) in UNTRACKABLE_TYPES) or (
+            isinstance(obj, int) and (-5 <= obj <= 256)
+        ):
+            return
+        flow_run_context.task_run_results[id(obj)] = state
 
     flow_run_context = FlowRunContext.get()
```
madkinsz, line 1316: You should retrieve this above the inline function for clarity.

madkinsz (approved):

PR #6350, opened 2022-08-10
Fixed example for prefect deployment inspect

hugolmn:
### Summary
The example of running `prefect deployment inspect` contained single quotes around the deployment name.
However, these quotes are encoded as %27 in the GET request to the API, which in returns a 404.

>"GET /deployments/name/%27leonardo_dicapriflow/leo-deployment%27 HTTP/1.1" 404 Not Found

Removing the quotes solves this issue. I'm running prefect on windows, could the behavior differ using another OS?

### Checklist

This pull request is:

	- No tests or issue needed

### DIFF
--- a/docs/tutorials/deployments.md
+++ b/docs/tutorials/deployments.md
@@ -265,7 +265,7 @@ Use `prefect deployment inspect` to display details for a specific deployment.
 
 <div class='terminal'>
 ```bash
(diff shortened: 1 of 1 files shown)


tpdorsey:
@anna-geller can you verify this? I don't currently have a Windows environment to test. Seems like this is perhaps a bug, not a docs issue.

anna-geller (approved):

PR #6352, opened 2022-08-10
Reduce opportunity for race condition in flow run idempotency check

jlowin:
### Summary

Flow run idempotency is checked with an inequality comparison of the `created` time of the run to the time at the start of the API call. This works in almost all practical cases, but is subject to two failure modes:
- a race condition in which two runs are inserted concurrently. Both calls measure the time at their start and attempt to insert the run. If call A measures its start time, then call B measures its start time, then call A's run is created, then call B's run is created, then both call A and call B will conclude that **it** was the call that actually inserted the run.
- an implicit dependency on SQLAlchemy's python default for the `created` column, which uses `pendulum.now()` to populate the time _when the ORM model is instantiated_. If that behavior were ever changed or removed for any reason, the database would supply the time as a SQL default on the `created` column. In Postgres, that default uses `CURRENT_TIMESTAMP` which is the time _at the beginning of the transaction_. This means that if the transaction were opened prior to creating the flow run, it would appear as if the flow run predated the API call and the logic would conclude it was created idempotently. Since the idempotent creation route uses SQLalchemy Core, not ORM, it's actually unclear to me if this is relevant or not, but since tests for creating flow runs with states and idempotency keys pass on Postgres I believe somehow it is.

This PR modifies the idempotency logic to explicitly set the `created` value in the API call and checking for equality to determine whether idempotency was enforced. This mitigates the failure modes in the following ways:
- the opportunity for race condition is reduced to two calls generating the created timestamp in the exact same microsecond because the check is now an equality check not a range check
- the implicit reliance on SQLAlchemy's default is removed (if it exists at all). In effect, this PR is implementing the same behavior as SQLAlchemy ORM, but doing so explicitly rather than implicitly by generating the created timestamp in Python, not the database.

### DIFF
--- a/src/prefect/orion/models/flow_runs.py
+++ b/src/prefect/orion/models/flow_runs.py
@@ -48,38 +48,40 @@ async def create_flow_run(
     Returns:
         db.FlowRun: the newly-created flow run
     """
-
     now = pendulum.now("UTC")
-    # if there's no idempotency key, just create the run
+
+    flow_run_dict = dict(
+        **flow_run.dict(
+            shallow=True,
+            exclude={
+                "created",
+                "state",
+                "estimated_run_time",
+                "estimated_start_time_delta",
+            },
+            exclude_unset=True,
+        ),
+        created=now,
+    )
+
+    # if no idempotency key was provided, create the run directly
     if not flow_run.idempotency_key:
-        model = db.FlowRun(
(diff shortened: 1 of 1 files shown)


zangell44 (approved):

madkinsz (approved):
Nice!

Issue #6244, opened 2022-08-04
Allow developers to use feature flags for Prefect development


abrookins:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the proposed behavior

A developer writing code for a new feature should be able to integrate their in-progress code behind a feature flag.

### Describe the current behavior

No feature flagging system exists. To integrate in-progress work, a developer could add a Prefect setting that enabled or disabled the work based on an environment variable.

### Example Use

See https://github.com/PrefectHQ/prefect/pull/6218

### Additional context

This PR includes the core utilities for flagging and a docs file. Future work will include:

- CLI commands
- A flags REST API that the UI consumes to read flag state

bunchesofdonald:
@abrookins should we close this one out?

madkinsz:
Sounds good to me — see #7819.

abrookins:
Closed by #7819.

PR #6218, opened 2022-08-03
Add flagging utilities

abrookins:
### Summary
Add a feature flagging system so that developers can integrate in-progress work behind a flag.

This PR adds the core flagging utilities, tests, and docs. Future PRs will include:
- A CLI command
- A REST API to get flags (to be used by the UI)

Parent issue: #6244 

### Steps Taken to QA Changes
1. Set `PREFECT_FEATURE_FLAGGING_ENABLED=true` in my shell
2. Ran `ipython`
3. Created a feature flag
4. Checked if the feature flag existed
5. Enabled the flag
6. Checked if the flag was disabled

### Checklist

This pull request is:

- [ ] A documentation / typographical error fix
	- No tests or issue needed
- [ ] A short code fix
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a bug report issue 
	- Please include tests. One-line fixes without tests will not be accepted unless it's related to the documentation only.
- [x] A new feature implementation
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a feature enhancement issue 
	- Please include tests
    - Please make sure that your QA steps are both thorough and easy to reproduce by somebody with limited knowledge of the feature that you are submitting

**Happy engineering!**

PR #6369, opened 2022-08-10
Add a feature flags CLI

abrookins:
### Summary

Building off #6218, add a feature flags CLI that allows users to list configured feature flags and whether the flags are enabled or disabled.

When a future PR adds a REST API for listing, enabling, and disabling flags, I'll expand the CLI to use it instead.

Dependent on: #6218 
Parent issue: #6244 

**NOTE**: This will be a lot easier to read when #6218 is merged, so you might want to come back later!

### Steps Taken to QA Changes

1. I ran `prefect flags ls` with `PREFECT_FEATURE_FLAGGING_ENABLED` set to true
2. I ran `prefect flags ls` with `PREFECT_FEATURE_FLAGGING_ENABLED` set to false
3. I created a disabled feature flag in the `src/prefect/utilities/feature_flags.py` file and performed steps 1 and 2
4. I created an enabled feature flag in the `src/prefect/utilities/feature_flags.py` file and performed steps 1 and 2

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**


docs/contributing/feature_flags.md
```diff
@@ -0,0 +1,140 @@
...
+1. Testing a new feature before it's complete. With feature flags,
+   a developer can test, merge, and deploy an incomplete feature.
+   This lets us move faster and integrate in-progress systems early.
+
+2. Releasing a feature to a subset of users, AKA "canary releases."
+   On this technique, we release a new (complete) feature to a small
+   group of users, monitor the results, and gradually release the
+   feature to more users until 100% of users see the new feature.
+
+A crucial fact about both of these uses of feature flags is that soon
```
chrisguidry, line 37: 💯

src/prefect/settings.py
```diff
@@ -397,6 +397,13 @@ def warn_on_database_password_value_without_usage(values):
     prefetched. Defaults to `10`.""",
 )
 
+PREFECT_FEATURE_FLAGGING_ENABLED = Setting(
```
chrisguidry, line 400: You feature flagged the feature flags.

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
+"""
+This module contains code related to feature flagging.
```
chrisguidry, line 2: > contains code

What else does a module contain?  :P 

How about something like:

```suggestion
Support for feature flags enabling experimental behavior in Prefect.
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+        if not batch:
+            break
+
+        flags.extend(batch)
+        offset += batch_size
+
+    return flags
+
+
+# Feature flags currently in use go here.
```
chrisguidry, line 160: It would be nice if the definitions of flags could be in a separate module from the support functions.  I don't have a great idea for how to organize them, though...

We want the flags to be registered automatically at any point we would want to check them, and having them at the bottom of this module makes that true (you have to import this module to get the functions to check them).  Is there another way we could organize things and preserve that?

tests/cli/test_flags.py
```diff
@@ -0,0 +1,69 @@
...
+            expected_code=0,
+        )
+
+
+def test_lists_enabled_feature_flags(enabled_flag):
+    with temporary_settings(updates={PREFECT_FEATURE_FLAGGING_ENABLED: "true"}):
+        invoke_and_assert(
+            command=["flags", "ls"],
+            expected_output=f"""
+            Feature Flags:
```
chrisguidry, line 45: YES

zangell44:
Are we worried about edge cases with in-memory stored feature flags here?
For example, let's say I'm working on a feature in the API and using the ephemeral API. How would I turn the flag "on" in a way that guarantees it's propagated to subprocesses when I run a flow?

Actually, is there a meaningful way to turn a flag "on" if it only exists in memory? Only the REST API is allowed to keep the state? 🤔

Perhaps it's worth exploring profiles + environment variables for this. E.g. `export PREFECT_FEATURE_FLAGGING_<FEATURE-NAME>=true`. Profiles would be the natural places to carry around feature flag context.

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+    conditions: Optional[Iterable[Condition]] = None,
+    client: FeatureFlagClient = None,
+) -> Optional[FeatureFlag]:
+    """
+    Create a feature flag if a flag matching the given name does not
+    already exist.
+
+    Args:
+        flag_name: The name of the feature flag.
+        is_enabled: the initial enabled/disabled state of the flag if
```
tpdorsey, line 46:
```suggestion
        is_enabled: The initial enabled/disabled state of the flag if
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+                    this function creates it
```
tpdorsey, line 47:
```suggestion
                    this function creates it.
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+        client_data: arbitrary data that we should store with the flag
```
tpdorsey, line 48:
```suggestion
        client_data: Arbitrary data that we should store with the flag.
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+        bucketer: an optional "bucketter" from the `flipper` module, e.g.
```
tpdorsey, line 49:
```suggestion
        bucketer: An optional "bucketter" from the `flipper` module, e.g.
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+                  `PercentageBucketer`, to use when determining if the
+                  flag is enabled
+        conditions: an optional iterable of `Condition` instances from the
+                    `flipper` module against which we will check input data
+                    to determine if a flag is enabled
```
tpdorsey, line 54:
```suggestion
                    to determine if a flag is enabled.
```

src/prefect/utilities/feature_flags.py
```diff
@@ -0,0 +1,171 @@
...
+    This function always returns an empty list if the setting
+    `PREFECT_CLOUD_ENABLE_FEATURE_FLAGGING` is false.
+
+    Args:
+        batch_size: batch size of flags to retrieve at a time
+        client: The `FeatureFlagClient` instance to use. Defaults to a client
+                configured to look at an in-memory feature store.
+
+    Returns:
+        List[FeatureFlag]: list of all feature flags in the store
```
tpdorsey, line 137:
```suggestion
        List[FeatureFlag]: List of all feature flags in the store.
```

tpdorsey (commented):
Looks good. Minor suggested doc/docstring edits.

madkinsz:
Could we nest this CLI under `prefect config`? I'd rather not clutter the top-level CLI. I'd also be interested in calling it `feature` rather than `flags` e.g. `prefect config feature ls`, `prefect config feature enable ...`

src/prefect/utilities/feature_flags.py
abrookins: Yeah, I'm thinking about that. I already need to reorganize this a bit because I want to use a separate client instance for testing, which with my "simple functions" approach now means threading that through every call.

Using module side-effects is an antipattern, but it works in Cloud where we can be super minimal. But ultimately, the code in OSS should support changing out the "store" so that our default and simple OSS implementation is easy to replace with the Redis store we use in Cloud -- while all the feature flags and code paths in OSS continue to work and can be toggled/tested in Cloud.

abrookins:
@zangell44 Right, in this version there is no way to enable or disable flags for exactly that reason. I'm going to keep it minimal for now, but I also have a branch going that uses an ORM model to persist feature flags (which we'd only use in OSS).

madkinsz:
Closing as stale and partially superseded by #7819 — we may want to use this approach in the future though!

PR #6371, opened 2022-08-10
Update `visit_collection` to be a synchronous function

madkinsz:
The `visit_collection` function was complicated by the need to support async visitor functions and concurrent gathers. These features are important for resolving futures, but there are other ways to gather the futures. The primary work done by `visit_collection` is synchronous and often we needed to make the visit function async just to satisfy the interface. Sometimes, we need the functionality of `visit_collection` from a synchronous context. Instead of maintaining two copies of a complicated function, this pull request converts `visit_collection` to be synchronous and callers can make use of AnyIO worker threads to run asynchronous visit functions.

This should also unblock repairing recursion issues which were mostly unsolvable due to the concurrent gathers.


peytonrunyan (approved):
LGTM!

bunchesofdonald (approved):
Makes sense to me.

src/prefect/utilities/collections.py
```diff
@@ -289,20 +283,18 @@ def visit_nested(expr):
...
         result = typ(items) if return_data else None
 
     elif typ in (dict, OrderedDict):
         assert isinstance(expr, (dict, OrderedDict))  # typecheck assertion
         keys, values = zip(*expr.items()) if expr else ([], [])
-        keys = await gather(*[visit_nested(k) for k in keys])
-        values = await gather(*[visit_nested(v) for v in values])
+        keys = [visit_nested(k) for k in keys]
+        values = [visit_nested(v) for v in values]
         result = typ(zip(keys, values)) if return_data else None
```
ahuang11, line 294: I wonder if these can be combined into one statement now now that the await/gather is removed?
```
typ((visit_nested(k), visit_nested(v)) for k, v in expr.items())
```

madkinsz: I took a poke at this but we don't want to construct the new type unless `return_data` is `True`, we still need to visit all of the keys and values. I think I've got a simplification but it's still two lines

PR #6380, opened 2022-08-11
Update for breaking pydantic API change

anticorrelator:
A small patch to ensure that we are compatible with the latest release of Pydantic, which includes a breaking APi change. I'm switching the behavior on the Pydantic version here, in case users aren't ready to update their dependencies but I'm not sure what the right choice is.

### DIFF
--- a/src/prefect/orion/utilities/schemas.py
+++ b/src/prefect/orion/utilities/schemas.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, List, Set, Type, TypeVar
 from uuid import UUID, uuid4
 
 import pendulum
+import pydantic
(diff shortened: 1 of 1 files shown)


madkinsz:
Nice fix! I'd be down with excluding v1.9.1 in our requirements if that solves it without branching?

madkinsz:
Original discussion by us at https://github.com/PrefectHQ/orion/pull/1840#discussion_r877528105

Can we just leave this field unset and use the default behavior with 1.9.1 excluded?

madkinsz (approved):

jlowin (approved):

PR #6390, opened 2022-08-11
update docs to recommend using conda 

discdiver:
### Summary

Users have had difficulties upgrading with venv and pipenv in recent days. 
conda also installs the necessary sqlite version in windows environemnts.
This recommendation should reduce user difficulties.

### Checklist

This pull request is:

- [ x] A documentation / typographical error fix
	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -2,11 +2,11 @@
 
 The first step to getting started with Prefect is installing the Prefect Python package. 
 
-## Set up Python
+## Use a Virtual Environment with Python 
 
 Prefect requires Python 3.7 or later.
 
-We recommend installing Prefect 2.0 using a Python virtual environment manager such as `pipenv`, `conda`, or `virtualenv`/`venv`.
+We recommend installing Prefect 2.0 into `conda` virtual environment.
 
 ## Install Prefect 
 


madkinsz (changes requested):
I don't think we should suggest using conda quite like this. If conda is our preferred installation method, we should encourage use of the conda-forge version of Prefect rather than mixing in `pip`. Mixing dependency management tools can lead to a lot of problems.

Instead, I'd prefer to highlight using conda on Windows or specifically to resolve SQLite version issues.

Which venv and pipenv upgrade issues are you referring to? We should resolve those. We should have a rock-solid installation story on popular environment tools.

madkinsz:
Thanks for cleaning this up. @tpdorsey / @anna-geller / @discdiver it'd still be great if ya'll could coordinate on adding a section that covers SQLite version errors and our suggested path to resolving it (basically only use conda right now).

discdiver:
Agree @madkinsz . 

Still strange that conda won't install with conda-forge channel with miniforge on m1 (osx-arm64) - I've tried with Python 3.9 and 3.10. Verbose output not helpful. Prefect from conda-forge installs fine on an intel chip mac with miniconda.

- venv appears to still be creating issues with agents sometimes
- pipenv appears to have an issue but outside our control

Conda environment with pip appears to work on all architectures. I think we should recommend users take that route.

madkinsz:
I've installed Prefect from conda-forge on my M1 without any issues. I haven't tested `miniforge` though, is there an issue for that?

We need to reproduce and fix the venv issue. It's the standard library environment tool.

PR #6401, opened 2022-08-12
Add error message if `get_run_logger` receives context of unknown type

madkinsz:
### Summary

Throw a type error instead of falling through to a "no context available" error if `get_run_logger` is called with an invalid context type. This should help narrow the cause of errors like the following where we can see the function is called with a value.

```python
Traceback (most recent call last):
  File "/root/venv/lib/python3.8/site-packages/prefect_dask/task_runners.py", line 236, in wait
    return await future.result(timeout=timeout)
  File "/root/venv/lib/python3.8/site-packages/distributed/client.py", line 294, in _result
    raise exc.with_traceback(tb)
  File "/root/venv/lib/python3.8/site-packages/prefect/engine.py", line 959, in begin_task_run
    get_run_logger(flow_run_context).debug(
  File "/root/venv/lib/python3.8/site-packages/prefect/logging/loggers.py", line 91, in get_run_logger
    raise RuntimeError("There is no active flow or task run context.")
RuntimeError: There is no active flow or task run context.
```

From https://github.com/PrefectHQ/prefect/issues/6086#issuecomment-1213240222

### Steps Taken to QA Changes

Added unit test.

### Checklist

This pull request is:

- [ ] A documentation / typographical error fix
	- No tests or issue needed
- [x] A short code fix
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a bug report issue 
	- Please include tests. One-line fixes without tests will not be accepted unless it's related to the documentation only.
- [ ] A new feature implementation
	- Please reference the related issue by including "closes `<link to issue>`" in this Pull Request's summary section. 
        - If no issue exists, please create a feature enhancement issue 
	- Please include tests
    - Please make sure that your QA steps are both thorough and easy to reproduce by somebody with limited knowledge of the feature that you are submitting

**Happy engineering!**

PR #6402, opened 2022-08-12
Fix log on abort when the flow run context is not available

madkinsz:
### Summary

Fixes log added in #6307 when the task is run on a remote worker.

Related to bug noted in #6401 

### Steps Taken to QA Changes

This is hard to reproduce, a task run needs to double-execute on a remote worker.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -921,7 +921,7 @@ async def begin_task_run(
     --> `begin_task_run` executes on a different event loop than the flow
     --> Current settings is not set or does not match, settings A is entered
     """
-    flow_run_context = prefect.context.FlowRunContext.get()
+    maybe_flow_run_context = prefect.context.FlowRunContext.get()
 
     async with AsyncExitStack() as stack:
 
@@ -931,13 +931,13 @@ async def begin_task_run(
             stack.enter_context(settings)
             setup_logging()
 
-        if flow_run_context:
+        if maybe_flow_run_context:
             # Accessible if on a worker that is running in the same thread as the flow
-            client = flow_run_context.client
+            client = maybe_flow_run_context.client
             # Only run the task in an interruptible thread if it in the same thread as
             # the flow _and_ the flow run has a timeout attached. If the task is on a
             # worker, the flow run timeout will not be raised in the worker process.
-            interruptible = flow_run_context.timeout_scope is not None
+            interruptible = maybe_flow_run_context.timeout_scope is not None
         else:
             # Otherwise, retrieve a new client
             client = await stack.enter_async_context(get_client())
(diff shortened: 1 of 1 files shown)


src/prefect/engine.py
```diff
@@ -921,7 +921,7 @@ async def begin_task_run(
     --> `begin_task_run` executes on a different event loop than the flow
     --> Current settings is not set or does not match, settings A is entered
     """
-    flow_run_context = prefect.context.FlowRunContext.get()
+    maybe_flow_run_context = prefect.context.FlowRunContext.get()
```
cicdw, line 924: This PR could be a fun one to discuss for a low stakes design / convention discussion; I think the original name is good, as putting `maybe` is kind of like baking in the fact that the type is `Optional[FlowRunContext]` into the variable name.  But I also see the value of including that in the name to make sure attribute access is conditional; could be fun to hear others' thoughts on this (independently of merging this PR).

cicdw (approved):

src/prefect/engine.py
madkinsz: Yeah I'd love to chat about it more in general. I changed the name of the variable here because I, who wrote the original code a while back, introduced a bug because I did not realize the variable was nullable. I figured it'd bite someone else. If we start running type checking in CI it'd catch the bug and I agree it feels weird to replicate the optional type information in the variable name.

jlowin: (Very) weak opinion that I prefer it without `maybe_` because the code using it appears to be exclusively inside a [block](https://github.com/PrefectHQ/prefect/pull/6402/files#diff-ae6f1b6d423c9f28188aedf4ab0f0afbce428d61b84feed78ca9c716ce3e26caR934) that effectively asserts it is not None (`if maybe_flow_run_context:`). Within that block, I think the code is much less readable (since it is not "maybe" a flow run context, but rather "definitely" a flow run context, in any of the places it's being used). In other words I think this trades off increased clarity when you're using this object in the wrong place for reduced clarity when it's being used properly, but that latter case is where most developers will encounter it.

madkinsz: Interestingly the flow run context is basically expected not to be available here. We are only retrieving it to optimize performance when tasks happen to be occurring in the same context as the flow. I like your point though Jeremiah, perhaps there's a better way to emphasize this.

I'm going to merge this to keep things moving along, but it'd be nice to have an engineering review session related to this discussion and we can change it then.

madkinsz: cc @abrookins

PR #6428, opened 2022-08-16
Enhance docs search

jlowin:
Two minor flags following discussion with @tpdorsey :
- search auto-suggestions
- search highlights


tpdorsey (approved):
<highlight>Nice!</highlight>

PR #6432, opened 2022-08-16
Use a fully no op policy when creating a task run

cicdw:
### Summary

This PR introduces a minor optimization when creating task runs; specifically it avoids any concurrency limit logic to reduce the number of database interactions.  

### Steps Taken to QA Changes

Ran test suite - no new functionality was introduced, so assuming all critical behavior is captured via test we will be good to go.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/orion/models/task_runs.py
+++ b/src/prefect/orion/models/task_runs.py
@@ -15,7 +15,7 @@ import prefect.orion.schemas as schemas
 from prefect.orion.database.dependencies import inject_db
 from prefect.orion.database.interface import OrionDBInterface
 from prefect.orion.exceptions import ObjectNotFoundError
-from prefect.orion.orchestration.core_policy import MinimalTaskPolicy
+from prefect.orion.orchestration.core_policy import MinimalTaskPolicy, NoOpTaskPolicy
 from prefect.orion.orchestration.global_policy import GlobalTaskPolicy
 from prefect.orion.orchestration.policies import BaseOrchestrationPolicy
 from prefect.orion.orchestration.rules import (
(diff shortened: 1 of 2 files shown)


src/prefect/orion/models/task_runs.py
```diff
@@ -76,11 +76,13 @@ async def create_task_run(
     model = result.scalar()
 
     if model.created == now and task_run.state:
+        # creating a run should never involve concurrency limits, so we use a NoOp policy
         await models.task_runs.set_task_run_state(
             session=session,
             task_run_id=model.id,
             state=task_run.state,
             force=True,
+            task_policy=NoOpTaskPolicy(),
```
zangell44, line 85: After a second look, I don't think we can do this safely. There are a few rules that seem like they need to run here https://github.com/PrefectHQ/prefect/blob/main/src/prefect/orion/orchestration/global_policy.py#L21

zangell44 (approved):

anticorrelator (approved):

anticorrelator:
This looks good: I should point out that no database interactions to check concurrency limits happen as the code is currently as well: the transitions associated with creating a task run (`None -> Pending`) is invalid against the `ReleaseTaskConcurrencySlot` rule so the context manager will `pass` both on entry and exit

cicdw:
@anticorrelator ah that's awesome to hear thanks for the info!  In that case I would rather us rely on robust rule structure like you describe than to begin incentivizing naive overrides as in my PR, so I'm going to close.

Issue #6434, opened 2022-08-16
Enhancement: UI support for new work queues


znicholasbrown:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the proposed behavior

The UI should support new work queues

### Describe the current behavior

Work queues currently create old-style work queues with filters, tags, and deployments

PR #6435, opened 2022-08-16
Feature: UI work queue updates

znicholasbrown:
### Summary

This PR updates UI work queue, deployment, and flow run types, work queue pages and forms, and deployment pages and forms to match work in #6342

Closes: #6434

cc @stackoverfloweth for persistent toast pattern
cc @zhen0 for updated code snippet

### Steps Taken to QA Changes

- Verify that existing (deprecated) work queues are rendered properly
- Deprecated work queues should be denoted as such with a small toast in the bottom right of the screen
- Deprecated work queues should not be editable but you should be able to turn them on/off and delete them
- New work queues should be editable and should not include tags or deployments
- Non-deprecated deployments without an associated work queue should be denoted as such with a small toast in the bottom right of the screen (old work queues should not have this as they will continue to operate against old work queues)
- Deployment editing should now include a selector for existing work queues

### Checklist

This pull request is:

	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**


orion-ui/src/compositions/useToast.ts
```diff
@@ -0,0 +1,19 @@
+import { showToast, Toast, ToastOptions, ToastType } from '@prefecthq/prefect-design'
+import { onUnmounted, Component } from 'vue'
+
+/**
+ * The useToast composition takes the same arguments as showToast but also dismisses the toast when the caller component is unmounted
```
pleek91, line 5: This is interesting in clever. But I think we could make `dismissOnUnmount` or something similar part of `ToastOptions` right? Rather than using a composition.

orion-ui/src/pages/Deployment.vue
```diff
@@ -76,4 +77,11 @@
   function routeToDeployments(): void {
     router.push(routes.deployments())
   }
+
+  watchEffect(() => {
```
pleek91, line 81: as nice as `watchEffect` is I think we should use a `watch` with `{ immediate: true }` here so we can watch just the deployment. `watchEffect` will watch reactive data even in method calls. Meaning if `showToast` ever has reactive data referenced those will trigger the effect. `watch` is just a lot safer.

orion-ui/src/compositions/useToast.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param message string | Component
+ * @param type ToastType
+ * @param options ToastOptions
+ * @returns Toast
+ */
+export function useToast(message: string | Component, type?: ToastType, options?: ToastOptions): Toast {
```
stackoverfloweth, line 11: I think this composition is slightly wrong. 

since `useToast` has an `onUnmounted` hook it needs to be called from a setup function, not necessarily at the same time the developer wants a toast message to appear. 

I would have expected this composition to return a function, so it would always be "initiated" in the setup and called whenever the developer needs the toast to show up. 

<script lang="ts" setup>
    // no options needed 
    const showToast = useToast()
    
    async function doSomethingAsync(): Promise<void> {
        await something()
        // options go here like regular toast
        showToast(....)
    }
</script>

stackoverfloweth (changes requested):

orion-ui/src/pages/WorkQueue.vue
```diff
@@ -65,6 +66,12 @@
   const routeToQueues = (): void => {
     router.push(routes.workQueues())
   }
+
+  watchEffect(() => {
```
pleek91, line 70: same `watch` suggestion here.

orion-ui/src/compositions/useToast.ts
znicholasbrown: Ultimately we might want to do that but @stackoverfloweth and I explored this a little and it would require a non-trivial refactor of the toast plugin

orion-ui/src/compositions/useToast.ts
pleek91: +1

orion-ui/src/pages/Deployment.vue
znicholasbrown: Yup you're correct, updated 👍🏻

orion-ui/src/compositions/useToast.ts
znicholasbrown: Good call - let me see if I can find a better way to surface that

znicholasbrown: Made an update to that, lmk what you think

orion-ui/src/compositions/useToast.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * The useToast composition returns a callable that can be used in place of showToast to automatically dismiss any toasts created during the component's lifecycle
+ * @returns () => Toast
+ */
+export function useToast(): typeof showToast {
+  const toasts: Toast[] = []
+
+  onUnmounted(() => toasts.forEach(toast => toast.dismiss()))
+
+  return (message, type, options) => {
```
pleek91, line 13: can we type these args

```suggestion
  return (...args: Paramaters<typeof showToast>) => {
```

znicholasbrown: Yeah this is more future-proof than just the return type, I like it (didn't commit the suggestion cause the args needed to be passed as well)

pleek91 (approved):
LGTM

stackoverfloweth (approved):
LGTM

Issue #6335, opened 2022-08-09
Excessive amount of AssertionError: feed_data before feed_eof in the log


drfraser:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Everything seems to be working fine, but when I looked in the orion.log file (SERVER_LEVEL:WARNING) I see the exception as below. Unfortunately I get the impression I can't dynamically change the logging level using SIGHUP/re-read configuration so I can't add any more context and I don't want to stop the Orion server etc.

If I could  get some background info on why this might be happening over and over, I can debug it further.  Perhaps my code is doing something incorrectly?

### Reproduction

```python
Need more info before I can provide code
```

### Error

```
17:35:38.216 | ERROR   | asyncio - Exception in callback SubprocessStreamProtocol.pipe_data_received(1, b'INFO:     1...1.1" 200 OK\n')
handle: <Handle SubprocessStreamProtocol.pipe_data_received(1, b'INFO:     1...1.1" 200 OK\n')>
Traceback (most recent call last):
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.8/asyncio/subprocess.py", line 73, in pipe_data_received
    reader.feed_data(data)
  File "/usr/lib/python3.8/asyncio/streams.py", line 472, in feed_data
    assert not self._eof, 'feed_data after feed_eof'
AssertionError: feed_data after feed_eof
```

### Versions

```
Version:             2.0.3
API version:         0.8.0
Python version:      3.8.10
Git commit:          2f1cf4ac
Built:               Fri, Aug 5, 2022 3:57 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         hosted

```

madkinsz:
This error comes from AnyIO's parsing of subprocess output streams. I've seen it locally a few times and have generally tried to squash it, but I'm not sure what the cause is.

`prefect.utilities.processutils.run_process` is the relevant implementation. If you run that directly in a script, can you reproduce this error?

drfraser:
[...]

PR #6445, opened 2022-08-17
Improve robustness of subprocess text streaming

madkinsz:
### Summary

Previously, this utility would call blocking writes to stderr and stdout directly. This appears to introduce the possibility of race conditions and blocked event loops when many processes are run concurrently. Since we use these utilities to launch flows in parallel processes from the agent, it is important that they are robust to concurrency. `wrap_file` _may_ not be thread safe, but this is still an improvement from where we were. There is not a clear suggested pattern for this, see extensive discussion at https://github.com/python-trio/trio/issues/174.

Attempts to address https://github.com/PrefectHQ/prefect/issues/6335

### Steps Taken to QA Changes

I wrote a recursive subprocess test

```python
from prefect.utilities.processutils import run_process
import anyio
import sys


async def main(count):
    if count <= 0:
        return

    print(f"Running {count} children")

        for _ in range(count):
            await tg.start(
                run_process, [sys.executable, "./process-test.py", str(count - 2)], True
            )

if __name__ == "__main__":
    anyio.run(main, int(sys.argv[1]))
```

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/utilities/processutils.py
+++ b/src/prefect/utilities/processutils.py
@@ -8,7 +8,7 @@ import anyio
 import anyio.abc
 from anyio.streams.text import TextReceiveStream, TextSendStream
 
-TextSink = Union[TextIO, TextSendStream]
+TextSink = Union[anyio.AsyncFile, TextIO, TextSendStream]
 
 
 @asynccontextmanager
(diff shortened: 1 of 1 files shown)


chrisguidry (approved):
Excellent find!

peytonrunyan (approved):
lgtm!

Issue #6446, opened 2022-08-17
Prefect agent crashes when subprocess exits on Windows


nicholasjng:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When running a flow to completion in Prefect on Windows via a Prefect agent tied to a deployment, I encounter an exception related to the decoding of some data chunks in the shutdown process.

### Reproduction

```python
# Code is private right now unfortunately - I will try to come up with an MWE in the following days and amend the issue.
```

### Error

```
Traceback (most recent call last):
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\prefect\cli\_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\prefect\utilities\asyncutils.py", line 193, in wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\prefect\utilities\asyncutils.py", line 140, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\anyio\_core\_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\anyio\_backends\_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "C:\Users\nju\AppData\Local\Programs\Python\Python310\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "C:\Users\nju\AppData\Local\Programs\Python\Python310\lib\asyncio\base_events.py", line 646, in run_until_complete
    return future.result()
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\anyio\_backends\_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\prefect\cli\agent.py", line 93, in start
    async with OrionAgent(
  File "C:\Users\nju\Workspaces\python\venv\lib\site-packages\prefect\agent.py", line 271, in __aexit__
[...]

PR #6456, opened 2022-08-17
Update agent to prevent crash when infrastructure errors after submission

madkinsz:
~Builds on the simple change in #6453 — this can supersede that pull request entirely if desired.~

Requires #6610 as it uses the agent test refactor there.

### Summary

Running infrastructure is a two step process:
- Submission: infrastructure is created for the flow run
- Monitoring: the infrastructure is watched for success/failure

AnyIO's `TaskStatus` is used to report the completion of submission
allowing a caller to use `task_group.start` to await submission without
blocking on monitoring. If an error occurs during submission, the
agent captures it and marks the flow run as failed. However, if an
error occured during monitoring, the agent does not capture this
exception. Instead, it is raised in the agent's task group which
can cause the agent and any other concurrent tasks to fail.

A single flow run should not be able to crash the agent. To resolve
this, the agent is updated to capture exceptions during the
monitoring phase as well. In contrast to submission, we do not mark
the flow run as failed as this could be unrelated to execution of
the flow itself. The error and traceback are logged.

Prevents an exception in #6446 from crashing the agent. Does not address the cause of that exception.

### Future work

- Consider changing the reported state for flow runs that fail to submit to CRASHED.
- Consider using the exit code of the infrastructure to report CRASHED flow runs that fail to report their own failure.
- Consider separate `start` and `watch` interfaces for the `InfrastructureBlock` which would simplify this handling and allow us to resume watching jobs on agent restart.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -8,13 +8,14 @@ from uuid import UUID
 import anyio
 import anyio.to_process
 import pendulum
-from anyio.abc import TaskGroup
+from anyio.abc import TaskGroup, TaskStatus
 
 from prefect.blocks.core import Block
 from prefect.client import OrionClient, get_client
 from prefect.engine import propose_state
 from prefect.exceptions import Abort, ObjectNotFound
-from prefect.infrastructure import Infrastructure, Process
+from prefect.infrastructure import Infrastructure, InfrastructureResult, Process
+from prefect.infrastructure.submission import submit_flow_run
 from prefect.logging import get_logger
 from prefect.orion.schemas.core import BlockDocument, FlowRun, WorkQueue
 from prefect.orion.schemas.data import DataDocument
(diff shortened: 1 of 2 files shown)


src/prefect/agent.py
```diff
@@ -193,19 +193,42 @@ async def submit_run(self, flow_run: FlowRun) -> None:
...
+            )
+            self.logger.info(f"Completed submission of flow run '{flow_run.id}'")
+
+        self.submitting_flow_run_ids.remove(flow_run.id)
+
+    async def _submit_run_and_capture_errors(
+        self,
+        flow_run: FlowRun,
+        infrastructure: Infrastructure,
+        task_status: TaskStatus = None,
```
bunchesofdonald, line 209: It looks like `task_status`, will always be `None`. I'd drop it from the function definition and the call to `submit_flow_run` and then we can add it back when/if needed.

bunchesofdonald (approved):
Just a minor nitpick, but otherwise I think this makes sense

billpalombi (approved):
Agreed with the motivation for this change. We should definitely take up the future work to set the flow run state to CRASHED when runs fail to submit.

src/prefect/agent.py
madkinsz: `task_group.start` passes a `task_status`. It's a special keyword argument that lets the task group block until `.started()` is called.

madkinsz:
If anyone has some spare time to fix the tests here, I've got a few to many things going on to get this across the line in the near future.

madkinsz:
@bunchesofdonald this implementation has changed significantly.

src/prefect/agent.py
```diff
@@ -194,25 +194,67 @@ async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
...
```
bunchesofdonald, line 217: Would it be worth namespacing `Task` related things that come from anyio? It took me a bit to realize this wasn't a Prefect concept, whereas if this was `anyio.abc.TaskStatus` it would have been a lot more obvious.

src/prefect/agent.py
```diff
@@ -194,25 +194,67 @@ async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
...
+    ) -> Union[InfrastructureResult, Exception]:
+
+        # Note: There is not a clear way to determine if task_status.started() has been
+        #       called without peeking at the internal `_future`. Ideally we could just
```
bunchesofdonald, line 221: Maybe throw a PR at `anyio` that exposes the `canceled`, `running`, and `done` methods on the Future or the future itself?

bunchesofdonald (approved):
I have no memory of what this was like before, but I like it now!

src/prefect/agent.py
madkinsz: I think this pattern of passing `task_status` through several functions would be considered depraved by them :D I'm hoping that this is resolved by splitting `run` into two separate methods: `submit` and `watch`.

src/prefect/agent.py
madkinsz: Right now we never use the `anyio.abc` namespace explicitly, but I'm happy to do so in the future.

PR #6472, opened 2022-08-18
Move task tracing to engine and add StateDetails field for untrackable states

peytonrunyan:
### Summary

This pull requests updates our tracing to be called from within the engine in `propose_state` instead of in `State.result()`. It also adds a new field `StateDetails.returns_untrackable_result` which is included as part of the dependency graph response.

### Steps Taken to QA Changes

Ran `prefect-qa` and checked to make sure that dependency is still visualized correctly. I also updated tests to make sure that `returns_untrackable_result` is updated correctly.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


src/prefect/engine.py
```diff
@@ -1420,6 +1425,7 @@ def link_if_trackable(obj: Any) -> None:
         if (type(obj) in UNTRACKABLE_TYPES) or (
             isinstance(obj, int) and (-5 <= obj <= 256)
         ):
+            state.state_details.returns_untrackable_result = True
```
madkinsz, line 1428: We should highlight in this method's docstring that it's mutating the state object. This also suggests that we may be doing things in the incorrect order. Perhaps the state should update this attribute when data is assigned to it. This function can then just check the value of this boolean.

madkinsz: Although trackable types are a client-side concept and we don't have a client-side `State` object yet.. so actually probably not worth going down that road yet.

src/prefect/orion/schemas/states.py
```diff
@@ -46,6 +46,7 @@ class StateDetails(PrefectBaseModel):
     scheduled_time: DateTimeTZ = None
     cache_key: str = None
     cache_expiration: DateTimeTZ = None
+    returns_untrackable_result: bool = False
```
madkinsz, line 49: I have a preference for `untrackable_result`. The state does not "return" the result, it's assigned to it. We happen to use the return value of the user's function most of the time, but we also manually set this data to other things like exceptions.

madkinsz (approved):

Issue #6289, opened 2022-08-08
Deployment build using Samba Share does not accept the keyword overwrite


LukasK13:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When building a deployment and using a samba share as remote file system, an invalid keyword argument is passed to the filesystem library. The bug was introduced with this completely unrelated commit: https://github.com/PrefectHQ/prefect/commit/46b323bc3700bfd82a329b90ee2029c55652e01b

### Reproduction

```python
prefect deployment build ./catfact.py:catfacts_flow -n catfact -t test --storage-block remote-file-system/samba-share
```

### Error

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/prefect/utilities/asyncutils.py", line 193, in wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/prefect/utilities/asyncutils.py", line 140, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/usr/local/lib/python3.8/dist-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/usr/local/lib/python3.8/dist-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.8/dist-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/usr/local/lib/python3.8/dist-packages/prefect/cli/deployment.py", line 469, in build
    file_count = await storage.put_directory(ignore_file=".prefectignore")
  File "/usr/local/lib/python3.8/dist-packages/prefect/filesystems.py", line 288, in put_directory
    self.filesystem.put_file(f, fpath, overwrite=True)
[...]

PR #6475, opened 2022-08-19
Bug Fix: "Samba Share (using straight RemoteFileSystem block) does not accept the keyword overwrite"

darrida:
### Summary
- Bug fix suggestion for #6289
- This is fixes the SMB issue related to `overwrite=True` around filesystems.py:290
- The scaffolding needed is already in the `RemoteFileSystem.put_directory` method -- this flips the default `overwrite` parameter from `True` to `False`, then handles the need for `overwrite=True` in the respective `put_directory` methods in the S3, GCS, and Azure storage blocks themselves.
- The end result is RemoteFileSystem defaulting back to how it functioned before the issue report above, and the Azure issue the original `overwrite=True` change was created to fix also remains resolved.

### Steps Taken to QA Changes
- Similarly to when I submitted the SMB storage block, it doesn't seem like there are examples of python unit tests related to storage blocks (which I assume is because storage blocks rely primarily on downstream functionality in fsspec). I did do the following manual testing though...
- RemoteFileSystem User SMB
  - created RemoteFileSystem storage block using 'smb://', and required 'settings' values
  - `deployment build flow.py:main -sb remote-file-system/smb-generic-storage -n smb_generic_flow` to push to S3 storage bucket
  - applied and ran flow from Cloud 2.0, pulling from RemoteFileSystem storage block
  - adjusted and ran deployment build/apply, then ran flow again to ensure that SMB via the RemoteFileSystem block could be overwritten
- S3 block
  - created S3 storage block
  - `deployment build flow.py:main -sb s3/s3-storage -n s3_flow` to push to S3 storage bucket
  - applied and ran flow from Cloud 2.0, pulling from S3 storage block
  - adjusted and ran deployment build/apply, then ran flow again to ensure that S3 block could be overwritten
- Azure block
  - created Azure storage block
  - `deployment build flow.py:main -sb azure/azure-storage -n azure_flow` to push to Azure blob storage
  - applied and ran flow from Cloud 2.0, pulling from Azure storage block
  - adjusted and ran deployment build/apply, then ran flow again to ensure that Azure block could be overwritten
- SMB block
  - created SMB storage block
  - `deployment build flow.py:main -sb smb/smb-storage -n smb_flow` to push to SMB Windows share
  - applied and ran flow from Cloud 2.0, pulling from SMB storage block
  - adjusted and ran deployment build/apply, then ran flow again to ensure that SMB block could be overwritten
- GCS block
  - I don't have a GCS account
  - That said, these small adjustments shouldn't actually change any existing behavior for S3, Azure, GCS, or SMB
  - If this is really wanted/needed, I can create a free account and do my own testing

### Note
- It seems like Azure might be the only of the 3 cloud storage blocks that really needs `overwrite=True`, since S3 and GCS were available for quite a quite before Azure (and `overwrite=True`) was added. But, it also seems like S3 and GCS continued functioning fine after that as well (based on lack of cloud service specific bug reports related to that change -- and my own first hand testing here).

### Checklist

This pull request is:

- [x] A short code fix


peytonrunyan:
Thanks a ton for the thorough write-up and QA! I want to speak with somebody more familiar with blocks just to make sure that I'm not missing context on the decision behind the default, but this otherwise looks good to me.

darrida:
> Thanks a ton for the thorough write-up and QA! I want to speak with somebody more familiar with blocks just to make sure that I'm not missing context on the decision behind the default, but this otherwise looks good to me.

Makes total sense to me. Thanks for following up.

darrida:
Btw -- Rereading your comment makes me realize that I have some level of context to offer (obviously not the whole picture).

The original default was deeper inside the  RemoteFileSystem.put_directory method, hardcoded in the "file systems.put_file" method call, so there was no toggle. My pull request submitting the SMB block is what hoisted it out to the class method parameter level, so that I could set it to False for the SMB block (since otherwise it throws the error the related/linked-to Issue specifies). 

I set it to default to True so that it preserved the original behavior it had for S3, Azure, GCS, and the general RemoteFileSystem block with the fewest changes possible.

In some ways I think the flip that I'm suggesting here is what I should've included with the SMB block in the first place, since it preserves the behavior of the original cloud storage blocks, enables the SMB to leave out the "overwrite" parameter entirely (and other hypothetical future blocks), and allows the RemoteFileSystem block to function the way it did originally by default.

peytonrunyan (dismissed):

darrida:
If there is a concern that some users have already created custom blocks around RemoteFileSystem that depend on the "overwrite=True", the parameter could always be hoisted further up into the constructor at RemoteFileSystem class level. It could default to "True", but it would be available to change to "False" for `fsspec` file system options that the "overwrite=True" default might render unusable.

peytonrunyan:
Hey @darrida I'm sorry it's taken me so long to get back to you! I had a chance to talk to with @madkinsz about this. He'd like to take a bit more time to think this over, but it might be a week or two.

madkinsz:
Closing this as stale; please feel free to respond if you want to continue work on this.

@desertaxle please keep this in mind during new feature work.

Issue #6476, opened 2022-08-19
Port in DockerContainer image is interpreted as image tag


MSSandroid:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

`DockerContainer` does not correctly parse images with a port in the url when pulling. The offending function is `DockerContainer._get_image_and_tag`. The port is interpreted as the tag.

### Reproduction

```python
from prefect.infrastructure.docker import DockerContainer

d = DockerContainer(image='some_url_to_a_docker_repo.com:5050/path/to/image:tag1') 
image, tag = d._get_image_and_tag()

assert image == 'some_url_to_a_docker_repo.com:5050/path/to/image'
assert tag == 'tag1'
```

### Error

```

# Copy complete stack trace and error message here, including log output if applicable.

```

### Versions

2.1.1

PR #6477, opened 2022-08-19
Correct parsing of DockerContainer image with port 

MSSandroid:
### Summary

closes https://github.com/PrefectHQ/prefect/issues/6476.
When defining an image in DockerContainer with a port in the registry url, the function `_get_image_and_tag` confuses the port with the tag. I corrected this behavior by assuming that tags never have a `/` slash and after the port a `/` must be present. 
### Steps Taken to QA Changes

Created test file
### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


tests/docker/test_dockercontainer_image_parsing.py
```diff
@@ -0,0 +1,12 @@
+from prefect.infrastructure.docker import DockerContainer
+
+def docker_image_parsing(image, tags):
+    d = DockerContainer(image=':'.join([image]+tags))
+    parsed_image, parsed_tag = d._get_image_and_tag()
+    assert parsed_image == image
+    assert parsed_tag == (tags[0] if len(tags)>0 else None)
+
+def test_dockercontainer_image_parsing():
```
madkinsz, line 9: Could you use `pytest.parametrize` to test your image and tag combinations? We'll also want these tests to be in `tests/docker/test_parsing.py` for a parsing utility and `tests/infrastructure/test_docker_container.py` for anything that uses the `DockerContainer` type.

src/prefect/infrastructure/docker.py
```diff
@@ -268,8 +268,12 @@ def _create_and_start_container(self) -> str:
...
     def _get_image_and_tag(self) -> Tuple[str, Optional[str]]:
         parts = self.image.split(":")
-        image = parts.pop(0)
-        tag = parts[0] if parts else None
+        try:
+            image = ':'.join(parts[:2]) if '/' in parts[1] else parts[0]
+            tag = parts[2] if '/' in parts[1] else parts[1]
+        except IndexError:
+            image= self.image
+            tag = None
```
madkinsz, line 276: I'm having a hard time following this code now and I'm worried it doesn't account for all of the cases we'll encounter. I think we should introduce a utility in `prefect.docker` called `parse_image_tag` that is reused here. We should include a docstring that explains the expected format and output. It looks like there's some valuable information in this thread https://stackoverflow.com/questions/42115777/parsing-docker-image-tag-into-component-parts

cc @chrisguidry

madkinsz (changes requested):
Thanks for contributing a fix! We appreciate it a lot. 

I've requested some changes to improve this implementation, but it might be kind of hard. Let us know if you have any questions.

tests/docker/test_dockercontainer_image_parsing.py
MSSandroid: I mode the requested changes, but saw no need to include a test specific for DockerContainer.

src/prefect/infrastructure/docker.py
MSSandroid: I wrote a more legible version of the function. The functionallity is nearly the same. Your original script could cope with more than on tag and my old version could do that to (by only using the first tag). Since that is not really format compliant, I now through an error in these cases.

tests/docker/test_parsing.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from prefect.docker import parse_image_tag
+from itertools import product
+
+image_names = ['some_url_to_a_docker_repo.com:5050/path/to/image','some_url_to_a_docker_repo.com/path/to/image']
+image_tags = [[], ['tag1']]
+@pytest.mark.parametrize('docker_image,image_name,image_tag', [(':'.join([image]+tag), image, tag) for image, tag in product(image_names, image_tags)])
+def test_docker_image_parsing(docker_image,image_name,image_tag):
```
peytonrunyan, line 8: @MSSandroid thanks for adding the tests and moving it to a separate function! I'm reviewing this now, so I may have a few more comments in a bit. 

A couple of things on the tests - a lot of us (and other users) wind up going straight to the tests to quickly determine how a function is supposed to behave, because we can see the expected inputs and outputs. Pytest's parameterize is really helpful for this. It's also helpful because it allows us to at a glance check for edge case coverage. 

The way your have your values setup inside of parameterize, it takes a quite a bit of effort to understand what the input is, and what the output is. Could you explicitly list the input values and the expected outputs similar to 
```
    @pytest.mark.parametrize(
        "value,expected",
        [
            (<input_value_1>, <expected_output_1>),
            (<input_value_2>, <expected_output_2>),
            (<input_value_3>, <expected_output_3>),
        ],
    )
    def test_my_func(self, value, expected)
```

Could you also add a couple of parameterized tests to make sure that your error is being raised under the expected conditions?

tests/docker/test_parsing.py
```diff
@@ -0,0 +1,12 @@
...
+    parsed_image, parsed_tag = parse_image_tag(docker_image)
+    assert parsed_image == image
```
peytonrunyan, line 10: This line raises an error `NameError: name 'image' is not defined`. 

Just in case you're not familiar with our testing setup, you can install and test the forked version of Prefect by starting a virtual environment, going into the root of the repo, and running `pip install -e ".[dev]"`, and then running `pytest tests/docker/test_parsing.py `. It's also worth running pytest for any other code that yours might impact, so `pytest tests/docker` and `pytest tests/infrastructure/test_docker_container.py`.

peytonrunyan: One of your cases fails when parsed (the second case), so that's probably worth checking out in the function itself.
```
-> assert parsed_image == image_name
(Pdb) parsed_image
'5:0:5:0:/:p:a:t:h:/:t:o:/:i:m:a:g:e'
```

madkinsz:
Thanks for working on this and sorry about the ping. We'll merge this with #6564 to create the final pull request.

Issue #6485, opened 2022-08-19
Flow-level retries don't work with deployed flow runs


anna-geller:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When running this flow locally, it will retry 2 times as expected:

```python
from prefect import flow


@flow(retries=2, retry_delay_seconds=2)
def healthcheck():
    raise ValueError("fail and retry")


if __name__ == "__main__":
    healthcheck()
```

But when this flow run is deployed, flow-level retries don't apply:
![image](https://user-images.githubusercontent.com/86264395/185681248-c03d6f24-b725-4dce-87ad-f8d989e43fba.png)

### Reproduction

```python
see above
```

### Versions

```
Version:             2.1.1
API version:         0.8.0
Python version:      3.9.12
Git commit:          dc2ba222
Built:               Thu, Aug 18, 2022 10:18 AM
OS/Arch:             darwin/arm64
Profile:             dev
Server type:         hosted
```

PR #6486, opened 2022-08-19
Add `empirical_policy` to flow run update route

madkinsz:
### Summary

This will facilitate updating orchestration rules discovered from flow-level annotations for deployment flow runs.

Related to #6485 
See usage in https://github.com/PrefectHQ/prefect/pull/6489

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/client.py
+++ b/src/prefect/client.py
@@ -624,15 +624,21 @@ class OrionClient:
         parameters: Optional[dict] = None,
         name: Optional[str] = None,
         tags: Optional[Iterable[str]] = None,
-    ) -> None:
+        empirical_policy: Optional[schemas.core.FlowRunPolicy] = None,
+    ) -> httpx.Response:
         """
         Update a flow run's details.
 
         Args:
-            flow_run_id: the run ID to update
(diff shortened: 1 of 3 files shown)


chrisguidry (approved):
Looks good!  It took my a while to understand why the parameters to `update_flow_run` don't line up with `create_flow_run`.  Is it because we don't have access to the `flow` when we update, and/or we want to limit to a specific set of things that may be updated?

madkinsz:
A different set of things can be updated compared to initial creation and generally you don't want to update everything. I think we should consider having the client methods be a little less magical and consume an explicit list of parameters rather than a `Flow` directly in the future.

cicdw (approved):

cicdw:
@madkinsz does this raise any Cloud API considerations given that there's a new field on the `FlowRunUpdate`?

madkinsz:
At the very least Cloud will need to bump its Orion pin to support clients with this change.

PR #6501, opened 2022-08-21
Docs add task call from another task example

discdiver:
Updates tasks concepts page to include calling task form another task. Also updates tasks docstring.

### Summary

The task docs currently incorrectly state that a task cannot be called from another task. Corrects that and adds an example.

### Steps Taken to QA Changes

Tested the code.

### Checklist

This pull request is:

- [ x] A documentation / typographical error fix
	- No tests or issue needed
- [ ] A short code fix
	- Please include tests

**Happy engineering!**


docs/concepts/tasks.md
```diff
@@ -53,6 +55,29 @@ Tasks are uniquely identified by a task key, which is a hash composed of the tas
 
     To be clear, there's nothing stopping you from putting all of your code in a single task &mdash; Prefect will happily run it! However, if any line of code fails, the entire task will fail and must be retried from the beginning. This can be avoided by splitting the code into multiple dependent tasks.
 
+**Calling a task from another task**
```
anna-geller, line 58: I like the idea of showing how to call a task's function, but the wording of calling a task from another task could be a little misleading since in the end you only run `my_second_task` which calls `my_first_task`'s function but it doesn't trigger its task run -- this could get confusing in the UI 
image

I think we could keep it but phrase it as "Calling a task's function from another task" and add a note sth like:

Note that this way you are only calling the task's function without actually generating a task run. It has a negative effect on your observability - Prefect won't track task execution in your Prefect backend if you call it this way.

anna-geller (changes requested):
Thanks for creating this! I'd rephrase this a little otherwise LGTM

discdiver:
Good points.

docs/concepts/tasks.md
discdiver, line 58:
```suggestion
**Calling a task's function from another task**
```

docs/concepts/tasks.md
```diff
@@ -53,6 +55,29 @@ Tasks are uniquely identified by a task key, which is a hash composed of the tas
...
+def my_second_task(msg):
+    my_first_task.fn(msg)
+
+@flow
+def my_flow():
+    my_second_task("Trillian")
+
+```
+
+
```
discdiver, line 79:
```suggestion
Note that in the example above you are only calling the task's function without actually generating a task run. This pattern has a negative effect on your observability. Prefect won't track task execution in your Prefect backend if you call the task function this way.
```

discdiver (commented):
updated to reflect Anna's suggested changes regarding drawbacks

docs/concepts/tasks.md
```diff
@@ -53,6 +55,29 @@ Tasks are uniquely identified by a task key, which is a hash composed of the tas
 
     To be clear, there's nothing stopping you from putting all of your code in a single task &mdash; Prefect will happily run it! However, if any line of code fails, the entire task will fail and must be retried from the beginning. This can be avoided by splitting the code into multiple dependent tasks.
 
+**Calling a task's function from another task**
+
+A task can be called from another task by use of the `.fn()` method. Pass any arguments to `fn`.
```
madkinsz, line 60: This isn't actually the `fn` method, you're accessing the `fn` attribute which happens to be your function. Perhaps...

```suggestion
Prefect does not allow execution of tasks within other tasks. If you would like to call your task's function directly, you can access it at `task.fn`. You may then call it without any Prefect orchestration or observation.
```

docs/concepts/tasks.md
```diff
@@ -53,6 +55,29 @@ Tasks are uniquely identified by a task key, which is a hash composed of the tas
...
+def my_second_task(msg):
+    my_first_task.fn(msg)
+
+@flow
+def my_flow():
+    my_second_task("Trillian")
+
+```
+
+Note that in the example above you are only calling the task's function without actually generating a task run. This pattern has a negative effect on your observability. Prefect won't track task execution in your Prefect backend if you call the task function this way.
```
madkinsz, line 79: There also won't be any orchestration behavior like retries.

tpdorsey:
Perhaps consider leaving the main text as stating that you cannot call a _task_ from a _task_, and move discussion of calling a function from a task as a `!!! warning` callout. I'm concerned that parsing out the details inline might be considered promoting what is actually an anti-pattern.

docs/concepts/tasks.md
discdiver, line 79: ``
```suggestion
Note that in the example above you are only calling the task's function without actually generating a task run. This pattern has a negative effect on observability and orchestration. Prefect won't track this call's task execution in your Prefect backend. And will you be able to add orchestration benefits such as retries.
```

discdiver: Good point.

discdiver (commented):
add @madkinsz 's improvements, in case we decide to leave the changes inline

discdiver:
> Perhaps consider leaving the main text as stating that you cannot call a _task_ from a _task_, and move discussion of calling a function from a task as a `!!! warning` callout. I'm concerned that parsing out the details inline might be considered promoting what is actually an anti-pattern.

Good idea @tpdorsey . Added warning. I think I've incorporates all requested changes.

docs/concepts/tasks.md
```diff
@@ -30,7 +30,9 @@ Tasks are special because they receive metadata about upstream dependencies and
 
 Tasks also take advantage of automatic Prefect [logging](/concepts/logs) to capture details about task runs such as runtime, tags, and final state. 
 
-You can define your tasks within the same file as your flow definition, or you can define tasks within modules and import them for use in your flow definitions. All tasks must be called from within a flow. Tasks may not be called from other tasks.
+You can define your tasks within the same file as your flow definition, or you can define tasks within modules and import them for use in your flow definitions. All tasks must be called from within a flow or another task.
```
madkinsz, line 33: Should probably restore this old line
```suggestion
You can define your tasks within the same file as your flow definition, or you can define tasks within modules and import them for use in your flow definitions. All tasks must be called from within a flow. Tasks may not be called from other tasks.
```

docs/concepts/tasks.md
```diff
@@ -53,6 +55,29 @@ Tasks are uniquely identified by a task key, which is a hash composed of the tas
 
     To be clear, there's nothing stopping you from putting all of your code in a single task &mdash; Prefect will happily run it! However, if any line of code fails, the entire task will fail and must be retried from the beginning. This can be avoided by splitting the code into multiple dependent tasks.
 
+!!! warning "Calling a task's function from another task"
+
+    Prefect does not allow execution of tasks within other tasks. If you would like to call your task's function directly, you can access it at `task.fn`. You may then call it without any Prefect orchestration or observation.
```
anna-geller, line 60:
```suggestion
    Prefect does not allow triggering task runs from other tasks. If you want to call your task's function directly, you can use `task.fn()`. Note that this way you are only calling the task's function and you lose orchestration and observability provided by Prefect.
```

anna-geller (approved):
Added some comments, otherwise LGTM

PR #6513, opened 2022-08-22
Only calculate lengths of iterable arguments to `map`

### LINKED ISSUE #6383: Optional arguments must be explicitly passed with map
### First check

### Bug summary

https://discourse.prefect.io/t/task-map-interaction-with-optional-arguments/1372/3

For mapped tasks with optional single value arguments I have found that I need to wrap the optional value in the function definition with unmapped as shown below. The unmapped optional task parameter forces the task to be invoked with .map or it fails.

This is different than behavior in v1.

### Reproduction

```python
**v2**

from typing import List
from prefect.utilities.annotations import unmapped

@task
def mytask(x: int, y: int = unmapped(5)) -> int:
    return x + y

@flow
def myflow(c: int) -> List[int]:
    return mytask.map(list(range(c)))

if __name__ == "__main__":
    res = myflow(4)
    print(res)


**v1**
```python
from typing import List
from prefect import task, Flow, unmapped

@task
# NOTE that unmapped(5) is not needed on the y optional input
# to make this work in v1
def mytask(x: int, y: int = 5) -> int: 
    return x + y

if __name__ == "__main__":
    with Flow("MyFlow") as flow:
        mytask.map(list(range(4)))

    state = flow.run()
    print(state.result[flow.get_tasks()[0]].result)
```
```


### Error

_No response_

### Versions

```
Version:             2.0.4
API version:         0.8.0
Python version:      3.8.13
Git commit:          39db6fb1
Built:               Wed, Aug 10, 2022 1:19 PM
Profile:             default
Server type:         hosted
```


### Additional context

Initially opened the discussion on discourse
https://discourse.prefect.io/t/task-map-interaction-with-optional-arguments/1372/4

### PR DESCRIPTION
**Pull request by bunchesofdonald**
### Summary
Issue #6383 calls out that a task with a default keyword argument fails when being mapped unless the task has defaulted that keyword argument to an `unmapped` value. This PR alters the way `map` works so that it only calculates the length of iterable arguments, which makes the use of `unmapped` un-needed.

This does result in a change to behavior. When `map` is called without any parameters, previously it would run the task a single time, now it will fail with a `MappingMissingIterable` exception. The thinking behind the exception is that if a user calls `.map` without an iterable then it's no different than calling the task directly and the user is either using an incorrect way of calling `map` or the arguments they're passing are incorrect.

Closes #6383 

### Steps Taken to QA Changes
- Unit tests
- Tested example code in #6383 

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### CODE REVIEW

**Review by madkinsz — COMMENTED** [review]
This seems like really reasonable behavior, but I want to point out a change that I wasn't expecting. The original issue was that parameter defaults were not implicitly treated as `unmapped` e.g.

```python
@task
def foo(x: int, y: int = 2)
    return x + y

foo.map(x=[1, 2, 3])  # Fails since `y` is not an iterable
```

The writer of the task should not need to account for it being used in mapping, hence this bug fix. However, this implementation comes with an interesting side-effect where the following will succeed as well:

```python
@task
def foo(x: int, y: int = 2)
    return x + y + z 

foo.map(x=[1,2,3], y=10)
```

The user has explicitly provided a value for `y` but it is not mapped. I think this is a nice way to prevent the user from having to explicitly use `unmapped` in most cases! However, it might be confusing if they intended to pass an iterable and did not get a error when it was a scalar. I think this trade-off is worth it, but we should probably get feedback from @billpalombi

bunchesofdonald:
Agreed, and I started down the route of implicitly marking the default arguments as unmapped (you can check out that version over here: https://github.com/PrefectHQ/prefect/commit/60f3d8bc846ef2208816211bd8d669e7e8e11007 But it felt like we were just coding around our own limitations.

madkinsz (approved):
Implementation LGTM. Can you explain this behavior in the `map` docstring? Is there a good place to clarify this behavior in the concepts?

billpalombi (approved):
Thanks for tagging me in @madkinsz! I agree that the trade off is worth it. This change is definitely an improvement over the current state. I'll keep a look out for users that are confused by the implicitly unmapped values, but it seems pretty intuitive to me.

Issue #6497, opened 2022-08-19
Sync flows cannot run async flows


ahuang11:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Not sure if this is a bug or feature request, but if I try to run an async flow inside a sync flow, it says:
```
/var/folders/tb/fksvh8ms2hl5xnzwrftnx66r0000gq/T/ipykernel_96002/1050154831.py:10: RuntimeWarning: coroutine 'create_and_begin_subflow_run' was never awaited
```

### Reproduction

```python
import asyncio
from prefect import flow

@flow
async def test_async_flow():
    asyncio.sleep(3)

@flow
def test_sync_flow():
    test_async_flow()

test_sync_flow()
```

### Error

```
/var/folders/tb/fksvh8ms2hl5xnzwrftnx66r0000gq/T/ipykernel_96002/1050154831.py:10: RuntimeWarning: coroutine 'create_and_begin_subflow_run' was never awaited
```

### Versions

```
Version:             2.0.4+292.gdc2ba2221
API version:         0.8.0
Python version:      3.9.7
Git commit:          dc2ba222
Built:               Thu, Aug 18, 2022 10:18 AM
OS/Arch:             darwin/arm64
Profile:             cloud
Server type:         hosted
```

### Additional context

There's a prefect-databricks flow that is async, but would like to be used in sync flows.
https://github.com/PrefectHQ/prefect-databricks/blob/main/prefect_databricks/flows.py#L57

Also a dbt flow:
https://github.com/PrefectHQ/prefect-dbt/blob/main/prefect_dbt/cloud/jobs.py#L172

PR #6514, opened 2022-08-22
Fixes sync calling async subflows

ahuang11:
### Summary

Closes https://github.com/PrefectHQ/prefect/issues/6497

The async subflow gets handled downstream:
```
    # Sync subflow run
    if not parent_flow_run_context.flow.isasync:
        return run_async_from_worker_thread(begin_run)
    else:
        return parent_flow_run_context.sync_portal.call(begin_run)
```

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


bunchesofdonald (approved):
I was just looking at this issue and this is essentially what I was about to do, so it's a +1 from me.

ahuang11:
> I was just looking at this issue and this is essentially what I was about to do, so it's a +1 from me.

Yes, right after I made a PR and was linking the issue, I saw that! Sorry I didn't self-assign it earlier, and thanks for helping review

src/prefect/engine.py
```diff
@@ -136,8 +136,8 @@ def enter_flow_run_engine_from_flow_call(
         return_type=return_type,
     )
 
-    # Async flow run
-    if flow.isasync:
+    # Async flow run (not async subflow)
```
madkinsz, line 139: Can you update the comments to mirror those for `enter_task_run_engine`?

madkinsz (commented):
We'll need product sign-off on this. I don't remember why we weren't supporting this in the first place. cc @cicdw / @billpalombi

billpalombi:
Good catch! I'll defer to @cicdw here but I don't think it was a deliberate decision not to support calling async flows from sync flows. As long as there aren't any unintuitive side effects, I totally support this.

cicdw (approved):
LGTM, agree with everyone here that this was most likely not intentional

PR #6521, opened 2022-08-23
Revert breaking schema change

zangell44:
### Summary

All clients from `2.1.1` and before will break by making these properties now optional.

Instead of making the properties optional, we have marked them as deprecated and introduced two new properties.

For consistency, the updates have also been applied to `TaskRunPolicy`.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/orion/schemas/core.py
+++ b/src/prefect/orion/schemas/core.py
@@ -106,6 +106,21 @@ class FlowRunPolicy(PrefectBaseModel):
         None, description="The delay time between retries, in seconds."
     )
 
+    @root_validator
+    def populate_deprecated_fields(cls, values):
+        """
+        If deprecated fields are provided, populate the corresponding new fields
+        to preserve orchestration behavior.
+        """
+        if not values.get("retries", None) and values.get("max_retries", 0) != 0:
+            values["retries"] = values["max_retries"]
+        if (
+            not values.get("retry_delay", None)
+            and values.get("retry_delay_seconds", 0) != 0
+        ):
+            values["retry_delay"] = values["retry_delay_seconds"]
+        return values
+
 
 class FlowRun(ORMBaseModel):
     """An ORM representation of flow run data."""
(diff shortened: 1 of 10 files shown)


src/prefect/orion/schemas/core.py
```diff
@@ -90,8 +90,21 @@ class FlowRunPolicy(PrefectBaseModel):
...
+    max_retries: int = Field(
+        0,
+        description="The maximum number of retries. Field is not used. Please use `retries` instead.",
+        deprecated=True,
+    )
+    retry_delay_seconds: float = Field(
+        0,
+        description="The delay between retries. Field is not used. Please use `retry_delay` instead.",
+        deprecated=True,
+    )
```
chrisguidry, line 102: I'm not sure of the current status of these fields and how/where they're provided by clients.  Is there any value to having a `@root_validator` that uses the old values to populate the new values (if the new values are `None` and the old values are non-zero)?

chrisguidry (approved):
This looks solid, just a question about whether we could take one more backwards compatibility step...

src/prefect/orion/schemas/core.py
zangell44: Hmmm that's a fair point. Would that break any assumptions for old clients @madkinsz?

madkinsz: We're here with these changes, right?

Old client -> Sends old fields
New server -> Ignores orchestration behavior

New client -> Sends new fields
Old server -> Throws an error

And introducing a root validator would get us to

Old client -> Sends old fields
New server -> Uses root validator to populate new fields with old
New server -> Expected orchestration behavior is preserved

tests/orion/schemas/test_core.py
```diff
@@ -95,6 +95,14 @@ async def test_multiple_message_template_variables_are_validated(self):
 
 
 class TestFlowRunPolicy:
+    class OldFlowRunPolicy(PrefectBaseModel):
```
chrisguidry, line 98: I love this idea of including the previous version too, so cool

madkinsz (approved):

tests/orion/schemas/test_core.py
abrookins: Should this go farther and identify in the name which version of the model this is?

abrookins: Ah, I see it's in the test docstring, maybe we can change the name of the class if we ever need to do so (i.e. there are _two_ old flow run policies we test...)

PR #6527, opened 2022-08-23
Refactor Task Run creation

anticorrelator:
### Summary
Task runs are created by the engine before submitting the work to the Task Runner. Because the main thread waits for task run creation before the information can be passed to the Task Runner, the rate of task run submission can be slow, especially when using a concurrent Task Runner to submit many tasks at once. Here we background Task Run creation when using a concurrent task runner and return a future immediately instead. This brings task submission performance in line with Prefect 1.

co-authored with @zangell44 and @madkinsz 

### Steps Taken to QA Changes
Manually tested by submitting hundreds of tasks concurrently before and after this change.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**


src/prefect/futures.py
```diff
@@ -104,18 +106,23 @@ class PrefectFuture(Generic[R, A]):
...
-        self.run_key = run_key
+        self.key = key
+        self.name = name
         self.asynchronous = asynchronous
+        self.task_run = None
         self._final_state = _final_state
         self._exception: Optional[Exception] = None
         self._task_runner = task_runner
+        self._submitted = anyio.Event()
+        import asyncio
```
bunchesofdonald, line 123: Why avoid the import until here? Might be worth a comment.

src/prefect/futures.py
```diff
@@ -263,11 +272,23 @@ async def _get_state(self, client: OrionClient = None) -> State[R]:
...
+        else:
+            await self._submitted.wait()
+
     def __hash__(self) -> int:
-        return hash(self.run_key)
+        return hash(self.key)
 
     def __repr__(self) -> str:
-        return f"PrefectFuture({self.run_key!r})"
+        return f"PrefectFuture({self.name!r})"
```
bunchesofdonald, line 291: 👏 This should make for more readable output.

src/prefect/task_runners.py
```diff
@@ -303,29 +271,29 @@ async def wait(
                 "serialization."
             )
 
-        return await self._get_run_result(prefect_future.run_key, timeout)
+        return await self._get_run_result(key, timeout)
 
-    async def _run_and_store_result(self, run_key: str, run_fn, run_kwargs):
+    async def _run_and_store_result(self, key: UUID, call):
```
bunchesofdonald, line 276:
```suggestion
    async def _run_and_store_result(self, key: UUID, call: Callable[[], Awaitable[State[R]]]):
```

tests/test_engine.py
```diff
@@ -83,24 +83,50 @@ def flow_run_caplog(caplog):
         logger.propagate = False
 
 
+# @pytest.fixture
```
bunchesofdonald, line 86: Commented out code here, should we remove it?

tests/test_engine.py
```diff
@@ -1226,29 +1261,14 @@ def raise_unspecified_exception(*args, **kwargs):
...
     async def test_handles_bad_parameter_types(
-        self, orion_client, parameterized_flow, local_filesystem
+        self,
+        orion_client,
+        parameterized_flow,
+        get_flow_run_context,
     ):
-        ctx = await self.create_test_context(orion_client, local_filesystem)
-        with ctx:
+        # ctx = await self.create_test_context(orion_client, local_filesystem)
```
bunchesofdonald, line 1270: More commented out code, I'd remove it.

bunchesofdonald (approved):
I love this, I think it makes things more performant as well as pulling Future creation up a level which I _think_ will make it all a bit easier to reason about.

src/prefect/futures.py
madkinsz: This is left over from my sketch, should be moved I think.

src/prefect/futures.py
madkinsz: It's the same content actually, just a clearer usage.

bunchesofdonald: Oh, it seemed like it was dropping some other parts, but better name is good too.

src/prefect/futures.py
```diff
@@ -104,18 +106,23 @@ class PrefectFuture(Generic[R, A]):
 
     def __init__(
         self,
-        task_run: TaskRun,
-        run_key: str,
+        name: str,
+        key: UUID,
```
zangell44, line 110: I changed key to be `f"{flow_run_context.flow_run.id}-{task.task_key}-{dynamic_key}"
    )` in line 880 of `engine.py` and forgot to change the types here.

Are we alright with that? I think it makes sense given it's a unique key for the task.
But hmmm perhaps we need to include flow and task retry counts too?

madkinsz: Task retries will occur _within_ the submitted call. Flow run retries would need to be included though. I don't think we should mix real information about the task into this key. This should be a unique identifier to facilitate lookup of the future's result and as you've noted collisions are a concern.

tests/test_tasks.py
```diff
@@ -558,9 +558,9 @@ def flaky_function():
         @flow
         def test_flow():
             future = flaky_function.submit()
-            return future.wait()
+            return future.wait(), ...
```
madkinsz, line 561: Was this updating the flow run with the task run state? You can use `quote` if so.

src/prefect/futures.py
zangell44: Hmm okay should we change this back to a `uuid` then?

PR #6529, opened 2022-08-23
maintenance: playwright tests prep

stackoverfloweth:
base work that I build upon for work-queue tests
- mostly just moves utils into separate files
- ~adds `useToggle` util~ will add useToggle in follow up PR where I need it
- uses mocker for dummy data 🚀


orion-ui/tests/utilities/useContextSidebar.ts
```diff
@@ -0,0 +1,24 @@
...
+  navigate: (destination: string) => Promise<void>,
+}
+
+export function useContextSidebar(page: Page | Locator = PAGE): UseContextSidebar {
+  const sidebar = page.locator('.app__sidebar')
+
+  return {
+    sidebar,
+    navigate: async (destination) => {
+      const { link } = useLink(destination, sidebar)
```
pleek91, line 16: ❤️

orion-ui/tests/utilities/useToggle.ts
```diff
@@ -0,0 +1,55 @@
+import { Locator, Page } from '@playwright/test'
+import { PAGE } from './fixtures'
+
+export type UseToggle = {
+  toggle: Locator,
+  getState: () => Promise<boolean | undefined>,
+  setState: (state: boolean) => Promise<void>,
+}
+
+export function useToggle(page: Page | Locator = PAGE): UseToggle {
```
pleek91, line 10: This appears to act on all toggles within the page or locator which doesn't seem ideal. Could this work on a single toggle and fail if multiple toggles are found within a locator? I can't think of a use case where we would want to act on multiple toggles at the same time. And if we do, useToggle should probably be called multiple times to intentionally select the desired toggles.

pleek91: And by fail I mean just have it return the locates toggle(s) but any methods that do actions would fail because of playwrights strict check on actions being called on single locators

orion-ui/tests/utilities/useToggle.ts
```diff
@@ -0,0 +1,55 @@
...
+function getToggle(page: Page | Locator): Locator {
+  return page.locator('.p-toggle')
+}
+
+async function setToggleState(toggle: Locator, state: boolean): Promise<void> {
+  const current = await getToggleState(toggle)
+
+  if (current !== state) {
+    await toggle.click()
+    await toggle.locator(`[aria-checked="${state}"]`).waitFor()
```
pleek91, line 49: I would remove this wait for. Generally we shouldn't have to be waiting for much and let playwright locators do their thing.

pleek91 (changes requested):

madkinsz:
Can you use the `maintenance` label for this since it's not a user-facing feature?

orion-ui/tests/utilities/useToggle.ts
stackoverfloweth: without this wait you can't write a test like

```ts
const { getState, setState } = useToggle(worksQueue)
const currentState = await getState()
await setState(!currentState) // this is where the `waitFor` is currently
const newState = await getState() // without the `waitFor` this `getState` won't be the updated value
```

orion-ui/tests/utilities/useToggle.ts
stackoverfloweth: yeah I like that. I was trying to follow what I thought was the existing expectation that these utilities work the same on a single element as they do a list

pleek91: No I don't think any of the utilities work on a list.

orion-ui/tests/utilities/useToggle.ts
stackoverfloweth: @pleek91 what do you think about moving it into a separate method

```ts
export function useToggle(page: Page | Locator = PAGE): UseToggle {
  return {
    toggle: getToggle(page),
    getState: (toggle: Locator = getToggle(page)) => {
      return isChecked(toggle)
    },
    setState: async (state, toggle: Locator = getToggle(page)) => {
      const current = await isChecked(toggle)

      if (current !== state) {
        toggle.click()
      }
    },
    waitFor: (state, toggle: Locator = getToggle(page)) => {
      return toggle.locator(`[aria-checked="${state}"]`).waitFor()
    },
  }
}
```

pleek91 (approved):

PR #6534, opened 2022-08-23
Fix `S3List` error on empty list

madkinsz:
Updates #6028 with `1.x`.


madkinsz:
@desertaxle this is ready for review now

desertaxle (approved):

PR #6540, opened 2022-08-24
testing pydantic v1.10.0a2

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,7 @@ kubernetes >= 24.2.0
 packaging >= 21.3
 pathspec >= 0.8.0
 pendulum >= 2.1.2
-pydantic >= 1.8.2, != 1.9.1
+pydantic == 1.10.0a2
 python-slugify >= 5.0
 pytz >= 2021.1
 pyyaml >= 5.4.1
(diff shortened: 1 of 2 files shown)


samuelcolvin:
Hi, please can you kick off tests so I can check if v1.10 will break anything.

madkinsz:
Welcome! Please let us know if there's anything we can do to help. Feel free to ping me here our in your tracking issue.

Some brief notes:
- The Python 3.7 tests run with the lowest version pin. Looks like the script doesn't support alpha labels so it failed on install.
- We test with SQLite and Postgres as backing databases
- Our `--only-services` tests run integration tests with things like Docker that are slow to test against

samuelcolvin:
So from what I can tell, except this issue with lowest version pinning on 3.7, pydantic v1.10 seems to be compatible with prefect. Is that correct?

madkinsz:
Here's a fix for that issue https://github.com/PrefectHQ/prefect/pull/6543 if you're worried specifically about 3.7 compatibility. I imagine that's not a concern though.

Looks like we're compatible yes :)

samuelcolvin:
hopefully I've copied the important part of that commit.

Please could you kick off tests, by the way [you should consider changing the "awaiting approval" settings](https://twitter.com/samuel_colvin/status/1562374350287867904).

samuelcolvin:
okay, hopefully final time :crossed_fingers:.

madkinsz:
🎉

samuelcolvin:
great, thanks so much.

PR #6541, opened 2022-08-24
Prevent non-UUID slugs from raising errors on the BlockDocuments APIs.

chrisguidry:
### Summary

In Prefect Cloud, we observed some errors when clients would send requests for
`.../block_documents/null`, which should really be handled at the routing layer
with 404s when the path UUIDs can't be parsed.

Note that this is just correcting the server-side issue, but does not attempt
to diagnose the client-side issue at this time.  Also, this does not attempt
to go through every route in Orion that includes UUIDs in its path.

### Steps Taken to QA Changes

Additional regression tests added for this symptom, and no other tests were modified.

### Checklist

This pull request is:

	- No tests or issue needed
- [x] A short code fix
	- Please include tests

**Happy engineering!**

### DIFF
--- a/src/prefect/orion/api/block_documents.py
+++ b/src/prefect/orion/api/block_documents.py
@@ -5,16 +5,7 @@ from typing import List, Optional
 from uuid import UUID
 
 import sqlalchemy as sa
-from fastapi import (
-    Body,
-    Depends,
-    HTTPException,
-    Path,
-    Query,
-    Response,
-    responses,
-    status,
-)
+from fastapi import Body, Depends, HTTPException, Path, Query, Response, status
 
 from prefect.orion import models, schemas
 from prefect.orion.api import dependencies
@@ -74,7 +65,7 @@ async def read_block_documents(
     return result
 
 
-@router.get("/{id}")
+@router.get("/{id:uuid}")
 async def read_block_document_by_id(
     block_document_id: UUID = Path(
         ..., description="The block document id", alias="id"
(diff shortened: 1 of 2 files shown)


src/prefect/orion/api/block_documents.py
```diff
@@ -74,7 +65,7 @@ async def read_block_documents(
     return result
 
 
-@router.get("/{id}")
+@router.get("/{id:uuid}")
```
zangell44, line 68: Is this necessary to get the desired behavior?

chrisguidry: It is, because without that, we get 422s instead of 404s, because the FastAPI kind of assumes that if it makes it as far as a `Depends`, then the URL must be correct it's something in the body of the request that's wrong.  This is the more correct form anyway, using a [starlette path converter](https://www.starlette.io/routing/#path-parameters).

zangell44 (approved):

src/prefect/orion/api/block_documents.py
zangell44: Hmmm is 422 the correct status to return though?
Providing something other than a UUID means we can't process the request. And it's probably useful for the client to know they've provided a bad data type rather than "we couldnt find this" right?

chrisguidry: I believe 404 is the correct status here.  They are requesting an object that doesn't exist ("/block_documents/null")

Here's how I interpret the status codes:
400: something in the _request line_ or _header_ is incorrect
401: you aren't allowed, maybe try authenticating
403: you aren't allowed, and authenticating won't help you

chrisguidry: We could get this error even on a `GET` request today, which doesn't even have a body, so I'm pretty sure 422 doesn't work.

zangell44 (approved):

PR #6545, opened 2022-08-24
Agent version display

cicdw:
Minor change to display prefect version running on agents; accessing this information quickly and efficiently has come up a few times when debugging (**cc:** @dylanbhughes )

### DIFF
--- a/src/prefect/cli/agent.py
+++ b/src/prefect/cli/agent.py
@@ -6,6 +6,7 @@ from uuid import UUID
 
 import typer
 
+import prefect
 from prefect.agent import OrionAgent
 from prefect.cli._types import PrefectTyper, SettingsOption
 from prefect.cli._utilities import exit_with_error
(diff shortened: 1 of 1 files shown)


src/prefect/cli/agent.py
```diff
@@ -106,15 +107,19 @@ async def start(
...
+            app.console.print(
+                f"Starting v{prefect.__version__} agent with ephemeral API..."
+            )
 
     async with OrionAgent(work_queues=work_queues) as agent:
         if not hide_welcome:
             app.console.print(ascii_name)
             app.console.print(
-                "Agent started! Looking for work from "
+                f"Agent started! Looking for work from "
```
chrisguidry, line 122: You don't actually need this `f` here.

chrisguidry (approved):

src/prefect/cli/agent.py
cicdw: 🤦 this is evidence that I spent a lot of time trying to figure out where to put this information without printing a new line

cicdw, line 122:
```suggestion
                "Agent started! Looking for work from "
```

Issue #6544, opened 2022-08-24
Unable to start Orion on new install


richard-alexander:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I have a fresh install of prefect 2.2 into a new conda environment, but orion will not start. It immediately stops with a message `ModuleNotFoundError: No module named 'prefect.orion'`.

### Reproduction

```python
conda activate myprefectenv

pip install prefect

export PREFECT_ORION_DATABASE_PASSWORD='MyPassword'
export PREFECT_ORION_DATABASE_CONNECTION_URL='postgresql+asyncpg://MyUsername:${PREFECT_ORION_DATABASE_PASSWORD}@MyDatabase:5432/prefect'

prefect orion start

```

#### Output of `prefect orion start`

```


 ___ ___ ___ ___ ___ ___ _____    ___  ___ ___ ___  _  _
| _ \ _ \ __| __| __/ __|_   _|  / _ \| _ \_ _/ _ \| \| |
|  _/   / _|| _|| _| (__  | |   | (_) |   /| | (_) | .` |
|_| |_|_\___|_| |___\___| |_|    \___/|_|_\___\___/|_|\_|

Configure Prefect to communicate with the server with:

    prefect config set PREFECT_API_URL=http://127.0.0.1:4200/api

View the API reference documentation at http://127.0.0.1:4200/docs

Check out the dashboard at http://127.0.0.1:4200



Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/prefect2/bin/uvicorn", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniconda3/envs/prefect2/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/prefect2/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/miniconda3/envs/prefect2/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/miniconda3/envs/prefect2/lib/python3.10/site-packages/click/core.py", line 760, in invoke
[...]

PR #6547, opened 2022-08-24
Set uvicorn `--app-dir` when starting orion

bunchesofdonald:
### Summary
In issue #6544 it was reported that a user was unable to get orion to start and it would fail with `No module named 'prefect.orion'` when starting uvicorn. The only way I'm able to reproduce this issue is if I create a `prefect` folder containing an `__init__.py`. By default `uvicorn` uses `.` as its "app dir", this PR updates the uvicorn command to set the app dir to point at the folder where prefect is installed.

### Steps Taken to QA Changes

- Ran unittests
- Started orion in a folder without a prefect folder.
- Started orion in a folder with a prefect folder.

### Checklist

This pull request is:

- [x] A short code fix


src/prefect/cli/orion.py
```diff
@@ -124,6 +125,8 @@ async def start(
                 run_process,
                 command=[
                     "uvicorn",
+                    "--app-dir",
+                    str(Path(prefect.__file__).parent.parent),
```
madkinsz, line 129: We have some constants in `__init__.py` to make this clearer
```

# The absolute path to this module
__module_path__ = pathlib.Path(__file__).parent
# The absolute path to the root of the repository
__root_path__ = __module_path__.parents[1]
# The absolute path to the built UI within the Python module
__ui_static_path__ = __module_path__ / "orion" / "ui"
```

This path is just going to be `site-packages` when a non-editable install is used. Is this the intention?

bunchesofdonald: Yep, it just needs to point at site-packages, uvicorn overrides the python path to `.` so we kind of just need to point it back in the right direction.

bunchesofdonald: I updated it to use `__root_path__`.

madkinsz: Sorry I've mislead you here. `__root_path__` is only valid during development usage. We should clarify that in a command. You want `__module_path__.parent` but root path is another level up because we need to step out of `src` when the repository is being used.

bunchesofdonald: Ha, no worries. Updated again and added a comment to the `__root_path__`.

madkinsz (approved):

PR #6553, opened 2022-08-24
Improve performance of flow run serialization for `/flow_runs/filter` endpoint

zangell44:
In testing we noticed performance issues when using jsonable_encoder with large, nested data structures. FlowRun.parameters is often such a structure.

Not only can it be slow, but FastAPI does not run the call in a worker thread, so it blocks the main event loop.
We also now use `orjson` by default for our pydantic model serialization. In testing, this approach resulted in a 75% reduction in CPU time needed to process problematic flow runs.

See [issue](https://github.com/PrefectHQ/nebula/issues/1301) for more info.

### Checklist

### DIFF
--- a/src/prefect/orion/api/flow_runs.py
+++ b/src/prefect/orion/api/flow_runs.py
@@ -9,6 +9,7 @@ from uuid import UUID
 import pendulum
 import sqlalchemy as sa
 from fastapi import Body, Depends, HTTPException, Path, Response, status
+from fastapi.responses import ORJSONResponse
 
 import prefect.orion.api.dependencies as dependencies
 import prefect.orion.models as models
@@ -155,7 +156,7 @@ async def read_flow_run_graph(
     )
 
 
-@router.post("/filter")
+@router.post("/filter", response_class=ORJSONResponse)
 async def read_flow_runs(
     sort: schemas.sorting.FlowRunSort = Body(schemas.sorting.FlowRunSort.ID_DESC),
     limit: int = dependencies.LimitBody(),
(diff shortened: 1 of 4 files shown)


src/prefect/orion/api/flow_runs.py
```diff
@@ -155,7 +156,7 @@ async def read_flow_run_graph(
     )
 
 
-@router.post("/filter")
+@router.post("/filter", response_class=ORJSONResponse)
```
madkinsz, line 159: Will this break our typing in the API reference?

Can we introduce a custom response type that includes a generic e.g. `JSONResponse[List[FlowRun]]`? Or another more elegant solution?

madkinsz: It does break the API reference documentation:
- https://deploy-preview-6553--prefect-orion.netlify.app/api-ref/rest-api/#/Flow%20Runs/read_flow_runs_flow_runs_filter_post 
- https://main--prefect-orion.netlify.app/api-ref/rest-api/#/Flow%20Runs/read_flow_runs_flow_runs_filter_post

(Yay deploy previews)

zangell44: Good catch! Adding back the type hint is fine as long as we return `ORJSONResponse` directly. Docs looked good for me locally.

madkinsz: Nice

src/prefect/orion/api/flow_runs.py
```diff
@@ -180,6 +181,16 @@ async def read_flow_runs(
...
 
+    # Instead of relying on fastapi.encoders.jsonable_encoder to convert the
+    # response to JSON, we do so more efficiently ourselves.
+    # In particular, the FastAPI encoder is very slow for large, nested objects.
+    # See: https://github.com/tiangolo/fastapi/issues/1224
+    encoded = [
+        schemas.core.FlowRun.from_orm(fr).dict(json_compatible=True)
+        for fr in db_flow_runs
+    ]
+    return ORJSONResponse(content=encoded)
```
chrisguidry, line 192: To @madkinsz point, how about this as a compromise:

```suggestion
    return cast(List[schemas.core.FlowRun], ORJSONResponse(content=encoded))
```

chrisguidry (approved):
🐎 Excellent!

src/prefect/orion/api/flow_runs.py
madkinsz: Zach restored the original type annotation on the function itself. We'll probably get a `mypy` error now but we can tackle that and casting separately when actually start running type checks here?

Issue #6376, opened 2022-08-11
prefect deployment build should not upload data to the block 


felix-ht:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the proposed behavior

Prefect deployment build should not upload data to the storage block. Instead this should be done by apply. 

```bash
prefect deployment build flow/test.py:test -ib s3/some_block
prefect deployment apply hello_world-depolyment.yaml  #updates the flow on prefect an uploads to s3 block
```
The yaml holds all the required information anyhow.

### Describe the current behavior

```bash
prefect deployment build flow/test.py:test -ib s3/some_block # uploads to s3
prefect deployment apply hello_world-depolyment.yaml #updates the flow on prefect
```

This feels very unintuitive - build should not deploy anything. 

With the current state it even feels like there is not even a need for two separate commands. As you always **have** to run apply after build - otherwise you end up with an inconsistent deployment.

bunchesofdonald:
Thanks for the idea @felix-ht! I just wanted to let you know that this feature is available as of 2.3.0, you can now pass `--skip-upload` to the `build` command to have it skip uploading. The `--apply` option was added as well which lets you do  build and apply at the same time. https://docs.prefect.io/concepts/deployments/

Issue #6555, opened 2022-08-24
Add a flag `--apply` to build + apply in one step for ease of use


anna-geller:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the proposed behavior

`-a` for `--apply` - allows to directly apply the manifest, effectively creating deployment with a single command while still retaining the YAML manifest

### Describe the current behavior

build + apply are separate

PR #6560, opened 2022-08-25
Deployment upload flags

cicdw:
This PR makes some adjustments to deployment uploads to remote storage:
- calling `deployment.upload_to_storage` no longer adjusts attributes of the deployment object other than storage
- exposes `skip_upload` flag on `Deployment.build_from_flow` for feature parity with the CLI
- exposes new `upload` flags on both the `deployment.apply` method as well as the `apply` CLI

The `--skip-upload` / `--upload` combo addresses #6376 in a non-breaking way.

### Example

```
prefect deployment build ./flows/example.py:main -n "demo" -sb s3/dev --skip-upload
prefect deployment apply main-deployment.yaml --upload
```

### Checklist

### DIFF
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -255,6 +255,11 @@ async def apply(
         ...,
         help="One or more paths to deployment YAML files.",
     ),
+    upload: bool = typer.Option(
+        False,
+        "--upload",
+        help="A flag that, when provided, uploads this deployment's files to remote storage.",
+    ),
 ):
     """
     Create or update a deployment from a YAML file.
@@ -266,6 +271,21 @@ async def apply(
             app.console.print(f"Successfully loaded {deployment.name!r}", style="green")
         except Exception as exc:
             exit_with_error(f"'{path!s}' did not conform to deployment spec: {exc!r}")
+
+        if upload:
+            file_count = await deployment.upload_to_storage()
+            if file_count and deployment.storage:
+                location = (
+                    deployment.storage.basepath + "/"
+                    if not deployment.storage.basepath.endswith("/")
+                    else ""
+                )
+                if deployment.path:
+                    location += deployment.path
+                app.console.print(
+                    f"Successfully uploaded {file_count} files to {location}",
+                    style="green",
+                )
 
         deployment_id = await deployment.apply()
         app.console.print(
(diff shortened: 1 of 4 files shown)


cicdw:
@anna-geller identified an issue with `apply --build` that I'm currently investigating -- I understand the issue, but the solution is running into deeper async issues that I want to resolve.

anna-geller:
> @anna-geller identified an issue with `apply --build` that I'm currently investigating -- I understand the issue, but the solution is running into deeper async issues that I want to resolve.

Afaik we planned the other way around, though i.e. to add --apply to the build step - or perhaps you mean the same? https://github.com/PrefectHQ/prefect/issues/6555

src/prefect/cli/deployment.py
```diff
@@ -267,6 +272,21 @@ async def apply(
         except Exception as exc:
             exit_with_error(f"'{path!s}' did not conform to deployment spec: {exc!r}")
 
+        if upload:
+            file_count = await deployment.upload_to_storage()
+            if file_count and deployment.storage:
+                location = (
```
bunchesofdonald, line 278: Is the `location` something the `Deployment` class should know rather than building it in the CLI?

tests/test_deployments.py
```diff
@@ -131,12 +132,6 @@ async def test_raises_informatively_if_cant_pull(self):
 
 
 class TestDeploymentUpload:
```
bunchesofdonald, line 134: I'd add some targeted tests for the CLI's `prefect deployment apply --upload ...`.

bunchesofdonald (approved):
I agree with the issue submitter, upload on build is non-intuitive, this seems like a nice way to not break the current behavior and possibly even a way to slowly deprecate upload-on-build if we want.

src/prefect/cli/deployment.py
cicdw: I could expose it as a computed attribute for sure, and I think that would help users discover how entrypoint, path, and storage are all related.

Would you be OK if I did that in a follow-up PR?

cicdw: Actually if I'm going to write some more tests I can do it here :+1:

tests/test_deployments.py
cicdw: we agreed offline to do this in a follow up PR once I have a plan for mocking remote storage uploads

PR #6565, opened 2022-08-25
Add error when sync compatible methods are used incorrectly

madkinsz:
Attempts to resolve a common confusion around `sync_compatible` calls. You cannot call a `sync_compatible` method from a synchronous function that was called by an asynchronous function. For example:

```python
async def foo():
    # Here the main thread is async
    bar()


def bar():
    # Here we are in an async thread but we are in a sync function
    # To support execution of `foobar`, we must be able to send it to an event loop
    # If we send it to the main thread where we are now, it will attempt to run it and wait for
    # it on the same thread resulting in a deadlock
    foobar()


@sync_compatible
async def foobar():
    pass


run_async_in_new_loop(foo)
```

The proper pattern is

```python
async def foo():
    # Here the main thread is async
    await run_sync_in_worker_thread(bar)


def bar():
    # Here we are in a worker thread. `sync_compatible` will send foobar
    # back to the main thread for completion. Since it is using `await` to wait
    # for the result of `bar`, the event loop has yielded and is free to execute
    # `foobar`.
    foobar()


@sync_compatible
async def foobar():
    pass


run_async_in_new_loop(foo)
```

There is likely a small performance hit from peaking at the frame of the caller to see if it is an async function. We attempt to minimize this by retrieving at a single frame instead of the full stack.

Instead of throwing an error, we could support execution of this case by creating a _new_ thread with a _new_ event loop and executing the given coroutine in it. However, this is a dangerous pattern and may cause more issues in the long run. For now, I'd like to continue discouraging patterns that attempt to run `sync_compatible` methods in a blocked event loop.

### Example

The first snippet above would have previously returned a coroutine that could not be awaited. Now, instead, we raise an exception:

> RuntimeError: A 'sync_compatible' method was called from a context that was previously async but is not sync. The sync call must be changed to run in a worker thread to support sending the coroutine for 'foobar' to the main thread.

### Checklist

### DIFF
--- a/src/prefect/utilities/asyncutils.py
+++ b/src/prefect/utilities/asyncutils.py
@@ -3,6 +3,7 @@ Utilities for interoperability with async functions and workers from various con
 """
 import ctypes
 import inspect
+import sys
 import threading
 import warnings
 from contextlib import asynccontextmanager
@@ -181,8 +182,20 @@ def sync_compatible(async_fn: T) -> T:
     @wraps(async_fn)
     def wrapper(*args, **kwargs):
         if in_async_main_thread():
-            # In the main async context; return the coro for them to await
-            return async_fn(*args, **kwargs)
+            caller_frame = sys._getframe(1)
+            caller_async = caller_frame.f_code.co_flags & inspect.CO_COROUTINE
+            if caller_async:
+                # In the main async context; return the coro for them to await
+                return async_fn(*args, **kwargs)
+            else:
+                # In the main thread but call was made from a sync method
+                raise RuntimeError(
+                    "A 'sync_compatible' method was called from a context that was "
+                    "previously async but is now sync. The sync call must be changed "
+                    "to run in a worker thread to support sending the coroutine for "
+                    f"{async_fn.__name__!r} to the main thread."
+                )
+
         elif in_async_worker_thread():
             # In a sync context but we can access the event loop thread; send the async
             # call to the parent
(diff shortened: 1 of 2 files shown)


src/prefect/utilities/asyncutils.py
```diff
@@ -181,8 +182,20 @@ def sync_compatible(async_fn: T) -> T:
...
+            caller_async = caller_frame.f_code.co_flags & inspect.CO_COROUTINE
+            if caller_async:
+                # In the main async context; return the coro for them to await
+                return async_fn(*args, **kwargs)
+            else:
+                # In the main thread but call was made from a sync method
+                raise RuntimeError(
+                    "A 'sync_compatible' method was called from a context that was "
+                    "previously async but is now sync. The sync call must be changed "
+                    "to run in a worker thread to support sending the coroutine for "
```
cicdw, line 195: is there any way to simplify this message in an actionable way?  I believe this message is _most_ helpful for folks who are calling this and don't fully understand sync / async

madkinsz: Yeah I'm not excited about this message either. I think that it should mostly be internal facing — if a user encounters this, we've failed somehow. Even with this message, it seems like an improvement over just receiving a coroutine. Happy for any better message suggestions though.

bunchesofdonald: Maybe write up some documentation on this aspect and have this message point at that documentation?

bunchesofdonald (approved):

src/prefect/utilities/asyncutils.py
madkinsz: That makes sense to me. I'd prefer to do so in a follow-up though as it'll be a significant amount of work to explain how we're doing sync/async compatibility

cicdw (approved):

PR #6566, opened 2022-08-25
Update database docs for postgres

rpeden:
Prefect Orion requires functionality from the pg_trgm extension when using PostgreSQL. This extension is installed and enabled by default in many cases (like when using the Postgres Docker image), but it is not _always_ available. For example, in Postgres packages included in the package repositories of many Linux distros, the extension is not enabled by default; you must install postgres-contrib and then run a Postgres command to add the extension. 

I worry that mentioning this in the docs adds friction, but in my experience it's common for engineers to run Postgres in Docker on their dev machines even though their production environment uses distro-installed Postgres. They figure that it is Postgres on Linux either way, so if it works in dev it should work in production, but this isn't always true. Overall, I _think_ mentioning the extension here is a friction-reducer that will prevent unwelcome surprises.

### Example
N/A - added a sentence to the docs. 

### Checklist

### DIFF
--- a/docs/concepts/database.md
+++ b/docs/concepts/database.md
@@ -23,7 +23,7 @@ The Prefect Orion database persists data used by many features of Prefect to per
 Currently Prefect Orion supports the following databases:
 
 - SQLite: The default in Prefect Orion, and our recommendation for lightweight, single-server deployments. SQLite requires essentially no setup.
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
Thanks!

tpdorsey (approved):

PR #6571, opened 2022-08-25
Always provide an idempotency key to avoid duplicate flow runs

madkinsz:
When the client retries a request (i.e. there is a network issue), the flow run can be created multiple times on the server.

This updates the client to _always_ set an idempotency key for `create_flow_run`.  This key defaults to a UUID4.

### Example

Previously when creating a run timed out, the client would retry 6 times which resulted in 6 duplicate flow runs.
Now, a single flow run will be created.

I've got a massive flow at https://staging.prefect.io/prefect-staging/flow/485d0f11-718e-4276-876b-b9c3ce4ce3eb that is so slow to create that the client can timeout waiting for the server to respond that I noticed this issue for originally.

### Checklist


zangell44 (approved):

cicdw (approved):
this seems reasonable to me

PR #6572, opened 2022-08-25
Fix `FlowRunView` task run queries when all task runs are cached

madkinsz:
Closes bug reported in https://prefecthq.slack.com/archives/C028BPL85C3/p1661452093250219 where if all of the task runs are cached an error is thrown when retrieving cache runs

```python
  File "/usr/local/lib/python3.9/site-packages/prefect/engine/task_runner.py", line 880, in get_task_run_state
    value = prefect.utilities.executors.run_task_with_timeout(
  File "/usr/local/lib/python3.9/site-packages/prefect/utilities/executors.py", line 468, in run_task_with_timeout
    return task.run(*args, **kwargs)  # type: ignore
  File "/redacted.py", line 122, in redacted
    tasks = view.get_all_task_runs()
  File "/usr/local/lib/python3.9/site-packages/prefect/backend/flow_run.py", line 739, in get_all_task_runs
    task_run_data = TaskRunView._query_for_task_runs(
  File "/usr/local/lib/python3.9/site-packages/prefect/backend/task_run.py", line 367, in _query_for_task_runs
    raise ValueError(
ValueError: No task runs found while querying for task runs where ....
```

This is easily fixed by toggling the `error_on_empty` bool.
Includes a change that excludes of cached task runs from the `load_static_tasks` query.
Includes a fix that prevents duplicate task runs from being returned when the object is shared by multiple threads.

### Example

```python
from prefect.backend import FlowRunView

view = FlowRunView.from_flow_run_id("030ee8ba-3e2c-439f-96c4-44c0ef2ad618")

# Calling repeatedly does not fail
view.get_all_task_runs()
view.get_all_task_runs()  # Previously would fail here
view.get_all_task_runs()

# Getting the latest is fine
view.get_latest()

# Getting the latest and making a query for static tasks is fine
view.get_latest(load_static_tasks=True)

# Instantiating a new instance with static loading works still
view = FlowRunView.from_flow_run_id("030ee8ba-3e2c-439f-96c4-44c0ef2ad618", load_static_tasks=True)
```
### Checklist

### DIFF
--- a/src/prefect/backend/flow_run.py
+++ b/src/prefect/backend/flow_run.py
@@ -738,11 +738,16 @@ class FlowRunView:
                 "Please load the tasks you are interested in individually."
             )
 
+        # Freeze the cached task runs we will return to prevent race conditions if
+        # this object is retrieving task runs in multiple threads
+        cached_task_runs = list(self._cached_task_runs.values())
+        cached_task_run_ids = {task_run.task_run_id for task_run in cached_task_runs}
+
         # Run a single query instead of querying for each task run separately
         task_run_data = TaskRunView._query_for_task_runs(
             where={
                 "flow_run_id": {"_eq": self.flow_run_id},
-                "id": {"_nin": list(self._cached_task_runs.keys())},
+                "id": {"_nin": list(cached_task_run_ids)},
             },
             error_on_empty=False,
         )
@@ -751,7 +756,7 @@ class FlowRunView:
             TaskRunView._from_task_run_data(data) for data in task_run_data
         ]
 
-        task_runs = new_task_runs + list(self._cached_task_runs.values())
+        task_runs = new_task_runs + cached_task_runs
 
         for task_run in new_task_runs:
             self._cache_task_run_if_finished(task_run)
--- a/src/prefect/backend/flow_run.py
+++ b/src/prefect/backend/flow_run.py
@@ -583,13 +583,16 @@ class FlowRunView:
             A populated `FlowRunView` instance
         """
         flow_run_data = cls._query_for_flow_run(where={"id": {"_eq": flow_run_id}})
+        _cached_task_runs = list(_cached_task_runs or [])
 
         if load_static_tasks:
             task_run_data = TaskRunView._query_for_task_runs(
                 where={
                     "map_index": {"_eq": -1},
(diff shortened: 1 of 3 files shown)


src/prefect/backend/flow_run.py
```diff
@@ -583,13 +583,16 @@ def from_flow_run_id(
...
         """
         flow_run_data = cls._query_for_flow_run(where={"id": {"_eq": flow_run_id}})
+        _cached_task_runs = list(_cached_task_runs or [])
 
         if load_static_tasks:
             task_run_data = TaskRunView._query_for_task_runs(
                 where={
                     "map_index": {"_eq": -1},
                     "flow_run_id": {"_eq": flow_run_id},
+                    "id": {"_nin": [view.task_run_id for view in _cached_task_runs]},
```
bunchesofdonald, line 593: Forgive my v1 ignorance, but I assume that `_nin` means `not in`? So this is effectively saying "Get all of the task runs associated with this flow run, except for the ones that are cached."

bunchesofdonald (approved):
Assuming my assumptions are right this looks good to me.

src/prefect/backend/flow_run.py
madkinsz: Yep! We only cache finished task runs so excluding them from the query results will reduce the amount of traffic with no change in behavior.

src/prefect/backend/flow_run.py
```diff
@@ -735,19 +738,25 @@ def get_all_task_runs(self) -> List["TaskRunView"]:
                 "Please load the tasks you are interested in individually."
             )
 
+        # Freeze the cached task runs we will return to prevent race conditions if
+        # this object is retrieving task runs in multiple threads
+        cached_task_runs = list(self._cached_task_runs.values())
+        cached_task_run_ids = {task_run.task_run_id for task_run in cached_task_runs}
+
```
madkinsz, line 745: @bunchesofdonald A new change here that prevents a race condition if `self._cached_task_runs` is updated by another thread while the query is running. We wouldn't include those task runs in the `nin` call so they'd be returned by the API then on L759 we'd append the cached runs resulting in duplicates.

zangell44 (approved):

PR #6575, opened 2022-08-26
Add a context manager to disable run logger

ahuang11:
Closes https://github.com/PrefectHQ/prefect/issues/6576

This PR makes it possible to bypass `RuntimeError("There is no active flow or task run context.")` when a task contains a call to `get_run_logger` and its underlying function gets called outside a flow/task.

### Example

If I don't care about the logs and I just want the functionality of a task by calling `fn`, this will raise the aforementioned error:
```
from prefect import flow, task, get_run_logger

@task
def test_disable_run_logger():
    logger = get_run_logger()
    logger.info("hey I shouldn't be showing up")
    return 42

test_disable_run_logger.fn()
```

With this PR, it's possible to bypass that error and get the return value of 42.
```
from prefect import flow, task, get_run_logger
from prefect.logging import disable_run_logger

@task
def test_disable_run_logger():
    logger = get_run_logger()
    logger.info("hey I shouldn't be showing up")
    return 42

with disable_run_logger():
   test_disable_run_logger.fn()
```

Or if users don't want to use this contextmanager, they can still bypass it with:
```
    flow_run_logger = get_logger("prefect.flow_run")
    task_run_logger = get_logger("prefect.task_run")
    flow_run_logger.disabled = True
    task_run_logger.disabled = True
```
This is because there's now an additional elif clause to check whether both loggers are disabled.

### Context 
Anna mentioned that this came up a lot; e.g. https://discourse.prefect.io/t/unit-testing-best-practices-for-prefect-flows-subflows-and-tasks/1070/7

I personally encountered this when I was trying to [reuse functionality from an existing task](https://github.com/PrefectHQ/prefect-gcp/pull/41/files#diff-714c7d2c23ae895d8069740bb275473f8c7945952a02657a0297ac5ef7db3e76R611) in prefect-gcp, but since it's a class method, and the [underlying function of the task contained `get_run_logger`](https://github.com/PrefectHQ/prefect-gcp/pull/41/files#diff-714c7d2c23ae895d8069740bb275473f8c7945952a02657a0297ac5ef7db3e76R268-R269), it failed.

### Checklist


anna-geller (approved):
this is awesome, works perfecty!

src/prefect/logging/loggers.py
```diff
@@ -92,6 +93,12 @@ def get_run_logger(context: "RunContext" = None, **kwargs: str) -> logging.Logge
         logger = flow_run_logger(
             flow_run=flow_run_context.flow_run, flow=flow_run_context.flow, **kwargs
         )
+    elif (
+        get_logger("prefect.flow_run").disabled
+        and get_logger("prefect.task_run").disabled
+    ):
+        logger = logging.getLogger("null")
+        logger.addHandler(logging.NullHandler())
```
madkinsz, line 101: This always retrieves the same logger so this will add a new handler each time this is called

ahuang11: Do you think I should edit logging.yml or just set a NULL_LOGGER in code?

madkinsz: Is there a reason we need to set a null handler on it? Can't we just return a logger that's not hooked up to anything?

madkinsz (approved):

PR #6578, opened 2022-08-26
Use Github Action's cache for image layer caches.

chrisguidry:
This should improve the runtime of our release process by using cached image
layers between runs.  Note, though, that the biggest benefits here come from
optimizing the order of the Dockerfile to maximize layer reuse.  This is
something we can work on over time.

Reference for the cache directives: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#github-cache

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -97,6 +97,8 @@ jobs:
           tags: ${{ steps.metadata.outputs.tags }}
           labels: ${{ steps.metadata.outputs.labels }}
           outputs: type=docker,dest=/tmp/image.tar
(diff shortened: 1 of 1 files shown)


bunchesofdonald (approved):
Makes sense to me.

chrisguidry:
Unfortunately, there's not a great way to test this until the next release, but I have done this with Prefect Cloud and the results are good.

madkinsz:
@chrisguidry We should be able to test this for our image builds during service tests?

See also https://github.com/PrefectHQ/orion/pull/811

PR #6582, opened 2022-08-26
Update iterable detection to exclude strings and bytes

madkinsz:
Updates our `isiterable` check to treat `str` and `bytes` as non-iterable types. Our usage of `isiterable` is to detect if a parameter should be mapped over or not. These objects are singletons that happen to be iterable and should not require an explicit `unmapped` during mapped calls. Passing a string to a mapped call and expecting it to be treated as a constant is a very common pattern.

I've also updated the `Task.map` documentation to clarify this while simplifying and fixing some examples.

I noticed this in https://github.com/PrefectHQ/prefect/pull/6579.

### Example

```python
@task
def display(prefix, item):
   print(prefix, item)

@flow
def my_flow():
    # Previously, this would throw a mapping length mismatch error
    display.map("Check it out: ", [1, 2, 3])

my_flow()
```

### Checklist

### DIFF
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -638,17 +638,19 @@ class Task(Generic[P, R]):
             >>> my_flow()
             [6, 7, 8]
 
-            Use unmapped to treat an iterable arg as static
+            Use `unmapped` to treat an iterable argument as static
             >>> from prefect import unmapped
             >>>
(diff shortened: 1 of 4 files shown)


bunchesofdonald (approved):
This is definitely more intuitive.

PR #6583, opened 2022-08-26
Including the layer caching configuration from #6578 during python tests.

chrisguidry:
Pretty much what is says on the tin ;)


madkinsz (approved):

chrisguidry:
Before:
image

(still working on "after")

chrisguidry:
image

A modest saving, maybe a minute from the main image build.  This will get better with more fine-tuning of how Python packages are installed.

PR #6594, opened 2022-08-27
work-queue playwright tests

stackoverfloweth:
added `useToggle` utility (MUCH simpler with non headless version of toggle)
added the following tests for work-queues

- Can create work queue
- Can toggle workQueue from list
- Can delete worksQueue from list
- Can toggle workQueue


orion-ui/tests/utilities/useToggle.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function useToggle(page: Page | Locator = PAGE): UseToggle {
+  return {
+    toggle: getToggle(page),
+  }
+}
+
+function getToggle(page: Page | Locator): Locator {
+  return page.locator('.p-toggle__control')
+}
```
pleek91, line 16:
```suggestion
export function useToggle(page: Page | Locator = PAGE): UseToggle {
  const toggle = page.locator('.p-toggle__control')

  return {
    toggle,
  }
}
```

orion-ui/tests/workQueues.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  const newWorksQueues = await worksQueues.count()
+
+  expect(newWorksQueues).toBe(existingWorksQueues - 1)
+})
+
+test('Can toggle workQueue', async ({ page }) => {
+  await createWorkQueue()
+  await page.waitForNavigation({
+    url: /\/work-queue\//,
+  })
```
pleek91, line 67: We should make a utility for waiting for our known pages.

orion-ui/tests/workQueues.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const { toggle } = useToggle()
+  await toggle.waitFor()
```
pleek91, line 70: Does this work without the wait for? I thought the isChecked would wait on its own

orion-ui/tests/workQueues.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  const currentState = await toggle.isChecked()
+  await toggle.setChecked(!currentState)
+  const newState = await toggle.isChecked()
+
+  expect(newState).toBe(!currentState)
+})
+
+async function createWorkQueue(): Promise<void> {
```
pleek91, line 78: Can this take an argument for the name rather than using the same const each time? That way the places that use it can pass in a name. That would make this more friendly for future tests as well.

orion-ui/tests/workQueues.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  const { heading } = usePageHeading()
+  const { link } = useLink(pages.workQueuesCreate, heading)
+  await link.click()
+
+  const { control: name } = useLabel('Name')
+  const input = name.locator('input')
```
pleek91, line 84: we need a `useInput` utility

orion-ui/tests/workQueues.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  const newWorkQueues = await workQueues.count()
+
+  expect(newWorkQueues).toBe(existingWorkQueues + 1)
+})
+
+test('Can toggle workQueue from list', async () => {
+  const { rows: worksQueues } = useTable()
+  const worksQueue = worksQueues.filter({
+    hasText: workQueueNameToCreate,
+  })
```
pleek91, line 32: If you accept name in your createWorkQueue function. Could we have each test create its own work queue and run these in parallel?

Would have to tweak how we're testing that something is created. We'd have to locate a table row that contains the unique name rather than checking the count is 1 greater. But I think that's better right?

orion-ui/tests/workQueues.spec.ts
pleek91: Even better. Have it mock the name and return it

pleek91 (commented):
I love these tests. Couple ideas but overall looking great. Curious what your experience was writing these. We're the existing utilities and tests intuitive? How was the developer experience?

orion-ui/tests/workQueues.spec.ts
stackoverfloweth: this relates to a bigger conversation below

orion-ui/tests/workQueues.spec.ts
stackoverfloweth: it does not, this `waitFor` is just the standard playwright version though, similar to how we have `waitFor` tables, since the wait for navigation doesn't wait for the work queue subscription to finish and the toggle actually become visible

orion-ui/tests/workQueues.spec.ts
stackoverfloweth: perhaps, this particular example is not a full `page`, since the actual url is `/work-queue/:id` and I don't have the id.

stackoverfloweth:
@pleek91 

It might be best to sync up on these tests, here are some thoughts in the mean time.

 - I'm increasingly confident that running the full test suite is going to take a long time, even just notifications and work-queues takes a few seconds for < 20 tests.
 - these tests did take me a long time to write, but part of that is my lack of practice and also the lack of utilities. I wasted a bunch of time on the complicated p-toggle component, which once I rewrote become stupid simple with built in playwright. Like most tests, if the test is hard to write its a good indicator that _what_ you're testing need work. The more we can just use vanilla playwright the better.
 - I really want to define these tests as individual units that run in parallel (true for notifications too). I don't like the idea if that if the first test fails they all "fail". Unfortunately I feel even worse about using playwright logic (find element, click element) to create for each test (would need to do the same for delete to clean up after tests). That puts the same dependency of if create test fails - they all fail,  except now it takes longer to run which defeats the purpose. What I really want is to use the API class to create/delete but that's proving difficult.
 
 here's an example of what I'm thinking for that
```ts
async function useWorkQueue(callback: (workQueue: WorkQueue) => Promise<void>): Promise<void> {
  const workQueue = await workQueuesApi.createWorkQueue({
    name: mocker.create('string'),
  })

  await callback(workQueue)

  workQueuesApi.deleteWorkQueue(workQueue.id)
}
```

which would be used like so

```ts
test('Can toggle workQueue', ({ page }) => {
  useWorkQueue(async (workQueue) => {
    const url = pages.workQueue(workQueue.id)
    await page.waitForNavigation({ url })

    const { toggle } = useToggle()
    await toggle.waitFor()
    const currentState = await toggle.isChecked()
    await toggle.setChecked(!currentState)
    const newState = await toggle.isChecked()

    expect(newState).toBe(!currentState)
  })
})
```

stackoverfloweth:
@pleek91 take a look at this last commit, I was able to update the tests to run in parallel with their own setup/tear down using the api directly 

https://github.com/PrefectHQ/prefect/pull/6594/commits/a0767b1613ede143ee89892401baf8f695f4c88d

orion-ui/tests/workQueues.spec.ts
pleek91: Right, but we could make a utility for waiting for our pages. Not a specific work queue, but the "single work queue" page

orion-ui/tests/workQueues.spec.ts
pleek91: I guess it's similar to count and how count will return 0 if there's nothing on the page. I expected `isChecked` to wait for the locator though 🤔

pleek91:
@stackoverfloweth Latest changes are a great step in the right direction. Gave me so many ideas. Let's sync up next week 👍

orion-ui/tests/utilities/fixtures.ts
```diff
@@ -1,10 +1,48 @@
...
+      try {
+        await deleteWorkQueue(request, workQueue.id)
+      } catch {
+        // do nothing
+      }
+    }
+  },
+})
+
+const baseUrl = 'http://127.0.0.1:4200/api'
```
pleek91, line 37: This probably should be customizable via .env

pleek91 (approved):

PR #6595, opened 2022-08-27
Move logic for getting work queue runs to DB

jlowin:
Presently, all work queue logic is contained in `orion.models.work_queues.get_runs_in_work_queue`, including handling `paused` queues and queues with `concurrency limits`. This route is quite performant for a single queue but can't efficiently scale beyond that. As I'm looking at broad services that monitor multiple queues, this becomes a bottleneck.  

In the worst case, `get_runs_in_work_queue` makes three separate database calls: one to load the work queue, a second to count how many flow runs are currently running (if the work queue has a concurrency limit) and a third to load scheduled flow runs. All of these queries are well-scoped and quite performant, but three separate network calls can have a latency consequence. This explodes when working with multiple queues: checking two queues requires up to six calls, three queues up to nine, etc.

In this PR, I'm seeking to make a single db query for an arbitrary number of work queues. This moves the work queue logic out of the single Python function and into a scalable db process. I've implemented it as a `QueryComponent` method so aspects of it can be overwritten in dialect-specific ways; specifically, it takes advantage of Postgres lateral joins when available (SQLite has to do a slower correlated subquery). 

For a single queue, the primary advantage of this approach is reducing network calls, as the single-queue queries are (necessarily) already pretty efficient. To ensure that I was not introducing any significant regressions by moving to a broad SQL query, I ran benchmarks against a local database, which minimizes the impact of network calls and highlights differences in query performance itself. For a single, unlimited work queue -- the best case scenario for the old function -- the new function is roughly the same to slightly (~5%) slower. For a single, concurrency-limited work queue, the new function is roughly the same to slightly (~10%) faster. For multiple work queues it is WAY faster - this query can load all scheduled runs from all Prefect Cloud work queues (and there are thousands of such queues) in about 10x the time it takes the current function to load one queue. These benchmarks give me confidence in the approach; network latency will only make the new function more preferable.


src/prefect/orion/database/query_components.py
```diff
@@ -437,3 +553,53 @@ async def get_flow_run_notifications_from_queue(
...
+        limit_per_queue: Optional[int],
+        scheduled_before: Optional[datetime.datetime],
+    ):
+
+        # select scheduled flow runs, ordered by scheduled start time per queue
+        scheduled_flow_runs = (
+            sa.select(
+                sa.func.row_number()
+                .over(
+                    partition_by=[db.FlowRun.work_queue_name],
```
chrisguidry, line 570: I double-checked that `WorkQueue.name` is unique (`uq_work_queue__name`) 👍

src/prefect/orion/database/query_components.py
```diff
@@ -437,3 +553,53 @@ async def get_flow_run_notifications_from_queue(
...
+
+        join_criteria = sa.and_(
+            self._join_flow_run_to_work_queue(
+                flow_run=scheduled_flow_runs.c, work_queue=db.WorkQueue
+            ),
+            scheduled_flow_runs.c.rank
+            <= sa.func.min(
+                sa.func.coalesce(work_queue_query.c.available_slots, limit_per_queue),
+                limit_per_queue,
+            ),
```
chrisguidry, line 603: 👏 This is brilliant

chrisguidry (approved):
This looks fundamentally sound to me, even if there could be some edge cases in the query I'm not seeing.  It looks like you anticipated a couple of `NULL`-related things, and I can't think of any more.  Such a cool approach.

I think these queries would be far more legible if we expressed them as SQL (rather than contorting ourselves to say them with SQLAlchemy), but I that could just be a me thing.

jlowin:
@chrisguidry yeah, the rendered SQL is definitely "more readable" than the SQLA representation, but this translates more easily across postgres and sqlite. I just made some minor changes to try to improve readability slightly.

src/prefect/orion/models/work_queues.py
```diff
@@ -165,15 +165,64 @@ async def delete_work_queue(
...
+            session=session,
+            work_queue_id=work_queue_id,
+            db=db,
+            scheduled_before=scheduled_before,
+            limit=limit,
+        )
+
+
+@inject_db
+async def _legacy_get_runs_in_work_queue(
```
zangell44, line 216: can we modify this function to only handle legacy work queues?
at the moment it handles both old and new work queues

zangell44 (approved):

madkinsz:
@jlowin is there anything holding this one up?

jlowin:
Nope!

PR #6598, opened 2022-08-28
GitHub-based deployments

### LINKED ISSUE #6585: Integrate GitHub read-only storage block with the deployment interface
### First check

### Prefect Version

2.x

Related to https://github.com/PrefectHQ/prefect/pull/6517 

follow up to the previous PR that introduced the GitHub block

### Example Use

_No response_

### Additional context

_No response_

cicdw:
This PR wires up `GitHub` remote storage so that it can be used with deployments seamlessly.   Specifically, users can now specify:
- a remote repository (currently required to be public, but we will expand support for private repos)
- a reference branch or tag of that repository
- and an optional subdirectory of that repository to use as a workflow deployment project structure

Closes #6585 

### Example
We can actually illustrate this rather nicely off the current branch `github-deployments` that this PR is based on! 

Start by navigating to `tests/test-projects/nested-project` ([see its file structure here](https://github.com/PrefectHQ/prefect/tree/main/tests/test-projects/nested-project)).  We begin by creating a `GitHub` remote storage block:
```python
from prefect.filesystems import GitHub

g = GitHub(repository="https://github.com/PrefectHQ/prefect.git", reference="github-deployments")
g.save("example")
```
and then we proceed to create a corresponding deployment from the CLI for one of our test flows:
```bash
prefect deployment build ./implicit_relative.py:foobar \
 -n "GitHub Example" \
 -sb github/example \
 --path tests/test-projects/nested-project \
 --apply

prefect deployment run 'foobar/GitHub Example'

# this next command can be run from within *any* directory
prefect agent start -q default
...
17:53:46.052 | INFO    | Flow run 'intrepid-newt' - Finished in state Completed()
```

### Checklist

### DIFF
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -283,6 +283,11 @@ async def apply(
                         f"Successfully uploaded {file_count} files to {deployment.location}",
                         style="green",
                     )
+            else:
+                app.console.print(
+                    f"Deployment storage {deployment.storage} does not have upload capabilities; no files uploaded.",
+                    style="red",
+                )
 
         deployment_id = await deployment.apply()
         app.console.print(
@@ -580,6 +585,11 @@ async def build(
                     f"Successfully uploaded {file_count} files to {deployment.location}",
                     style="green",
                 )
+        else:
+            app.console.print(
+                f"Deployment storage {deployment.storage} does not have upload capabilities; no files uploaded.  Pass --skip-upload to suppress this warning.",
+                style="green",
+            )
 
     deployment_loc = output_file or f"{obj_name}-deployment.yaml"
     await deployment.to_yaml(deployment_loc)
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -273,12 +273,16 @@ async def apply(
             exit_with_error(f"'{path!s}' did not conform to deployment spec: {exc!r}")
 
         if upload:
-            file_count = await deployment.upload_to_storage()
-            if file_count:
-                app.console.print(
-                    f"Successfully uploaded {file_count} files to {deployment.location}",
-                    style="green",
-                )
+            if (
+                deployment.storage
+                and "put-directory" in deployment.storage.get_block_capabilities()
+            ):
+                file_count = await deployment.upload_to_storage()
+                if file_count:
+                    app.console.print(
+                        f"Successfully uploaded {file_count} files to {deployment.location}",
+                        style="green",
+                    )
 
         deployment_id = await deployment.apply()
         app.console.print(
@@ -548,12 +552,16 @@ async def build(
         )
 
     if not skip_upload:
-        file_count = await deployment.upload_to_storage()
-        if file_count:
-            app.console.print(
-                f"Successfully uploaded {file_count} files to {deployment.location}",
-                style="green",
-            )
+        if (
+            deployment.storage
+            and "put-directory" in deployment.storage.get_block_capabilities()
+        ):
+            file_count = await deployment.upload_to_storage()
+            if file_count:
+                app.console.print(
+                    f"Successfully uploaded {file_count} files to {deployment.location}",
+                    style="green",
+                )
 
     deployment_loc = output_file or f"{obj_name}-deployment.yaml"
     await deployment.to_yaml(deployment_loc)
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -394,7 +394,7 @@ async def build(
         None,
         "--storage-block",
         "-sb",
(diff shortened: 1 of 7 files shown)


src/prefect/filesystems.py
```diff
@@ -750,23 +751,38 @@ async def get_directory(
...
 
         if local_path is None:
             local_path = Path(".").absolute()
 
+        # in this case, we clone to a temporary directory and move the subdirectory over
+        tmp_dir = None
+        if from_path:
+            tmp_dir = tempfile.TemporaryDirectory(suffix="prefect")
+            path_to_move = str(Path(tmp_dir.name).joinpath(from_path))
+            cmd += f" {tmp_dir.name} && cp -R {path_to_move}/."
```
cicdw, line 774: I recognize this is a potentially OS specific command to run; we can probably refactor this to rely on `shutil` but in the meantime this works on *nix machines.

src/prefect/filesystems.py
```diff
@@ -750,23 +751,38 @@ async def get_directory(
         """
         Clones a GitHub project specified in `from_path` to the provided `local_path`; defaults to cloning
         the repository reference configured on the Block to the present working directory.
+
+        Args:
+            - from_path: if provided, interpreted as a subdirectory of the underlying repository that will
```
cicdw, line 756: Testing the `from_path` argument is tricky without cloning a repo with every test run; will need to think about this more.

anna-geller (approved):
Works great with CLI! 

image

src/prefect/cli/deployment.py
```diff
@@ -273,12 +273,16 @@ async def apply(
...
         if upload:
-            file_count = await deployment.upload_to_storage()
-            if file_count:
-                app.console.print(
-                    f"Successfully uploaded {file_count} files to {deployment.location}",
-                    style="green",
-                )
+            if (
+                deployment.storage
+                and "put-directory" in deployment.storage.get_block_capabilities()
```
chrisguidry, line 278: 👏 I love how the capabilities worked out from the calling side.  We may want to add a `Block.has_capability(self, capability: str) -> bool` method to make these read even more clearly, but it's already pretty good.

chrisguidry: Also, you didn't just enable Github deployments, you enabled using any read-only storage block :bow:

src/prefect/filesystems.py
chrisguidry: In less time than it would take to clone a reference repo, we could `git init` and `git commit` what we wanted into a `tmp_path`.  Should be pretty easy with pytest fixtures:


```
from functools import partial
from pathlib import Path
from subprocess import check_call

@pytest.fixture
def repo(tmp_path: Path) -> Path:
    in_repo = partial(check_call, cwd=tmp_path)

    in_repo(['git', 'init', '-b', 'main'])

    (tmp_path / "my_flow.py").write_text(...some flow code...)

    in_repo(['git', 'add', '.'])
    in_repo(['git', 'commit', '--message', 'initial commit')

    return tmp_path

@pytest.fixture
def prod_branch(repo: Path) -> Path:
    in_repo = partial(check_call, cwd=tmp_path)

    in_repo(['git', 'checkout', '-b', 'prod'])

    return repo
```

chrisguidry (approved):
Let's go!

src/prefect/cli/deployment.py
```diff
@@ -273,12 +273,16 @@ async def apply(
...
+            ):
+                file_count = await deployment.upload_to_storage()
+                if file_count:
+                    app.console.print(
+                        f"Successfully uploaded {file_count} files to {deployment.location}",
+                        style="green",
+                    )
```
bunchesofdonald, line 285: Should we add an `else` here that catches the case where a user is using a read-only storage block, but they've said to upload? Without it this seems like a silent failure.

src/prefect/cli/deployment.py
```diff
@@ -566,12 +570,16 @@ async def build(
...
-        file_count = await deployment.upload_to_storage()
-        if file_count:
-            app.console.print(
-                f"Successfully uploaded {file_count} files to {deployment.location}",
-                style="green",
-            )
+        if (
+            deployment.storage
+            and "put-directory" in deployment.storage.get_block_capabilities()
+        ):
```
bunchesofdonald, line 576: This side also feels like it needs at least a message saying 'skipping upload since the block is read-only', but it feels wrong to fail the command (possibly more reason to deprecate this flag).

src/prefect/cli/deployment.py
bunchesofdonald: Agreed, this `x in deployment.storage.get_block_capabilities()` pattern is common enough at this point that it probably warrants a helper.

bunchesofdonald (approved):

src/prefect/cli/deployment.py
cicdw: good call, updated

src/prefect/cli/deployment.py
cicdw: updated

PR #6602, opened 2022-08-29
Fix edge case in `sync_compatible` detection when using AnyIO task groups

madkinsz:
The recently added check (https://github.com/PrefectHQ/prefect/pull/6565) for an async caller will fail when using task groups since AnyIO uses internal methods that are not async to create the coroutine

### Example

```python
@sync_compatible
async def foo(task_status=None):
   if task_status:
       task_status.started()

    # We would throw an error on this call before
    await tg.start(foo)

foo()
```

### Checklist


bunchesofdonald (approved):

bunchesofdonald:
Good catch!

Issue #6461, opened 2022-08-18
Can't run deployment on MacOS


semcha:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Can't run deployment on MacOS

### Reproduction

```python
pip install -U "prefect>=2.1.0" prefect-shell prefect-dask prefect-dbt
prefect orion database reset
prefect orion start
# Create Work Queue in UI named dev
prefect agent start --work-queue "dev"

prefect deployment build dbt-test.py:dbt_flow -n main__dbt-test -q dev
prefect deployment apply dbt_flow-deployment.yaml
```

dbt-test.py:
```python
from prefect import flow, task, get_run_logger
from prefect_dbt.cli.commands import trigger_dbt_cli_command


@flow
def dbt_flow():
    logger = get_run_logger()
    result = trigger_dbt_cli_command(
        "dbt debug",
        project_dir="/some_empty_dbt_project/",
        return_all=True,
    )
    logger.info(result)
    return result
```

### Error

```

10:49:49.848 | INFO    | prefect.agent - Submitting flow run '18ed4ecd-2002-414b-ab73-8fb2637fbe90'
10:49:49.908 | INFO    | prefect.infrastructure.process - Opening process 'honest-warthog'...
10:49:49.914 | ERROR   | prefect.agent - Flow runner failed to submit flow run '18ed4ecd-2002-414b-ab73-8fb2637fbe90'
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/prefect/agent.py", line 199, in submit_run
    await self.task_group.start(submit_flow_run, flow_run, infrastructure)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 807, in start
    return await future
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/prefect/infrastructure/submission.py", line 48, in submit_flow_run
    return await infrastructure.run(task_status=task_status)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/prefect/infrastructure/process.py", line 66, in run
    process = await run_process(
[...]

PR #6610, opened 2022-08-29
Update default infrastructure command to be set at runtime

madkinsz:
Originally, this command was set as a default on infrastructure so that when a deployment is created the command is frozen. This is helpful for containerized deployments: if we ever change the entrypoint command, the agent needs to send the correct command to the container. In the frozen implementation, we do not need to worry about any version tracking to achieve this. However, there are a few issues introduced by this:

- If the infrastructure is a subprocess and the entrypoint changes, the agent would use the wrong command
- Maintaining the `prefect.engine` entrypoint in the future is very low cost
- Setting the command to call `python` statically can cause flow runs to run outside the agent's virtual environment
- Infrastructure blocks can be used for running arbitrary commands, not just flow runs, but the default command was flow run specific
 
Since every other flow run related property is set when a flow run is submitted to infrastructure by an agent, the inclusion of this on the infrastructure itself is confusing. This pull request:

- Restores the type of `command` to `Optional[List[str]]`
- Eliminates language around flow run commands for infrastructure fields
- Uses `sys.executable` for processes to ensure agents submit flows to the correct Python version and environment
- Retains the ability to override the command
- Clarifies any confusion about whether or not the command has been provided by the user or should be the default
- Includes some minor changes to the descriptions of other base infrastructure fields
- Refactors infrastructure preparation to be a property of infrastructure
- Refactors the agent to prepare infrastructure for the flow run and call `run` directly

The refactor to move submission logic to the base infrastructure type was motivated by concerns raised by @cicdw where we needed to perform conditional logic based on infrastructure type to determine the flow run entrypoint command to use. This refactor removes the `submission` module entirely, adding a `prepare_for_flow_run` method to `Infrastructure` and flow run submission utilities as static methods on the infrastructure base class. 

Closes https://github.com/PrefectHQ/prefect/issues/6461

Remaining work:
- [x] Update tests to account for change
- [x] Add checks for null commands at infrastructure runtime
- [ ] Ensure this is backwards compatible 

### Checklist


madkinsz:
Note that @cicdw and I chatted briefly about this and hope to think about the interaction between infrastructure, agents, and flow runs holistically. An alternative design is likely to improve clarity in the long run. For now, I think that consolidating the override behavior to submission by the agent simplifies this pattern.

madkinsz:
One option for a more holistic approach is to allow `command` to be passed to `run`. The `command` field would then just be the _default_ command for the infrastructure.

bunchesofdonald (approved):
This makes sense to me. It seems best to have that base block agnostic about the eventual command that's going to run. I _do_ like the sound of `run` also taking a command, but that could just be my docker fanboy kicking in.

src/prefect/infrastructure/submission.py
```diff
@@ -1,14 +1,23 @@
...
 from prefect.orion.schemas.core import FlowRun
 
 MIN_COMPAT_PREFECT_VERSION = "2.0b12"
 
 
+def get_flow_run_entrypoint(infrastructure):
+    if isinstance(infrastructure, Process):
+        return [sys.executable, "-m", "prefect.engine"]
+    else:
+        return ["python", "-m", "prefect.engine"]
```
cicdw, line 18: The design of this feels off to me - the location of this command specification is so far removed from where a developer will look when debugging

cicdw: Sad differently, it's far too implicit

madkinsz: This single module deals with all of the overrides for running flows in infrastructure. I'm hoping that consolidation here makes it easier to find when debugging in contrast to v1 where you need to look through the implementation details of each agent. I'm not excited about the specific toggle for the Python call in process infrastructure, but I'm not sure where else to implement it. Without the `sys.executable` switch we're too likely to call into the wrong `python` for subprocesses.

madkinsz: Are you okay with improving the design in the future? Which user stories are you worried this degrades?

madkinsz: This was addressed with a refactor agreed upon in a team review session 🥳

cicdw (approved):
Nice, this looks clean!

cicdw (approved):

PR #6626, opened 2022-08-30
Minor touchups to release notes for v2.3.0

madkinsz:
We should probably not merge the release notes until right before release as it turns out I like to futz.


bunchesofdonald (approved):

discdiver:
I pushed two other minor changes to the release notes.

madkinsz:
Thanks!

PR #6633, opened 2022-08-30
Swap order of "current" and "proposed" behavior in enhancement issue template

madkinsz:
Every time I open an enhancement request I want to explain how things currently work and what's wrong with that before proposing a solution. If users do not have a clear story of the current behavior, a feature request does not make much sense. Currently, our issue template prompts for your proposed change _then_ prompts for current behavior. This pull request swaps the order of these fields. This should make issues easier to read and write.

### Checklist


anna-geller (approved):
makes so much sense!

PR #6637, opened 2022-08-31
Allow the DB interface class to be modified

jlowin:
Currently, the database permits dependency injection in a variety of ways -- the ORM config, customizable queries, and the database's own configuration. One thing that can't be changed is the parent interface itself, which is always an `OrionDBInterface`.  

The `OrionDBInterface` re-exposes many of its component attributes as properties or methods, which provides convenient access (e.g. `db.json_arr_agg` instead of `db.queries.json_arr_agg`, a more difficult-to-discover access point. However, this means that if dependency injection is used to add a new query, it can not be accessed in the same way. By allowing the `OrionDBInterface` class itself to be injected, developers adding new queries and properties specific to their use case can  modify the parent interface as well.

### Checklist

- [ ] This pull request includes a `fix`, `feature`, `enhancement`, `docs`, or `maintenance` label categorizing the change.

### DIFF
--- a/src/prefect/orion/database/dependencies.py
+++ b/src/prefect/orion/database/dependencies.py
@@ -29,6 +29,7 @@ MODELS_DEPENDENCIES = {
     "database_config": None,
     "query_components": None,
     "orm": None,
+    "interface_class": None,
 }
 
 
@@ -44,6 +45,7 @@ def provide_database_interface() -> OrionDBInterface:
     database_config = MODELS_DEPENDENCIES.get("database_config")
     query_components = MODELS_DEPENDENCIES.get("query_components")
     orm = MODELS_DEPENDENCIES.get("orm")
+    interface_class = MODELS_DEPENDENCIES.get("interface_class")
     dialect = get_dialect(connection_url)
 
     if database_config is None:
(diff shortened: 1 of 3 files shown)


src/prefect/orion/database/dependencies.py
```diff
@@ -178,11 +183,30 @@ def temporary_orm_config(tmp_orm_config: BaseORMConfiguration):
...
+    Args:
+        tmp_interface_class: Orion interface class to inject.
+
+    """
+    starting_interface_class = MODELS_DEPENDENCIES["interface_class"]
+    try:
+        MODELS_DEPENDENCIES["interface_class"] = tmp_interface_class
+        yield
+    finally:
+        MODELS_DEPENDENCIES["interface_class"] = starting_interface_class
```
chrisguidry, line 201: What do you think about:

```suggestion
    try:
        set_interface_class(tmp_interface_class)
        yield
    finally:
        set_interface_class(starting_interface_class)
```

which would let us trigger some additional behavior any time the interface class is set?

chrisguidry (approved):
Legit!

zangell44 (approved):

PR #6645, opened 2022-08-31
Refactor block documents queries for speed ⚡️

jlowin:
Closes https://github.com/PrefectHQ/nebula/issues/1492

@zangell44 noticed that block document queries are starting to take longer and longer in Prefect Cloud, even when querying for a nonexistent block document ID (see attached issue for detail, internal repo so re-summarizing here). Query complexity exploded into the hundreds of thousands, the culprit being an odd `seq scan` of the block document table, even though a block document ID was being provided.

When we looked closely, we identified that the blame lies with [this join condition](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/orion/models/block_documents.py#L148-L153):
```python
            sa.or_(
                db.BlockDocument.id == block_document_id,
                recursive_block_document_references_cte.c.parent_block_document_id.is_not(
                    None
                ),
            )
```
(note: this shows up as a where clause but is implicitly the join condition of an outer join specified immediately before)

The reason this is problematic is not the predicates, but the `OR`. In fact, both predicates should actually be instantaneous. However, many query planners opt for seq scans when evaluating `OR` conditions over multiple tables, and Postgres is no exception. This is why it was doing a seq scan even when a single, nonexistent ID was provided. 

In cases like this one, this resolution is straightforward: replace the `OR` condition with a `UNION` (or in our case, a `UNION ALL`). This lets the query planner take advantage of indexed lookups for each predicate, then trivially stitch the result together.

This change appears to work well and bring query times from hundreds of milliseconds down to the expected fraction of a millisecond. However, in implementing it I noticed that we form this query three different places: `read_block_document_by_id`, `read_block_document_by_name`, and `read_block_documents`. Therefore, I took the opportunity to refactor this code in the following ways:
- `read_block_document_by_id` and `read_block_document_by_name` are now just passthroughs to `read_block_documents`
- `read_block_documents` gains additional filter options to support those passthroughs: `block_type.slug`, `block_document.name`, and `block_document.id`.
- `read_block_documents` was making an additional DB query to load the ID's of block documents that met its filter criteria, then forming the "full" query from those materialized IDs; I refactored this so only a single DB query gets made.

Reviewed changes with @desertaxle and @zangell44 

Edit: see https://github.com/PrefectHQ/prefect/pull/6645#issuecomment-1234223411 for notes on implementation and a weird situation that prevented SQLite from being *quite* as efficient as we want (still makes two DB calls) but no longer suffers from the seq scan, so overall a big win.
### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ profile_default/
 ipython_config.py
 *.ipynb_checkpoints/*
 
+# Profiling
+/prof
+
 # Environments
 .python-version
 .env
--- a/src/prefect/orion/api/block_documents.py
+++ b/src/prefect/orion/api/block_documents.py
@@ -43,6 +43,7 @@ async def create_block_document(
 async def read_block_documents(
     limit: int = dependencies.LimitBody(),
     block_documents: Optional[schemas.filters.BlockDocumentFilter] = None,
+    block_types: Optional[schemas.filters.BlockTypeFilter] = None,
     block_schemas: Optional[schemas.filters.BlockSchemaFilter] = None,
     include_secrets: bool = Body(
         False, description="Whether to include sensitive values in the block document."
@@ -56,6 +57,7 @@ async def read_block_documents(
     result = await models.block_documents.read_block_documents(
         session=session,
         block_document_filter=block_documents,
+        block_type_filter=block_types,
         block_schema_filter=block_schemas,
         include_secrets=include_secrets,
         offset=offset,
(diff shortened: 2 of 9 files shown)


madkinsz (approved):
Sweet!

desertaxle (approved):
This is fire.

Not sure why the order of block documents is different than what the tests expect. A comment says that the block documents are sorted by block type name and block document name, but I don't think that was true before these changes.

jlowin:
:/ looking into the test failures - I'm not seeing them locally so I think they might have some phantom state from running the whole suite? I'll figure it out.

jlowin:
@desertaxle 
> A comment says that the block documents are sorted by block type name and block document name, but I don't think that was true before these changes.

The `order by` was being applied prior to a final select which is why they were showing up in the wrong order (fixed now) -- I've also removed that comment which is no longer true (and wasn't prior to these changes)

desertaxle (approved):

PR #6654, opened 2022-08-31
Add sync compatibility to `run` for all infrastructure types

madkinsz:
This is particularly helpful for testing infrastructure, but is also relevant for the many cases in which a user may want to run commands in infrastructure from sync flows or tasks.

### Example

Previously the following would not run the process as it returned a coroutine

```python
from prefect import flow
from prefect.infrastructure import Process

@flow
def foo():
    process = Process(command=["echo", "hello!"])
    process.run()

foo()
```

### Checklist

### DIFF
--- a/src/prefect/infrastructure/base.py
+++ b/src/prefect/infrastructure/base.py
@@ -51,9 +51,10 @@ class Infrastructure(Block, abc.ABC):
         identifier for the infrastructure.
 
         The call will then monitor the created infrastructure, returning a result at
-        the end containing a status code indicating if the infrastructure exited cleanly\
+        the end containing a status code indicating if the infrastructure exited cleanly
         or encountered an error.
         """
+        # Note: implementations should include `sync_compatible`
 
     @abc.abstractmethod
     def preview(self) -> str:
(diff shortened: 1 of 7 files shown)


madkinsz:
It is unclear to me if this is worth adding explicit test coverage for at this time

desertaxle (approved):
LGTM

Having sync-compatible methods on blocks seems to be a reoccuring theme. Do you think that sync-compatibility is worth implementing automatically or enforcing for all blocks?

bunchesofdonald (approved):
In general, I _personally_ feel like if it's worth adding, it's worth testing. But agree that sometimes, with changes like this, it feels a little silly.

madkinsz:
> Having sync-compatible methods on blocks seems to be a reoccuring theme. Do you think that sync-compatibility is worth implementing automatically or enforcing for all blocks?

Yeah, I think so. I considered finding some way to do that here, but decided it was out of scope.

> In general, I personally feel like if it's worth adding, it's worth testing. But agree that sometimes, with changes like this, it feels a little silly.

I think that I'd rather have an explicit check that all methods on a block are sync compatible defined in some generic block tests. Let me see if I can add some implicit coverage here though.

Issue #6648, opened 2022-08-31
Setting auto remove to true in Docker container blocks crashes agent


WillRaphaelson:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Using a docker infrastructure block and setting auto-remove to true seems to crash the agent with a 404 error from the docker API

### Reproduction

```python
Create a Docker Container Block block in the UI with auto remove set to true 

build and apply the deployment using that infra block

prefect deployment build flows/flow.py:hello_flow --name live-2 -q live -ib docker-container/prod --apply

run the agent and submit a flow run from the deployment you just created
```

### Error

```

(orion) ➜  prefect-live git:(main) ✗ prefect agent start -q live
Starting v2.3.0 agent connected to 
https://api.prefect.cloud/api/accounts/12242a57-9f05-4bf5-8853-9bff595d4bab/workspaces/38f22bf8-0f9d-4d15-9aa9-324e60ed3009...

  ___ ___ ___ ___ ___ ___ _____     _   ___ ___ _  _ _____
 | _ \ _ \ __| __| __/ __|_   _|   /_\ / __| __| \| |_   _|
 |  _/   / _|| _|| _| (__  | |    / _ \ (_ | _|| .` | | |
 |_| |_|_\___|_| |___\___| |_|   /_/ \_\___|___|_|\_| |_|


Agent started! Looking for work from queue(s): live...
12:01:46.092 | INFO    | prefect.agent - Submitting flow run 'bd6fa147-e94d-48d3-ab59-a6eb20e16504'
12:01:46.685 | INFO    | prefect.infrastructure.docker-container - Pulling image 'wraphaelson/live'...
12:01:47.501 | INFO    | prefect.infrastructure.docker-container - Creating Docker container 'prod'...
12:01:47.527 | INFO    | prefect.infrastructure.docker-container - Creating Docker container 'prod-1'...
12:01:47.556 | INFO    | prefect.infrastructure.docker-container - Creating Docker container 'prod-2'...
12:01:48.000 | INFO    | prefect.agent - Completed submission of flow run 'bd6fa147-e94d-48d3-ab59-a6eb20e16504'
12:01:48.046 | INFO    | prefect.infrastructure.docker-container - Docker container 'prod-2' has status 'running'
[...]

PR #6656, opened 2022-08-31
Update Docker container runs to be robust to container removal

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6648

If a container was removed while streaming logs, we threw an exception instead of returning the correct result. This  pull request updates our container watching implementation to repeatedly yield the latest container metadata it has seen so a wrapper can catch any `NotFound` errors and still produce a `DockerContainerResult`.

I imagine there are some edge cases here where we end up with a version of the container metadata that does not contain the exit code, but we can tackle those in a later as they'd require more redesign of how we stream logs. For example, the call to `wait()` could be fully decoupled from the log streaming calls  by running them in separate worker threads.

### Example

As described in the linked issue, the following would fail as the container was removed before a final status could be retrieved:
 
```python
from prefect.infrastructure import DockerContainer
from prefect.logging.configuration import setup_logging

setup_logging()

DockerContainer(command=["echo", "hello!"], auto_remove=True).run()
```

### Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,8 @@ filterwarnings =
     ignore:`DeploymentSpec` has been replaced by `Deployment`:DeprecationWarning
     # This warning is raised on Windows by Python internals
     ignore:the imp module is deprecated:DeprecationWarning
+    # Dockerpy is behind on this one 
+    ignore:distutils Version classes are deprecated:DeprecationWarning
 
 [isort]
 skip = __init__.py
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -4,7 +4,7 @@ import sys
 import urllib.parse
 import warnings
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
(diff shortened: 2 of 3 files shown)


WillRaphaelson:
Nice, so this is definitely an improvement as my agent doesn't crash anymore, but I still get this ominous warning I'm not quite sure what to make of?

```
16:59:16.093 | WARNING | prefect.infrastructure.docker-container - Docker container prod-3 was removed before we could wait for its completion.
```

madkinsz:
> Nice, so this is definitely an improvement as my agent doesn't crash anymore, but I still get this ominous warning I'm not quite sure what to make of?

Yeah... hm. Basically we call `container.wait()` to wait for the final status code but we do that _after_ streaming all of the logs from the container. If the container is auto-removed it looks like it is sometimes removed so fast that we don't get a chance to stream the logs and the function is interrupted before we can call `container.wait()`. Here I basically add handling for this early interrupt, preventing it from raising the error to the user, and we will _attempt_ to retrieve the status code of your container but it could theoretically be missing. I think that log is an important signal that we failed to complete our streaming and such, perhaps it could be worded differently? Removing the need for the log entirely would require some more significant changes.

Issue #6663, opened 2022-09-01
Parameters in UI of Github repository deployment don't update when flow code is changed


lennertvandevelde:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When adding new parameters to flow-code, pushing it to Github and deployment build and apply, the deployment in the UI is not update to include these parameters.

### Reproduction
Commands:
``` prefect deployment build -n ghtest -q prod -sb github/test --apply test-flow.py:test_flow```
```python
# Original flow code
from prefect import flow

@flow
def test_flow(name):
    print("Hello, {}!".format(name))

# Changed flow code
from prefect import flow, task

@flow
def test_flow(name, name2):
    print("Hello, {} and {}!".format(name, name2))
```
deployment.yaml:
```
entrypoint: test-flow.py:test_flow
parameter_openapi_schema:
  title: Parameters
  type: object
  properties:
    name:
      title: name
    name2:
      title: name2
  required:
  - name
  - name2
  definitions: null
```

### Error

Parameters can not be changed, the flow run does however require the new parameter:

```
Validation of flow parameters failed with error: ParameterTypeError('Flow run received invalid parameters:\n - name2: field required')
Traceback (most recent call last):
  File "/home/lennert/.pyenv/versions/export-venv/lib/python3.7/site-packages/prefect/engine.py", line 284, in retrieve_flow_then_begin_flow_run
    parameters = flow.validate_parameters(flow_run.parameters)
  File "/home/lennert/.pyenv/versions/export-venv/lib/python3.7/site-packages/prefect/flows.py", line 269, in validate_parameters
    raise ParameterTypeError.from_validation_error(exc) from None
prefect.exceptions.ParameterTypeError: Flow run received invalid parameters:
 - name2: field required
```

![image](https://user-images.githubusercontent.com/61053080/187876739-521fb125-9095-4b72-9cf9-d3258a344a71.png)

### Versions

```

2.3.0

```

### Additional context

A solution is to delete the deployment before reapplying, but in CI/CD this is not always possible.

PR #6668, opened 2022-09-01
Add parameter_openapi_schema to inject database

lennertvandevelde:
Closes [#6663](https://github.com/PrefectHQ/prefect/issues/6663)
When building and reapplying after changing parameters of a flow, the parameters in the Orion UI are not changed. 
Fixed by adding 'parameter_openapi_schema' to include for insertion in Orion database.

#### Code:
```

@flow
def test_flow(name):
    print("Hello, {} and {}!".format(name))
```
#### Command:
```
prefect deployment build -n ghtest -q prod --apply test-flow.py:test_flow
```

#### UI result:
image

#### Code change:
```

@flow
def test_flow(name, name2):
    print("Hello, {} and {}!".format(name, name2))
```
#### Command:
```
prefect deployment build -n ghtest -q prod --apply test-flow.py:test_flow
```
#### UI result:
No change

After adding ```parameter_openapi_schema```:
#### UI result
image

### Checklist

### DIFF
--- a/src/prefect/orion/models/deployments.py
+++ b/src/prefect/orion/models/deployments.py
@@ -86,7 +86,7 @@ async def create_deployment(
                         "path",
                         "entrypoint",
                         "infra_overrides",
-                        "parameter_openapi_schema"
+                        "parameter_openapi_schema",
                     },
                 ),
             },
--- a/src/prefect/orion/models/deployments.py
+++ b/src/prefect/orion/models/deployments.py
@@ -72,21 +72,10 @@ async def create_deployment(
             set_={
                 **deployment.dict(
                     shallow=True,
-                    include={
-                        "schedule",
-                        "is_schedule_active",
-                        "description",
-                        "tags",
-                        "parameters",
-                        "updated",
-                        "work_queue_name",
-                        "storage_document_id",
-                        "infrastructure_document_id",
-                        "manifest_path",
-                        "path",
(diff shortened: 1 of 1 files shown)


src/prefect/orion/models/deployments.py
```diff
@@ -86,6 +86,7 @@ async def create_deployment(
                         "path",
                         "entrypoint",
                         "infra_overrides",
+                        "parameter_openapi_schema"
```
zangell44, line 89: Great catch!

I think we should be explicitly excluding instead of including here, as this is likely to break in the future when we add more fields.

Could we change
```python 
set_={
                **deployment.dict(
                    shallow=True,
                    include={
...
```

to

```python
set_={
                **deployment.dict(
                    shallow=True,
                    exclude_unset=True,
                    exclude={"id", "created", "updated"}
}
```

lennertvandevelde: ```"updated"``` should be included I assume? Otherwise, Last Updated will not change in the Orion Database

madkinsz:
Hi! Can you update a test to include coverage for this change? I believe https://github.com/PrefectHQ/prefect/blob/main/tests/orion/models/test_deployments.py#L78 is the relevant test

madkinsz (approved):
We've updated the test coverage separately. Thanks for contributing!

PR #6670, opened 2022-09-01
Add error display to UI when backend is unreachable

marichka-offen:
This PR adds error display to UI when backend is unreachable 
Closes https://github.com/PrefectHQ/prefect/issues/6442

### Checklist

### DIFF
--- a/orion-ui/src/App.vue
+++ b/orion-ui/src/App.vue
@@ -62,7 +62,7 @@
     workQueuesRouteKey
   } from '@prefecthq/orion-design'
   import { PGlobalSidebar, PIcon, media, showToast } from '@prefecthq/prefect-design'
-  import { computed, onMounted, provide, ref, watchEffect } from 'vue'
+  import { computed, provide, ref, watchEffect } from 'vue'
   import { blockDocumentsApi } from './services/blockDocumentsApi'
   import { blockSchemasApi } from './services/blockSchemasApi'
   import { blockTypesApi } from './services/blockTypesApi'
--- a/orion-ui/src/App.vue
+++ b/orion-ui/src/App.vue
@@ -61,14 +61,12 @@
     workQueuesApiKey,
     workQueuesRouteKey
   } from '@prefecthq/orion-design'
-  import { PGlobalSidebar, PIcon, media, showToast } from '@prefecthq/prefect-design'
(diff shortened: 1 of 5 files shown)


orion-ui/src/App.vue
```diff
@@ -129,7 +130,19 @@
     mobileMenuOpen.value = false
   }
 
-  watchEffect(() => document.body.classList.toggle('body-scrolling-disabled', showMenu.value && !media.lg))
+  const healthCheck = async (): Promise<void> => {
+    try {
+      await healthApi.getHealth()
+    } catch (error) {
+      showToast('Network error', 'error', { timeout: false })
```
pleek91, line 137: I think we can give a more intuitive message here. Maybe check with product?

orion-ui/src/App.vue
```diff
@@ -129,7 +130,19 @@
...
+      console.warn(error)
+    }
+  }
+
+  watchEffect(() => {
+    healthCheck()
```
pleek91, line 143: Why call this in a watch effect?

marichka-offen: I haven't thought it through yet, just added it there to see if API returns an error since watchEffect was already there. Would it be better to put it in onMounted?

pleek91: I'm not sure it even needs to be in a hook. Should just be able to call it directly

marichka-offen: True, I don't think I've seen such a pattern of us calling functions directly in our code

pleek91: I would probably move the health check function into a `api` utility file. Then just import and call it in `App.vue`

orion-ui/src/services/healthApi.ts
```diff
@@ -0,0 +1,7 @@
+import { HealthApi } from '@prefecthq/orion-design'
+import { createActions } from '@prefecthq/vue-compositions'
+import { ApiRoute } from '@/mixins/ApiRoute'
+
+export class OrionHealthApi extends ApiRoute(HealthApi) {}
+
+export const healthApi = createActions(new OrionHealthApi())
```
pleek91, line 7: I'm so glad we won't have to do this anymore soon!

pleek91 (approved):

PR #6672, opened 2022-09-01
Use windows specific signals when stopping Orion in windows

### LINKED ISSUE #5880: Shut down PowerShell session to release server port
## Description

With 2.0b6, prefect server instance does not release port when shut down.

## Reproduction / Example

When running a Prefect server instance on Windows 10 in PowerShell:

- With 2.0b6
- If you stop the server with Ctrl-C, the port is not released 
- If you try to start a new server instance, it fails 
- You must close the PowerShell console instance (or tab in Windows Terminal) and launch another server instance in a new PowerShell console.

bunchesofdonald:
As detailed in this [Python bug report](https://bugs.python.org/issue26350), while implemented `signal.SIGTERM` doesn't really work on Windows. The signal module supplies a Windows specific equivalent called `signal.CTRL_BREAK_EVENT`.  This PR makes `prefect orion start` use that signal when shutting down in windows.

Closes #5880 

### Example

No real example to provide here, but I did test it in Windows via Parallels and ensured that the non-Windows version continues to function.

### Checklist


src/prefect/cli/orion.py
```diff
@@ -138,22 +139,33 @@ async def start(
...
+        else:
+
+            def stop_orion(*args):
+                app.console.print("\nStopping Orion...")
+                os.kill(orion_process_id, signal.SIGTERM)
+                signal.signal(signal.SIGINT, kill_orion)
+
+            def kill_orion(*args):
+                app.console.print("\nKilling Orion...")
+                os.kill(orion_process_id, signal.SIGKILL)
```
madkinsz, line 168: This patterns seems useful as a `send_kill_signal(pid: int)` utility in `processutils` so future consumers of this pattern don't need to worry about Windows specific logic.

madkinsz: (It'd also make it easier to test)

bunchesofdonald: Good idea! I'll work down that route.

sm-Fifteen:
Again, why run uvicorn as a subprocess instead of spinning up an event loop for it and then `run_forever`, which is exactly what running it as an executable module does? Uvicorn is perfectly capable of stopping itself on its own and `prefect orion start` has nothing else to do once the server is up.

sm-Fifteen:
> There are a few reasons off the top of my head:
> 
>     * Previously, `prefect orion start` did do other things like start an agent. We may have multiple actions again in the future.
> 
>     * We set environment variables to pass configuration changes to the child process.
> 
>     * We're already in an event loop. Running applications on event loops outside the main thread can be troublesome.

@madkinsz: Now that you mention it, I actually did run into [a very specific uvicorn issue](https://github.com/encode/uvicorn/issues/526) way back when, concerning being unable to set signal handlers on Windows on any thread besides the main one, and how that was [causing issues with running uvicorn as a Windows service](https://github.com/mhammond/pywin32/issues/1452), so you're probably right on these, especially the last one. Nevermind all that, then.

> Is there an issue where you explain the benefits of running it in the current process? This pull request is not a great place to air the concern.

I mentioned this as being a potential solution in https://github.com/PrefectHQ/prefect/issues/5880#issuecomment-1198229351, since that's how I normally run my ASCI apps through Uvicorn and this seemed like a nice way to work around these signal issues, but I'd forgotten how this came with a lot of complications in some contexts, that you seem to have already run into prior.

madkinsz:
@sm-Fifteen thanks for the additional context and links! Also thanks for being understanding about our approach :) it does seem nice to side-step the complexity of managing the subprocess but yeah there are other demons lurking. I imagine that we could change this sometime in the future still.

madkinsz:
@bunchesofdonald 

```
        # Send first SIGINT and expect a SIGTERM
>       signal.raise_signal(signal.SIGINT)
E       AttributeError: module 'signal' has no attribute 'raise_signal'
```

On Python 3.7

bunchesofdonald:
The tests should pass under Python 3.7 and I manually verified that they pass in Windows as well.

bunchesofdonald:
Looks like the the signal isn't always being caught by the tests and instead by pytest. I'll see if I can find a better way to handle this.

madkinsz (approved):
What an adventure

PR #6673, opened 2022-09-01
Protect against race condition with deployment schedules

jlowin:
If a deployment with a huge number of scheduled runs has its schedule made inactive, then it becomes highly likely that it will be visited by the scheduler service _while_ its auto-scheduled runs are being deleted (simply because the delete is a long-running operation). The net result is a race condition in which a deployment's schedule is disabled, but it has a bunch of scheduled runs. The only recourse is to toggle the schedule on and off a second time.

This PR solves this by `commit()`-ing the inactive schedule before deleting the auto-scheduled runs. This has the effect of preventing the scheduler from picking up the deployment, since the inactive schedule change has been committed. However, because our `get_session` dependency opens a transaction with the `begin()` method, we can't manually `commit()` it. Instead, this manually opens a new session to allow explicit transaction control. 

Note: a future enhancement could move the auto-scheduled runs delete to a background task.

### Checklist


madkinsz:
I imagine we'll need to make sure this behaves well with Nebula's session management but this looks great to me! Is there a way to write a test?

jlowin:
I think it's a tricky one to test in an affirmative way because it's replacing a flush with a commit, and the race condition in question was difficult to see regardless (and probably not possible at all in SQLite). What I can say is that the mid-function `.commit()` _failed_ in the current test suite with the `get_session()` dependency, so the fact that schedule unit tests are now passing is a good indication that both commits are happening as expected.

madkinsz (approved):

PR #6674, opened 2022-09-01
Update `Deployment.infrastructure` to accept types outside of the core library

madkinsz:
Updates deployment `infrastructure` field type to the `Infrastructure` base class instead of enumerating all types. This is required for the upcoming infrastructure blocks in collections e.g. `ECSTask` and `CloudRunJob`.

#6622 should allow us to do this without any issues.

### Example

```python
from prefect_aws.ecs import ECSTask

Deployment(name="foo", infrastructure=ECSTask(...))
```

### Checklist

### DIFF
--- a/tests/test_deployments.py
+++ b/tests/test_deployments.py
@@ -6,7 +6,7 @@ from prefect.blocks.core import Block
 from prefect.deployments import Deployment
 from prefect.exceptions import BlockMissingCapabilities
 from prefect.filesystems import S3, GitHub, LocalFileSystem
-from prefect.infrastructure import DockerContainer, Process
+from prefect.infrastructure import DockerContainer, Infrastructure, Process
 
 
 class TestDeploymentBasicInterface:
@@ -29,6 +29,19 @@ class TestDeploymentBasicInterface:
     async def test_that_infrastructure_defaults_to_process(self):
         d = Deployment(name="foo")
         assert isinstance(d.infrastructure, Process)
+
+    async def test_infrastructure_accepts_arbitrary_infra_types(self):
+        class CustomInfra(Infrastructure):
+            type = "CustomInfra"
+
+            def run(self):
+                return 42
+
+            def preview(self):
+                return "woof!"
+
+        d = Deployment(name="foo", infrastructure=CustomInfra())
+        assert isinstance(d.infrastructure, CustomInfra)
 
     async def test_default_work_queue_name(self):
         d = Deployment(name="foo")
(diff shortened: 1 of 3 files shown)


madkinsz:
@peytonrunyan do you want to give this a test with your work? It'd be nice to include an explicit test for this by defining a dynamic `Infrastructure` type and creating a deployment with it if you're up to it.

madkinsz:
Sorry this should build on #6622, I thought it was merged already but it's not. There's no way this should be passing tests as-is, I'm guessing there's some important coverage missing for deployments.

peytonrunyan:
I'm in!

madkinsz:
@peytonrunyan I rebased this onto #6622 — let me know how it goes. It'd be good to add a unit test here with a fake infrastructure block type if you have the chance.

peytonrunyan:
I was experimenting with a GUI tool to squash commits and it looks like I went back one commit too far. I'm going to hold off on trying to fix this until I've got a slightly better idea of the right solution. Git isn't a strong point for me.

src/prefect/orion/database/configurations.py
```diff
@@ -109,7 +109,12 @@ async def engine(self) -> AsyncEngine:
             if connect_args:
                 kwargs["connect_args"] = connect_args
 
-            engine = create_async_engine(self.connection_url, echo=self.echo, **kwargs)
+            engine = create_async_engine(
+                self.connection_url,
+                echo=self.echo,
+                execution_options={"isolation_level": "AUTOCOMMIT"},
```
cicdw, line 115: what are these changes related to? I'm not as familiar with our database configuration but my understanding is that changing the isolation level is a pretty significant change

cicdw (changes requested):
Deployment changes look good to me, but i have a question about the DB change

src/prefect/orion/database/configurations.py
madkinsz: This was a weird merge with `main`. Should be gone now.

tests/test_deployments.py
```diff
@@ -30,6 +30,19 @@ async def test_that_infrastructure_defaults_to_process(self):
...
+        class CustomInfra(Infrastructure):
+            type = "CustomInfra"
+
+            def run(self):
+                return 42
+
+            def preview(self):
+                return "woof!"
+
+        d = Deployment(name="foo", infrastructure=CustomInfra())
```
madkinsz, line 43: Can you add a test with `infrastructure=CustomInfra().dict()` too? We need that to ensure it works from YAML.

madkinsz: (or an actual test that loads from a YAML config which would be better)

madkinsz: @peytonrunyan want to make sure you've seen this as we kick off the new week

madkinsz:
Not merged. Something went wrong. Will fix :D

madkinsz:
Alright, I've rebased the history once again.

anna-geller (approved):

PR #6691, opened 2022-09-02
Enable stale action to handle aging issues and pull requests

bunchesofdonald:
This enables the [stale action](https://github.com/actions/stale), but sets it in debug mode so that we can look at the changes it would make if it were fully enabled. This way we can change the settings around till we're happy.

To see the debug output we'll have to set the `ACTIONS_STEP_DEBUG` secret in the repo to `true`.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,21 @@
+name: "Close stale issues and PRs"
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          enable-statistics: true
+          stale-issue-message: "This issue is stale because it has been open 30 days with no activity. To keep this issue open remove stale label or comment."
+          close-issue-message: "This issue was closed because it has been stale for 7 days with no activity. If this issue is important or you have more to add feel free to re-open it."
+          stale-pr-message: "This pull request is stale because it has been open 30 days with no activity. To keep this pull request open remove stale label or comment."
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
I do not have permissions to add the secret, cc @abrookins 

I'm basically fine to turn this on without debugging too though.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          enable-statistics: true
+          stale-issue-message: "This issue is stale because it has been open 30 days with no activity. To keep this issue open remove stale label or comment."
+          close-issue-message: "This issue was closed because it has been stale for 7 days with no activity. If this issue is important or you have more to add feel free to re-open it."
+          stale-pr-message: "This pull request is stale because it has been open 30 days with no activity. To keep this pull request open remove stale label or comment."
+          close-pr-message: "This pull request was closed because it has been stale for 7 days with no activity. If this pull request is important or you have more to add feel free to re-open it."
+          days-before-stale: 30
+          days-before-close: 7
```
cicdw, line 19: 7 seems tight to me, I'd probably vote to be more generous with 14 but easy enough to change

cicdw (approved):
I added the secret 👍

.github/workflows/stale.yml
madkinsz: They'd have 7 days to bump then another 30 days after that before it gets marked stale again 🤷‍♀️. I also lean generous on this though.

cicdw: Gotcha gotcha - yea this is probably one of those things that will be easier to decide once we have a feel for it in practice.

bunchesofdonald: I'm cool with 14, I'll bump that and merge this.

Issue #6563, opened 2022-08-25
Adding additional parameters to flows and recreating the deployments does not reflect the new param in the UI


andnig:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When one creates a flow + deployment and afterwards adds an additional parameter to this flow and recreates the deployment, the new parameter is NOT reflected in the UI. 

### Reproduction

1. Create a flow with one or more parameters
2. Run ``prefect deployment build``
3. Add the parameter value to the deployment.yaml (this step can also be done with the UI - the bug is not related to adding the parameter here)
4. Run ``prefect deployment apply``
--> Up until here, everything is fine. The deployment is shown in the UI, with parameters + param values.

5. Add an additional parameter to the flow
6. Run ``prefect deployment build``
7. Add the new parameter value to the deployment.yaml
8. Run ``prefect deployment apply``
--> The deployment in the UI still only shows the old parameters - the newe parameter is not visible and you can't set values for it

### Error

The deployment in the UI still only shows the old parameters - the newe parameter is not visible and you can't set values for it

### Versions

```
Version:             2.2.0
API version:         0.8.0
Python version:      3.9.13
Git commit:          e3651362
Built:               Tue, Aug 23, 2022 2:18 PM
OS/Arch:             linux/x86_64
Profile:             rm_dwh_200
Server type:         hosted
```

### Additional context

As a workaround, one can delete the first deployment and fully recreate one from scratch then the parameter is available.

discdiver:
Noted by another user in community Slack and I was able to replicate.

PR #6696, opened 2022-09-03
Adds test for parameter openapi schema

cicdw:
This PR ensures that the Parameter OpenAPI schema is updated with each deployment creation; previously this update was not occurring, leading to the UI not displaying the correct information.

Closes https://github.com/PrefectHQ/prefect/issues/6563

### Example
Take a flow defined as:
```python
from prefect import flow

@flow 
def my_flow(old: str):
    pass
```
and create a basic deployment for it (CLI or Python, doesn't matter).  Add an override for `old` in the UI.

Then update the flow:
```python
from prefect import flow

@flow 
def my_flow(old: str, new: bool = False):
    pass
```
and re-create a basic deployment for it.  Visit the UI - on `main`, the `new` parameter would not be displayed, but on this branch it will be.

### DIFF
--- a/src/prefect/orion/models/deployments.py
+++ b/src/prefect/orion/models/deployments.py
@@ -78,6 +78,7 @@ async def create_deployment(
                         "description",
                         "tags",
                         "parameters",
+                        "parameter_openapi_schema",
                         "updated",
                         "work_queue_name",
                         "storage_document_id",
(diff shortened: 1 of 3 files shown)


madkinsz:
Ah we've got a user contribution for this at https://github.com/PrefectHQ/prefect/pull/6668 that's just missing a test. Let's merge both since Zach requested that we switch from a list of `includes` to `excludes` here.

cicdw:
Ah brilliant! If you want to merge the user PR right now, I can resolve merge conflicts and my test will persist 👍

madkinsz (approved):

Issue #6686, opened 2022-09-02
When a task is mapped over a list of iterables the task hierarchy is not correctly displayed.


robalar:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

When a task is mapped over a list of iterables the task hierarchy is not correctly displayed. The mapped tasks are shown as connected to the flow root, not the task their data came from (see notes).

### Reproduction

1.  Run the following flow code:
```python
from prefect import task, flow, get_run_logger


@task
def sum_task(chunk: list[int]) -> int:
    logger = get_run_logger()
    logger.info(chunk)
    return sum(chunk)


@task
def numbers() -> list[tuple[int, int]]:
    return [(0, 1), (2, 3), (3, 4)]


@flow
def test_flow():
    logger = get_run_logger()

    n = numbers()
    sums = sum_task.map(n)

    logger.info("Sums: %s", [x.result() for x in sums])


if __name__ == "__main__":
    test_flow()
```
2. Open Orion dashboard
3. Radar visualisation displays incorrect hierarchy

### Error

![Screenshot 2022-09-02 at 1 56 49 pm](https://user-images.githubusercontent.com/67542061/188169125-55f4f98d-f05c-469d-baf5-5600be0956cb.png)

### Browers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```
Version:             2.2.0
API version:         0.8.0
Python version:      3.9.13
Git commit:          e3651362
Built:               Tue, Aug 23, 2022 2:18 PM
OS/Arch:             darwin/x86_64
Profile:             default
Server type:         hosted
```

bunchesofdonald:
Thanks @robalar, this does appear to be broken, even when using `.submit` for the task run. We'll look into it.

madkinsz:
Is the data correct in the API and just displayed incorrectly?

cc @zhen0 if this is broken for all task runs we may want to bump this to `priority:high`.

robalar:
This is the API response from the `graph` endpoint (from a modified version of the provided flow code using submit):

```json
[
  {
    "id": "c83a2a71-fdf9-44f8-8734-5bac0136e929",
    "upstream_dependencies": [],
    "state": {
[...]

PR #6701, opened 2022-09-03
Preserve task dependencies to futures passed as parameters in `.map`

robalar:
Preserves relationships between a mapped task and its parameters if that parameter is a future, closes https://github.com/PrefectHQ/prefect/issues/6686.

### Example

```python

from prefect import task, flow, get_run_logger


@task
def log_task(x: int):
    logger = get_run_logger()
    logger.info(x)


@task
def flat_numbers() -> list[int]:
    return [0, 1, 2, 3, 4]


@flow
def test_flow():
    x = flat_numbers.submit()
    log_task.map(x)
```

Will now correctly produce this graph:

Screenshot 2022-09-03 at 3 44 34 pm

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -55,7 +55,7 @@ from prefect.logging.loggers import (
     task_run_logger,
 )
 from prefect.orion.schemas import core
-from prefect.orion.schemas.core import FlowRun, TaskRun
+from prefect.orion.schemas.core import FlowRun, TaskRun, TaskRunInput
 from prefect.orion.schemas.data import DataDocument
 from prefect.orion.schemas.filters import FlowRunFilter
 from prefect.orion.schemas.responses import SetStateStatus
@@ -749,10 +749,16 @@ async def begin_task_map(
     task_runner: Optional[BaseTaskRunner],
 ) -> List[Union[PrefectFuture, Awaitable[PrefectFuture]]]:
     """Async entrypoint for task mapping"""
+    # We need to resolve some futures to map over their data, collect the upstream
+    # links beforehand to retain relationship tracking.
+    task_inputs = {
+        k: await collect_task_run_inputs(v, max_depth=0) for k, v in parameters.items()
+    }
 
-    # Resolve any futures / states that are in the parameters as we need to
-    # validate the lengths of those values before proceeding.
-    parameters.update(await resolve_inputs(parameters))
+    # Resolve the top-level parameters in order to get mappable data of a known length.
+    # Nested parameters will be resolved in each mapped child where their relationships
+    # will also be tracked.
+    parameters = await resolve_inputs(parameters, max_depth=1)
 
     iterable_parameters = {}
     static_parameters = {}
@@ -760,7 +766,7 @@ async def begin_task_map(
         if isinstance(val, unmapped):
             static_parameters[key] = val.value
         elif isiterable(val):
-            iterable_parameters[key] = val
+            iterable_parameters[key] = list(val)
         else:
             static_parameters[key] = val
 
@@ -795,15 +801,14 @@ async def begin_task_map(
                 wait_for=wait_for,
                 return_type=return_type,
                 task_runner=task_runner,
+                extra_task_inputs=task_inputs,
             )
         )
 
     return await gather(*task_runs)
 
 
-async def collect_task_run_inputs(
-    expr: Any,
-) -> Set[Union[core.TaskRunResult, core.Parameter, core.Constant]]:
+async def collect_task_run_inputs(expr: Any, max_depth: int = -1) -> Set[TaskRunInput]:
     """
     This function recurses through an expression to generate a set of any discernable
     task run inputs it finds in the data structure. It produces a set of all inputs
@@ -835,6 +840,7 @@ async def collect_task_run_inputs(
         expr,
         visit_fn=add_futures_and_states_to_inputs,
         return_data=False,
+        max_depth=max_depth,
     )
 
     return inputs
@@ -847,13 +853,17 @@ async def get_task_call_return_value(
     wait_for: Optional[Iterable[PrefectFuture]],
     return_type: EngineReturnType,
     task_runner: Optional[BaseTaskRunner],
+    extra_task_inputs: Optional[Dict[str, Set[TaskRunInput]]] = None,
 ):
+    extra_task_inputs = extra_task_inputs or {}
+
     future = await create_task_run_future(
         task=task,
         flow_run_context=flow_run_context,
         parameters=parameters,
         wait_for=wait_for,
         task_runner=task_runner,
+        extra_task_inputs=extra_task_inputs,
     )
     if return_type == "future":
         return future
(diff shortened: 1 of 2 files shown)


src/prefect/engine.py
```diff
@@ -750,6 +750,11 @@ async def begin_task_map(
 ) -> List[Union[PrefectFuture, Awaitable[PrefectFuture]]]:
     """Async entrypoint for task mapping"""
 
+    # Preserve task dependencies to futures passed as parameters,
+    # as when they are resolved the relationship will be lost.
+    wait_for = [] if wait_for is None else wait_for
+    wait_for.extend([x for x in parameters.values() if isinstance(x, PrefectFuture)])
```
madkinsz, line 756: Thanks a bunch for digging into this issue! It's really helpful. Unfortunately there a couple issues with this implementation:
- This will drop the parameter _key_ that the future is passed under. We need to preserve that data but it looks like we'll need to do some refactoring to pass it downstream.
- This will not find futures nested inside parameters. We use a very magical `visit_collections` utility for this.

We can take this on, but if you're interested in looking further the typical pattern for dependency collection is at https://github.com/PrefectHQ/prefect/blob/a48eb57ded87300a165090517a829afcf54e5c4b/src/prefect/engine.py#L959

robalar: Thanks for taking the time to take a look!

> This will drop the parameter key that the future is passed under. We need to preserve that data but it looks like we'll need to do some refactoring to pass it downstream.

I agree that dropping the key is a bad idea, however, I'm not too clear on where I should be persisting that data to. 

> This will not find futures nested inside parameters. We use a very magical `visit_collections` utility for this.

Ah cheers, I realised that this would not work for chained maps so have a version working for those, but this should be what I need to generalise it even further.

Very happy to keep working away at this, but if it's going to be too much bandwidth for you guys to review and you'd rather do it in-house just let me know!

madkinsz: If you're interested you're welcome to take it further. I think you'll need to handle a few complex cases though and then there will be a little bit of engine refactoring to pass the relevant data downstream.

I think the most helpful next step is to write test cases (that will fail for now) covering the each way that task relationships can get created by futures with map. Here are a few example cases I'm thinking about:

```python

@task
def foo(x):
   return x

@task
def bar(x, y):
   return x + y

@task
def many_foo():
    return [1,2,3]
 
@flow
def my_flow():
   x = foo.submit(1)
   bar.map([x, 2], y=1)   # Relationship between mapped child bar-0 and foo-0 but not mapped child bar-1
   bar.map([1, 2, 3], y=x)   # Relationship between all mapped children of bar and foo-0
   x2 = foo.submit(2)
   bar.map([x, x2], y=0)  # Relationships between foo-0/mapped child bar-0 and foo-1/mapped child bar-1
   xs = many_foo.submit()
   bar.map(xs, y=0) # Relationships between all mapped children of bar and many_foo
```

I'm not sure what the best way to handle this will be yet! We're off until Tuesday, but we can try to provide some  more specific guidance on the best way to pass the data around then. If it looks like it'll be painful or really hard we can take a more active role.

robalar: Excellent! I have added those test cases plus one for more deeply nested maps. I will push a `wait_for` based solution to handle these cases, which can hopefully be useful during the refactor.

robalar:
@madkinsz Have pushed a commit that passes all the test cases. It obviously uses `wait_for` but feel like it should extend when we start persisting the keys too. 

I don't love that I had to add a special case to `visit_collection` but could not find any other way to recuse into resolved values and avoid infinite recursion. 

I also found some tests in the same file that test relationships but only via values returned, I believe my tests would sit better next to them. As a side note, have you considered breaking up this file? It's really quite large and hard to orient/find things in.

robalar:
Ah, that makes so much sense! The nesting was giving me a massive headache, but it didn't need to be! Feels obvious looking at it..... :smile: 

>I did not remove any of the functions you introduced and there may be some additional polish here. Are you interested in cherry-picking that commit into here and cleaning things up?

Yes very happy to tidy up!

> I'd prefer to keep all the mapping tests separate from those other inputs tests for now since they're a special case. If you want to create a TestTaskMapInputs section that's fine with me.

Noted, will split into its own section.

madkinsz:
Excellent work! Since I wrote some of this I'm going to request @bunchesofdonald as a reviewer. This looks good to me though.

robalar:
@madkinsz thank you for your guidance on this! Not sure I can claim to have done too much of it, but glad to have helped it along!

src/prefect/engine.py
```diff
@@ -795,15 +801,14 @@ async def begin_task_map(
                 wait_for=wait_for,
                 return_type=return_type,
                 task_runner=task_runner,
+                extra_task_inputs=task_inputs,
```
bunchesofdonald, line 804: I'm not _excited_ about `extra_task_inputs`, but I also can't think of a better approach at the moment. It may point at a possible need to refactor the way we validate the parameters to a map call or the way we gather at task inputs at some point.

tests/test_tasks.py
```diff
@@ -1948,6 +1951,195 @@ def my_flow():
...
+        graph = await models.flow_runs.read_task_run_dependencies(
+            session=session, flow_run_id=flow_run_id
+        )
+
+        return {x["id"]: [d.id for d in x["upstream_dependencies"]] for x in graph}
+
+    async def test_map_preserves_dependencies_between_futures_all_mapped_children(
+        self, session
+    ):
+        """
```
bunchesofdonald, line 1972: I love these docstrings, such a great way of showing the complexity involved while giving a better understanding of the test. We should consider doing this in other places, some tests are pretty hard to decode.

bunchesofdonald: A very small nitpick: Without more of a description of the _test_ these should probably be standard comments instead of docstrings.

bunchesofdonald (approved):
I think this looks great! That test suite is fantastic.

src/prefect/engine.py
madkinsz: Yeah me neither. I think ideally we have a way to track lengths on futures and create indexed children of futures. When the future has no known length, we resolve it to determine the length for mapping then access its indices (which creates child futures) that we pass downstream. That way, our existing tracking will work.

That said, users want to create dependencies between tasks in their graph without necessarily _waiting for_ them to finish. `extra_task_inputs` may be useful for user-defined relationships that do not have extra implications.

tests/test_tasks.py
robalar: Thank you! I was drawing these to better understand the relationship so thought why not include them for future readers. Will update to standard comments.

src/prefect/engine.py
madkinsz: cc @WillRaphaelson for reference

tests/test_tasks.py
robalar: Updated to docstrings of the flow as that is what they are a description of

robalar:
@madkinsz do you require any more action from me to get this merged? all tests are passing, netlify seems to be failing due to authentication issues to a private repo.

madkinsz:
See #6728 for some notes on that failure, it's expected right now. Thanks a bunch for contributing!

PR #6703, opened 2022-09-03
Simplify the nested block document query

jlowin:
In #6645 we refactored the block document query for significant performance gains, but noted an unfortunate need to create a slightly different implementation for sqlite.

In #6702 we unified the sqlite and postgres implementations of this query.

#6645 was motivated by replacing a slow `OR` condition with `UNION`; in this PR we restructure the query to remove that UNION entirely and cut the query complexity by about half.

In the original query, we began by using the provided filters to load all _parent_ block document ids, then used a recursive CTE to find all _referenced_ block documents, including potentially nested ones. We then queried the block document table, using the information in those two queries to know which documents to load. It is because we were using those two disjoint queries that we needed an `OR` or `UNION` statement at all.

In this query, we modify the recursive CTE so that instead of initializing it with any reference blocks contained in the parent documents, we initialize it with the parent block documents themselves. This means that after the recursive query finishes, it contains not only the IDs of any nested references, but the parent document IDs as well! This eliminates the need for the `OR` / `UNION` entirely: all the IDs we need are in the recursive output and we can do an efficient inner join to the block document table to load all the information we need.

The net result of #6702 and this PR is that we no longer need to relegate this query to the `QueryComponents` as it no longer relies on any db-specific implementations, and we're able to remove #6645's `UNION` entirely.

### DIFF
--- a/src/prefect/orion/database/configurations.py
+++ b/src/prefect/orion/database/configurations.py
@@ -210,7 +210,7 @@ class AioSqliteConfiguration(BaseDatabaseConfiguration):
             # use `named` paramstyle for sqlite instead of `qmark` in very rare
             # circumstances, we've seen aiosqlite pass parameters in the wrong
             # order; by using named parameters we avoid this issue
-            # see e.g. https://github.com/PrefectHQ/prefect/pull/6645
+            # see https://github.com/PrefectHQ/prefect/pull/6702
             kwargs["paramstyle"] = "named"
 
             # ensure a long-lasting pool is used with in-memory databases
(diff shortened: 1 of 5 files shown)


zangell44 (approved):
LGTM

src/prefect/orion/database/configurations.py
```diff
@@ -207,6 +207,12 @@ async def engine(self) -> AsyncEngine:
             if self.timeout is not None:
                 kwargs["connect_args"] = dict(timeout=self.timeout)
 
+            # use `named` paramstyle for sqlite instead of `qmark` in very rare
+            # circumstances, we've seen aiosqlite pass parameters in the wrong
+            # order; by using named parameters we avoid this issue
+            # see https://github.com/PrefectHQ/prefect/pull/6702
+            kwargs["paramstyle"] = "named"
```
chrisguidry, line 214: Knowledge :bow:

chrisguidry (approved):
So nice.  Would have loved to read it in the original SQL :P

PR #6714, opened 2022-09-05
Remove unnecessary SQL transactions

zangell44:
At the moment, all requests made against the Prefect 2.0 API explicitly begin and commit/rollback a SQL transaction. Many API requests don't need need to do so.

This PR refactors session management to allow route handlers to control when a session is created, when a connection is established, when transactions are committed, etc.

Unblocks the second part of https://github.com/PrefectHQ/nebula/issues/1470

### Example

Route handlers now rely on retrieving the database interface object directly in order to get a session via FastAPI dependency injection.

```
@router.post("/", status_code=status.HTTP_201_CREATED)
async def create_work_queue(
    work_queue: schemas.actions.WorkQueueCreate,
) -> schemas.core.WorkQueue:
    """
    Creates a new work queue.

    If a work queue with the same name already exists, an error
    will be raised.
    """
```

The route handler can decide how to manage its own session. The simplest action is getting a SQLAlchemy session from the database

```python
session = await.db.session()
# ... do stuff with session
```

I have also included a context manager to simplify the two most common operations 
- (1) creating a connection, querying, and closing the connection
- (2) all of (1) + opening a transaction and commit/rollback

```python
async with db.session_context() as session:
    # get a connection and make some queries

async with db.session_context(begin_transaction=True) as session:
    # make changes that will be committed
```

### Checklist


chrisguidry (approved):
This feels much more precise to me, and the overhead is very low with the new context managers.

One bit of feedback is that we might want to write some kind of guide for when to use each.  It's not as clear-cut as "use `transaction_context` for writes, and `session_context` for reads", because transactions also give you a consistent view of the database for _reading multiple objects_ as well.  I think I'd tell folks to use `transaction_context` if they have any doubts, and then optimize down to `session_context` when the know they are only making single reads?

madkinsz:
+1 to Chris' comments regarding the context to use. Seems good to cover in the docstring and perhaps as a short section in the contributing documentation as well.

src/prefect/orion/database/interface.py
```diff
@@ -79,11 +80,34 @@ async def engine(self):
...
     async def session(self):
         """
-        Provides a SQLAlchemy session
+        Provides a SQLAlchemy session.
         """
         engine = await self.engine()
         return await self.database_config.session(engine)
 
+    @asynccontextmanager
+    async def session_context(self):
```
jlowin, line 89: To resolve the two-context-manager confusion, perhaps
```suggestion
    async def session_context(self, begin_transaction:bool=False):
```

madkinsz: I kind of like that. Then it can just be `with db.session() as session:` or `with db.session(begin_transaction=True)` or `with_transaction=True`.

anticorrelator (approved):

src/prefect/orion/database/interface.py
zangell44: I like this, I've implemented it

src/prefect/orion/api/deployments.py
```diff
@@ -191,70 +198,72 @@ async def schedule_deployment(
...
+            session=session, deployment_id=deployment_id
         )
-    deployment.is_schedule_active = True
-    await session.flush()
+        if not deployment:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Deployment not found"
+            )
+        deployment.is_schedule_active = True
+        # await session.flush() # TODO - do we need this?
```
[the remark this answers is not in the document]
zangell44: I went through and audited all of the places we use `session.flush` and removed what we could.

Part of the need (maybe all of it?) is the way we use `models` functions in tests. However, I don't want to include that refactoring in this PR

src/prefect/orion/database/interface.py
madkinsz: Is there a case where you should use `session` without a context manager or should we just have `session` be a context and have something else for the lower-level session retrieval?

jlowin (approved):

PR #6720, opened 2022-09-06
Only load flow if necessary

cicdw:
I believe this may resolve the weird bug @znicholasbrown is seeing by avoiding a redundant load.  My theory is that objects loaded with our utility have non-standard attributes making this code path brittle.

I unfortunately don't have time to write a test for this right now, so if anyone else wants to take over this PR or I can follow up later (assuming it closes the issue).


madkinsz (approved):
Makes sense regardless of the bug

anna-geller (approved):
QAed with CLI and works now!

PR #6728, opened 2022-09-07
Make Netlify deploy previews resilient on forks

cicdw:
Currently every PR opened from a fork results in a Netlify notification asking to approve the deploy preview; because many of our open source PRs are documentation related, it is convenient to see these builds often.  Right now, we have sensitive environment variables for interacting with our MkDocs insiders repo so we don't want to use those for "untrusted" builds. 

This PR attempts to change our Netlify build command so that it first attempts to install our Insiders package, but if that fails it falls back to a standard MkDocs build (which should work).  We can then turn builds on for unknown authors that don't set sensitive environment variables.  This should hopefully let us see useful deploy previews at the cost of not knowing if our insider builds are failing (this should be readily apparent though).

### Example

Let's see how the deploy previews go in practice!

### Checklist

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,7 +12,7 @@ command = """
 pip install --upgrade pip \
 && pip install --upgrade --upgrade-strategy eager  -e ".[dev]" \
 && pip install git+https://${MKDOCS_INSIDERS_PAT}@github.com/PrefectHQ/mkdocs-material-insiders.git; \
-&& prefect dev build-docs  \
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
Seems worth adding a comment in the toml file explaining what's happening, otherwise seems good to me.

I did notice a while ago that you get direct notifications for each of those preview approvals and was curious how long that would last :D

Issue #6776, opened 2022-09-10
Using DaskExecutor raises RuntimeWarning: coroutine 'rpc.close_rpc' was never awaited


leonarduschen:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Getting a RuntimeWarning when using DaskExecutor

```
/home/leonarduschen/src/jd/data/venv/lib/python3.9/site-packages/prefect/executors/dask.py:314: RuntimeWarning: coroutine 'rpc.close_rpc' was never awaited
  scheduler_comm.close_rpc()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Also, I am getting a lot of zombie processes when running my flow (scheduled for every 5 mins), I suspect it is related to this warning.

### Reproduction

```python
# hello_world.py
from prefect import Flow, task
from prefect.executors import DaskExecutor


@task
def hello() -> None:
    print("Hello")


if __name__ == "__main__":
    with Flow("hello_world") as flow:
        hello()

    flow.executor = DaskExecutor(
        cluster_kwargs={"n_workers": 4, "threads_per_worker": 1},
        debug=True,
    )

    flow.run()
```

### Error

```
2022-09-10 09:52:20+0700 - INFO - prefect.FlowRunner[618577][MainThread] - (flow_runner.py).run - Beginning Flow run for 'hello_world'
2022-09-10 09:52:20+0700 - INFO - prefect.DaskExecutor[618577][MainThread] - (dask.py).start - Creating a new Dask cluster with `distributed.deploy.local.LocalCluster`...
2022-09-10 09:52:21+0700 - INFO - prefect.DaskExecutor[618577][MainThread] - (dask.py).start - The Dask dashboard is available at http://127.0.0.1:8787/status
/home/leonarduschen/src/jd/data/venv/lib/python3.9/site-packages/distributed/scheduler.py:4830: UserWarning: Scheduler already contains a plugin with name worker-status; overwriting.
  warnings.warn(
2022-09-10 09:52:22+0700 - INFO - prefect.TaskRunner[618625][Dask-Default-Threads-618625-0] - (task_runner.py).run - Task 'hello': Starting task run...
Hello
[...]

PR #6793, opened 2022-09-12
Fix bug where `scheduler_comm.close_rpc()` was not awaited

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6776


bunchesofdonald (approved):
Yep, that's the problem.

PR #6813, opened 2022-09-13
Use `copy_model_fields` utility for action schemas

madkinsz:
Pulls the `copy_model_fields` utility from  https://github.com/PrefectHQ/nebula/commit/9874ae454ce6fa64ad22291fb906584be937b617 and uses it for our actions models instead of the `pydantic_subclass` utility. This means action types no longer violate the Liskov substitution principle in which we were claiming they were subclasses of the parent type but we were excluding fields. We no longer use the subclass utility and it could be removed but I think it may be useful still.

I discovered several cases of undesirable and confusing behavior from the old pattern. I've done my best to match the old behavior but in some cases it was wrong and needed to be changed. This includes one "breaking" change to the set state routes for flow runs and task runs: Previously, these routes would raise a 422 if the client provided a timestamp. Now, they will ignore the timestamp and set the state. This now matches the behavior of flow run and task run creation.

I renamed the associated `From` utility to `FieldFrom` as I think it makes its usage more intuitive. After this is merged, I can open a pull request to Nebula to use the utilities from the core library instead.

This utility will be useful in upcoming work on result persistence where we will need to introduce client-side copies of server-side models with extended capabilities.

Requires #6829 

### Checklist

### DIFF
--- a/src/prefect/orion/schemas/actions.py
+++ b/src/prefect/orion/schemas/actions.py
@@ -61,21 +61,26 @@ class DeploymentCreate(PrefectBaseModel):
     """Data used by the Orion API to create a deployment."""
 
     name: str = FieldFrom(schemas.core.Deployment)
-    version: Optional[str] = FieldFrom(schemas.core.Deployment)
-    description: str = FieldFrom(schemas.core.Deployment)
     flow_id: UUID = FieldFrom(schemas.core.Deployment)
-    schedule: schemas.schedules.SCHEDULE_TYPES = FieldFrom(schemas.core.Deployment)
     is_schedule_active: bool = FieldFrom(schemas.core.Deployment)
-    infra_overrides: Dict[str, Any] = FieldFrom(schemas.core.Deployment)
     parameters: Dict[str, Any] = FieldFrom(schemas.core.Deployment)
     tags: List[str] = FieldFrom(schemas.core.Deployment)
+
+    manifest_path: Optional[str] = FieldFrom(schemas.core.Deployment)
     work_queue_name: Optional[str] = FieldFrom(schemas.core.Deployment)
-    parameter_openapi_schema: Dict[str, Any] = FieldFrom(schemas.core.Deployment)
-    path: str = FieldFrom(schemas.core.Deployment)
-    entrypoint: str = FieldFrom(schemas.core.Deployment)
-    manifest_path: str = FieldFrom(schemas.core.Deployment)
     storage_document_id: Optional[UUID] = FieldFrom(schemas.core.Deployment)
     infrastructure_document_id: Optional[UUID] = FieldFrom(schemas.core.Deployment)
+    schedule: Optional[schemas.schedules.SCHEDULE_TYPES] = FieldFrom(
+        schemas.core.Deployment
+    )
+    description: Optional[str] = FieldFrom(schemas.core.Deployment)
+    parameter_openapi_schema: Optional[Dict[str, Any]] = FieldFrom(
+        schemas.core.Deployment
+    )
+    path: Optional[str] = FieldFrom(schemas.core.Deployment)
+    version: Optional[str] = FieldFrom(schemas.core.Deployment)
+    entrypoint: Optional[str] = FieldFrom(schemas.core.Deployment)
+    infra_overrides: Optional[Dict[str, Any]] = FieldFrom(schemas.core.Deployment)
 
     class Config:
         extra = "forbid"
@@ -184,7 +189,7 @@ class FlowRunCreate(PrefectBaseModel):
     infrastructure_document_id: Optional[UUID] = FieldFrom(schemas.core.FlowRun)
     empirical_policy: schemas.core.FlowRunPolicy = FieldFrom(schemas.core.FlowRun)
     tags: List[str] = FieldFrom(schemas.core.FlowRun)
-    idempotency_key: str = FieldFrom(schemas.core.FlowRun)
+    idempotency_key: Optional[str] = FieldFrom(schemas.core.FlowRun)
 
 
 @copy_model_fields
@@ -205,7 +210,7 @@ class DeploymentFlowRunCreate(PrefectBaseModel):
     infrastructure_document_id: Optional[UUID] = FieldFrom(schemas.core.FlowRun)
     empirical_policy: schemas.core.FlowRunPolicy = FieldFrom(schemas.core.FlowRun)
     tags: List[str] = FieldFrom(schemas.core.FlowRun)
-    idempotency_key: str = FieldFrom(schemas.core.FlowRun)
+    idempotency_key: Optional[str] = FieldFrom(schemas.core.FlowRun)
 
 
 @copy_model_fields
@@ -293,7 +298,6 @@ class BlockDocumentCreate(PrefectBaseModel):
     class Config:
         extra = "forbid"
 
-    # validators
     _validate_name_format = validator("name", allow_reuse=True)(
         validate_block_document_name
     )
--- a/src/prefect/orion/schemas/actions.py
+++ b/src/prefect/orion/schemas/actions.py
@@ -143,7 +143,9 @@ class StateCreate(ActionBaseModel):
     # DEPRECATED
 
     timestamp: Optional[schemas.core.DateTimeTZ] = Field(
-        default=None, repr=False, ignored=True
+        default=None,
+        repr=False,
+        ignored=True,
     )
     id: Optional[UUID] = Field(default=None, repr=False, ignored=True)
 
--- a/src/prefect/orion/schemas/actions.py
+++ b/src/prefect/orion/schemas/actions.py
@@ -5,7 +5,7 @@ import re
 from typing import Any, Dict, List, Optional, Union
 from uuid import UUID
 
-from pydantic import Field, validator
+from pydantic import Field, root_validator, validator
 
 import prefect.orion.schemas as schemas
 from prefect.orion.utilities.schemas import (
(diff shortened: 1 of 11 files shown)


src/prefect/orion/schemas/actions.py
```diff
@@ -31,159 +35,133 @@ def validate_block_document_name(value):
...
+    type: schemas.states.StateType = FieldFrom(schemas.states.State)
+    name: str = FieldFrom(schemas.states.State)
+    timestamp: schemas.core.DateTimeTZ = FieldFrom(schemas.states.State)
+    message: str = FieldFrom(schemas.states.State)
+    data: schemas.data.DataDocument = FieldFrom(schemas.states.State)
+    state_details: schemas.states.StateDetails = FieldFrom(schemas.states.State)
+
     class Config:
-        extra = "forbid"
+        extra = "ignore"
```
madkinsz, line 132: This is the only real change here. This is necessary to support `StateCreate` in action models. Without this, I got errors that `StateCreate` does not accept ids. The client is just sending normal `State` objects. I believe that previously we actually were not actually overriding the `state` field type where we thought we were.

src/prefect/orion/schemas/actions.py
```diff
@@ -2,13 +2,17 @@
...
-from typing import List, Optional
+from typing import Any, Dict, List, Optional, Union
 from uuid import UUID
 
 from pydantic import Field, validator
 
 import prefect.orion.schemas as schemas
-from prefect.orion.utilities.schemas import PrefectBaseModel
+from prefect.orion.utilities.schemas import (
+    FieldFrom,
```
chrisguidry, line 12: I like that, `FieldFrom`, looks tight!

src/prefect/orion/schemas/actions.py
```diff
@@ -31,159 +35,133 @@ def validate_block_document_name(value):
...
-            "infrastructure_document_id",
-            "infra_overrides",
-        ],
-    )
-):
+@copy_model_fields
+class DeploymentCreate(PrefectBaseModel):
     """Data used by the Orion API to create a deployment."""
 
+    name: str = FieldFrom(schemas.core.Deployment)
```
chrisguidry, line 63: I find there's a lot of visual noise from using the extra module `schemas.` in every field definition.  What do you think about `from prefect.orion.schemas import core, states` and having these all be `core.Deployment`, `core.FlowRun`, `states.State`, etc?

chrisguidry (approved):
Nice!  Looking forward to the backport

zangell44 (approved):

src/prefect/orion/schemas/actions.py
madkinsz: I'd be willing to do that or just import the objects directly.

madkinsz:
Failing
```
FAILED tests/orion/api/test_block_documents.py::TestCreateBlockDocument::test_create_block_document_with_invalid_characters[my block]
FAILED tests/orion/api/test_block_documents.py::TestCreateBlockDocument::test_create_block_document_with_invalid_characters[my\\block]
FAILED tests/orion/api/test_block_documents.py::TestCreateBlockDocument::test_create_block_document_with_invalid_characters[my\U0001f44dblock]
FAILED tests/orion/api/test_block_documents.py::TestCreateBlockDocument::test_create_block_document_with_invalid_characters[my|block]
FAILED tests/orion/api/test_block_types.py::TestCreateBlockType::test_create_block_type_with_invalid_slug_fails
FAILED tests/orion/api/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_policy_states
FAILED tests/orion/api/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_policy_active
FAILED tests/orion/api/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_policy_message_template
FAILED tests/orion/api/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_missing_policy
FAILED tests/orion/api/test_flow_runs.py::TestCreateFlowRun::test_create_flow_run_with_state_sets_timestamp_on_server
FAILED tests/orion/api/test_flow_runs.py::TestSetFlowRunState::test_set_flow_run_errors_if_client_provides_timestamp
FAILED tests/orion/api/test_logs.py::TestCreateLogs::test_create_logs_with_flow_run_id_and_returns_number_created
FAILED tests/orion/api/test_logs.py::TestCreateLogs::test_create_logs_with_task_run_id_and_returns_number_created
FAILED tests/orion/api/test_task_runs.py::TestCreateTaskRun::test_create_task_run_with_state_sets_timestamp_on_server
FAILED tests/orion/api/test_task_runs.py::TestSetTaskRunState::test_set_task_run_errors_if_client_provides_timestamp
FAILED tests/orion/models/test_block_documents.py::TestCreateBlockDocument::test_create_block_document_errors_if_no_name_provided
FAILED tests/orion/models/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_policy_states
FAILED tests/orion/models/test_flow_run_notification_policies.py::TestUpdateFlowRunNotificationPolicy::test_update_policy_active
FAILED tests/orion/models/test_flow_run_notification_policies.py::TestQueueNotificationsFromPolicy::test_inactive_notifications_not_queued
FAILED tests/orion/models/test_logs.py::TestCreateLogs::test_create_logs_succeeds
FAILED tests/orion/services/test_flow_run_notifications.py::test_service_uses_message_template
ERROR tests/orion/api/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_policies
ERROR tests/orion/api/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_active_policies
ERROR tests/orion/api/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_inactive_policies
ERROR tests/orion/models/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_policies
ERROR tests/orion/models/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_active_policies
ERROR tests/orion/models/test_flow_run_notification_policies.py::TestReadFlowRunNotificationPolicies::test_read_inactive_policies
```

Likely I've just missed some fields or need to update some to be `Optional` in update payloads.

madkinsz:
Requires https://github.com/PrefectHQ/prefect/pull/6829; please do not merge until after it.

src/prefect/orion/schemas/actions.py
chrisguidry: Thoughts about shortening these?  It could also be done in a follow-up PR

src/prefect/orion/utilities/schemas.py
```diff
@@ -347,3 +348,97 @@ class Config:
...
+
+def copy_model_fields(model_class: Type[B]) -> Type[B]:
+    """
+    A class decorator which copies field definitions and field validators from other
+    Pydantic BaseModel classes.  This does _not_ make the model a subclass of any of
+    the copied field's owning classes, nor does this copy root validators from any of
+    those classes.  Note that you should still include the type hint for the field in
+    order to make typing explicit.
+
+    Use this decorator and the corresponding `From` field to compose response and
```
chrisguidry, line 376:
```suggestion
    Use this decorator and the corresponding `FieldFrom` to compose response and
```

src/prefect/orion/utilities/schemas.py
```diff
@@ -347,3 +348,97 @@ class Config:
...
+        >>> from prefect.orion.utilities.schemas import copy_model_fields, From
+        >>>
+        >>> class Parent(BaseModel):
+        ...     name: str
+        ...     sensitive: str
+        >>>
+        >>> @copy_model_fields
+        >>> class Derived(BaseModel):
+        ...     name: str = From(Parent)
+        ...     my_own: str
```
chrisguidry, line 391:
```suggestion
        >>> from prefect.orion.utilities.schemas import copy_model_fields, FieldFrom
        >>>
        >>> class Parent(BaseModel):
        ...     name: str
        ...     sensitive: str
        >>>
        >>> @copy_model_fields
        >>> class Derived(BaseModel):
        ...     name: str = FieldFrom(Parent)
        ...     my_own: str
```

tests/orion/api/test_task_runs.py
```diff
@@ -77,7 +77,11 @@ async def test_create_task_run_with_state_sets_timestamp_on_server(
             json=schemas.actions.TaskRunCreate(
                 flow_run_id=flow_run.id,
                 task_key="a",
-                state=states.Completed(timestamp=pendulum.now().add(months=1)),
+                state=schemas.actions.StateCreate(
+                    type=schemas.states.StateType.COMPLETED,
+                    # This is no longer allowed
+                    # timestamp=pendulum.now().add(months=1),
```
chrisguidry, line 83: I just realized this test doesn't exactly test what it sounds like.  I think the assertion at the end should be:

```
start_of_test = pendulum.now()

... do the POST ...

assert start_of_test <= task_run.state.timestamp <= pendulum.now()
```

Right?

chrisguidry (approved):

src/prefect/orion/schemas/actions.py
madkinsz: I didn't make this change because we're using `schemas.` frequently in this library — it's kind of our standard pattern. I'd like to address this later perhaps more holistically but for now I want to be consistent.

tests/orion/api/test_task_runs.py
madkinsz: Ah that was a holdover from when I was going to ban timestamps but I realized I had to keep them for backwards compatibility. I've updated this test to include the timestamp again and changed the name for clarity.

We're sending a timestamp with +1 month so we can do a simpler check. We could include a lower bound but I'm not sure what it adds?

madkinsz: Let me know if you have additional concerns; this is how the tests were structured before so I'll leave it as is here.

PR #6826, opened 2022-09-14
UI: Provide new api service for orion-design

pleek91:
# Description
Part of a multi step effort to replace our current ui api services with a much simpler and more flexible approach. 

Related to https://github.com/PrefectHQ/orion-design/pull/545

### DIFF
--- a/orion-ui/src/App.vue
+++ b/orion-ui/src/App.vue
@@ -12,7 +12,7 @@
     </template>
     <ContextSidebar v-if="showMenu" class="app__sidebar" @click="close" />
     <suspense>
-      <RouterView />
+      <AppRouterView />
     </suspense>
   </div>
 </template>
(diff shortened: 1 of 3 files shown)


orion-ui/src/App.vue
```diff
@@ -12,13 +12,7 @@
...
     <ContextSidebar v-if="showMenu" class="app__sidebar" @click="close" />
     <suspense>
-      <router-view class="app__router-view">
-        <template #default="{ Component }">
-          <transition name="app__router-view-fade" mode="out-in">
-            <component :is="Component" />
-          </transition>
-        </template>
-      </router-view>
+      <RouterView />
```
stackoverfloweth, line 15: I like this change but I'd use a different name. It's too easy to assume that's the same router view component provided by vue-router

pleek91: Open to alternatives, any suggestions?

stackoverfloweth: AppRouterView, RootRouterView, ProvideRouterView?

pleek91: Went with `AppRouterView`

stackoverfloweth (approved):

PR #6829, opened 2022-09-14
Update model fields for consistent usage of `Optional` and default

madkinsz:
Makes two readability and consistency changes to our models:
- Use `Optional` for types when field defaults are `None`
   - Avoids relying on implicit optional behavior
   - Avoids confusion over unvalidated default values of `None`
- Always use `default=` explicitly
   - Static typing for default values will not display correctly otherwise, see
      https://github.com/pydantic/pydantic/issues/3753
      https://github.com/microsoft/pyright/blob/main/specs/dataclass_transforms.md#field-descriptors
   - Explicit is better than implicit
  
There should be no change in behavior here. This was performed with some crazy regular expressions then manual edits for edge cases. Please give it a second look.

The lack of explicit `Optional` types lead to issues in #6813 — this change should unblock that.

### Example

The following:
```python
class Model(BaseModel):
   foo: str = Field(None)
```

has changed to:
```python
class Model(BaseModel):
   foo: Optional[str] = Field(default=None)
```
### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Add workflow features like retries, distributed execution, scheduling, caching, 
 
 ```python
 from prefect import flow, task
-from typing import List
+from typing import Optional, List
 import httpx
 
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Add workflow features like retries, distributed execution, scheduling, caching, 
 
 ```python
 from prefect import flow, task
-from typing import Optional, List
+from typing import List
 import httpx
 
 
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -7,7 +7,7 @@ Example:
 
     >>> from prefect import flow, task
     >>> from prefect.task_runners import SequentialTaskRunner
-    >>> from typing import List
+    >>> from typing import Optional, List
     >>>
     >>> @task
     >>> def say_hello(name):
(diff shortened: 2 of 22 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Add workflow features like retries, distributed execution, scheduling, caching,
 
 ```python
-from typing import List
+from typing import Optional, List
```
bunchesofdonald, line 43: It doesn't look like `Optional` is used in this code block.

```suggestion
from typing import List
```

src/prefect/task_runners.py
```diff
@@ -7,7 +7,7 @@
 
     >>> from prefect import flow, task
     >>> from prefect.task_runners import SequentialTaskRunner
-    >>> from typing import List
+    >>> from typing import Optional, List
```
bunchesofdonald, line 10:
```suggestion
    >>> from typing import List
```

bunchesofdonald (approved):
Looks good! I was a bit cross-eyed by the end, but all I found a couple places where `Optional` didn't need to be imported.

README.md
madkinsz, line 43:
```suggestion
from typing import List
```

chrisguidry (approved):
This makes a lot of sense to me, and the only change we might need to check in on with Prefect Cloud is the `ORMBaseModel`.  I'm not expecting any trouble, what you've done here is totally compatible and just making the implicit explicit (thank you!)

Issue #6831, opened 2022-09-14
"Rate Limit Encountered" logged 680k times even when Diagnostic Logger is turned off


robfreedy:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When rate limits are encountered, the log statement get repeatedly logged, both when the diagnostic logger is on and off. 
This is the piece of code that the logger pertains to: 
https://github.com/PrefectHQ/prefect/blob/4c363a2cde15d411bace54b8bb3d315c100f3807/src/prefect/client/client.py#L541

Raised by John from Recursion: 
 I’m seeing 680k log statements of Rate limit encountered and why e.g. 3300 of them are sleeping for 192.2335378385163s(the same random jitter) even on the 3rd attempt all occurring within a few microseconds of each other.

### Reproduction

In order to reproduce, a 429 error code (rate limit error code) has to be raised on a graphQL request. Once a 429 is hit, then the statement will be logged over and over again.

### Error

```

[2022-08-08 16:34:05+0000] WARNING - Diagnostics | Rate limit encountered (attempt 1); sleeping for 187.4598413938345s...

```

### Versions

```

Prefect 1.0 

```

PR #6832, opened 2022-09-14
Use a singleton diagnostics logger for `Client`

madkinsz:
Each client "creates" a new logger which includes attaching handlers. Since Python loggers are singletons behind the scenes, this results in attaching duplicate handlers to the logger. Messages will then be duplicated per number of clients. Since task runs each have their own client, this can result in many log messages.

To resolve this, we define `logger` as a class variable instead of instance variable. The logger will be created once at import-time.

Closes https://github.com/PrefectHQ/prefect/issues/6831

### DIFF
--- a/src/prefect/client/client.py
+++ b/src/prefect/client/client.py
@@ -95,6 +95,8 @@ class Client:
             `prefect auth` CLI
     """
 
+    logger = create_diagnostic_logger("client.diagnostics")
+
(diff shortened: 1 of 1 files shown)


chrisguidry (approved):
Great catch!

johnurbanik:
> Great catch!

Thank you!

PR #6833, opened 2022-09-14
Respect PREFECT__CLOUD__SEND_FLOW_RUN_LOGS

### LINKED ISSUE #5521: Allow disabling sending flow run logs on a per flow run basis via PREFECT__CLOUD__SEND_FLOW_RUN_LOGS environment variable

**david.beck**: Hi Prefect Team! I'm looking to disable cloud logging since our implementation will be routing our logs via k8s directly to splunk. I tried setting an env value of `PREFECT__CLOUD__SEND_FLOW_RUN_LOGS=False` in our global container config which I see being recognized by Prefect Cloud, however there are logs still populating for the run (see attached). Do I perhaps have the wrong value defined or could I be missing something?

**kevin701**: That looks right so confused why it’s behaving like that. I can ask the team but you can also try the `--no-cloud-logs` on the agent when you start per <https://docs.prefect.io/core/idioms/logging.html#logging-with-a-backend|this>

**david.beck**: <@U01QEJ9PP53> okay I thought I set that up correctly. Please let me know what the team says. I have to look into our CI/CD process to see if/how we can include that run flag. We are submitting scheduled flow runs, so I'd have to be sure that they are receiving the `--no-cloud-logs` flag

**michael054**: Hey David, the agent overrides this environment variable value with its own setting.

**michael054**: I’d accept a pull request to change this behavior to allow override as you’ve done. This isn’t intended behavior.

**anna**: <@ULVA73B9P> open "Allow disabling sending flow run logs on a per flow run basis via PREFECT__CLOUD__SEND_FLOW_RUN_LOGS environment variable"

Original thread can be found [here](https://prefect-community.slack.com/archives/C0192RWGJQH/p1646778174513709?thread_ts=1646778174.513709&cid=C0192RWGJQH).

EmilRex:
The `PREFECT__CLOUD__SEND_FLOW_RUN_LOGS` environment variables is ignored when explicitly set. This change avoids overwriting it with a default value when set.

Closes #5521

### DIFF
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -562,7 +562,8 @@ class KubernetesAgent(Agent):
         env.update(self.env_vars)
         if run_config.env:
             env.update(run_config.env)
-        if env.get("PREFECT__CLOUD__SEND_FLOW_RUN_LOGS", "").lower() != "false":
+        # Allow logs to be disabled by flow run settings but not enabled
+        if not self.log_to_cloud or "PREFECT__CLOUD__SEND_FLOW_RUN_LOGS" not in env:
             env["PREFECT__CLOUD__SEND_FLOW_RUN_LOGS"] = str(self.log_to_cloud).lower()
         env.update(
             {
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -562,6 +562,8 @@ class KubernetesAgent(Agent):
         env.update(self.env_vars)
         if run_config.env:
             env.update(run_config.env)
+        if "PREFECT__CLOUD__SEND_FLOW_RUN_LOGS" not in env:
+            env["PREFECT__CLOUD__SEND_FLOW_RUN_LOGS"] = str(self.log_to_cloud).lower()
         env.update(
             {
                 "PREFECT__BACKEND": config.backend,
@@ -575,7 +577,6 @@ class KubernetesAgent(Agent):
                     else ""
                 ),
                 "PREFECT__CLOUD__USE_LOCAL_SECRETS": "false",
-                "PREFECT__CLOUD__SEND_FLOW_RUN_LOGS": str(self.log_to_cloud).lower(),
                 "PREFECT__CONTEXT__FLOW_RUN_ID": flow_run.id,
                 "PREFECT__CONTEXT__FLOW_ID": flow_run.flow.id,
                 "PREFECT__CONTEXT__IMAGE": image,
--- a/src/prefect/agent/kubernetes/agent.py
+++ b/src/prefect/agent/kubernetes/agent.py
@@ -562,7 +562,7 @@ class KubernetesAgent(Agent):
         env.update(self.env_vars)
         if run_config.env:
             env.update(run_config.env)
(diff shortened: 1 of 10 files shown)


madkinsz:
Hm. Looking back at https://github.com/PrefectHQ/prefect/pull/4487 I do not think this setting is _intended_ to be used this way. The engineer running the agent determines if flow run logs can be sent to Cloud. If I configure my agent not to send logs to Cloud, I would be pretty upset if any user could send logs by adding an environment variable.

Perhaps we should allow this to be turned _off_ per flow run but never turned _on_ unless allowed at the agent level. We'll want to check that this meets Recursion's requirements.

As is, I think this could be seen as a security risk.

EmilRex:
> I would be pretty upset if any user could send logs by adding an environment variable.
Good point, I hadn't considered that.

> Perhaps we should allow this to be turned off per flow run but never turned on unless already allowed at the agent level. We'll want to check that this meets Recursion's requirements.
Agreed.

johnurbanik:
Yes, I think that agent having flow logs off should take precedence, but if the agent has flow logs on, it should be possible to turn it off at the flow run level.

Updating the docs to communicate that clearly once implemented would also be great.

madkinsz:
> Yes, I think that agent having flow logs off should take precedence, but if the agent has flow logs on, it should be possible to turn it off at the flow run level.

Great! Definitely down for that implementation. @EmilRex you'll also need to implement this for all of the other agent types.

src/prefect/agent/kubernetes/agent.py
```diff
@@ -562,6 +562,8 @@ def generate_job_spec(self, flow_run: GraphQLResult) -> dict:
         env.update(self.env_vars)
         if run_config.env:
             env.update(run_config.env)
+        if env.get("PREFECT__CLOUD__SEND_FLOW_RUN_LOGS", "").lower() != "false":
+            env["PREFECT__CLOUD__SEND_FLOW_RUN_LOGS"] = str(self.log_to_cloud).lower()
```
madkinsz, line 566: I think what we want here is:
```suggestion
        # Allow logs to be disabled by flow run settings but not enabled
        if not self.log_to_cloud or "PREFECT__CLOUD__SEND_FLOW_RUN_LOGS" not in env:
           env["PREFECT__CLOUD__SEND_FLOW_RUN_LOGS"] = str(self.log_to_cloud).lower()
```

EmilRex: Gotcha

madkinsz (changes requested):
Can you write tests covering our new expected behavior as well?

tests/agent/test_docker_agent.py
```diff
@@ -222,6 +222,37 @@ def test_populate_env_vars_sets_log_to_cloud(flag, api, config_with_api_key):
     assert env_vars["PREFECT__LOGGING__LOG_TO_CLOUD"] == str(not flag).lower()
 
 
+@pytest.mark.parametrize(
```
EmilRex, line 225: @madkinsz Before I implement for the rest of the agents, is this what you had in mind?

madkinsz: I'd rename them for clarity to: `disabled_on_agent,enabled_by_run,expected`

Otherwise yeah this lgtm,

PR #6852, opened 2022-09-15
Prevent redundant transitions

anticorrelator:
Here we add a simple orchestration rule preventing redundant transitions. Among other things, this will prevent multiple Agents from orchestrating an active run in competing and inconsistent ways.

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1169,22 +1169,15 @@ async def orchestrate_task_run(
         # Resolve futures in any non-data dependencies to ensure they are ready
         await resolve_inputs(wait_for, return_data=False)
     except UpstreamTaskError as upstream_exc:
-        if task_run.state.is_pending():
+
+        return await propose_state(
+            client,
+            Pending(name="NotReady", message=str(upstream_exc)),
+            task_run_id=task_run.id,
             # if orchestrating a run already in a pending state, force orchestration to
             # update the state name
-            return await propose_state(
-                client,
-                Pending(name="NotReady", message=str(upstream_exc)),
-                task_run_id=task_run.id,
-                force=True,
-            )
-        else:
-            return await propose_state(
-                client,
-                Pending(name="NotReady", message=str(upstream_exc)),
-                task_run_id=task_run.id,
-                force=False,
-            )
+            force=task_run.state.is_pending(),
+        )
 
     # Generate the cache key to attach to proposed states
     cache_key = (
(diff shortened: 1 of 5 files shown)


src/prefect/orion/orchestration/core_policy.py
```diff
@@ -407,3 +410,40 @@ async def before_transition(
...
+        context: OrchestrationContext,
+    ) -> None:
+        initial_state_type = initial_state.type if initial_state else None
+        proposed_state_type = proposed_state.type if proposed_state else None
+        if (
+            self.STATE_PROGRESS[proposed_state_type]
+            <= self.STATE_PROGRESS[initial_state_type]
+        ):
+            await self.abort_transition(
+                reason=f"This run cannot to the {proposed_state.type} state."
```
jlowin, line 448: I think a word is missing here - "transition"?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -407,3 +410,40 @@ async def before_transition(
...
+    """
+    Prevents redundant transitions.
+
+    Under normal operation, this rule prevents the "backwards" progress of a run. This
+    rule will also help prevent multiple agents from attempting to orchestrate a run by
+    preventing transitions into the same state type. If any of these disallowed
+    transitions are attempted, this rule will abort the transition.
+    """
+
+    STATE_PROGRESS = {
```
jlowin, line 425: This is really neat

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -34,6 +35,7 @@ class CoreFlowPolicy(BaseOrchestrationPolicy):
     def priority():
         return [
             PreventTransitionsFromTerminalStates,
+            PreventBackwardsTransitions,
```
jlowin, line 38: Where is this class? Maybe a rename?

jlowin (changes requested):
Great stuff @anticorrelator and a win for the policy design to see how compact it is!

I don't think it's critical but I'd love to see an API-level unit test that proves this from a user's point of view (rather than the engine's point of view) by calling `set_flow_run_state(RUNNING)` twice in a row.

src/prefect/orion/orchestration/core_policy.py
anticorrelator: oops! yes, this was a rename

madkinsz:
Just wondering, should we use REJECT and provide the current state allowing the client to decide if it should retry or give up? Or is this always going to be a case where the run is already being handled elsewhere and consequently we should force an ABORT?

anticorrelator:
I believe that if these transitions should never be allowed, the only reason why a client might attempt them is if we're in a bad state; I prefer aborting here instead of trying to solve for the cases where we might be able to recover

src/prefect/engine.py
```diff
@@ -1169,11 +1169,22 @@ async def orchestrate_task_run(
...
-            task_run_id=task_run.id,
-        )
+        if task_run.state.is_pending():
+            # if orchestrating a run already in a pending state, force orchestration to
+            # update the state name
+            return await propose_state(
+                client,
+                Pending(name="NotReady", message=str(upstream_exc)),
+                task_run_id=task_run.id,
+                force=True,
```
madkinsz, line 1179: Can we just do `force=task_run.state.is_pending()` with that comment for clarity?

jlowin:
I agree, REJECT means "you asked to transition to X but you transitioned to Y instead" which isn't the semantic here

jlowin (approved):

PR #6847, opened 2022-09-15
Refactor client into new namespaces

madkinsz:
In preparation for client-side schemas, introduces a `prefect.client` namespace.

- Moves `CloudClient` from `prefect.cli.cloud` to `prefect.client.cloud`
- Moves `PrefectHttpxClient`, `PrefectResponse`, and similar to `prefect.client.base`
- Moves `OrionClient` to `prefect.client.orion`

Imports of `prefect.client.get_client` and `prefect.client.OrionClient` are backwards compatible. Other objects previously in the module are not available without specifying `prefect.client.base` or `prefect.client.orion`.

No changes in behavior have been made.

### Example

```
client
├── __init__.py
├── base.py
├── cloud.py
└── orion.py
```

### Checklist

- ~[ ] This pull request references any related issue by including "closes `<link to issue>`"~
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a `fix`, `feature`, `enhancement`, `docs`, `collections`, or `maintenance` label categorizing the change.

PR #6858, opened 2022-09-16
Add toggle for block registration on application start

madkinsz:
Adds the setting `PREFECT_ORION_BLOCKS_REGISTER_ON_START` which is used to determine if blocks should be registered as a server startup step. This may also solve issues being worked on in #6841.

This setting defaults to `True`, but I could see the argument for defaulting it to `False` and then setting it to `True` during `prefect orion start`. This would be a significant improvement to ephemeral client performance, but #6841 could accomplish something similar without changing the default.

Automatic registration of blocks was creating integrity errors during our test seeding process as we seed type models that happened to match a block type that was used in another fixture and auto-registered when a client was started. Previously, we implemented a workaround for this but additional issues were encountered indicating that our workaround was not robust. 

Now that we have a toggle, we can disable registration on application startup during test runs. This resolves the integrity errors and, as an added bonus, appears to improve test suite runtimes significantly:
- Py3.8 on Postgres went from 12m -> 7m: **40%** faster
- Py3.8 on SQLite went from 24m -> 9m: **62%** faster

This is part 2/2 resolving issues with blocks failing tests in https://github.com/PrefectHQ/prefect/pull/6847

### Checklist

### DIFF
--- a/src/prefect/orion/api/server.py
+++ b/src/prefect/orion/api/server.py
@@ -272,6 +272,9 @@ def create_app(
 
     async def add_block_types():
         """Add all registered blocks to the database"""
+        if not prefect.settings.PREFECT_ORION_BLOCKS_REGISTER_ON_START:
+            return
+
         from prefect.blocks.core import Block
         from prefect.orion.database.dependencies import provide_database_interface
         from prefect.orion.models.block_schemas import create_block_schema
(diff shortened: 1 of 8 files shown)


madkinsz:
Regarding the test suite timing reductions: This suggests to me that there is some low-hanging fruit for improving performance in our test suite and for our users. Additionally, it seems as though block registration is expensive and we may want to consider optimizing those routes. The massive improvement for SQLite also indicates that block registration was particularly slow for SQLite and we may want to optimize the queries used there. cc @abrookins

madkinsz:
@abrookins do you hear me now? ;)

zangell44 (approved):

chrisguidry (approved):
Looks good, and you had me at 40% reduction in the test runtime :P

PR #6857, opened 2022-09-16
Inject `OrionClient` during block registration, save, and load

madkinsz:
Updates `Block.save`, `Block.load`, and `Block.register_types_and_schema` to use `@inject_client` which:
- Simplifies implementation and reduces nesting for readability
- Allows users to provide clients to `save` and `load` to avoid extra client startup
- Allows us to perform future optimizations for client reuse by updating `inject_client`

### Example

```python
async with get_client() as client:
    # Use client for other things
    ....
    Block.load("test", client=client)

   # Previously, `client` could not be passed and it would always create a new one
```

### Checklist

- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a `fix`, `feature`, `enhancement`, `docs`, `collections`, or `maintenance` label categorizing the change.

PR #6859, opened 2022-09-16
Optimize client retrieval using flow and task run contexts

madkinsz:
Requires #6847 to avoid merge conflicts.

Updates `inject_client` to use an existing client from a flow run or task run context if present. This will reduce the number of clients that we instantiate (which can be expensive due to lifetime management) automatically when users call methods requiring a client. Combined with #6857 this should dramatically improve performance of block save and load operations when used in flows or tasks.

### Example

 ```python
@inject_client
@sync_compatible
async def foo(client: OrionClient):
    # Use client!
    ...


foo()  # Sets up and tears down new client

with FlowRunContext(...):
   # Previously, the following would still set up and tear down a new client
   # Now, it will use the client from the `FlowRunContext`
   foo()
```

### Checklist


madkinsz:
I'd like to add some brief tests for this as well.

bunchesofdonald (approved):
Makes sense to me!

PR #6862, opened 2022-09-16
Drop cache upload for Conda flavor Docker images

madkinsz:
Conda flavored Docker images are taking 8+ minutes to build compared to the base image build of 2 minutes. The build time is dominated by ~7 minutes of uploading image layers to a cache. Here, we disable the caching as it cannot possible be worth the runtime.


chrisguidry (approved):
Heck yeah!

image

PR #6864, opened 2022-09-17
Return a better exception when decorator is associated to a method

jcfaracco:
There is a huge consensus that decorators are being used for functions due to the ability to create stateless procedures. Classes have state and according to several discussions methods cannot be turned into tasks. Then, it should be properly captured and raised by the flow.

Signed-off-by: Julio Faracco <jcfaracco@gmail.com>

### Example
```python
class Tasks:
    @task
    def task1(self, in):
        return in

    @task
    def task2(self, in1, in2):
        return in1, in2
```
The code above is not technically allowed and when we run a pipeline, the error is not trivial. The fix is a small one just to throw a better and clear exception.

### Checklist

- [ ] This pull request includes a `fix`, `feature`, `enhancement`, `docs`, `collections`, or `maintenance` label categorizing the change.

### DIFF
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -27,7 +27,10 @@ def get_call_parameters(
     Raises a ParameterBindError if the arguments/kwargs are not valid for the function
     """
     try:
-        bound_signature = inspect.signature(fn).bind(*call_args, **call_kwargs)
+        signature_fn = inspect.signature(fn)
+        if "self" in signature_fn.parameters:
+            raise TypeError(f"Decorator does not support class methods")
+        bound_signature = signature_fn.bind(*call_args, **call_kwargs)
     except TypeError as exc:
         raise ParameterBindError.from_bind_failure(fn, exc, call_args, call_kwargs)
     bound_signature.apply_defaults()


src/prefect/utilities/callables.py
```diff
@@ -27,7 +27,10 @@ def get_call_parameters(
     Raises a ParameterBindError if the arguments/kwargs are not valid for the function
     """
     try:
-        bound_signature = inspect.signature(fn).bind(*call_args, **call_kwargs)
+        signature_fn = inspect.signature(fn)
+        if "self" in signature_fn.parameters:
```
madkinsz, line 31: We'll want to do a more robust check. Probably `inspect.ismethod`.

I think we'll also want to move enforcement for this to the `Task.__init__` method instead. That way, a user will get an error when they define the task rather than at runtime.

madkinsz (commented):
Hi! Thanks for contributing. I think this makes sense, but I've suggested a different implementation and I'd like to rope in our product team.

Can you share an example of the error you get so we can understand the current user experience?

madkinsz:
Closing as stale. Feel free to reopen!

Issue #6821, opened 2022-09-14
Prefect Orion fails to start on Windows when there is a space in Prefect package's path


rpeden:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If the Prefect package is installed on Windows in a directory containing a space in the pathname, Prefect Orion fails to start because it cannot launch Uvicorn. 

The problem happens when we set the `--app-dir` flag just before launching Uvicorn.

### Reproduction
* Run cmd or Powershell on Windows
* Create a virtual environment in a directory containing a space character in the path. 
* Activate the virtual environment. 
* Install Prefect.
* Run `prefect orion start`

You can also observe the problem directly by running something like:

```python
uvicorn --app-dir c:\my python site packages\prefect --factory prefect.orion.api.server:create_app
```

### Error

```
___ ___ ___ ___ ___ ___ _____    ___  ___ ___ ___  _  _
| _ \ _ \ __| __| __/ __|_   _|  / _ \| _ \_ _/ _ \| \| |
|  _/   / _|| _|| _| (__  | |   | (_) |   /| | (_) | .` |
|_| |_|_\___|_| |___\___| |_|    \___/|_|_\___\___/|_|\_|
Configure Prefect to communicate with the server with:
    prefect config set PREFECT_API_URL=http://127.0.0.1:4200/api
View the API reference documentation at http://127.0.0.1:4200/docs
Check out the dashboard at http://127.0.0.1:4200/
Usage: uvicorn [OPTIONS] APP
Try 'uvicorn --help' for help.
Error: Got unexpected extra argument (prefect.orion.api.server:create_app)
Orion stopped!
```

### Versions

```
Version:             2.4.0
API version:         0.8.0
Python version:      3.10.4
Git commit:          513639e8
Built:               Tue, Sep 13, 2022 2:15 PM
OS/Arch:             win32/AMD64
Profile:             default
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.39.2
```

### Additional context

Place where this occurs in Prefect's code: https://github.com/PrefectHQ/prefect/blob/2f27976a4dd376fea4fd06fd265dfe8a2751a764/src/prefect/cli/orion.py#L127

bunchesofdonald:
[...]

PR #6880, opened 2022-09-19
Fix Prefect Orion fails to start on Windows when there is a space in Prefect package's path

bansalkanav:
Issue - https://github.com/PrefectHQ/prefect/issues/6821

## Summary
- Bug fix suggestion for https://github.com/PrefectHQ/prefect/issues/6821
- Prefect Orion fails to start on Windows when there is a space in Prefect package's path

## Bug Fix
- `str(prefect.__module_path__.parent)` represents the absolute path. Prefect Orion fails to start in Windows OS if the directory path contains spaces.
```
Error: Got unexpected extra argument (prefect.orion.api.server:create_app)
Orion stopped!
```

- `"../.."` represents relative path. **This solves the problem.**


madkinsz:
Hi @bansalkanav! Unfortunately the relative path won't work once the code is packaged, just during development. I think we need to look into "escaping" the spaces in the path instead or include `'` around the path in the subprocess call.

bansalkanav:
@madkinsz Requesting a review. As per your suggestion, updated the code.

peytonrunyan:
Hi @bansalkanav!  It looks like you're a first-time contributor, so first, thanks for taking the time to contribute to Prefect! 

Before digging into the PR itself, I want to make sure that you've got your local development environment setup so that we're able to see the same thing when looking at proposed solutions. Our guide is located here: https://docs.prefect.io/contributing/overview/

Do you have a windows development environment that you are able to run the Prefect tests on?

bansalkanav:
Hi @peytonrunyan. Thanks for reverting back. 

Noted.

I will set up the environment and let you know. Thanks for the guidance.

discdiver:
Hi all! 👋 How are things looking for the PR?

madkinsz:
@discdiver replaced by #7224 — @peytonrunyan make sure to note superseded pull requests when opening.

PR #6895, opened 2022-09-20
Introduce `prefect.client.schemas` and client-side state type

madkinsz:
- Adds `prefect.client.schemas` with:
   - `State`: extends server-side schema to include result retrieval
   - `FlowRun`: uses client-side state type for field
   - `TaskRun`: uses client-side state type for field
   - `OrchestrationResult`: uses client-side state type for field
- Add `cls` argument to `prefect.orion.schemas.states` state type utilities for reuse client-side
- Removes `result` from server-side `State` type
- Updates the `OrionClient` to parse API responses as client types
- Moves `OrchestrationResult` to `schemas.responses` to fix circular imports

### Checklist

### DIFF
--- a/src/prefect/orion/schemas/states.py
+++ b/src/prefect/orion/schemas/states.py
@@ -165,13 +165,15 @@ class State(IDBaseModel, Generic[R]):
 
 
 def Scheduled(
-    cls: Type[State] = State, scheduled_time: datetime.datetime = None, **kwargs
+    scheduled_time: datetime.datetime = None, cls: Type[State] = State, **kwargs
 ) -> State:
     """Convenience function for creating `Scheduled` states.
 
     Returns:
         State: a Scheduled state
     """
+    # NOTE: `scheduled_time` must come first for backwards compatibility
+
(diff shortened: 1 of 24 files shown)


src/prefect/orion/schemas/states.py
```diff
@@ -64,108 +62,6 @@ class Config:
...
-    def result(state_or_future: "State[R]", raise_on_failure: bool = True) -> R:
-        ...
-
-    @overload
-    def result(
-        state_or_future: "State[R]", raise_on_failure: bool = False
-    ) -> Union[R, Exception]:
-        ...
-
-    def result(self, raise_on_failure: bool = True):
```
madkinsz, line 77: Should I make this backwards compatible by importing from the client-side dynamically? I this I ought to with a warning.

bunchesofdonald: Agreed, seems like we should have it backwards compatible for at least a single release.

bunchesofdonald (approved):
I love where this is going

src/prefect/orion/schemas/states.py
madkinsz: Done

PR #6899, opened 2022-09-20
Add `JSONSerializer` and `PickleSerializer` to `prefect.serializers`

madkinsz:
This prepares `prefect.serializers` for use with results.

- Copies base serializer model from `prefect.packaging`
     - This model has support for dispatch so we will be able to dump/load subclasses from JSON
- Copies `PickleSerializer` from `prefect.packaging.serializers` and adds tests
- Adds `JSONSerializer` with support for:
    - Alternative JSON libraries (e.g. `orjson`)
    - Pydantic JSON encoders by default, extending support to include `Set`, `UUID`, etc.
    - User-specified JSON object encoders for custom types`
  
Future work:
- Deprecation of data document serializers (#6900)
- Removal of packaging serializers

### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,7 @@ orjson >= 3.7
 packaging >= 21.3
 pathspec >= 0.8.0
 pendulum >= 2.1.2
-pydantic >= 1.8.2, != 1.9.1
+pydantic >= 1.10.0
 python-slugify >= 5.0
 pytz >= 2021.1
 pyyaml >= 5.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,5 +32,5 @@ slack-sdk>=3.15.1
 sqlalchemy[asyncio] >= 1.4.20, != 1.4.33
 toml >= 0.10.0
 typer >= 0.4.1
-typing_extensions >= 4.0.1
+typing_extensions >= 4.1.0
 uvicorn  >= 0.14.0
--- a/src/prefect/serializers.py
+++ b/src/prefect/serializers.py
@@ -15,6 +15,128 @@ if TYPE_CHECKING:
     from prefect.packaging.base import PackageManifest
     from prefect.results import _Result
 
+import abc
+import base64
+import json
+import warnings
+from typing import Any, Generic, TypeVar
+
+import cloudpickle
+import pydantic
+from anyio._core._exceptions import ExceptionGroup
+from pydantic import BaseModel
+from typing_extensions import Literal
+
+from prefect.orion.serializers import register_serializer
+from prefect.utilities.importtools import from_qualified_name
+from prefect.utilities.pydantic import add_type_dispatch
+
+D = TypeVar("D")
+
+
+def _exception_group_reduce_patch(self: ExceptionGroup):
+    return (self.__class__, (self.exceptions,))
+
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -22,7 +22,7 @@ orjson >= 3.7
 packaging >= 21.3
 pathspec >= 0.8.0
 pendulum >= 2.1.2
-pydantic >= 1.8.2, != 1.9.1
+pydantic >= 1.10.0
```
madkinsz, line 25: This is required for JSON serdes to work with dataclasses. Otherwise, we get this failure

```
assert _Pydantic_MyDataclass_140486686115840(x=1, y='test') == MyDataclass(x=1, y='test')
```

If this Pydantic upgrade is an issue for our users, we can revert this and add custom handling for dataclasses _or_ throw an error when serializing dataclasses if the Pydantic version is not sufficient.

src/prefect/serializers.py
```diff
@@ -15,9 +15,245 @@
...
+
+        for module in self.pickle_modules:
+            pickler.register_pickle_by_value(from_qualified_name(module))
+
+        blob = pickler.dumps(obj)
+
+        for module in self.pickle_modules:
+            # Restore the pickler settings
+            pickler.unregister_pickle_by_value(from_qualified_name(module))
+
```
ahuang11, line 173: Should this be a context manager / finally if error occurs in dumping, still unregister pickle?

src/prefect/serializers.py
```diff
@@ -15,9 +15,245 @@
...
+    Wraps pickles in base64 for safe transmission.
+    """
+
+    type: Literal["pickle"] = "pickle"
+
+    picklelib: str = "cloudpickle"
+    picklelib_version: str = None
+    pickle_modules: List[str] = pydantic.Field(default_factory=list)
+
```
ahuang11, line 98: Does this need load_kwargs and dumps_kwargs similar to JSONSerializer?

madkinsz: I don't think there are as many useful argument as there are for JSON. For example, I can see users wanting the `sort_keys` setting but all the pickle ones are obscure with the exception of `protocol` which I'd rather have support for directly. I'd prefer to wait and see if there are requests. I was on the fence about including it in JSON even.

src/prefect/serializers.py
madkinsz: Yeah it's also not thread-safe so we could easily bleed context here. I'm not sure what the best. I've just copied from this from `packagers` for now, I think we can address doing this correctly before release but in a separate pull request.

tests/test_serializers.py
```diff
@@ -0,0 +1,266 @@
...
+            "prefect.fake_object_encoder", fake_object_encoder, raising=False
+        )
+        monkeypatch.setattr(
+            "prefect.serializers.prefect_json_object_encoder",
+            prefect_object_encoder,
+        )
+
+        serializer = JSONSerializer(object_encoder="prefect.fake_object_encoder")
+
+        # Encoder hooks are only called for unsuipported objects
```
bunchesofdonald, line 216:
```suggestion
        # Encoder hooks are only called for unsupported objects
```

Issue #6836, opened 2022-09-15
Add MS Teams notifications to the UI


sti0:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Only SLACK notifications and email are available through UI ([docs](https://docs.prefect.io/ui/notifications/)).

### Describe the proposed behavior

Like I requested for Prefect v1.0 back then (https://github.com/PrefectHQ/prefect/issues/4129), it would be nice to create notifications for MS Teams for global alerts at the UI. Like SLACK this feature should be implemented in Orion server so its available on-prem as well.

### Example Use

End users who uses MS Teams in their tool stack could be alerted when flows are failing / crashing.

PR #6920, opened 2022-09-22
Add Apprise and Microsoft Teams notification block

rpeden:
The changes in this PR add a Microsoft Teams notification block that uses Apprise for sending notifications. Since Apprise also supports Slack, the PR also move the Slack notification block to Apprise, allowing the removal of the slack-sdk dependency. 

The PR includes a base `AppriseNotificationBlock` class that both the Slack and Teams webhook blocks inherit from. We can use this base class to quickly add other notification blocks that we might want in the UI because users will look for them by name, such as Discord or IFTTT. With additional documentation, we could easily make the Apprise notification block visible to users, allowing them to use any of the notification targets [Apprise supports](https://github.com/caronc/apprise).

Closes https://github.com/PrefectHQ/prefect/issues/6836.
### Example

The only user-visible change is the addition of the a Microsoft Teams webhook notification block. It shows up in the Blocks UI with a Teams logo, and can be used in Python in the same way as the Slack webhook notification block:
```
from prefect.blocks.notifications import MicrosoftTeamsWebhook
teams_webhook_block = MicrosoftTeamsWebhook.load("BLOCK_NAME")
teams_webhook_block.notify("Hello from Prefect!")
```

### Checklist


src/prefect/blocks/notifications.py
```diff
@@ -21,9 +23,54 @@ async def notify(self, body: str, subject: Optional[str] = None):
         """
 
 
+class PrefectNotifyType(NotifyType):
```
rpeden, line 26: A custom NotifyType was needed because Apprise adds images with an Apprise logo to notifications that use one of Apprise's built-in notification types.

tests/blocks/test_notifications.py
```diff
@@ -1,54 +1,86 @@
...
+    except UserWarning as ex:
+        # ignore the warning Prefect gives when reloading the notifications module
+        # because we reload prefect itself immediately afterward.
+        pass
+
+    reload(prefect)
+
+
+# A list of the notification classes Pytest should use as parameters to each method in TestAppriseNotificationBlock
+notification_classes = AppriseNotificationBlock.__subclasses__()
```
rpeden, line 29: It might be better to explicitly list the classes we want to test instead of including all subclasses. On the other hand, including them all lets us use this test class as a way of validating expected behavior for any new Apprise-based notification blocks we add.

desertaxle: I'm a fan of this approach. I like that it standardizes the behavior that we should expect from all Apprise notification blocks and does require either the implementer or review to remember to add a new Apprise notification block to a list somewhere. 

I think that when parameterizing tests, a deterministic order is important because the different test runs are distributed across different workers in the CI pipeline. I'd recommend adding a sort function so that `notification_classes` is always in the same order for each execution of this module.

src/prefect/blocks/notifications.py
```diff
@@ -51,11 +98,27 @@ class SlackWebhook(NotificationBlock):
...
+    Examples:
+        Load a saved Teams webhook and send a message:
+        ```python
+        from prefect.blocks.notifications import MicrosoftTeamsWebhook
+        teams_webhook_block = MicrosoftTeamsWebhook.load("BLOCK_NAME")
+        teams_webhook_block.notify("Hello from Prefect!")
+        ```
+    """
+
+    _block_type_name = "Microsoft Teams Webhook"
```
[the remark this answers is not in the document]
anna-geller: fair point, how about just Teams?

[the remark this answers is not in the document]
discdiver: Maybe **Microsoft Teams** for the name and **ms-teams** for the slug? **Teams** alone sounds like it could be for some team functionality unrelated to Microsoft Teams.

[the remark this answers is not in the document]
desertaxle: Keeping Webhook is important to avoid future collision/confusion if a Microsoft Team Channel block or some other Microsoft Teams block are added later. I think that `ms-teams-webhook` for the slug strikes the right balance between brevity and clarity.

tests/blocks/test_notifications.py
```diff
@@ -1,54 +1,86 @@
...
+
+            apprise_instance_mock = AppriseMock.return_value
+            apprise_instance_mock.async_notify = AsyncMock()
+
+            url = "https://example.com/notification"
+
+            @prefect.flow
+            async def test_flow():
+                block = block_class(url=url)
+                await block.notify("test")
```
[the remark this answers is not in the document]
rpeden: Thanks - the flow is gone now.

tests/blocks/test_notifications.py
rpeden: Thanks for the great suggestion! `notification classes` is now sorted.

tests/blocks/test_notifications.py
```diff
@@ -1,54 +1,84 @@
...
+
+            apprise_instance_mock = AppriseMock.return_value
+            apprise_instance_mock.async_notify = AsyncMock()
+
+            url = "https://example.com/notification"
+
+            @prefect.flow
+            def test_flow():
+                block = block_class(url=url)
+                block.notify("test")
```
desertaxle, line 70: This flow can also be removed.

rpeden: Makes sense. It is gone now. I had left that one because it was in the original `test_notify_sync`, but I suppose we're calling the block synchronously whether we do it in a flow or not.

desertaxle (approved):
LGTM!

madkinsz:
@rpeden Please squash merge in the future. It makes it easy to deal with the history on `main`.

rpeden:
@madkinsz Will do! Thanks for the heads up on that.

Issue #6468, opened 2022-08-18
Performance difference between Sequential and ConcurrentTaskRunner for tasks with synchronous network IO


AT-SyT:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

There is a difference in performance between running a SequentialTaskRunner and ConcurrentTaskRunner with synchronous tasks which perform heavy network IO when using synchronous tasks.

I am running into performance issues when downloading large files (via httpx) from an internal API. Those issues only arise when I use a ConcurrentTaskRunner with synchronous tasks. If I use the SequentialTaskRunner or only async tasks the performance is fine. 
Namely the throughput of the API download should be around 800Mbps which is about the limit of the connection and the API. This is the case for the SequentialTaskRunner. If I use the ConcurrentTaskRunner then the speeds fluctuate and I am between 1 and about 25Mbps. 

The issue goes away if I change the download task to use the async functionality of httpx and make the flow and the tasks async. See attached screenshots for the comparisons

The reproduction code includes all four variations in a minimal example to reproduce the issue by downloading a 1GB file from a public url. 

![Screenshot 2022-08-18 at 18 46 01](https://user-images.githubusercontent.com/12372752/185450310-7d9e7360-6b8f-4325-b769-48eb2bd16d8c.png)
![Screenshot 2022-08-18 at 18 46 33](https://user-images.githubusercontent.com/12372752/185450318-38d055f7-e1d2-40c7-97ae-fe8c76cb4f64.png)
![Screenshot 2022-08-18 at 18 46 43](https://user-images.githubusercontent.com/12372752/185450319-fee8d3f8-e475-4e30-a540-1cc8dd68e86e.png)
![Screenshot 2022-08-18 at 18 46 54](https://user-images.githubusercontent.com/12372752/185450320-73ec46dc-82b8-4f2f-b4a9-3a07c18ef076.png)
![Screenshot 2022-08-18 at 18 47 08](https://user-images.githubusercontent.com/12372752/185450322-74ad0e6a-d547-4a1e-b93b-06c36f654f38.png)

### Reproduction

```python
import io
import time

import httpx
from prefect import flow, task, get_run_logger
from prefect.task_runners import ConcurrentTaskRunner, SequentialTaskRunner


[...]

PR #6948, opened 2022-09-23
Improve concurrent task runner performance

madkinsz:
Improves performance of the concurrent task runner by using events instead of spinlocks to wait for results where feasible. I believe this only doesn't apply to use of async tasks from sync flows. If performance is a major concern in that case, users should be using sync or async consistently. This follows a pattern established in `PrefectFuture.wait`.

Closes https://github.com/PrefectHQ/prefect/issues/6468
Supersedes #6239 

### Example

In #6468 the concurrent task runner was demonstrated to be 7233% slower when performing synchronous IO. Replicated here, the concurrent task runner is 16% faster when performing synchronous IO. I did not attempt to reproduce the original slow download as my internet is not as fast as theirs and  their example would take a theoretical 3 hours to download on my machine.

I failed to demonstrate an improvement in total flow runtime with a benchmark using CPU intensive tasks derived from #6239.

### Checklist

### DIFF
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -294,10 +294,14 @@ class ConcurrentTaskRunner(BaseTaskRunner):
 
         self._result_events[key].set()
 
-    async def _get_run_result(self, key: UUID, timeout: float = None):
+    async def _get_run_result(
+        self, key: UUID, timeout: float = None
+    ) -> Optional[State]:
         """
         Block until the run result has been populated.
         """
+        result = None  # Return value on timeout
+
         with anyio.move_on_after(timeout):
 
             # Attempt to use the event to wait the result. This is much more efficient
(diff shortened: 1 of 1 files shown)


src/prefect/task_runners.py
```diff
@@ -287,15 +292,31 @@ async def _run_and_store_result(
...
+        self, key: UUID, timeout: float = None
+    ) -> Optional[State]:
         """
         Block until the run result has been populated.
         """
+        result = None  # Return value on timeout
+
         with anyio.move_on_after(timeout):
+
+            # Attempt to use the event to wait the result. This is much more efficient
```
bunchesofdonald, line 307:
```suggestion
            # Attempt to use the event to wait for the result. This is much more efficient
```

bunchesofdonald (approved):
I like it!

PR #6958, opened 2022-09-24
Warn if user tries to login with API key from Cloud 1

anna-geller:
### Problem 

We've seen some users trying to log in to Cloud 2 with Cloud 1 API key. It would be great to direct them to the right Cloud URL.

### Example

image

### Checklist

- [x] This pull request includes tests 

### DIFF
--- a/src/prefect/cli/cloud.py
+++ b/src/prefect/cli/cloud.py
@@ -160,7 +160,9 @@ async def login(
                 for workspace in workspaces
             }
         except CloudUnauthorizedError:
-            if not key.startswith("pnu"):
+            if key.startswith("pcu"):
+                help_message = "It looks like you're using API key from Cloud 1 (https://cloud.prefect.io). Make sure that you generate API key using Cloud 2 (https://app.prefect.cloud)"
+            elif not key.startswith("pnu"):
                 help_message = "Your key is not in our expected format."
             else:
                 help_message = "Please ensure your credentials are correct."
--- a/src/prefect/cli/cloud.py
+++ b/src/prefect/cli/cloud.py
@@ -160,6 +160,10 @@ async def login(
                 for workspace in workspaces
             }
         except CloudUnauthorizedError:
+            if not key.startswith("pnu"):
+                exit_with_error(
+                    "Make sure you are using API key from https://app.prefect.cloud (Prefect 2) rather than https://cloud.prefect.io (Prefect 1)"
+                )
             exit_with_error(
                 "Unable to authenticate with Prefect Cloud. Please ensure your credentials are correct."
             )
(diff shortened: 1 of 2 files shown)


src/prefect/cli/cloud.py
```diff
@@ -160,6 +160,10 @@ async def login(
                 for workspace in workspaces
             }
         except CloudUnauthorizedError:
+            if not key.startswith("pnu"):
+                exit_with_error(
+                    "Make sure you are using API key from https://app.prefect.cloud (Prefect 2) rather than https://cloud.prefect.io (Prefect 1)"
+                )
             exit_with_error(
                 "Unable to authenticate with Prefect Cloud. Please ensure your credentials are correct."
             )
```
madkinsz, line 169: Something in this style would make more sense to me:

```suggestion
            help_message = "Please ensure your credentials are correct."
            if not key.startswith("pnu"):
                help_message = "Your key is not in our expected format."
            exit_with_error(
                "Unable to authenticate with Prefect Cloud. {help_message}"
            )
```

It seems nice retain the Cloud 1 messaging, but with your implementation if I provide the key "foo" I'll get a message about Prefect 1 vs 2 which doesn't make much sense. Is there a way we can detect that the key is from Cloud 1 from the format?

madkinsz:
Since this improves the UX without fixing a bug, I've updated this to an `enhancement`. This is also a great type of change to include an example for. We should also add a test covering whatever behavior we decide on.

src/prefect/cli/cloud.py
anna-geller: > if I provide the key "foo" I'll get a message about Prefect 1 vs 2 which doesn't make much sense

It does make sense because the right Cloud2 key should start with `pnu_` and `foo` doesn't

anna-geller:
> Since this improves the UX without fixing a bug, I've updated this to an `enhancement`. This is also a great type of change to include an example for in the PR body. We should also add a test covering whatever behavior we decide on.

Thanks so much, great point, I updated the PR body

src/prefect/cli/cloud.py
madkinsz: It doesn't make sense to hint that the key was retrieved from Cloud 1 when it's just arbitrary text. This could be confusing. I'd prefer to have separate warnings for these cases:
- Warn that it's not in the expected format if it doesn't start with the expected Cloud 2 prefix
- Warn that it's from the wrong application (1 vs 2) if it actually looks like a Cloud 1 key

anna-geller: thx a lot, 100% agree. Does this look good or would you change the wording/structure?

image

Issue #6977, opened 2022-09-26
Add a custom exception type for situation where there is "no active flow or run context"


tonycombocurve:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

When calling **get_run_logger** there is "no active flow or run context", a generic Runtime exception is thrown. 
```
RuntimeError: There is no active flow or task run context.
```

### Describe the proposed behavior

It would be helpful if I could catch a specific exception type such as PrefectContextError.
```
PrefectContextError: There is no active flow or task run context.
```

### Example Use

```
try:
   logger = get_run_logger()
except PrefectContextError:
   # Houston, we have a problem
```

peytonrunyan:
@tonycombocurve just for the sake of my curiosity, do you have an example of where you might be calling `get_run_logger` outside of a flow or task run?

tonycombocurve:
It came up while refactoring some existing utility code that configures and returns a Python logger. When the function is called within the context of a flow/task, it will return the Prefect run logger. Otherwise, it will return an appropriately configured Python logger. The ideal scenario would be to add a handler for the Prefect run logger to our existing Python logger. That way we could log to both Prefect and Cloud Logging (we are running on GCP). Unfortunately, Prefect doesn't appear to provide any support for that.

peytonrunyan:
Awesome, thanks for that!

madkinsz:
> Unfortunately, Prefect doesn't appear to provide any support for that.

Whoa hey now :D We do too support that. You can customize https://github.com/PrefectHQ/prefect/blob/main/src/prefect/logging/logging.yml for full control over all of the logging happening in Prefect. You could also attach the Orion handler to your logger if you want https://github.com/PrefectHQ/prefect/blob/main/src/prefect/logging/logging.yml#L58-L60

https://docs.prefect.io/concepts/logs/#logging-configuration

PR #6980, opened 2022-09-26
Raise a `MissingContextError` when `get_run_logger` is called outside a run context

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6977

Note changing exception types is generally a breaking change. However, `MissingContextError` inherits from `RuntimeError` so this is backwards compatible.

### Example

```python
try:
   logger = get_run_logger()
except MissingContextError:
   # Can catch the specific error instead of all `RuntimeError`
```

### Checklist

### DIFF
--- a/src/prefect/logging/loggers.py
+++ b/src/prefect/logging/loggers.py
@@ -4,6 +4,7 @@ from functools import lru_cache
 from typing import TYPE_CHECKING
 
 import prefect
+from prefect.exceptions import MissingContextError
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


peytonrunyan (approved):
This seems like a harmless change, so approved, but I am curious about the usecase. What is he doing where he might be calling the prefect logger outside of a flow or task?

madkinsz:
I'm not sure! You should ask :)

I think it was an oversight that it wasn't this error type in the first place which is why I just fixed it without digging in.

PR #7004, opened 2022-09-28
Add configuration of results

madkinsz:
Changes:
- Extensive documentation of user experience for configuration of results; see [the preview](https://deploy-preview-7004--prefect-orion.netlify.app/concepts/results/#persisting-results)
- Adds `result_serializer`, `persist_result`, `result_storage` to `Flow` and `Task` interfaces
- Adds `ResultFactory` utility to determine the result settings to be used
- Adds `ResultFactory` to flow and task run contexts
- Insantiates and passes result factories in the engine

This does not:
- Implement result types or creation in the factory
- Replace existing result mechanisms

### Example

See documentation for examples.

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -213,7 +213,9 @@ Persistence of results requires a [**serializer**](#result-serializers) and a [*
 - `result_storage`: Where to store the result when persisted
 - `result_serializer`: How to convert the result to a storable form
 
-Persistence of results can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect future used by the flow or task. Persistence of results can be manually toggled on or off:
+#### Enabling and disabling persistence
+
+Persistence of results can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task. Persistence of results can be manually toggled on or off:
 
 ```python
 from prefect import flow, task
@@ -230,6 +232,8 @@ def my_task():
     ...
 ```
 
+
+#### Configuring the result storage location
 
 [The result storage location](#result-storage) can be configured with the `result_storage` option. The `result_storage` option defaults to a null value, which infers storage from the context.
 Generally, this means that tasks will use the result storage configured on the flow unless otherwise specified.
@@ -263,6 +267,8 @@ You can configure this to use a specific storage using one of the following:
 - A UUID, e.g. `'cae1dda0-5000-4ca2-a18c-727d400145f2'`
 
 
+#### Configuring the result serializer
+
 [The result serializer](#result-serializer) can be configured with the `result_storage` option. The `result_serialzier` option defaults to a null value, which infers the serializer from the context.
 Generally, this means that tasks will use the result serializer configured on the flow unless otherwise specified.
 If there is no context to load the serializer from, the serializer defined by `PREFECT_RESULTS_DEFAULT_SERIALIZER` will be used. This setting defaults to Prefect's pickle serializer.
@@ -272,15 +278,32 @@ You may set the configure the result serializer using:
 - A type name, e.g. `"json"` or `"pickle"` — this corresponds to an instance with default values
 - An instance, e.g. `JSONSerializer(jsonlib="orjson")`
 
-### Result storage
+### Storage of results in Prefect
+
+The Prefect API does not store your results in most cases for the following reasons:
+
+- Results can be large and slow to send to and from the API
+- Results often contain private information or data
+- Results would need to be stored in the database or complex logic implemented to hydrate from another source
+
+There are a few cases where Prefect _will_ store your results directly in the database. This is an optimization to reduce the overhead of reading and writing to result storage.
+
+The following data types will be stored by the API without persistence to storage:
+
+- booleans (`True`, `False`)
+- nulls (`None`)
+
+If `persist_result` is set to `False`, these values will never be stored.
+
+
+## Result storage
 
 Result storage is responsible fo reading and writing serialized data to an external location. At this time, any file system block can be used for result storage.
-
-### Result serializers
+## Result serializers
 
 A result serializer is responsible for converting your Python object to and from bytes. This is necessary to store the object outside of Python and retrieve it later.
 
-#### Pickle serializer
+### Pickle serializer
 
 Pickle is a standard Python protocol for encoding arbitrary Python objects. We supply a custom Pickle serializer at `prefect.serializers.PickleSerializer`. Prefect's pickle serializer uses the [cloudpickle](https://github.com/cloudpipe/cloudpickle) project by default to support more object types. Alternative pickle libraries can be specified:
 
@@ -302,7 +325,7 @@ Drawbacks of the pickle serializer:
 - Serialized objects cannot be easily shared across different programming languages
 - Serialized objects are not human readable
 
-#### JSON serializer
+### JSON serializer
 
 
 We supply a custom JSON serializer at `prefect.serializers.JSONSerializer`. Prefect's JSON serializer uses custom hooks by default to support more object types. Specifically, we add support for all types supported by [Pydantic](https://pydantic-docs.helpmanual.io/).
@@ -327,7 +350,30 @@ Drawbacks of the JSON serializer:
 - Supported types are limited
 - Implementing support for additional types must be done at the serializer level
 
-### Result references
+
+## Result types
+
+Prefect uses the internal `Result` type to capture information about the result attached to a state. There are two types of `Result`:
+
+- `ResultLiteral`
+- `ResultReference`
+
+All result types include a `get()` method that can be called to return the value of the result. This is done behind the scenes when the `result()` method is used on states or futures.
+
+### Result literals
+
+Result literals are used to represent [results stored in the Prefect database](#storage-of-results-in-prefect). The values contained by these results must always be JSON serializable.
+
+Example:
+```
+result = ResultLiteral(value=None)
+result.json()
+# {"type": "result", "value": "null"}
+```
+
+Result literals reduce the overhead required to persist simple results.
+
+###  Result references
 
 Result references contain all of the information needed to retrieve the result from storage. This includes:
 
@@ -335,21 +381,4 @@ Result references contain all of the information needed to retrieve the result f
 - Key: indicates where this specific result is in storage
 - Serializer: description of the [result serializer](#result-serializers) that can be used to deserialize the result
 
-Result references implement the `get()` method which retrieves the data from storage, deserializes it, and returns the original object. This is done automatically when the `result()` method is used on states or futures. The `get()` operation will cache the resolved object to reduce the overhead of subsequent calls.
-
-### Storage of results in Prefect
-
-The Prefect API does not store your results in most cases for the following reasons:
-
-- Results can be large and slow to send to and from the API
-- Results often contain private information or data
-- Results would need to be stored in the database or complex logic implemented to hydrate from another source
-
-There are a few cases where Prefect _will_ store your results directly in the database. This is an optimization to reduce the overhead of reading and writing to result storage.
-
-The following data types will be stored by the API without persistence to storage:
-
-- booleans (`True`, `False`)
-- nulls (`None`)
-
-If `persist_result` is set to `False`, these values will never be stored.
+The `get()` method on result references retrieves the data from storage, deserializes it, and returns the original object. The `get()` operation will cache the resolved object to reduce the overhead of subsequent calls.
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -10,7 +10,7 @@ tags:
 
 # Results
 
-Results represent the data returned by a flow or a task. 
+Results represent the data returned by a flow or a task.
 
 ## Retrieving results
 
@@ -87,7 +87,7 @@ def my_flow():
         my_task()
     except ValueError:
         print("Oh no! The task failed.")
-    
+
     return True
 
 my_flow()
@@ -118,7 +118,7 @@ result = my_flow()
 assert result == 2
 ```
 
-If you retrieve the result from a failed state, the exception will be raised. For this reason, it's often best to check if the state is failed first. 
+If you retrieve the result from a failed state, the exception will be raised. For this reason, it's often best to check if the state is failed first.
 
 ```python
 from prefect import flow, task
@@ -135,7 +135,7 @@ def my_flow():
         result = state.result()
     except ValueError:
         print("Oh no! The state raised the error!")
-    
+
     return True
 
 my_flow()
@@ -160,7 +160,7 @@ def my_flow():
         result = 1
     else:
         result = maybe_result
-    
+
     return result + 1
 
 result = my_flow()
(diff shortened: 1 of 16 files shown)


madkinsz:
👋 The next step for this pull request is to add tests asserting that:
1. Result settings are being merged as expected into factories
2. Factories are not saving blocks multiple times

Follow-ups pull requests will include:
1. Addition of `Result` types
2. Implementation of result instance creation from the factory
3. Replacing existing result handling with calls to the result factory

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+- booleans (`True`, `False`)
+- nulls (`None`)
+
+If `persist_result` is set to `False`, these values will never be stored.
+
+
+## Result storage types
+
+Result storage is responsible fo reading and writing serialized data to an external location. At this time, any file system block can be used for result storage.
```
bunchesofdonald, line 300:
```suggestion
Result storage is responsible for reading and writing serialized data to an external location. At this time, any file system block can be used for result storage.
```

src/prefect/results.py
```diff
@@ -43,3 +56,201 @@ async def _retrieve_result(state):
...
+            storage_block = await Block.load(result_storage, client=client)
+            storage_block_id = storage_block._block_document_id
+            assert storage_block_id is not None, "Loaded storage blocks must have ids"
+        else:
+            raise TypeError(
+                "Result storage must be one of the following types: 'UUID', 'Block', "
+                f"'str'. Got unsupported type {type(result_storage).__name__!r}."
+            )
+
+        return storage_block_id, storage_block
```
bunchesofdonald, line 240: Do we need to check if the block has file read/write capability?

bunchesofdonald (approved):
That's a great start! I'm excited to see this come together.

src/prefect/results.py
madkinsz: Yeah I think we need a utility to do so. It'd be great to check earlier (e.g. at flow declaration) rather than at runtime (when this object is created).

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+#### Result serializer
+
+[The result serializer](#result-serializer-types) can be configured with the `result_serializer` option. The `result_serializer` option defaults to a null value, which infers the serializer from the context.
+Generally, this means that tasks will use the result serializer configured on the flow unless otherwise specified.
+If there is no context to load the serializer from, the serializer defined by `PREFECT_RESULTS_DEFAULT_SERIALIZER` will be used. This setting defaults to Prefect's pickle serializer.
+
+You may set the configure the result serializer using:
+
+- A type name, e.g. `"json"` or `"pickle"` — this corresponds to an instance with default values
```
tpdorsey, line 277:
```suggestion
- A type name, e.g. `"json"` or `"pickle"` &mdash; this corresponds to an instance with default values
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- An instance, e.g. `JSONSerializer(jsonlib="orjson")`
+
+### Storage of results in Prefect
+
+The Prefect API does not store your results in most cases for the following reasons:
+
+- Results can be large and slow to send to and from the API
```
tpdorsey, line 284:
```suggestion
- Results can be large and slow to send to and from the API.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Results often contain private information or data
```
tpdorsey, line 285:
```suggestion
- Results often contain private information or data.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Results would need to be stored in the database or complex logic implemented to hydrate from another source
```
tpdorsey, line 286:
```suggestion
- Results would need to be stored in the database or complex logic implemented to hydrate from another source.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+## Result storage types
+
+Result storage is responsible fo reading and writing serialized data to an external location. At this time, any file system block can be used for result storage.
+## Result serializer types
+
+A result serializer is responsible for converting your Python object to and from bytes. This is necessary to store the object outside of Python and retrieve it later.
+
+### Pickle serializer
+
+Pickle is a standard Python protocol for encoding arbitrary Python objects. We supply a custom Pickle serializer at `prefect.serializers.PickleSerializer`. Prefect's pickle serializer uses the [cloudpickle](https://github.com/cloudpipe/cloudpickle) project by default to support more object types. Alternative pickle libraries can be specified:
```
tpdorsey, line 307:
```suggestion
Pickle is a standard Python protocol for encoding arbitrary Python objects. We supply a custom pickle serializer at `prefect.serializers.PickleSerializer`. Prefect's pickle serializer uses the [cloudpickle](https://github.com/cloudpipe/cloudpickle) project by default to support more object types. Alternative pickle libraries can be specified:
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+```python
+from prefect.serializers import PickleSerializer
+
+PickleSerializer(picklelib="custompickle")
+```
+
+Benefits of the pickle serializer:
+
+- Many object types are supported
```
tpdorsey, line 317:
```suggestion
- Many object types are supported.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Objects can define custom pickle support
```
tpdorsey, line 318:
```suggestion
- Objects can define custom pickle support.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+Drawbacks of the pickle serializer:
+
+- When nested attributes of an object cannot be pickled, it is hard to determine the cause
```
tpdorsey, line 322:
```suggestion
- When nested attributes of an object cannot be pickled, it is hard to determine the cause.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- When deserializing objects, your Python and pickle library versions must match the one used at serialization time
```
tpdorsey, line 323:
```suggestion
- When deserializing objects, your Python and pickle library versions must match the one used at serialization time.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Serialized objects cannot be easily shared across different programming languages
```
tpdorsey, line 324:
```suggestion
- Serialized objects cannot be easily shared across different programming languages.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Serialized objects are not human readable
```
tpdorsey, line 325:
```suggestion
- Serialized objects are not human readable.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+```python
+from prefect.serializers import JSONSerializer
+
+JSONSerializer(jsonlib="orjson")
+```
+
+Benefits of the JSON serializer:
+
+- Serialized objects are human readable
```
tpdorsey, line 343:
```suggestion
- Serialized objects are human readable.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Serialized objects can often be shared across different programming languages
```
tpdorsey, line 344:
```suggestion
- Serialized objects can often be shared across different programming languages.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Deserialization of serialized objects is generally version agnostic
```
tpdorsey, line 345:
```suggestion
- Deserialization of serialized objects is generally version agnostic.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+
+Drawbacks of the JSON serializer:
+
+- Supported types are limited
```
tpdorsey, line 349:
```suggestion
- Supported types are limited.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Implementing support for additional types must be done at the serializer level
```
tpdorsey, line 350:
```suggestion
- Implementing support for additional types must be done at the serializer level.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+# {"type": "result", "value": "null"}
+```
+
+Result literals reduce the overhead required to persist simple results.
+
+###  Result references
+
+Result references contain all of the information needed to retrieve the result from storage. This includes:
+
+- Storage: a reference to the [result storage](#result-storage-types) that can be used to read the serialized result
```
tpdorsey, line 379:
```suggestion
- Storage: a reference to the [result storage](#result-storage-types) that can be used to read the serialized result.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Key: indicates where this specific result is in storage
```
tpdorsey, line 380:
```suggestion
- Key: indicates where this specific result is in storage.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- Serializer: description of the [result serializer](#result-serializer-types) that can be used to deserialize the result
```
tpdorsey, line 381:
```suggestion
- Serializer: description of the [result serializer](#result-serializer-types) that can be used to deserialize the result.
```

src/prefect/flows.py
```diff
@@ -94,6 +97,19 @@ class Flow(Generic[P, R]):
         retries: An optional number of times to retry on flow run failure.
         retry_delay_seconds: An optional number of seconds to wait before retrying the
             flow after failure. This is only applicable if `retries` is nonzero.
+        persist_result: An optional toggle indiciating if the result of this flow
```
tpdorsey, line 100:
```suggestion
        persist_result: An optional toggle indicating whether the result of this flow
```

src/prefect/flows.py
```diff
@@ -94,6 +97,19 @@ class Flow(Generic[P, R]):
...
+            should be persisted to result storage. Defaults to `None` which indicates
```
tpdorsey, line 101:
```suggestion
            should be persisted to result storage. Defaults to `None`, which indicates
```

src/prefect/flows.py
```diff
@@ -94,6 +97,19 @@ class Flow(Generic[P, R]):
...
+            that Prefect should choose if the result should be persisted depending on
```
tpdorsey, line 102:
```suggestion
            that Prefect should choose whether the result should be persisted depending on
```

tpdorsey (commented):
Some minor suggested edits.

madkinsz:
@abrookins this actually doesn't change the REST API so I have removed that label.

src/prefect/results.py
```diff
@@ -43,3 +56,201 @@ async def _retrieve_result(state):
...
+
+    @staticmethod
+    def resolve_serializer(serializer: ResultSerializer) -> Serializer:
+        """
+        Result one of the valid `ResultSerializer` input types into a serializer
```
madkinsz, line 245:
```suggestion
        Resolve one of the valid `ResultSerializer` input types into a serializer
```

madkinsz:
Thank you so much for the thorough documentation improvements @tpdorsey !

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
 
 my_flow()
 ```
+
+## Persisting results
+
+The Prefect API does not store your results [except in special cases](#storage-of-results-in-prefect). Instead, the result is **persisted** to a storage location in your infrastructure and Prefect stores a **reference** to the result.
```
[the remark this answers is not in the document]
madkinsz: `reference` will be a link soon as well
```suggestion
The Prefect API does not store your results [except in special cases](#storage-of-results-in-prefect). Instead, the result is _persisted_ to a storage location in your infrastructure and Prefect stores a _reference_ to the result.
```

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
 ```
+
+## Persisting results
+
+The Prefect API does not store your results [except in special cases](#storage-of-results-in-prefect). Instead, the result is _persisted_ to a storage location in your infrastructure and Prefect stores a _reference_ to the result.
+
+The following Prefect features require results to be persisted:
+
+- Task cache keys
+- Flow run retries
```
jlowin, line 204: I think this won't be a universally true statement -- a flow run composed of tasks that don't exchange data should be able to retry regardless of this setting. Recognizing that this is a pedantic point, perhaps `- Retrying flows that exchange data` would be more specific?

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+- `result_serializer`: How to convert the result to a storable form.
+
+#### Toggling persistence
+
+Persistence of results can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task. Persistence of results can be manually toggled on or off:
+
+```python
+from prefect import flow, task
+
+@flow(persist_result=True)
```
jlowin, line 223: For flows, `persist_result` makes me think that this is only for the (single) result of the flow, rather than a setting that will be inherited by all tasks (and subflows?) inside the flow. `persist_results` (plural) is what I would expect to see there. I realize this would break symmetry with the task decorator, but is a more accurate description of this behavior.

docs/concepts/results.md
```diff
@@ -193,3 +193,191 @@ def my_flow():
...
+new_flow = my_flow.with_options(result_storage=S3(bucket="my-bucket"))
+
+new_flow()  # The flow and task within it will use S3 for result storage.
+```
+
+You can configure this to use a specific storage using one of the following:
+
+- A storage instance, e.g. `LocalFileSystem(basepath=".my-results"`
+- A storage slug, e.g. `'s3/dev-s3-block'`
+- A UUID, e.g. `'cae1dda0-5000-4ca2-a18c-727d400145f2'`
```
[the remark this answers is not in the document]
madkinsz: Okay! I'll remove it.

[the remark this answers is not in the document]
madkinsz: Well... it seems kind of nice to be able to pin to a specific block document. Should I just improve this documentation or remove it?

docs/concepts/results.md
madkinsz: That's true, but as-is we're going to turn on results if you have flow run retries enabled whether or not your tasks exchange data. We don't know if your task's data will be used downstream when it runs and we need to decide if it should be persisted or not.

docs/concepts/results.md
madkinsz: It only affects the flow. Unlike the other options, this setting will not be inherited. How can I clarify this in the persistence toggle doc?

A separate setting i.e. `persist_child_results` or `persist_task_results` may be added in the future to toggle the default for children, but Bill and I decided it was not necessary in the first go.

madkinsz: > Persistence of results can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task.

vs 

> [The result storage location](#result-storage-types) can be configured with the `result_storage` option. The `result_storage` option defaults to a null value, which infers storage from the context.
Generally, this means that tasks will use the result storage configured on the flow unless otherwise specified.

docs/concepts/results.md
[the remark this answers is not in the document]
jlowin: I think adding UUID would become a creeping path to (what would become) a legitimate request / confusion of how we refer to block documents across all features -- for now I would say keep the access pattern universal and only `type/name`.

docs/concepts/results.md
jlowin: Oh sweet 👍

docs/concepts/results.md
jlowin: Sorry, I think I misunderstood -- I guess I assumed this was inherited because the other options were. In that case I retract my comment! I'm not sure if my confusion is universal (it could have just been me reading quickly) so I'm not sure if requires explicit documentation at this time.

docs/concepts/results.md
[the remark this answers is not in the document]
madkinsz: Great.

docs/concepts/results.md
anna-geller: `persist_results` in the plural form is actually not bad and works for both -- 1) what if the task returns a tuple 2) one could see it as _persist results of multiple runs_ for this task/flow

PR #7009, opened 2022-09-28
Update `sync_compatible` handling of wrapped async functions and generators

madkinsz:
While investigating https://github.com/PrefectHQ/prefect-dask/pull/33 two issues were found with `sync_compatible`.

1. It does not recognize async functions wrapped by sync decorators
2. Async generators are not detected as async by `iscoroutinefn` / `is_async_fn`

These were fixed by:

1. Updating `sync_compatible` to call `is_async_fn`
2. Updating `is_async_fn` to unwrap functions
3. Adding `is_async_gen_fn` to detect async generators

I briefly explored adding support to `sync_compatible` for async generators. I'm not actually sure it's possible. I've done some minor refactoring to allow us to do so in the future, but for now we will not allow it. The error message is clearer now, at least.

### Example

```python
async def foo():
   yield True

async def bar():
   return True

is_async_fn(foo)  # False
is_async_fn(bar)  # True

# New utility
is_async_gen_fn(foo)  # True
is_async_gen_fn(bar)  # False


# Support for wrappers that define a sync function
def wrapper(fn):
    @wraps(fn)
    def inner():
        return fn()

    return inner

@wrapper
async def foo():
    return True

is_async_fn(foo) # True (previously False)

sync_compatible(foo)  # Works (previously errored)
```

### Checklist

### DIFF
--- a/src/prefect/utilities/asyncutils.py
+++ b/src/prefect/utilities/asyncutils.py
@@ -32,12 +32,24 @@ def is_async_fn(
     func: Union[Callable[P, R], Callable[P, Awaitable[R]]]
 ) -> TypeGuard[Callable[P, Awaitable[R]]]:
     """
-    This wraps `iscoroutinefunction` with a `TypeGuard` such that we can perform a
-    conditional check and type checkers will narrow the expected type.
+    Returns `True` if a function returns a coroutine.
 
     See https://github.com/microsoft/pyright/issues/2142 for an example use
     """
+    while hasattr(func, "__wrapped__"):
+        func = func.__wrapped__
+
     return inspect.iscoroutinefunction(func)
+
+
+def is_async_gen_fn(func):
+    """
+    Returns `True` if a function is an async generator.
+    """
+    while hasattr(func, "__wrapped__"):
+        func = func.__wrapped__
+
+    return inspect.isasyncgenfunction(func)
 
 
 async def run_sync_in_worker_thread(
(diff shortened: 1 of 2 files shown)


madkinsz:
@ahuang11 feels like we'll need to provide a different utility for async for now.

ahuang11:
Are non-async contextmanagers supported?

madkinsz:
> Are non-async contextmanagers supported?

No need for `sync_compatible` in that case! What kind of pattern do you need?

ahuang11:
> > Are non-async contextmanagers supported?
> 
> No need for `sync_compatible` in that case! What kind of pattern do you need?

Oh yeah brain wasn't working after staring at code too long.

ahuang11:
I wonder if we can use something like:
https://anyio.readthedocs.io/en/stable/threads.html?highlight=async%20context%20manager#using-asynchronous-context-managers-from-worker-threads

```
from anyio.from_thread import start_blocking_portal


class AsyncContextManager:
    async def __aenter__(self):
        print('entering')

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        print('exiting with', exc_type)


async_cm = AsyncContextManager()
with start_blocking_portal() as portal, portal.wrap_async_context_manager(async_cm):
    print('inside the context manager block')
```

madkinsz:
We can definitely do something like that! The trick is getting it work efficiently from a worker thread and in the "already async" case. I think we'll want to solve this for a sync-compatible `OrionClient` as well, but I don't see the Dask client context as worth diving into this right now unless you want to take it on in some spare time.

ahuang11:
I think I'll just separate the sync and async utils for now!

src/prefect/utilities/asyncutils.py
```diff
@@ -32,14 +32,26 @@ def is_async_fn(
...
 
     See https://github.com/microsoft/pyright/issues/2142 for an example use
     """
+    while hasattr(func, "__wrapped__"):
+        func = func.__wrapped__
+
     return inspect.iscoroutinefunction(func)
 
 
+def is_async_gen_fn(func):
```
ahuang11, line 45:
```suggestion
def is_async_gen_fn(func: Callable) -> bool:
```

ahuang11 (approved):

src/prefect/utilities/asyncutils.py
madkinsz: I'm actually scared to try to reproduce the correct signature here which would be a type guard as above. I'm hoping by omitting it the type is inferred from the inner call which is a type guard.

Issue #6906, opened 2022-09-21
Subpaths and remote subdirectory paths can't be used on deployments with GCS storage block


taljaards:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If you create a deployment with a storage block that has a directory as part of its "base path" _and_ a `path`, the files are not downloaded correctly from the storage and the flow run fails.

(GCS) Storage Blocks can have a directory added [as part of their bucket path](https://docs.prefect.io/tutorials/storage/#create-a-storage-block).
[As described in the tutorial](https://docs.prefect.io/tutorials/storage/#sharing-storage), you can also share Storage Blocks between deployments and put each deployment's file in its own subdirectory by making use of `path` on a deployment.

---

I also tested by removing the directory from the "base path". Then it runs fine. When I add a `path`, it still runs fine.
When I add another folder level, it fails (see Reproduction method 2).

### Reproduction

There's two ways to get multiple dir levels as part of the flow storage.

#### Method 1:

I followed the tutorial (https://docs.prefect.io/tutorials/deployments/#setting-up).

Differences:
- I'm running inside a Poetry venv.
- My GCS block has an additional directory as part of its base path:
    
![image](https://user-images.githubusercontent.com/6089256/191446769-a9d7960e-e6d2-470a-987a-72f37f51418e.png)

Everything works fine if the deployment does not have a `path` specified:
```
poetry run prefect deployment build ./log_flow.py:log_flow -n log-simple -q test 
  -sb gcs/poc-gcs-bucket-dev --apply
```

When adding a `path`, then running a flow from that deployment, fails.
```
poetry run prefect deployment build ./log_flow.py:log_flow -n log-simple -q test 
  -sb gcs/poc-gcs-bucket-dev --path log-flow-directory --apply
```
Errors posted in error section.

---

#### Method 2:

Remove the base path from the storage block.
Add more than one level to the path here:
```
poetry run prefect deployment build ./log_flow.py:log_flow -n log-simple -q test 
  -sb gcs/poc-gcs-bucket-dev --path testing/log-flow-directory --apply
```

### Error

<details>
  <summary>On Windows</summary>

```
[...]

PR #7012, opened 2022-09-29
Expose `ignore_file` to `Deployment.build_from_flow()` Python object

taljaards:
With a [Python object `Deployments.build_from_flow()`](https://docs.prefect.io/concepts/deployments/#create-a-deployment-from-a-python-object) (with the `skip_upload=True` "shortcut"), specifying the `ignore_file` is not exposed to the user.

### Example/context

<details>
  <summary>Besides the use-case here, I'm sure it would be beneficial to be able to specify this file.</summary>
  
  By default the `.prefectignore` in the root dir is used.
  If you have a list of flows in different subdirs, a workaround for #6906 is to `os.chdir()` into each flow's subdir before registering it, causing the path to only have one level (and thus the flow run pulls the remote file(s) successfully).
  However, now using the `Deployment.build_from_flow(..., skip_upload=False,  apply=True)` shortcuts, it now expects a `.prefectignore` file in each of the subdirs.
  
</details>

### Alternative

The current working alternative is to do the actions separately:
```py
dep = Deployment(..., skip_upload=True, apply=True)
dep.upload_to_storage(ignore_file=path_to_shared_custom_ignore_file)
```
From https://medium.com/the-prefect-blog/prefect-2-3-0-adds-support-for-flows-defined-in-docker-images-and-github-repositories-79a8797a7371#dc27, the more streamlined single step seems like a great selling point. But since there is a working alternative, you can also decline this PR.

### Checklist


taljaards:
Let me know your thoughts - we could also apply this change here https://github.com/PrefectHQ/prefect/blob/27ae06a0ed2fb0601aaafcb9f88d1de2fdca0655/src/prefect/deployments.py#L501

peytonrunyan:
Hi @taljaards I just wanted to let you know that I've seen this and will be giving it a look over soon!

anna-geller:
@taljaards, thanks so much for this PR, it looks good. But have you QAed if this doesn't break the build/apply from CLI? CLI remains the standard interface for creating deployments, so we can accept this PR only if this is implemented in a way that doesn't break CLI 

I can QA as well, but wanted to ask first

peytonrunyan:
@taljaards an update on where I'm at with this:

We call `deployment.upload_to_storage()` from a few different functions at the moment, and we don't expose the ignore file at any of the calling functions. If we want to make the path to the ignore file accessible in places other than `upload_to_storage`, we'll probably want to make sure that it's done consistently, rather than in a single location. Before going down that path, I want to talk with somebody who was more involved in the design of deployments and make sure that's coherent with the plans for deployments. This conversation should hopefully happen soon. Thanks for your patience here!

peytonrunyan:
Hey @taljaards - I had the conversation and the consensus is that this is a great way to go! I'll approve and merge this PR shortly. Thanks again for taking the initiative on this, and for your patience.

We'll also be exposing it in the other methods:

* the `apply` CLI command
* the `build` CLI command 
* `deployment.apply()`

I'm happy to do these myself, but if you'd rather make a follow up PR, we can also do it that way. The next one should move a bit quicker 😅.

peytonrunyan (approved):

PR #7013, opened 2022-09-29
Add capacity limiter to ConcurrentTaskRunner

jashbycu:
Adds a CapacityLimiter to the ConcurrentTaskRunner to limit the number of concurrent threads. This is helpful for flows that would otherwise spawn hundreds of threads.

### Example

```
from prefect.task_runners import ConcurrentTaskRunner

@flow(task_runner=ConcurrentTaskRunner(max_workers=10))
def example_flow():
    # Flow that runs with at most 10 threads ...
```

### Checklist

### DIFF
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -234,6 +234,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         self._results: Dict[UUID, Any] = {}
         self._keys: Set[UUID] = set()
         self._max_workers = max_workers
+        self._limiter = None
 
         super().__init__()
 
@@ -288,8 +289,8 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         task crashes from crashing the flow run.
         """
         try:
-            if self._max_workers is not None:
-                async with CapacityLimiter(self._max_workers):
+            if self._limiter is not None:
+                async with self._limiter:
                     self._results[key] = await call()
             else:
                 self._results[key] = await call()
@@ -327,6 +328,8 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         """
         Start the process pool
         """
+        if self._max_workers is not None:
+            self._limiter = CapacityLimiter(self._max_workers)
         self._task_group = await exit_stack.enter_async_context(
             anyio.create_task_group()
         )
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -65,7 +65,6 @@ from typing import (
 from uuid import UUID
 
 import anyio
-from anyio import CapacityLimiter
 
 from prefect.utilities.collections import AutoEnum
 
@@ -76,6 +75,7 @@ from prefect.logging import get_logger
 from prefect.orion.schemas.states import State
 from prefect.states import exception_to_crashed_state
 from prefect.utilities.collections import AutoEnum
+from prefect.utilities.asyncutils import asyncnullcontext
 
 T = TypeVar("T", bound="BaseTaskRunner")
 R = TypeVar("R")
@@ -234,7 +234,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         self._results: Dict[UUID, Any] = {}
         self._keys: Set[UUID] = set()
         self._max_workers = max_workers
-        self._limiter = None
+        self._limiter = asyncnullcontext
 
         super().__init__()
 
@@ -289,10 +289,7 @@ class ConcurrentTaskRunner(BaseTaskRunner):
         task crashes from crashing the flow run.
         """
         try:
-            if self._limiter is not None:
-                async with self._limiter:
-                    self._results[key] = await call()
-            else:
+            async with self._limiter:
                 self._results[key] = await call()
         except BaseException as exc:
             self._results[key] = exception_to_crashed_state(exc)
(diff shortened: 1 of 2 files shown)


src/prefect/task_runners.py
```diff
@@ -323,6 +328,8 @@ async def _start(self, exit_stack: AsyncExitStack):
         """
         Start the process pool
         """
+        if self._max_workers is not None:
+            self._limiter = CapacityLimiter(self._max_workers)
```
madkinsz, line 332: Could we set this to an `asyncnullcontext` when `_max_workers` is null so that the code in `_run_and_store_result` doesn't need a branch?

src/prefect/task_runners.py
```diff
@@ -65,6 +65,7 @@
 from uuid import UUID
 
 import anyio
+from anyio import CapacityLimiter
```
madkinsz, line 68: Can you use `anyio.CapacityLimiter` instead of importing the name? It's nice to have the clarity of where the implementation is.

madkinsz (changes requested):
Thanks for contributing! I've been meaning to add something like this. A couple comments on the implementation. We'll also need some sort of test before we can accept it.

madkinsz:
> Looking at test_task_runners, I'm not sure the best way to test this. Do we just need an additional test suite that yields (for example) `ConcurrentTaskRunner(max_workers=10)`?

That'd be a good start! It seems quite challenging to "prove" that it's being limited to some number of workers. Maybe would be useful to look at the AnyIO capacity limiter tests and see how they check it.

src/prefect/task_runners.py
```diff
@@ -226,14 +227,14 @@ class ConcurrentTaskRunner(BaseTaskRunner):
...
-        # TODO: Consider adding `max_workers` support using anyio capacity limiters
-
+    def __init__(self, max_workers: int =None):
         # Runtime attributes
         self._task_group: anyio.abc.TaskGroup = None
         self._result_events: Dict[UUID, anyio.abc.Event] = {}
         self._results: Dict[UUID, Any] = {}
         self._keys: Set[UUID] = set()
+        self._max_workers = max_workers
+        self._limiter = asyncnullcontext
```
madkinsz, line 237:
```suggestion
        self._limiter = asyncnullcontext()
```

I'm also curious, is there a reason we need to instantiate the limiter in `_start` instead of `__init__`? Does it need the event loop?

madkinsz:
We might be able to set `max_workers=1` then assert that the task runner behaves sequentially instead of concurrently?

jashbycu:
Thanks Michael. I've pushed some test ideas, but I can't get the tests to run with `pytest tests/test_task_runners.py`. I get `pytest: error: unrecognized arguments: --cov=src/ --cov=tests/ --no-cov-on-fail --cov-report=term`. I don't see any detailed testing instructions in this repo. Can you help me?

One thing I'm not sure of is if `max_workers=1` will result in truly sequential task running, or if the ConcurrentTask Runner might run nearly-simultaneously-submitted tasks in an arbitrary order. So if the tests fail due to this, I have a workaround idea. But I have to be able to run the tests to see.

tests/test_task_runners.py
```diff
@@ -32,3 +34,55 @@ class TestConcurrentTaskRunner(TaskRunnerStandardTestSuite):
...
+
+class TestConcurrentTaskRunnerLimitedThreads(TaskRunnerStandardTestSuite):
+    @pytest.fixture
+    def task_runner(self):
+        yield ConcurrentTaskRunner(max_workers=3)
+
+class TestConcurrentTaskRunnerSingleThreaded(TaskRunnerStandardTestSuite):
+    @pytest.fixture
+    def task_runner(self):
+        yield ConcurrentTaskRunner(max_workers=1)
```
madkinsz, line 46: You may need to / be able to change the the `concurrency_type` on this instance of the task runner from `TaskConcurrencyType.CONCURRENT` to `TaskConcurrencyType.SEQUENTIAL` so it goes through the sequential standard tests

madkinsz:
>  I've pushed some test ideas, but I can't get the tests to run with pytest tests/test_task_runners.py. I get pytest: error: unrecognized arguments: --cov=src/ --cov=tests/ --no-cov-on-fail --cov-report=term. I don't see any detailed testing instructions in this repo. Can you help me?

Sounds like you're missing some test requirements. Did you install with the `dev` extra? https://docs.prefect.io/contributing/overview/#setting-up-a-development-environment

> One thing I'm not sure of is if max_workers=1 will result in truly sequential task running, or if the ConcurrentTask Runner might run nearly-simultaneously-submitted tasks in an arbitrary order. So if the tests fail due to this, I have a workaround idea. But I have to be able to run the tests to see

Also not sure about what will happen there :) that seems like a possibility. If it's really a pain, we can explore a different test pattern.

jashbycu:
Ok I see that about the tests. I adjusted it to simply override the concurrency type, which hopefully will let it pass the tests. But I'm having trouble getting a dev env working on my computer, and unfortunately I have to focus on some other high priority work for me now. Short of tinkering with the tests, is there anything else I can do to help? I've added you as a collaborator to my fork.

madkinsz:
I've run the tests in CI and it looks like there are some issues. We don't have the bandwidth to prioritize this right now, but if it languishes we'll eventually get an engineer to get it across the line. Futzing with the tests is definitely the biggest part of this change :)

jashbycu:
Some more testing on my end shows that with more complex flows, this method clogs up and results in flows indefinitely stalling. I think it's because pending tasks count toward the thread limit. For my use case, I ended up using task tag concurrency limits, where each flow run I generate a unique tag to limit all tasks of that flow run, and then delete it when I'm done. I'm not sure what the Prefect solution should be, but maybe it needs to follow a similar pattern to that used by the task tags. In any case, I'm going to close this. Thank you!

madkinsz:
Thanks for investigating! It seems like we'd need to take the concurrency limit right before calling the user's function which is indeed trickier.

Issue #7040, opened 2022-09-30
Sync flow calling async flow calling sync task results in sync_portal == None -> no attribute 'call'`


ahuang11:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When I use a sync flow to call an async subflow to call a sync task, there is no sync portal available.

cc @anna-geller 

Related:
https://github.com/PrefectHQ/prefect-dbt/issues/67

### Reproduction

```python3
from prefect import flow, task

@task
def test_subtask(obj):
    return obj.get("id")

@flow
async def test_async_flow():
    obj = {"id": 123}
    return test_subtask(obj)

@flow
async def test_sync_flow():
    await test_async_flow()

await test_sync_flow()
```
```


### Error

```python3
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [2], line 16
     12 @flow
     13 def test_sync_flow():
     14     test_async_flow()
---> 16 test_sync_flow()

File ~/Applications/python/prefect/src/prefect/flows.py:390, in Flow.__call__(self, return_state, *args, **kwargs)
    386 parameters = get_call_parameters(self.fn, args, kwargs)
    388 return_type = "state" if return_state else "result"
--> 390 return enter_flow_run_engine_from_flow_call(
    391     self, parameters, return_type=return_type
    392 )

File ~/Applications/python/prefect/src/prefect/engine.py:157, in enter_flow_run_engine_from_flow_call(flow, parameters, return_type)
    153 elif in_async_main_thread():
    154     # An event loop is already running and we must create a blocking portal to
    155     # run async code from this synchronous context
    156     with start_blocking_portal() as portal:
--> 157         return portal.call(begin_run)
    158 else:
    159     # An event loop is not running so we will create one
    160     return anyio.run(begin_run)

File ~/mambaforge/envs/prefect/lib/python3.9/site-packages/anyio/from_thread.py:283, in BlockingPortal.call(self, func, *args)
    268 def call(
    269     self,
    270     func: Callable[..., Union[Coroutine[Any, Any, T_Retval], T_Retval]],
    271     *args: object
    272 ) -> T_Retval:
    273     """
    274     Call the given function in the event loop thread.
    275 
   (...)
    281 
    282     """
[...]

Issue #7039, opened 2022-09-30
Running forever when async flow calls async subflow that calls sync task


ahuang11:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When I use an async flow to call an async subflow that calls a sync task, I get a forever running flow.

### Reproduction

```python3
from prefect import flow, task

@task
def test_subtask(obj):
    return obj.get("id")

@flow
async def test_async_subflow():
    obj = {"id": 123}
    return test_subtask(obj)

@flow
async def test_async_flow():
    await test_async_subflow()

await test_async_flow()
```

### Error

```python3
Added a debug statement after Connecting to API to ensure it's not stuck connecting.

11:21:33.284 | DEBUG   | prefect.client - Connecting to API at https://api.prefect.cloud/api/accounts/
11:21:33.285 | DEBUG   | prefect.client - Started API at https://api.prefect.cloud/api/accounts/
False
11:21:33.645 | INFO    | prefect.engine - Created flow run 'gentle-cuckoo' for flow 'test-sync-flow'
11:21:33.647 | DEBUG   | Flow run 'gentle-cuckoo' - Starting 'ConcurrentTaskRunner'; submitted tasks will be run concurrently...
11:21:33.648 | DEBUG   | prefect.task_runner.concurrent - Starting task runner...
11:21:33.684 | DEBUG   | prefect.client - Connecting to API at https://api.prefect.cloud/api/accounts/f1baedc9-7ad7-4d21-b630-cf33e5f1c328/workspaces/d55b8483-9106-4fc3-841e-6c146d011adb/
11:21:33.686 | DEBUG   | prefect.client - Started API at https://api.prefect.cloud/api/accounts/
11:21:34.428 | DEBUG   | Flow run 'gentle-cuckoo' - Executing flow 'test-sync-flow' for flow run 'gentle-cuckoo'...
11:21:34.429 | DEBUG   | Flow run 'gentle-cuckoo' - Executing test_sync_flow()
11:21:34.446 | DEBUG   | Flow run 'gentle-cuckoo' - Resolving inputs to 'test-async-flow'
True
11:21:34.799 | INFO    | Flow run 'gentle-cuckoo' - Created subflow run 'bright-hound' for flow 'test-async-flow'
[...]

PR #7041, opened 2022-09-30
Fix no sync portal

ahuang11:
Closes https://github.com/PrefectHQ/prefect/issues/7040, but at the moment leads to https://github.com/PrefectHQ/prefect/issues/7039

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -157,16 +157,16 @@ def enter_flow_run_engine_from_flow_call(
         else:
             # An event loop is not running so we will create one
             return anyio.run(begin_run)
-
-    if not parent_flow_run_context.flow.isasync:
-        # Async subflow run in sync flow run
-        return run_async_from_worker_thread(begin_run)
-    elif parent_flow_run_context.flow.isasync and flow.isasync:
-        # Async subflow run in async flow run
-        return begin_run()
     else:
-        # Sync subflow run in async flow run
-        return parent_flow_run_context.sync_portal.call(begin_run)
+        if not parent_flow_run_context.flow.isasync:
+            # Async subflow run in sync flow run
+            return run_async_from_worker_thread(begin_run)
+        elif parent_flow_run_context.flow.isasync and flow.isasync:
+            # Async subflow run in async flow run
+            return begin_run()
+        else:
+            # Sync subflow run in async flow run
+            return parent_flow_run_context.sync_portal.call(begin_run)
 
 
 def enter_flow_run_engine_from_subprocess(flow_run_id: UUID) -> State:
@@ -488,6 +488,18 @@ async def create_and_begin_subflow_run(
             )
             task_runner = await stack.enter_async_context(flow.task_runner.start())
 
+            parent_sync_portal = parent_flow_run_context.sync_portal
+            if parent_sync_portal is None:
+                # for sync flow -> async subflow -> sync task
+                sync_portal = (
+                    stack.enter_context(start_blocking_portal())
+                    if flow.isasync
+                    else None
+                )
+            else:
+                # for async flow -> async subflow
+                sync_portal = parent_sync_portal
+
             terminal_state = await orchestrate_flow_run(
                 flow,
                 flow_run=flow_run,
@@ -498,7 +510,7 @@ async def create_and_begin_subflow_run(
                 client=client,
                 partial_flow_run_context=PartialModel(
                     FlowRunContext,
-                    sync_portal=parent_flow_run_context.sync_portal,
+                    sync_portal=sync_portal,
                     result_filesystem=parent_flow_run_context.result_filesystem,
                     task_runner=task_runner,
                     background_tasks=parent_flow_run_context.background_tasks,
--- a/src/prefect/client/orion.py
+++ b/src/prefect/client/orion.py
@@ -1761,6 +1761,8 @@ class OrionClient:
         # Enter the httpx client's context
         await self._exit_stack.enter_async_context(self._client)
 
+        self.logger.debug(f"Started API at {self.api_url}")
+
         self._started = True
 
         return self


src/prefect/engine.py
```diff
@@ -157,16 +157,16 @@ def enter_flow_run_engine_from_flow_call(
...
-        return parent_flow_run_context.sync_portal.call(begin_run)
+        if not parent_flow_run_context.flow.isasync:
+            # Async subflow run in sync flow run
+            return run_async_from_worker_thread(begin_run)
+        elif parent_flow_run_context.flow.isasync and flow.isasync:
+            # Async subflow run in async flow run
+            return begin_run()
+        else:
+            # Sync subflow run in async flow run
+            return parent_flow_run_context.sync_portal.call(begin_run)
```
madkinsz, line 169: Is there a change here...? Try not to make stylistic changes when changing behavior. It makes it harder to narrow down what changes are actually relevant.

ahuang11: Wanted to triple check something didn't leak out the first if clause.

ahuang11:
Turns out I can reproduce with even simpler example:
```

@task
def test_subtask():
    return 123

@flow
async def test_async_flow():
    test_subtask()

await test_async_flow()
```

ahuang11:
Not sure why `get_task_call_return_value` is able to store to `flow_run_context.task_run_futures`, but it gets lost when `begin_task_map`

src/prefect/engine.py
```diff
@@ -343,24 +343,22 @@ async def begin_flow_run(
...
         )
 
         # If the flow is async, we need to provide a portal so sync tasks can run
-        flow_run_context.sync_portal = (
-            stack.enter_context(start_blocking_portal()) if flow.isasync else None
+        flow_run_context.sync_portal = stack.enter_context(start_blocking_portal())
+
+        # Create a task group for background tasks
+        flow_run_context.background_tasks = await stack.enter_async_context(
+            anyio.create_task_group()
```
ahuang11, line 355: Moved this after `task_runner` because if not, it complains that the task runner hasn't started yet

src/prefect/engine.py
```diff
@@ -343,24 +343,22 @@ async def begin_flow_run(
...
-        flow_run_context.background_tasks = await stack.enter_async_context(
-            anyio.create_task_group()
+        flow_run_context.task_runner = await stack.enter_async_context(
+            flow.task_runner.start()
         )
 
         # If the flow is async, we need to provide a portal so sync tasks can run
-        flow_run_context.sync_portal = (
-            stack.enter_context(start_blocking_portal()) if flow.isasync else None
+        flow_run_context.sync_portal = stack.enter_context(start_blocking_portal())
```
ahuang11, line 351: Always have sync portal up; this cleaned up logic downstream, but depending on whether running `start_blocking_portal` is expensive, add back https://github.com/PrefectHQ/prefect/pull/7041/commits/7c16f0298aa84a3ce9c5b03481c3003c226436b5

ahuang11:
At the moment, this is blocked because `task_run_futures` is lost in `flow_run_context` when running `begin_task_map` even though the `id(flow_run_context.task_run_futures)` is the same inside and outside of `begin_task_map`

src/prefect/engine.py
```diff
@@ -737,8 +735,10 @@ def enter_task_run_engine(
 
     # Sync task run in async flow run
     else:
-        # Call out to the sync portal since we are not in a worker thread
-        return flow_run_context.sync_portal.call(begin_run)
+        # Submit to background tasks since we are not in a worker thread
+        flow_run_context.background_tasks.start_soon(begin_run)
+        # get futures from the context
+        return flow_run_context.task_run_futures
```
ahuang11, line 741: These should be appended and available from https://github.com/PrefectHQ/prefect/pull/7041/files#diff-ae6f1b6d423c9f28188aedf4ab0f0afbce428d61b84feed78ca9c716ce3e26caL919

However, when I execute a flow, it's empty when returned. I think it has something to do with begin_run running in the background and results not yet available.

src/prefect/engine.py
madkinsz: Huh? Background tasks shouldn't require the task runner to be started.

src/prefect/engine.py
madkinsz: Yeah, these are not populated because your function has been sent to the background

> I haven't found a way to await the results in a sync function.

That's the whole point of the blocking portal :D

src/prefect/engine.py
ahuang11: With the current changes:
```
File ~/Applications/python/prefect/src/prefect/task_runners.py:250, in ConcurrentTaskRunner.submit(self, key, call)
    249 if not self._started:
--> 250     raise RuntimeError(
    251         "The task runner must be started before submitting work."
    252     )
```

And I'm using background_tasks here because:
```
RuntimeError: Task <Task pending name='' coro=<create_task_run_then_submit() running at /Users/mz/dev/prefect/src/prefect/engine.py:940> cb=[TaskGroup._spawn.<locals>.task_done() at /opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py:726]> got Future <Future pending> attached to a different loop
```

src/prefect/engine.py
ahuang11: Right, but if the blocking portal and background tasks are in a separate loop, it leads to:
```
Right, but I based it off
```
RuntimeError: Task <Task pending name='' coro=<create_task_run_then_submit() running at /Users/mz/dev/prefect/src/prefect/engine.py:940> cb=[TaskGroup._spawn.<locals>.task_done() at /opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py:726]> got Future <Future pending> attached to a different loop
```
```

madkinsz: That looks like we're incorrectly attempting to wait for the result of a future from a different loop than it was created in. Futures need to consistently belong to a single event loop. We have some definite issues with this and it's proven a bit complicated to untangle.

src/prefect/engine.py
madkinsz: Hm interesting. Perhaps this is because of context exit, i.e. the task runner exits before the background tasks are all done if its context is entered second.

src/prefect/engine.py
ahuang11: Thanks for the insight.

Put this in `enter_task_run_engine` and `begin_task_map` and reverting back to sync_portal:
```
    import asyncio
    loop = asyncio.get_running_loop()
    loop.set_debug(True)
    print(id(loop))
```

Results in two separate loop IDs
```
4445688688
4765149456
```

So I wonder if there's a way to pass the existing event loop `sync_portal`

madkinsz: The whole point of the sync portal is to open a new event loop in a new thread. If we use the existing loop, when we send something to it it'd deadlock as the main loop in the main thread waits for the result of the call in the portal's thread which needs to execute code in the main loop.

madkinsz: (I know this hurts the brain — we can find some time for a sync chat about this soon)

ahuang11:
To add some context, this only deadlocks when using Cloud, and does not deadlock using a local server.

ahuang11:
Also, this is no longer blocking prefect-dbt; a workaround is made here: https://github.com/PrefectHQ/prefect-dbt/pull/71

abrookins:
@ahuang11, are you still actively working on this PR or can we close it?

ahuang11:
I got stuck so not actively working on this anymore. We can close this, but https://github.com/PrefectHQ/prefect/commit/7c16f0298aa84a3ce9c5b03481c3003c226436b5 might be useful for an active `sync_portal`.

PR #7047, opened 2022-09-30
Flow runs from deployments

### LINKED ISSUE #6850: Expose a utility that creates a flow run from a deployment
### First check

### Prefect Version

2.x

Currently, to create a flow run from a deployment, you need to write async code that is aware of the deployment's UUID, like this [example taken from Discourse](https://discourse.prefect.io/t/how-to-create-a-flow-run-from-deployment-orchestrator-pattern/803):

```python
import asyncio
from prefect.client import get_client


async def main():
        depl_id = "074db2e5-229a-460e-85ad-fca31b379fd2"
        response = await client.create_flow_run_from_deployment(depl_id)
        print(response)


if __name__ == "__main__":
    asyncio.run(main())
```

There are a few rough spots with this:
1. Our client is async, so using it directly requires that you await it in an async context.
2. To create a flow run from a deployment, you need the deployment's UUID. The name would be easier and make a nicer interface.
3. Linking a flow run created from a deployment to the parent flow (as a subflow) should be straightforward, but it's not.

A utility should exist that lets users create a flow run from a deployment, and it should have the following desirable qualities:
1. Creating a flow run from a deployment should work with both sync and async Python code
2. Creating a flow run from a deployment should be possible with the deployment's name
3. The method to link a flow run created from a deployment to its parent run as a subflow should be straightforward.

### Example Use

This is not at all a suggestion about how this utility should work or what its name should be:

```python
from prefect import flow, get_run_logger, create_flow_run_from_deployment

DEPLOYMENT_NAME = "my-deployment"

@flow
def no_discourse_post_needed():
    logger = get_flow_run_logger()
    logger.info("Look, no Discourse post was necessary")

    # The UI will display this flow run as a subflow of the
    # `no_discourse_post_needed` run.
    create_flow_run_from_deployment(DEPLOYMENT_NAME)
```

### Additional context

Suppose I have a Flask application and want to use my deployed Prefect flow to schedule a run immediately, but my Flask application exists in a separate repository from my flows. If the utility to create a flow run from a deployment exists in the `prefect` repository, then I need to add Prefect as a dependency in my Flask app -- but I'd rather not do this because I'd like to keep my API Docker images slim.

Is it possible that this functionality could exist in an external package that Prefect depends on so that environments that don't want to depend on Prefect can still schedule flow runs with a deployment?

**Note**: If I schedule a flow run in my Flask app, I probably want to save the flow run ID and use it to check the status of the run, so that I can tell users of my app that the job is processing or completed.

anticorrelator:
Adds a utility to create flow runs from a deployment and waits for a specified amount of time before returning the FlowRun object. Users can both override the default Deployment parameters as well as specify unlimited (or no) polling before returning from this function.

Closes #6850

### Checklist


src/prefect/coordination.py
```diff
@@ -0,0 +1,119 @@
...
+                flow_state = flow_run.json()["state"]["type"]
+            except KeyError:
+                raise MissingFlowRunError("Error polling flow run")
+
+            if flow_state in TERMINAL_STATE_STRINGS:
+                return flow_state
+
+        raise DeploymentTimeout(
+            f"Deployment run did not terminate and is in the {flow_state} state"
+        )
```
anna-geller, line 97: Adding handling for this edge case for max_polls <= 0 would be a nice extra feature (fire-and-forget mode) + would prevent a bug where flow_state can't be reached

```suggestion
    This function will block until the deployment run enters a terminal state or until
    the polling duration has been exceeded. If max_polls is not a positive integer, 
    flow run is triggered without waiting for completion (fire-and-forget).
    """

    with _minimal_client() as client:
        body = {"parameters": parameters}

        flow_run_res = client.post(
            f"/deployments/name/{deployment_name}/schedule_flow_run",
            json=body,
        )
        flow_run_id = flow_run_res.json()
        
        if max_polls <= 0:
            pass
        
        for poll in range(max_polls):
            time.sleep(poll_interval)
            try:
                flow_run = client.get(f"/flow_runs/{flow_run_id}")
                flow_state = flow_run.json()["state"]["type"]
            except KeyError:
                raise MissingFlowRunError("Error polling flow run")

            if flow_state in TERMINAL_STATE_STRINGS:
                return flow_state

            raise DeploymentTimeout(
                f"Deployment run did not terminate and is in the {flow_state} state"
            )
```

anna-geller (changes requested):
QAed and it works well 🎉  

One thing to consider is that currently, both `run_deployment` and `schedule_deployment` result in **no** auto-scheduled flag being set, some might expect that `schedule_deployment` leads to `auto_scheduled=True`. The current behavior seems correct to me since those don't run on a regular cadence (the `schedule_deployment` is technically a single ad-hoc run but just not executed immediately) but this is a matter of interpretation

madkinsz:
@anna-geller `auto_scheduled=True` is used for us to determine if runs should be deleted when a schedule is changed or toggled so we don't want to use it here. Good thing to acknowledge though!

abrookins (changes requested):
Adding some notes here from our sync review and after thinking about it a bit:
1. For the top-level package, can we use `experimental`? This would open up some exciting possibilities for shipping valuable features to users before we've completely designed how they should work.
2. Can the two utility functions that users call have both a sync and an async form?
3. I think that `schedule_flow_run_from_deployment` should use a Pydantic model to validate `parameters`. I don't think the Pydantic model has to be an existing schema to be useful -- simply having one would eliminate manual error-handling that we would need to add here while producing validation errors consistent with the others that our API returns. Having said all that, I'm not actually familiar with what values we would accept here -- another reason that a Pydantic model could be useful, as it would produce an OpenAPI specification for this field.
4. The decision to build feature without using the existing client and by building an alternative, simpler server-side API endpoint is notable. I think these are interesting choices to explore given open questions we have about how our events client should work, especially if we can house this code in an `experiments` package.

abrookins:
Ah, right, and thank you @madkinsz  -- I misunderstood that the schema for that payload was deployment-dependent! I agree that we shouldn't try to validate that in this PR.

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,52 @@
+import time
+
+from prefect.client.orion import get_client
+from prefect.utilities.asyncutils import sync_compatible
+
+
+@sync_compatible
+async def run_deployment(
```
cicdw, line 8: Not necessary for the issue itself but FYI you could expose a `scheduled_start_time: datetime.datetime` field here and then pass `state = schemas.states.Scheduled(scheduled_time=scheduled_start_time)` to the `create_flow_run_from_deployment` client method in order to expose the ability to optionally schedule the run for the future.

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,52 @@
...
+    Runs a deployment immediately and returns a FlowRun object.
+
+    This function will return when the created flow run enters a terminal state or until
+    the polling duration has been exceeded.
+
+    Args:
+        name: The deployment name in the form: '<flow-name>/<deployment-name>'
+        parameters: Parameter overrides for this flow run. Merged with the deployment
+            defaults
+        max_polls: The maxinum number of times to poll the flow run before returning.
```
cicdw, line 24:
```suggestion
        max_polls: The maximum number of times to poll the flow run before returning.
```

src/prefect/orion/models/deployments.py
```diff
@@ -436,6 +437,10 @@ async def _generate_scheduled_flow_runs(
         n=max_runs, start=start_time, end=end_time
     )
 
+    tags = deployment.tags
+    if auto_scheduled:
+        tags = ["auto-scheduled"] + tags
```
cicdw, line 442: Is this worth writing a test for?  It doesn't seem like you used this logic for this PR but either way I agree your change makes for a much better design

cicdw (commented):
a few minor notes, this LGTM otherwise!

src/prefect/client/utilities.py
anticorrelator: 👍

src/prefect/client/utilities.py
anticorrelator: ah good catch, this is a good idea

src/prefect/orion/models/deployments.py
anticorrelator: yeah I think this deserves a test, I used it for the previous implementation but left it

anna-geller (approved):
Works great! 

for docs - code used for QA
```python
from prefect.client.utilities import run_deployment

# run parametrized flow from deployment and wait for its completion before starting the next task
flow_run_obj = run_deployment(
    "parametrized/dev",
    parameters=dict(user="QA Run - waiting"),
)

# run parametrized flow from deployment and wait for its completion before starting the next task
run_deployment(
    "parametrized/dev",
    max_polls=-1,
    parameters=dict(user="QA Run - waiting forever until completion - no poll limits"),
)
# run parametrized flow from deployment and don't wait for completion
run_deployment(
    "parametrized/dev",
    max_polls=0,
    parameters=dict(user="QA Run - fire-and-forget 🔥"),
)

# run parametrized flow from deployment and don't wait for completion
not_finished_flow_run = run_deployment(
    "parametrized/dev",
    max_polls=2,
    poll_interval=2,
    parameters=dict(user="QA Run - very short polling"),
)
print(not_finished_flow_run.state)  # Pending
```

abrookins (approved):
Looks good to me!

cicdw (approved):

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,62 @@
...
+from prefect.client.orion import get_client, schemas
+from prefect.utilities.asyncutils import sync_compatible
+
+
+@sync_compatible
+async def run_deployment(
+    name: str,
+    parameters: dict = None,
+    scheduled_time: datetime = None,
+    max_polls: int = 60,
```
madkinsz, line 15: In the past, I've preferred `timeout` or `timeout_seconds` where you can check the elapsed time at the start of each poll interval. This way, the user does not need to do math and it's clear that if you set a timeout you may get a timeout error.

It'd also be great to have it default to `-1` or `None` to to match other functions that can timeout (e.g. they'll run to completion unless otherwise specified). I could see the argument for `0` or returning immediately as well. Setting a default timeout doesn't feel as helpful though since it will not be clear on first-use to a user that it may timeout.

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,62 @@
...
+            deployment.id,
+            state=schemas.states.Scheduled(scheduled_time=scheduled_time),
+            parameters=parameters,
+        )
+
+        flow_run_id = flow_run.id
+
+        polls = 0
+        while max_polls == -1 or polls < max_polls:
+            time.sleep(poll_interval)
```
madkinsz, line 54: We'll want to use anyio here
```suggestion
            await anyio.sleep(poll_interval)
```

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,62 @@
...
+        flow_run = await client.create_flow_run_from_deployment(
+            deployment.id,
+            state=schemas.states.Scheduled(scheduled_time=scheduled_time),
+            parameters=parameters,
+        )
+
+        flow_run_id = flow_run.id
+
+        polls = 0
+        while max_polls == -1 or polls < max_polls:
```
madkinsz, line 53: I've also realized that since this is async you can just use an AnyIO timeout context which is easier to read and makes implementing seconds as a parameter very simpler.

src/prefect/client/utilities.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    if max_polls < -1:
+        raise ValueError(
+            "`max_polls` must be -1 (unlimited polling), 0 (return immediately) or any positive integer"
+        )
+
+    if scheduled_time is None:
+        scheduled_time = pendulum.now("UTC")
+
+    async with get_client() as client:
```
madkinsz, line 42: If you use `inject_client` then we won't create a new client when this is called in a flow or a task (this is a recent optimization) which is more performant.

madkinsz (commented):
Exciting! Mostly asking for a change around timeout handling.

src/prefect/client/utilities.py
anna-geller: Right, default of -1 makes sense -- if we poll, we should by default poll until completion 👍

madkinsz:
Thanks for working through so many people's feedback!

Issue #6549, opened 2022-08-24
Settings are missing from the API reference


madkinsz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

The API reference for settings at https://docs.prefect.io/api-ref/prefect/settings/#prefect.settings.Settings should include all of the settings and their descriptions but they are not present. I believe this is also breaking all links to specific settings, e.g. https://docs.prefect.io/faq/?h=postgres#what-external-requirements-does-prefect-have

The original implementation at https://github.com/PrefectHQ/orion/pull/1014 was careful to ensure that settings were visible in the API reference.

### Reproduction

```python
See Prefect documentation build steps.
```

### Error

No error.

### Versions

N/A

dylanmcreynolds:
Just another note, I was following this [document](https://docs.prefect.io/faq/?h=postgres#how-do-i-choose-between-sqlite-and-postgres). It would be pretty handy if this had a reference back to this documentation about construction the database connection string.

madkinsz:
Yep that document is supposed to have a link but it's broken due to this issue.

madkinsz:
@tpdorsey can you look into this? It's not easy to see setting help strings otherwise. If not, let me know and I'll find someone with some time.

tpdorsey:
I think the issue is that the docstring for settings assignments needs to be specified as an attribute docstring. The parser can't read the description directly. I'll have a PR to demonstrate shortly.

https://peps.python.org/pep-0258/#attribute-docstrings

PR #7067, opened 2022-10-05
Update docstrings to render settings in API reference

tpdorsey:
Adds attribute docstrings to settings, ensuring that they appear correctly in rendered API documentation.

Minor formatting corrections to settings assignments.

Update mkdocstrings options.

Closes https://github.com/PrefectHQ/prefect/issues/6549

### Checklist

### DIFF
--- a/docs/api-ref/prefect/settings.md
+++ b/docs/api-ref/prefect/settings.md
@@ -8,8 +8,6 @@ tags:
 ---
 
 ::: prefect.settings
-    selection:
-      inherited_members: true
+    options:
       filters: ["!^_", "!Config", "!copy", "!dict", "!json"]
-    rendering:
       members_order: source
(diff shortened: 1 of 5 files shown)


madkinsz:
@tpdorsey We didn't need attribute docstrings for these to render correctly before. I'm down to make the change but I'm curious why this is? I'd also like to drop the existing descriptions in favor of this style if we're going this route instead of duplicating them.

tpdorsey:
@madkinsz I'm not sure why the build behavior changed. My reading of the handler docs and our templates suggests that it should _not_ have worked previously. Using attribute docstrings falls back on default behavior of the handler, which I think is a good thing. I am admittedly still learning the deeper workings here.

ahuang11:
Maybe something related to mkdocstrings[python] vs mkdocstrings legacy

I just encountered this in prefect collection generator and I found the option `show_if_no_docstring` from
https://mkdocstrings.github.io/python-legacy/usage/#globallocal-options

I used in a `credentials.md` (rather than `mkdocs.yml`):
```
::: prefect_databricks.credentials
    selection:
      members:
        - _generated
```

madkinsz (approved):

PR #6926, opened 2022-09-22
Update `State.result` to be async

madkinsz:
Part of #6908.

This pull request updates `State.result` to be an async method. This is a breaking change for **users calling `State.result` directly in async tasks and flows**. I believe this will be a small group of users. 

This will **not be a breaking change** for:

- Users using `Future.result` or calling flows/tasks directly
- Users with synchronous flows
- Users relying on automatic resolution of states e.g. just calling their tasks/flows or passing futures around 

The result method must be async to support retrieval of the result's file system block.

### Example

```python
from prefect import flow, task

@task
async def my_task(x = 0):
    return 1 + x

@flow
async def my_flow():

    # Not breaking: Calling to get result
    assert await my_task() == 1

    # Not breaking: Getting the result from the future
    future = await my_task.submit()
    assert await future.result() == 1

    # Not breaking: Passing the future or state downstream
    assert await my_task(future) == 2
    state = await future.wait()
    assert await my_task(state) == 2

    # Breaking: Use of `return_state` then `State.result`
    state = await my_task(return_state=True)
    assert await state.result() == 1  # Now requires await

    # Breaking: Use of `wait` then `State.result`
    future = await my_task.submit()
    state = await future.wait()
    assert await state.result() == 1  # Now requires await

    return 1
```

### Checklist

- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

PR #7071, opened 2022-10-05
Add async support to `State.result()`

madkinsz:
Supersedes https://github.com/PrefectHQ/prefect/pull/6926; includes backwards compatibility.

`State.result()` needs async support for automatic retrieval of results that have been persisted. For our synchronous users, we can change the method to be `async` and just use the `sync_compatible `decorator and they will see no change in behavior but we will be able to make async calls. For our asynchronous users, this would be a breaking change since they would need to await the call. To avoid making a breaking change, we will _not_ make asynchronous calls if the user calls `State.result()` from an async context unless they opt-in. If they opt-in, we will return the coroutine for them to await. If they do not opt-in (or explicitly opt-out), we will display a warning:

> State.result() was called from an async context but not awaited. This method will be updated to return a coroutine by default in the future. Pass `fetch=True` and `await` the call to get rid of this warning.

See the [new documentation](https://deploy-preview-7071--prefect-orion.netlify.app/concepts/results/#working-with-async-results) which covers these changes for user-facing messaging.

Affected users may opt-in per call with `State.result(fetch=True)` or globally with the `PREFECT_ASYNC_FETCH_STATE_RESULT` setting. The defaults for these settings can change after a deprecation cycle.

### Example

```python
from prefect.settings import temporary_settings, PREFECT_OPT_IN_ASYNC_STATE_RESULT

@task
async def my_task(x = 0):
    return 1 + x

@flow
async def my_flow():

    # Not breaking: Calling to get result
    assert await my_task() == 1

    # Not breaking: Getting the result from the future
    future = await my_task.submit()
    assert await future.result() == 1

    # Not breaking: Passing the future or state downstream
    assert await my_task(future) == 2
    state = await future.wait()
    assert await my_task(state) == 2

    # Would be breaking: Use of `return_state` then `State.result`
    state = await my_task(return_state=True)
    # The result type will be returned; a warning will be displayed
    assert state.result() == ResultReference(...)

    # Would be breaking: Use of `wait` then `State.result`
    future = await my_task.submit()
    state = await future.wait()
    # The result type will be returned; a warning will be displayed
    assert state.result() == ResultReference(...)

    # Opt-in to new behavior
    state = await my_task(return_state=True)
    assert await state.result(fetch=True) == 1

    # Opt-in to new behavior globally
    with temporary_settings({PREFECT_OPT_IN_ASYNC_STATE_RESULT: True}):
          state = await my_task(return_state=True)
          # No `fetch` needed
          assert await state.result() == 1

    return 1
```

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -241,12 +241,14 @@ asyncio.run(main())
 !!! important "Resolving results"
     In Prefect 2.6.0, we added automatic retrieval of persisted results.
     Prior to this version, `State.result()` did not require an `await`.
-    For backwards compatibility, when used from an asynchronous context, `State.result()` will return a raw result type.
+    For backwards compatibility, when used from an asynchronous context, `State.result()` returns a raw result type.
+    
     You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
     If you would like this behavior to be used automatically, you may enable the `PREFECT_ASYNC_FETCH_STATE_RESULT` setting.
-    If you do not opt-in to this behavior you will see a warning.
+    If you do not opt-in to this behavior, you will see a warning.
+    
     You may also opt-out by setting `fetch=False`.
-    This will silence the warning but you will need to retrieve your result manually from the result type.
+    This will silence the warning, but you will need to retrieve your result manually from the result type.
 
 When submitting tasks to a runner, the result can be retreived with the `Future.result()` method:
 
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -227,8 +227,9 @@ async def my_task():
 
 @flow
 async def my_flow():
-    state = my_task(return_state=True)
-    return state.result() + 1
+    state = await my_task(return_state=True)
+    result = await state.result(fetch=True)
+    return result + 1
 
 async def main():
     state = await my_flow(return_state=True)
@@ -241,10 +242,11 @@ asyncio.run(main())
     In Prefect 2.6.0, we added automatic retrieval of persisted results.
     Prior to this version, `State.result()` did not require an `await`.
     For backwards compatibility, when used from an asynchronous context, `State.result()` will return a raw result type.
-    You may opt-in to the new behavior by passing `fetch=True`.
-    If you would like this behavior to be used everywhere without passing `fetch=True`, you may set the 
-    `PREFECT_OPT_IN_ASYNC_STATE_RESULT` variable.
-
+    You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
+    If you would like this behavior to be used automatically, you may enable the `PREFECT_OPT_IN_ASYNC_STATE_RESULT` setting.
+    If you do not opt-in to this behavior you will see a warning.
+    You may also opt-out by setting `fetch=False`.
+    This will silence the warning but you will need to retrieve your result manually from the result type.
 
 When submitting tasks to a runner, the result can be retreived with the `Future.result()` method:
 
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -239,7 +239,7 @@ asyncio.run(main())
 ```
 
 !!! important "Resolving results"
-    In Prefect 2.6.0, we added automatic retrieval of persisted results.
+    Prefect 2.6.0 added automatic retrieval of persisted results.
     Prior to this version, `State.result()` did not require an `await`.
     For backwards compatibility, when used from an asynchronous context, `State.result()` returns a raw result type.
     
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -194,6 +194,79 @@ def my_flow():
 my_flow()
 ```
 
+### Working with async results
+
+When **calling** flows or tasks, the result is returned directly:
+
+```python
+import asyncio
+from prefect import flow, task
+
+@task
+async def my_task():
+    return 1
+
+@flow
+async def my_flow():
+    task_result = await my_task()
+    return task_result + 1
+
+result = asyncio.run(my_flow())
+assert result == 2
+```
+
+When working with flow and task states, the result can be retreived with the `State.result()` method:
+
+```python
+import asyncio
+from prefect import flow, task
+
+@task
+async def my_task():
+    return 1
+
+@flow
+async def my_flow():
+    state = my_task(return_state=True)
+    return state.result() + 1
+
+async def main():
+    state = await my_flow(return_state=True)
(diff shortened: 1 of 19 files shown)


tests/results/test_result_fetch.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # A result type is returned
+    assert isinstance(result, ResultLiteral)
+    assert await result.get() is True
+
+
+async def test_async_result_warnings_are_not_raised_by_engine():
+    # Since most of our tests are run with the opt-in globally enabled, this test
+    # covers a bunch of features to cover remaining cases where we may internally
+    # call `State.result` incorrectly.
```
madkinsz, line 40: Not really enthused about this kind of test but it'll be useful as I update the engine to use the new result types.

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+
+!!! important "Resolving results"
+    In Prefect 2.6.0, we added automatic retrieval of persisted results.
+    Prior to this version, `State.result()` did not require an `await`.
+    For backwards compatibility, when used from an asynchronous context, `State.result()` will return a raw result type.
+    You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
+    If you would like this behavior to be used automatically, you may enable the `PREFECT_ASYNC_FETCH_STATE_RESULT` setting.
+    If you do not opt-in to this behavior you will see a warning.
+    You may also opt-out by setting `fetch=False`.
+    This will silence the warning but you will need to retrieve your result manually from the result type.
```
tpdorsey, line 249:
```suggestion
    This will silence the warning, but you will need to retrieve your result manually from the result type.
```

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+    assert await state.result(fetch=True) == 2
+
+asyncio.run(main())
+```
+
+!!! important "Resolving results"
+    In Prefect 2.6.0, we added automatic retrieval of persisted results.
+    Prior to this version, `State.result()` did not require an `await`.
+    For backwards compatibility, when used from an asynchronous context, `State.result()` will return a raw result type.
+    You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
```
tpdorsey, line 245:
```suggestion
    
    You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
```

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+    If you would like this behavior to be used automatically, you may enable the `PREFECT_ASYNC_FETCH_STATE_RESULT` setting.
+    If you do not opt-in to this behavior you will see a warning.
+    You may also opt-out by setting `fetch=False`.
```
tpdorsey, line 248:
```suggestion
    
    You may also opt-out by setting `fetch=False`.
```

docs/concepts/results.md
tpdorsey: Break up concepts into some paragraphs for easier scanning.

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+
+async def main():
+    state = await my_flow(return_state=True)
+    assert await state.result(fetch=True) == 2
+
+asyncio.run(main())
+```
+
+!!! important "Resolving results"
+    In Prefect 2.6.0, we added automatic retrieval of persisted results.
```
billpalombi, line 242:
```suggestion
    Prefect 2.6.0 added automatic retrieval of persisted results.
```

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+    Prior to this version, `State.result()` did not require an `await`.
+    For backwards compatibility, when used from an asynchronous context, `State.result()` will return a raw result type.
```
billpalombi, line 244:
```suggestion
    For backwards compatibility, when used from an asynchronous context, `State.result()` returns a raw result type.
```

docs/concepts/results.md
```diff
@@ -194,6 +194,81 @@ def my_flow():
...
+    You may opt-in to the new behavior by passing `fetch=True` as shown in the example above.
+    If you would like this behavior to be used automatically, you may enable the `PREFECT_ASYNC_FETCH_STATE_RESULT` setting.
+    If you do not opt-in to this behavior you will see a warning.
```
billpalombi, line 247:
```suggestion
    If you do not opt-in to this behavior, you will see a warning.
```

billpalombi (approved):
LGTM, thanks for figuring this out. I proposed a couple of minor revisions to the callout.

src/prefect/settings.py
```diff
@@ -417,6 +417,24 @@ def warn_on_database_password_value_without_usage(values):
     prefetched. Defaults to `10`.""",
 )
 
+PREFECT_ASYNC_FETCH_STATE_RESULT = Setting(
+    bool,
+    default=False,
+    description=textwrap.dedent(
+        """
+        Determines whether `State.result()` fetches results automatically or not.
```
tpdorsey, line 425: See https://github.com/PrefectHQ/prefect/pull/7067 RE using attribute docstrings for the API docs

tpdorsey (approved):
Minor suggested doc edits. Looks great. Ty

src/prefect/settings.py
madkinsz: Thanks!

src/prefect/utilities/asyncutils.py
```diff
@@ -182,26 +181,8 @@ def sync_compatible(async_fn: T) -> T:
...
-                # In the main thread but call was made from a sync method
-                raise RuntimeError(
-                    "A 'sync_compatible' method was called from a context that was "
-                    "previously async but is now sync. The sync call must be changed "
-                    "to run in a worker thread to support sending the coroutine for "
-                    f"{async_fn.__name__!r} to the main thread."
-                )
-
+            # In the main async context; return the coro for them to await
+            return async_fn(*args, **kwargs)
```
abrookins, line 185: Interesting -- does this resolve potential async issues outside of the change in this PR?

abrookins (approved):
This looks good! I applaud you for finding a way to preserve backward compatibility while also moving the product forward. 👏

src/prefect/utilities/asyncutils.py
madkinsz: Ah this change was cherry-picked into #7073 for independent review. I discuss it more there, but tldr not really.

PR #7081, opened 2022-10-06
Update `run_deployment` to link created runs to the parent flow

madkinsz:
Update `run_deployment` to link the created flow run to the parent flow run. The parent flow run is detected by checking if a flow run context is present. When found, we will create a dummy task run in the parent flow run then link it to the child. Flow runs created with `run_deployment` appear on the subflows page and radar of the parent flow. Upstream task results used as parameters for the flow run will also result in a visible dependency.

### Checklist


src/prefect/deployments.py
```diff
@@ -68,11 +69,44 @@ async def run_deployment(
...
+            version=flow_run_ctx.flow.version,
+        )
+        parent_task_run = await client.create_task_run(
+            task=dummy_task,
+            flow_run_id=flow_run_ctx.flow_run.id,
+            dynamic_key=_dynamic_key_for_task_run(flow_run_ctx, dummy_task),
+            task_inputs=task_inputs,
+            state=Pending(),
+        )
+        parent_task_run_id = parent_task_run.id
```
madkinsz, line 100: I'd like to consolidate this logic somewhere but would prefer to do so independently of this change.

anna-geller (approved):
This is fantastic, so grateful for this PR!

image

I QAed it and it works perfectly
image

image

src/prefect/deployments.py
```diff
@@ -68,11 +70,47 @@ async def run_deployment(
...
+            _dynamic_key_for_task_run,
+            collect_task_run_inputs,
+        )
+
+        task_inputs = {
+            k: await collect_task_run_inputs(v) for k, v in parameters.items()
+        }
+
+        # Generate a task in the parent flow run to represent the result of the subflow
+        dummy_task = Task(
```
anticorrelator, line 91: not that it matters but I always thought that "virtual tasks" are a better name for these

anticorrelator (changes requested):
lgtm! but I think we need to bump the API patch version first

src/prefect/deployments.py
madkinsz: :) something to consider when we create an abstract function for this logic

anticorrelator (approved):

PR #7090, opened 2022-10-06
Improve client stability

### LINKED ISSUE #6996: Proposal to add retries on httpx exceptions
### First check

### Prefect Version

2.x

Not Implemented

This is a proposal to add retry handling on `post`, `get`, and other methods of `PrefectHttpxClient`
https://github.com/PrefectHQ/prefect/blob/1a43c7d7160e7f27fb3ae5eb01552c648c3c017d/src/prefect/client/base.py#L141

As on unstable connection environments, this can force flow runs to fail. I usually encounter `ReadError` however I believe that a retry on `NetworkError` would be more suitable as it handles more cases 

https://www.python-httpx.org/exceptions/ 

### Example Use

_No response_

### Additional context

_No response_

zangell44:
This PR tweaks settings on the `orion` httpx client to improve stability of highly concurrent requests, like running a large number of mapped tasks.

Part of ongoing investigation of https://github.com/PrefectHQ/nebula/issues/1727.

Closes #6996 

### Checklist


src/prefect/client/orion.py
```diff
@@ -185,9 +215,8 @@ async def api_healthcheck(self) -> Optional[Exception]:
         If successful, returns `None`.
         """
         try:
-            with anyio.fail_after(10):
-                await self._client.get("/health")
-                return None
+            await self._client.get("/health")
```
zangell44, line 218: I think it's a bad idea to handle this ourselves with `fail_after`. Best case scenario is we fail a bit faster. Worst case (and I think more common) scenario is we disrupt `httpx`'s error handling

requirements.txt
```diff
@@ -15,7 +15,7 @@ docker >= 4.0
 fastapi >= 0.70
 fsspec >= 2022.5.0
 griffe >= 0.20.0
-httpx >= 0.22
+httpx[http2] >= 0.23
```
chrisguidry, line 18: 💥

src/prefect/client/orion.py
```diff
@@ -163,6 +160,39 @@ def __init__(
...
+        httpx_settings.setdefault(
+            "limits",
+            httpx.Limits(
+                # We see instability when allowing the client to open many connections at once.
+                # Limiting concurrency results in more stable performance.
+                max_connections=16,
+                max_keepalive_connections=8,
+                # The Prefect Cloud LB will keep connections alive for 30s.
+                # Only allow the client to keep connections alive for 25s.
+                keepalive_expiry=25,
```
chrisguidry, line 183: 🥇

src/prefect/client/orion.py
chrisguidry: +1

src/prefect/client/orion.py
madkinsz: Should we consider making this configurable or default to a different value if the user is not using Cloud?

src/prefect/client/orion.py
madkinsz: I agree

src/prefect/client/orion.py
```diff
@@ -163,6 +183,17 @@ def __init__(
...
             )
 
+        # See https://www.python-httpx.org/advanced/#timeout-configuration
+        httpx_settings.setdefault(
+            "timeout",
+            httpx.Timeout(
+                connect=PREFECT_API_REQUEST_TIMEOUT.value(),
+                read=PREFECT_API_REQUEST_TIMEOUT.value(),
+                write=PREFECT_API_REQUEST_TIMEOUT.value(),
+                pool=PREFECT_API_REQUEST_TIMEOUT.value(),
```
madkinsz, line 193: Fyi I don't think this changes anything as `timeout=X` is the same as `timeout=Timeout(a=X, b=X, ...)`

src/prefect/client/orion.py
```diff
@@ -163,6 +160,39 @@ def __init__(
...
+                max_connections=16,
+                max_keepalive_connections=8,
+                # The Prefect Cloud LB will keep connections alive for 30s.
+                # Only allow the client to keep connections alive for 25s.
+                keepalive_expiry=25,
+            ),
+        )
+        # See https://www.python-httpx.org/advanced/#custom-transports
+        # `retries` specifies the number of retries on connection errors
+        httpx_settings.setdefault("transport", httpx.AsyncHTTPTransport(retries=3))
```
BitTheByte: @zangell44 is it possible to have a configurable number or retires instead of the hardcoded 3 times?

src/prefect/client/orion.py
```diff
@@ -152,6 +149,29 @@ def __init__(
...
+                    max_connections=16,
+                    max_keepalive_connections=8,
+                    # The Prefect Cloud LB will keep connections alive for 30s.
+                    # Only allow the client to keep connections alive for 25s.
+                    keepalive_expiry=25,
+                ),
+            )
+            # See https://www.python-httpx.org/advanced/#custom-transports
+            # `retries` specifies the number of retries on connection errors
+            httpx_settings.setdefault("transport", httpx.AsyncHTTPTransport(retries=3))
```
j-tr, line 167: it seems to me that this prevents that the other transport related parameters such as connection limits and http2 settings are properly applied.
httpx only creates a new transport object if no such object is provided as a parameter (see [https://github.com/encode/httpx/blob/master/httpx/_client.py#L1439](https://github.com/encode/httpx/blob/master/httpx/_client.py#L1439)). 
i don't see how one could manipulate the retry behavior other than via the transport parameter. so this could be possibly fixed by moving the limits and http2 setting definition into the transport constructor.

PR #6908, opened 2022-09-21
Tracking: Add first-class results

madkinsz:
Tracking branch for consolidation of result configuration changes.

Includes:
- #6895
- #6899 
- #6896
- #6900
- #6973
- #7004
- #7044
- #7071
- #7048
- #7094
- #7103
- #7108 
- #7106
- #7128
- #7129

Related changes, but into `main`:
- #6829 
- #6813 
- #6857 
- #6859 
- #6847 
- #6992
- #7073 

Planned:
- Add option to reduce memory consumption

Future:
- Ensure task run results persisted to local file systems on remote workers respect relative paths
- Investigate storing results after return from the remote worker for task runs with local file systems

May fix:
- https://github.com/PrefectHQ/prefect/issues/6845
- https://github.com/PrefectHQ/prefect/issues/6086

Unblocks:
- https://github.com/PrefectHQ/prefect/issues/5841

Closes issues:
- Closes https://github.com/PrefectHQ/prefect/issues/6040
- Closes https://github.com/PrefectHQ/prefect/issues/6397
- Closes https://github.com/PrefectHQ/prefect/issues/5888
- Closes https://github.com/PrefectHQ/prefect/issues/7061

PR #7094, opened 2022-10-06
Update engine to use new results

madkinsz:
Hooks everything up! Everywhere in the engine where we were creating `DataDocument` types, we are now creating  new result types. 

The largest changes here are dealing with cases where we previously would pickle exceptions and attach them to states. We do this all over the place — sometimes where we do not have access to the user's result settings. To support those cases, I've refactored failed/crashed results to allow generation of an exception from a string. This means we do not need to actually serialize an exception to report a failure, e.g. when an agent fails to submit a flow run.

Unlike the other pull requests in #6908 this one is a little sprawling — at some point everything's got to switch over and little edge cases need to be addressed. I think this makes this a good candidate for synchronous review.

I have not added tests here covering usage of results. Instead, I've focused on fixing the existing tests which make extensive usage of results anyway. I'm going to add dedicated tests in a follow-up pull request but want to get this merged ASAP to facilitate QA of this feature.

### DIFF
--- a/src/prefect/client/schemas.py
+++ b/src/prefect/client/schemas.py
@@ -140,6 +140,13 @@ class State(schemas.states.State.subclass(exclude_fields=["data"]), Generic[R]):
             return get_state_result(self, raise_on_failure=raise_on_failure)
 
     def to_state_create(self) -> schemas.actions.StateCreate:
+        """
+        Convert this state to a `StateCreate` type which can be used to set the state of
+        a run in the API.
+
+        This method will drop this state's `data` if it is not a result type. Only
+        results should be sent to the API. Other data is only available locally.
+        """
         from prefect.results import BaseResult
 
         return schemas.actions.StateCreate(
--- a/src/prefect/client/schemas.py
+++ b/src/prefect/client/schemas.py
@@ -1,12 +1,12 @@
 import datetime
 import warnings
-from typing import TYPE_CHECKING, Generic, Optional, Type, TypeVar, Union, overload
+from typing import TYPE_CHECKING, Any, Generic, Optional, Type, TypeVar, Union, overload
 
 from pydantic import Field
 
 from prefect.orion import schemas
 from prefect.settings import PREFECT_ASYNC_FETCH_STATE_RESULT
-from prefect.utilities.asyncutils import in_async_main_thread, sync_compatible
+from prefect.utilities.asyncutils import in_async_main_thread
 
 if TYPE_CHECKING:
     from prefect.deprecated.data_documents import DataDocument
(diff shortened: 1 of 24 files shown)


abrookins (approved):
Wow, there's a lot here! I don't think I can give a detailed review, but I read through parts and never felt shocked. 😅 I'm more interested in getting this merged and into some users' hands to try it out!

src/prefect/client/schemas.py
```diff
@@ -134,25 +135,20 @@ def result(self, raise_on_failure: bool = True, fetch: Optional[bool] = None):
...
-            return result_from_state_with_data_document(
-                self, raise_on_failure=raise_on_failure
-            )
-        elif isinstance(self.data, BaseResult):
-            return await self.data.get()
-        else:
-            raise ValueError(
-                f"State data is of unknown result type {type(self.data).__name__!r}."
-            )
+    def to_state_create(self) -> schemas.actions.StateCreate:
```
madkinsz, line 142: Add docstring

src/prefect/engine.py
```diff
@@ -1096,6 +1063,8 @@ async def begin_task_run(
                 anyio.create_task_group()
             )
 
+        await stack.enter_async_context(report_task_run_crashes(task_run, client))
```
madkinsz, line 1066: Investigate if this change can be split out

src/prefect/engine.py
```diff
@@ -1319,24 +1257,30 @@ async def wait_for_task_runs_and_report_crashes(
         if not state.type == StateType.CRASHED:
             continue
 
-        exception = await state.result(raise_on_failure=False, fetch=True)
+        exception = await get_state_exception(state)
```
madkinsz, line 1260: Figure out why we call this here instead of `state.result`

src/prefect/exceptions.py
```diff
@@ -56,6 +56,22 @@ class PrefectException(Exception):
...
+    """
+
+
+class FailedRun(PrefectException):
+    """
+    Raised when the result from a failed run is retrieved and an exception is not
+    attached.
+
+    This occurs when a string is attached to the state instead of an exception
+    or the attached exception is a `BaseException` and is not safe to raise.
```
madkinsz, line 71: Check if removed?

src/prefect/states.py
```diff
@@ -121,51 +174,124 @@ async def return_value_to_state(result: Any, serializer: str = "cloudpickle") ->
...
     else:
         raise TypeError(
-            f"Unexpected result for failure state: {result!r} —— "
+            f"Unexpected result for failed state: {result!r} —— "
             f"{type(result).__name__} cannot be resolved into an exception"
         )
 
 
+@sync_compatible
+async def raise_state_exception(state: State) -> None:
```
madkinsz, line 285: Determine if this is used?

src/prefect/engine.py
madkinsz: No. This is needed because now we raise base exceptions instead of returning them.

PR #7099, opened 2022-10-06
Agents with dynamic work queue matching

### LINKED ISSUE #5623: Add toggle for automatic concurrency limit of one run per deployment
## Description
Currently, if a user wants to ensure that there is no more than one active flow run from a specific deployment, they have to create a work queue with a concurrency limit of 1 and spin up a separate agent process for each such work queue. Implementing this for many flows becomes unmanageable.

## Excerpt from the request made via Slack

[**View in #prefect-community on Slack**](https://slack.com/archives/CL09KU1K7/p1648631937216979)

![Jacobo_Blanco] **@Jacobo_Blanco:** One pattern that emerges very often for us is Singleton flows (For the sake of clarity, only one run of this specific flow can ever be active at any one time). The other hidden constraint is that due to Change Management/Security/Auditing requirements that are in place it’s not a simple procedure to log into the EC2 instance and execute commands.

First assumption check: I assume that the agent commands have to be executed on the instance, is that right?

![Anna_Geller] **@Anna_Geller:** correct, they must be executed on the agent

regarding singleton-flow-runs: I believe we may add a feature allowing to have only one active flow run of each specific flow without having to explicitly set a work-queue concurrency limit of 1 for each of them

![Jacobo_Blanco] **@Jacobo_Blanco:** If that is the case then this covers the biggest pain point. It would be very unproductive to have to create 50 queues and agents for 50 singleton flows.

Thanks for the quick response.

[Jacobo_Blanco]: https://avatars.slack-edge.com/2019-12-03/857109811792_877584fe7c70c84fc3d2_24.png
[Anna_Geller]: https://avatars.slack-edge.com/2021-10-04/2585679938480_3d7e8c7969cc36b0c7ca_24.jpg

anticorrelator:
closes #5623

This PR adds two small features:

- a convenience option to set the concurrency limit on deployment work queues as they are being created
- the ability to start an agent with a prefix matching string instead of an explicit list of work queues, the agent will dynamically update the set of work queues it polls from, obviating the need to restart agents to support new work queues and concurrency limits

Note: the regex filtering can be implemented either as a serverside feature or purely clientside. I'm currently opting for a clientside-only implementation to avoid (even an incremental) API version bump.

### Example

These two features work together to ensure that deployments running on all work queues that match a pattern will be picked up without restarting the agent. In the following example, three deployments are created against a work queue with a name that matches the `"dev-"` prefix pattern. By using the `--limit` option, the concurrency limit on the associated work queues can be conveniently set. Even though the work queues are created after the agent has started, the work is still picked up.
```
prefect agent start --match "dev-"
...
prefect deployment build -n dev -q dev-healthcheck --limit 1 -a flows/healthcheck.py:healthcheck
prefect deployment build -n dev -q dev-parametrized --limit 1 -a flows/parametrized.py:parametrized
prefect deployment build -n dev -q dev-hello --limit 1 -a flows/hello.py:hello
```

### Checklist


anna-geller:
Agent polls for all work queues, which works great ✅  
But concurrency limits have not been applied ❌
image

**Minor detail**
When starting the agent, the current message is: `Agent started! Looking for work from queue(s): ...`

Maybe we can rewrite this to: `Agent started! Looking for work from queue(s) that match pattern: dev-.*...`

src/prefect/cli/work_queue.py
```diff
@@ -206,7 +206,10 @@ async def inspect(
 async def ls(
     verbose: bool = typer.Option(
         False, "--verbose", "-v", help="Display more information."
-    )
+    ),
+    work_queue_regex: str = typer.Option(
+        None, "--regex", "-r", help="Python regex string used to match work queue names"
```
madkinsz, line 211: Maybe we can call this `--match` or `--pattern` instead? That it's a regex feels like a detail for the description.

madkinsz:
This looks really cool!

src/prefect/agent.py
```diff
@@ -36,7 +37,8 @@ def __init__(
                 "Provide only one of 'default_infrastructure' and 'default_infrastructure_document_id'."
             )
 
-        self.work_queues: Set[str] = set(work_queues)
+        self.work_queues: Set[str] = set(work_queues) if work_queues else set()
+        self.work_queue_regex = work_queue_regex
```
madkinsz, line 41: If you update the client to accept a compiled regex as well we can compile it here which will be helpful for performance since we're using it repeatedly.

src/prefect/agent.py
```diff
@@ -108,6 +130,8 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
 
         self.logger.debug("Checking for flow runs...")
 
+        await self.update_matched_agent_work_queues()
```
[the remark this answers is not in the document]
anticorrelator: on my branch I've bundled this with the cached work queue update, so it only happens every 30 seconds

src/prefect/agent.py
anticorrelator: yeah that's a good idea

src/prefect/cli/work_queue.py
anna-geller: plus, who knows maybe one day we would decide to match differently than by regex, so good call

src/prefect/agent.py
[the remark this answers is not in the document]
anna-geller: optimizing a query pattern sounds like a topic for @jlowin :)

anna-geller (approved):
Leaving naming (--regex vs. replacing it by --match) and performance (querying for work queues) concerns aside, this looks great, and QA passes 👍

anna-geller:
closes https://github.com/PrefectHQ/prefect/issues/5623 for now (we may consider a more dedicated solution to the singleton problem, but this PR addresses it already)

zangell44:
I'm wondering if it would be better for us to only allow matching using `LIKE`-style queries and perform the matching when we query for work queues vs. allowing for full regex and doing the matching client side.
The changes would be
- trigram index on `work_queue.name`
- adding a `LIKE` style filter to work queue filters

anticorrelator:
@zangell44 yeah! I think a serverside implementation would be a lot more performant; my thought right now is to leave this implementation completely on the serverside if possible, since that would minimize the maintenance burden/cost of making breaking changes. Do you have any thoughts about this?

anticorrelator:
@zangell44 @anna-geller after thinking about this for a bit I think it's worth it to wait to understand usage before committing to a new index/migration. Adding this feature later should be relatively pain-free on the scale of things: the match pattern is only used on agent start and we can document (and warn) users if our match syntax changes from regex to sql matching

anna-geller:
Comments after the change to startswith prefix-pattern:

image

so nice and simple <3 

image

concurrency limit of 1 is respected:
image

LGTM!

madkinsz (approved):

PR #7101, opened 2022-10-07
Exempt `status:accepted` labels from being marked as stale

bunchesofdonald:
Adds `status:accepted` to the Stale action's exempt label list.


madkinsz (approved):
Ah nice. We can remove `status:backlog`. I renamed that label to accepted.

PR #7102, opened 2022-10-07
Always use _delete_auto_scheduled_runs utility

zangell44:
Centralizes logic for deleting deployment auto-scheduled runs.
Part of the fix for https://github.com/PrefectHQ/nebula/issues/2113

### Checklist

### DIFF
--- a/src/prefect/orion/models/deployments.py
+++ b/src/prefect/orion/models/deployments.py
@@ -343,11 +343,7 @@ async def delete_deployment(
     """
 
     # delete scheduled runs, both auto- and user- created.
-    delete_query = sa.delete(db.FlowRun).where(
-        db.FlowRun.deployment_id == deployment_id,
-        db.FlowRun.state_type == schemas.states.StateType.SCHEDULED.value,
-    )
-    await session.execute(delete_query)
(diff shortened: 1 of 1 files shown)


src/prefect/orion/models/deployments.py
```diff
@@ -343,11 +343,7 @@ async def delete_deployment(
     """
 
     # delete scheduled runs, both auto- and user- created.
-    delete_query = sa.delete(db.FlowRun).where(
-        db.FlowRun.deployment_id == deployment_id,
-        db.FlowRun.state_type == schemas.states.StateType.SCHEDULED.value,
-    )
-    await session.execute(delete_query)
+    await _delete_auto_scheduled_runs(session=session, deployment_id=deployment_id)
```
chrisguidry, line 346: Hmm, the `_delete_auto_scheduled_runs` function filters down to only `auto_scheduled=True`, but this comment says both "auto- and user-created".  Should we change it to `_delete_scheduled_runs` with an additional option for whether we do all or just the auto-scheduled ones?

chrisguidry (commented):
Love where you're heading here, just think we might have a slightly different use case to cover here (user-created versus auto-scheduled)

PR #7104, opened 2022-10-07
Experiment using new python handler for mkdocstrings

tpdorsey:
Experimental use of new python handler for mkdocstrings (replacing python-legacy handler) for creating API reference docs.

Why change handler? 

- The legacy handler will be deprecated eventually, might as well be ahead of the timeline if the new handler suits our needs.
- New handler uses griffe. Seems to be much faster and more configurable, and I believe we have some in-house expertise using griffe elsewhere in our stack.
- New handler will support jinja templating (at least that's the plan).
- New handler does a much nicer job natively handling modules.

Issues/work to be explored:

- New handler uses completely different templates. The default templates work quite well, but we'll have to start over with customization on the new templates. (This PR currently is using the defaults with only minor CSS-level customization.)
- It's possible to generate API docs entirely on the fly, without having to create skeleton .md files or setting up nav in mkdocs.yml. This could be an advantage for adding new modules. I will continue to look at this option.
- Unknown unknowns

### Checklist


tpdorsey:
Closed in favor of https://github.com/PrefectHQ/prefect/pull/7855

I encountered unexpected build errors and decided to re-step through implementation to figure out what caused the errors.

PR #7107, opened 2022-10-07
Add private repo support for GitHub

peytonrunyan:
This PR adds support for private GitHub repositories using Personal Access Tokens. It also moves our directory copying function from bash to `copy_tree` which should hopefully provide windows support. I still need to finish QA, but wanted to get eyes on the implementation before going any further. 

Shoutout to @ahuang11 and @desertaxle for working me on the PR for https://github.com/PrefectHQ/prefect-github/pull/30 which was 99.9% of the work for this one.

Behavior worth noting: a bad token results in a prompt to log in to GitHub. A token with insufficient privileges will result in an OSError that has the message `remote: Write access to repository not granted.` in the body.
### Example

```python
token = "my-personal-access-token"
repo = "https://github.com/PrefectHQ/prefect.git"
g = GitHub(
    repository=repo,
    access_token=token,
)
```

### QA
Successfully ran the following:
```python
from prefect.filesystems import GitHub

repo = GitHub(
    repository=<my_private_repo>,
    access_token=<my_token>
)

repo.get_directory()
```
### Checklist


src/prefect/filesystems.py
```diff
@@ -791,47 +793,103 @@ class GitHub(ReadableDeploymentStorage):
         default=None,
         description="An optional reference to pin to; can be a branch name or tag.",
     )
+    access_token: Optional[SecretStr] = Field(
+        name="Personal Access Token",
+        default=None,
+        description="A github Personal Access Token (PAT).",
```
ahuang11, line 799:
```suggestion
        description="A GitHub Personal Access Token (PAT).",
```

src/prefect/filesystems.py
```diff
@@ -791,47 +793,103 @@ class GitHub(ReadableDeploymentStorage):
...
 
-        # in this case, we clone to a temporary directory and move the subdirectory over
-        tmp_dir = None
-        tmp_dir = tempfile.TemporaryDirectory(suffix="prefect")
-        path_to_move = str(Path(tmp_dir.name).joinpath(from_path))
-        cmd += f" {tmp_dir.name} && cp -R {path_to_move}/."
+        # Clone to a temporary directory and move the subdirectory over
+        with TemporaryDirectory(suffix="prefect") as tmp_dir:
+            tmp_path_str = tmp_dir
+            cmd += f" {tmp_path_str}"
```
ahuang11, line 882:
```suggestion
            cmd += f" {tmp_dir}"
```

src/prefect/filesystems.py
```diff
@@ -791,47 +793,103 @@ class GitHub(ReadableDeploymentStorage):
         default=None,
         description="An optional reference to pin to; can be a branch name or tag.",
     )
+    access_token: Optional[SecretStr] = Field(
+        name="Personal Access Token",
+        default=None,
+        description="A github Personal Access Token (PAT).",
+    )
+
+    @validator("access_token")
```
cicdw, line 802: Could be useful to throw a comment above this about why you made this an `access_token` validator instead of a `repository` validator (I think I get it - you only want to validate when a PAT is provided)

src/prefect/filesystems.py
```diff
@@ -791,47 +793,103 @@ class GitHub(ReadableDeploymentStorage):
...
+                        "using the 'HTTPS' format. You must either remove the "
+                        "credential if you wish to use the 'SSH' format and are not "
+                        "using a private repository, or you must change the repository "
+                        "url to the 'HTTPS' format. "
+                    )
+                )
+
+        return v
+
+    def _create_repo_url(self) -> str:
```
cicdw, line 819: Should this be a `repository` validator?

ahuang11 (commented):
Looks good! Just minor suggestions.

cicdw (commented):
so fast! a few minor comments / thoughts

WillRaphaelson:
100% wouldn't slow this down - but wondering, does this implementation preclude a UX like what we have with docker registry creds being composed into the docker container block? I actually think all of the web service blocks should be decoupled from the web service credential blocks, so I can have multiple s3 buckets that all use the same aws creds block, multiple github repository blocks that all use the same github PAT block. Again, dont block on this, just a note.

image

jlowin:
Agree with @WillRaphaelson that the “full” implementation should introduce GH credentials as a composable, standalone block

peytonrunyan:
I wish I could take credit for knocking this one out that quickly. Almost all of the work had been finished at https://github.com/PrefectHQ/prefect-github/pull/30. That one is a fuller implementation, including using the GitHub credential.

src/prefect/filesystems.py
```diff
@@ -791,47 +793,103 @@ class GitHub(ReadableDeploymentStorage):
         default=None,
         description="An optional reference to pin to; can be a branch name or tag.",
     )
+    access_token: Optional[SecretStr] = Field(
```
madkinsz, line 796: Should we try loading the `GITHUB_TOKEN` environment variable if not set?

src/prefect/filesystems.py
peytonrunyan: I figured this way keeps us from having an attribute that exposes the secret value.

src/prefect/filesystems.py
peytonrunyan: Is this a common pattern? I see a `GITHUB_TOKEN` for our release notes. My gut is against it because we aren't offering environment variable credentials elsewhere for storage.

cicdw: Because a block ultimately persists its state in the DB, we should avoid implicit assignment of properties.

peytonrunyan:
@anna-geller and I did some QA on Friday and it's working as expected.

src/prefect/filesystems.py
```diff
@@ -791,47 +793,105 @@ class GitHub(ReadableDeploymentStorage):
...
+        """
+        if v is not None:
+            if urllib.parse.urlparse(values["repository"]).scheme != "https":
+                raise InvalidRepositoryURLError(
+                    (
+                        "Crendentials can only be used with GitHub repositories "
+                        "using the 'HTTPS' format. You must either remove the "
+                        "credential if you wish to use the 'SSH' format and are not "
+                        "using a private repository, or you must change the repository "
+                        "url to the 'HTTPS' format. "
```
tpdorsey, line 817:
```suggestion
                        "URL to the 'HTTPS' format. "
```

tpdorsey (approved):
LGTM. Thanks!

ahuang11 (approved):

madkinsz:
@peytonrunyan please remember to include the `enhancement` label so this is correctly categorized in the changelog. This also should have the body checklist filled out.

PR #7108, opened 2022-10-07
Refactor `State` types into `prefect.states`

madkinsz:
Requires #7106 

I was uncomfortable with the implementation details present in `prefect.client.schemas` as well as the lack of clarity on where state types should be imported from. This moves implementation details of the `State` type and related utilities into `prefect.states`. `prefect.client.schemas` now contains schemas with no logic. A bit of shuffling was required to resolve circular imports here. I've moved `inject_client` to a dedicated module as we use it in nearly every client module and it is a constant source of circular import issues.


src/prefect/client/utilities.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+# This module must not import from `prefect.client` when it is imported to avoid
+# circular imports for decorators such as `inject_client` which are widely used.
+
+from functools import wraps
+
+from prefect.utilities.asyncutils import asyncnullcontext
+
+
+def inject_client(fn):
```
peytonrunyan, line 12: Big fan of the move!

Issue #7098, opened 2022-10-06
Allow tasks to wait_for any terminal state, not just Completed


EmilRex:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

A task's `wait_for` argument only allows it to run when all of its upstream tasks have succeeded, or more specifically reached a `Completed` state.

### Describe the proposed behavior

There should exist a way to declare that a task should wait for its upstream tasks to reach any [terminal state](https://docs.prefect.io/concepts/states/#state-types).

### Potential Solution
```python
@task
def task_one():
    pass

@task
def task_two():
    raise ValueError()

@task(wait_for=[task_one, allow_failure(task_two)]):
def task_three():
    pass
```

david-elliott-deliveroo:
Huge ➕ 1 on this please, we have a blocking use-case whereby we _need_ to be able to run tasks downstream of other failed tasks!!

PR #7120, opened 2022-10-10
Add `allow_failure` annotation to allow failed runs to be passed downstream

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/7098

### Example

```python
@task
def fails():
    raise ValueError("Fail task!")

@task
def bar(y):
    return y

@flow
def test_flow():
    f = fails.submit()

    # Previously, `bar` would fail with a `NotReady` state
    # Now, it will receive an exception
    b = bar.submit(allow_failure(f))
    assert isinstance(b.result(), ValueError)
    
    # Previously, `bar` would fail with a `NotReady` state
    # Now, it will run after the upstream completes
    b = bar.submit(1, wait_for=[allow_failure(f)])
    assert b.result() == 1
```

### Checklist

### DIFF
--- a/src/prefect/utilities/annotations.py
+++ b/src/prefect/utilities/annotations.py
@@ -4,14 +4,39 @@ T = TypeVar("T")
 
 
 class unmapped:
-    """A container that acts as an infinite array where all items are the same
-    value. Used for Task.map where there is a need to map over a single
-    value"""
+    """
+    Wrapper for iterables.
+
+    Indicates that this input should be sent as-is to all runs created during a mapping
+    operation instead of being split.
+    """
 
     def __init__(self, value: Any):
         self.value = value
 
     def __getitem__(self, _) -> Any:
+        # Internally, this acts as an infinite array where all items are the same value
+        return self.value
+
+
+class allow_failure:
+    """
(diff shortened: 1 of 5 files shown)


madkinsz:
I'd like to add some documentation before merging, but otherwise this is pretty straightforward.

EmilRex (approved):
🚀

madkinsz:
@tpdorsey do you know where documentation would best fit for this?

src/prefect/utilities/annotations.py
```diff
@@ -4,14 +4,39 @@
...
 
 class unmapped:
-    """A container that acts as an infinite array where all items are the same
-    value. Used for Task.map where there is a need to map over a single
-    value"""
+    """
+    Wrapper for iterables.
+
+    Indicates that this input should be sent as-is to all runs created during a mapping
+    operation instead of being split.
```
bunchesofdonald, line 11: Thank you for updating this.

bunchesofdonald (approved):

madkinsz:
There's no `wait_for` documentation right now. I can write it but I'd like to get this merged and do it as a follow-up.

PR #7128, opened 2022-10-11
Additional test coverage for results

madkinsz:
### Changes

- Abstract `PersistedResult._read_blob` for testing persisted results
- Consolidate result testing utilities into `prefect.testing.utilities`
- Add pytest assertion rewrites for testing utilities
- Add flow and test result coverage
- Add `MissingResult` exception instead of `ValueError` when data is null


bunchesofdonald (approved):
👏 That's a lot of tests and edge cases, nice work!

PR #7129, opened 2022-10-11
Remove old result handling from the engine and documentation

madkinsz:
Previously this was left untouched even though it no longer used (since #7094), now it has been purged.


madkinsz:
Thanks for the speedy review!!

PR #7138, opened 2022-10-12
History query refactor

zangell44:
This PR adds `state_timestamp` to the `flow_run` and `task_run` tables. This denormalization will speed up run history queries, which now do not need to join to the `flow_run_state` and `task_run_state` tables.

See https://github.com/PrefectHQ/nebula/issues/2110

### Example

n/a

### Checklist

### DIFF
--- a/src/prefect/orion/database/orm_models.py
+++ b/src/prefect/orion/database/orm_models.py
@@ -266,16 +266,13 @@ class ORMRun:
         """Total run time is incremented in the database whenever a RUNNING
         state is exited. To give up-to-date estimates, we estimate incremental
         run time for any runs currently in a RUNNING state."""
-        if self.state and self.state_type == schemas.states.StateType.RUNNING:
+        if self.state_type and self.state_type == schemas.states.StateType.RUNNING:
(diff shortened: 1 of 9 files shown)


src/prefect/orion/database/orm_models.py
```diff
@@ -265,34 +266,29 @@ def estimated_run_time(self):
...
-        if self.state and self.state_type == schemas.states.StateType.RUNNING:
-            return self.total_run_time + (pendulum.now() - self.state.timestamp)
+        if self.state_type and self.state_type == schemas.states.StateType.RUNNING:
+            return self.total_run_time + (pendulum.now() - self.state_timestamp)
         else:
             return self.total_run_time
 
     @estimated_run_time.expression
     def estimated_run_time(cls):
-        # use a correlated subquery to retrieve details from the state table
```
chrisguidry, line 275: This expression property was on my radar for that other performance issue involving loading the run state (flow_runs/filter).  This might improve the performance of that endpoint too!

chrisguidry (approved):
This makes a lot of sense to me.  We'll have to backfill old values, though, won't we?

PR #7140, opened 2022-10-12
Proactively look for duplicate blocks instead of relying on `IntegrityError`.

chrisguidry:
On PostgreSQL, relying on the `IntegrityError` to indicate a constraint violation does work, but leads to a lot of noise in the PostgreSQL logs.  Also, tracing libraries see these spans as errors, even though they are normal ops.

This is a backported fix from Prefect Cloud: PrefectHQ/nebula#2165

### Checklist

### DIFF
--- a/src/prefect/orion/models/block_documents.py
+++ b/src/prefect/orion/models/block_documents.py
@@ -7,6 +7,7 @@ from typing import Dict, List, Optional, Tuple
 from uuid import UUID, uuid4
 
 import sqlalchemy as sa
+from sqlalchemy.ext.asyncio import AsyncSession
 
 import prefect.orion.models as models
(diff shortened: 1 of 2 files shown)


src/prefect/orion/models/block_documents.py
```diff
@@ -211,7 +228,7 @@ async def _find_parent_block_document(
 
 @inject_db
 async def read_block_document_by_name(
-    session: sa.orm.Session,
+    session: AsyncSession,
```
chrisguidry, line 231: While I was in this module, I fixed these incorrect type hints.  We should get these corrected across the `models` package by and by.

zangell44 (approved):

madkinsz (approved):

PR #7141, opened 2022-10-12
Save log batches in individual transactions.

chrisguidry:
When saving very large numbers of logs, it's possible to overwhelm a PostgreSQL instance by having a single extremely large transaction open.  Instead, we should commit as we go.

Part of PrefectHQ/nebula#2192

### Checklist

### DIFF
--- a/src/prefect/orion/models/logs.py
+++ b/src/prefect/orion/models/logs.py
@@ -4,8 +4,8 @@ Intended for internal use by the Orion API.
 """
 from typing import List
 
-import sqlalchemy as sa
 from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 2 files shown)


src/prefect/orion/models/logs.py
```diff
@@ -37,13 +42,12 @@ async def create_logs(
         None
     """
     log_insert = await db.insert(db.Log)
-    for batch in batched_iterable(logs, LOG_BATCH_SIZE):
-        await session.execute(log_insert.values([log.dict() for log in batch]))
```
chrisguidry, line 41: I wanted to preserve the convention that all `models` methods take a `session` argument, so I've moved the batching out in the API.

zangell44 (approved):

PR #7152, opened 2022-10-13
Manual flow retries

### LINKED ISSUE #7127: Restart Flow Runs from Failed Tasks
**User Story**
As a Prefect 2 user with a flow run that has completed in a failed state, I want to retry the flow run, from either the UI or the CLI, perhaps because I know that something about the flow run context has changed. I expect that any task run in that flow run that finished in a failed state will be restarted with all of its retries reset to zero.

**Implementation Considerations**
- This change may require updates to orchestration rules
- We should consider the tradeoffs of using the existing `/api/task_runs/{id}/set_state` route or creating a new route, such as `/api/flow_runs/{id}/restart`.
- Caching must be enabled to achieve this. The design must account for scenarios in which there is no cache. In particular, the UI would ideally know whether it is possible to retry before inviting the user to do so.

**Prompts**
- Is "Restart" the best name for this feature? To me, "Restart" implies that the flow run will start at the beginning. "Retry" might be more appropriate. Or "Restart from Failure" might be more precise.

**Resources**
- [Prefect 1 implementation](https://docs-v1.prefect.io/orchestration/ui/flow-run.html#restarting)

anticorrelator:
closes #7127

Implements the ability to manually retry flow runs in terminal states by proposing an `AwaitingRetry` scheduled state

This PR also includes a refactor of flow retry logic

- we no longer check for failed task runs and marking them rerunnable on retry/restart
- instead, we track whether or not a flow is retrying
- while attempting to retry a flow, the engine attempts to orchestrate any tasks it discovers

This should significantly reduce the overhead of retrying flows with many task runs

### Checklist


billpalombi:
This design is consistent with what we discussed yesterday. It's somewhat more complex than the "soft" restart, but I appreciate the protection against potential failure modes.

src/prefect/orion/schemas/core.py
```diff
@@ -268,6 +271,10 @@ class TaskRunPolicy(PrefectBaseModel):
     retry_delay: Optional[int] = Field(
         default=None, description="The delay time between retries, in seconds."
     )
+    flow_restart_index: Optional[int] = Field(
```
cicdw, line 274: why wouldn't this always be equal to `restarts`?

src/prefect/orion/api/flow_runs.py
```diff
@@ -260,3 +260,43 @@ async def set_flow_run_state(
         response.status_code = status.HTTP_201_CREATED
 
     return orchestration_result
+
+
+@router.post("/{id}/restart")
```
cicdw, line 265: If we think making this an orchestration rule is correct, I don't see the benefit of this route? Seems like setting the flow run state to `AwaitingRestart` is all that's need?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -450,3 +465,137 @@ async def before_transition(
...
+        self,
+        initial_state: Optional[states.State],
+        proposed_state: Optional[states.State],
+        context: OrchestrationContext,
+    ):
+        from prefect.orion.models import task_runs
+
+        trs = await flow_runs.read_task_runs(context.session, context.run.id)
+        for task_run in trs:
+            if task_run.empirical_policy.flow_restart_index is None:
```
cicdw, line 549: Could you explain this logic?

cicdw (commented):
A few questions for ya

src/prefect/orion/api/flow_runs.py
anticorrelator: the new route is mostly a convenient way to specify an entirely different orchestration policy: for instance we'd be moving the state of a run "backwards" which is normally prohibited by the standard policy

src/prefect/orion/orchestration/core_policy.py
anticorrelator: The hard restart strategy keeps all the old task runs and states (also important to caching), but it's important to indicate that the the task runs came from a different "attempt" at traversing the flow logic. If a user presses the restart button twice on a flow with 5 tasks, the flow run will end up with 15 tasks runs which we can distinguish in the UI with this metadata

image

src/prefect/orion/schemas/core.py
anticorrelator: if a flow run has been restarted multiple times, some of its task runs will be from the first attempt at running the flow and some other task runs will be from a later attempt

src/prefect/orion/orchestration/core_policy.py
cicdw: Hm that doesn't sound right to me - we shouldn't create any new task runs with a flow run restart; I also don't see where any new task runs are created in the PR?  But tracking what attempt a task run was successful on makes sense to me

anticorrelator: if we don't make new task runs the restart will always fail to work properly if there was a change in execution environment that would change the "realized task graph" of the flow run, given that this graph can be different from restart to restart (e.g. if I misconfigured a block or I forget to turn on a server--anything outside of Prefect's control) it's very difficult to the tasks to each other properly

anticorrelator: the idea here is: if users have configured their cache keys and result storage properly, the restarted task runs never enter a running state, they are retrieved from the cached state created from the original attempt instead.

We do work to try and match up "old runs" with "new runs" in our flow retry logic. It relies on idempotency key and will break even if the same graph is realized in a different order—something as simple as a task taking longer to run when the target bucket is filled with data can mess up the task run matching; I do think if we really want to try and match up task runs there are options there to do so based on the cache key instead

anticorrelator: sorry--to answer your question about where the task runs come from: the flow run is put in a scheduled state, so an agent will pick up the flow again and create new task runs as it runs the flow

cicdw: Two points:
- putting the flow run into a scheduled state won't create any new task runs (and that's a good thing)
- it's OK for us to tell users they need to write stable graphs to maximally benefit from our feature set; creating new runs is going to be very noisy and not representative of what's happening (in reality the runs are the same, they just have a longer state history)

madkinsz:
Perhaps we can separate the changes required for a hard-restart from the changes required to expose the existing retry mechanism? If they're in two pull requests we can first review exposure of the retry/restart mechanism via an API route _then_ , separately and without blocking the feature we need to deliver, we can have a more extended discussion about adding an option to that route to perform a _hard_ restart.

src/prefect/orion/api/flow_runs.py
jlowin: This "feels" very wrong to me -- either restarting flows is an expected orchestration path, or it's an exception, but needing to load an entire new policy to handle it goes against the design of the rules engine.

For example, the `PreventRestartingFlowsWithoutDeployments` rule is defined on every single transition, meaning it can not be used except with the `FlowRestartPolicy` that is loaded only on specific transitions called from a specific route -- the purpose of the rules engine is to provide a generic policy that embraces all expected transitions. In this model, `PreventRestartingFlowsWithoutDeployments` should be defined on `[<acceptable prior states>] → [AwaitingRetry]` only and included in the global flow run policy. It would only be enforced when attempting to retry.

If the rule about moving the run state backwards is too strict to account for this new behavior, then that rule can be relaxed rather than introducing an entirely new orchestration state machine to work around it.

src/prefect/orion/orchestration/core_policy.py
jlowin: We can't loop over every task run and modifying it when setting a flow run state - this is the exact pattern that frequently kills Prefect 1 deployments with large number of task runs (imagine mapping over thousands of tasks and trying to retry!)

I suspect that the information about what "flow run iteration" a task run belongs to can be inferred by their relative timestamps if necessary, but if we want to encode it at the task run level then the time to do so is when the task run state is recorded. E.g. when marking a task run as complete, also record the flow run's retry index at that time. This will achieve the same outcome with no additional overhead.

madkinsz: We are already doing this for automated flow run retries. Perhaps we need to take a step back and take a look at our mechanisms for performing a rerun holistically?

max13fr:
Hello,

Thanks for your contribution for this really useful feature !

I'm not sure to fully understand how is working your pull request or the internals of prefect2 so sorry if i'm off topic.
I would like to know if your feature could be extended to allow to launch a new flow with some tasks already validated with a specified result. It would be incredible to test the deployments by mocking some steps without touching to the code.

Best regards,
Max

src/prefect/orion/api/flow_runs.py
```diff
@@ -260,3 +260,43 @@ async def set_flow_run_state(
...
+async def restart_flow_run(
+    flow_run_id: UUID = Path(..., description="The flow run id", alias="id"),
+    force: bool = Body(
+        False,
+        description=(
+            "If false, orchestration rules will be applied that may alter "
+            "or prevent the state transition. If True, orchestration rules are not applied."
+        ),
+    ),
```
madkinsz, line 274: What does this mean in the context of a restart?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -305,6 +292,55 @@ async def before_transition(
...
+    TO_STATES = [states.StateType.SCHEDULED]
+
+    async def before_transition(
+        self,
+        initial_state: Optional[states.State],
+        proposed_state: Optional[states.State],
+        context: FlowOrchestrationContext,
+    ):
+        if context.run.parent_task_run_id:
+            await self.abort_transition("Cannot restart a subflow run.")
```
madkinsz, line 306: You should be allowed to do this e.g. when `run_deployment` has been used you could want to restart a run that has been linked. I think your next check that a deployment id is present should be sufficient here.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -305,6 +292,55 @@ async def before_transition(
...
+        # reset run count to 0
+        self.original_run_count = context.run.run_count
+        context.run.run_count = 0  # reset run count to preserve retry behavior
+
+        # update empirical settings
+        self.original_settings = context.run_settings.copy()
+        self.restarts = self.original_settings.restarts + 1
+        new_settings = context.run_settings.copy()
+        new_settings.restarts = self.restarts
+        flow_run_update = actions.FlowRunUpdate(empirical_policy=new_settings)
```
madkinsz, line 322: Before merge, we should definitely be tracking this in the state details or on the flow run since empirical settings is intended to be static.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -406,13 +442,20 @@ class PreventTransitionsFromTerminalStates(BaseOrchestrationRule):
...
+
     async def before_transition(
         self,
         initial_state: Optional[states.State],
         proposed_state: Optional[states.State],
         context: OrchestrationContext,
     ) -> None:
-        await self.abort_transition(reason="This run has already terminated.")
+        if proposed_state.name not in self.NAMED_STATE_EXCEPTIONS:
+            await self.abort_transition(reason="This run has already terminated.")
```
madkinsz, line 458: This makes me a bit uncomfortable. I guess I don't see a clear path around it at this time though.

src/prefect/orion/api/flow_runs.py
```diff
@@ -260,3 +260,43 @@ async def set_flow_run_state(
...
+
+    async with db.session_context(begin_transaction=True) as session:
+        # create the state
+        orchestration_result = await models.flow_runs.set_flow_run_state(
+            session=session,
+            flow_run_id=flow_run_id,
+            state=state,
+            force=force,
+            flow_policy=flow_policy,
+        )
```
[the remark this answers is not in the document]
anticorrelator: yep, I've mentioned this elsewhere--this is no longer necessary

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -450,3 +493,68 @@ async def before_transition(
...
+                await models.task_runs.update_task_run(
+                    context.session, context.run.id, task_run_update
+                )
+                await self.rename_state("Restarting")
+        elif self.flow_run.run_count > context.run.empirical_policy.flow_retry_attempt:
+            # if the flow run count is > 1, the flow is retrying
+            updated_settings = context.run_settings.copy()
+            updated_settings.flow_retry_attempt += 1
+
+            task_run_update = actions.TaskRunUpdate(empirical_policy=updated_settings)
```
madkinsz, line 541: Same as the above comment re usage of the empirical policy for state.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -406,13 +442,20 @@ class PreventTransitionsFromTerminalStates(BaseOrchestrationRule):
     FROM_STATES = TERMINAL_STATES
     TO_STATES = ALL_ORCHESTRATION_STATES
 
+    NAMED_STATE_EXCEPTIONS = {
+        "Restarting",
```
madkinsz, line 446: Is this the name that's used for RUNNING on a restart? It mirrors "Retrying" but is unfortunately less clear and feels easily confused with the scheduled state. Perhaps @billpalombi has a suggestion here.

I think that AwaitingRetry -> Retrying makes sense for retries but for restarts the action isn't automated so it feels more like Restarting -> Running?

billpalombi: I think:
- the named SCHEDULED equivalent should be "AwaitingRestart" to mirror "AwaitingRetry"
- the named RUNNING equivalent should be "Rerunning." I don't love it either, but I can't think of anything better.

anticorrelator: I don't have an opinion about the named state here--I think `Retrying` makes just as much sense as anything else

billpalombi: Do IU understand correctly that retry counts are zeroed out when a flow run is restarted? If so, my only concern with `Retrying` is that a user might expect that it counts toward their retry count, but perhaps we don't need to signal that with the state name.

madkinsz: Yeah the retry counts are zeroed so it can do its full set of retries again. 

> the named SCHEDULED equivalent should be "AwaitingRestart" to mirror "AwaitingRetry"

The problem with this is that it makes it seem like you're waiting for someone to click the restart button but this is actually what happens right after you click it. I'm not super opposed though.

> the named RUNNING equivalent should be "Rerunning." I don't love it either, but I can't think of anything better.

Can we just use "Running" as a name and consider a custom name later? A custom name was necessary for Dustin's original implementation but it should not be now. I do not think we necessarily need to try to clarify this as we might as well see if using the normal name is sufficient.

billpalombi: If we don't need custom names, I agree that they may only create confusion. For example, if my flow run retries after having been restarted, we'll enter the usual state names and the indication that the flow run had been restarted will be lost. I'm on board with using "Scheduled" and "Running" as usual.

anticorrelator:
> Hello,
> 
> Thanks for your contribution for this really useful feature !
> 
> I'm not sure to fully understand how is working your pull request or the internals of prefect2 so sorry if i'm off topic. I would like to know if your feature could be extended to allow to launch a new flow with some tasks already validated with a specified result. It would be incredible to test the deployments by mocking some steps without touching to the code.
> 
> Best regards, Max

Hi @max13fr! Sorry for not getting back to you yesterday. This feature is intended to make it easy to *rerun* flows that have failed for some reason--while still being able to reuse any tasks in that flow that have completed successfully. If you want that behavior across flows, you might be able to do this already using custom cache functions on your tasks!

max13fr:
> > Hello,
> > Thanks for your contribution for this really useful feature !
> > I'm not sure to fully understand how is working your pull request or the internals of prefect2 so sorry if i'm off topic. I would like to know if your feature could be extended to allow to launch a new flow with some tasks already validated with a specified result. It would be incredible to test the deployments by mocking some steps without touching to the code.
> > Best regards, Max
> 
> Hi @max13fr! Sorry for not getting back to you yesterday. This feature is intended to make it easy to _rerun_ flows that have failed for some reason--while still being able to reuse any tasks in that flow that have completed successfully. If you want that behavior across flows, you might be able to do this already using custom cache functions on your tasks!

Hello,

No worries, there is no urgency, thanks for your reply.

The downside with the cache method is the need to modify the deployment for edge cases like :
- testing (mock one or more steps for example, force specific error on a step, ...)
- tempary hack (to bypass a step on a specific manual flow run)
- upgrading a deployment (eg: if we have an issue in the code of a task, we need to fix it then launch a new flow run without replaying the succeded steps).

For last the one, I guess it is the behaviour of your feature ? Or you force the reuse of the same version of a deployment ?

Thanks in advance,

Best regards, Max

madkinsz:
@max13fr Would you mind opening an enhancement issue explaining your desired use-case? That way, our product team can get involved and we can have a discussion dedicated to your needs there. Here in this pull request, we'll want to focus on getting the implementation details right for this feature that we've already decided on the functionality of.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -270,41 +274,96 @@ async def before_transition(
...
         retry_state = states.AwaitingRetry(
             scheduled_time=scheduled_start_time,
             message=proposed_state.message,
             data=proposed_state.data,
         )
+
+        # support old-style flow run retries for older clients
+        # older flow retries require us to loop over failed tasks to update their state
+        # this is not required after API version 0.8.3
+        api_version = context.parameters.get("api-version", None)
```
cicdw, line 297: this implies to me that the new logic is incompatible with old clients but abstractly I don't see why that should be the case - could you help me understand that?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -450,3 +524,69 @@ async def before_transition(
...
+            self.flow_run.run_count == 1
+            and self.flow_run.restarts > 0
+            and (self.flow_run.restarts > context.run.flow_restart_attempt)
+        )
+
+        retrying = not restarting and (
+            self.flow_run.run_count <= (self.flow_run.empirical_policy.retries + 1)
+        )
+
+        if restarting:
```
cicdw, line 569: The distinction between restarting and retrying feels arbitrary to me and also hard to internalize - is there a technical reason we need it?

cicdw (commented):
A few questions

src/prefect/orion/orchestration/core_policy.py
anticorrelator: It's really minor actually—we require a small client change to handle subflows (hence the new integration test, they all would have passed before its addition). Because we want the subflow to be rerun (as per the existing tests that define flow retry behavior) we need to handle it here: https://github.com/PrefectHQ/prefect/blob/flow-restarts/src/prefect/engine.py#L423-L425

src/prefect/orion/orchestration/core_policy.py
anticorrelator: Even if we remove any bookkeeping around retrying and restarting there still is a distinction: flow run run counts are reset on a restart, otherwise they will not retry as defined, they would be the same if restarts consumed a retry, but then we'd need special handling for restarts that exceed the retry limit

src/prefect/orion/orchestration/core_policy.py
anticorrelator: I should add that if we don't enforce subflow retry behavior to be preserved exactly we can remove this compatibility code, flows with just tasks will work just as well on old clients

cicdw: I think it's OK to keep around for a bit, thanks for the clarification!

src/prefect/orion/orchestration/core_policy.py
cicdw: I see, so because clients can set _automatic_ retry configuration on a flow, we need a way to allow "restart from failure" that ignores that automatic retry setting

anticorrelator: yeah exactly

jlowin: This is still really confusing to me (and I'm not sure, even after reading the code, which is a "retry" and which is a "restart"). 

To me a "restart" is a state transition from TERMINAL to SCHEDULED that can be orchestrated at any time. In addition, the system will automatically restart flows that have been configured to do so.

Consider a flow that has no retries configured (e.g. `retries=0`). That means the system won't automatically retry the flow on failure but of course I can _manually_ put that flow back into a scheduled state to retry it if it fails. In addition, a flow with `retries=3` would automatically retry up until the third attempt, but I could still continue to manually retry it above that.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -412,6 +474,18 @@ async def before_transition(
         proposed_state: Optional[states.State],
         context: OrchestrationContext,
     ) -> None:
+        if proposed_state.is_running() and context.parameters.get(
```
jlowin, line 477: This setup couples `RestartFlowRun` and `PreventTransitionsFromTerminalStates` in a really tight manner - they must both be in the orchestration policy and in the right order. It seems cleaner to have just one rule `HandleTerminalTransition` that allows restarts if appropriate and blocks transitions otherwise.

src/prefect/orion/schemas/core.py
```diff
@@ -207,16 +206,16 @@ class FlowRun(ORMBaseModel):
     state_name: Optional[str] = Field(
         default=None, description="The name of the current flow run state."
     )
-
     run_count: int = Field(
         default=0, description="The number of times the flow run was executed."
     )
-
+    restarts: int = Field(
```
jlowin, line 212: Just to echo my comment above, I can't understand why `run_count` doesn't already contain this information. How can run count increment if the flow wasn't restarted?

jlowin (changes requested):
This still seems way overcomplicated to me, in particular the need to separately track "retries", "restarts" and "run counts" suggests that the core motivation of running the flow again is being dominated by some degree of edge cases.

src/prefect/orion/schemas/core.py
anticorrelator: Just for posterity: the run count increments already with flow retries, restarts were conceived as a separate concept--we decided to make them the same so `restarts` -> `manual_retries` which consume a retry attempt

src/prefect/orion/orchestration/core_policy.py
anticorrelator: sounds good! I've split them up into two rules: one for Tasks and one for Flows

src/prefect/orion/orchestration/core_policy.py
anticorrelator: In this model, retries are transitions from running -> scheduled, the terminal transitions were never hit for automatic retries. Restarts were intended to encompass the terminal -> scheduled case; with different behavior. As mentioned in another comment we've decided to unify running -> scheduled and terminal -> scheduled for simplicity

jlowin (approved):
🎉 I really like how clean this feels now, nice work @anticorrelator!

PR #7154, opened 2022-10-13
Add create notification policy to OrionClient

jmg-duarte:
This change introduces `create_flow_run_notification_policy`, which enables programmatic creation of notification policies.

In the current project I am working on, we would like to be able to programatically create notification policies, however, the `OrionClient` API is lacking a method to do so, hence this change.

### Example
```py
    client.create_flow_run_notification_policy(
        "5a146297-2372-4afa-a4c4-d26dd1f638b3",
        is_active=True,
        state_names=["Completed", "Failed"]
    )
```

### Checklist

### Notes
- I am currently missing tests for this change and extra guidance to implement them would be amazing!


src/prefect/client/orion.py
```diff
@@ -1682,6 +1687,47 @@ async def create_logs(self, logs: Iterable[Union[LogCreate, dict]]) -> None:
...
+
+        policy = FlowRunNotificationPolicyCreate(
+            block_document_id=block_document_id,
+            is_active=is_active,
+            tags=tags,
+            state_names=state_names,
+            message_template=message_template,
+        )
+        response = await self._client.post(
+            "/flow_run_notification_policies/",
```
madkinsz, line 1721: Might not work with the trailing `/` — tests will show though
```suggestion
            "/flow_run_notification_policies",
```

madkinsz:
@jmg-duarte Take a look at the tests in `tests/client/test_orion_client.py` — I think you'll want to have a write/read test that creates the policy then reads it and asserts everything looks good. You'll need to implement a method for reading too. Do you have any specific questions?

src/prefect/client/orion.py
madkinsz: Looks like I mislead you here and it's actually the other way around :D

madkinsz:
> Not quite related to this PR but related nonetheless, how can I add a new filter to FlowRunNotificationPolicyFilter?
I think that filtering by block_document_id would be useful (at least in my use-case) and (actually) for this test too.

You'd need to add a `FlowRunNotificationPolicyBlockDocumentFilter` 🤯 that would look a bit like `BlockDocumentFilter` with a subset of support. We'd also probably need to investigate an index on that column in the database so filtering would be performant. I'd definitely prefer a separate pull request for that change as we'd need input from @chrisguidry or others who are focusing on the API.

madkinsz (approved):

PR #7161, opened 2022-10-13
Add `flow(persist_results_by_default=...)` to quickly set persistence for all children

madkinsz:
Adds `persist_results_by_default` to flows allowing you to toggle persistence for the flow _and_ its children rather than just for the flow.

The [documentation](https://deploy-preview-7161--prefect-orion.netlify.app/concepts/results/#toggling-persistence) should give a comprehensive overview.

### Example

```python

# Flow result is persisted
@flow(persist_results=True)
def foo():
   one()
   two()
   three()
   return 0

# Task result is persisted
@task 
def one():
   return 1

# Task result is not persisted
@task(persist_result=False)
def two():
   return 2

# Subflow result is persisted
@flow
def three():
   four()
   return 3

# Subflow task result is not persisted
@task
def four():
   return 4
```

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -290,9 +290,13 @@ Persistence of results requires a [**serializer**](#result-serializers) and a [*
 - `result_storage`: Where to store the result when persisted.
 - `result_serializer`: How to convert the result to a storable form.
 
+The flow decorator also includes the option:
+
+- `persist_results`: Whether the result of this flow and any tasks or subflows in it should be persisted to storage.
+
 #### Toggling persistence
 
-Persistence of the result of a task or flow can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task.
+Persistence of the results in a flow can be configured with the `persist_results` option. The `persist_results` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used in the flow.
 
 For example, the following flow has retries enabled. Flow retries require that all task results are persisted, so the task's result will be persisted:
 
@@ -337,7 +341,35 @@ def my_flow():
     my_other_task()
 ```
 
-Persistence of results can be manually toggled on or off:
+You may enable or disable persistence of all of the results in a flow manually by setting `persist_results`:
+
+```python
+from prefect import flow, task
+
+@flow(persist_results=True)
+def my_flow():
+    # This flow will persist its result
+
+    # This task will persist its result
+    my_task()
+
+    # This subflow will persist its result
+    my_subflow()
+
+
+@task
+def my_task():
+    ...
+
+@flow
+def my_subflow():
+    # Any tasks called in this subflow will not persist their results by default
+    ...
+```
+
+If you disable persistence with `persist_results`, results will not be persisted even if required for a feature you are using.
+
+Persistence of results can be manually toggled on or off per flow or task with `persist_result`:
 
 ```python
 from prefect import flow, task
@@ -355,6 +387,20 @@ def my_task():
 ```
 
 Toggling persistence manually will always override any behavior that Prefect would infer.
+
+You may toggle persistence of all of the results in a flow, but not persist the return value of the flow itself:
+
+```python
+from prefect import flow, task
+
+@flow(persist_result=False, persist_results=True)
+def my_flow():
+    # This task will persist its result
+    my_task()
+
+    # The return value of this flow will not be persisted
+    return 1
+```
 
 #### Result storage location
 
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -101,6 +101,12 @@ class Flow(Generic[P, R]):
             should be persisted to result storage. Defaults to `None`, which indicates
             that Prefect should choose whether the result should be persisted depending on
             the features being used.
+        persist_results: An optional toggle indicating whther the results of tasks and
+            flows within this flow should be persisted to result storage. Defaults to
+            `None` which follows the bhavior of `persist_result`. If `True``, this flow
+            and all tasks or flows within it will persist their results unless they have
+            opted out. If `False,` this flow and all tasks or flows within it will not
+            persist their results unless they have opted in.
         result_storage: An optional block to use to perist the result of this flow.
             This value will be used as the default for any tasks in this flow.
             If not provided, the local file system will be used unless called as
@@ -126,6 +132,7 @@ class Flow(Generic[P, R]):
         timeout_seconds: Union[int, float] = None,
         validate_parameters: bool = True,
         persist_result: Optional[bool] = None,
+        persist_results: Optional[bool] = None,
         result_storage: Optional[ResultStorage] = None,
         result_serializer: Optional[ResultSerializer] = None,
     ):
@@ -183,6 +190,7 @@ class Flow(Generic[P, R]):
                 ) from exc
 
         self.persist_result = persist_result
+        self.persist_results = persist_results
         self.result_storage = result_storage
         self.result_serializer = result_serializer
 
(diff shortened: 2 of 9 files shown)


src/prefect/flows.py
```diff
@@ -101,6 +101,12 @@ class Flow(Generic[P, R]):
             should be persisted to result storage. Defaults to `None`, which indicates
             that Prefect should choose whether the result should be persisted depending on
             the features being used.
+        persist_results: An optional toggle indicating whther the results of tasks and
```
anna-geller, line 104:
```suggestion
        persist_results: An optional toggle indicating whether the results of tasks and
```

src/prefect/flows.py
```diff
@@ -101,6 +101,12 @@ class Flow(Generic[P, R]):
...
+            flows within this flow should be persisted to result storage. Defaults to
+            `None` which follows the bhavior of `persist_result`. If `True``, this flow
```
anna-geller, line 106:
```suggestion
            `None`, which follows the behavior of `persist_result`. If `True`, this flow
```

src/prefect/flows.py
```diff
@@ -101,6 +101,12 @@ class Flow(Generic[P, R]):
...
+            and all tasks or flows within it will persist their results unless they have
+            opted out. If `False,` this flow and all tasks or flows within it will not
```
anna-geller, line 108:
```suggestion
            opted out. If `False`, this flow and all tasks or flows within it, will not
```

src/prefect/flows.py
```diff
@@ -543,6 +559,11 @@ def flow(
             should be persisted to result storage. Defaults to `None`, which indicates
             that Prefect should choose whether the result should be persisted depending on
             the features being used.
+        persist_results: An optional toggle indicating whther the results of tasks and
```
anna-geller, line 562:
```suggestion
        persist_results: An optional toggle indicating whether the results of tasks and
```

src/prefect/flows.py
```diff
@@ -543,6 +559,11 @@ def flow(
...
+            flows within this flow should be persisted to result storage. Defaults to
+            `None` which follows the bhavior of `persist_result`. If set, this flow and
```
anna-geller, line 564:
```suggestion
            `None`, which follows the behavior of `persist_result`. If set, this flow and
```

anna-geller (approved):
Looks great. I cross-checked/QAed that this indeed fixes the issue of with_options:

```python
from prefect import flow


@flow(result_serializer="json")
def flow_dont_persist_result():
    result = "Hi from Results! 👋"
    print(result)
    return result


if __name__ == "__main__":
    flow_dont_persist_result.with_options(name="customname", persist_result=True)()
    # gives pickle rather than json

    # workaround until 7161  gets released
    flow_dont_persist_result.with_options(
        name="customname",
        persist_result=True,
        result_serializer="json",
        result_storage=None,
    )()
    # PR fixes it https://github.com/PrefectHQ/prefect/pull/7161
```

and this correctly generates 4 result files:
```python


# Flow result is persisted
@flow(persist_results=True)
def persist_results_flow():
    one()  # 1st file
    two()  # 2nd file
    three()  # 3rd and 4th file
    return 0


# Task result is persisted
@task
def one():
    return 1


# Task result is not persisted
@task(persist_result=False)
def two():
    return 2


# Subflow result is persisted
@flow
def three():
    four()
    return 3


# Subflow task result is not persisted
@task
def four():
    return 4


if __name__ == "__main__":
    persist_results_flow()
```

src/prefect/flows.py
```diff
@@ -543,6 +559,11 @@ def flow(
...
+            all tasks or flows within it will persist their results unless they have
+            opted out. If
```
tpdorsey, line 566:
```suggestion
            opted out.
```

tpdorsey (approved):
Looks good. Minor edit suggested.

docs/concepts/results.md
```diff
@@ -290,9 +290,13 @@ Persistence of results requires a [**serializer**](#result-serializers) and a [*
 - `result_storage`: Where to store the result when persisted.
 - `result_serializer`: How to convert the result to a storable form.
 
+The flow decorator also includes the option:
+
+- `persist_results`: Whether the result of this flow and any tasks or subflows in it should be persisted to storage.
```
chrisguidry, line 295: 😬 Are we sure it's a good idea to have two options that differ only by `s`?  I feel like it will be super easy to mix that up, and hard to verbally talk about with peers.  Could we consider renaming this to something like `persist_all_results`?

src/prefect/flows.py
```diff
@@ -282,13 +291,18 @@ def with_options(
             persist_result=(
                 persist_result if persist_result is not NotSet else self.persist_result
             ),
+            persist_results=(
+                persist_results
+                if persist_results is not NotSet
+                else self.persist_results
+            ),
             result_storage=(
-                result_storage if result_storage is not NotSet else self.persist_result
```
chrisguidry, line 286: Good catches on these copypastas!

chrisguidry (approved):
The implementation and tests look great, and thanks for the bug fixes.  I just can't help but think it's a mistake to have two options so close in spelling with different meanings.  Can we run that by some product folks?

docs/concepts/results.md
cicdw: Yea, I agree with @chrisguidry that this will inevitably confuse folks - is there a production use case for disabling all result persistence, or is it just for testing?

madkinsz: I was also concerned and a few votes that is confusing is enough for sure. I do not think this should proceed with this variable name. I kind of like `persist_all_results`.

For other's watching, we are chatting about the use-case internally at https://prefecthq.slack.com/archives/C01QFU3549W/p1665775026247199

anna-geller: @cicdw to give context: some users want to globally disable result persistence for a flow because e.g. they operate on large dataframes with multiple tasks and they may not even need retries/caching, so this feature is very useful and something we've been asked about in the community

madkinsz: Persistence is off by default in 2.6+ — unless you have enabled retries or caching, there is no persistence unless you explicitly opt int.

I see the use-case here as more focused on _enabling_ persistence for all tasks without turning on a feature.

chrisguidry: I wish there was a way to say `persist_results_by_default` with fewer words.

Boolean options are clean and simple to explain, but there's another possibility here: 
 deprecate `persist_result=bool` and use an enum to get very explicit:

```
persist_results=Persist.Mine
persist_results=Persist.MineAndMySubTasks
persist_results=Persist.OnlyMySubtasks
```

These names are bad, but you see where I'm going...

`persist_result=bool` would translate internally to `persist_results=Persist.Mine` (I think that's the current behavior)

Naturally, this is a lot more complicated, but it gives us a way to avoid ambiguity

madkinsz: Whilst I appreciate the clarity of enums, I don't think they're particularly nice for user discoverability and I think they're useful when there are ~3 cases at most. Here we'd have to cover disabling options and clarify which options affect subflows as well.

> I wish there was a way to say persist_results_by_default with fewer words.

I actually don't think that's a bad suggestion. It's clearer than anything else we've discussed here :)

billpalombi: I think it makes sense to err on the side of explicitness with `persist_results_by_default`

anna-geller: +1, LGTM :)

madkinsz: After further thought, this may unfortunately not the clearest option. If `persist_results_by_default` is `False`, we will not persist results when a feature is used that would normally toggle persistence on. In this case, it is not changing the "default" but rather forcing all results to be disabled. I can't think of many use-cases where setting it to `False` makes sense, maybe we'd just want to throw an error if you were to do so when using features that typically require persistence. I'm not sure yet.

madkinsz:
https://github.com/PrefectHQ/prefect/pull/7228 is the preferred method for this for now. This can be revisited later.

Issue #7155, opened 2022-10-13
RRule Producing Incorrect Schedule


Sahiler:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

When inputting an RRule `RDATE:20221012T134000Z,20221012T230000Z,20221013T120000Z,20221014T120000Z,20221015T120000Z` RDATE into the deployment.yaml, it results in the UI showcasing 'every year' on the deployments page. The RRule dates are not consistent with the schedule and does not show any future flow runs. 

### Reproduction

- Install Prefect
- Create a generic hello-world flow
- Create a deployment for the flow
- Use RRule of `RDATE:20221012T134000Z,20221012T230000Z,20221013T120000Z,20221014T120000Z,20221015T120000Z`
- Check the UI and see that '_every year_' populates in the deployment schedule
- No future flow runs exist

### Error

- The deployment.yaml breakdown
<img width="754" alt="Screen Shot 2022-10-13 at 9 15 32 AM" src="https://user-images.githubusercontent.com/24235898/195621522-e4e6c149-d66d-46f8-b34a-a69e050b06ae.png">

- Yearly schedule in the UI

![Screen Shot 2022-10-12 at 11 18 01 AM](https://user-images.githubusercontent.com/24235898/195623107-596a07e5-14e9-4feb-81d4-52913ff23e56.png)

- No future runs existing

![Screen Shot 2022-10-12 at 11 17 51 AM](https://user-images.githubusercontent.com/24235898/195623224-7fe25a8f-a2eb-434e-a6ec-7ab5149b0a77.png)

### Browers

- [X] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Additional context

- Happens with other iterations of RRule that uses RDATE to schedule

EmilRex:
Notably, when the `RDATE`s are in the past, the schedules do seem to populate but seemingly repeat every 5 days (`RDATE`s shouldn't repeat).

It's also worth noting that `dateutil.rrule` supports `RDATE`s:
```py
In [1]: import pendulum

In [2]: import dateutil.rrule

In [3]: print(pendulum.now("UTC"))
2022-10-13T15:13:19.287422+00:00

[...]

PR #7165, opened 2022-10-14
Improve rrule backend parsing

anticorrelator:
`RRuleSchedule` cannot currently handle compound RRules, `dateutil`'s parser will sometimes return a complex `rruleset` object instead of an `rrule` when parsing an RRule string. I'm adding handling for `rrulesets`.

I could really use someone with some experience with working with schedules to give this a careful once-over to make sure I haven't missed/changed anything critical. I tried to add some tests that ensure that we can properly roundtrip a schedule using the `from_rrule` and `to_rrule` methods, but because of all the inconsistencies in naming timezones across the various timezone parsing libraries, this has proven to be difficult.

### Checklist

### DIFF
--- a/src/prefect/orion/schemas/schedules.py
+++ b/src/prefect/orion/schemas/schedules.py
@@ -344,7 +344,6 @@ class RRuleSchedule(PrefectBaseModel):
 
     rrule: str
     timezone: Optional[str] = Field(default=None, example="America/New_York")
-    timezones: Optional[list[str]] = Field(default=None, example=["America/New_York", "UTC"])
 
     @validator("rrule")
     def validate_rrule_str(cls, v):
@@ -359,7 +358,7 @@ class RRuleSchedule(PrefectBaseModel):
         return v
 
     @classmethod
-    def from_rrule(cls, rrule: dateutil.rrule.rrule):
+    def from_rrule(cls, rrule: dateutil.rrule.rrule, timezone: Optional[str] = None):
         if isinstance(rrule, dateutil.rrule.rrule):
             if rrule._dtstart.tzinfo is not None:
                 timezone = rrule._dtstart.tzinfo.name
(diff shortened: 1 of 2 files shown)


src/prefect/orion/schemas/schedules.py
```diff
@@ -359,32 +359,65 @@ def validate_rrule_str(cls, v):
...
+            if unique_dstarts and unique_timezones:
+                timezone = dtstarts[0].tzinfo.name
+            else:
+                timezone = "UTC"
+            rruleset_string = "\n".join(str(r) for r in rrule._rrule)
+            return RRuleSchedule(rrule=rruleset_string, timezone=timezone)
         else:
-            timezone = "UTC"
-        return RRuleSchedule(rrule=str(rrule), timezone=timezone)
+            raise ValueError(f"Invalid RRule object: {rrule}")
```
chrisguidry, line 386: I might use `NotImplementedError` here, but it's not a huge deal.

src/prefect/orion/schemas/schedules.py
```diff
@@ -359,32 +359,65 @@ def validate_rrule_str(cls, v):
...
+            else:
+                timezone = "UTC"
+            return RRuleSchedule(rrule=str(rrule), timezone=timezone)
+        elif isinstance(rrule, dateutil.rrule.rruleset):
+            dtstarts = [rr._dtstart for rr in rrule._rrule if rr._dtstart is not None]
+            unique_dstarts = set(pendulum.instance(d).in_tz("UTC") for d in dtstarts)
+            unique_timezones = set(d.tzinfo for d in dtstarts if d.tzinfo is not None)
+
+            if len(unique_timezones) > 1:
+                raise ValueError(f"rruleset has too many dtstart timezones: {rrule}")
```
chrisguidry, line 374: Should we also include which timezones we found?

Something like:
```suggestion
                raise ValueError(f"rruleset {rrule} has too many dtstart timezones: {unique_timezones}")
```

I'm not sure what the `repr(rrule)` here will look like, maybe it has all the info we'd need?

tests/orion/schemas/test_schedules.py
```diff
@@ -647,6 +647,52 @@ async def test_rrule_validates_rrule_str(self):
...
+            rrule="DTSTART:19970902T090000\n"
+            "RRULE:FREQ=YEARLY;COUNT=2;BYDAY=TU\n"
+            "RRULE:FREQ=YEARLY;COUNT=1;BYDAY=TH\n"
+        )
+        s2 = RRuleSchedule.from_rrule(s1.to_rrule())
+        s1_dates = await s1.get_dates(5, start=datetime(1900, 1, 1, tz="UTC"))
+        s2_dates = await s2.get_dates(5, start=datetime(1900, 1, 1, tz="UTC"))
+        assert s1_dates == s2_dates
+
+    async def test_rrule_schedule_rejects_rrulesets_with_many_dtstarts(self):
```
chrisguidry, line 673: Hmm, is this actually the test for many timezones?  Or am I misreading this (since the two dtstarts are actually the same UTC time)?

I was expecting to see two exception tests: one for multiple timezones and one for multiple dtstarts.

tests/orion/schemas/test_schedules.py
```diff
@@ -647,6 +647,52 @@ async def test_rrule_validates_rrule_str(self):
...
+        dt_nyc = datetime(2018, 1, 11, 4, tz="America/New_York")
+        dt_chicago = datetime(2018, 1, 11, 3, tz="America/Chicago")
+        rrset = rrule.rruleset(cache=True)
+        rrset.rrule(rrule.rrule(rrule.HOURLY, count=10, dtstart=dt_nyc))
+        rrset.rrule(rrule.rrule(rrule.HOURLY, count=10, dtstart=dt_chicago))
+
+        with pytest.raises(ValueError):
```
chrisguidry, line 680: Can we add the `match="exception message"` part to this assertion so it's clear which of the two exception conditions we're testing?

chrisguidry (commented):
The change looks good and I believe it will address the errors we're seeing in Prefect Cloud, thank you!  I could use some more tests for the different new branches, though.  Nice job only _adding_ tests, that's a great way to instill confidence (instead of also modifying existing tests)

src/prefect/orion/schemas/schedules.py
anticorrelator: it doesn't, specifying the timezones is a good idea

chrisguidry (approved):
Thanks for the new tests, looks good!

EmilRex:
@anticorrelator do you think this would help with #7155?

anticorrelator:
Hey @EmilRex; I did some light manual testing and I don't think it will--I will try to add handling for this as well

PR #7170, opened 2022-10-14
Add last_polled and tracking to work queues

zangell44:
This PR adds a `last_polled` field to work queues and updates that field each time `/get_runs` is called, allowing users to monitor the health of their work queues.

`last_polled` will not be updated if the api request originated from the UI. This PR should go out after the relevant UI changes.

Backend work for https://github.com/PrefectHQ/orion-design/issues/671. See ticket for full discussion.

### Example

```
async def test_read_work_queue_runs_updates_work_queue_last_polled_time(
        self,
        client,
        work_queue,
        session,
    ):
        now = pendulum.now("UTC")
        response = await client.post(
            f"/work_queues/{work_queue.id}/get_runs",
            json=dict(),
        )
        assert response.status_code == status.HTTP_200_OK

        session.expunge_all()
        updated_work_queue = await models.work_queues.read_work_queue(
            session=session, work_queue_id=work_queue.id
        )
        assert updated_work_queue.last_polled > now

        # The Prefect UI often calls this route to see which runs are enqueued.
        # We do not want to record this as an actual poll event.
        ui_response = await client.post(
            f"/work_queues/{work_queue.id}/get_runs",
            json=dict(),
            headers={"X-PREFECT-UI": "true"},
        )
        assert ui_response.status_code == status.HTTP_200_OK

        session.expunge_all()
        ui_updated_work_queue = await models.work_queues.read_work_queue(
            session=session, work_queue_id=work_queue.id
        )
        assert ui_updated_work_queue.last_polled == updated_work_queue.last_polled
```

### Checklist

### DIFF
--- a/src/prefect/orion/api/work_queues.py
+++ b/src/prefect/orion/api/work_queues.py
@@ -6,7 +6,7 @@ from uuid import UUID
 
 import pendulum
 import sqlalchemy as sa
-from fastapi import Body, Depends, Header, HTTPException, Path, status
+from fastapi import BackgroundTasks, Body, Depends, Header, HTTPException, Path, status
 
 import prefect.orion.api.dependencies as dependencies
 import prefect.orion.models as models
@@ -104,6 +104,7 @@ async def read_work_queue(
 
 @router.post("/{id}/get_runs")
 async def read_work_queue_runs(
+    background_tasks: BackgroundTasks,
     work_queue_id: UUID = Path(..., description="The work queue id", alias="id"),
     limit: int = dependencies.LimitBody(),
     scheduled_before: DateTimeTZ = Body(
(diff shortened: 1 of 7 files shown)


madkinsz (approved):
Makes sense to me! Seems like a simpler mechanism than introducing agent identifiers.

src/prefect/orion/api/work_queues.py
```diff
@@ -126,6 +131,17 @@ async def read_work_queue_runs(
...
+        # We do not want to record this as an actual poll event.
+        if not x_prefect_ui:
+            await models.work_queues.update_work_queue(
+                session=session,
+                work_queue_id=work_queue_id,
+                work_queue=schemas.actions.WorkQueueUpdate(
+                    last_polled=pendulum.now("UTC")
+                ),
+            )
+
```
chrisguidry, line 144: It feels like both this and the `record_agent_poll` below should be done in the background.  It would mean opening a new session for those updates, so there's a tradeoff there in terms of latency versus connection capacity.

chrisguidry (approved):
Looks straightforward!  Have a think about whether we should background these...

zangell44:
After discussing internally, we're going to take a slightly different approach here.

PR #7171, opened 2022-10-14
Remove agent table

zangell44:
This PR removes server side tracking of agent polls. This mechanism is unused at the moment. It will be immediately replaced by `work_queue.last_polled` and eventually replaced by a solution with more robust agent observability. 

This is a "breaking" API change in the sense that providing the same payload will not have the same effect. However, the API never exposed `agent.last_activity_time` or any other agent info. 

See also https://github.com/PrefectHQ/prefect/pull/7170 and https://github.com/PrefectHQ/orion-design/issues/671

### Checklist


madkinsz:
I actually want to use the agent table for persisting agent state sometime in the future, is there a pressing reason to delete it?

zangell44:
> I actually want to use the agent table for persisting agent state sometime in the future, is there a pressing reason to delete it?

Nope we can leave it if we have plans for future use! Just trying to keep the code clean

PR #7174, opened 2022-10-14
Add option to drop task or flow return values from memory

madkinsz:
Adds a `cache_result_in_memory` option to flows and tasks to drop a result from memory instead of holding it. You _can_ use this while disabling persistence which is the equivalent of saying "I shall not use this result downstream". By default, using it will enable persistence.

### Example

See the examples in the [documentation](https://deploy-preview-7174--prefect-orion.netlify.app/concepts/results/#caching-of-results-in-memory)

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -294,6 +294,12 @@ Persistence of results requires a [**serializer**](#result-serializers) and a [*
 
 Persistence of the result of a task or flow can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task.
 
+The effect of features on persistence of results:
+- Flow retries: enables persistence of task and subflow results
+- Disabling in-memory caching: enables persistence of the given flow or task result
+- Task cache key: enables persistence of the given task result
+- Task retries: does not affect persistence of results
+
 For example, the following flow has retries enabled. Flow retries require that all task results are persisted, so the task's result will be persisted:
 
 ```python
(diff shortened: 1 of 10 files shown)


docs/concepts/results.md
```diff
@@ -417,9 +417,68 @@ The following data types will be stored by the API without persistence to storag
 If `persist_result` is set to `False`, these values will never be stored.
 
 
+## Caching of results in memory
+
+When running your workflows, Prefect will keep the results of all tasks and flows in memory so they can be passed downstream. In some cases, it is desirable to override this behavior. For example, if you are returning a large amount of data from a task it can be costly to keep it memory for the entire duration of the flow run.
+
+Flows and tasks both include an option to drop the result from memory:
```
tpdorsey, line 424:
```suggestion
Flows and tasks both include an option to drop the result from memory with `cache_result_in_memory`:
```

src/prefect/flows.py
```diff
@@ -236,6 +239,8 @@ def with_options(
             persist_result: A new option for enabling or disabling result persistence.
             result_storage: A new storage type to use for results.
             result_serializer: A new serializer to use for results.
+            cache_result_in_memory: A new toggle indicating if the flow's result should
```
tpdorsey, line 242: Important to note "new" here? Seems like this could become stale.

tpdorsey (approved):
Looks good. Minor comments

src/prefect/flows.py
madkinsz: This is in the `with_options` section for overriding settings on a new copy
```suggestion
            cache_result_in_memory: A new value indicating if the flow's result should
```

docs/concepts/results.md
```diff
@@ -282,7 +282,7 @@ The following Prefect features require results to be persisted:
 
 If results are not persisted, these features will not be usable.
 
-### Configuring results
+### Configuring persistence results
```
madkinsz, line 285:
```suggestion
### Configuring persistence of results
```

madkinsz:
I'm thinking that I should toggle `persist_result` from `None` to `True` when this is used so that the default experience of turning it on is better.

docs/concepts/results.md
madkinsz, line 424:
```suggestion
Flows and tasks both include an option to drop the result from memory with `cache_result_in_memory`:
```

billpalombi (approved):
Both the proposed names and defaults make sense to me.

bunchesofdonald (approved):
I like it!

PR #7187, opened 2022-10-14
Allow deployments to be sorted in the api

pleek91:
# Description
Add DeploymentSort to `/deployments/filter` to allow sorting deployments in the same way that flows are currently sorted.

### DIFF
--- a/src/prefect/orion/api/deployments.py
+++ b/src/prefect/orion/api/deployments.py
@@ -136,7 +136,9 @@ async def read_deployments(
     flow_runs: schemas.filters.FlowRunFilter = None,
     task_runs: schemas.filters.TaskRunFilter = None,
     deployments: schemas.filters.DeploymentFilter = None,
-    sort: schemas.sorting.DeploymentSort = Body(schemas.sorting.DeploymentSort.NAME_ASC),
+    sort: schemas.sorting.DeploymentSort = Body(
+        schemas.sorting.DeploymentSort.NAME_ASC
+    ),
     db: OrionDBInterface = Depends(provide_database_interface),
 ) -> List[schemas.core.Deployment]:
     """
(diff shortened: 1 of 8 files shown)


src/prefect/orion/api/deployments.py
```diff
@@ -136,6 +136,7 @@ async def read_deployments(
     flow_runs: schemas.filters.FlowRunFilter = None,
     task_runs: schemas.filters.TaskRunFilter = None,
     deployments: schemas.filters.DeploymentFilter = None,
+    sort: schemas.sorting.DeploymentSort = Body(schemas.sorting.DeploymentSort.NAME_ASC),
```
madkinsz, line 139: Does sorting by default cost us query performance?

madkinsz (commented):
Nice! Can you add support in `OrionClient.read_deployments` too?

src/prefect/orion/api/deployments.py
madkinsz: cc @chrisguidry ?

pleek91: I do not know. I just tried to follow what flows is doing currently.

pleek91:
> Nice! Can you add support in `OrionClient.read_deployments` too?

Added 👍

src/prefect/orion/api/deployments.py
madkinsz: 👍 We'll want to check for indices on these columns before merging as well.

zangell44: Sorting by default should be fine performance-wise. We should add an index on `deployment.created` though. Happy to commit to this branch if you'd like

pleek91: Please do! Thanks @zangell44

zangell44: pushed!

zangell44 (approved):
LGTM, will get this deployed to Cloud on Monday

PR #7189, opened 2022-10-14
Add backwards compatibility for return of server-states from flows and tasks

madkinsz:
Fixes a backwards compatibility edge-case where an `isinstance` check in `return_value_to_state` did not include server-side state schemas.

### Example

 
 Did not work prior to change, flow ended in `Completed` state and treated returned value as a result:
```python

from prefect.orion.schemas.states import Cancelled

@flow
def foo():
    return Cancelled()
```
The above works now and displays deprecation warnings.

Did work prior to change:
```python

from prefect.states import Cancelled

@flow
def foo():
    return Cancelled()
```

### Checklist

### DIFF
--- a/src/prefect/orion/schemas/states.py
+++ b/src/prefect/orion/schemas/states.py
@@ -141,12 +141,12 @@ class State(IDBaseModel, Generic[R]):
 
     def to_state_create(self):
         # Backwards compatibility for `to_state_create`
-        from prefect.orion.schemas import State
+        from prefect.client.schemas import State
 
         warnings.warn(
-            "`to_state_create` is not supported by `prefect.orion.schemas.states.State` "
-            "and will be removed in a future release. When working with states, use "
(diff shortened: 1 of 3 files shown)


src/prefect/orion/schemas/states.py
```diff
@@ -141,12 +141,12 @@ def result(self, raise_on_failure: bool = True, fetch: Optional[bool] = None):
 
     def to_state_create(self):
         # Backwards compatibility for `to_state_create`
-        from prefect.orion.schemas import State
+        from prefect.client.schemas import State
 
         warnings.warn(
-            "`to_state_create` is not supported by `prefect.orion.schemas.states.State` "
```
madkinsz, line 147: Most of the time, users do not call this directly so I shall provide a bit more generic of a message.

jakekaplan (approved):
Looks good to me!

PR #7192, opened 2022-10-16
UI: fix naming of subflow runs tab on flow run page

stackoverfloweth:
This PR removes a space between sub and flow per documentation standards

> Task run, flow run, and subflow are the current usage in the docs

### DIFF
--- a/orion-ui/src/pages/FlowRun.vue
+++ b/orion-ui/src/pages/FlowRun.vue
@@ -84,7 +84,12 @@
   const flowRunId = useRouteParam('id')
 
   const tabs = computed(() => {
-    const values = ['Logs', 'Task Runs', 'SubFlow Runs', 'Parameters']
+    const values = [
+      'Logs',
+      'Task Runs',
+      'Subflow Runs',
+      'Parameters',
+    ]
 
     if (!media.xl) {
       values.push('Details')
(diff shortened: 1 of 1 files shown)


orion-ui/src/pages/FlowRun.vue
```diff
@@ -84,7 +84,7 @@
   const flowRunId = useRouteParam('id')
 
   const tabs = computed(() => {
-    const values = ['Logs', 'Task Runs', 'Sub Flow Runs', 'Parameters']
+    const values = ['Logs', 'Task Runs', 'SubFlow Runs', 'Parameters']
```
jawnsy, line 87: Should this be Subflow runs, with a lowercase F and R? Are we using sentence case for Task Runs also?

We might also consider putting each element on its own line as well, to make diffs easier to read

jawnsy (approved):

orion-ui/src/pages/FlowRun.vue
stackoverfloweth: good points, both are implemented

madkinsz:
@stackoverfloweth  Please include the `fix` or `enhancement` label on pull requests so they are categorized correctly in the changelog.

PR #7217, opened 2022-10-18
Add support for using callable objects as tasks

### LINKED ISSUE #7196: Tasks do not support callable objects
### First check

### Bug summary

As the title states, callable objects (i.e. those that implement `__call__`) are not supported by tasks.

This happens because callable objects don't have two attributes: `__name__` and `__qualname__`.
The `__name__` restriction can be bypassed by defining the `name` attribute for the task - in which case, the error below is different.

Without the name, the task fails when calling `prefect.utilities.importtools.to_qualified_name`.
https://github.com/PrefectHQ/prefect/blob/8294cb1e9b64660deffb912c9c3c0c9246ba5af6/src/prefect/utilities/importtools.py#L18-L29

Lambdas work because both `__name__` and `__qualname__` are set to `<lambda>`.

### Reproduction

```python3
from typing import Any

import prefect


class CallableObj:
    def __init__(self) -> None:
        self.x = 10

    def __call__(self, *args: Any, **kwds: Any) -> Any:
        print(self.x)


@prefect.flow
def test_flow():
    t = prefect.Task(fn=CallableObj())
    t()
```

#### Without declaring the `name`

```
File "/Users/duarte/Documents/data/pdbt/dystematic/pdbt/flows/raw/epfr/materializations/test_flow.py", line 18, in test_flow
    t = prefect.Task(fn=CallableObj())
  File "/Users/duarte/Documents/data/pdbt/.venv/lib/python3.8/site-packages/prefect/context.py", line 163, in __register_init__
  File "/Users/duarte/Documents/data/pdbt/.venv/lib/python3.8/site-packages/prefect/tasks.py", line 136, in __init__
    self.name = name or self.fn.__name__
AttributeError: 'CallableObj' object has no attribute '__name__'
```

#### Declaring the `name`
```
  File "/Users/duarte/Documents/data/pdbt/.venv/lib/python3.8/site-packages/prefect/context.py", line 163, in __register_init__
  File "/Users/duarte/Documents/data/pdbt/.venv/lib/python3.8/site-packages/prefect/tasks.py", line 142, in __init__
    self.task_key = to_qualified_name(self.fn)
  File "/Users/duarte/Documents/data/pdbt/.venv/lib/python3.8/site-packages/prefect/utilities/importtools.py", line 29, in to_qualified_name
    return obj.__module__ + "." + obj.__qualname__
AttributeError: 'CallableObj' object has no attribute '__qualname__'
```

### Versions

```Text
[11:58:43] Φ prefect version
Version:             2.5.0
API version:         0.8.2
Python version:      3.8.9
Git commit:          eac37918
Built:               Thu, Oct 6, 2022 12:41 PM
Profile:             default
Server:
  Database:          sqlite
  SQLite version:    3.32.3
```

### Additional context

_No response_

jmg-duarte:
Fixes #7196 

### Example

The following example should now work:

```py
from typing import Any

import prefect

class CallableObj:
    def __init__(self) -> None:
        self.x = 10

    def __call__(self, *args: Any, **kwds: Any) -> Any:
        print(self.x)


@prefect.flow
def test_flow():
    t = prefect.Task(fn=CallableObj())
    t()
```

### Checklist

### DIFF
--- a/src/prefect/utilities/importtools.py
+++ b/src/prefect/utilities/importtools.py
@@ -26,6 +26,8 @@ def to_qualified_name(obj: Any) -> str:
     Returns:
         str: the qualified name
     """
+    if not hasattr(obj, "__qualname__"):
+        return obj.__module__ + "." + type(obj).__qualname__
     return obj.__module__ + "." + obj.__qualname__
 
 
--- a/src/prefect/utilities/importtools.py
+++ b/src/prefect/utilities/importtools.py
@@ -26,8 +26,6 @@ def to_qualified_name(obj: Any) -> str:
     Returns:
         str: the qualified name
     """
-    if not hasattr(obj, "__qualname__"):
-        return obj.__module__ + "." + type(obj).__qualname__
     return obj.__module__ + "." + obj.__qualname__
 
 
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -153,7 +153,11 @@ class Task(Generic[P, R]):
         raise_for_reserved_arguments(self.fn, ["return_state", "wait_for"])
 
         self.tags = set(tags if tags else [])
-        self.task_key = to_qualified_name(self.fn)
+
+        if not hasattr(self.fn, "__qualname__"):
+            self.task_key = to_qualified_name(type(self.fn))
+        else:
+            self.task_key = to_qualified_name(self.fn)
 
         self.cache_key_fn = cache_key_fn
         self.cache_expiration = cache_expiration
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -1,7 +1,7 @@
 import datetime
 import inspect
 import warnings
-from typing import Dict, List
+from typing import Any, Dict, List
 from unittest.mock import MagicMock
 from uuid import UUID
 
@@ -154,6 +154,15 @@ class TestTaskCall:
         assert state.is_failed()
         with pytest.raises(ValueError, match="Test"):
             state.result()
+
+    def test_task_supports_callable_objects(self):
+        class A:
+            def __call__(self, *_args: Any, **_kwargs: Any) -> Any:
+                return "hello"
+
+        a = A()
+        task = Task(fn=a, name="Task")
+        assert task.fn is a
 
 
 class TestTaskRun:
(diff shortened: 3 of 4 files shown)


src/prefect/utilities/importtools.py
```diff
@@ -26,6 +26,8 @@ def to_qualified_name(obj: Any) -> str:
     Returns:
         str: the qualified name
     """
+    if not hasattr(obj, "__qualname__"):
```
madkinsz, line 29: Can we update this in `tasks/flows.py` to call `to_qualified_name(type(obj))` instead? `to_qualified_name` _should_ throw an error for instances because we will not be able to recreate it in `from_qualified_name`.

jmg-duarte: Hmmm, I don't think that is a good idea because it will break existing behavior (?)
```
In [1]: def f():
   ...:     pass
   ...: 

In [2]: type(f).__qualname__
Out[2]: 'function'
```

jmg-duarte: Ah, moving the logic to `Task.__init__`, something like:
```py
if not hasattr(obj, "__qualname__"):
    self.task_key = to_qualified_name(type(obj))
else:
    self.task_key = to_qualified_name(obj)
```

madkinsz: Yep!

madkinsz (approved):

madkinsz:
@jmg-duarte The following test fails

```python

    def test_task_run_from_callable_object(self):
        class Foo:
            message = "hello"

            def __call__(self, prefix: str, suffix: str) -> Any:
                return prefix + self.message + suffix

        obj = Foo()
        foo = Task(fn=obj)

        @flow
        def bar():
            return foo("a", suffix="b")

        assert bar() == "ahellob"
```
>  `E       AttributeError: 'Foo' object has no attribute '__name__'`

I may have to revert this change. Do you have time to look at a fix?

jmg-duarte:
@madkinsz as a quick fix: adding `name=...` to the `Task` call works.
As for an actual fix, I suggest using a similar strategy to the one for the `task_key`. If that's ok with you, I can submit a fix using that

madkinsz:
Ah I see I didn't even notice that it was in the `__init__` block. Yeah can you remove the `name=` from your existing test and also include this test that it runs okay? I'm worried we'll inadvertently break something by accessing something that's not present at runtime as well.

jmg-duarte:
> also include this test that it runs okay?

Sorry, I don't understand what you mean by this and by extension, which path forward to take: revert or apply the suggestion I gave

madkinsz:
Sorry! Can you:

- Apply the suggestion you gave in a new pull request
- Remove the `name=...` from the test you added in this pull request so the fix is covered by the test suite
- Include the test I wrote here to ensure that the task can be used in a flow in creation

PR #7222, opened 2022-10-18
Add support for multiple work queue prefixes

### LINKED ISSUE #7221: Support multiple match patterns for `agent start`
### First check

### Prefect Version

2.x

The `-m` flag only supports a single pattern.

Multiple `-m` flags should support several match patterns:

```
$ prefect agent start -m "dev-" -m "stag-"
.....
Listening to `dev-` and `stag-`
```

### Example Use

Refer to the previous section

### Additional context

_No response_

jmg-duarte:
Fixes #7221 

### Example
```
prefect agent start -m "q1-" -m "q2-"
```

### Checklist

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -25,7 +25,7 @@ class OrionAgent:
     def __init__(
         self,
         work_queues: List[str] = None,
-        work_queue_prefix: str = None,
+        work_queue_prefix: List[str] = None,
         prefetch_seconds: int = None,
         default_infrastructure: Infrastructure = None,
         default_infrastructure_document_id: UUID = None,
@@ -63,7 +63,9 @@ class OrionAgent:
 
     async def update_matched_agent_work_queues(self):
         if self.work_queue_prefix:
-            matched_queues = await self.client.match_work_queues(self.work_queue_prefix)
+            matched_queues = []
+            for prefix in self.work_queue_prefix:
+                matched_queues += await self.client.match_work_queues(prefix)
             matched_queues = set(q.name for q in matched_queues)
             if matched_queues != self.work_queues:
                 new_queues = matched_queues - self.work_queues
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -25,7 +25,7 @@ class OrionAgent:
     def __init__(
         self,
         work_queues: List[str] = None,
-        work_queue_prefix: List[str] = None,
+        work_queue_prefix: Union[str, List[str]] = None,
         prefetch_seconds: int = None,
         default_infrastructure: Infrastructure = None,
         default_infrastructure_document_id: UUID = None,
@@ -44,6 +44,8 @@ class OrionAgent:
         self.task_group: Optional[anyio.abc.TaskGroup] = None
         self.client: Optional[OrionClient] = None
 
+        if isinstance(work_queue_prefix, str):
+            work_queue_prefix = [work_queue_prefix]
         self.work_queue_prefix = work_queue_prefix
 
         self._work_queue_cache_expiration: pendulum.DateTime = None
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -63,9 +63,7 @@ class OrionAgent:
 
     async def update_matched_agent_work_queues(self):
         if self.work_queue_prefix:
-            matched_queues = []
-            for prefix in self.work_queue_prefix:
-                matched_queues += await self.client.match_work_queues(prefix)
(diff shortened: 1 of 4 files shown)


madkinsz (commented):
I think we'd want a test that includes coverage where multiple are passed. Otherwise this looks great to me! Thanks for all the contributions :)

src/prefect/agent.py
```diff
@@ -63,7 +63,9 @@ def __init__(
 
     async def update_matched_agent_work_queues(self):
         if self.work_queue_prefix:
-            matched_queues = await self.client.match_work_queues(self.work_queue_prefix)
+            matched_queues = []
+            for prefix in self.work_queue_prefix:
+                matched_queues += await self.client.match_work_queues(prefix)
```
anticorrelator, line 68: I think we'll want to modify `match_work_queues` to accept a list of prefixes, otherwise we'll be unnecessarily polling the server for work queues--in the event we choose to move the matching serverside, accepting multiple prefix patterns should be fairly straightforward as well

jmg-duarte: Definitely a nice catch! Will fix

jmg-duarte:
Functionality is all there, you might want to change the message the agent prints though
```
Agent started! Looking for work from queue(s) that start with the prefix: ['a', 'b']...
```

src/prefect/client/orion.py
```diff
@@ -816,9 +816,11 @@ async def match_work_queues(
...
             if not new_queues:
                 break
-            filtered_queues = list(
-                filter(lambda q: q.name.startswith(prefix), new_queues)
-            )
+            filtered_queues = []
+            for q in new_queues:
+                for p in prefixes:
+                    if q.name.startswith(p):
+                        filtered_queues.append(q)
```
anticorrelator, line 823: this looks good! I think it might be a good idea to convert this to a set or something similar to avoid duplicate matches?

jmg-duarte: I thought about it, but that implies that `new_queues` is a misnomer and it doesn't in fact, just hold new queues.

I just extended the already present logic to handle several filters.
https://github.com/PrefectHQ/prefect/pull/7222/files#diff-ed60480b777e77bdeeced26ed733ea8d0da54e3da5a6dce53054258a1625b004L811-L815

jmg-duarte: @anticorrelator following up, do you have further comments?

anticorrelator: nope, this looks good to me!

anticorrelator: @jmg-duarte Actually, taking a look at the failing test it does look like queues might be being matched multiple times--for example if we have two prefixes specified but a queue name matches against both, it'll be added twice to the list

jmg-duarte: Alright, I'll make a set! Thanks for the feedback!

src/prefect/client/orion.py
```diff
@@ -794,35 +794,37 @@ async def read_work_queues(
...
             )
             if not new_queues:
                 break
-            filtered_queues = list(
-                filter(lambda q: q.name.startswith(prefix), new_queues)
-            )
-            work_queues += filtered_queues
+            filtered_queues = set()
+            for q in new_queues:
+                for p in prefixes:
```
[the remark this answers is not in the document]
jmg-duarte: Oh I see my mistake now 🤦 I totally get the confusion now. My bad

src/prefect/agent.py
```diff
@@ -25,7 +25,7 @@ class OrionAgent:
     def __init__(
         self,
         work_queues: List[str] = None,
-        work_queue_prefix: str = None,
+        work_queue_prefix: List[str] = None,
```
madkinsz, line 28: This and the change in the client are technically breaking changes as people can be using these interfaces. Can you update this to take a `Union[str, List[str]]` and just coerce the `str` to ` List[str]` in `__init__`?

src/prefect/client/orion.py
```diff
@@ -794,13 +794,13 @@ async def read_work_queues(
 
     async def match_work_queues(
         self,
-        prefix: str,
+        prefixes: List[str],
```
madkinsz, line 797: I think that it is less likely that anyone is using this interface directly and am okay with a breaking change here. Please keep an eye out of interface changes in the future though.

jmg-duarte: Agreed on this case.

src/prefect/agent.py
jmg-duarte: Done.

anticorrelator (approved):
thank you so much for the contribution!

PR #7227, opened 2022-10-18
Fix handling of `cache_result_in_memory` in `Task.with_options`

madkinsz:
This one has bitten me once again. We'll need to be careful about test coverage for this in the future.

### Example

The bug:

```python
@task
def foo():
   pass

bar = foo.with_options()

# Using `bar` will fail since `cache_result_in_memory` will be `None` which is not allowed
# To workaround, specify the value:

bar = foo.with_options(cache_result_in_memory=True)
```

After this fix, `with_options` works as intended.

### Checklist

### DIFF
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -284,7 +284,7 @@ class Task(Generic[P, R]):
             ),
             cache_result_in_memory=(
                 cache_result_in_memory
-                if cache_result_in_memory is None
+                if cache_result_in_memory is not None
                 else self.cache_result_in_memory
             ),
         )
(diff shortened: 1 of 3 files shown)


madkinsz:
The failing test in CI: https://github.com/PrefectHQ/prefect/actions/runs/3275902168/jobs/5391316583 is the one that alerted me to this issue.

src/prefect/tasks.py
```diff
@@ -284,7 +284,7 @@ def with_options(
             ),
             cache_result_in_memory=(
                 cache_result_in_memory
-                if cache_result_in_memory is None
+                if cache_result_in_memory is not None
```
bunchesofdonald, line 287: I swear I stared at this line for a while in the last PR thinking something was wrong, but convinced myself it was right. Good catch!

bunchesofdonald (approved):

PR #7228, opened 2022-10-18
Add `PREFECT_RESULTS_PERSIST_BY_DEFAULT` to globally toggle the result persistence default

madkinsz:
Adds a global setting to change the default value for persistence e.g. when opt-in/out is not provided and a feature is not used.

Simplified version of #7161 

### Example

To enable persistence of all flow, task, and subflow results by default:
```
$ prefect config set PREFECT_RESULTS_PERSIST_BY_DEFAULT=true
```

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -293,7 +293,7 @@ Persistence of results requires a [**serializer**](#result-serializers) and a [*
 
 #### Toggling persistence
 
-Persistence of the result of a task or flow can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task.
+Persistence of the result of a task or flow can be configured with the `persist_result` option. The `persist_result` option defaults to a null value, which will automatically enable persistence if it is needed for a Prefect feature used by the flow or task. Otherwise, persistence is disabled by default.
 
 For example, the following flow has retries enabled. Flow retries require that all task results are persisted, so the task's result will be persisted:
 
@@ -310,6 +310,8 @@ def my_flow():
     # so Prefect will persist its result at runtie
     my_task()
 ```
+
+Flow retries do not require the flow's result to be persisted, so it will not be.
 
 In this next example, one task has caching enabled. Task caching requires that the given task's result is persisted:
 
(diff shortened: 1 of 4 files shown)


docs/concepts/results.md
```diff
@@ -357,6 +359,14 @@ def my_task():
 
 Toggling persistence manually will always override any behavior that Prefect would infer.
 
+You may also change Prefect's default persistence behavior with the `PREFECT_RESULTS_PERSIST_BY_DEFAULT` setting. To persist results by default, even if they are not needed for a feature change the value to a truthy value:
+
+```
+$ prefect config set PREFECT_RESULTS_PERSIST_BY_DEFAULT=true
+```
+
+Task and flows with `persist_result=False` will not persist their results even if `PREFECT_RESULTS_PERSIST_BY_DEFAULT` is `true`.
```
anna-geller, line 368: this is awesome!

anna-geller (approved):
Very cool!

docs/concepts/results.md
```diff
@@ -357,6 +359,14 @@ def my_task():
 
 Toggling persistence manually will always override any behavior that Prefect would infer.
 
+You may also change Prefect's default persistence behavior with the `PREFECT_RESULTS_PERSIST_BY_DEFAULT` setting. To persist results by default, even if they are not needed for a feature change the value to a truthy value:
```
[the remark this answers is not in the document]
madkinsz: I'll change it in the code block too but gotta commit locally.

src/prefect/results.py
```diff
@@ -49,6 +50,13 @@ def get_default_result_serializer() -> ResultSerializer:
     return PREFECT_RESULTS_DEFAULT_SERIALIZER.value()
 
 
+def get_default_persist_setting() -> bool:
```
bunchesofdonald, line 53: I don't have a suggestion for a better name, but this makes me think that we're getting the default setting, a setting a user could override and not the 'default' set by the user. Any reason to not just use `PREFECT_RESULTS_PERSIST_BY_DEFAULT.value()` directly?

bunchesofdonald (approved):

src/prefect/results.py
madkinsz: I could use it directly, but I'm following this pattern because we use it for all the other default values. We are likely to extend these to pull settings from the API in the future. I can try to think of a clearer name.

madkinsz: I don't really have a good idea and since all three of these `get_default_...` methods read settings I'm going to leave it as is for now.

PR #7231, opened 2022-10-18
Fixes integration tests for 2.6 to include the current client

madkinsz:
The client tests are being skipped e.g. https://github.com/PrefectHQ/prefect/actions/runs/3268773790/jobs/5375567361

This led to a bad release in 2.6.2 and was definitely my fault as I added it incorrectly. I want to generate the matrix dynamically at some point, perhaps that point is soon as it would guard against this.

## Example

https://github.com/PrefectHQ/prefect/actions/runs/3276226541/jobs/5392036236


abrookins (approved):
LGTM 👍

PR #7242, opened 2022-10-19
Handle __name__ for callable objects

### LINKED ISSUE #7241: Creating a Task from a callable object fails if the task does not have a name
@jmg-duarte The following test fails

```python

    def test_task_run_from_callable_object(self):
        class Foo:
            message = "hello"

            def __call__(self, prefix: str, suffix: str) -> Any:
                return prefix + self.message + suffix

        obj = Foo()
        foo = Task(fn=obj)

        @flow
        def bar():
            return foo("a", suffix="b")

        assert bar() == "ahellob"
```
>  `E       AttributeError: 'Foo' object has no attribute '__name__'`

I may have to revert this change. Do you have time to look at a fix?

_Originally posted by @madkinsz in https://github.com/PrefectHQ/prefect/issues/7217#issuecomment-1284169478_

jmg-duarte:
Fixes #7241 

### Example

```py
 def test_task_run_from_callable_object(self):
        class Foo:
            message = "hello"

            def __call__(self, prefix: str, suffix: str) -> Any:
                return prefix + self.message + suffix

        obj = Foo()
        foo = Task(fn=obj)

        @flow
        def bar():
            return foo("a", suffix="b")

        assert bar() == "ahellob"
```

### Checklist

### DIFF
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -152,7 +152,7 @@ class Task(Generic[P, R]):
             if not hasattr(self.fn, "__name__"):
                 self.name = type(self.fn).__name__
             else:
-                self.task_key = self.fn.__name__
+                self.name = self.fn.__name__
         else:
             self.name = name
 
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -148,7 +148,14 @@ class Task(Generic[P, R]):
         self.fn = fn
         self.isasync = inspect.iscoroutinefunction(self.fn)
 
-        self.name = name or self.fn.__name__
+        if not name:
+            if not hasattr(self.fn, "__name__"):
+                self.name = type(self.fn).__name__
+            else:
+                self.task_key = self.fn.__name__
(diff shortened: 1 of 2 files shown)


jmg-duarte:
@madkinsz for the sake of completeness, I've kept both tests. I can try to parametrize them to shorten the length of the file but I don't see that as a blocker

madkinsz (approved):
Thanks again!

src/prefect/tasks.py
```diff
@@ -148,7 +148,14 @@ def __init__(
         self.fn = fn
         self.isasync = inspect.iscoroutinefunction(self.fn)
 
-        self.name = name or self.fn.__name__
+        if not name:
+            if not hasattr(self.fn, "__name__"):
+                self.name = type(self.fn).__name__
+            else:
+                self.task_key = self.fn.__name__
```
madkinsz, line 155:
```suggestion
                self.name = self.fn.__name__
```

PR #7244, opened 2022-10-19
Allow image/namespace fields to be loaded from Kubernetes job manifest

### LINKED ISSUE #7245: Kubernetes manifest image and namespace shortcuts should be optional
### First check

### Prefect Version

2.x

Reported by one of our community members: https://prefect-community.slack.com/archives/CL09KU1K7/p1666199414970169

The `image` and `namespace` shortcuts on the `KubernetesJob` block are not optional, so you cannot pass them `None` to avoid injecting them.  Even worse, if you provide a full manifest file (like with `KubernetesJob.job_from_file`) the manifest's values will be overwritten by the Prefect defaults.

Instead, we should make the `image`/`namespace` attributes optional (with their current defaults) but all `None` to disable the shortcuts.

### Example Use

_No response_

### Additional context

_No response_

madkinsz:
Closes #7245

Only sets a default value for `image` and `namespace` if they are not provided in the given manifest. This allows users to provide a manifest that contains these preset.

### Example

Given `job.yaml`:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  generateName: prefect-job-d41d8cd98f00b204e9800998ecf8427e
  labels: {}
  namespace: my-namespace
spec:
  template:
    spec:
      completions: 1
      containers:
      - env: []
        image: my-image
        name: prefect-job
      parallelism: 1
      restartPolicy: Never
```

The image and namespace should be respected by this block:

```python
job = KubernetesJob(job=KubernetesJob.from_file("job.yaml"))
print(job.preview())
```

However, we overwrite `namespace` and `image` with Prefect's default values. This pull request updates this to respect the values in the job manifest if they exist.

### Checklist

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -172,9 +172,10 @@ class KubernetesJob(Infrastructure):
 
     @root_validator
     def default_namespace(cls, values):
-        job = values["job"]
+        job = values.get("job")
+
         namespace = values.get("namespace")
-        job_namespace = job["metadata"].get("namespace")
+        job_namespace = job["metadata"].get("namespace") if job else None
 
         if not namespace and not job_namespace:
             values["namespace"] = "default"
@@ -183,9 +184,13 @@ class KubernetesJob(Infrastructure):
 
     @root_validator
     def default_image(cls, values):
-        job = values["job"]
+        job = values.get("job")
         image = values.get("image")
-        job_image = job["spec"]["template"]["spec"]["containers"][0].get("image")
+        job_image = (
+            job["spec"]["template"]["spec"]["containers"][0].get("image")
+            if job
+            else None
+        )
 
         if not image and not job_image:
             values["image"] = get_prefect_image_name()
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -75,11 +75,11 @@ class KubernetesJob(Infrastructure):
         default="kubernetes-job", description="The type of infrastructure."
     )
     # shortcuts for the most common user-serviceable settings
-    image: str = Field(
+    image: Optional[str] = Field(
         default_factory=get_prefect_image_name,
         description="The tag of a Docker image to use for the job. Defaults to the Prefect image.",
     )
-    namespace: str = Field(
+    namespace: Optional[str] = Field(
         default="default", description="The Kubernetes namespace to use for this job."
     )
     service_account_name: Optional[str] = Field(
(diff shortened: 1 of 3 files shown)


src/prefect/infrastructure/kubernetes.py
```diff
@@ -278,18 +278,25 @@ def _shortcut_customizations(self) -> JsonPatch:
...
+        if self.image:
+            shortcuts.append(
+                {
+                    "op": "add",
+                    "path": "/spec/template/spec/containers/0/image",
+                    "value": self.image,
+                }
+            )
 
         shortcuts += [
```
chrisguidry, line 301: We _could_ move the labels up as the initial list assignment (not a big deal):

```
shortcuts = [...the labels...]

if self.namespace:
  shortcuts.append(...)
```

src/prefect/infrastructure/kubernetes.py
```diff
@@ -75,11 +75,11 @@ class KubernetesJob(Infrastructure):
         default="kubernetes-job", description="The type of infrastructure."
     )
     # shortcuts for the most common user-serviceable settings
-    image: str = Field(
+    image: Optional[str] = Field(
         default_factory=get_prefect_image_name,
         description="The tag of a Docker image to use for the job. Defaults to the Prefect image.",
```
chrisguidry, line 80:
```suggestion
        description="The tag of a Docker image to use for the job. Defaults to the Prefect image.  Use `None` if you will be customizing the `job` and already include an `image` in your manifest",
```

src/prefect/infrastructure/kubernetes.py
```diff
@@ -75,11 +75,11 @@ class KubernetesJob(Infrastructure):
...
     )
-    namespace: str = Field(
+    namespace: Optional[str] = Field(
         default="default", description="The Kubernetes namespace to use for this job."
```
chrisguidry, line 83:
```suggestion
        default="default", description="The Kubernetes namespace to use for this job.  Use `None` if you will be customizing the `job` and already define the `namespace` in your manifest"
```

chrisguidry (approved):
Just a thought for the parameter descriptions, but otherwise I think this will do the trick!

src/prefect/infrastructure/kubernetes.py
madkinsz: Not taken verbatim, but addressed in https://github.com/PrefectHQ/prefect/pull/7244/commits/cd7629b06bca087576371610f6d4ef537974ace9

peytonrunyan (approved):
It reads reasonably to me!

PR #7248, opened 2022-10-19
Add support for annotations to `visit_collection`

madkinsz:
Closes https://github.com/PrefectHQ/prefect-dask/issues/43

Items inside annotations like `allow_failure` are not visited by `visit_collection` which means that when we wrap futures e.g. in https://github.com/PrefectHQ/prefect-dask/issues/43 they may not be resolved correctly. This pull request adds a base class for annotations and support for visiting.

### Checklist

### DIFF
--- a/src/prefect/utilities/annotations.py
+++ b/src/prefect/utilities/annotations.py
@@ -1,9 +1,27 @@
-from typing import Any, Generic, TypeVar
+from abc import ABC
+from typing import Generic, TypeVar
 
 T = TypeVar("T")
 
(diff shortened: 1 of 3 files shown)


madkinsz:
This breaks `allow_failure` in some cases because we visit the future separately from the annotation. I'll need to poke around at a way to fix that. We could also just add annotation support to the `visit_fn` in `prefect-dask`.

bunchesofdonald (approved):
❤️

PR #7249, opened 2022-10-19
New flow run path to reflect notifications path provided by the API

nicolascribbles:
#### Problem
Solves Issue #7250 

- The API notifications was providing a route to `/flow-runs/flow-run/:id` and our URL was `/flow-run/:id`, so users would land on a 404 page when clicking on the URL provided by the notification.
- The "Flow Runs" navigation item on the menu to the left of the dashboard was not highlighted when you traveled to an individual route because the routes are not children of the `/runs` route

#### What Changed
- `/runs` is going away, redirect was created to reflect new path `/flow-runs`
- The individual flow run is registered as a child of the flow-runs route therefore `Flow Runs` navigation link is now active when we are on an individual flow run and its radar.

### Example
[screen-capture (35).webm](https://user-images.githubusercontent.com/39339887/196779909-131b54f1-9073-4148-b04d-58ea3442526a.webm)
```
  {
    name: 'flow-runs-old',
    path: '/runs',
    redirect: routes.flowRuns(),
  },
  {
    path: '/flow-runs',
    children: [
      {
        name: 'flow-runs',
        path: '',
        component: FlowRunsPage,
      },
      {
        path: 'flow-run/:id',
        children: [
          {
            name: 'flow-run',
            path: '',
            component: (): RouteComponent => import('@/pages/FlowRun.vue'),
          },
          {
            name: 'radar',
            path: 'radar',
            component: (): RouteComponent => import('@/pages/FlowRunRadar.vue'),
          },
        ],
      },
    ],
  },
```

### Checklist


orion-ui/src/router/index.ts
```diff
@@ -11,22 +11,32 @@ const routeRecords: AppRouteRecord[] = [
     redirect: routes.flowRuns(),
   },
   {
-    name: 'flow-runs',
```
zhen0, line 14: I wonder if we could just take this section out?  Any ideas if we need to keep it for anything @pleek91?

orion-ui/src/router/routes.ts
```diff
@@ -11,6 +11,7 @@ export const routes = {
   flowRun: (id: string) => ({ name: 'flow-run', params: { id } }) as const,
   flowRunCreate: (deploymentId: string) => ({ name: 'flow-run.create', params: { deploymentId } }) as const,
   flowRuns: () => ({ name: 'flow-runs' }) as const,
+  flowRunsOld: () => ({ name: 'flow-runs-old' }) as const,
```
zhen0, line 14: Same question about whether we need to keep this but thanks for considering it!

zhen0:
Great job figuring this one out @nicolascribbles!  The routing looks good to me!  I'll let @pleek91 comment on whether we need to keep the old 'runs' route or if we can remove it.

orion-ui/src/router/index.ts
nicolascribbles: I only put this here in case someone has the page "starred" but they can always reaccess the page by clicking!

orion-ui/src/router/routes.ts
nicolascribbles: Happy to remove it if its requested though!

zhen0:
We chatted this one through and agreed we should follow this pattern for flows, deployments, work queues etc to make them highlighted correctly in the sidebar.

PR #7251, opened 2022-10-19
UI: Add `resolve.dedupe` settings for vite

pleek91:
# Description
We've been having quite a few issues when linking orion-design locally for development. Adding `vue` to vite's dedupe config seems to resolve those issues.

### DIFF
--- a/orion-ui/vite.config.ts
+++ b/orion-ui/vite.config.ts
@@ -8,6 +8,7 @@ export default defineConfig({
   base: process.env.ORION_UI_SERVE_BASE ?? '',
   resolve: {
     alias: [{ find: '@', replacement: resolve(__dirname, './src') }],
+    dedupe: ['vue'],
   },
   css: {
     devSourcemap: true,


nicolascribbles (approved):
😱 nice - 1 liner

znicholasbrown:
Vue is listed as a peer dependency for `orion-design` so this solution is a little confusing... they should share the same installation, no?

pleek91:
@znicholasbrown I don't totally understand but dedupe is a package feature. Npm even has its own dedupe flag.  @stackoverfloweth tested trying to dedupe using npm and didn't help. I found this in a [vite github issue](https://github.com/vitejs/vite/issues/7454) where someone was having the same issue we're experience and this was the suggested solution. 

There are lots of issues with peer dependencies in general. But its my understanding that marking something as a peer dependency only enforces the consuming package meets that requirement. It doesn't change at all how the actual code is bundled or packaged. Similar to how rollup's external settings work. From my testing, setting vue as external (as described [here](https://vitejs.dev/guide/build.html#library-mode)) fixes this same issue when consuming published packages. But the issue can be reintroduced when using packages locally because of how npm treats symlinks and how vite is producing the bundles. The [vite docs](https://vitejs.dev/config/shared-options.html#resolve-dedupe) for `resolve.dedupe` mentions "linked packages in monorepos" as a reason you might want to use the `dedupe` setting. 

I'm definitely not 100% confident in this fix. But from my testing it fixes the development issue we've been experience and doesn't seem to have any negative side effects with the build itself.

madkinsz:
@pleek91 if this is a user-facing enhancement, can you update the title to reflect the change in UX? If it's not user facing, can you change the label to `maintenance` so it is not included in the changelog?

PR #7263, opened 2022-10-20
Extend `visit_collection` with support for annotations

madkinsz:
Extends https://github.com/PrefectHQ/prefect/pull/7248 but solves https://github.com/PrefectHQ/prefect-dask/issues/43 without requiring the changes in https://github.com/PrefectHQ/prefect-dask/pull/45

There is a new `context` dictionary on `visit_collection`. It is `None` by default and ignored. If set, it can be used to store contextual information that persists as `visit_collection` descends into nested items. A shallow copy is used to prevent the changes from leaking "up".

This `context` dictionary is populated with an annotation when descending into annotations. Previously, `visit_collection` would never descend into annotations. We now explicitly do not parse object within quotes in the engine instead of relying on `visit_collection` to ignore it.

Additionally, `visit_collection` now has a `remove_annotations: bool  = False` setting that is used during `resolve_inputs` to strip annotations from inputs while preserving other behavior i.e. visiting the items within the annotation.

### Checklist


src/prefect/engine.py
```diff
@@ -1418,7 +1412,7 @@ def resolve_input(expr):
         # Do not allow uncompleted upstreams except failures when `allow_failure` has
         # been used
         if not state.is_completed() and not (
-            should_allow_failure and state.is_failed()
+            isinstance(context.get("annotation"), allow_failure) and state.is_failed()
```
madkinsz, line 1415: Hm. This will break when annotations are nested e.g. `allow_failure(quote("foo"))` — maybe I should make the context have a list or set of annotations?

tests/utilities/test_collections.py
```diff
@@ -400,6 +407,44 @@ def test_visit_collection_max_depth(self, inp, depth, expected):
...
+
+        def visit(expr, context):
+            if isinstance(expr, list):
+                context["depth"] += 1
+                return expr
+            else:
+                return expr + context["depth"]
+
+        result = visit_collection(foo, visit, context={"depth": 0}, return_data=True)
+        assert result == [2, 3, [5, 6], [7, [9, 10]], 9, 10]
```
peytonrunyan, line 421: nit - a comment here giving a tldr of why we're expecting this, or a short docstring in the visit function would reduce the overhead and context needed to understand the behavior of `visit_collection` from this test.

src/prefect/engine.py
cicdw: That makes sense to me - it seems that you could even change the key name to `annotations` without any backwards compatible breakage since the flow is re-parsed with each run

madkinsz: I'm going to wait for this to be an actual issue so we can keep this simple and get it out there

madkinsz:
@peytonrunyan please re-review

peytonrunyan (approved):
lgtm!

PR #7276, opened 2022-10-20
Fix with_options override with 0s

ahuang11:
Previously, using `or` made it go back to the initially set `retries`, not the overrided one.

### Example

```python

@task(retries=1, retry_delay_seconds=1)
def test_task():
    1 / 0

@flow
def test_flow():
    return test_task.with_options(retries=0)()


test_flow()
```

Would still retry at least once before this PR.

### Checklist


ahuang11:
Don't think the error is related to this PR.

src/prefect/tasks.py
```diff
@@ -280,8 +280,10 @@ def with_options(
             tags=tags or copy(self.tags),
             cache_key_fn=cache_key_fn or self.cache_key_fn,
             cache_expiration=cache_expiration or self.cache_expiration,
-            retries=retries or self.retries,
-            retry_delay_seconds=retry_delay_seconds or self.retry_delay_seconds,
+            retries=retries if retries is not NotSet else self.retries,
+            retry_delay_seconds=retry_delay_seconds
+            if retry_delay_seconds is not NotSet
+            else self.retry_delay_seconds,
```
madkinsz, line 286: Can you add some parens here so it formats more readable?

madkinsz (approved):

madkinsz:
@ahuang11 this is being rewritten at "Allow `with_options` to reset retries and retry delays" in the release notes — try to make the title more clear in future changes :) Thanks!

PR #7282, opened 2022-10-21
Add timezone param to deployment build

### LINKED ISSUE #7202: Timezone parameter to deployment cli for cron schedule
### First check

### Prefect Version

2.x

Currently when we add a deployment through cli with a cron schedule the default timezone "UTC" is used for the schedule. 
When deploying with a Python script we can use `CronSchedule()` and specify a timezone there.

There should be a additional parameter to specify the timezone eg. `--crontz` which only can be set if `--cron` was specified. This is fully optional. If not specified UTC is used as a fallback.

Another solution would be to specifiy a default timezone for the whole orion instance.

### Example Use

`prefect deployment build --cron "0 4 * * *"....` -> results in "0 4 * * *" in UTC timezone (current behavior).

`prefect deployment build --cron "0 4 * * *" --crontz "Europe/Berlin"....` -> results in "0 4 * * *" in Europe/Berlin timezone.

### Additional context

_No response_

anticorrelator:
closes #7202

Adds timezone handling to `prefect deployment build` to bring its behavior more in line with `prefect deployment set-schedule`. Also improves handling of anchor dates for interval schedules and explicit timezone overrides for RRule schedules.

### Checklist

### DIFF
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -238,6 +238,9 @@ async def set_schedule(
     cron_schedule = {"cron": cron_string, "day_or": cron_day_or, "timezone": timezone}
     if rrule_string is not None:
         rrule_schedule = json.loads(rrule_string)
+        if timezone:
+            # override timezone if specified via CLI argument
+            rrule_schedule.update({"timezone": timezone})
     else:
         # fall back to empty schedule dictionary
         rrule_schedule = {"rrule": None}
@@ -686,7 +689,7 @@ async def build(
     timezone: str = typer.Option(
         None,
         "--timezone",
-        help="A timezone that will be passed to cron or interval schedules.",
+        help="Deployment schedule timezone string e.g. 'America/New_York'",
     ),
     path: str = typer.Option(
         None,
@@ -800,8 +803,11 @@ async def build(
     elif rrule:
         try:
             schedule = RRuleSchedule(**json.loads(rrule))
+            if timezone:
+                # override timezone if specified via CLI argument
+                schedule.timezone = timezone
         except json.JSONDecodeError:
-            schedule = RRuleSchedule(rrule=rrule)
+            schedule = RRuleSchedule(rrule=rrule, timezone=timezone)
 
     # parse storage_block
     if storage_block:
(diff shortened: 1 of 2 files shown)


src/prefect/cli/deployment.py
```diff
@@ -675,11 +678,19 @@ async def build(
         "--interval",
         help="An integer specifying an interval (in seconds) that will be used to set an IntervalSchedule on the deployment.",
     ),
+    interval_anchor: Optional[str] = typer.Option(
+        None, "--anchor-date", help="The anchor date for an interval schedule"
```
anna-geller, line 682: do we expect an ISO-formatted string? it might be worth mentioning the expected format in this help message; I'd intuitively understand that you would expect 2022-12-24 but someone could try to do 2022/24/12

tests/cli/test_deployment_cli.py
```diff
@@ -372,6 +440,35 @@ def test_parsing_rrule_schedule_json(self, patch_import, tmp_path):
...
+                "build",
+                "fake-path.py:fn",
+                "-n",
+                "TEST",
+                "-o",
+                str(tmp_path / "test.yaml"),
+                "--rrule",
+                '{"rrule": "DTSTART:20220910T110000\\nRRULE:FREQ=HOURLY;BYDAY=MO,TU,WE,TH,FR,SA;BYHOUR=9,10,11,12,13,14,15,16,17", "timezone": "America/New_York"}',
+                "--timezone",
+                "Europe/Berlin",
```
anna-geller, line 460: sanity check: is this expected that the user can set the timezone on both rrule and separately? I'd expect if the one on rrule dict is set, no extra --timezone flag should be provided, but no strong opinions on what the right behavior should be here

anna-geller (approved):
Looks great! QA passes, see screenshot

image

src/prefect/cli/deployment.py
anticorrelator: I think it's anything that's parse-able with `pendulum.parse`, I'd have to look up the details there

tests/cli/test_deployment_cli.py
anticorrelator: I do think it's important to be able to override the timezone explicitly here even if one is provided, `RRule` strings can be complex and specify multiple timezones for each of its components

anna-geller: I see, some users could be confused which one is the one that will get used in the end, but this is probably sth we can tackle with docs, thx for clarifying

madkinsz (approved):

PR #7284, opened 2022-10-21
UI: Redirect back to block form after creating a nested block

pleek91:
# Description
If there is a redirect in the url of the block create page navigate to that rather than the newly created block. 

Needs https://github.com/PrefectHQ/orion-design/pull/713


stackoverfloweth (approved):

WillRaphaelson:
Nice

PR #7316, opened 2022-10-25
Add persistence of data when users return states from tasks/flows

madkinsz:
### Example

Previously, the following example would fail with a `MissingResult` exception since the user did not provide `data` for the state. Now it will work with `None` as the result.

```python
@flow
def foo():
  return Completed(message="test")

foo()
```

Previously, the following example would result in unpersisted data. Now, it will persist the data attached to the state:
```python
@flow(persist_result=True)
def foo():
  return Completed(data="test")

foo()

# If read from the API later, the flow run state would have no data
```

The following example does not have any change in behavior, the user can attach a custom `BaseResult`:

```python
@flow
def foo():
  return Completed(data=LiteralResult(value=False))

foo()
```

The following example does not have any change in behavior, the user can attach data without persistence:

```python
@flow(persist_result=False)
def foo():
  return Completed(data="foo!")

foo()
```

### Checklist

### DIFF
--- a/src/prefect/states.py
+++ b/src/prefect/states.py
@@ -214,7 +214,14 @@ async def return_value_to_state(retval: R, result_factory: ResultFactory) -> Sta
         and not retval.state_details.flow_run_id
         and not retval.state_details.task_run_id
     ):
-        return retval
+        state = retval
+
+        # Unless the user has already constructed a result explicitly, use the factory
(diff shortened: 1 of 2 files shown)


anna-geller (approved):
Looks great! I QAed and it's working as described in PR

bunchesofdonald (approved):

PR #7328, opened 2022-10-26
Restores proxy-awareness for the Orion client for remote Orion servers.

### LINKED ISSUE #7204: Add configuration for all Prefect API calls to be proxied
### First check

### Prefect Version

2.x

httpx honors setting the HTTPS_PROXY env var which proxies all traffic, which means that the Prefect API calls will be proxied, but it also means that private network traffic will also get proxied.

Adding a configuration option in the Prefect configuration for a proxy you can pass it through to httpx and aiohttp internally, meaning that only traffic to Prefect API will use the proxy.

The libraries mentioned in the docs, httpx and aiohttp, both provide support for specifying proxies ([here](https://www.python-httpx.org/advanced/#http-proxying) and [here](https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support)).

A similar feature for a slack notification handler was implemented here https://github.com/PrefectHQ/prefect/issues/5236.

### Example Use

In CLI `prefect config set HTTPS_PROXY=https://proxy.quackme.co` for example

### Additional context

_No response_

chrisguidry:
We were overriding the `transport` on our Orion connections in order to set the number of retries.  When doing this, we would need to include all of the logic for how to discover if the user has proxies configured in their environment. Instead, just set the retries _after_ the fact.

Fixes PrefectHQ/nebula#2356
Fixes #7204

See the background thread in the [Prefect Community Slack](https://linen.prefect.io/t/2930204/Hi-I-m-having-some-issues-with-accessing-prefect-2-0-cloud-f) for more information.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -13,7 +13,7 @@ tags:
 
 # Installation
 
-The first step to getting started with Prefect is installing the Prefect Python package. 
+The first step to getting started with Prefect is installing the Prefect Python package.
 
 <p align="left">
     <a href="https://pypi.python.org/pypi/prefect/" alt="PyPI version">
@@ -43,7 +43,7 @@ We recommend installing Prefect 2 using a Python virtual environment manager suc
 !!! warning "Windows and Linux requirements"
     See [Windows installation notes](#windows-installation-notes) and [Linux installation notes](#linux-installation-notes) for details on additional installation requirements and considerations.
 
-## Install Prefect 
+## Install Prefect
 
 The following sections describe how to install Prefect in your development or execution environment.
 
@@ -122,12 +122,12 @@ Server:
 
 ## Windows installation notes
 
-Prefect 2 supports running Prefect flows on Windows. 
+Prefect 2 supports running Prefect flows on Windows.
 
 !!! note "Prefect on Windows requires Python 3.8 or later"
     Make sure to use Python 3.8 or higher when running a Prefect agent on Windows.
 
-You can install and run Prefect as described above via Windows PowerShell, the Windows Command Prompt, or [`conda`](https://docs.conda.io/projects/conda/en/latest/user-guide/install/windows.html). Note that, after installation, you may need to manually add the Python local packages `Scripts` folder to your `Path` environment variable. 
+You can install and run Prefect as described above via Windows PowerShell, the Windows Command Prompt, or [`conda`](https://docs.conda.io/projects/conda/en/latest/user-guide/install/windows.html). Note that, after installation, you may need to manually add the Python local packages `Scripts` folder to your `Path` environment variable.
 
 The `Scripts` folder path looks something like this (the username and Python version may be different on your system):
 
@@ -140,10 +140,10 @@ Watch the `pip install` installation output messages for the `Scripts` folder pa
 If using Windows Subsystem for Linux (WSL), see [Linux installation notes](#linux-installation-notes).
 
 !!! note "Windows support is under development"
-    Support for Prefect on Windows is a work in progress. 
-    
-    Right now, we're focused on your ability to develop and run flows and tasks on Windows, along with running the API server, orchestration engine, and UI. 
-    
+    Support for Prefect on Windows is a work in progress.
+
+    Right now, we're focused on your ability to develop and run flows and tasks on Windows, along with running the API server, orchestration engine, and UI.
+
     If you encounter unexpected issues, please let us know via a [GitHub issue](https://github.com/PrefectHQ/prefect/issues), [Prefect Discourse](https://discourse.prefect.io/) discussion groups, or the [Prefect Community Slack](https://www.prefect.io/slack/).
 
 ## Linux installation notes
@@ -156,7 +156,7 @@ Known compatible releases include:
 
 - Ubuntu 20.04 LTS
 
-You can also: 
+You can also:
 
 - Use [Prefect Cloud](/ui/cloud/) as your API server and orchestration engine.
 - Use the `conda` virtual environment manager, which enables configuring a compatible SQLite version.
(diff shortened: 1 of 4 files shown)


zangell44 (approved):

src/prefect/client/orion.py
```diff
@@ -161,6 +160,18 @@ def __init__(
...
+        # See https://www.python-httpx.org/advanced/#custom-transports
+        # If we're using an HTTP/S client (not the ephemeral client), adjust the
+        # transport to add retries _after_ it is instantiated. If we alter the transport
+        # before instantiation, the transport will not be aware of proxies unless we
+        # reproduce all of the logic to make it so.
+        if isinstance(api, str):
+            orion_transport = self._client._transport_for_url(httpx.URL(api))
+            if isinstance(orion_transport, httpx.AsyncHTTPTransport):
+                if isinstance(orion_transport._pool, httpcore.AsyncConnectionPool):
+                    orion_transport._pool._retries = 3
```
madkinsz, line 173: Since we were using `setdefault` before this no longer lets a user pass in a transport without our retry count. I don't think that's a big deal, but I'm also not excited that we're using all these private attributes. Is there no better way to do this?

setup.cfg
```diff
@@ -39,8 +39,10 @@ filterwarnings =
     ignore:`DeploymentSpec` has been replaced by `Deployment`:DeprecationWarning
     # This warning is raised on Windows by Python internals
     ignore:the imp module is deprecated:DeprecationWarning
-    # Dockerpy is behind on this one 
+    # Dockerpy is behind on this one
     ignore:distutils Version classes are deprecated:DeprecationWarning
+    # distutils is deprecated, but we are using it directly in prefect/filesystems.py
```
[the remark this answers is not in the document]
chrisguidry:
```
$ pytest tests/client/test_orion_client.py
ImportError while loading conftest '/Users/chris/src/github.com/PrefectHQ/prefect/tests/conftest.py'.
tests/conftest.py:36: in <module>
    import prefect
src/prefect/__init__.py:24: in <module>
    from prefect.states import State
src/prefect/states.py:21: in <module>
    from prefect.results import BaseResult, R, ResultFactory
src/prefect/results.py:12: in <module>
    from prefect.filesystems import LocalFileSystem, ReadableFileSystem, WritableFileSystem
src/prefect/filesystems.py:9: in <module>
    from distutils.dir_util import copy_tree
../../../../.pyenv/versions/3.10.5/lib/python3.10/distutils/__init__.py:19: in <module>
    warnings.warn(_DEPRECATION_MESSAGE,
E   DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

This is on `main`

src/prefect/client/orion.py
chrisguidry: Ahh that's a great catch, I can skip this if the users provides a transport themselves.

As far as alternatives, it's looked pretty rough to me (and seems a lot worse than poking into the transport's pool to set the retries).  I believe we'd have to reproduce what's happening here:

https://github.com/encode/httpx/blob/master/httpx/_client.py#L213-L231
https://github.com/encode/httpx/blob/master/httpx/_client.py#L689-L702

I feel like with the `isinstance` checks and the unit tests, we're protecting ourselves pretty well against changes in the library.

I welcome alternatives, though!  We'll need to have solid support for proxies in many enterprise, government, and academic settings, so it's well worth it.

setup.cfg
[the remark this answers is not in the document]
madkinsz: 👍 no need to address here.

@peytonrunyan I believe you used this for compatibility on old Python versions or something?

tests/client/test_orion_client.py
```diff
@@ -56,6 +59,107 @@ def test_get_client_cache_uses_profile_settings(self):
...
+        assert isinstance(httpx_client, httpx.AsyncClient)
+
+        transport_for_orion = httpx_client._transport_for_url(remote_https_orion)
+        assert isinstance(transport_for_orion, httpx.AsyncHTTPTransport)
+
+        pool = transport_for_orion._pool
+        assert isinstance(pool, httpcore.AsyncConnectionPool)
+        assert pool._retries == 3  # set in prefect.client.orion.get_client()
+
+    def test_users_can_still_provide_transport(self, remote_https_orion: httpx.URL):
```
chrisguidry, line 90: @madkinsz Here's the new test case for allowing the user to provide a `transport`, great catch!

src/prefect/client/orion.py
madkinsz: Hm. The httpx proxy logic is complicated because its for a client that can contact many hosts. Our client is only being used for one host though and we know what it is when we start the client. Have you looked at https://github.com/PrefectHQ/prefect/issues/7204?

How are we configuring proxies in Cloud for this? `HTTPS_PROXY`?

> I feel like with the isinstance checks and the unit tests, we're protecting ourselves pretty well against changes in the library.

I'm more worried about compatibility breaking in released versions of Prefect, e.g. we release this then httpx changes the name of an attribute in a minor version a few months later. Users of the old version of Prefect will have to manually pin their httpx lower or they won't be able to use any server.

> We'll need to have solid support for proxies in many enterprise, government, and academic settings, so it's well worth it.

I strongly agree :)

tests/client/test_orion_client.py
madkinsz: Thanks!

src/prefect/client/orion.py
madkinsz: After a synchronous discussion, we'll add some more guards here but this looks like the best path forward. We'll also be reaching out to the encode/httpx team to see if we can contribute upstream to avoid this pattern.

docs/getting-started/installation.md
```diff
@@ -253,6 +253,16 @@ pip3 install prefect
 ```
 </div>
 
+## Using Prefect in an environment with HTTP proxies
```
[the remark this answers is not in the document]
tpdorsey: I am okay having this in installation until we have a better place for it.

tpdorsey (approved):

chrisguidry:
I see the `Test UI` integration tests are failing, but that appears to be unrelated to my changes (here's another failure of the same kind from a dependabot bump: https://github.com/PrefectHQ/prefect/actions/runs/3316571118/jobs/5478514617)

setup.cfg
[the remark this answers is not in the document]
peytonrunyan: Yup. `distutils` has a 3.7 compatible ways of doing the equivalent of `shutil.copytree(..., dirs_exist_ok=True)`. `dirs_exist_ok` was introduced in 3.8. We can get around it in a bit of a hacky way with shutil while maintaining 3.7 compatibility (see https://stackoverflow.com/a/22331852 ) if we want to get rid of `distutils`.

PR #7339, opened 2022-10-26
Fix pre-commit checks for `orion.py`

madkinsz:


src/prefect/client/orion.py
```diff
@@ -162,7 +162,7 @@ def __init__(
         )
 
         # See https://www.python-httpx.org/advanced/#custom-transports
-        # 
+        #
```
madkinsz, line 165: @chrisguidry :P

chrisguidry: The _one_ thing I quickly tweaked in the GH UI hahaha

chrisguidry (approved):
Mea culpa

PR #7343, opened 2022-10-26
subflows can `wait_for` upstream dependencies

### LINKED ISSUE #5704: Add `wait_for` for flows
## User story
As a Prefect 2 user, I want to use a flow as a subflow, with a dependency on an upstream flow or a task, even though they do not do not exchange data. I want to do this with the `wait_for` keyword argument, just as I can for tasks.

**digangidomenico**: Hi all,
I am experimenting with tasks and subflows in prefect 2.0. While in tasks I can use the wait_for keyword to enforce task dependency, the same keyword does not work for subflows. I did not found much on the subflows docs, but it seems that in prefect 1.0 dependency between subflows was possible. Is such a feature available also in prefect 2 ?
**michael054**: This just hasn’t been added yet
**michael054**: However, since a subflow will block until it is complete, the `wait_for` kwarg is less useful
**michael054**: You might as well just wait for all the tasks to finish using `.wait()` before calling the subflow
**digangidomenico**: Thanks for the quick reply <@U01CEUST9B5>.
If I understand correctly, using wait_for in a task also injects a "logical dependency" not only a temporal ordering. For example, in the following

```
@task
def task_a():
    pass

@task
def task_b():
    pass

@flow
def my_flow():
    a = task_a()
    b = task_b(wait_for=[a])
```
if task_a  fails, task_b does not execute, right?
I am looking for a similar dependency between subflows or a subflow depending on a task
**michael054**: Ah yes, I see your point. You can pass the task futures to the subflow directly but when you do not want to add a data dependency this introduces a problem
**michael054**: You can add a `wait_for` argument to your subflow that takes a list of upstream tasks and just not use the data if you don’t need it. We’ll likely support this as a first-class feature in the future. I’ll open an issue for tracking.
**digangidomenico**: I think this will work. Many thanks for your help <@U01CEUST9B5>
**michael054**: <@ULVA73B9P> open “Add `wait_for` for flows”

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1650646059038279?thread_ts=1650646059.038279&cid=CL09KU1K7).

## References
[Frequently Asked Questions - Enforcing ordering](https://docs.prefect.io/faq/?h=wait_for#how-do-i-enforce-ordering-between-tasks-that-dont-share-data)

anticorrelator:
closes #5704

Adds a `wait_for` keyword to flows. If the flow is called with this keyword and is running as a subflow, it will resolve these dependencies first before attempting to run the subflow. Otherwise the keyword will be ignored (should we throw an error instead?)

These dependencies are also attached to the virtual task that represents the subflow.

### Checklist


anna-geller (approved):
Backend ✅ works great

But front end is broken, Radar is completely empty in this case, not sure why this broke Radar entirely, but no subflows are shown

Example parent flow:

```python
"""
https://discourse.prefect.io/t/how-can-i-structure-my-flows-both-in-the-repository-and-the-ui-and-set-project-based-permissions-rbac/706
"""
from prefect.task_runners import ConcurrentTaskRunner


@task
def crm():
    pass


@task
def erp():
    pass


@task
def google_analytics():
    pass


@task
def shop_system():
    pass


@task
def shipment_api():
    pass


@flow(task_runner=ConcurrentTaskRunner())
def staging_area():
    crm()
    erp()
    google_analytics()
    shop_system()


@flow
def business_vault():
    pass


@flow
def data_mart():
    pass


@flow
def data_services():
    pass


@flow(description="Fail immediately if any stage/subflow fails")
def refresh():
    stage = staging_area()
    bv = business_vault(wait_for=[stage])
    mart = data_mart(wait_for=[bv])
    data_services(wait_for=[mart])


if __name__ == "__main__":
    refresh()
```

Radar:
image
image
image

src/prefect/engine.py
```diff
@@ -131,6 +134,7 @@ def enter_flow_run_engine_from_flow_call(
         create_and_begin_subflow_run if is_subflow_run else create_then_begin_flow_run,
         flow=flow,
         parameters=parameters,
+        wait_for=wait_for if is_subflow_run else None,
```
madkinsz, line 137: Is this just a safety guard?

madkinsz (approved):
This looks good to me!

src/prefect/engine.py
madkinsz: I'm not sure we actually require it to be a subflow, it'd be weird to pass things and have them be ignored. Maybe we should throw an exception a few lines up if it's not a subflow run but it is not none?

madkinsz: I have a preference for an exception instead of silently ignoring it. I'm not really sure what users would be passing in, but I think it'll be clearer to discourage it with an error.

jmg-duarte:
This is a breaking change.
Source: I've had my project break due to this addition.

Seems to me that this change could've been made to be non-breaking if the argument was added last with `None` as default.
Since the argument is positional, moving forward, another breaking change is required to make is "non-breaking".

You can argue that my project shouldn't be dependent on the engine's module interface but if that was to be the case, the module should be made private or some warning should be put there. 

Or maybe I am the only one using this """internal""" API (?)

anna-geller:
@jmg-duarte could you explain a bit more which part of the code is breaking for you (link to the code) and provide a minimal example of how you were using this before and what error it gives you now? ideally, **could you open that as a separate GitHub issue** that will be linked to this PR? This would help us to prioritize a potential fix, thanks so much!

PR #7352, opened 2022-10-27
UI: removed date-fns

stackoverfloweth:
[This PR](https://github.com/PrefectHQ/orion-design/pull/730) introduces an overload to all date-fns functions that make them aware of users timezone preference. Once that is merged, this PR removes `date-fns` in favor of these orion-design methods

### DIFF
--- a/orion-ui/package-lock.json
+++ b/orion-ui/package-lock.json
@@ -25,7 +25,6 @@
         "@types/d3": "^7.4.0",
         "@vitejs/plugin-vue": "^2.3.3",
         "autoprefixer": "10.4.12",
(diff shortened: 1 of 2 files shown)


pleek91:
Do we even need to wait for that PR? This project never imports from date-fns. Should be safe to remove? I don't think date-fns is marked as a peer in any of the dependant projects.

stackoverfloweth:
> Do we even need to wait for that PR? This project never imports from date-fns. Should be safe to remove? I don't think date-fns is marked as a peer in any of the dependant projects.

I don't have any reason to wait 👍

pleek91 (approved):

PR #7361, opened 2022-10-28
Agent: Add limit to control number of concurrent flow runs

eudyptula:
Added a --limit option on 'prefect agent start' to limit the number of flow runs an agent will start simultaneously.

### Example

`prefect agent start -l 4`
`prefect agent start --limit 4`

### Checklist

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -2,7 +2,6 @@
 The agent is responsible for checking for flow runs that are ready to run and starting
 their execution.
 """
-import platform
 from typing import Iterator, List, Optional, Set, Union
 from uuid import UUID
 
@@ -188,7 +187,9 @@ class OrionAgent:
                     flow_run,
                 )
 
-        return list(filter(lambda run: run.id in self.submitting_flow_run_ids, submittable_runs))
+        return list(
+            filter(lambda run: run.id in self.submitting_flow_run_ids, submittable_runs)
+        )
 
     async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
         deployment = await self.client.read_deployment(flow_run.deployment_id)
@@ -360,7 +361,9 @@ class OrionAgent:
     async def start(self):
         self.started = True
         self.task_group = anyio.create_task_group()
-        self.limiter = anyio.CapacityLimiter(self.limit) if self.limit is not None else None
+        self.limiter = (
+            anyio.CapacityLimiter(self.limit) if self.limit is not None else None
+        )
         self.client = get_client()
         await self.client.__aenter__()
         await self.task_group.__aenter__()
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -184,7 +184,6 @@ class OrionAgent:
                 self.task_group.start_soon(
                     self.submit_run,
                     flow_run,
-                    self.limiter,
                 )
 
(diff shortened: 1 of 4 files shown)


andreas-ntonas:
Hi! Any news on the state of this PR? Having a limit on the agent level is a really nice feature that my projects could benefit as well

madkinsz:
@eudyptula thanks for the pull request! This looks reasonable to me. We may want to clarify that this is a limit on _running_ flows as we could also introduce a separate limit on concurrent _submissions_. We'll need tests before this can be accepted.

eudyptula:
@madkinsz The agent should run no more than N flow runs concurrently with this new parameter. It will not take more flow runs from the work queue, than it can handle. Meaning that the remaining (scheduled) flow runs will stay in the queue for any other agent to pick up. At least if I made the update as I intended :D seems to work for us though.

Came out of this question on slack, btw.:
https://prefect-community.slack.com/archives/CL09KU1K7/p1666864461860479

madkinsz:
Yep it looks like that's what it does :)

We could also release capacity limit slots after `task_status.started()` is called which would make the limit N flow runs submitting concurrently. Submission concurrency could be useful for some users as well and we'll want to make sure it's clear that the `--limit` here applies to the full life cycle of the flow run infrastructure so when/if we add another type of limit the interface is still intuitive. I'm thinking ahead to a name for the submitting limit e.g. `--submitting-limit` and wondering if we should do something similar here e.g. `--running-limit`, but I think `--limit` is actually more accurate in this case.

eudyptula:
@madkinsz Not sure I completely understand what submission limit is actually limiting...

madkinsz:
@eudyptula Well the process of creating infrastructure can be rate limited by cloud providers so you may want to have that concurrency limited. Anyway, I wouldn't worry about that.

If you add tests here, we can accept this :)

eudyptula:
@madkinsz: I doubt it, but just realised that we may have to clarify if this update could somehow cause containers to not be shut down on crashes (in regard to my issue https://github.com/PrefectHQ/prefect/issues/7560)

MuFaheemkhan:
@madkinsz is  "limit on concurrent submissions" in pipeline? This feature would be really useful.

madkinsz:
@MuFaheemkhan it should be trivial to add following this pattern if it works well.

andreas-ntonas:
I want to add that I have been using this PR directly for a couple of weeks and I can confirm it works the way it is supposed to! :) 
Also I noticed that https://github.com/PrefectHQ/prefect/pull/7362 could supplement this feature really well, like adding a tag that helps us identify the agent where the flow was executed.

madkinsz (approved):
@eudyptula thanks for your patience!

eudyptula:
@andreas-ntonas, Happy to hear you're finding it useful :)

@madkinsz, Just glad to see it making it through :) Looking forward to ditch my fork and use the official releases again, and when the network issues are resolved, hopefully we can finally complete our migration to Prefect 2.

madkinsz:
@eudyptula while the real issue is buried deep in some http core libraries, #7593 should alleviate all of the issues for ya'll :)

rsampaths16:
Hey @madkinsz is this released for public use? If not, what is the timeline for its release? This seems like an important feature. Thanks :)

PR #7372, opened 2022-10-28
Fix GitHub storage error for Windows

peytonrunyan:
A user encountered an error with the GitHub block: https://prefect-community.slack.com/archives/C03CY667GGM/p1666901179640649

Running `my_github_block.get_directory()` on a Windows machine results in an error `'g' is not recognized as an internal or external command, operable program or batch file.` . This is because we pass a string with the git pull command to `run_process`, which in turn calls `open_process`, inside of which we check `if sys.platform == "win32": command = " ".join(command)`.

This pull request changes the command to be a list of command components instead of a string. After replicating the error on a Windows VM, I made the code changes and tested running `get_directory()`, which behaved as desired. 

I updated the tests to reflect that we expect a list instead of a string.

### Checklist

### DIFF
--- a/src/prefect/filesystems.py
+++ b/src/prefect/filesystems.py
@@ -894,16 +894,18 @@ class GitHub(ReadableDeploymentStorage):
             local_path: A local path to clone to; defaults to present working directory.
         """
         # CONSTRUCT COMMAND
-        cmd = f"git clone {self._create_repo_url()}"
+        cmd = ["git", "clone", self._create_repo_url()]
         if self.reference:
-            cmd += f" -b {self.reference}"
+            cmd.append("-b")
+            cmd.append(self.reference)
 
         # Limit git history
-        cmd += " --depth 1"
+        cmd.append("--depth")
+        cmd.append("1")
 
         # Clone to a temporary directory and move the subdirectory over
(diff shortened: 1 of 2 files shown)


src/prefect/filesystems.py
```diff
@@ -894,16 +894,18 @@ async def get_directory(
...
+        cmd = ["git", "clone", self._create_repo_url()]
         if self.reference:
-            cmd += f" -b {self.reference}"
+            cmd.append("-b")
+            cmd.append(self.reference)
 
         # Limit git history
-        cmd += " --depth 1"
+        cmd.append("--depth")
+        cmd.append("1")
```
bunchesofdonald, line 904: It _could_ be more readable to use concatenation here:

```suggestion
        cmd += ['--depth', '1']
```

bunchesofdonald (approved):
Looks great!

src/prefect/filesystems.py
ahuang11: Another way is to use `shlex.split` when finished joining the string.

peytonrunyan: Both of those are cleaner solutions - updated!

PR #7236, opened 2022-10-18
Fix the UI to the deployment when running prefect deployment apply

khuyentran1401:
This PR fixes the UI to the deployment when running prefect deployment apply

### Example
When running `prefect deployment apply`, the UI to the deployment is incorrect:
```
Successfully loaded 'test flow'
Deployment 'test-flow/test flow' successfully created with id 'dc008392-680f-45b9-9ce6-42c79efd93bb'.
View Deployment in UI: 
https://app.prefect.cloud/account/804d0209-2a32-420b-8f8b-6ce3eb1d0211/workspace/ac926a68-ea8d-4d20-b18
6-4c7beac6e8a3/deployment/dc008392-680f-45b9-9ce6-42c79efd93bb
```

This PR makes sure the UI is in the correct form. This means that the UI above will become `https://app.prefect.cloud/account/804d0209-2a32-420b-8f8b-6ce3eb1d0211/workspace/ac926a68-ea8d-4d20-b18
6-4c7beac6e8a3/deployments/deployment/dc008392-680f-45b9-9ce6-42c79efd93bb`.

closes #7208 

### Checklist

- [x] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

PR #7376, opened 2022-10-28
Fix deployment UI link

anticorrelator:
Fixes a broken link

### Checklist


abrookins:
Hmmm -- there is something about this that I don't understand. The new URL works on Cloud, but doesn't work in the OSS UI. I.e., locally,`http://localhost:4200/deployments/deployment/b63d7e2b-260e-43df-a8f1-f59781220987` is a 404, but `http://localhost:4200/deployment/b63d7e2b-260e-43df-a8f1-f59781220987` works. We may need some advice from @zhen0.

madkinsz:
Yeah I'm super confused about the divergence of URLs for Cloud/Orion — ideally we would not have to implement client logic to split between the two. It'd be great if we had an overview of the stable UI URLs we are supposed to use somewhere.

khuyentran1401:
I created a PR #7236 that is related to this and encountered the same issue

anticorrelator:
@abrookins @madkinsz @khuyentran1401 There is active work by @zhen0 to unify routes. I can branch this behavior to work for both cloud and self-hosted due to the route divergence but it'd likely require another fix to undo later. If we unify the routes by changing, won't that be a breaking API change?

zhen0:
Pr to handle both routes is here: https://github.com/PrefectHQ/prefect/pull/7391

As @anticorrelator notes - this is only for UI links.  We discussed adding a redirect but since all routes are being updated here I think it's easier to update all together.

anticorrelator:
@abrookins with https://github.com/PrefectHQ/prefect/pull/7391 this change should be now consistent across both cloud and the OSS UI, mind if I get a re-review?

abrookins (approved):
LGTM! 👍 Thanks for jumping on it!

PR #7394, opened 2022-10-31
Allow filtering of Work Queues via the API based on name prefix

### LINKED ISSUE #7205: Add a route to filter work queues by name
### First check

### Prefect Version

2.x

`/work_queues/filter` only supports `limit` and `offset` 

We have two use cases that require filtering by work queue name
1. UI work queue page
2. Agents polling for runs by work queue names matching a given pattern as discussed here https://github.com/PrefectHQ/prefect/pull/7099#issuecomment-1273690600 

- Add a route to filter for work queues by name 
- Implement it in a way that queries in a LIKE-style to allow filtering for work queues that start with a given string
- Add necessary index, e.g. trigram index on `work_queue.name` for performance

### Example Use

_No response_

### Additional context

_No response_

bunchesofdonald:
Adds support to the `work_queues/filter` route in the Orion API to filter Work Queues based on the name. The filter can perform `startswith_` queries that enable finding work queues with a matching prefix, as well as the standard `all_` query.

Closes #7205 

### Checklist

### DIFF
--- a/src/prefect/client/orion.py
+++ b/src/prefect/client/orion.py
@@ -828,13 +828,13 @@ class OrionClient:
 
     async def match_work_queues(
         self,
-        prefix: str,
+        prefixes: List[str],
     ) -> List[schemas.core.WorkQueue]:
         """
         Query Orion for work queues with names with a specific prefix.
 
         Args:
-            prefix: a string used to match work queue name prefixes
+            prefixes: a list of strings used to match work queue name prefixes
 
         Returns:
             a list of [WorkQueue model][prefect.orion.schemas.core.WorkQueue] representations
@@ -849,7 +849,7 @@ class OrionClient:
                 offset=current_page * page_length,
                 limit=page_length,
                 work_queue_filter=schemas.filters.WorkQueueFilter(
-                    name=schemas.filters.WorkQueueFilterName(startswith_=prefix)
+                    name=schemas.filters.WorkQueueFilterName(startswith_=prefixes)
                 ),
             )
             if not new_queues:
--- a/src/prefect/client/orion.py
+++ b/src/prefect/client/orion.py
@@ -800,6 +800,7 @@ class OrionClient:
         self,
         limit: int = None,
         offset: int = 0,
+        work_queue_filter: schemas.filters.WorkQueueFilter = None,
     ) -> List[schemas.core.WorkQueue]:
         """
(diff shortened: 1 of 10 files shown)


anticorrelator (commented):
thanks for building this @bunchesofdonald! should we consider updating the orion client method [here](https://github.com/PrefectHQ/prefect/blob/filter-work-queue-name/src/prefect/client/orion.py#L822-L853) to use this route? I think we'd reduce the load on the database for agents using this matching feature by a fair bit

bunchesofdonald:
That makes sense to me @anticorrelator

src/prefect/orion/schemas/filters.py
```diff
@@ -1231,3 +1231,48 @@ def _get_filter_list(self, db: "OrionDBInterface") -> List:
...
+        description="A list of work queue names to include",
+        example=["wq-1", "wq-2"],
+    )
+
+    startswith_: Optional[str] = Field(
+        default=None,
+        description=(
+            "A case-insensitive starts-with match. For example, "
+            " passing 'marvin' will match "
+            "'marvin', and 'Marvin-robot', but not 'sad-marvin'."
```
anna-geller, line 1250: sad-marvin ❤️ those Easter eggs!

anna-geller (approved):
looks great

bunchesofdonald:
@anticorrelator That method is updated to use the new filter as well.

src/prefect/client/orion.py
```diff
@@ -815,19 +817,24 @@ async def read_work_queues(
...
+                if work_queue_filter
+                else None
+            ),
         }
         response = await self._client.post(f"/work_queues/filter", json=body)
         return pydantic.parse_obj_as(List[schemas.core.WorkQueue], response.json())
 
     async def match_work_queues(
         self,
-        prefixes: List[str],
```
anticorrelator, line 824: The feature that relies on this agent method may pass in a list of prefixes to match, could we keep that?

anticorrelator (changes requested):

src/prefect/client/orion.py
bunchesofdonald: It'll take some rework, but yeah. The tests are all setup to just send through a string (no list) so I _assumed_ the type here was just wrong.

anticorrelator (approved):

PR #7398, opened 2022-11-01
Enhancement: Return API setting for the late runs service

znicholasbrown:
This PR adds the `PREFECT_ORION_SERVICES_LATE_RUNS_AFTER_SECONDS` setting to the payload returned by the `/ui-settings` route; this allows the UI to operate on that setting when determining lateness thresholds to show or not.

### Example
```json
{
  "api_url": "...",
  "late_runs_threshold_seconds": 60,
}
```

### Checklist


src/prefect/orion/api/server.py
```diff
@@ -225,6 +225,7 @@ def create_ui_app(ephemeral: bool) -> FastAPI:
     def ui_settings():
         return {
             "api_url": prefect.settings.PREFECT_ORION_UI_API_URL.value(),
+            "late_runs_threshold_seconds": prefect.settings.PREFECT_ORION_SERVICES_LATE_RUNS_AFTER_SECONDS.value().total_seconds(),
```
madkinsz, line 228: It seems like we should stay as close to the original name as possible. 

Separately, how exactly is the UI going to use this? We may want to introduce a second setting that defaults to the value of this setting, like the `PREFECT_ORION_UI_API_URL` so we can have consistent "UI" settings.

znicholasbrown: The UI will use this to decide cutoffs for showing degrees of "lateness", which is a little TBD but will likely take the form of showing/hiding lateness indicators at some cutoff and then emphasizing lateness more or less after the cutoff.

I don't really understand what a separate setting would give us in this situation?

madkinsz: Well, right now UI settings are prefixed by `PREFECT_ORION_UI_` and this breaks from that pattern. Here, you're accessing the setting for the late run service. This is not some global concept of lateness, but rather the value that a specific service uses to change state names.

znicholasbrown: Updated the name

madkinsz: So the UI is going to use this setting to compute several different levels of lateness to display?

znicholasbrown: More to figure out how pressing lateness should be, e.g. if a run is late by 1 second it's probably not important but if a run is late by 5 minutes it might be more so

madkinsz: It seems like it's a separate setting then? The existing setting is an absolute cutoff at which the state name is changed by a service. This is for a different purpose, i.e. highlighting late runs in the UI. I don't think this particular setting is a huge deal, but I think in general we'll want the settings for the UI to be specific. I don't think it should reach into API service settings.

madkinsz: For example, if we removed the service the setting `PREFECT_ORION_SERVICES_LATE_RUNS_AFTER_SECONDS` would not have meaning anymore. It'd be unexpected that it's consumed by the UI. If there was a `PREFECT_LATE_RUN_AFTER_SECONDS=...` setting I'd be more okay with reusing it here, but it's pretty explicitly for the service as-is.

PR #7400, opened 2022-11-01
docs: fix typo in concepts/task-runner.md document

marwan116:
This is a very simple merge request that resolves a typo in the concepts/task-runner docs which confused me when I was trying to follow the docs.

### Example

The code example shown used to be:
```python

@task
def my_task(name):
    return "I'm a task!"

@flow
def my_flow():
    future = my_task()
    result = future.result(raise_on_failure=False)
    if future.get_state().is_failed():
        # `result` is an exception! handle accordingly
        ...
    else:
        # `result` is the expected return value of our task
        ...
```
What confused me is the call to `my_task`:
- `my_task.__call__` expects a `name` argument but we don't pass it anything 
- invoking `my_task.__call__` should not return a future but instead should execute the underlying python function. 

### Checklist

### DIFF
--- a/docs/concepts/task-runners.md
+++ b/docs/concepts/task-runners.md
@@ -258,12 +258,13 @@ The `.result()` method will wait for the task to complete before returning the r
 from prefect import flow, task
 
 @task
-def my_task(name):
+def my_task():
     return "I'm a task!"
+
 
 @flow
 def my_flow():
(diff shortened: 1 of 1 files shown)


tpdorsey (approved):
Good catch! Thank you for the contribution.

PR #7334, opened 2022-10-26
Add login with a browser to `prefect cloud login`

madkinsz:
Merges into #7332. Requires a UI `/authorize` route. Authored with @znicholasbrown.

Adds login via the web browser to `prefect cloud login`, allowing you to login without creating an API key first.

This is a result of the UI workshop during TDP in which we were attempting to improve the user experience during onboarding. The user is prompted to create a workspace, then prompted to use the workspace from the CLI before they have logged in with an API key. Instead of adding more steps to the UI onboarding experience, we can improve the user experience by making it very easy to authenticate the CLI.

This works by spinning up a little API server from the CLI then opening the user's web browser to a dedicated UI page. When the user authorizes the login, the browser will send a new API key to the CLI's API server. The CLI then uses the key to login as normal.

We preserve the ability to pass a key via CLI argument or pasted into a prompt.

This also includes:
- Check if the user is logged in and prompt y/N to reauthenticate
- Check if the user is logged in via another profile and prompt a profile switch
- Login can be performed without any interactive elements
- `CloudClient.read_workspaces` parses workspaces into a schema

### Example

Screen Shot 2022-11-03 at 1 04 31 PM

### Checklist

- [x] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

PR #7173, opened 2022-10-14
Flow Run Notification URL Bugfix

padbk:
Our Prefect instance runs on `.ai` TLD. In the notifications the TLD was being stripped as the characters are also in `/api` and therefore the URL in Slack was broken. This change should keep the TLD, and so the URL should work when linked in Slack.

No issue created as this is only a small bugfix.

Thank you,

### Example
Before:
ourdomain./flow-runs/flow-run/...
After:
ourdomain.ai/flow-runs/flow-run/...

### Checklist
- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

PR #7411, opened 2022-11-02
Create setting for UI URL and clarify Cloud API URL

madkinsz:
Changes `PREFECT_CLOUD_URL` to `PREFECT_CLOUD_API_URL` and adds `PREFECT_UI_URL` and `PREFECT_CLOUD_UI_URL`  to consolidate inference of UI URLs. 

`PREFECT_CLOUD_URL` is marked as deprecated and has backwards compatibility support.
`PREFECT_UI_URL` infers its value from `PREFECT_API_URL` by default, but can be set explicitly. For example, if you want to test against a UI preview on netlify.
`PREFECT_CLOUD_UI_URL` infers its value from `PREFECT_CLOUD_API_URL` by default, but can be set explicitly.

When using Cloud, the `PREFECT_UI_URL` is workspace scoped. The `PREFECT_CLOUD_UI_URL` is not.
When using the open source server, `PREFECT_CLOUD_UI_URL` will still point to Cloud. This is important for login e.g. #7334 

Required for #7334 
Helps with issues like https://github.com/PrefectHQ/prefect/pull/7173
### Example

```
❯ PREFECT_CLOUD_API_URL="bar" python -c "from prefect.settings import PREFECT_CLOUD_API_URL; print(PREFECT_CLOUD_API_URL.value())"
bar
```

```
❯ PREFECT_CLOUD_URL="foo" PREFECT_CLOUD_API_URL="bar" python -c "from prefect.settings import PREFECT_CLOUD_API_URL; print(PREFECT_CLOUD_API_URL.value())"
* Deprecation warning
foo
```

```
❯ PREFECT_CLOUD_API_URL="bar" python -c "from prefect.settings import PREFECT_CLOUD_URL; print(PREFECT_CLOUD_URL.value())" 
* Deprecation warning
bar
```
### Checklist

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -93,7 +93,7 @@ class Setting(Generic[T]):
 
         self.__doc__ = self.field.description
 
-    def value(self) -> T:
+    def value(self, bypass_callback: bool = False) -> T:
         """
         Get the current value of a setting.
 
@@ -103,9 +103,9 @@ class Setting(Generic[T]):
         PREFECT_API_URL.value()
         ```
         """
-        return self.value_from(get_current_settings())
+        return self.value_from(get_current_settings(), bypass_callback=bypass_callback)
 
-    def value_from(self, settings: "Settings") -> T:
+    def value_from(self, settings: "Settings", bypass_callback: bool = False) -> T:
         """
         Get the value of a setting from a settings object
 
@@ -115,7 +115,7 @@ class Setting(Generic[T]):
         PREFECT_API_URL.value_from(get_default_settings())
         ```
         """
-        return settings.value_of(self)
+        return settings.value_of(self, bypass_callback=bypass_callback)
 
     def __repr__(self) -> str:
         return f"<{self.name}: {self.type.__name__}>"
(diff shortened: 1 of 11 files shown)


madkinsz:
See also https://github.com/PrefectHQ/prefect/pull/7334/commits/1a2c92f7058a31ce9a6466389a3c3d0854766149 — which adds `PREFECT_CLOUD_UI_URL` for cases where the URL should not include the workspace

src/prefect/settings.py
```diff
@@ -234,6 +234,43 @@ def warn_on_database_password_value_without_usage(values):
...
+    ui_url = api_url = PREFECT_API_URL.value_from(settings)
+
+    if not api_url:
+        return "http://ephemeral-orion"
+
+    if api_url.endswith("/api"):
+        ui_url = ui_url[:-4]
+
+    if api_url.startswith("https://api."):
+        ui_url = ui_url.replace("https://api.", "https://app.")
```
peytonrunyan, line 269: Pure curiosity - is this a common issue?

madkinsz: Hm idk about issue, this is just how we transform our API URl to a UI URL. I'm a little hesitant about having it be this broad honestly, maybe it should only replace `api.prefect.cloud` with `app.prefect.cloud`.

anticorrelator (commented):
hey michael, will these changes require an update to the files [here](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/cli/orion_utils.py#L1)

madkinsz:
Ah thanks for the note! Yeah we should probably look into changing those. Also looks like there are some additional transformations for Cloud that I'm missing :'( 

https://github.com/PrefectHQ/prefect/blob/main/src/prefect/cli/orion_utils.py#L76-L77

src/prefect/settings.py
peytonrunyan: Is there any downside to restricting it?

madkinsz: Nope! I think we should. Otherwise we may incorrectly coerce someone's custom domain.

madkinsz:
I've cherry-picked https://github.com/PrefectHQ/prefect/commit/1a2c92f7058a31ce9a6466389a3c3d0854766149 and updated handling to match the code linked by @anticorrelator

anticorrelator (approved):
lgtm!

PR #7430, opened 2022-11-03
Add `is_schedule_active` to client `Deployment` class

EmilRex:
This PR exposes the `is_schedule_active` option on the `Deployment` class. Currently, the default behavior is to enable the schedule, but users may want to apply their deployments and not schedule runs.

### Example
```python
from my_project.flows import my_flow
from prefect.orion.schemas.schedules import CronSchedule

    flow=my_flow,
    name="example",
    schedule=CronSchedule(cron="0 0 * * *"),
    is_schedule_active=False
)
deployment.apply()
```

### Checklist


src/prefect/deployments.py
```diff
@@ -362,6 +364,9 @@ def _yaml_dict(self) -> dict:
         description="One of more tags to apply to this deployment.",
     )
     schedule: schemas.schedules.SCHEDULE_TYPES = None
+    is_schedule_active: Optional[bool] = Field(
+        default=True, description="Whether or not the schedule is active."
```
madkinsz, line 368: I think we may want to default to `None` to respect the server-side value during updates.

I'm tempted to suggest another name for this as well, `schedule_active` seems nicer for this user-facing toggle whereas `is_schedule_active` makes more sense on the API object itself.

madkinsz: cc @billpalombi

EmilRex: >I think we may want to default to None to respect the server-side value during updates.
Yep, good point!

madkinsz (approved):

madkinsz:
@EmilRex looks like this requires additional changes

madkinsz (changes requested):
We'll also want test coverage for the behavior of `None`, `True`, and `False`.

madkinsz:
@EmilRex are you going to be able to finish this? Do you want help from someone on the OSS team?

src/prefect/deployments.py
```diff
@@ -505,6 +510,10 @@ async def load(self) -> bool:
         Raises:
             - ValueError: if both name and flow name are not set
         """
+        client_field_to_api_field = {
+            "schedule_active": "is_schedule_active",
+        }
+
```
peytonrunyan, line 516: If somebody else has a preferred way of handling this, I'd love to hear it.

peytonrunyan: That seems reasonable

peytonrunyan:
@madkinsz any other thoughts?

src/prefect/orion/schemas/core.py
```diff
@@ -559,6 +559,16 @@ def validate_name_characters(cls, v):
...
 
+    @validator("is_schedule_active", pre=True)
+    def validate_is_schedule_active(cls, v, field):
+        """
+        Primary responsibility of this validator is to make sure that actions that use this field
+        are able to safely accept a `None` value.
+        """
+        if v is None:
+            return field.default
+        return v
```
madkinsz, line 570: I'm pretty confused by this validator, it seems like a no-op?

tests/client/test_orion_client.py
```diff
@@ -1470,3 +1470,52 @@ async def test_ephemeral_app_check(orion_client):
...
+    deployment = await orion_client.read_deployment(deployment_id)
+    assert deployment.is_schedule_active == is_schedule_active
+    schedule = IntervalSchedule(interval=timedelta(days=1))
+    await orion_client.update_deployment(deployment, schedule=schedule)
+    deployment = await orion_client.read_deployment(deployment_id)
+    assert deployment.is_schedule_active == is_schedule_active
+
+
+@pytest.mark.parametrize(
+    "initial, initial_expected, updated",
```
madkinsz, line 1498: Maybe `on_create`, `after_create`, `on_update`, `after_update` for clarity?

tests/client/test_orion_client.py
```diff
@@ -1470,3 +1470,52 @@ async def test_ephemeral_app_check(orion_client):
 async def test_ephemeral_app_check_when_using_hosted_orion(hosted_orion_api):
     async with OrionClient(hosted_orion_api) as orion_client:
         assert (await orion_client.using_ephemeral_app()) is False
+
+
+@pytest.mark.parametrize(
+    "schedule_active, is_schedule_active", [(True, True), (False, False), (None, True)]
+)
+async def test_update_deployment_schedule_active_does_not_overwrite_when_not_provided(
```
madkinsz, line 1478: Similarly, could these be `on_create`, `after_update`?

tests/test_deployments.py
```diff
@@ -526,14 +572,38 @@ async def test_deployment(patch_import, tmp_path):
...
+        self, patch_import, tmp_path, orion_client
+    ):
+        d = Deployment(name="TEST", flow_name="fn")
+        deployment_id = await d.apply(work_queue_concurrency=424242)
+        queue_name = d.work_queue_name
+        work_queue = await orion_client.read_work_queue_by_name(queue_name)
+        assert work_queue.concurrency_limit == 424242
+
+    @pytest.mark.parametrize(
+        "is_active, expected_schedule_active",
```
madkinsz, line 586: Similarly, `on_update` and `after_update` or something like `provided`, `expected`?

tests/test_deployments.py
```diff
@@ -526,14 +572,38 @@ async def test_deployment(patch_import, tmp_path):
...
+            flow_function,
+            name="foo",
+            tags=["A", "B"],
+            description="foobar",
+            version="12",
+            is_schedule_active=is_active,
+        )
+        assert d.flow_name == flow_function.name
+        assert d.name == "foo"
+        assert d.is_schedule_active == is_active
```
madkinsz, line 602: Nit: we should probably use `is` for comparisons with this field

src/prefect/orion/schemas/core.py
peytonrunyan: `core.Deployment.is_schedule_active` is a bool. 

`client.create_deployment` has `is_schedule_active` as a `bool = None`. The goal here is to have it using using the server default if None is provided. 

Inside of there we do `DeploymentCreate(..., is_schedule_active=is_schedule_active)` which, if we pass `None` without the validator, results in 

```
E   pydantic.error_wrappers.ValidationError: 1 validation error for DeploymentCreate
E   is_schedule_active
E     none is not an allowed value (type=type_error.none.not_allowed)
```

tests/test_deployments.py
peytonrunyan: Confirming my understanding - this is because we're using a bool here?

src/prefect/orion/schemas/core.py
peytonrunyan: @madkinsz I think I addressed your concerns here: https://github.com/PrefectHQ/prefect/pull/7430/files/bc7b7d78d2df68fbfada5d0f93b18d8dff7c4501..7b3b2d4ecb4504cec6f530661b6413fc360ff165#diff-99d27244d265f50d916c514ca3d3afdb591ab196fb390cea38f3b55c2e11847c

tests/test_deployments.py
madkinsz: Yep!

madkinsz (approved):

PR #7433, opened 2022-11-04
Adjust the scheduler to create the smallest number of runs 

jlowin:
Draft PR that I'm exploring with @zangell44 

Currently, the scheduler attempts to create `N` runs for every deployment. `N` was previously 100 and Zach recently lowered it to 20. This has a couple negative effects:
- the runs view is "crowded" because of the volume of upcoming, scheduled runs
- the scheduler is constantly trying to insert runs to the database (runs are rejected on insert via idempotency keys)

This PR tries a new approach. Instead of always trying to create 20 runs, the scheduler now tries to create the fewest possible runs that satisfy a relatively heuristic:
- at least 3 but no more than 100 runs
- scheduled at least one hour in the future but not more than 100 days

The result is that for "typical" deployments having intervals of of > 20 minutes, only the next three runs will be scheduled at any time (I chose three because it seems like a nice but inoffensive look into the future at any moment). This should alleviate significant write pressure on the database at scale.

For "fast" deployments, more runs will be scheduled to guarantee visibility at least one hour out (e.g. if the interval is 5 minutes, 12 or 13 runs will be scheduled [depending on timing]).

However, under any circumstances, more than 100 runs won't be scheduled (for super fast deployments) and runs won't be scheduled more than 100 days in the future (for super slow deployments).

### Future enhancements
The scheduler performs this logic for every deployment no matter whether its runs were just scheduled. For typical deployments, this means that the same runs are being created over and over until one of them finally executes and permits a new run to be scheduled. We could evaluate skipping deployments that were recently scheduled or have more than 3 auto-scheduled runs already, depending on performance implications.

### DIFF
--- a/docs/concepts/schedules.md
+++ b/docs/concepts/schedules.md
@@ -190,7 +190,14 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
 
 The `Scheduler` service is started automatically when `prefect orion start` is run and it is a built-in service of Prefect Cloud. 
 
-By default, the `Scheduler` service visits deployments on a 60-second loop and attempts to create up to 100 scheduled flow runs up to 100 days in the future. These defaults can be shown with the terminal command `prefect config view --show-defaults`:
+By default, the `Scheduler` service visits deployments on a 60-second loop, though recently-modified deployments will be visited more frequently. The `Scheduler` evaluates each deployment's schedule and creates new runs appropriately. For typical deployments, it will create the next three runs, though more runs will be scheduled if the next 3 would all start in the next hour. 
+
+More specifically, the `Scheduler` tries to create the smallest number of runs that satisfy the following constraints, in order:
+- no more than 100 runs will be scheduled
+- runs will not be scheduled more than 100 days in the future
+- at least 3 runs will be scheduled
+- runs will be scheduled until at least one hour in the future
+These behaviors can all be adjusted through the relevant settings that can be viewed with the terminal command `prefect config view --show-defaults`:
 
 <div class='terminal'>
 ```bash
@@ -198,7 +205,9 @@ PREFECT_ORION_SERVICES_SCHEDULER_DEPLOYMENT_BATCH_SIZE='100'
 PREFECT_ORION_SERVICES_SCHEDULER_ENABLED='True' 
 PREFECT_ORION_SERVICES_SCHEDULER_INSERT_BATCH_SIZE='500' 
 PREFECT_ORION_SERVICES_SCHEDULER_LOOP_SECONDS='60.0' 
+PREFECT_ORION_SERVICES_SCHEDULER_MIN_RUNS='3' 
 PREFECT_ORION_SERVICES_SCHEDULER_MAX_RUNS='100' 
+PREFECT_ORION_SERVICES_SCHEDULER_MIN_SCHEDULED_TIME='1:00:00' 
 PREFECT_ORION_SERVICES_SCHEDULER_MAX_SCHEDULED_TIME='100 days, 0:00:00' 
 ```
 </div>
--- a/src/prefect/orion/api/deployments.py
+++ b/src/prefect/orion/api/deployments.py
@@ -201,18 +201,33 @@ async def schedule_deployment(
     deployment_id: UUID = Path(..., description="The deployment id", alias="id"),
     start_time: DateTimeTZ = Body(None, description="The earliest date to schedule"),
     end_time: DateTimeTZ = Body(None, description="The latest date to schedule"),
+    min_time: DateTimeTZ = Body(
+        None, description="Runs will be scheduled until at least this date"
+    ),
+    min_runs: int = Body(None, description="The minimum number of runs to schedule"),
     max_runs: int = Body(None, description="The maximum number of runs to schedule"),
     db: OrionDBInterface = Depends(provide_database_interface),
 ) -> None:
     """
     Schedule runs for a deployment. For backfills, provide start/end times in the past.
+
+    This function will generate the minimum number of runs that satisfy the min
+    and max times, and the min and max counts. Specifically, the following order
+    will be respected:
+        - runs will be generated starting on or after the `start_time`
+        - no more than `max_runs` runs will be generated
+        - no runs will be generated after `end_time` is reached
+        - at least `min_runs` runs will be generated
+        - runs will be generated until at least `min_time` is reached
     """
     async with db.session_context(begin_transaction=True) as session:
         await models.deployments.schedule_runs(
             session=session,
             deployment_id=deployment_id,
             start_time=start_time,
+            min_time=min_time,
             end_time=end_time,
+            min_runs=min_runs,
             max_runs=max_runs,
         )
 
--- a/src/prefect/orion/models/deployments.py
+++ b/src/prefect/orion/models/deployments.py
@@ -19,6 +19,8 @@ from prefect.orion.utilities.database import json_contains
 from prefect.settings import (
     PREFECT_ORION_SERVICES_SCHEDULER_MAX_RUNS,
     PREFECT_ORION_SERVICES_SCHEDULER_MAX_SCHEDULED_TIME,
+    PREFECT_ORION_SERVICES_SCHEDULER_MIN_RUNS,
+    PREFECT_ORION_SERVICES_SCHEDULER_MIN_SCHEDULED_TIME,
 )
 
 
@@ -367,6 +369,8 @@ async def schedule_runs(
     deployment_id: UUID,
     start_time: datetime.datetime = None,
     end_time: datetime.datetime = None,
+    min_time: datetime.datetime = None,
+    min_runs: int = None,
     max_runs: int = None,
     auto_scheduled: bool = True,
 ) -> List[UUID]:
@@ -377,28 +381,48 @@ async def schedule_runs(
         session: a database session
         deployment_id: the id of the deployment to schedule
         start_time: the time from which to start scheduling runs
-        end_time: a limit on how far in the future runs will be scheduled
+        end_time: runs will be scheduled until at most this time
+        min_time: runs will be scheduled until at least this time
+        min_runs: a minimum amount of runs to schedule
         max_runs: a maximum amount of runs to schedule
+
+    This function will generate the minimum number of runs that satisfy the min
+    and max times, and the min and max counts. Specifically, the following order
+    will be respected:
+        - runs will be generated starting on or after the `start_time`
+        - no more than `max_runs` runs will be generated
+        - no runs will be generated after `end_time` is reached
+        - at least `min_runs` runs will be generated
+        - runs will be generated until at least `min_time` is reached
 
     Returns:
         a list of flow run ids scheduled for the deployment
     """
+    if min_runs is None:
+        min_runs = PREFECT_ORION_SERVICES_SCHEDULER_MIN_RUNS.value()
     if max_runs is None:
         max_runs = PREFECT_ORION_SERVICES_SCHEDULER_MAX_RUNS.value()
     if start_time is None:
         start_time = pendulum.now("UTC")
-    start_time = pendulum.instance(start_time)
     if end_time is None:
         end_time = start_time + (
             PREFECT_ORION_SERVICES_SCHEDULER_MAX_SCHEDULED_TIME.value()
         )
+    if min_time is None:
+        min_time = start_time + (
+            PREFECT_ORION_SERVICES_SCHEDULER_MIN_SCHEDULED_TIME.value()
+        )
+    start_time = pendulum.instance(start_time)
     end_time = pendulum.instance(end_time)
+    min_time = pendulum.instance(min_time)
 
     runs = await _generate_scheduled_flow_runs(
         session=session,
         deployment_id=deployment_id,
         start_time=start_time,
         end_time=end_time,
+        min_time=min_time,
+        min_runs=min_runs,
         max_runs=max_runs,
         auto_scheduled=auto_scheduled,
     )
@@ -411,6 +435,8 @@ async def _generate_scheduled_flow_runs(
     deployment_id: UUID,
     start_time: datetime.datetime,
     end_time: datetime.datetime,
+    min_time: datetime.datetime,
+    min_runs: int,
     max_runs: int,
     db: OrionDBInterface,
     auto_scheduled: bool = True,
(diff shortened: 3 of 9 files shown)


src/prefect/orion/schemas/schedules.py
```diff
@@ -242,7 +263,32 @@ async def get_dates(
...
+                schedule's timezone.
+
+        Returns:
+            List[pendulum.DateTime]: a list of dates
+        """
+        return sorted(
+            [i async for i in self._get_dates_generator(n=n, start=start, end=end)]
+        )
+
+    async def _get_dates_generator(
```
madkinsz, line 285: It looks like this doesn't need to be async anymore, probably shouldn't use an async generator unless we need `await`.

madkinsz:
This is a great idea! Looking forward to it, the implementation looks good so far too.

jlowin:
@madkinsz thanks for the suggestion! I've removed the async generators. I've left the `get_dates()` function async because it would technically be a breaking change to public schedules API otherwise. Note this has one minor behavior change (that I don't think is an issue but if it is we can revert): when the generator is async, control is yielded back to the control loop after each date is generated (because of the `yield` in this PR's previous version and an explicit `await sleep` in the current codebase); now the entire date-generation loop runs before control is yielded. This has minor potential to constitute a blocking loop but I think that paired with this PR's early-stopping logic it's probably not a problem.

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
     deployment_id: UUID = Path(..., description="The deployment id", alias="id"),
     start_time: DateTimeTZ = Body(None, description="The earliest date to schedule"),
     end_time: DateTimeTZ = Body(None, description="The latest date to schedule"),
+    min_time: DateTimeTZ = Body(
```
zangell44, line 204: should this be a `timedelta` instead?

zangell44 (approved):
Nice work!

docs/concepts/schedules.md
```diff
@@ -190,15 +190,24 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
 
 The `Scheduler` service is started automatically when `prefect orion start` is run and it is a built-in service of Prefect Cloud. 
 
-By default, the `Scheduler` service visits deployments on a 60-second loop and attempts to create up to 100 scheduled flow runs up to 100 days in the future. These defaults can be shown with the terminal command `prefect config view --show-defaults`:
+By default, the `Scheduler` service visits deployments on a 60-second loop, though recently-modified deployments will be visited more frequently. The `Scheduler` evaluates each deployment's schedule and creates new runs appropriately. For typical deployments, it will create the next three runs, though more runs will be scheduled if the next 3 would all start in the next hour. 
+
+More specifically, the `Scheduler` tries to create the smallest number of runs that satisfy the following constraints, in order:
+- no more than 100 runs will be scheduled
```
tpdorsey, line 196:
```suggestion

- No more than 100 runs will be scheduled.
```

docs/concepts/schedules.md
```diff
@@ -190,15 +190,24 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
...
+- runs will not be scheduled more than 100 days in the future
```
tpdorsey, line 197:
```suggestion
- Runs will not be scheduled more than 100 days in the future.
```

docs/concepts/schedules.md
```diff
@@ -190,15 +190,24 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
...
+- at least 3 runs will be scheduled
```
tpdorsey, line 198:
```suggestion
- At least 3 runs will be scheduled.
```

docs/concepts/schedules.md
```diff
@@ -190,15 +190,24 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
...
+- runs will be scheduled until at least one hour in the future
```
tpdorsey, line 199:
```suggestion
- Runs will be scheduled until at least one hour in the future.
```

docs/concepts/schedules.md
```diff
@@ -190,15 +190,24 @@ See the Python [RRuleSchedule class docs](/api-ref/orion/schemas/schedules/#pref
...
+These behaviors can all be adjusted through the relevant settings that can be viewed with the terminal command `prefect config view --show-defaults`:
```
tpdorsey, line 200:
```suggestion

These behaviors can all be adjusted through the relevant settings that can be viewed with the terminal command `prefect config view --show-defaults`:
```

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
...
     max_runs: int = Body(None, description="The maximum number of runs to schedule"),
     db: OrionDBInterface = Depends(provide_database_interface),
 ) -> None:
     """
     Schedule runs for a deployment. For backfills, provide start/end times in the past.
+
+    This function will generate the minimum number of runs that satisfy the min
+    and max times, and the min and max counts. Specifically, the following order
+    will be respected:
+        - runs will be generated starting on or after the `start_time`
```
tpdorsey, line 217:
```suggestion

        - Runs will be generated starting on or after the `start_time`
```

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
...
+        - no more than `max_runs` runs will be generated
```
tpdorsey, line 218:
```suggestion
        - No more than `max_runs` runs will be generated
```

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
...
+        - no runs will be generated after `end_time` is reached
```
tpdorsey, line 219:
```suggestion
        - No runs will be generated after `end_time` is reached
```

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
...
+        - at least `min_runs` runs will be generated
```
tpdorsey, line 220:
```suggestion
        - At least `min_runs` runs will be generated
```

src/prefect/orion/api/deployments.py
```diff
@@ -201,18 +201,33 @@ async def schedule_deployment(
...
+        - runs will be generated until at least `min_time` is reached
```
tpdorsey, line 221:
```suggestion
        - Runs will be generated until at least `min_time` is reached
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
         start_time: the time from which to start scheduling runs
-        end_time: a limit on how far in the future runs will be scheduled
+        end_time: runs will be scheduled until at most this time
+        min_time: runs will be scheduled until at least this time
+        min_runs: a minimum amount of runs to schedule
         max_runs: a maximum amount of runs to schedule
 
+    This function will generate the minimum number of runs that satisfy the min
+    and max times, and the min and max counts. Specifically, the following order
+    will be respected:
```
tpdorsey, line 391:
```suggestion
    will be respected:
   
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
+        - runs will be generated starting on or after the `start_time`
```
tpdorsey, line 392:
```suggestion
        - Runs will be generated starting on or after the `start_time`
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
+        - no more than `max_runs` runs will be generated
```
tpdorsey, line 393:
```suggestion
        - No more than `max_runs` runs will be generated
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
+        - no runs will be generated after `end_time` is reached
```
tpdorsey, line 394:
```suggestion
        - No runs will be generated after `end_time` is reached
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
+        - at least `min_runs` runs will be generated
```
tpdorsey, line 395:
```suggestion
        - At least `min_runs` runs will be generated
```

src/prefect/orion/models/deployments.py
```diff
@@ -377,28 +381,48 @@ async def schedule_runs(
...
+        - runs will be generated until at least `min_time` is reached
```
tpdorsey, line 396:
```suggestion
        - Runs will be generated until at least `min_time` is reached
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
         start_time: the time from which to start scheduling runs
-        end_time: a limit on how far in the future runs will be scheduled
+        end_time: runs will be scheduled until at most this time
+        min_time: runs will be scheduled until at least this time
+        min_runs: a minimum amount of runs to schedule
         max_runs: a maximum amount of runs to schedule
 
+    This function will generate the minimum number of runs that satisfy the min
+    and max times, and the min and max counts. Specifically, the following order
+    will be respected:
```
tpdorsey, line 464:
```suggestion
    will be respected:
    
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
+        - runs will be generated starting on or after the `start_time`
```
tpdorsey, line 465:
```suggestion
        - Runs will be generated starting on or after the `start_time`
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
+        - no more than `max_runs` runs will be generated
```
tpdorsey, line 466:
```suggestion
        - No more than `max_runs` runs will be generated
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
+        - no runs will be generated after `end_time` is reached
```
tpdorsey, line 467:
```suggestion
        - No runs will be generated after `end_time` is reached
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
+        - at least `min_runs` runs will be generated
```
tpdorsey, line 468:
```suggestion
        - At least `min_runs` runs will be generated
```

src/prefect/orion/models/deployments.py
```diff
@@ -428,9 +454,20 @@ async def _generate_scheduled_flow_runs(
...
+        - runs will be generated until at least `min_time` is reached
```
tpdorsey, line 469:
```suggestion
        - Runs will be generated until at least `min_time` is reached
```

src/prefect/orion/schemas/schedules.py
```diff
@@ -472,7 +511,30 @@ async def get_dates(
...
+            n (int): The number of dates to generate
+            start (datetime.datetime, optional): The first returned date will be on or
+                after this date. Defaults to None.  If a timezone-naive datetime is
+                provided, it is assumed to be in the schedule's timezone.
+            end (datetime.datetime, optional): The maximum scheduled date to return. If
+                a timezone-naive datetime is provided, it is assumed to be in the
+                schedule's timezone.
+
+        Returns:
+            List[pendulum.DateTime]: a list of dates
```
tpdorsey, line 527:
```suggestion
            List[pendulum.DateTime]: A list of dates
```

src/prefect/orion/schemas/schedules.py
```diff
@@ -242,7 +261,30 @@ async def get_dates(
...
```
tpdorsey, line 277:
```suggestion
            List[pendulum.DateTime]: A list of dates
```

src/prefect/orion/schemas/schedules.py
```diff
@@ -108,7 +107,30 @@ async def get_dates(
...
```
tpdorsey, line 123:
```suggestion
            List[pendulum.DateTime]: A list of dates
```

src/prefect/orion/services/scheduler.py
```diff
@@ -183,12 +193,34 @@ async def _generate_scheduled_flow_runs(
...
+            deployment_id: the id of the deployment to schedule
+            start_time: the time from which to start scheduling runs
+            end_time: runs will be scheduled until at most this time
+            min_time: runs will be scheduled until at least this time
+            min_runs: a minimum amount of runs to schedule
+            max_runs: a maximum amount of runs to schedule
+
+        This function will generate the minimum number of runs that satisfy the min
+        and max times, and the min and max counts. Specifically, the following order
+        will be respected:
```
tpdorsey, line 209:
```suggestion
        will be respected:
        
```

src/prefect/orion/services/scheduler.py
```diff
@@ -183,12 +193,34 @@ async def _generate_scheduled_flow_runs(
...
+            - runs will be generated starting on or after the `start_time`
```
tpdorsey, line 210:
```suggestion
            - Runs will be generated starting on or after the `start_time`
```

src/prefect/orion/services/scheduler.py
```diff
@@ -183,12 +193,34 @@ async def _generate_scheduled_flow_runs(
...
+            - no more than `max_runs` runs will be generated
```
tpdorsey, line 211:
```suggestion
            - No more than `max_runs` runs will be generated
```

tpdorsey (approved):
Looks good. Minor suggested edits.

src/prefect/orion/api/deployments.py
jlowin: Yes, I like that approach

PR #7439, opened 2022-11-04
Redirect to the logs tab when the flowRunId changes 

zhen0:
Uses the new v-model on the tabs component to redirect to the logs tab on the flow runs page when the flow run ID changes.  This is mainly for sub-flows where a user clicks on a link to a subflow and the flow run info updates but they are left on an empty and not useful subflow tab. 

### Example

 
 Current:

https://user-images.githubusercontent.com/40272060/200025325-bd4ac4d3-c9da-465b-a2e9-63db65a6e0c0.mov

Update:

https://user-images.githubusercontent.com/40272060/200025441-fedae1a2-e8f3-4fa1-9d42-17bdaf5c75d8.mov

### Checklist


pleek91:
@zhen0 this totally works and can be merged as is if you'd like. But something we could and probably should do is add the tab to the url and v-model the tab using a `useRouteParam`. Which we can do by adding another route child after the radar route like
```typescript
{
  name: 'flow-run',
  path: ':flowRunTab',
  component: (): RouteComponent => import('@/pages/FlowRun.vue'),
}
```

zhen0:
Can we add that as a follow up @pleek91?  Happy to make a ticket for it. This was already an add on to my original parent id work and I don't want to go too off topic.

pleek91 (approved):

Issue #6666, opened 2022-09-01
Flow won't execute with Python 3.7


Tostino:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I am having (file permissions?) failures when trying to run on my deployed infrastructure rather than a local docker container for the exact same flows mentioned in: https://github.com/PrefectHQ/prefect/issues/6665
The source files exist in the specified deploy location, they work in my local docker container, they don't work on this VM.

### Reproduction

```python
Use Debian 10, try to run a local storage and a process, with the code deployed where specified in the deployment.
```

### Error

```
Sep  1 01:41:12 check-prefectagent01 prefect[14263]: 01:41:12.436 | INFO    | prefect.agent - Submitting flow run '49507f89-c9e5-49b5-8b28-4e522f9f4253'
Sep  1 01:41:14 check-prefectagent01 prefect[14263]: 01:41:14.575 | INFO    | prefect.infrastructure.process - Opening process 'abstract-roadrunner'...
Sep  1 01:41:14 check-prefectagent01 prefect[14263]: 01:41:14.585 | INFO    | prefect.agent - Completed submission of flow run '49507f89-c9e5-49b5-8b28-4e522f9f4253'
Sep  1 01:41:16 check-prefectagent01 prefect[14263]: 01:41:16.336 | ERROR   | Flow run 'abstract-roadrunner' - Flow could not be retrieved from deployment.
Sep  1 01:41:16 check-prefectagent01 prefect[14263]: Traceback (most recent call last):
Sep  1 01:41:16 check-prefectagent01 prefect[14263]:   File "/usr/local/lib/python3.7/dist-packages/prefect/engine.py", line 256, in retrieve_flow_then_begin_flow_run
Sep  1 01:41:16 check-prefectagent01 prefect[14263]:     flow = await load_flow_from_flow_run(flow_run, client=client)
Sep  1 01:41:16 check-prefectagent01 prefect[14263]:   File "/usr/local/lib/python3.7/dist-packages/prefect/client.py", line 103, in with_injected_client
Sep  1 01:41:16 check-prefectagent01 prefect[14263]:     return await fn(*args, **kwargs)
[...]

Issue #7307, opened 2022-10-24
LocalFileSystem does not upload files in subfolders when used as Deployment storage block


hateyouinfinity:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

LocalFileSystem does not upload files in subfolders when used as Deployment storage block.
The problem lies in `LocalFileSystem._get_ignore_func` ([ref](https://github.com/PrefectHQ/prefect/blob/51e92dda965431c5f1beb486f0bcc4058035a868/src/prefect/filesystems.py#L148)).
`_get_ignore_func` calculates which files are to be included based on patterns from `.prefectignore` and packs the list of files to include into a function to be [used](https://github.com/PrefectHQ/prefect/blob/51e92dda965431c5f1beb486f0bcc4058035a868/src/prefect/filesystems.py#L187) by `shutil.copytree`.
Quoting `shutil.copytree`'s docstring:
> The optional ignore argument is a callable. If given, it
    is called with the `src` parameter, which is the directory
    being visited by copytree(), and `names` which is the list of
    `src` contents, as returned by os.listdir():

        callable(src, names) -> ignored_names
However, `_get_ignore_func.ignore_func` [ignores](https://github.com/PrefectHQ/prefect/blob/51e92dda965431c5f1beb486f0bcc4058035a868/src/prefect/filesystems.py#L154) what `shutil.copytree` refers to as `src`, seemingly assuming that each filename is a full path.
Basically, `_get_ignore_func.ignore_func` expects calls like `(directory="dirname", files=["dirname/filename"])`, but is instead called with `(directory="dirname", files=["filename"])`.

### Reproduction

```shell
mkdir testproj
cd testproj
echo "" > .prefectignore 
touch LICENSE
mkdir myflow
echo "from prefect import flow

@flow
def flow_func():
    return 2+2" > myflow/myflow.py

echo "from prefect.deployments import Deployment
from prefect.filesystems import LocalFileSystem

from myflow.myflow import flow_func

fs = LocalFileSystem(basepath='/tmp/testfolder')
fs.save('test-fs', overwrite=True)

depl = Deployment.build_from_flow(flow=flow_func, name='myflow_deploy', storage=fs)
depl.apply()" > deploy.py
python deploy.py

ls /tmp/testfolder/*
[...]

PR #7441, opened 2022-11-04
Fix error with LocalFileSystem deployments in python 3.7

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/6666 and https://github.com/PrefectHQ/prefect/issues/7307

Python did not add support for writing to existing directories using `shutil.copy_tree` until python 3.8. This results in errors when creating deployments using `LocalFileSystem` storage when using python 3.7.

To fix this, I used the code from https://stackoverflow.com/a/22331852 to replicate the behavior in a manner that is compatible with python 3.7. This implementation is located in `src/prefect/utilities/compat.py` and is used if the python version is 3.7.

In the process of fixing this, I found that our ignore function was written in such a way that it kept us from uploading subdirectory contents. That's fixed now too.

I've updated the tests to check the actual behavior of the functions and methods in temporary directories. 

### Example

```python
copytree(src=local_path, dst=to_path, ignore=ignore_func, dirs_exist_ok=True)
```
### Checklist


src/prefect/utilities/filesystem.py
```diff
@@ -117,3 +120,45 @@ def relative_path_to_current_platform(path_str: str) -> Path:
...
+            _copytree_37(source_path, destination_path, original_src, symlinks, ignore)
+        else:
+            shutil.copy2(source_path, destination_path)
+
+
+def prefect_copytree(src, dst, ignore=None):
+    if sys.version_info < (3, 8):
+        _copytree_37(src=src, dst=dst, original_src=src, ignore=ignore)
+    else:
+        shutil.copytree(src=src, dst=dst, ignore=ignore, dirs_exist_ok=True)
```
madkinsz, line 164: I think `prefect.utilities.compat.copytree` makes more sense to me for all of this. I'd do something like

if if sys.version_info < (3, 8):
   def copytree(...):
      ..
else:
   from shutil import copytree
```

No reason to define this function if not running on 3.7.

**Comment by madkinsz** [comment]
@peytonrunyan this looks like it got some unrelated changes in it, let me know if you want any help getting it sorted out.

File: src/prefect/filesystems.py [hunk 2]
```diff
@@ -161,24 +161,22 @@ async def put_directory(
...
 
         if ignore_file:
             ignore_func = await self._get_ignore_func(ignore_file)
         else:
             ignore_func = None
 
-        copytree(src=local_path, dst=to_path, ignore=ignore_func, dirs_exist_ok=True)
+        if local_path == to_path:
+            pass
+        else:
```
madkinsz, line 176: Probably can be a != :)

peytonrunyan: Ah yeah - that is from the commit to revert all of the changes that weren't strictly part of fixing this bug I think. It's not a new addition by me. It'll get cleaned up again in a follow up PR that addresses the LocalFileSystem bugs.

src/prefect/filesystems.py
```diff
@@ -138,18 +136,14 @@ async def get_directory(
...
 
-    async def _get_ignore_func(self, local_path: str, ignore_file: str):
+    async def _get_ignore_func(self, ignore_file: str):
         with open(ignore_file, "r") as f:
             ignore_patterns = f.readlines()
 
-        included_files = filter_files(local_path, ignore_patterns)
-
         def ignore_func(directory, files):
+            included_files = filter_files(directory, ignore_patterns)
```
james-tisato, line 146: I wondered if a better solution to this bit would be to keep the call to `filter_files` in `_get_ignore_func` and then modify `ignore_func` as below:
```
def ignore_func(directory, files):
    relative_path = Path(directory).relative_to(local_path)
    return_val = [f for f in files if str(relative_path / f) not in included_files]
    return return_val
```

The list of included files would be generated once, relative to the root directory where the `.prefectignore` file lives, and then `ignore_func` checks each file against that list of relative paths (rather than just checking the filename with no path like it does in the original code).

This solution allows you to write patterns in your `.prefectignore` that take advantage of the directory structure beyond the current directory that `ignore_func` is running for, e.g.
```
/some/dir/relative/to/ignore/file
**/subdir/*
```

peytonrunyan: I like this! Just made the update and added a test.

madkinsz:
@peytonrunyan fwiw I generally wouldn’t bother testing anything in the `_compat` module. We just expect them to be perfect backports and they should have good coverage from usage in our code where we're calling them.

tests/test_filesystems.py
```diff
@@ -52,6 +81,129 @@ async def test_get_directory_duplicate_directory(self, tmp_path):
...
+
+    async def test_dir_contents_copied_correctly_with_put_directory_and_directory_pattern(
+        self,
+    ):
+        """Make sure that ignore file behaves properly."""
+
+        sub_dir_name = "puppy"
+        skip_sub_dir = "kitty"
+
+        with TemporaryDirectory() as tmp_src:
```
madkinsz, line 174: In the future, I'd use the `tmp_path` fixture and just make two directories in it for a source and destination — no need for all this nesting

madkinsz: Then your `setup_test_directory` could just be a fixture that uses `tmp_path`.

madkinsz (approved):
LGTM!

tests/test_filesystems.py
peytonrunyan: Awesome - I'll refactor this once I'm back from break.

PR #7450, opened 2022-11-06
Improve scheduler selectivity for performance

jlowin:
Follow up to #7433 (and builds on it).

If #7433 is merged, the scheduler will begin scheduling the _smallest_ number of runs that satisfy certain minimum criteria (a minimum number of runs and a minimum duration to schedule into the future). This has the major effect of reducing the number of runs the scheduler attempts to insert, which is important because the scheduler runs on a relatively tight loop and relies on idempotency to allow it to eagerly attempt to create redundant runs.

However, the fact that the scheduler is constantly trying to create runs that already exist is, itself, a larger source of potential performance problems. Even if every single deployment had just been scheduled and nothing had changed, on its next loop the scheduler will revisit every single deployment and attempt to recreate all of its runs. 

This PR makes a small change to the query that loads deployments to schedule. Instead of just looking for ones with active schedules, this PR adds additional criteria to only get ones with fewer than `min_runs` runs, or no runs outside `min_scheduled_time`. In other words: only deployments whose future work would actually be affected by scheduling. A new index is introduced to make this efficient.

### Please note
The performance gains of modifying the scheduler selectivity in this PR may outweigh those gained by adopting the "least number of runs" in #7433. The logic in this PR can be implemented independently of #7433 with a minor modification (namely, removing the `min_scheduled_time` constraint and changing `min_runs` back to `max_runs`). This could be done if the approach in #7433 is deemed to be confusing or overly complicated. The result would be that the scheduler would maintain `20` scheduled runs for every deployment, but only query those with fewer than `20` runs at any time - old scheduling behavior, new way of figuring out what to schedule.

### DIFF
--- a/src/prefect/orion/services/scheduler.py
+++ b/src/prefect/orion/services/scheduler.py
@@ -120,6 +120,7 @@ class Scheduler(LoopService):
         Returns a sqlalchemy query for selecting deployments to schedule.
 
         The query gets the IDs of any deployments with:
+        
             - an active schedule
             - EITHER:
                 - fewer than `min_runs` auto-scheduled runs
--- a/src/prefect/orion/services/scheduler.py
+++ b/src/prefect/orion/services/scheduler.py
@@ -138,7 +138,7 @@ class Scheduler(LoopService):
                 db.FlowRun,
                 # join on matching deployments, only picking up future scheduled runs
                 sa.and_(
-                    db.Deployment.id == db.FlowRun.id,
+                    db.Deployment.id == db.FlowRun.deployment_id,
                     db.FlowRun.state_type == StateType.SCHEDULED,
                     db.FlowRun.next_scheduled_start_time >= now,
                     db.FlowRun.auto_scheduled.is_(True),
--- a/src/prefect/orion/services/scheduler.py
+++ b/src/prefect/orion/services/scheduler.py
@@ -120,7 +120,7 @@ class Scheduler(LoopService):
         Returns a sqlalchemy query for selecting deployments to schedule.
 
         The query gets the IDs of any deployments with:
-        
+
             - an active schedule
             - EITHER:
                 - fewer than `min_runs` auto-scheduled runs
--- a/src/prefect/orion/services/scheduler.py
+++ b/src/prefect/orion/services/scheduler.py
@@ -12,6 +12,7 @@ import sqlalchemy as sa
 import prefect.orion.models as models
 from prefect.orion.database.dependencies import inject_db
 from prefect.orion.database.interface import OrionDBInterface
+from prefect.orion.schemas.states import StateType
 from prefect.orion.services.loop_service import LoopService, run_multiple_services
 from prefect.settings import (
     PREFECT_ORION_SERVICES_SCHEDULER_DEPLOYMENT_BATCH_SIZE,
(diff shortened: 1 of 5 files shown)


src/prefect/orion/services/scheduler.py
```diff
@@ -116,14 +117,47 @@ async def run_once(self, db: OrionDBInterface):
...
+            # TODO: on Postgres, this could be replaced with a lateral join that
+            # sorts by `next_scheduled_start_time desc` and limits by
+            # `self.min_runs` for a ~ 50% speedup. At the time of writing,
+            # performance of this universal query appears to be fast enough that
+            # this optimization is not worth maintaining db-specific queries
+            .join(
+                db.FlowRun,
+                # join on matching deployments, only picking up future scheduled runs
+                sa.and_(
+                    db.Deployment.id == db.FlowRun.id,
```
zangell44, line 141:
```suggestion
                    db.Deployment.id == db.FlowRun.deployment_id,
```

madkinsz:
> However, the fact that the scheduler is constantly trying to create runs that already exist is, itself, a larger source of potential performance problems. Even if every single deployment had just been scheduled and nothing had changed, on its next loop the scheduler will revisit every single deployment and attempt to recreate all of its runs.

I _thought_ our scheduler might be doing this but I wasn't sure. Definitely a weird behavior before and a great optimization to avoid it! Given the current consensus that https://github.com/PrefectHQ/prefect/pull/7433 is sensible I think we can merge both. We can get an additional opinion from @bunchesofdonald.

src/prefect/orion/services/scheduler.py
```diff
@@ -116,14 +117,47 @@ async def run_once(self, db: OrionDBInterface):
...
+                    db.FlowRun.state_type == StateType.SCHEDULED,
```
madkinsz, line 142: What happens if I move a scheduled run to a "CANCELLED" state. Will this schedule it again on the next scheduler iteration?

bunchesofdonald (approved):
Both PRs seem sensible to me. I do agree that this PR is likely to have a much greater performance advantage upfront, but I also see value in adding fewer rows to the database in general even if that doesn't result in a huge gain.

jlowin:
@zangell44  

> One alternative implementation would be maintaining some schedule_at column on deployments that indicates the next time a deployment would need to be scheduled. Each time a deployment is scheduled / updated, we could update that column with the next time the deployment should be revisited.

I have a weak preference not to introduce this type of state that ties the scheduler implementation to the database if we can avoid it. If it becomes the only way to get the performance we require than by all means, it's a very effective solution -- but if we can get to the same place by reusing existing "core" data, that's preferable

src/prefect/orion/services/scheduler.py
jlowin: face...palm...

src/prefect/orion/services/scheduler.py
jlowin: Good question. I think that would fall back into something similar to the current behavior where the scheduler would see, for example, that there are only two scheduled runs instead of three and seek to create the "missing" run. However the run wouldn't actually be created because of the idempotency key. 

In the scheme of things, while this would represent a temporary performance regression for that deployment, it does recover the correct behavior and I don't think we need to worry about it.

madkinsz:
@jlowin I've rebased this after merge of https://github.com/PrefectHQ/prefect/pull/7433

src/prefect/orion/services/scheduler.py
```diff
@@ -116,14 +117,47 @@ async def run_once(self, db: OrionDBInterface):
     @inject_db
     def _get_select_deployments_to_schedule_query(self, db: OrionDBInterface):
         """
-        Returns a sqlalchemy query for selecting deployments to schedule
+        Returns a sqlalchemy query for selecting deployments to schedule.
+
+        The query gets the IDs of any deployments with:
```
tpdorsey, line 122:
```suggestion
        The query gets the IDs of any deployments with:
        
```

tpdorsey (approved):
Docstrings look good. Minor formatting suggestion added. (Lists don't render correctly if not preceded by a blank line)

zangell44 (approved):

src/prefect/orion/services/scheduler.py
jlowin: @madkinsz i added a test for this situation

zangell44 (approved):

PR #7457, opened 2022-11-07
Ensure that queries are sorted correctly when limits are used

jlowin:
Tests in #7450 are failing because of a non-deterministic bug (that only shows up there because the introduction of an index led to slightly different row selectivity).

On Postgres, we use a lateral join to efficiently grab runs in each work queue that needs scheduling. We then sort the query to ensure that runs are returned in ascending order of scheduled start time. However, if the work queue has a concurrency limit or the `limit_per_queue` flag is passed, we apply a `LIMIT` in that lateral join. The limit is applied _before_ the sort, so it's possible for the lateral join to grab arbitrary runs in the future (not necessarily the next scheduled runs). The fix is to add a sort inside the lateral join. This also includes a test that maximizes the probability of observing the bug.

### DIFF
--- a/src/prefect/orion/database/query_components.py
+++ b/src/prefect/orion/database/query_components.py
@@ -239,6 +239,8 @@ class BaseQueryComponents(ABC):
                 scheduled_before_clause,
             )
             .with_for_update(skip_locked=True)
+            # priority given to runs with earlier next_scheduled_start_time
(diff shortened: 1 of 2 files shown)


chrisguidry (approved):
That is a lovely test!

jlowin:
@chrisguidry always love getting that +1 from you 😃

Issue #5941, opened 2022-06-25
Hide  password in PREFECT_ORION_DATABASE_CONNECTION_URL in Prefect 2.0 UI setting page


gq1:
PREFECT_ORION_DATABASE_CONNECTION_URL: postgresql+asyncpg://prefect:***********@tol-prefect-postgresql.prefect:5432/orion

We don't have any restriction to use UI page, and the database password shown in the setting page.

zhen0:
Thanks for the request. That info is there because we think it could be useful to users. I'm wary of removing it since that would affect others. I'm ok to leave this ticket open to see if others share your use case or have other preferences or ideas.

gq1:
The problem is we don't have a user name and password to protect the UI pages, and then everybody can see your database's password.

madkinsz:
We can update the settings routes to handle secrets the same way we do for blocks now that we've added a workaround to our Pydantic models.

austinnichols101:
> we think it could be useful to users

Yes - **_malicious users_**
I can think of no good reason why the DB password should EVER be displayed in the UI.

voycey:
Wouldn't it be simpler to simply create authentication for the Orion Server UI?

madkinsz:
Unfortunately adding authentication is vastly more complicated than obfuscating our settings. For some users, that may be the preferred path though.

voycey:
Couldn't it be as simple as:

- start docker container and pass token into env
- require the same token for all requests if it exists?

That way the implementation falls to the users and is flexible enough to be handled as they wish

jannekeskitalo:
A new setting PREFECT_ORION_DATABASE_PASSWORD was added to allow injecting the password from a secret to the connection url, but then the password setting itself is shown in cleartext on the settings -page. Would be really good to hide it. For now I need to block access to the settings -path completely.

https://github.com/PrefectHQ/prefect/pull/5759

madkinsz:
We are going to hide it, we just need to duplicate secret handling from blocks -> settings. If anyone is interested in taking that on, feel free to chime in. Otherwise it's on our backlog and we'll get to it as soon as possible.

PR #7465, opened 2022-11-07
Add secret flag to settings and obfuscate by default when displayed

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/5941

- Adds `Setting.is_secret: bool`.
- Adds `Settings.with_obfuscated_secrets()` to get a copy of the settings with secret values replaced with `********`.
- Adds obfuscation of secrets to `prefect config view`; `--show-secrets` can be used to view them.
- Adds obfuscation of secrets to `GET /admin/settings`
- Moves `prefect.orion.utilities.names` to `prefect.utilities.names` so `obfuscate_string` can be used client-side
- Adds `obfuscate` utility that handles nulls and non-string data types

Secret settings in `/admin/settings` are always obfuscated. This is sort of a breaking change although I think it is acceptable. We could add `include_secrets: bool = True` if we want to preserve compatibility. I am not sure I see value in ever returning secret settings via the API though.

### Example

```
❯ PREFECT_API_KEY="foo" 
❯ prefect config view              
PREFECT_PROFILE='test'
PREFECT_API_KEY='*********' (from env)

❯ prefect config view --show-secrets
PREFECT_PROFILE='test'
PREFECT_API_KEY='foo' (from env)
```

Screen Shot 2022-11-07 at 2 57 22 PM

Note there is no reason to override `repr` or `str` at this time:
```
❯ ipython -c "from prefect.settings import PREFECT_API_KEY; print(str(PREFECT_API_KEY))"
<PREFECT_API_KEY: str>
```

### Checklist


src/prefect/cli/root.py
```diff
@@ -70,7 +74,8 @@ def main(
         soft_wrap=not PREFECT_CLI_WRAP_LINES.value(),
     )
 
-    setup_logging()
+    if not PREFECT_TEST_MODE:
+        setup_logging()
```
madkinsz, line 78: Otherwise the `CliRunner` will attempt to setup logging multiple times when the logging config changes

madkinsz:
@ahuang11 I believe you poked at this before if you're interested in reviewing :)

src/prefect/settings.py
```diff
@@ -427,8 +433,8 @@ def warn_on_database_password_value_without_usage(values):
 PREFECT_LOGGING_COLORS = Setting(
     bool,
     default=True,
-    description="""Whether to style console logs.""",
 )
+"""Whether to style console logs with color"""
```
ahuang11, line 437:
```suggestion
"""Whether to style console logs with color."""
```

src/prefect/settings.py
```diff
@@ -787,6 +795,22 @@ def copy_with_update(
             }
         )
 
+    def with_obfuscated_secrets(self):
+        """
+        Returns a copy of this settings object with secret setting values obfsucated.
+        """
+        settings = self.copy(
+            update={
+                setting.name: "*" * 8
```
ahuang11, line 804: Maybe `"*" * 8` should be a constant since it gets used a couple times?

src/prefect/settings.py
```diff
@@ -787,6 +795,22 @@ def copy_with_update(
             }
         )
 
+    def with_obfuscated_secrets(self):
+        """
+        Returns a copy of this settings object with secret setting values obfsucated.
```
ahuang11, line 800:
```suggestion
        Returns a copy of this settings object with secret setting values obfuscated.
```

src/prefect/cli/root.py
ahuang11, line 78:
```suggestion
        # Otherwise the CliRunner will attempt to setup logging
        # multiple times when the logging config changes
        setup_logging()
```

ahuang11 (commented):
It looks great!

madkinsz:
Thanks for the review!

bunchesofdonald (approved):
I like it!

abrookins:
Looks great. No concerns from me about the change in behavior.

zangell44 (approved):

jlowin:
> I am not sure I see value in ever returning secret settings via the API though.

100% agree.

PR #7471, opened 2022-11-08
Minor doc fixes to resolve warnings and errors

tpdorsey:
Fixes to configuration and links to resolve doc build warnings and mkdocstring errors.

- Remove deprecated watch setting for mkdocstrings. Add mkdocs watch setting.
- Remove page reference from nav that no longer references a valid source file.
- Update mkdocstrings local options to use `options` syntax, removing deprecated syntax.
- Fix broken link references in API docs.
- Fix missing or malformed docstrings causing errors.

### Checklist


src/prefect/client/orion.py
```diff
@@ -1433,7 +1433,7 @@ async def read_flow_run(self, flow_run_id: UUID) -> FlowRun:
             flow_run_id: the flow run ID of interest
 
         Returns:
-            a [Flow Run model][prefect.orion.FlowRun] representation of the flow run
+            a [Flow Run model][prefect.orion.schemas.core.FlowRun] representation of the flow run
```
madkinsz, line 1436: Actually it's `prefect.client.schemas.FlowRun` now; should we even have these links? They're in the type annotation.

tpdorsey: `prefect.client.schemas.FlowRun` doesn't resolve to a valid link address for mkdocstrings

madkinsz: I think we should fix that or remove the link then. There are other cases here where we are incorrectly referencing the orion schemas as well. I'd prefer not mislead users on the return types.

madkinsz (approved):

PR #7479, opened 2022-11-09
Add deployment and flow metadata to infrastructure labels

madkinsz:
Currently, infrastructure is given labels for flow run metadata. However, in cases like https://github.com/PrefectHQ/prefect-aws/issues/148 it is useful for the infrastructure to have additional metadata like the deployment and flow names.

This adds optional `deployment` and `flow` parameters to `Infrastructure.prepare_for_flow_run` which will then be used to add the labels:
- `prefect.io/deployment-name`
- `prefect.io/deployment-updated`
- `prefect.io/flow-name`

These can be used by infrastructure blocks to configure namespaces. For example, the `ECSTask` block will use the deployment and flow names for the "family" name to avoid rate limits. Right now, we're using "prefect" for the family name.

### Example

The full list of labels for an example deployment is now:

```python
{
    "prefect.io/flow-run-id": "69da0440-1aa0-4aa0-9f5d-c2d43a9752d0",
    "prefect.io/flow-run-name": "olive-nightingale",
    "prefect.io/version": "2.6.6+53.g4c0c88af2.dirty",
    "prefect.io/deployment-name": "My Deployment",
    "prefect.io/deployment-updated": "2022-11-09T16:50:47.307155Z",
    "prefect.io/flow-name": "my-flow",
}
```

### Checklist

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -179,6 +179,7 @@ class OrionAgent:
 
     async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
         deployment = await self.client.read_deployment(flow_run.deployment_id)
+        flow = await self.client.read_flow(deployment.flow_id)
 
         # overrides only apply when configuring known infra blocks
         if not deployment.infrastructure_document_id:
@@ -213,9 +214,14 @@ class OrionAgent:
         doc_dict["data"] = infra_dict
         infra_document = BlockDocument(**doc_dict)
         infrastructure_block = Block._from_block_document(infra_document)
+
         # TODO: Here the agent may update the infrastructure with agent-level settings
 
-        prepared_infrastructure = infrastructure_block.prepare_for_flow_run(flow_run)
+        # Add flow run metadata to the infrastructure
+        prepared_infrastructure = infrastructure_block.prepare_for_flow_run(
+            flow_run, deployment=deployment, flow=flow
+        )
+
         return prepared_infrastructure
 
     async def submit_run(self, flow_run: FlowRun) -> None:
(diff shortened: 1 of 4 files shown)


peytonrunyan:
This seems really useful. I want to double-check that I understand how labels are intended to be used: their description says `Labels applied to the infrastructure for metadata purposes`. We're still technically passing metadata here, but it's essentially being used for providing runtime configuration. I don't have any example off the top of my head, but are we open to passing additional runtime configurations as labels in the future?

madkinsz:
@peytonrunyan yeah this is a little tricky. The idea here is:

1. Infrastructure should be usable fully independently of a flow run so infrastructure fields are entirely specific to running infrastructure. 
2. Infrastructure can be used for flow runs, this should be a rich experience as it is the primary way it is used in practice (right now, at least). 
3. The base infrastructure type includes labels though not infrastructure has labels so flow run metadata can be passed to all infrastructure types.
4. This metadata can be used to facilitate richer experiences for flow run infrastructure but should never be required. 

It's kind of a fine line. I intentionally did not include this previously; no decisions were made at infrastructure runtime based on labels like this. `prepare_for_flow_run` would set the correct flow run command and environment variable which is the minimum we need to execute the flow.  Now that there's a need for deployment related information, we need to find a way to get it to infrastructure.

Now that you've got me thinking about it a bit harder, we could do the following as an alternative:

1. Add a `family` field to `ECSTask` (it should have one anyway) that defaults to `None`
2. Add a custom `prepare_for_flow_run` to `ECSTask` that updates the family if not set to this value

This would preserve the idea that labels are only metadata and keep `prepare_for_flow_run` the only point at which infrastructure configuration choices are made based on flow run information.

madkinsz:
I've updated the ECS pull request to reflect the above changes, no changes needed here.

peytonrunyan (approved):

Issue #7134, opened 2022-10-12
Flow Run search should be by `start_time` not `expected_start_time`


kevin868:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

The Flow Run history page has some unexpected behavior. If your Flow Run starts at 9:10, but you adjust the filter to Start Date = '9:00' and End Date (any time after 9:10), your Flow Run is not shown in the results.

That's because this Flow Run was expected at 08:00.. (70 minutes late) so, we actually need to set the Start Date to 08:00 to make it show up.

I believe the most natural usage of the Start Date filter should be the actual `start_date`, not the `expected_start_date`
Proposal: Change the UI's param to /history to use `start_date`

### Reproduction

I verified that sending this Payload to /history
```
{"flow_runs":{"expected_start_time":{"after_":"2022-10-12T00:00:00.000Z","before_":"2022-10-12T03:59:59.000Z"}},"sort":"EXPECTED_START_TIME_DESC"}
```
produces an empty `[]` result..  and we get the expected result (flowRuns in a list) if we move the Start date filter back earlier on the UI.

### Error

There WERE Flow runs in this period. around 9:08PM

<img width="1244" alt="Screen Shot 2022-10-11 at 9 30 58 PM" src="https://user-images.githubusercontent.com/15519577/195237253-5190cae7-f5c6-40be-bdfb-478bba0adc9e.png">

It's just that I needed to set the Start Date back earlier, to surround the 'expected_start_time' values:
<img width="1244" alt="Screen Shot 2022-10-11 at 9 52 22 PM" src="https://user-images.githubusercontent.com/15519577/195237355-5d23c604-6346-4b41-a991-6fd16de16dd6.png">
It's odd that they appear only in the below plot, when the above plot should also have them. 
ie - the X-axis seems to mean Start time :)

### Browers

- [X] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Prefect version

```Text
2.5.0
```

billpalombi:
This filter should use `start_date` if one exists and `expected_start_date` if `start_date` does not exist.

zhen0:
This would need a double sort from the api so is not currently possible.

billpalombi:
[...]

PR #7496, opened 2022-11-10
Add start-time sort for Flow Runs

anticorrelator:
implements the API half of https://github.com/PrefectHQ/prefect/issues/7134
to close, we'd need to change the UI to use the new `start_time` sort methods

Not all Runs have a `start_time` so we only implemented an `expected_start_time` sort instead. In #7134 we note that this leads to quite a bit of UI confusion, so instead of permitting compound sorting, here I'm experimenting with a special sort method for `start_time` which falls back to `expected_start_time` if a Run's `start_time` is not available.

### Checklist


src/prefect/orion/database/migrations/versions/postgresql/2022_11_10_171740_8caf7c1fd82c_add_coalesced_start_time_indices.py
```diff
@@ -0,0 +1,53 @@
...
+    with op.get_context().autocommit_block():
+        op.execute(
+            """
+            CREATE INDEX CONCURRENTLY
+            ix_flow_run__coalesce_start_time_expected_start_time_asc
+            ON flow_run (coalesce(start_time, expected_start_time) ASC);
+            """
+        )
+
+    with op.get_context().autocommit_block():
```
chrisguidry, line 28: You can do them both in one block, but it's not a big deal either way

chrisguidry (approved):
That's 🌶️ !

PR #7499, opened 2022-11-10
Introduces Prefect Events and the clients that can emit them

chrisguidry:
This introduces the shell for what will become a more fleshed-out implementation later.  In this change, I'm just carving out the `prefect.events` namespace, introducing the client-side `Resource` schemas and abstract client for emitting Events.

This includes a FastAPI dependency that we can use in API routes to have access to an event client.  The idea is that in the beginning, this will be a `NullEventsClient`, so there should be only a very tiny performance change to use it.  Later, we can replace the implementation using FastAPI's `dependency_overrides` to drop in a "real" client.

### Example

See the included tests.

### Checklist

### DIFF
--- a/src/prefect/events/clients.py
+++ b/src/prefect/events/clients.py
@@ -1,0 +1,57 @@
+import abc
+from types import TracebackType
+from typing import List, Optional, Type
+
+from .schemas import Event
+
+
+class EventsClient(abc.ABC):
+    """The abstract interface for a Prefect Events client"""
+
+    @abc.abstractmethod
+    async def emit(self, event: Event) -> None:
+        ...
+
+    async def __aenter__(self) -> "EventsClient":
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[Exception]],
+        exc_val: Optional[Exception],
+        exc_tb: Optional[TracebackType],
+    ) -> "EventsClient":
+        return self
+
+
+class NullEventsClient(EventsClient):
+    """An implementation of the Prefect Events client that does nothing"""
+
+    async def emit(self, event: Event) -> None:
+        pass
+
+
+class AssertingEventsClient(EventsClient):
+    """An implementation of the Prefect Events client that records all events sent
+    to it for inspection during tests."""
+
+    events: List[Event]
+
+    def __init__(self):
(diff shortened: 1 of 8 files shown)


src/prefect/events/clients.py
```diff
@@ -0,0 +1,57 @@
...
+    async def __aenter__(self) -> "EventsClient":
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[Exception]],
+        exc_val: Optional[Exception],
+        exc_tb: Optional[TracebackType],
+    ) -> "EventsClient":
+        return self
```
madkinsz, line 24: The return value has a meaning here, I don't think you want to return self.

src/prefect/events/clients.py
```diff
@@ -0,0 +1,57 @@
+import abc
+from types import TracebackType
+from typing import List, Optional, Type
+
+from .schemas import Event
```
madkinsz, line 5: We discourage relative imports.

src/prefect/events/clients.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class NullEventsClient(EventsClient):
+    """An implementation of the Prefect Events client that does nothing"""
+
+    async def emit(self, event: Event) -> None:
+        pass
+
+
+class AssertingEventsClient(EventsClient):
```
madkinsz, line 34: Should this and the `NullEventsClient` be in `prefect.testing.events`? It may be nice to keep them out of the "maintained" interface.

src/prefect/events/clients.py
chrisguidry: Oops, good call!

src/prefect/events/clients.py
chrisguidry: It does make sense to move the `AssertingEventsClient` there, but the `NullEventsClient` will be the one that runs for most people unless they configure another one.  I think that's a "real" one, so to speak.

src/prefect/events/clients.py
chrisguidry: Oof, that's _really_ bad, it would be a truthy value, thus quietly swallowing exceptions ha!  Thank you!

PR #7514, opened 2022-11-11
Improve orchestration injection semantics

anticorrelator:
A few of our dependency injection semantics assumed a bit of structure that can complicate extending our orchestration engine. These changes tweak a couple things:

- we push logic that prevents firing `UniversalTransforms` on nullified transitions down into the transform itself. We may wish to extend bookkeeping to fire on all transitions—nullified or not—in the future.
- we unify the interfaces for `UniversalTransforms` and `OrchestrationRules`, a `UniversalTransform` is usefully equipped with the intended transition when initialized, we may want to track bookkeeping on the specifics of the state transition being orchestrated.
- we allow orchestration parameters to be injectable via FastAPI

### Checklist


src/prefect/orion/api/flow_runs.py
```diff
@@ -34,6 +34,10 @@ async def create_flow_run(
     flow_run: schemas.actions.FlowRunCreate,
     db: OrionDBInterface = Depends(provide_database_interface),
     response: Response = None,
+    orchestration_parameters: dict = Depends(
+        orchestration_dependencies.provide_flow_orchestration_parameters
```
chrisguidry, line 38: DELICIOUS

src/prefect/orion/orchestration/dependencies.py
```diff
@@ -31,6 +33,28 @@ async def provide_flow_policy():
...
 
+async def provide_task_orchestration_parameters():
+    provided_parameters = ORCHESTRATION_DEPENDENCIES.get(
+        "task_orchestration_parameters"
+    )
+
+    if provided_parameters is None:
+        provided_parameters = dict()
+
+    return provided_parameters
```
chrisguidry, line 44: What do you think about having `ORCHESTRATION_DEPENDENCIES["task_orchestration_parameters"]` be a _function_ that returns the `dict` of parameters, instead of just a static `dict` of parameters?  Seems like it would cover more cases?

src/prefect/orion/orchestration/global_policy.py
```diff
@@ -78,7 +79,8 @@ class SetRunStateName(BaseUniversalTransform):
     async def before_transition(self, context: OrchestrationContext) -> None:
 
         # record the new state's name
-        context.run.state_name = context.proposed_state.name
+        if not self.nullified_transition():
+            context.run.state_name = context.proposed_state.name
```
chrisguidry, line 83: I'm more into:

```
if self.nullified_transition():
    return

...do the things...
```

Just because it keeps the indentation lower and it's a little clearer that the check is a guard for the whole function in a lot of these cases

chrisguidry (approved):
This looks solid, just a couple notes.

I notice this doesn't include a facility to dependency-inject the `GlobalFlowPolicy`/`GlobalTaskPolicy`.  Is the idea that we'd just augument the regular core policies but now we can use universal transforms in that stack of policies?

src/prefect/orion/orchestration/dependencies.py
anticorrelator: absolutely!

anticorrelator:
@chrisguidry thanks for the review! With these changes we can keep the `GlobalFlow/TaskPolicy` special, as those transforms implement `Orion`'s critical bookkeeping functionality as close as possible to the point we transition to a new state. Instead we're making `Transforms` interface-compliant with `Rules`, so they can be injected together in the same policy object in any order. Personally I think this makes more sense, as a `Policy` should be able to include any kind of orchestration primitive

chrisguidry (approved):
Excellent!  Thank you, this definitely improves the dependency injection and customization for orchestration.

PR #7532, opened 2022-11-14
Add tests that demonstrate deadlocks

bunchesofdonald:


madkinsz (approved):
Wonderful, these are nice and readable. Not sure about the `test_deadlocks` namespace once they're no longer deadlocks but we can move them later :)

PR #7537, opened 2022-11-15
Fix linking to Collections catalog

tpdorsey:
Reorganizing the collections catalog broke a previous URL to the collections. Fixing link in Blocks docs and adding a redirect to handle any user bookmarks.

### Checklist

### DIFF
--- a/docs/concepts/blocks.md
+++ b/docs/concepts/blocks.md
@@ -257,7 +257,7 @@ Blocks can be registered from a Python module available in the current virtual e
 $ prefect block register --module prefect_aws.credentials
 ```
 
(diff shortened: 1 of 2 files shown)


discdiver (approved):
LGTM! Thanks for the quick change!  One question - should /collections by itself direct to collections/overview? Not sure if we used /collections previously anywhere.

tpdorsey:
@discdiver /collections/ has not previously pointed to anything. I will look into whether there's a straightforward way to do that, should we need it

desertaxle (approved):
LGTM!

PR #7538, opened 2022-11-15
Add CLI Table utility to generate searchable, pageable tables from iterables

peytonrunyan:
I'm mostly seeking a design review at the moment.

This PR would add a utility to create searchable, pageable CLI tables from an iterable.

There is one issue with the paging, which is that the the terminal window needs to be minimum size, otherwise the top of the table re-renders, which looks ugly if you scroll up in your terminal. If you start the table, see that it is re-rendering, and increase the size of the window, it will stop doing it. 

Ugly paging
Screen Shot 2022-11-15 at 11 28 11 AM

The minimum size is pretty reasonable.
Screen Shot 2022-11-15 at 11 29 12 AM

### Example

```python
from prefect.cli._utilities import CliTable
import string

options = [
    s for s in string.ascii_lowercase
]

table = CliTable(elements=options, title="My demo table")
res = table.start()
print(res)
```
### Checklist


madkinsz:
Going to close as stale; we should revisit this though!

PR #7542, opened 2022-11-15
Fix positioning of Prefect Cloud link

tpdorsey:
Fixes positioning of "Go to Prefect Cloud" link in docs header.

This was floating right properly, but somewhere along the way it moved back to the left. I think this is a CSS difference between Material for Mkdocs and Insiders. 

Anyway minor adjustment to get it right-justified with Insiders builds again.

### Checklist

### DIFF
--- a/docs/overrides/partials/tabs.html
+++ b/docs/overrides/partials/tabs.html
@@ -31,7 +31,8 @@
       {% for nav_item in nav %}
         {% include "partials/tabs-item.html" %}
       {% endfor %}
-      <li class="md-tabs__item cloud-link" style="float: right;"><a class="md-tabs__link" target="_blank" rel="noopener noreferrer" href="https://app.prefect.cloud"><b>Go to Prefect Cloud &#187;</b></a></li>
+      <li class="md-tabs__item cloud-filler"></li>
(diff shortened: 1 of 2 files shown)


docs/overrides/partials/tabs.html
```diff
@@ -31,7 +31,8 @@
       {% for nav_item in nav %}
         {% include "partials/tabs-item.html" %}
       {% endfor %}
-      <li class="md-tabs__item cloud-link" style="float: right;"><a class="md-tabs__link" target="_blank" rel="noopener noreferrer" href="https://app.prefect.cloud"><b>Go to Prefect Cloud &#187;</b></a></li>
+      <li class="md-tabs__item cloud-filler"></li>
+      <li class="md-tabs__item cloud-link"><a class="md-tabs__link" target="_blank" rel="noopener noreferrer" href="https://app.prefect.cloud"><b>Go to Prefect Cloud &#187;</b></a></li>
```
znicholasbrown, line 35: If you wanted to avoid the extra list item, you could add an auto left margin to `.cloud-link` like this:

```css
.cloud-link {
  margin-left: auto;
}
```

which would let you remove the `.cloud-filler` class + element

tpdorsey: That's a far better solution. Thank you.

PR #7546, opened 2022-11-16
Added support for pickle-files in import_object-function

### LINKED ISSUE #7545: Add support for pickled files in importtools utilities
### First check

### Prefect Version

2.x

It is currently not possible to import pickled files.

It would be cool if we were able to load flows from pickled-files in the same fashion Prefect 1 did it.

### Example Use

If an object has been pickled:

```python
import cloudpickle

class MyClass:
  def __init__(self, a: int, b: str):
    self.a = a
    self.b = b

mc = MyClass(123, "foo")
open("mc.pickle", "wb+").write(cloudpickle.dumps(mc))
```

It should be possible to load it using the `importtools::import_object`-function:

```

mc = import_object("mc.pickle")
```

### Additional context

By adding pickle-support to `import_object`, deployments can specify pickled flows as entrypoint instead of python-files, similar to how flows were stored in Prefect 1.

The `.pickle` file-extension references that the file is a pickled python-object, however the extension `.flow` was used in Prefect 1.

thomafred:
Added support for pickled-files so that flows may be loaded the same way they were in Prefect 1. This also means that deployments can be created problematically together with the `Flow.with_options`-function in a more streamlined fashion.

Closes #7545 

### Example

Users can load a pickled object using the `importtools::import_object`-function:

```

mc = import_object("mc.pickle")
```

### Checklist

### DIFF
--- a/src/prefect/utilities/importtools.py
+++ b/src/prefect/utilities/importtools.py
@@ -9,6 +9,7 @@ from tempfile import NamedTemporaryFile
 from types import ModuleType
 from typing import Any, Dict, Union
 
+import cloudpickle
 import fsspec
 
 from prefect.exceptions import ScriptError
@@ -177,6 +178,17 @@ def load_module(module_name: str) -> ModuleType:
         sys.path.remove(working_directory)
 
 
+def load_object_from_pickle(pickle_path: str) -> Any:
+    """
+    Import an object from a pickle-file
+    """
+
+    with open(pickle_path, "rb") as fd:
+        obj = cloudpickle.load(fd)
+
+    return obj
+
+
 def import_object(import_path: str):
     """
     Load an object from an import path.
@@ -185,12 +197,15 @@ def import_object(import_path: str):
     - module.object
     - module:object
     - /path/to/script.py:object
+    - /path/to/object.pickle
 
     This function is not thread safe as it modifies the 'sys' module during execution.
     """
     if ".py:" in import_path:
         script_path, object_name = import_path.rsplit(":", 1)
         module = load_script_as_module(script_path)
+    elif import_path.endswith(".pickle"):
+        return load_object_from_pickle(import_path)
     else:
         if ":" in import_path:
             module_name, object_name = import_path.rsplit(":", 1)
--- a/tests/utilities/test_importtools.py
+++ b/tests/utilities/test_importtools.py
@@ -1,10 +1,12 @@
 import importlib.util
 import runpy
 import sys
+import tempfile
 from pathlib import Path
 from types import ModuleType
 from uuid import uuid4
 
+import cloudpickle
 import pytest
 
 import prefect
@@ -234,3 +236,16 @@ def test_import_object_from_module_with_relative_imports_expected_failures(
         # Python would raise the same error
         with pytest.raises((ValueError, ImportError)):
             runpy.run_module(import_path)
+
+
+def test_load_pickled_object():
+    with tempfile.TemporaryDirectory() as working_directory:
+        f = Foo()
+        f.a = 123
+
+        with open(f"{working_directory}/foo.pickle", "wb+") as fd:
+            fd.write(cloudpickle.dumps(f))
+
+        foobar = import_object(f"{working_directory}/foo.pickle")
+        assert foobar.__class__.__name__ == f.__class__.__name__
+        assert foobar.a == f.a


src/prefect/utilities/importtools.py
```diff
@@ -185,12 +197,15 @@ def import_object(import_path: str):
     - module.object
     - module:object
     - /path/to/script.py:object
+    - /path/to/object.pickle
```
madkinsz, line 200: What are your thoughts on `pickle:///path/to/object` instead?

thomafred: That seems intuitive, however the `file://`-format is typically used to specify format (`s3:///bucketName/path/to/blob` for S3 for example).

What do you think?

madkinsz: Hm, I'm thinking of how fsspec chains urls to indicate compression (https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining)

They'd do `zip::s3://....` to indicate reads/writes should work with compressed content. Similarly, a `pickle::file:///...` makes sense to me. I'm hesitant to rely to the file extension, but not entirely opposed. I'll see if other people on our team have thoughts.

thomafred: the fsspec-scheme seems like a good idea. Relying on the file-extensions is indeed not ideal, especially as Prefect 1 used the extension `.flow` for pickled flows.

anna-geller:
@thomafred could you provide more background on what problem motivated this? we have lots of options with respect to how you can store your flows on deployments, I'm curious to hear how you arrived at this

thomafred:
We have created a custom deployment mechanism based on docker. This comes with a config-file which is intended to separate running code and infrastructure and a small CLI that handles running and deploying pipelines. The primary motivation behind this CLI was initially authenticaion, as we are using an OAUTH2-based authentication flow on Azure AD.

In the config-file, the users specifies things like what storage to use, including what storage to use and packages to install in the case with Docker, schedule and of course what runtime to use.

Our runtimes are essentially wrappers around the `KubernetesJob` infrastrcture, but has some convenience-functions built in such as resource handling and what container image to use. That way our user only specifies that they want the "small" tier in the runtime and automatically is allocated 2 vCPUs.

While working on support of a Dask-runtime that would create ephemeral Dask clusters, we saw the need to dynamically set the taskrunner of a flow dynamically at deploy-time. The flow then needs to be saved in its modified state so that the correct task runner. This is achieved by storing the flow as a pickle, hence this change :)

Right now we have a work-around where the pickle is saved alongside a small python-file which unpickles the file and exposes the flow:

```python
open(os.path.join(dir, f"{flow_name}.py"), "w+").write(
    textwrap.dedent(
        f"""
    import cloudpickle
    import os
    os.chdir(os.path.dirname(__file__))
    {flow_name} = cloudpickle.loads(open("{flow_name}.pickle", "rb").read())
"""
    )
)
```

Not pretty, but it works.

Here's an example config btw:

```yaml
metadata:
  name: example-flow
  description: An example flow. No bells or whistles
runtime:
  cls: odp.compute.deploy.runtime.KubernetesTieredResource
  tier: small
  env:
    SOME_VAR: some_value
storage:
  cls: odp.compute.deploy.storage.Docker
  files:
    "{{flow_dir}}/tasks": "/tasks/"
```

anna-geller:
> We have created a custom deployment mechanism based on docker. This comes with a config-file which is intended to separate running code and infrastructure and a small CLI that handles running and deploying pipelines. The primary motivation behind this CLI was initially authenticaion, as we are using an OAUTH2-based authentication flow on Azure AD.
> 
> In the config-file, the users specifies things like what storage to use, including what storage to use and packages to install in the case with Docker, schedule and of course what runtime to use.
> 
> Our runtimes are essentially wrappers around the `KubernetesJob` infrastrcture, but has some convenience-functions built in such as resource handling and what container image to use. That way our user only specifies that they want the "small" tier in the runtime and automatically is allocated 2 vCPUs.
> 
> While working on support of a Dask-runtime that would create ephemeral Dask clusters, we saw the need to dynamically set the taskrunner of a flow dynamically at deploy-time. The flow then needs to be saved in its modified state so that the correct task runner. This is achieved by storing the flow as a pickle, hence this change :)
> 
> Right now we have a work-around where the pickle is saved alongside a small python-file which unpickles the file and exposes the flow:
> 
> ```python
> open(os.path.join(dir, f"{flow_name}.py"), "w+").write(
>     textwrap.dedent(
>         f"""
>     import cloudpickle
>     import os
>     os.chdir(os.path.dirname(__file__))
>     {flow_name} = cloudpickle.loads(open("{flow_name}.pickle", "rb").read())
> """
>     )
> )
> ```
> 
> Not pretty, but it works.
> 
> Here's an example config btw:
> 
> ```yaml
> metadata:
>   name: example-flow
>   description: An example flow. No bells or whistles
> runtime:
>   cls: odp.compute.deploy.runtime.KubernetesTieredResource
>   tier: small
>   env:
>     SOME_VAR: some_value
> storage:
>   cls: odp.compute.deploy.storage.Docker
>   files:
>     "{{flow_dir}}/tasks": "/tasks/"
> ```

Would this work for you? https://github.com/PrefectHQ/prefect/issues/5560#issuecomment-1274800069 
this would allow you to solve it without having to rely on pickle (which is often prone to errors with versioning and serialization issues)

thomafred:
@anna-geller This would work, however it exposes this detail to the engineers. I would still love for infrastructure config to live in a separate config

anna-geller:
Have you thought about a different way of solving that problem? asking because pickle has issues with portability (serialization tied to python and specific versions), e.g. if you don't want to expose this directly, maybe you could solve it with a shared config or packaging of that logic in some way

not suggesting any solution in particular, just wondering how one could solve it in general

madkinsz:
We're likely to add configuration of task runners with blocks like infrastructure blocks in the future.

Would you be able to use a JSON block for this? e.g. `@flow(task_runner=DaskTaskRunner(**JSON.load("deploy-time-config").value))`

thomafred:
That's an option, however the dask-clusters are not accessible outside of the Kubernetes cluster meaning that the pipeline have to be deployed in order to test it.

Having some kind of block doing this would make sense. It could for example dynamically load the taskrunner depending on whether it is running in the cluster or locally on the developer computer

anna-geller:
> That's an option, however the dask-clusters are not accessible outside of the Kubernetes cluster meaning that the pipeline have to be deployed in order to test it.
> 
> Having some kind of block doing this would make sense. It could for example dynamically load the taskrunner depending on whether it is running in the cluster or locally on the developer computer

wouldn't you prefer to aim for environment parity? Testing locally on local K8s cluster, testing in DEV on dev K8s cluster, running in PROD on prod K8s cluster, etc

thomafred:
We do posses DEV/PROD clusters, but before deploying to these our data engineers test execution on their personal computers first. Having them set up personal k8s clusters, even using minikube/kind/etc, is seen as additional complexity we can live without.

Our data engineers have full control over their own execution environment and is free to attach debuggers, set breakpoints, etc, without setting up complex development environments. We only require that the code they write is portable enough that it can run both locally and in the cluster, hence why the taskrunner must be dynamically configured.

thomafred:
From the input, it seems like supporting pickle may not be a good fit for Prefect core. Feel free to reopen in the future :)

anna-geller:
thanks for understanding and keep us posted via Slack if the other solution we were talking about shouldn't work as expected 🙌

PR #7555, opened 2022-11-16
Add basic benchmarking to CI

madkinsz:
Uses `pytest-benchmark` to run benchmarks for flows and tasks. Benchmarks are in the `benches/` folder and use `bench_` as a prefix instead of `test_`.

Benchmarks will display a comparison to those on the base branch of a pull request.

See the [benchmarking readme](https://github.com/PrefectHQ/prefect/tree/engine/benchmarks/benches) for details on local usage.

### Checklist


.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,58 @@
...
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  run-benchmarks:
+    name: Benchmark
+    # Runs with ephemeral API and SQLite database right now
```
bunchesofdonald, line 29: Is using SQLite just the easiest thing to do or do we want to make sure we have good benchmarks against SQLite _and_ Postgres at some point?

benches/bench_flows.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @flow
+    def test_flow():
+        for _ in range(num_tasks):
+            test_task.submit()
+
+    benchmark(test_flow)
+
+
+@pytest.mark.skip(reason="Runs longer than the 60s timeout")
```
bunchesofdonald, line 33: That we can't run 200 no-op tasks in 60s seems like a great, but sad, place to start benchmarking.

benches/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def reset_object_registry():
```
bunchesofdonald, line 5: I know I put something like this in other places in the test suite, might be worth culling those.

bunchesofdonald (approved):
This looks like an awesome start. I haven't used pytest-benchmark before but it looks solid.

.github/workflows/benchmarks.yaml
madkinsz: I think testing against both and including tests with a hosted API instead of ephemeral are all important, but I think they'll come after this PR.

benches/bench_flows.py
madkinsz: It's running multiple iterations of the benchmark to get more stable numbers, so the "test" exceeds that timeout but each bench itself does not. I'm going to bump the timeout here.

benches/conftest.py
madkinsz: We'll want to consolidate these in general, this is duplicated from our `tests/conftest.py` — I figure we'll tackle redundancies here in the future.

PR #6101, opened 2022-07-29
Add colored log levels

ahuang11:
Docs preview:
https://deploy-preview-6101--prefect-orion.netlify.app/concepts/logs/#styles

Adds coloring with the help of Rich; sample output below.

Implements `PrefectConsoleHandler` which is a custom implementation of logging.Handler, and mimics a small part of [RichHandler](https://rich.readthedocs.io/en/stable/_modules/rich/logging.html#RichHandler).

Michael's comment on the [old Orion PR](https://github.com/PrefectHQ/orion/pull/2334#issuecomment-1186549615):
"I'd prefer to add our own handler that properly colors by level. This doesn't distinguish between the different columns (e.g. the timestamp is boldly colored but is the least important piece of the log). I'm not sure rich's highlighting of objects in the message is useful for logging. We'll need this to be easily togglable by a global setting so users can disable it."

Currently PrefectConsoleHandler [defaults to PrefectConsoleHighlighter](https://github.com/PrefectHQ/prefect/pull/6101/files#diff-d980f72a718028b8f1396848bdbc658325de4c15380facf689cf49f8edfd8c17R367), which is able to highlight:
1. log levels
2. web and file urls
3. flow and task (run) names
4. states

Users are able to control the highlighted portions of these logs in `logging.yml` (if no color is specified, the highlighted sections do not get colored) and has a [global setting](https://github.com/PrefectHQ/prefect/pull/6101/files#diff-652b63566d2fcd4e36ea4e8509b4e0cd56e607b9e54e043acbde5bb6907a599bR397) so users can disable it.

Also, I decided to only highlight the flow / flow run names of the top level console handler and not `console_flow_runs` and `console_task_runs` because it can lead to too much highlighting

On my terminal:
image

On my  VSCode notebook:
image

On my browser notebook:
image

Test code:
```
import asyncio
from prefect import flow, get_run_logger

@flow
async def test_async_flow():
    await asyncio.sleep(3)
    logger = get_run_logger()
    logger.debug("a debug")
    logger.info("my info")
    logger.warning("some warning")
    logger.critical("super critical")
    logger.error("error!")
    print('all done')

@flow
def test_sync_flow():
    test_async_flow()

test_sync_flow()
```

I tried to match the UI's colors:
```
    LEVEL_COLORS = {
        "DEBUG": "#627288",
        "INFO": "#38BDF8",
        "WARNING": "#FBBF25",
        "ERROR": "#F43F5E",
        "CRITICAL": "#be123d",
    }
```
[...]

PR #7558, opened 2022-11-16
Fix `PrefectConsoleHandler` bug where log tracebacks were excluded

madkinsz:
The logging handler added in #6101 did not include the proper formatting and consequently dropped tracebacks from `logger.exception(...)`. This pull request:
- Fixes the bug, restoring tracebacks
- Disables word-wrapping/cropping at the same time because log output should not be wrapped
- Restores usage of stderr for logs instead of stdout

### Checklist

### DIFF
--- a/src/prefect/logging/handlers.py
+++ b/src/prefect/logging/handlers.py
@@ -361,9 +361,10 @@ class OrionHandler(logging.Handler):
         return super().close()
 
 
-class PrefectConsoleHandler(logging.Handler):
+class PrefectConsoleHandler(logging.StreamHandler):
     def __init__(
         self,
+        stream=None,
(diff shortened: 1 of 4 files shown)


abrookins:
I came to approve, but I can't quite make heads or tails of the test failures. Will check back again!

bunchesofdonald (approved):
Makes sense to me. The soft_wrap fixes a big issue with testing the CLI in general.

The test failures suggest that there's something going on with rich writing to `self.stream` though, but it could just be a test issue.

madkinsz:
> The soft_wrap fixes a big issue with testing the CLI in general.

This was already addressed at https://github.com/PrefectHQ/prefect/blob/main/tests/conftest.py#L282

This prevents cropping of logs which was never a thing before but was now that the rich is being used.

PR #7563, opened 2022-11-17
Make stale action more likely to close issues

bunchesofdonald:
This makes several small changes to the stale action:

- Disables PR stale-ing. When debug / analytics were turned on a lot of operations were spent (over half) checking PRs and given the relatively small set of PRs it's likely not a great use of those operations.
- Sets `ascending` to `true`. If I'm reading the docs correctly this should result in in looking at old issues first.
- Doubles the number of operations it can perform, default is 30, this gives it 60.
- Updates messaging.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -11,11 +11,10 @@ jobs:
         with:
           stale-issue-message: "This issue is stale because it has been open 30 days with no activity. To keep this issue open remove stale label or comment."
           stale-issue-label: "status:stale"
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
I do think it'd be nice to stale PRs in the long run still.

madkinsz:
What's the downside of a high number of operations?

bunchesofdonald:
It's about the rate-limit for the GitHub API, the more used by this the less we have for any other action that may be using the API. It's _probably_ not a huge issue, but I want to avoid other things failing because this is trying to make things stale.

PR #7477, opened 2022-11-09
LocalFileSystem check for empty paths

kielnino:
This PR adds check for the paths in LocalFileSystem. So instead of checking only for `None` checks for empty strings are added.

Edit by @peytonrunyan - closes https://github.com/PrefectHQ/prefect/issues/7317
### Checklist

- [x] This pull request references any related issue by including "closes" 
- [ ] This pull request includes tests or only affects documentation.
- [ ] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

Issue #7317, opened 2022-10-25
FileNotFoundError when building a Deployment with LocalFileSystem Block


prefectcboyd:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

By default, without any additional parameters for a "deployment build", it defaults to a temporary location, using the process infrastructure block.

If a LocalFileSystem block is registered, and passed as a storage-block, the deployment build fails with
```
FileNotFoundError: [Errno 2] No such file or directory: ''
```

This occurs in both cases:
1. When a path is specified, such as /tmp or my current working directory on my local system
2. When no path is specified (the block creation lists path as 'Optional')
3. 

### Reproduction

```bash
This is all completed on my local system.


#Create a new venv
python3 -m venv missing_dir_issue

#Activate
source missing_dir_issue/bin/activate

# Install prefect
pip3 install prefect

#Make folder for flows, and copy in healthcheck.py
mkdir missing_dir_issue/flows
cd missing_dir_issue/flows

#create healthcheck.py
cat <<EOF > healthcheck.py
import prefect
from prefect import task, flow
from prefect import get_run_logger

@task
def log_platform_info():
    import platform
    import sys
    from prefect.orion.api.server import ORION_API_VERSION
    logger = get_run_logger()
    logger.info("Host's network name = %s", platform.node())
    logger.info("Python version = %s", platform.python_version())
    logger.info("Platform information (instance type) = %s ", platform.platform())
    logger.info("OS/Arch = %s/%s", sys.platform, platform.machine())
    logger.info("Prefect Version = %s 🚀", prefect.__version__)
    logger.info("Prefect API Version = %s", ORION_API_VERSION)
@flow
def healthcheck():
    log_platform_info()
if __name__ == "__main__":
    healthcheck()
EOF


# Confirm version 

prefect version
Version:             2.6.3
API version:         0.8.2
Python version:      3.9.12
Git commit:          9e7da96e
Built:               Tue, Oct 18, 2022 1:55 PM
OS/Arch:             darwin/x86_64
Profile:             boyd
Server type:         cloud

#Authenticate to cloud 
prefect cloud login -k <>

#Create a local file system block ; two can be created, one with a path, and one left blank
prefect block create local-file-systemprefect block create local-file-system
Create a local-file-system block:
[...]

PR #7564, opened 2022-11-17
Fix error with empty paths using LocalFileSystem

peytonrunyan:
This is a follow on PR to https://github.com/PrefectHQ/prefect/pull/7477
It adds tests for the changes made in that PR, and also adds handling for cases where a `basepath` was not provided to the `LocalFileSystem` which is necessary to address part of https://github.com/PrefectHQ/prefect/issues/7317 .

Previously we did a check like `if not from_path: from_path = Path(self.basepath).expanduser().resolve()` in `get_directory` and `put_directory`. This would error if the `basepath` was None. Now if an LFS block without a basepath calls get/put directory, it will treat the cwd as the basepath.

This currently CANNOT be merged until we change how temporary directories are created when running the engine.

The Process infrastructure block's run method currently spins up a subprocess in a random directory. Then from within that process, the storage block's get_directory method is called. put_directory is not called from this directory. This means that this isn't working correctly even when you resolve the paths if you don't supply as basepath.

### Checklist

### DIFF
--- a/src/prefect/filesystems.py
+++ b/src/prefect/filesystems.py
@@ -92,13 +92,17 @@ class LocalFileSystem(WritableFileSystem, WritableDeploymentStorage):
             return str(value)
         return value
 
-    def _resolve_path(self, path: str) -> Path:
-        # Only resolve the base path at runtime, default to the current directory
-        basepath = (
+    def _get_basepath(self):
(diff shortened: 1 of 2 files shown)


madkinsz:
@peytonrunyan what's the status of this?

Issue #7117, opened 2022-10-10
Stream print statements as logs similarly to the `log_stdout` in v1


anna-geller:
The key value prop of Prefect is that users don't need to rewrite their code. Their code is workflow. But to get any logs streamed to the backend, we **do** require users to change their code to use the Prefect logger or to configure extra loggers. Simple print statements are currently ignored by Prefect.

This gets asked a lot and is important to support incremental adoption 

![Image](https://user-images.githubusercontent.com/86264395/194860461-741f568f-b701-43d2-afb6-976e0aa37fbe.png)

anna-geller:
@anticorrelator for more context, this is how it was done in v1: we were redirecting stdout to "log_stdout" logger https://github.com/PrefectHQ/prefect/blob/1.x/src/prefect/engine/task_runner.py#L873-L877 and then using this utility: https://github.com/PrefectHQ/prefect/blob/1.x/src/prefect/utilities/logging.py#L375

billpalombi:
We'll need to consider the implications of encouraging users sending us more logs. We may need to invest in backend improvements to support greater volume.

anna-geller:
Workaround from a SO user struggling with [the same issue](https://stackoverflow.com/questions/74138833/forwarding-stdout-and-stderr-to-logger-in-prefect-2-flows/74138948#74138948)

zangell44:
Fwiw I don't think this issue should be blocked by log volume scaling concerns. Seems like a great user win that has only a modest increase in risk.

gabcoyne:
This issue is blocking some significant migration from 1-2

anna-geller:
cc @anticorrelator 🙏

madkinsz:
After further thought this is much easier if we require it to be a global setting rather than at the flow or task level. This is still quite challenging though and is certain to result in unexpected behavior.

I think we'd be much better off focusing on https://github.com/PrefectHQ/prefect/issues/6767#issuecomment-1242484405 if that would unblock usage.

WillRaphaelson:
https://github.com/PrefectHQ/prefect/pull/7580

PR #7580, opened 2022-11-17
Add `log_prints` option to redirect print to logs

madkinsz:
Addresses the much requested `log_stdout` feature from v1 in v2 by adding a `log_prints` option. It is very hard for us to log standard output from _within_ the run as we have no way to tell where the bytes are coming from (i.e. you cannot disable capturing per task or flow). We can, however, patch the builtin `print` method, allowing us to identify the flow or task run the message comes from. 

This pull request introduces temporary patching of `print` to accomplish logging of prints from within tasks or flows. This will also apply to any functions within the scope of a task or a flow e.g. if they call into another library. If a flow or task run context is not present, we will just `print` as normal.

There are options at the flow, task, and settings level. Defaults for children (subflows and tasks) are pulled from the parent flow.

Closes https://github.com/PrefectHQ/prefect/issues/7117
### Example

Play with the following, try `None`, `False` and `True` for values:
```python
import prefect


@prefect.flow(log_prints=False)
def foo():
    print("foo")
    bar()


@prefect.task(log_prints=True)
def bar():
    print("bar")


foo()
```
### Checklist

### Future work

We may add a `log_stdout` feature but it would operate at the infrastructure level where we actually have access to the full standard output stream.

### DIFF
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -224,6 +224,7 @@ class Flow(Generic[P, R]):
         result_storage: Optional[ResultStorage] = NotSet,
         result_serializer: Optional[ResultSerializer] = NotSet,
         cache_result_in_memory: bool = None,
+        log_prints: Optional[bool] = NotSet,
     ):
         """
         Create a new flow from the current object, updating provided options.
@@ -304,6 +305,7 @@ class Flow(Generic[P, R]):
                 if cache_result_in_memory is not None
                 else self.cache_result_in_memory
             ),
+            log_prints=log_prints if log_prints is not NotSet else self.log_prints,
         )
 
     def validate_parameters(self, parameters: Dict[str, Any]) -> Dict[str, Any]:
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -582,7 +582,7 @@ def flow(
             will be used unless called as a subflow, at which point the default will be
             loaded from the parent flow.
         log_prints: If set, `print` statements in the flow will be redirected to the
-            Prefect logger for the flow run. Defaults to `None` which indicates that
+            Prefect logger for the flow run. Defaults to `None`, which indicates that
             the value from the parent flow should be used. If this is a parent flow,
             the default is pulled from the `PREFECT_LOGGING_LOG_PRINTS` setting.
 
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -130,6 +130,7 @@ class Flow(Generic[P, R]):
         result_storage: Optional[ResultStorage] = None,
         result_serializer: Optional[ResultSerializer] = None,
         cache_result_in_memory: bool = True,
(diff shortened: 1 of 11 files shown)


madkinsz:
Needs:
- [ ] Tests
- [ ] Link the the closed issue

madkinsz:
Nate is leading tests for this.
Will is leading documentation.

anna-geller:
works 🎉 

```python
from prefect import flow

@flow(log_prints=True)
def hello():
    print("Greetings from Prefect! 🤗")

if __name__ == "__main__":
    hello()
```

image

```python

@task(log_prints=True)
def print_sth_nice():
    print("This is nice")


@flow(log_prints=True)
def hello():
    print("Greetings from Prefect! 🤗")
    print_sth_nice()

if __name__ == "__main__":
    hello()
```
image

anna-geller (approved):
QA passes

src/prefect/tasks.py
```diff
@@ -807,6 +816,11 @@ def task(
         result_serializer: An optional serializer to use to serialize the result of this
             task for persistence. Defaults to the value set in the flow the task is
             called in.
+        timeout_seconds: An optional number of seconds indicating a maximum runtime for
```
peytonrunyan, line 819: 😅

peytonrunyan (approved):
LGTM! The only thing is possibly a test explicitly for checking that the default behavior is `False`?

Also, TIL about the `builtins` module. I'd always thought "builtins" was just a colloquialism.

src/prefect/tasks.py
```diff
@@ -120,6 +120,9 @@ class Task(Generic[P, R]):
             called in.
         timeout_seconds: An optional number of seconds indicating a maximum runtime for
             the task. If the task exceeds this runtime, it will be marked as failed.
+        log_prints: If set, `print` statements in the task will be redirected to the
+            Prefect logger for the task run. Defaults to `None`, which indicates
+            that the value from the flow should be used.s
```
zzstoatzz, line 125:
```suggestion
            that the value from the flow should be used.
```

src/prefect/flows.py
```diff
@@ -574,6 +581,10 @@ def flow(
             in this flow. If not provided, the value of `PREFECT_RESULTS_DEFAULT_SERIALIZER`
             will be used unless called as a subflow, at which point the default will be
             loaded from the parent flow.
+        log_prints: If set, `print` statements in the flow will be redirected to the
+            Prefect logger for the flow run. Defaults to `None` which indicates that
```
tpdorsey, line 585:
```suggestion
            Prefect logger for the flow run. Defaults to `None`, which indicates that
```

tpdorsey (approved):
LGTM. Only reviewed docs and docstrings.

bunchesofdonald (approved):
I like it!

madkinsz:
Thanks for the help everyone!

PR #7588, opened 2022-11-18
Add `PrefectFormatter` to reduce logging configuration duplication

madkinsz:
Previously, flow run and task run loggers required their own formatters and did not propagate to the `prefect` logger. We did this to allow customization of formatting for flow and task run logs. However, this makes it difficult to change the behavior of Prefect's loggers since you need to adjust multiple handlers. We can accomplish this same feature with a single formatter that has support for flow and task run formats.

### Example

Now, since there are not separate formatters for flow and task runs, you can change the format with a single setting e.g. to use JSON logging

```
PREFECT_LOGGING_HANDLERS_CONSOLE_FORMATTER="json" python example.py
```

### Checklist

### DIFF
--- a/src/prefect/logging/formatters.py
+++ b/src/prefect/logging/formatters.py
@@ -1,4 +1,5 @@
 import logging.handlers
+import sys
 import traceback
 from types import TracebackType
 from typing import Optional, Tuple, Type, Union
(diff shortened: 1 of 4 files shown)


rpeden:
This looks great! I tried it and found: 
* Everything worked as expected on all Python versions we support. 
* A customized logging config based on an old version of `logging.yml` continued to work the way it worked before.

rpeden (approved):

peytonrunyan:
It looks good to me, but I'm not as familiar with logging configuration. Got some reading pulled up for later.

PR #7589, opened 2022-11-18
Ignore latest httpx version during development

madkinsz:
We are seeing failures in CI, likely due to incompatibility with `respx`


madkinsz:
Solved by https://github.com/lundberg/respx/releases/tag/0.20.1

Issue #7512, opened 2022-11-11
Network failures with self-hosted servers


madkinsz:
This is tracking issue for various reports of network failures when self-hosting Prefect Orion.

Notably, these issues seem focused to usage of Prefect 2.6.6.
 
If adding a report to this issue, please include the following information:

- If using Prefect official Docker images for the client or server, provide the image tags
- On the server, we are interested in the Prefect version, the database, and server library versions

```
prefect version
pip freeze | grep -E '(uvicorn|starlette)'
```

- On the client, we are interested in Prefect versions and the client HTTP library versions

```
prefect version
pip freeze | grep -E '(httpx|httpcore)'
```

- Please include the **full traceback** for the error
- Check for any related error logs on the server

Related to: 
- https://github.com/PrefectHQ/prefect/issues/7472

carlo-catalyst:
**Note**
We observed this with flows calling **run_deployment**

**Server**
- We run with docker using this image: prefecthq/prefect:2.6.6-python3.9
- Using postgres

**Agent**
- Our own custom image, with dependencies installed
- using that image, pulled these: <coming soon>



**Stack Trace**
 ```
run_deployment(name=run_deployment_name(flow=**_flow, env=env), parameters=parameters)
  File "/usr/local/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 197, in coroutine_wrapper
    return run_async_from_worker_thread(async_fn, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 148, in run_async_from_worker_thread
    return anyio.from_thread.run(call)
  File "/usr/local/lib/python3.9/site-packages/anyio/from_thread.py", line 49, in run
    return asynclib.run_async_from_thread(func, *args)
  File "/usr/local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 970, in run_async_from_thread
    return f.result()
  File "/usr/local/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/usr/local/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.9/site-packages/prefect/client/utilities.py", line 47, in with_injected_client
    return await fn(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/prefect/deployments.py", line 127, in run_deployment
[...]

PR #7593, opened 2022-11-18
Add retry on specified httpx network errors

peytonrunyan:
Begins to address https://github.com/PrefectHQ/prefect/issues/7512 and #7442 

This pull request moves adds retry logic for exceptions encountered when sending requests with `prefect/client/base.py::PrefectHttpxClient`. It also moves the logic for the retry from the `send` method to `_send_with_retry`, which takes a partial of the send function with the associated args and kwargs, a list of exceptions to retry under, and a list of status codes to retry under. 

### Checklist


src/prefect/client/base.py
```diff
@@ -154,27 +155,53 @@ class PrefectHttpxClient(httpx.AsyncClient):
...
+            not response
+            or (response.status in retry_codes)
+            and try_count <= self.RETRY_MAX
         ):
-            retry_count += 1
+            try:
+                response = await request()
+            except retry_exceptions as exc:
+                if try_count >= self.RETRY_MAX:
+                    raise exc
```
madkinsz, line 176:
```suggestion
                    raise
```
including the exc adds a frame to the traceback

src/prefect/client/base.py
```diff
@@ -154,27 +155,53 @@ class PrefectHttpxClient(httpx.AsyncClient):
...
+                    # response was successful
+                    if response.status_code not in retry_codes:
+                        return response
+                    # begin retry logic
+                    retry_after = response.headers.get("Retry-After")
+                if retry_after:
+                    retry_seconds = float(retry_after)
+                else:
+                    retry_seconds = 2**try_count
+                await anyio.sleep(retry_seconds)
```
madkinsz, line 191: I find this hard to read. Does this belong in a `finally` block?

src/prefect/client/base.py
```diff
@@ -154,27 +155,69 @@ class PrefectHttpxClient(httpx.AsyncClient):
...
+        try_count = 0
+        response = None
+        retry = True
+
         while (
-            response.status_code
-            in {status.HTTP_429_TOO_MANY_REQUESTS, status.HTTP_503_SERVICE_UNAVAILABLE}
-            and retry_count < self.RETRY_MAX
+            not response
+            or (response.status_code in retry_codes)
```
jakekaplan, line 180: I am not certain but could this be:
```
while (not response or retry):
```

It looks like retry is being set on the same statements that the while loop is conditional on? I might be missing a case where the reverse doesn't make sense. Or maybe the `retry` variable is unneeded? I can't quite put my finger on it 😅

jakekaplan (approved):
nice work! left a couple questions

src/prefect/client/base.py
```diff
@@ -154,27 +155,69 @@ class PrefectHttpxClient(httpx.AsyncClient):
...
             else:
-                retry_seconds = 2**retry_count
+                if response.status_code not in retry_codes:
+                    retry = False
+                else:
+                    retry_after = response.headers.get("Retry-After")
+                    if retry_after:
+                        retry_seconds = float(retry_after)
+
+            finally:
```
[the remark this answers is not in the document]
peytonrunyan: Yup - good catch

src/prefect/client/base.py
peytonrunyan: Is the new arrangement more readable?

jakekaplan: I'm so sorry, I think my original comment was a little vague and unhelpful! I think a combination of the conditionals underneath the `try/except/else/finally` structure makes it harder to follow the code paths. Maybe something like this could work? (Just to qualify a bit I don't know if this works and the solution as written is great, just seeing if it's possible to consolidate a bit)

```
        try_count = 0
        response = None

        while True:
            try_count += 1
            retry_seconds = None
            retry = try_count <= self.RETRY_MAX

            try:
                response = await request()
            except retry_exceptions:
                if not retry:
                    raise
                retry_seconds = 2**try_count
            else:
                if response.status_code in retry_codes:
                    retry_seconds = float(response.headers.get("Retry-After", 2**try_count))

            if retry and retry_seconds is not None:
                await anyio.sleep(retry_seconds)
            else:
                break

        return response
```

src/prefect/client/base.py
```diff
@@ -154,27 +155,74 @@ class PrefectHttpxClient(httpx.AsyncClient):
...
+    async def send(self, *args, **kwargs) -> Response:
+        api_request = partial(super().send, *args, **kwargs)
+
+        response = await self._send_with_retry(
+            request=api_request,
+            retry_codes={
+                status.HTTP_429_TOO_MANY_REQUESTS,
+                status.HTTP_503_SERVICE_UNAVAILABLE,
+            },
+            retry_exceptions=(httpx.RemoteProtocolError, httpx.ReadError),
```
iovsn, line 224: Hi.

Just saw your PR and tested it. In our case it didn't help until I updated the initial set of exceptions to the following:

```python
retry_exceptions=(httpx.RemoteProtocolError, httpx.ReadError, httpx.PoolTimeout, httpx.ReadTimeout)
```
With this update the workflow runs without any problems, but without it we still got crashing executions.

peytonrunyan: Awesome - thanks a ton for giving it a shot and following up. I'll get those errors added. 

@madkinsz 
Something that I'm wondering about is if we would possibly want to consider making these exceptions configurable by an environment variable or something similar so that users can adjust on the fly?

madkinsz: There have been requests for that, but let's consider it separately.

madkinsz: @peytonrunyan let's get this finished today and include the local protocol error in https://github.com/PrefectHQ/prefect/issues/7442

peytonrunyan:
@madkinsz thanks a ton for jumping in on this!

src/prefect/client/base.py
madkinsz: @jakekaplan want to give it a quick look again?

jakekaplan (approved):

PR #7603, opened 2022-11-21
Return 404 on missing flow run

zangell44:
If a flow run is not found when orchestrating a task run, ensure we return a 404. Previously, `MissingFlowRunError` was unhandled, resulting in a 500 error.

Closes https://github.com/PrefectHQ/nebula/issues/2685

### Example
```
async def test_set_task_run_state_returns_404_on_missing_flow_run(
        self, task_run, client, session
    ):
        await models.flow_runs.delete_flow_run(
            session=session, flow_run_id=task_run.flow_run_id
        )
        await session.commit()
        response = await client.post(
            f"/task_runs/{task_run.id}/set_state",
            json=dict(state=dict(type="RUNNING", name="Test State")),
        )
        assert response.status_code == status.HTTP_404_NOT_FOUND
```

### Checklist

### DIFF
--- a/src/prefect/exceptions.py
+++ b/src/prefect/exceptions.py
@@ -81,12 +81,6 @@ class MissingFlowError(PrefectException):
     """
 
 
-class MissingFlowRunError(PrefectException):
-    """
-    Raised when a flow run cannot be found.
-    """
-
-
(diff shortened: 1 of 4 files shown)


jakekaplan (approved):
LGTM 👍

Issue #7291, opened 2022-10-22
Add state message to the UI Flow Run page


zhen0:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently, the UI does not show state message information.  This info can be useful if a flow run is not picked up and can give context about why. 

### Describe the proposed behavior

Add a field in the flow run details tab to show the last state message

### Example Use

I have an agent with incorrect permissions to access where my flow code is stored.  The flow run fails and never gets to running so there are no logs.  The UI does not give useful information but if I look in the network tab and check out flow_run.state.message I get useful info.

zhen0:
<img width="1167" alt="Screen Shot 2022-10-25 at 12 28 04 PM" src="https://user-images.githubusercontent.com/40272060/197830638-9fc772f0-f15c-4b99-9253-694491833491.png">

Add this info to the flow run details tab / well.

zhen0:
Better example 
<img width="1673" alt="image" src="https://user-images.githubusercontent.com/40272060/197831345-6b410473-722f-411b-83d4-38b6866887fb.png">

Issue #774, opened 2019-03-21
Create test harness for new additions to the task library


cicdw:
There are a few very common test patterns for new tasks in the task library, such as: 
- does initialization work
- does initialization set attributes
- do additional kwargs get passed to the base task init method
- do certain attributes raise runtime errors if not provided
- can certain attributes be provided either at init or at runtime

We should create a test harness that addresses these generic tests and makes it easier for new additions to get tested without reproducing a bunch of boilerplate.

Clearly new additions will have unique tests as well, but this will help test the high-level functionality.

dshahid380:
Can I take this issue?

joshmeek:
Go right ahead! If you take a look at the tests for the kubernetes, google, aws, etc... tasks under `tests/tasks` it should become evident that there is a pattern emerging of testing the things that @cicdw initially outlined in this issue.

PR #7609, opened 2022-11-21
Updating prefect-design

nicolascribbles:
### Main info
- Closes [Orion-design Pull #774](https://github.com/PrefectHQ/orion-design/pull/774) and  #7291 
Needs updated prefect-design for styles to work

### Checklist

### DIFF
--- a/orion-ui/package.json
+++ b/orion-ui/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@prefecthq/orion-design": "1.1.6",
-    "@prefecthq/prefect-design": "1.1.29",
+    "@prefecthq/prefect-design": "^1.1.32",
     "@prefecthq/vue-compositions": "0.2.6",
     "@types/lodash.debounce": "4.0.7",
     "axios": "0.27.2",
--- a/orion-ui/package-lock.json
+++ b/orion-ui/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@prefecthq/orion-design": "1.1.6",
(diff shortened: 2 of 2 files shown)


orion-ui/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@prefecthq/orion-design": "1.1.6",
-    "@prefecthq/prefect-design": "1.1.29",
+    "@prefecthq/prefect-design": "^1.1.32",
```
marichka-offen, line 14: To check my own memory, some time ago there was a reason we weren't using `^` for prefecthq packages. Is it still a thing, or I'm misremembering? @pleek91 @znicholasbrown

znicholasbrown: Yes, definitely still a thing - this is to prevent unintentional version bumps that might contain breaking changes

pleek91: If its a `dependency` it should be exact. If its a `peer` it should have the `^` is the general rule of thumb for packages.

madkinsz:
Can you update the title to reflect the user-facing change so it'll appear in the changelog?

orion-ui/package.json
pleek91, line 14: @nicolascribbles Can you please run `npm i @prefecthq/prefect-design@latest --save-exact` and push that change? The `^` shouldn't be here since its a direct dependency. We only use the `^` for peer dependencies.

pleek91 (changes requested):

pleek91:
orion-ui already has prefect-design 1.1.32

PR #7611, opened 2022-11-21
Make agent set `infrastructure_pid` after starting a flow run

bunchesofdonald:
Updates the agent to set the `infrastructure_pid` after starting a flow run.

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -280,6 +280,16 @@ class OrionAgent:
                 )
             return exc
 
+        try:
+            await self.client.update_flow_run(
+                flow_run_id=flow_run.id, infrastructure_pid=result.identifier
+            )
+        except Exception as exc:
+            self.logger.exception(
+                "An error occured while setting the `infrastructure_pid` on "
+                f"flow run {flow_run.id!r}. The flow run will not be cancellable."
+            )
+
         if not task_status._future.done():
             self.logger.error(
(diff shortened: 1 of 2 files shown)


src/prefect/agent.py
```diff
@@ -280,6 +280,16 @@ async def _submit_run_and_capture_errors(
                 )
             return exc
 
+        try:
+            await self.client.update_flow_run(
+                flow_run_id=flow_run.id, infrastructure_pid=result.identifier
+            )
+        except Exception as exc:
+            self.logger.exception(
```
bunchesofdonald, line 288: I debated what to do if this fails and this seemed like the most 'user-friendly' thing to do at this point. When we have cancellation implemented it could make sense for this to do a kill on the infrastructure.

madkinsz: This makes sense to me. Maybe best to let it forge onward later too.

madkinsz: You are, however, calling this after `await infrastructure.run` so it will not update until the flow run completes. You need to update this when the task status is reported (the return value of `task_group.start`). We can split the infrastructure `run` into `start` and `wait` commands if this is helpful but I was hoping we wouldn't need to do that until after this raft.

bunchesofdonald:
This doesn't set the `pid` early enough in the process, I'm going to close this out and figure out another way.

PR #7620, opened 2022-11-22
Changing default hash_algo to sha512 instead of md5

### LINKED ISSUE #7615: MD5 not supported without usedforsecurity=False in FIPS 140-2 environment
### First check

### Bug summary

The md5 algorithm is compromised and forbidden in FIPS environments. 

The Prefect library explicitly uses hashlib.md5 in multiple locations, which makes it impossible to run Prefect 2.x with strict FIPS enforcement. However, Prefect version 1.x can function in strict FIPS environments.

Dask was using md5 for simple hashing (not security) which is permitted in FIPS environments, but it requires an extra parameter introduced in Python 3.9+; see:
- [#8608](https://github.com/dask/dask/issues/8608)
- [#8762](https://github.com/dask/dask/pull/8762)
- [#48024](https://github.com/pandas-dev/pandas/issues/48024)

The `hash_algo` is an input parameter to many of the methods in `utilities/hashing.py`, but it doesn't seem like there's any method to allow the user to override this with a different hashing algorithm.

### Reproduction

```python3
1. Install Prefect 2.x on a system with strict FIPS 140-2 enforcement
2. Attempt to run a simple Prefect Flow locally
```

### Error

```python3
/opt/python/lib/python3.10/site-packages/prefect/flows.py:619: in flow
    Flow(
/opt/python/lib/python3.10/site-packages/prefect/context.py:176: in __register_init__
/opt/python/lib/python3.10/site-packages/prefect/flows.py:158: in __init__
    version = file_hash(flow_file)
/opt/python/lib/python3.10/site-packages/prefect/utilities/hashing.py:41: in file_hash
    return stable_hash(contents, hash_algo=hash_algo)
/opt/python/lib/python3.10/site-packages/prefect/utilities/hashing.py:22: in stable_hash
    h = hash_algo()
E   ValueError: [digital envelope routines] disabled for fips
```

### Versions

```Text
Version:             2.6.8
API version:         0.8.3
Python version:      3.10.8
Git commit:          68044e28
Built:               Thu, Nov 17, 2022 3:19 PM
Profile:             default
Server:
  Database:          sqlite
  SQLite version:    3.39.4
```

### Additional context

Ideally, stop using md5 for hashing; but if required follow the example from Dask. Perhaps a sha512 hash would be a better default?

Pandas opted to use UUID instead for their hashing to get around a similar problem, but I don't think that will work with the object hashing Prefect is doing.

Or make it possible for the end-user to easily configure which `default` hashing algorithm is used across the `utilities/hashing.py` methods.

szelenka:
To allow Prefect to run in FIPS 140-2 environments, it cannot use the md5 algorithm. For Python 3.9+ it's possible to add an extra parameter to the md5 call, but there's no firm need to have this as md5 AFAIK. So adjusting this to sha512 will resolve this issue for Python 3.7+

closes #7615 

### Example
Changes are transparent to developer, simply changing the hash_algo from md5 to sha512

### Checklist

### DIFF
--- a/src/prefect/utilities/hashing.py
+++ b/src/prefect/utilities/hashing.py
@@ -7,7 +7,7 @@ import cloudpickle
 from prefect.serializers import JSONSerializer
 
 
-def stable_hash(*args: Union[str, bytes], hash_algo=hashlib.md5) -> str:
+def stable_hash(*args: Union[str, bytes], hash_algo=hashlib.sha512) -> str:
     """Given some arguments, produces a stable 64-bit hash of their contents.
 
     Supports bytes and strings. Strings will be UTF-8 encoded.
@@ -27,7 +27,7 @@ def stable_hash(*args: Union[str, bytes], hash_algo=hashlib.md5) -> str:
     return h.hexdigest()
 
 
-def file_hash(path: str, hash_algo=hashlib.md5) -> str:
+def file_hash(path: str, hash_algo=hashlib.sha512) -> str:
     """Given a path to a file, produces a stable hash of the file contents.
 
     Args:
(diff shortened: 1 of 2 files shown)


madkinsz:
Hi! Thanks for contributing, this is a little tricky as this is a breaking change. These hashes have meaning for existing users. Perhaps we can expose this as a setting?

szelenka:
@madkinsz would it be okay to copy the Dask approach? This would still use the md5 algorithm, but simply add the `usedforsecurity=False` parameter .. AFAIK, these don't appear to be used for security. Can you confirm?

madkinsz:
That seems reasonable! These are definitely not used for security.

madkinsz (approved):

PR #7626, opened 2022-11-22
Add flush of task run logs when on remote workers

madkinsz:
Users reported task run logs being dropped when using Dask w/ Slurm. Logs were recovered by adding a sleep to the end of the task, indicating an issue during teardown.

When running on distributed workers, tasks rely on an Orion log worker that is running in the distributed process. The log worker is flushed at the end of flow runs to ensure all logs are sent to the API before process exit, but workers on other processes will not be flushed. Here, we add a flush when task runs finish orchestration. This will ensure every task run's logs are sent to the API before a remote worker process exits. We do not perform this call for local task runs to retain efficient batching of logs.

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1105,7 +1105,7 @@ async def begin_task_run(
             ) from connect_error
 
         try:
-            return await orchestrate_task_run(
+            state = await orchestrate_task_run(
                 task=task,
                 task_run=task_run,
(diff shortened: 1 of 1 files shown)


madkinsz:
I'm not sure how to test this well, we do not have any remote task runners in this core library.

bunchesofdonald (approved):

PR #7629, opened 2022-11-22
Add flow run cancel command

peytonrunyan:
This pull request adds the command `prefect flow-run cancel <run_id>`. 

This CLI command will change the `state.StateType` of a given flow run to `CANCELLED` and the state name to `Cancelling` if possible, and otherwise will exit with error saying that it could not be cancelled. If it could not cancel, it will provide the reason from the abort status. 

There is also a change in the orion client, which changes `set_flow_run_state` to raise `ObjectNotFound` if we get a 404. This brings it in-line with the behavior of the rest of the functions in the orion client. A test has been added for this.

### Example

Screen Shot 2022-11-22 at 2 25 12 PM

### Checklist

### DIFF
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -128,9 +128,10 @@ async def cancel(id: UUID):
             )
         except ObjectNotFound as exc:
             exit_with_error(f"Flow run '{id}' not found!")
-    # breakpoint()
+
     if result.status == SetStateStatus.ABORT:
         exit_with_error(
             f"Flow run '{id}' was unable to be cancelled. Reason: '{result.details.reason}'"
         )
+
     exit_with_success(f"Flow run '{id}' was succcessfully scheduled for cancellation.")
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -17,8 +17,10 @@ from prefect.cli.root import app
 from prefect.client import get_client
 from prefect.exceptions import ObjectNotFound
 from prefect.orion.schemas.filters import FlowFilter, FlowRunFilter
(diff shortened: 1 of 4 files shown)


madkinsz:
We probably shouldn't merge this until cancellation is available.

src/prefect/cli/flow_run.py
```diff
@@ -113,3 +115,23 @@ async def delete(id: UUID):
...
+        try:
+            result = await client.set_flow_run_state(
+                flow_run_id=id, state=cancelling_state
+            )
+        except ObjectNotFound as exc:
+            exit_with_error(f"Flow run '{id}' not found!")
+
+    if result.status == SetStateStatus.ABORT:
+        exit_with_error(
+            f"Flow run '{id}' was unable to be cancelled. Reason: '{result.details.reason}'"
```
bunchesofdonald, line 134: Instead of wrapping things in quotes manually you can, generally, use `!r` in the format string, it can read a bit nicer:

```suggestion
            f"Flow run {id!r} was unable to be cancelled. Reason: {result.details.reason!r}"
```

tests/cli/test_flow_run.py
```diff
@@ -200,3 +212,66 @@ def test_ls_limit(
             found_count += 1
 
     assert found_count == 2
+
+
+class TestCancelFlowRun:
+    @pytest.mark.parametrize(
```
bunchesofdonald, line 218: :clap:

bunchesofdonald (approved):
This looks great!

src/prefect/cli/flow_run.py
peytonrunyan: Awesome - thanks!

PR #7635, opened 2022-11-22
Add kill method to Process infrastructure

bunchesofdonald:
This adds the `kill` method to the `Process` infrastructure.

- It makes some low-level changes to the way the process `pid` is communicated back to the Process infrastructure.
- Uses `loop.call_later` to facilitate sending a SIGKILL after a SIGTERM has been sent.


src/prefect/infrastructure/process.py
```diff
@@ -121,6 +143,24 @@ async def run(
...
+        def send_sigkill():
+            try:
+                os.kill(pid, signal.SIGKILL)
+            except ProcessLookupError:
+                # The process likely exited due to the SIGTERM.
+                pass
+
+        os.kill(pid, signal.SIGTERM)
+        loop = asyncio.get_running_loop()
+        loop.call_later(grace_seconds, send_sigkill)
```
madkinsz, line 162: Hmm, this should probably block until the process is killed so instead we should `await anyio.sleep(grace_seconds)` then send the SIGKILL? The agent can submit it to a task group to avoid blocking. We may need a polling interval to actually watch for the kill. You'd want to return early on that `ProcessLookupError`

src/prefect/utilities/processutils.py
```diff
@@ -55,6 +56,7 @@ async def run_process(
     command: List[str],
     stream_output: Union[bool, Tuple[Optional[TextSink], Optional[TextSink]]] = False,
     task_status: Optional[anyio.abc.TaskStatus] = None,
+    started_callback: Optional[Callable[[anyio.abc.Process], None]] = None,
```
madkinsz, line 59: Why were these changes necessary? Now we can't use `TaskGroup.start` with this utility.

bunchesofdonald: Because I wanted the hostname to be part of the `pid` sent back to the Agent, but I also wanted `Process` to be in charge of how that was formatted.

madkinsz: Ah I see. Let's just introduce a new task group in the `Process` then i.e.

```python
   pid = tg.start(run_process, ....)
   task_status.started(f"{hostname}:{pid}")
```

bunchesofdonald: I tried this route, but since `run` needs to block on waiting for the process starting up a new task group was untenable. I went with a modified approach of my initial commit that allows for a 'handler' for what's sent back via task_status.

madkinsz: Ah it'll block until it's complete (when the task group context exits) but yeah it's a pain to get the return value (`process`) with task groups without introducing a new utility. This seems reasonable for now, thanks for looking into it!

src/prefect/infrastructure/process.py
```diff
@@ -121,6 +141,27 @@ async def run(
             status_code=process.returncode, identifier=str(process.pid)
         )
 
+    async def kill(self, infrastructure_pid: str, grace_seconds: int):
+        hostname, pid = _parse_infrastructure_pid(infrastructure_pid)
+
+        if hostname != socket.gethostname():
+            # The process is running on a different host.
+            return
```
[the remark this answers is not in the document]
bunchesofdonald: :+1: Seems fair to move that handling up to the agent.

madkinsz (approved):

PR #7637, opened 2022-11-23
Blocking pause/resume

anticorrelator:
Starts work on #7410

This PR introduces the semantics necessary to pause a flow and allow it to be resumed out-of-process.

```python
from prefect import task, flow, pause, resume

@task
def foo():
    return 42

@flow
def bar():
    x = foo()
    pause()
    return x + foo()
    
bar()  # execution is blocked until resumed
resume(BAR_FLOW_RUN_ID)  # when called out of process, resumes flow execution
> 84
```

### Checklist


zangell44 (approved):
Looks clean to me!

For Postgres, I think we'll need an extra migration to add Paused to the enum value. Something like `ALTER TYPE state_type ADD VALUE 'Paused'`.

anticorrelator:
> Should we consider more verbose function names? e.g. `pause_flow_run`, `resume_flow_run`? We can always expose them as these simpler options later, but it seems safer to be explicit to start.

I like this idea

madkinsz:
@anticorrelator, not a big deal but please try to have titles be appropriate for the changelog so it doesn't need to be edited on release.

PR #7638, opened 2022-11-23
Add error to invoke and assert test utility when called from async function

peytonrunyan:
`invoke_and_assert` fails silently when called from an async main thread. This adds a check in the function which raises an exception if that happens. The exception message is formatted so that it's easy to read in pytest. 

I thought briefly about adding a test for it, but that feels possibly too meta. I'm down to write one if anybody feels like it's the right way to go. 

### Example

Screen Shot 2022-11-22 at 8 13 10 PM

### Checklist


src/prefect/testing/cli.py
```diff
@@ -71,7 +72,20 @@ def invoke_and_assert(
...
+            You cannot run `invoke_and_assert` directly from an async 
+            function. If you need to run `invoke_and_assert` in an async 
+            function, run it with `run_sync_in_worker_thread`. 
+            Example: 
+                run_sync_in_worker_thread(
+                    invoke_and_assert, 
+                    command=['my', 'command'],
+                    expected_code=0,
+                )
+            """
```
madkinsz, line 87: This indentation is pretty wonky. I'd use `dedent` if you want to use `"""`.

madkinsz (approved):

PR #7639, opened 2022-11-23
Guard against orchestration race conditions

zangell44:
Closes https://github.com/PrefectHQ/prefect/issues/7640

When multiple processes call /flow_runs/{id}/set_state around the same time, it is possible for both API calls to succeed. This can result in issues like two agents trying to run the same flow run.

### Checklist

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -147,6 +147,10 @@ async def load_flow_from_flow_run(
     If `ignore_storage=True` is provided, no pull from remote storage occurs.  This flag
     is largely for testing, and assumes the flow is already available locally.
     """
+    if flow_run.deployment_id is None:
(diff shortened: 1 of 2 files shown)


abrookins (commented):
@anticorrelator Pointed out that we probably can't do this after all due to the agent query also using a lock. Let's discuss tomorrow -- feel free to talk it through without me if that's easier, due to timezones!

madkinsz:
@anticorrelator  / @zangell44 what's the status of this?

anticorrelator:
@madkinsz I think we should close this, I don't remember the exact details but we greatly reduced the errors seen here in other ways

cc @zangell44

madkinsz:
👍 let me know if that changes

PR #7504, opened 2022-11-10
Add assertion for single notification and expand notification test suite

madkinsz:
Guarding against behavior where notifications are dropped silently, see https://prefect-community.slack.com/archives/CL09KU1K7/p1668119051343889

PR #7644, opened 2022-11-23
Use new session to send each notification

jlowin:
Curious for comment on this change -- currently we're using a single session to load and send every queued notification. This means that if anything happens to the session (due to misconfigured block or something), all subsequent notifications won't get sent because the session is rolled back. Notifications themselves do *not* use the session to send, so it's only the _loading_ of the notifications that presents a surface area for errors. This PR avoids this potentially long-lived session and opens a new session for each notification.


zangell44 (approved):
Excellent

madkinsz:
Safe to ignore the GitHub related failures in the service tests.

madkinsz:
Any hope that this will affect #7504 ?

PR #7648, opened 2022-11-23
Prompt workspace selection if API key is set but API URL is not set

madkinsz:
If the `PREFECT_API_URL` does not contain a valid workspace but the `PREFECT_API_KEY` is set during `prefect cloud login` re-authentication will fail during display of the current workspace.

### Example

Encountered in https://prefecthq.slack.com/archives/C029PBS5JVB/p1669233780065449

### Checklist


peytonrunyan (approved):
lgtm!

PR #7658, opened 2022-11-27
Add version_group_id property to flow view

AMirzaeiyan:
I need to create cloud hooks programmatically via API however currently, there is no clean way to pull out the `version_group_id` using the client. In my scenario, upon registering the flows I need to get the `group_version_id` and put it in the mutation request:

```
mutation {
  create_cloud_hook(
    input: {
      type: EMAIL,
      version_group_id: <= The key is mandatory,
      ...
      ) {
    id
  }
}
```

In order to address that issue I simply updated the flow query and introduced a new property in the codebase. That way we could simply call the function `from_flow_name` and get the result through the newly added property.

### Checklist

### DIFF
--- a/src/prefect/backend/flow.py
+++ b/src/prefect/backend/flow.py
@@ -34,6 +34,7 @@ class FlowView:
         - storage: The deserialized Storage object used to store this flow
         - name: The name of the flow
         - flow_group_labels: Labels that are assigned to the parent flow group
+        - version_group_id: The group version used to register the flow
     """
 
     flow_id: str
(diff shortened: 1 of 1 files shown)


madkinsz:
Hi! This makes sense to me. I'm curious why you're using the old "Cloud Hooks" system instead of the new "Automations" system? The `create_action` / `create_...._hook` system is more powerful and robust. For that, you'd need the `flow_group_id`

AMirzaeiyan:
Thanks for your suggestion, yes you're right, although currently, our system relies on cloud hooks, so for now, we prefer to stick with the current approach and avoid making further changes.

PR #7659, opened 2022-11-27
Fix link in 'Agents and Work Queues' documentation

brian-pond:
The previous page link "agent-configuration" was invalid, and did not work. I've substituted with "starting-an-agent", which should take the user further down the webpage.

image


tpdorsey (approved):
Good catch! Thank you

PR #7671, opened 2022-11-28
Add cancellation loop to agent

madkinsz:
### Example

Flow:
```python
import prefect
import time

@prefect.flow
def foo():
    print("hi!")
    time.sleep(1000)

if __name__ == "__main__":

    d = Deployment.build_from_flow(foo, name="test")
    d.apply()
```

Trigger:
```
❯ prefect deployment run foo/test
❯ prefect flow-run cancel 9344f984-f5b9-46c6-936c-2f6e44d2d291
16:18:49.350 | DEBUG   | prefect.client - Connecting to API at https://api.prefect.cloud/api/accounts/f6f508a4-f8a7-49a1-bb4c-4970a9d8662e/workspaces/897bd37e-a9aa-49ea-9f5e-75c6f6b2622d/
Flow run '9344f984-f5b9-46c6-936c-2f6e44d2d291' was succcessfully scheduled for cancellation.
```

Agent:
```
16:18:51.147 | DEBUG   | prefect.agent - Checking for cancelled flow runs...
16:18:51.330 | INFO    | prefect.agent - Found 1 flow runs awaiting cancellation.
16:18:51.597 | INFO    | prefect.agent - Killing process Michaels-MacBook-Pro.local:45276 for flow run '9344f984-f5b9-46c6-936c-2f6e44d2d291'...
16:18:53.086 | ERROR   | prefect.infrastructure.process - Process 'hospitable-coucal' exited with status code: -15; This indicates that the process exited due to a SIGTERM signal. Typically, this is caused by manual cancellation.
```

### Checklist


src/prefect/agent.py
```diff
@@ -177,6 +189,85 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+            for flow_run in cancelling_flow_runs:
+
+                # Avoid duplicate cancellation calls
+                if flow_run.id in self.cancelling_flow_run_ids:
+                    continue
+                else:
+                    self.cancelling_flow_run_ids.add(flow_run.id)
+
+                tg.start_soon(self.cancel_run, flow_run)
+
```
peytonrunyan, line 226: Purely a stylistic thing - is an explicit `if <whatever>: continue; else: <do stuff>` preferred over something like

```python
if flow_run.id not in self.cancelling_flow_run_ids:
    self.cancelling_flow_run_ids.add(flow_run.id)
    tg.start_soon(self.cancel_run, flow_run)
```

src/prefect/agent.py
```diff
@@ -177,6 +189,85 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+        except Exception:
+            self.logger.exception(
+                f"Failed to get infrastructure for flow run '{flow_run.id}'. "
+                "Flow run cannot be cancelled."
+            )
+            # Note: We leave this flow run in the cancelling set because it cannot be
+            #       cancelled and this will prevent additional attempts.
+            return
+
+        async def mark_flow_run_as_cancelled():
```
peytonrunyan, line 244: Why this as an internal function instead of a private method?

src/prefect/agent.py
```diff
@@ -177,6 +189,85 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+            await self.client.set_flow_run_state(flow_run.id, state, force=True)
+            self.cancelling_flow_run_ids.remove(flow_run.id)
+
+        self.logger.info(
+            f"Killing {infrastructure.type} {flow_run.infrastructure_pid} for flow run '{flow_run.id}'..."
+        )
+        try:
+            await infrastructure.kill(flow_run.infrastructure_pid)
+        except InfrastructureNotFound as exc:
+            self.logger.warning(f"{exc} Marking flow run as cancelled.")
```
peytonrunyan, line 255: Should we mention that the infra wasn't found?

src/prefect/agent.py
madkinsz: We're not using it anywhere else yet and it's very simple, seemed simplest to just have it inline for now.

src/prefect/agent.py
madkinsz: Yeah this continue else clause is pretty silly. Usually I'll just exclude the `else` here...

```python
                # Avoid duplicate cancellation calls
                if flow_run.id in self.cancelling_flow_run_ids:
                    continue

                self.cancelling_flow_run_ids.add(flow_run.id)
                tg.start_soon(self.cancel_run, flow_run)
```

I think I prefer that because it reduces nesting and makes the comment really clear

src/prefect/agent.py
```diff
@@ -177,6 +189,85 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+        async def mark_flow_run_as_cancelled():
+            state = flow_run.state.copy(update={"name": "Cancelled"})
+            await self.client.set_flow_run_state(flow_run.id, state, force=True)
+            self.cancelling_flow_run_ids.remove(flow_run.id)
+
+        self.logger.info(
+            f"Killing {infrastructure.type} {flow_run.infrastructure_pid} for flow run '{flow_run.id}'..."
+        )
+        try:
+            await infrastructure.kill(flow_run.infrastructure_pid)
```
bunchesofdonald, line 253: Should we handle the case that `infrastructure_pid` is null?

bunchesofdonald (approved):

peytonrunyan (approved):
Left a couple small comments, otherwise lgtm!

src/prefect/agent.py
```diff
@@ -177,6 +189,85 @@ async def get_and_submit_flow_runs(self) -> List[FlowRun]:
...
+        self.logger.debug("Checking for cancelled flow runs...")
+
+        work_queue_names = set()
+        async for work_queue in self.get_work_queues():
+            work_queue_names.add(work_queue.name)
+
+        cancelling_flow_runs = await self.client.read_flow_runs(
+            flow_run_filter=FlowRunFilter(
+                state=FlowRunFilterState(
+                    name=FlowRunFilterStateName(any_=["Cancelling"])
```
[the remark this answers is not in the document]
zangell44: Yes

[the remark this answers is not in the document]
madkinsz: I added a state type filter here too.

src/prefect/agent.py
madkinsz: Yeah let's handle that and do a `hasattr` check for `kill`.

madkinsz: On second thought, we should maybe handle this during the proposal of a cancelled state? We can add a `kill-infrastructure` capability to infrastructure blocks that support killing and block cancellation when it is not supported?

madkinsz: I added these in https://github.com/PrefectHQ/prefect/pull/7671/commits/52f5565392b3d33257a373399a00a18343c10360 and we mark the run as cancelled so we don't pick it up forever. Kind of an awkward situation.

PR #7678, opened 2022-11-29
Fix bug in `StateGroup` that caused `all_final` to be wrong

### LINKED ISSUE #7677: StateGroup.all_final() is looking at non-final states and not final states.
### First check

### Bug summary

`StateGroup.all_final()` is using the count of not-final states rather than the count of final states resulting in incorrect messaging when a Flow finishes with all final states.

### Reproduction

```python3
prefect import flow, task
from prefect.orion.schemas.states import Cancelled


@task
def skippy(x: bool = False):
    return Cancelled(message="Skip it!")


@flow
def my_flow():
    return skippy()


if __name__ == "__main__":
    my_flow()


The end state of the flow is `Failed('0/1 states are not final.')` but `Cancelled` is a final state so this should be `Failed("Given states: total=1, 'CANCELLED'=1")`
```

### Error

_No response_

### Versions

```Text
Latest `main`
```

### Additional context

_No response_

bunchesofdonald:
While looking at how we're handling `Cancelled` states returned by tasks I noticed that there was a bug in `StateGroup` that caused `all_final` to return the opposite of what it should. This fixes that up and adds tests for the class as a whole.

Closes #7677

### DIFF
--- a/src/prefect/states.py
+++ b/src/prefect/states.py
@@ -371,7 +371,8 @@ class StateGroup:
         self.states = states
         self.type_counts = self._get_type_counts(states)
         self.total_count = len(states)
-        self.not_final_count = self._get_not_final_count(states)
+        self.final_count = sum(state.is_final() for state in states)
+        self.not_final_count = self.total_count - self.final_count
 
     @property
     def fail_count(self):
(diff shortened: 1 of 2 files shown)


madkinsz (approved):
nice!

PR #7666, opened 2022-11-28
Tracking: Flow run cancellation

madkinsz:
Includes:

- #7629
- #7635 
- #7671 
- #7684 
- #7701 
- https://github.com/PrefectHQ/prefect/pull/7672

Closes:

- https://github.com/PrefectHQ/prefect/issues/7654

PR #7684, opened 2022-11-29
Add kill docker method

peytonrunyan:
This PR adds the `kill` method for Docker containers as part of #7666

### DIFF
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -7,6 +7,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Dict, Generator, List, Optional, Tuple
 
 import anyio.abc
+import docker
 import packaging.version
 from pydantic import Field, validator
 from typing_extensions import Literal
@@ -14,6 +15,7 @@ from typing_extensions import Literal
 import prefect
 from prefect.blocks.core import Block, SecretStr
 from prefect.docker import get_prefect_image_name, parse_image_tag
+from prefect.exceptions import InfrastructureNotAvailable, InfrastructureNotFound
 from prefect.infrastructure.base import Infrastructure, InfrastructureResult
 from prefect.settings import PREFECT_API_URL
 from prefect.utilities.asyncutils import run_sync_in_worker_thread, sync_compatible
@@ -263,11 +265,13 @@ class DockerContainer(Infrastructure):
 
         # The `docker` library uses requests instead of an async http library so it must
         # be run in a thread to avoid blocking the event loop.
+        client = self._get_client()
+        self.logger.warning(client.api.base_url)
         container = await run_sync_in_worker_thread(self._create_and_start_container)
 
-        # Mark as started and return the container id
+        # Mark as started and return the infrastructure id
         if task_status:
-            task_status.started(container.id)
+            task_status.started(self._get_infrastructure_pid(container.id))
 
         # Monitor the container
         container = await run_sync_in_worker_thread(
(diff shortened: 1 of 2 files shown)


bunchesofdonald (commented):
Lookin' good!

src/prefect/infrastructure/docker.py
```diff
@@ -288,6 +312,22 @@ def preview(self):
         finally:
             docker_client.close()
 
+    def _get_infrastructure_pid(self, container_id: str) -> str:
+        """Generates a docker infrastructure_pid string in the form of
+        `<docker_host_base_url>:<container_id>`.
+        """
+        docker_client = self._get_client()
+        base_url = docker_client.api.base_url
```
madkinsz, line 320: Do Docker client urls ever have secrets in them?

tests/infrastructure/test_docker_container.py
```diff
@@ -87,6 +101,49 @@ def test_name_cast_to_valid_container_name(
...
+
+    mock_docker_client.containers.get.return_value.stop.assert_called_with(
+        timeout=GRACE_SECONDS
+    )
+
+
+async def test_kill_raises_infra_not_available_with_bad_host_url(mock_docker_client):
+    BAD_BASE_URL = "bad-base-url"
+    with pytest.raises(
+        InfrastructureNotAvailable,
```
madkinsz, line 127: I'd add a `match` here too.

tests/infrastructure/test_docker_container.py
```diff
@@ -56,15 +60,25 @@ def mock_docker_client(monkeypatch):
...
+    mock.api = fake_api
+
     monkeypatch.setattr("docker.from_env", MagicMock(return_value=mock))
     return mock
 
 
+@pytest.fixture
+def client_container_not_found(mock_docker_client):
+    mock_docker_client.containers.get.side_effect = [docker.errors.NotFound("msg")]
+
```
madkinsz, line 80: I'm not sure it's worth moving this out into a fixture if it's a single line and only used once.

src/prefect/infrastructure/docker.py
```diff
@@ -277,9 +279,31 @@ async def run(
...
+            identifier=self._get_infrastructure_pid(container.id),
         )
 
+    async def kill(self, infrastructure_pid: str, grace_seconds: int):
+        docker_client = self._get_client()
+        base_url, container_id = self._parse_infrastructure_pid(infrastructure_pid)
+
+        if docker_client.api.base_url != base_url:
+            raise InfrastructureNotAvailable(
+                f"Unable to stop container {container_id!r}: the current docker api ",
```
madkinsz, line 291: Let's capitalize "Docker" and "API" throughout.

src/prefect/infrastructure/docker.py
```diff
@@ -277,9 +279,31 @@ async def run(
...
+                f"base_url {docker_client.api.base_url!r} does not match the expected ",
+                f"api base_url {base_url}.",
```
madkinsz, line 293: I try not to use variable names in messages e.g. "api base_url" should just be "API URL"

src/prefect/infrastructure/docker.py
```diff
@@ -288,6 +312,22 @@ def preview(self):
         finally:
             docker_client.close()
 
+    def _get_infrastructure_pid(self, container_id: str) -> str:
+        """Generates a docker infrastructure_pid string in the form of
+        `<docker_host_base_url>:<container_id>`.
+        """
+        docker_client = self._get_client()
```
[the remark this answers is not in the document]
peytonrunyan: Yeah, I thought about it a bit, but it's a bit awkward. It's unclear when it makes the most sense to instantiate a client for sharing. 

`run` does not instantiate a client, but calls `_create_and_start_container` and `_watch_container_safe`

`_create_and_start_container` instantiates a client (either from a registry or the regular client), passes it to 
`_build_container_settings`
`_should_pull_image`  
`_pull_image`
`_create_container`
and then closes it.

`_watch_container_safe` instantiates a client and passes it to `_watch_container` and then closes it.

`preview` instantiates and then closes

I thought about setting a property for `self.client` and then instantiating it if it's `None` or closed, but there's not an obvious way to check for whether or not the connection is closed without hitting what would probably be a law of demeter violation.

What I definitely should do if nothing else is not call the function twice inside of run.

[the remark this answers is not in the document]
peytonrunyan: https://github.com/PrefectHQ/prefect/pull/7684/commits/f03d189d5ac42bc1bea255b17a3ee85625dc9cd7

src/prefect/infrastructure/docker.py
peytonrunyan: Ooph. I have no clue. I'll see if somebody on the platform team might have a better idea.

madkinsz: tldr; probably not.

PR #7703, opened 2022-12-01
Increase default agent query interval to 10s

madkinsz:
5s is aggressive and can increase load on agent infrastructure.


billpalombi (approved):
I don't have any concerns with this

chrisguidry (approved):
:bow:

zangell44 (approved):

PR #7704, opened 2022-12-01
Update `run_sync_in_interruptible_worker_thread` to use an event

madkinsz:
I'm working on threadsafe async events in the engine reliability raft and have learned some magic. Here, I use an async event in a threadsafe way which should improve performance as we are no longer running a busy loop on the event loop. We only run tasks in interruptible threads when timeouts are set right now, but to support #7150 we will want sync tasks to be cancellable by default.

### DIFF
--- a/src/prefect/utilities/asyncutils.py
+++ b/src/prefect/utilities/asyncutils.py
@@ -1,6 +1,7 @@
 """
 Utilities for interoperability with async functions and workers from various contexts.
 """
+import asyncio
 import ctypes
 import inspect
 import threading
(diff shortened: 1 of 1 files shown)


src/prefect/utilities/asyncutils.py
```diff
@@ -109,13 +112,14 @@ def capture_worker_thread_and_result():
         except BaseException as exc:
             result = exc
             raise
+        finally:
+            loop.call_soon_threadsafe(event.set)
```
madkinsz, line 116: tldr; async events are not threadsafe but you can call `set` via a threadsafe loop method and it's all great.

bunchesofdonald (approved):
Yay!

PR #7713, opened 2022-12-01
chore: update CodeQL settings

jawnsy:
* Run static code analysis in pull requests whenever code changes, and upload reports to CodeQL
* Add a schedule to run the scans weekly on Sunday at 20:30 UTC
* Cancel in-progress scans for pull requests, and serialize scans on main branch
* Do not install Python dependencies prior to scan

### Checklist

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,4 +1,4 @@
-name: CodeQL
+name: "CodeQL"
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -1,27 +1,61 @@
...
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │  ┌───────────── hour (0 - 23)
+    #        │  │  ┌───────────── day of the month (1 - 31)
+    #        │  │  │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │  │  │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │  │  │ │ │
+    #        │  │  │ │ │
+    #        │  │  │ │ │
+    - cron: "30 20 * * 0"
```
madkinsz, line 30: cyuuute

.github/workflows/codeql-analysis.yml
```diff
@@ -1,27 +1,61 @@
...
       - main
+    paths:
+      - .github/workflows/codeql-analysis.yaml
+      - "**/*.py"
+      - "**/*.js"
+      - "**/*.ts"
+      - requirements.txt
+      - requirements-dev.txt
+      - setup.cfg
+      - alembic.ini
```
madkinsz, line 19: We did not do this before because it's super slow and blocks rapid development — do we need to run this on every pull request?

madkinsz: Perhaps it's faster now?

jawnsy: we don't _need_ to, but it's convenient because it'll show changes when people are making them, and flag new issues -- it's also not a required check so shouldn't block merges (people can merge without waiting for the analysis to finish, but if they happen to have a long-open PR then they'll benefit from the checks)

madkinsz: 👍 alrighty

madkinsz (approved):

.github/workflows/codeql-analysis.yml
jawnsy: also if it becomes a problem again we can always adjust this; no biggie!

Issue #7692, opened 2022-11-30
Eager conversion of generator tasks into lists not working in version 2.6+


sveinugu:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Automatic conversion of generator tasks into lists stopped working in version 2.6.0. A simple task yielding the first five integers from the `range()` builtin (see code below) returned `[0, 1, 2, 3, 4]` prior to 2.6.0. In version 2.6.0, the same code returns `None`. From 2.6.1 and onwards, the code returns `[]`.

I believe this is a bug. If this is a deprecated feature, it should not fail silently, possibly causing hard-to-find bugs in code that relies on this to work.

### Reproduction

```python3
from prefect import task, flow

@task
def my_range(i: int):
    for i in range(i):
        yield i

@flow
def my_flow():
    return my_range(5)

print(my_flow())
```

### Error

```python3
# Prefect v2.5.0

05:50:31.671 | INFO    | prefect.engine - Created flow run 'foamy-mackerel' for flow 'my-flow'
05:50:31.783 | INFO    | Flow run 'foamy-mackerel' - Created task run 'my_task-20c6ece6-0' for task 'my_task'
05:50:31.784 | INFO    | Flow run 'foamy-mackerel' - Executing 'my_task-20c6ece6-0' immediately...
05:50:31.811 | INFO    | Task run 'my_task-20c6ece6-0' - Finished in state Completed()
05:50:31.821 | INFO    | Flow run 'foamy-mackerel' - Finished in state Completed()
[0, 1, 2, 3, 4]

# Prefect v2.6.0

05:52:13.256 | INFO    | prefect.engine - Created flow run 'arrogant-termite' for flow 'my-flow'
05:52:13.331 | INFO    | Flow run 'arrogant-termite' - Created task run 'my_task-20c6ece6-0' for task 'my_task'
05:52:13.331 | INFO    | Flow run 'arrogant-termite' - Executing 'my_task-20c6ece6-0' immediately...
05:52:13.356 | INFO    | Task run 'my_task-20c6ece6-0' - Finished in state Completed()
05:52:13.365 | INFO    | Flow run 'arrogant-termite' - Finished in state Completed()
None

# Prefect v2.6.1

[...]

PR #7714, opened 2022-12-01
Keep data from being lost when assigning a generator to `State.data`

bunchesofdonald:
As I outlined in the [pydantic issue](https://github.com/pydantic/pydantic/issues/4807), when a generator is assigned to `State.data` the first few items in the generator are lost as pydantic iterates the generator during assignment. This adds `Sequence` as the first data type which keeps the data from being lost.

### DIFF
--- a/src/prefect/client/schemas.py
+++ b/src/prefect/client/schemas.py
@@ -1,4 +1,13 @@
-from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar, Union, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Generic,
+    Optional,
+    Sequence,
+    TypeVar,
+    Union,
+    overload,
+)
 from uuid import UUID
 
 from pydantic import Field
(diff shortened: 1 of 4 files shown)


bunchesofdonald:
This is related to #7692 but doesn't fully solve the issue.

madkinsz:
@bunchesofdonald this is wild haha — where does it fall short saving the issue? Should we just cast generators to lists in `return_value_to_state` instead of changing the annotation here?

bunchesofdonald:
Haha, right? I considered going down that route, this seemed less _invasive_ for some reason, but casting it to a list would likely solve 7692 in total. I can see what that looks like.

bunchesofdonald:
Converting the generator to a list does work, however we end up with kind of an odd situation where the task is returning a generator, but the value received by the caller is a list which means you can't use things like `next` that expect a generator object.

madkinsz:
Generators can't cross process boundaries anyway, so they'd break for other task runners. I'm not sure we should support generator state data. I'm not even sure what it'd mean for persistence too.

madkinsz:
@bunchesofdonald can you get this one across the line?

bunchesofdonald:
Yep thanks for calling it out @madkinsz!

madkinsz (approved):
Thanks!

madkinsz:
🤯 now I've got to release this too!

PR #7738, opened 2022-12-02
Implement non-blocking style pauses

anticorrelator:
Continues work on #7410

In addition to blocking-style pauses that keep infrastructure running, here we implement a flag that enables a deployed flow to gracefully exit once paused. If the flow run is resumed (either via the CLI or UI) it will be put back in a scheduled state so an agent can pick it up to resume execution.

### Checklist


src/prefect/exceptions.py
```diff
@@ -349,3 +349,7 @@ class NotPausedError(PrefectException):
 
 class FlowPauseTimeout(PrefectException):
     """Raised when a flow pause times out"""
+
+
+class FlowPauseExit(PrefectException):
```
cicdw, line 354: Why did you decide to use exception raising vs. `return` statements here?

anticorrelator: The engine needs a signal to gracefully stop execution immediately, so I decided to raise an exception that's caught instead of waiting until the next task. The exception makes the downstream handling pretty convenient, e.g. knowing it's okay to leave the flow without a terminal state etc

madkinsz: Without an exception, the user would need to check the return value of the pause function and exit themselves to prevent additional code from being executed.

src/prefect/orion/api/flow_runs.py
```diff
@@ -175,6 +175,78 @@ async def read_flow_run_graph(
...
+                details=schemas.responses.StateAbortDetails(
+                    reason="Cannot resume a flow run that is not paused."
+                ),
+            )
+            return result
+
+        orchestration_parameters.update({"api-version": api_version})
+
+        if state.state_details.pause_reschedule:
+            if not flow_run.deployment_id:
```
zangell44, line 211: Is this check redundant with orchestration rules for handling paused states?

src/prefect/orion/orchestration/global_policy.py
```diff
@@ -159,7 +166,48 @@ async def before_transition(self, context: OrchestrationContext) -> None:
...
 
 
-class IncrementRunCount(BaseUniversalTransform):
+class IncrementFlowRunCount(BaseUniversalTransform):
+    """
+    Records the number of times a run enters a running state. For use with retries.
+    """
+
+    async def before_transition(self, context: OrchestrationContext) -> None:
+        if self.nullified_transition():
```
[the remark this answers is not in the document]
anticorrelator: It used to be default behavior but was removed for some cases, e.g. we want them to fire when instrumenting some kinds of transitions even if they are aborted

zangell44 (approved):
LGTM! A few minor comments

src/prefect/exceptions.py
```diff
@@ -349,3 +349,7 @@ class NotPausedError(PrefectException):
 
 class FlowPauseTimeout(PrefectException):
     """Raised when a flow pause times out"""
+
+
+class FlowPauseExit(BaseException):
```
madkinsz, line 354: Should we name this `PausedRun` like we have `CrashedRun` and `FailedRun`?

src/prefect/engine.py
```diff
@@ -717,21 +734,39 @@ async def pause_flow_run(timeout: int = 300, poll_interval: int = 10):
...
-    client = get_client()
     response = await client.set_flow_run_state(
         frc.flow_run.id,
-        Paused(),
+        Paused(timeout_seconds=timeout, reschedule=reschedule, pause_key=pause_key),
     )
 
+    if response.status == SetStateStatus.ABORT:
+        if response.details.reason == "This pause has already fired.":
+            return
```
madkinsz, line 764: 😬 Can we have this be a REJECT with RUNNING instead of an ABORT? I'd rather not perform logic based on the text in the reason and I think an ABORT should always signal an exit of the process.

src/prefect/engine.py
```diff
@@ -772,11 +807,13 @@ async def resume_flow_run(flow_run_id):
...
         raise NotPausedError("Cannot resume a run that isn't paused!")
 
-    await client.set_flow_run_state(
+    response = await client.resume_flow_run(
         flow_run_id,
-        Running(name="Resuming"),
     )
 
+    if response.status == SetStateStatus.REJECT:
+        raise FlowPauseTimeout("Flow run can no longer be resumed.")
```
madkinsz, line 815: Per the above on ABORT, this could be a REJECT with a FAILED state to distinguish.

src/prefect/orion/api/flow_runs.py
```diff
@@ -175,6 +175,68 @@ async def read_flow_run_graph(
...
+    api_version=Depends(dependencies.provide_request_api_version),
+) -> OrchestrationResult:
+    """
+    Resume a paused flow run.
+    """
+    async with db.session_context(begin_transaction=True) as session:
+        flow_run = await models.flow_runs.read_flow_run(session, flow_run_id)
+        state = flow_run.state
+
+        if state is None or state.type != "PAUSED":
```
madkinsz, line 198: Can we use the state type enum here?

src/prefect/orion/api/flow_runs.py
```diff
@@ -175,6 +175,68 @@ async def read_flow_run_graph(
...
+            result = OrchestrationResult(
+                state=None,
+                status=schemas.responses.SetStateStatus.ABORT,
+                details=schemas.responses.StateAbortDetails(
+                    reason="Cannot resume a flow run that is not paused."
+                ),
```
madkinsz, line 204: Here I think we'd want to return a `REJECT` with the current state of the run?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -308,6 +311,14 @@ async def before_transition(
                 # Reset the run count so that the task run retries still work correctly
                 run.run_count = 0
 
+        # Reset pause metadata on retry
+        api_version = context.parameters.get("api-version", None)
+        if api_version is None or api_version >= Version("0.8.4"):
```
madkinsz, line 316: Can you add a comment about why we are doing different behavior based on version here?

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -429,6 +440,106 @@ async def before_transition(
...
+    async def before_transition(
+        self,
+        initial_state: Optional[states.State],
+        proposed_state: Optional[states.State],
+        context: TaskOrchestrationContext,
+    ) -> None:
+        if initial_state is None or not initial_state.is_running():
+            await self.abort_transition(
+                "Cannot pause flows that are not currently running."
+            )
```
madkinsz, line 460: I think all of these should be REJECT not ABORT as we are not requesting that the client process to _exit_ just telling them they cannot make the requested transition.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -529,16 +640,34 @@ async def before_transition(
         proposed_state: Optional[states.State],
         context: FlowOrchestrationContext,
     ) -> None:
+        self.original_flow_policy = context.run.empirical_policy.dict()
```
madkinsz, line 643: Not blocking for this, but is this the recommend practice for storing state during a rule?

anticorrelator: I don't know about recommended, but the rule needs to remember the state specifically for cleanup so it's how all of the rules that implement a stateful cleanup do it

src/prefect/engine.py
anticorrelator: regarding this stuff, I was going to refine this in a future PR with a "CONTINUE" action

src/prefect/states.py
```diff
@@ -495,13 +500,34 @@ def Pending(cls: Type[State] = State, **kwargs) -> State:
     return schemas.states.Pending(cls=cls, **kwargs)
 
 
-def Paused(cls: Type[State] = State, **kwargs) -> State:
+def Paused(
+    cls: Type[State] = State,
+    timeout_seconds: int = None,
+    pause_timeout: datetime.datetime = None,
```
madkinsz, line 506: It's pretty hard to understand the difference between these without reading the implementation. Can we use clearer names?

src/prefect/orion/api/flow_runs.py
anticorrelator: I think this really is an `ABORT`, the process asking for this transition is either some utility, a UI interface or CLI command. Those things can't do anything with the current state and can only pause or error out

src/prefect/engine.py
anticorrelator: this has been changed to be a reject, that matches much better to what we want to see happen and is distinct by more than a message

src/prefect/orion/orchestration/core_policy.py
anticorrelator: I'll mirror my comment above, since pausing isn't in the "primary orchestration flow" and not something the engine receives, I do think the "pauser" or "resumer" should be get an abort signal back, there's nothing else to do

src/prefect/engine.py
```diff
@@ -702,7 +717,9 @@ async def orchestrate_flow_run(
 
 
 @sync_compatible
-async def pause_flow_run(timeout: int = 300, poll_interval: int = 10):
+async def pause_flow_run(
+    timeout: int = 300, poll_interval: int = 10, reschedule=False, key: str = None
+):
     """
     Pauses a flow run by stopping execution until resumed.
```
madkinsz, line 724:
```suggestion
    Pauses the current flow run by stopping execution until resumed.
```

jakekaplan (approved):
Looks good to me fwiw!

zangell44 (approved):

src/prefect/engine.py
```diff
@@ -717,21 +734,41 @@ async def pause_flow_run(timeout: int = 300, poll_interval: int = 10):
...
-        Paused(),
+        Paused(timeout_seconds=timeout, reschedule=reschedule, pause_key=pause_key),
     )
 
+    if response.status == SetStateStatus.REJECT:
+        if response.state.type == StateType.RUNNING:
+            return
+        raise RuntimeError(response.details.reason)
+    elif response.status == SetStateStatus.ABORT:
+        raise RuntimeError(response.details.reason)
```
madkinsz, line 767: I think we should refactor this to use `propose_state`, but I'm happy to poke at that in a follow-up.

madkinsz (approved):

madkinsz:
I don't understand these flaky tests or why they're only a problem on this PR — I'm a bit concerned about that but we can look into the failing bin tests separately if we need to.

Issue #7739, opened 2022-12-02
Agent fails to cancel kubernetes jobs 


BitTheByte:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Agent fails while canceling running Kubernetes jobs with `ConfigException` as there is no `~/.kube/config` file will be present or any cluster context information if the job is running inside a Kubernetes job

### Reproduction

```python3
-
```

### Error

```python3
01:52:27.718 | INFO    | prefect.agent - Found 1 flow runs awaiting cancellation.
01:52:28.155 | INFO    | prefect.agent - Killing kubernetes-job in-cluster-config:ingenious-oxpecker-r9vr4 for flow run '1d44c745-6d8d-4b3a-a338-6d89117c7da4'...
01:52:28.160 | ERROR   | prefect.agent - Encountered exception while killing infrastructure for flow run '1d44c745-6d8d-4b3a-a338-6d89117c7da4'. Flow run may not be cancelled.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/prefect/agent.py", line 285, in 
cancel_run
    await infrastructure.kill(flow_run.infrastructure_pid)
  File 
"/usr/local/lib/python3.10/dist-packages/prefect/infrastructure/kubernetes.py", 
line 282, in kill
    current_cluster = self._get_active_cluster_name()
  File 
"/usr/local/lib/python3.10/dist-packages/prefect/infrastructure/kubernetes.py", 
line 347, in _get_active_cluster_name
    _, active_context = kubernetes.config.list_kube_config_contexts()
  File 
"/usr/local/lib/python3.10/dist-packages/kubernetes/config/kube_config.py", line
789, in list_kube_config_contexts
    loader = _get_kube_config_loader(filename=config_file)
  File 
"/usr/local/lib/python3.10/dist-packages/kubernetes/config/kube_config.py", line
770, in _get_kube_config_loader
    raise ConfigException(
kubernetes.config.config_exception.ConfigException: Invalid kube-config file. No
configuration found.
```

### Versions

```Text
Version:             2.7.0
API version:         0.8.3
Python version:      3.10.8
Git commit:          c2833339
Built:               Thu, Dec 1, 2022 4:03 PM
OS/Arch:             linux/x86_64
[...]

PR #7747, opened 2022-12-02
Use cluster uid and namespace instead of cluster "name" for Kubernetes job identifiers

madkinsz:
Co-authored with @peytonrunyan 
Closes https://github.com/PrefectHQ/prefect/issues/7739

Converts Kubernetes job PIDs from `{cluster_name}:{job_name}` to `{cluster_uid}:{namespace}:{job_name}`. Cluster names can be arbitrarily set in the Kubernetes config and are not a reliable unique identifier. Additionally, the config does not exist when using an in-cluster agent or in cases like #7739. We generate a stable cluster UID from the `kube-system` namespace UID as described in https://github.com/kubernetes/kubernetes/issues/44954 — which tracks the lack of a Kubernetes API for retrieving a cluster identifier. I attempted to use the job UID instead of the job name because it is guaranteed to be unique across namespaces. However, there is not a clear way to look up a job by its UID later without listing all jobs. Instead, we include the namespace in the PID to ensure we do not kill jobs in the wrong namespace.

Additionally, in #7701 the `KubernetesJobResult` returns were not updated to include the new identifier. Here, I do some minor refactoring to fix this.

### Example

 
 ```python
 from prefect.infrastructure import KubernetesJob
import anyio


async def main():
        job = KubernetesJob(
            command=["sleep", "1000"], pod_watch_timeout_seconds=5
        )

        pid = await tg.start(job.run)
        print("PID:", pid)

        await job.kill(pid)
        print("Killed :)")


anyio.run(main)
```

### Checklist

### DIFF
--- a/docs/concepts/tasks.md
+++ b/docs/concepts/tasks.md
@@ -221,7 +221,7 @@ A real-world example might include the flow run ID from the context in the cache
 
 ```python
 def cache_within_flow_run(context, parameters):
-    return f"{context.flow_run_id}-{task_input_hash(context, parameters)}"
+    return f"{context.task_run.flow_run_id}-{task_input_hash(context, parameters)}"
 
 @task(cache_key_fn=cache_within_flow_run)
 def cached_task():
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -260,7 +260,7 @@ class KubernetesJob(Infrastructure):
     async def run(
         self,
         task_status: Optional[anyio.abc.TaskStatus] = None,
-    ) -> Optional[bool]:
+    ) -> KubernetesJobResult:
         if not self.command:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by bunchesofdonald — APPROVED** [review]
This looks great! Using the namespace UID sure seems like the right way forward.

**Comment by scottwalshqs** [comment]
FYI, this change appears to have created a dependency that the agent must now have access to the kube-system namespace in order to get the cluster UID. 

```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"namespaces \"kube-system\" is forbidden: User \"system:serviceaccount:default:default\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"kube-system\"","reason":"Forbidden","details":{"name":"kube-system","kind":"namespaces"},"code":403}
```

PR #7770, opened 2022-12-05
Fix scheduled time copy for flows

madkinsz:
Fixes https://github.com/PrefectHQ/prefect/pull/7585 by adding the rule to the flow run policy.

### Checklist


madkinsz:
@anticorrelator I'm pretty concerned the tests I added passed before. Do you know why?

anticorrelator:
> @anticorrelator I'm pretty concerned the tests I added passed before. Do you know why?

Ah yeah, I missed that too. The tests added before simulate orchestration but don't run it. A full cycle of orchestration requires testing at the API or model level

madkinsz:
@anticorrelator can you review again please? cc @bunchesofdonald worth knowing about this gotcha as well.

madkinsz:
@anticorrelator it seems nice to be consistent with copying the metadata, I can see us scheduling tasks to run at some time in the future and wanting the same presubmit behavior. I can remove it if you prefer though.

anticorrelator (approved):

madkinsz:
I think since we have `WaitForScheduledTime` on tasks we should keep it for now. We could remove both of those if we wanted, tasks don't have scheduled times right now.

PR #7641, opened 2022-11-23
Added params to run deployment for  #7237 and #7522

YtKC:
Exposes the `tags` and `idempotency_key` parameters within the `run_deployment` method. Closes #7237 and #7522.

### Example

Example use of `run_deployment`:

```
await run_deployment(
    name=f"{flow_name}/default",
    parameters={'q', 42},
    scheduled_time = my_time,
    flow_run_name = my_run_name,
    timeout = 0,
    tags=["I", "love", "prefect"],
    idempotency_key = my_external_id,
  )
```

### Checklist

- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [ ] This pull request includes tests or only affects documentation.
- [ ] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

Issue #7522, opened 2022-11-14
Support run_deployment with tag


kevin868:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

The downstream function `create_flow_run_from_deployment` supports tags, but `run_deployment` does not take this param.

### Describe the proposed behavior

If we want to add tags to flow_runs created from our script.. we need to use `update_flow_run()` immediately after the `run_deployment()` call.

Ideally we can do it in one call, during the flow_run creation (run_deployment)

### Example Use

```
await run_deployment(
    name=f"{flow_name}/default",
    parameters={'q', 42},
    scheduled_time = my_time,
    flow_run_name = my_run_name,
    tags = ['i', 'love', 'prefect'], # <-----
    timeout = 0,
    idempotency_key = my_external_id # <- also in Issue #7237
  )
```

### Additional context

This request goes hand in hand with #7237 . If accepted, could be efficient to tackle both together.

madkinsz:
Sounds good to me!

humantargetjoe:
I would also like this feature!

discdiver:
This functionality now exists. 
![Uploading Screenshot 2024-02-09 at 1.13.47 PM.png…]()

PR #7771, opened 2022-12-05
Update deployment docs to include tag and idempotency key

riquelmev:
Added docstring documentation about deployment tags and idempotency key to deployments.py. 
Original request: https://github.com/PrefectHQ/prefect/issues/7522
Although https://github.com/PrefectHQ/prefect/pull/7641 fixed the issue, no documentation was added, hence this pr.

### Checklist

- [X ] This pull request includes tests or only affects documentation.

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -73,7 +73,7 @@ async def run_deployment(
             Defaults to None
         poll_interval: The number of seconds between polls
         tags: An optional list of tags to associate with this flow run; note that tags are used only for organizational purposes.
-        idempotency_key: q unique value to be used by the server to recognize subsequent retries of the same run.
+        idempotency_key: a unique value to be used by the server to recognize subsequent retries of the same run.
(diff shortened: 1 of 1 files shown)


src/prefect/deployments.py
```diff
@@ -72,6 +72,8 @@ async def run_deployment(
             Setting `timeout` to None will allow this function to poll indefinitely.
             Defaults to None
         poll_interval: The number of seconds between polls
+        tags: An optional list of tags to associate with this flow run; note that tags are used only for organizational purposes.
```
kevin868, line 75: You can remove 'optional' here.. as many of these params are, without explicitly saying so.

Also consider adding in this PR, `Optional[type]` tags above to each of the items with defaults (Everything but `name`). but this style should be something the Prefect team should weigh in on. It's _nice_ but also clutters readability.

cicdw (approved):
Thank you!

PR #7772, opened 2022-12-05
Add `--start-in` and `--start-at` to `prefect deployment run`

### LINKED ISSUE #6946: Add scheduled start time to the `prefect deployment run` CLI
### First check

### Prefect Version

2.x

The CLI command: `prefect deployment run my_flow/my_deployment` currently triggers a run immediately.

It would be great to add a `--start-time` flag to schedule a run for later (the format might be ISO 8601 string):

```
prefect deployment run my_flow/my_deployment --start-time '2022-09-23T20:00:00+02:00'
```

### Example Use

make it aligned with UI: https://discourse.prefect.io/t/how-to-schedule-a-flow-run-from-a-deployment-for-a-specific-date/1654 

### Additional context

_No response_

madkinsz:
Co-authored with @serinamarie 
Closes #6946 

Uses [dateparser](https://dateparser.readthedocs.io/en/latest/) to support setting a start time with `prefect deployment run`.

### Example

```
❯ prefect deployment run foo/test --start-in "3 minutes"
Creating flow run for deployment 'foo/test'...
Created flow run 'crafty-herring'.
└── UUID: d10f0acb-0b33-45ab-b73a-e0c0961cff8a
└── Parameters: {}
└── Scheduled start time: 2022-12-05 17:37:40 (in 3 minutes)
└── URL: <no dashboard available>
```

```
❯ prefect deployment run foo/test --start-at "3pm tomorrow"
Creating flow run for deployment 'foo/test'...
Created flow run 'pompous-porpoise'.
└── UUID: 0ce7930e-8ec0-40cb-8a0e-65bccb7a9605
└── Parameters: {}
└── Scheduled start time: 2022-12-06 15:00:00
└── URL: <no dashboard available>
```

### Checklist


src/prefect/cli/deployment.py
```diff
@@ -426,20 +464,39 @@ async def run(
...
+    else:
+        scheduled_display = "now"
+
     app.console.print(f"Created flow run {flow_run.name!r}.")
     app.console.print(
         textwrap.dedent(
             f"""
         └── UUID: {flow_run.id}
         └── Parameters: {flow_run.parameters}
+        └── Scheduled start time: {scheduled_display}
```
serinamarie, line 499: I see that this will now always have a `Scheduled start time`, would it also make sense for one of the `--start-in` or `--start-at` to have a default of `now` rather than `None` or no?

madkinsz: Hm that might make the `--help` message clearer but then we won't be able to tell if they've passed both `--start-in` and `--start-at` or if one of them is just the default value. We could set `start_at` to "now" after we do that  check to simplify some of our code, but then we'll display a timestamp for the scheduled time instead of the literal string "now" which I thought might be a nicer UX? Maybe it's not though :)

serinamarie: Regarding your first point, ah yes, that's true! As for the second point, yes we could set `start_at` after we've checked that it wasn't set. I think it might be nice to be able to compare the timestamp in the UI or if for whatever reason you were for running it programmatically then perhaps you would have a nicer record of the time it was created.

madkinsz: Maybe we can have the best of both worlds and do "Scheduled start time: {TIMESTAMP} (now)" when it's scheduled for now?

serinamarie: Hehe I like that!

serinamarie: If they pass in `--start-in` and `--start-at`, `--start-at` is ignored because it's 2nd in the if/else statement.
```python
if start_in:
    do_stuff()
elif start_at:
    do_other_stuff()
```
 So technically a `start-at` default of "now" could be possible? Or maybe I overlooked something 🤔 But regardless, it might be nicer to try to handle it the way we already described above.

serinamarie:
Should we allow users to set a `—start-at` that is in the past? I feel like most people would do it by accident, and if anyone did it on purpose I don't think it should be encouraged. For the majority of folks, they might appreciate the guardrails? The effect currently is that the flow run begins immediately after being marked as `Late`.

serinamarie:
I think at least 1 test for this would be a `—start-in` ^ (xor) `—start-at` because I could see users possibly thinking they need to specify both when we want to leave no room for ambiguity on our end.

src/prefect/cli/deployment.py
madkinsz: Good point — we should detect this and exit with a nice message :)

madkinsz:
> Should we allow users to set a —start-at that is in the past? 

Probably? People use scheduled start times in the past for backfills (although why you'd use the scheduled start time instead of your own parameter, I'll never understand). It seems safer not to restrict it for now. We could always display the time delta (after the timestamp) which would hint that something is in the past so it'd at least be immediately obvious?

serinamarie:
> We could always display the time delta (after the timestamp) which would hint that something is in the past so it'd at least be immediately obvious?

Yes, that makes sense, it's especially consistent if we are saying that no specification of a start time is going to look like this:
```bash
Created flow run 'adaptable-oyster'.
└── UUID: 91b00cee-334c-44f2-882f-4f8952acf74a
└── Parameters: {'num': 10}
└── Scheduled start time: 2022-12-15 16:30:30 (now)
└── URL: ...
```

peytonrunyan (approved):

Issue #7720, opened 2022-12-01
(sqlite3.OperationalError) table flow_run has no column named state_timestamp


meystingray:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Python 3.8.10
Prefect 2.6.8
Flask 2.0.2

I'm calling a Prefect 2 flow using Flask, and am consistently getting errors with the sqlite3 flow_run table such as:

`table flow_run has no column named state_timestamp`

### Reproduction

```python3
`import os
import json
from flask import Flask, request
import logging
import traceback
from prefect.orion.schemas.states import StateType

app = Flask(__name__)

#################################################
#
#  Flask routes
#
#################################################
@app.route('/', methods=['POST'])
def default_route_main():

    force_exception = location_success = False
    try:
        # Run flow
        flow_state = my_flow(return_state=True)

        force_exception = flow_state.result()
        flow_success = flow_state.type == (StateType.COMPLETED)`
```

### Error

```python3
`Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions)(**req.view_args)
  File "/main.py", line 134, in default_route_main
    onboardupdate_state = onboardupdate_flow(ou_event, atcore_event, prelog, return_state=True)
  File "/usr/local/lib/python3.8/dist-packages/prefect/flows.py", line 442, in __call__
    return enter_flow_run_engine_from_flow_call(
  File "/usr/local/lib/python3.8/dist-packages/prefect/engine.py", line 157, in enter_flow_run_engine_from_flow_call
    return anyio.run(begin_run)
  File "/usr/local/lib/python3.8/dist-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
[...]

PR #7789, opened 2022-12-06
Add lock to ensure that alembic commands are not run concurrently

madkinsz:
Co-authored with @ahuang11 
Adds locks around alembic commands.

Alembic commands rely on mutation of shared global state. Running multiple commands concurrently can result in corrupted configurations. To prevent this, we add acquisition of a shared lock to all alembic commands.

See:
-  #7720
- https://github.com/PrefectHQ/prefect-dask/pull/50
- https://github.com/PrefectHQ/prefect-dask/issues/55

Note: In https://github.com/PrefectHQ/prefect/pull/7789/commits/b43e04eb3f616e8e7ba60e999b26b046c99ada25 I attempted to use async locks in the database interface instead. This resulted in deadlocks when alembic was used from multiple threads via separate event loops. Async locks are not thread safe and this approach is safer.
 

### Checklist

### DIFF
--- a/tests/orion/database/test_alembic_commands.py
+++ b/tests/orion/database/test_alembic_commands.py
@@ -76,9 +76,10 @@ class TestAlembicCommands:
         jobs = [run_sync_in_worker_thread(alembic_upgrade) for _ in range(0, 10)]
         await asyncio.gather(*jobs)
 
+    @pytest.mark.flaky(max_runs=3)
     async def test_concurrent_downgrade_upgrade(self):
         jobs = []
-        for _ in range(0, 3):
+        for _ in range(0, 2):
             jobs.append(run_sync_in_worker_thread(alembic_downgrade))
             jobs.append(run_sync_in_worker_thread(alembic_upgrade))
         await asyncio.gather(*jobs)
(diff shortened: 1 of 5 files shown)


madkinsz:
@ahuang11 is going to take ownership of this as it is related to some Dask work.

I have not checked if it resolves compatibility with Flask, cc @meystingray

jakekaplan (approved):
Nice!

tests/orion/database/test_alembic_commands.py
```diff
@@ -69,3 +71,15 @@ def test_alembic_stamp(self, mocked):
         _, kwargs = mocked.call_args
         assert mocked.call_count == 1
         assert kwargs["revision"] == "abcdef"
+
+    async def test_concurrent_upgrade(self):
+        jobs = [run_sync_in_worker_thread(alembic_upgrade) for _ in range(0, 10)]
+        await asyncio.gather(*jobs)
+
+    @pytest.mark.flaky(max_runs=3)
```
madkinsz, line 79: Does this need to be flaky?

ahuang11: I had it `range(0, 3)` and it timed out; let me see if 2 times out.

PR #7792, opened 2022-12-06
Bugfix: Fullscreen layouts on screens < `lg` should take up all the available space

znicholasbrown:
This PR fixes a long-standing bug where views like Radar that are full screen used a min-content height on screen sizes under the lg breakpoint.

### Example

Before:
Screen Shot 2022-12-06 at 2 50 16 PM

After:
Screen Shot 2022-12-06 at 2 49 44 PM

### Checklist

### DIFF
--- a/orion-ui/src/App.vue
+++ b/orion-ui/src/App.vue
@@ -12,7 +12,7 @@
     </template>
     <ContextSidebar v-if="showMenu" class="app__sidebar" @click="close" />
     <suspense>
-      <AppRouterView />
+      <AppRouterView class="app__router-view" />
     </suspense>
   </div>
 </template>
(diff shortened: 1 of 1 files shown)


orion-ui/src/App.vue
```diff
@@ -79,9 +83,14 @@
...
-
     display: grid;
     grid-template-columns: max-content minmax(0, 1fr);
   }
 }
+
+.app__router-view {
+  /* The 1px flex-basis is important because it allows us to use height: 100% without additional flexing */
+  flex: 1 0 1px;
+  height: 100%;
```
pleek91, line 94: Won't this make this div scroll rather than the whole page?

znicholasbrown: If I understand your meaning then no, this will cause the element to expand to 100% of the specified parent height which, with flex-basis set at 1px, will be the computed height of the cell after Flexbox calculations.

pleek91: Okay, I trust you lol.

pleek91 (approved):

PR #7794, opened 2022-12-06
Create a `CANCELLING` state type

### LINKED ISSUE #7735: Cancellation should use a CANCELLING state type instead of state name
### First check

### Description

Cancellation is requested now with a transition from RUNNING -> Cancelling (CANCELLED). After cancellation completes, the state changes to Cancelled (CANCELLED). The state name is used as a filter at the agent level to avoid cancelling already cancelled runs.

There are some problems with this approach:
- Concurrency slots are released immediately on transition to Cancelling — but the run is still running
- The state name is being used to determine orchestration behavior — against our design for state types.

An alternative may be keeping the state name Cancelling (RUNNING), but this requires orchestration rules to allow RUNNING -> RUNNING transitions.

### Impact

Improve the cancellation experience and internal coherence of orchestration.

### Additional context

_No response_

bunchesofdonald:
The MVP for cancellation used a `CANCELLED` state with a name of `Cancelling` for the agent to decide which flows need to be cancelled. The main issue with this is that concurrency slots are immediately released when that transition happens even though the tasks/flows are still technically running. There is also some general 'ick' factor in using a state name for an important operation like cancellation.

This introduces a `CANCELLING` state type and sets it up so that it maintains / occupies concurrency limit slots, both for tag-based task usage as well as work queues. 

Related to #7735 

### Checklist

### DIFF
--- a/src/prefect/orion/orchestration/core_policy.py
+++ b/src/prefect/orion/orchestration/core_policy.py
@@ -162,7 +162,8 @@ class SecureTaskConcurrencySlots(BaseOrchestrationRule):
 
 class ReleaseTaskConcurrencySlots(BaseUniversalTransform):
     """
-    Releases any concurrency slots held by a run upon exiting a Running state.
+    Releases any concurrency slots held by a run upon exiting a Running or
+    Cancelling state.
     """
 
     async def after_transition(
@@ -172,7 +173,10 @@ class ReleaseTaskConcurrencySlots(BaseUniversalTransform):
         if self.nullified_transition():
             return
 
-        if not context.validated_state.is_running():
+        if context.validated_state.type not in [
+            states.StateType.RUNNING,
+            states.StateType.CANCELLING,
+        ]:
             filtered_limits = (
                 await concurrency_limits.filter_concurrency_limits_for_orchestration(
                     context.session, tags=context.run.tags
--- a/src/prefect/orion/orchestration/core_policy.py
+++ b/src/prefect/orion/orchestration/core_policy.py
@@ -702,10 +702,23 @@ class PreventRedundantTransitions(BaseOrchestrationRule):
         StateType.SCHEDULED: 1,
         StateType.PENDING: 2,
         StateType.RUNNING: 3,
+        StateType.CANCELLING: 4,
     }
 
-    FROM_STATES = [StateType.SCHEDULED, StateType.PENDING, StateType.RUNNING, None]
-    TO_STATES = [StateType.SCHEDULED, StateType.PENDING, StateType.RUNNING, None]
+    FROM_STATES = [
+        StateType.SCHEDULED,
+        StateType.PENDING,
+        StateType.RUNNING,
+        StateType.CANCELLING,
+        None,
+    ]
+    TO_STATES = [
+        StateType.SCHEDULED,
+        StateType.PENDING,
+        StateType.RUNNING,
+        StateType.CANCELLING,
+        None,
+    ]
 
     async def before_transition(
         self,
@@ -715,6 +728,7 @@ class PreventRedundantTransitions(BaseOrchestrationRule):
     ) -> None:
         initial_state_type = initial_state.type if initial_state else None
         proposed_state_type = proposed_state.type if proposed_state else None
+
         if (
             self.STATE_PROGRESS[proposed_state_type]
             <= self.STATE_PROGRESS[initial_state_type]
--- a/tests/orion/models/test_work_queues.py
+++ b/tests/orion/models/test_work_queues.py
@@ -238,7 +238,12 @@ class TestDeleteWorkQueue:
 
 
 class TestGetRunsInWorkQueue:
-    running_flow_count = 4
+    running_flow_states = [
+        schemas.states.StateType.PENDING,
+        schemas.states.StateType.CANCELLING,
+        schemas.states.StateType.RUNNING,
+        schemas.states.StateType.RUNNING,
+    ]
 
     @pytest.fixture
     async def work_queue_2(self, session):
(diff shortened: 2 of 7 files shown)


src/prefect/orion/database/migrations/versions/sqlite/2022_12_06_161948_50ab89b8fb35_add_cancelling_to_state_type_enum.py
```diff
@@ -0,0 +1,22 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "50ab89b8fb35"
+down_revision = "7201de756d85"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("ALTER TYPE state_type ADD VALUE IF NOT EXISTS 'PAUSED';")
```
madkinsz, line 18:
```suggestion
    op.execute("ALTER TYPE state_type ADD VALUE IF NOT EXISTS 'CANCELLING';")
```

madkinsz:
Does this feel like the correct move?

src/prefect/orion/database/migrations/versions/sqlite/2022_12_06_161948_50ab89b8fb35_add_cancelling_to_state_type_enum.py
bunchesofdonald: Oh man, complete your copy paste Pickett.

tests/orion/models/test_work_queues.py
```diff
@@ -270,16 +272,23 @@ async def scheduled_flow_runs(self, session, deployment, work_queue, work_queue_
 
     @pytest.fixture
     async def running_flow_runs(self, session, deployment, work_queue, work_queue_2):
-        for i in range(3):
+        for i in range(self.running_flow_count):
```
madkinsz, line 275: It seems like this could just use a list of state types instead? The numerical index doesn't seem very clear.

bunchesofdonald: :+1: Works for me.

anticorrelator:
is there a specific interaction the cancelling state helps us spell out? I'm curious since I implemented something similar for `resuming` by adding a marker to a run

anticorrelator:
> @anticorrelator I'd have to know more about the 'marker to a run' to say if that would work here. The interaction that this specifically is about is the interaction between the agent and the running flows. It needs to know _which_ flow runs to Cancel and to also know which flow runs have _already_ been cancelled. Currently this is accomplished by setting the flow run into a `Cancelled` state with a name of `Cancelling`, then when the cancellation is complete the state is renamed to `Cancelled`.
> 
> This _works_, but as stated above, it cause two issues:
> 
>     1. The concurrency slots are released when transitioning from RUNNING -> CANCELLED, which isn't what we want as flows in this state are still technically running.
> 
>     2. There's an 'ick' factor of using the name of a state for this kind of operation.

There's definitely an ick factor on doing client operations on state names. I was thinking of a mechanism that looked like attaching metadata to the `Run` somehow, for example `run.cancelling = True` and doing clientside operations that way. I don't know if it feels that much better but if we should always treat a `Cancelling` state as a `Running` state for the sake of orchestration I was hoping we could maybe discuss some alternatives briefly

madkinsz:
I'm also a little worried about treating CANCELLING like RUNNING in most cases. There are some important differences though, like:
- We won't allow retries from a CANCELLING.
- We won't allow CRASHED or FAILED from CANCELLING.

bunchesofdonald:
@anticorrelator and I discussed this and we agreed that this is probably the best way forward given the needs of the agent.

anticorrelator (approved):
LGTM, as mentioned async: we would benefit from an extension to `PreventRedundantTransitions` or a new rule that only allows `Cancelling` -> `Cancelled` (and maybe crashed/failed?) transitions

madkinsz:
What kind of migration strategy do we have in mind for Cloud? How can we make it clear that this feature requires changes and client / server versions can't be mixed in the OSS?

tests/orion/models/test_work_queues.py
```diff
@@ -238,6 +238,13 @@ async def test_delete_work_queue_returns_false_if_does_not_exist(self, session):
 
 
 class TestGetRunsInWorkQueue:
+    running_flow_states = [
+        schemas.states.StateType.PENDING,
+        schemas.states.StateType.CANCELLING,
+        schemas.states.StateType.RUNNING,
```
zangell44, line 244:
```suggestion
```

zangell44 (approved):

bunchesofdonald:
> What kind of migration strategy do we have in mind for Cloud? 

After discussion with Zach and Dustin it appears that we'll need to:

- Bump the Orion API version
- Add a compatibility check for the rule update and the new rule I'm going to create for preventing `Cancelling` -> Anything but Cancelled.

> How can we make it clear that this feature requires changes and client / server versions can't be mixed in the OSS?

Isn't that what the bump to the Orion API version signals?

bunchesofdonald:
> What happens when I have a server with this code and a client with the old code calls `prefect flow-run cancel`? What happens when an agent is running on the old code and uses the wrong filter?

After discussing this with @abrookins, we decided that feature flagging was the way to go here. This change should be good to go as it is, but I'll follow up with another PR to make the CLI/Agent check that flag and change the state it's sending.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -172,7 +173,10 @@ async def after_transition(
         if self.nullified_transition():
             return
 
-        if not context.validated_state.is_running():
+        if context.validated_state.type not in [
+            states.StateType.RUNNING,
+            states.StateType.CANCELLING,
```
anticorrelator, line 178: this will need a matching update in cloud

anticorrelator (approved):

Issue #7752, opened 2022-12-03
Timed out tasks do not release concurrency slot


giorgiobasile:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When a tagged task times out, based on its `timeout_seconds` value, concurrency slots associated with its tags are not released properly.

### Reproduction

```python3
import time
import asyncio
from prefect import get_client, task, flow, get_run_logger


async def create_concurrency_limit():
    async with get_client() as client:
        limit_id = await client.create_concurrency_limit(
            tag="small_instance", 
            concurrency_limit=1
        )

@task(tags=["small_instance"], timeout_seconds=5, retries=1)
def my_task():
    logger = get_run_logger()
    logger.info("Task start")
    time.sleep(10)

@flow
def my_flow():
   my_task()

asyncio.run(create_concurrency_limit())
my_flow()
```

### Error

The first task run goes in `TimedOut` state, but it is not retried since the `small_instance` slot has not been released correctly.

```python3
19:30:09.580 | INFO    | prefect.engine - Created flow run 'giga-condor' for flow 'my-flow'
19:30:09.646 | INFO    | Flow run 'giga-condor' - Created task run 'my_task-20c6ece6-0' for task 'my_task'
19:30:09.646 | INFO    | Flow run 'giga-condor' - Executing 'my_task-20c6ece6-0' immediately...
19:30:09.665 | INFO    | Task run 'my_task-20c6ece6-0' - Task start
19:30:14.664 | ERROR   | Task run 'my_task-20c6ece6-0' - Task run exceeded timeout of 5.0 seconds
Traceback (most recent call last):
  File "/Users/giorgio/Projects/personal/playground/.venv/lib/python3.10/site-packages/prefect/engine.py", line 1346, in orchestrate_task_run
    result = await run_sync(task.fn, *args, **kwargs)
  File "/Users/giorgio/Projects/personal/playground/.venv/lib/python3.10/site-packages/prefect/utilities/asyncutils.py", line 130, in run_sync_in_interruptible_worker_thread
    async with anyio.create_task_group() as tg:
  File "/Users/giorgio/Projects/personal/playground/.venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 658, in __aexit__
    raise CancelledError
[...]

PR #7798, opened 2022-12-07
Release task concurrency slots when transition is rejected as long as the task is not in a running state

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/7752.
Massive shoutout to @anticorrelator for the help.

Task concurrency slots are released according to `ReleaseTaskConcurrencySlots` which is currently a child class of `BaseOrchestrationRule`. Orchestration rules have `before_transition`, `after_transition`, and `cleanup` methods. The `cleanup` method is fired instead of the `after_transition` method if, at any point in the series of calls in the orchestration rules, a transition is rejected. `ReleaseTaskConcurrencySlots` releases the concurrency slot in the `after_transition` method.

This was causing a problem when retrying tasks, because a retry rejects a proposed final state and sets the state to `SCHEDULED` instead. This meant that `ReleaseTaskConcurrencySlots` was not able to release the concurrency slot.

 After discussing this with @anticorrelator there appeared to be two possible solutions. One was to make the `cleanup` method mirror the `after_transition` method. This would also require modifying the priority of `ReleaseTaskConcurrencySlots`. The alternative was changing `ReleaseTaskConcurrencySlots` from a `BaseOrchestrationRule` to a `BaseUniversalTransform`.

The `after_transition` method on a `BaseUniversalTransform` will fire on each state transition regardless of `ACCEPT` or `REJECT` status. `ABORT` is handled by a check at the start of the method. 

We elected to make it a `BaseUniversalTransform`. This newly modified `ReleaseTaskConcurrencySlots` will release a concurrency slot as long as the task transitions out of running.

I'd appreciate a second set of eyes on these tests. Yesterday was my first time looking at the orchestration rules, so I might be missing something. 

### Checklist

### DIFF
--- a/tests/orion/orchestration/test_core_policy.py
+++ b/tests/orion/orchestration/test_core_policy.py
@@ -1494,7 +1494,6 @@ class TestTaskConcurrencyLimits:
         task2_running_ctx = await initialize_orchestration(
             session, "task", *running_transition, run_tags=["shrinking limit"]
         )
-
         async with contextlib.AsyncExitStack() as stack:
             for rule in concurrency_policy:
                 task2_running_ctx = await stack.enter_async_context(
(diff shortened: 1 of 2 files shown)


tests/orion/orchestration/test_core_policy.py
```diff
@@ -1553,3 +1552,321 @@ async def test_concurrency_race_condition_limit_decreases_dont_impact_existing_r
         # the concurrency slot is released as expected
         assert task1_completed_ctx.response_status == SetStateStatus.ACCEPT
         assert (await self.count_concurrency_slots(session, "shrinking limit")) == 0
+
+    async def test_returning_concurrency_slots_when_transitioning_out_of_running_even_on_fizzle_with_release_first(
```
anticorrelator, line 1556: super pedantic: `fizzling` implies that the slot was released before the state mutation, so we don't need the `with_release_first` piece to the test name

tests/orion/orchestration/test_core_policy.py
```diff
@@ -1553,3 +1552,321 @@ async def test_concurrency_race_condition_limit_decreases_dont_impact_existing_r
...
+                task1_pending_ctx = await stack.enter_async_context(
+                    rule(task1_pending_ctx, *pending_transition)
+                )
+            await task1_pending_ctx.validate_proposed_state()
+
+        assert task1_pending_ctx.response_status == SetStateStatus.REJECT
+        assert task1_pending_ctx.validated_state.type != states.StateType.RUNNING
+        assert (await self.count_concurrency_slots(session, "small")) == 0
+
+    async def test_returning_concurrency_slots_when_transitioning_out_of_running_even_on_fizzle_with_release_last(
```
anticorrelator, line 1640: for this test, we aren't "fizzling" the concurrency slot release, we'd be invalidating it instead (though of course these concepts don't really apply to transforms, only rules)

anticorrelator (approved):
this looks great! thanks for fixing this @peytonrunyan—I just had a few small notes about test names, but other than that could you check with @bunchesofdonald to make sure the updated concurrency slot release logic is compatible with his cancelling work?

PR #7799, opened 2022-12-07
Fix cancellation of flows running in processes on Windows

### LINKED ISSUE #7742: Windows: Cancelling process flow run kills agent too
### First check

### Bug summary

I have a prefect agent running in a windows command prompt.
I also have a separate command prompt running prefect orion

If I have a scheduled or manually submitted flow run and I press cancel in Prefect Orion the job correctly transitions to a cancelling state. 

This causes the prefect agent to kill the processes for the flow run however this also sends a kill signal to the agent process itself:

image

Note "Terminate Batch Job Y/N" was not caused by me pressing ctrl+c - this happened from the external signal via prefect orion.

Restarting the agent leaves it in a confused state since it never logs the successful kill of the flow:

image

This may be linked to issue #7736 

### Reproduction

```python3
Windows, have an agent running, prefect orion running and a flow running.

Cancel a flow run from the UI.

Agent process receives kill signal, locking it at a "Terminate Batch Job Y/N" state.
```

### Error

```python3
After restarting the agent:

14:00:59.727 | ERROR   | prefect.agent - Encountered exception while killing infrastructure for flow run '2c154be8-e775-413a-afa1-19fc994084f9'. Flow run may not be cancelled.
  File "D:\prefect\prefect_agent\lib\site-packages\prefect\agent.py", line 283, in cancel_run
    await infrastructure.kill(flow_run.infrastructure_pid)
  File "D:\prefect\prefect_agent\lib\site-packages\prefect\infrastructure\process.py", line 159, in kill
    os.kill(pid, signal.CTRL_BREAK_EVENT)
OSError: [WinError 87] The parameter is incorrect
```

### Versions

```Text
Version:             2.7.0
API version:         0.8.3
Python version:      3.10.6
Git commit:          c2833339
Built:               Thu, Dec 1, 2022 4:03 PM
OS/Arch:             win32/AMD64
Profile:             prod
Server:
  Database:          sqlite
  SQLite version:    3.39.4
```

### Additional context

_No response_

rpeden:
### Summary
On Windows, canceling flows created by the Process infrastructure block also stops the Prefect agent. 

This happens because all processes in Windows process group share a [console](https://learn.microsoft.com/en-us/windows/console/consoles). And by default, Windows subprocesses spawn in the same process group as their parent. Since CTRL-BREAK (which we are sending to kill the subprocess) is a console event, an agent sending it to a subprocess means it also sends CTRL-BREAK to itself and any other subprocesses it started, so they all terminate.

Then, when the agent restarts, it looks for processes it thinks it needs to cancel, even though they've already been terminated. Our code checks for this and tries to catch a `ProcessLookupError` indicating the process is already gone. On Windows, however, a `WindowsError` is raised instead. 

### Changes Included in this Pull Request

To solve both problems, this PR:

* Creates an `asyncio`-specific implementation of `anyio.open_process` that allows us to pass additional keyword arguments needed to create Windows subprocesses in new process groups. This implementation is necessary because `asyncio`'s subprocess creation methods support the `creationflags` parameter we need to use, but anyio's default `open_process` does not provide a way to forward the parameter to the `asyncio` backend.
    * I discussed this approach with @madkinsz to ensure it's okay to do to ship this fix. See the *Additional Considerations* section below for comments on a possible path forward in a future PR.
* Updates the `Process` infrastructure block to create Windows subprocesses in a new process group by setting the `creationflags` parameter to `subprocess.CREATE_NEW_PROCESS_GROUP` on process creation.
* Uses [SetConsoleCtrlHandler](https://learn.microsoft.com/en-us/windows/console/setconsolectrlhandler) to add a Ctrl-C handler to the main process that terminates flow runner subprocesses before handing control over to any remaining handlers (i.e., anything the agent normally does in response to SIGINT or KeyboardInterrupt)
* Catches `WindowsError` in addition to `ProcessLookupError` when trying to terminate a process (Python may still raise `ProcessLookError` on other some versions of Windows or on Python interpreters compiled with MinGW, so I think we should keep handling it even though I did not encounter it.)
* Updates the test suite for the Process infrastructure block to work correctly on both Windows and Unix-like operating systems.
* Adds tests to cover new functionality.

Closes #7742.
 
#### Additional Considerations
* Since upcoming Prefect engine changes may tie us more closely to `asyncio`, we could update `processutils` to avoid using `anyio`'s process and stream wrappers and instead work directly with `asyncio`. I avoided doing so in this PR to keep its scope reasonable. 
    * We use some of the convenience methods added by `anyio`'s process wrapper, so we'd either have to create our own equivalent wrapper or rework our code and tests to work without it.
* Or, we could open an upstream issue in the `anyio` repository to see if they'd be open to adding this feature. Per the comments [this PR comment](https://github.com/agronholm/anyio/pull/279#issuecomment-832112951), changes that add specific arguments supported by all `anyio` backends seem acceptable. 
   * This would fit our current use case but may land us back in the same situation if we find we need to pass more arguments through to the `asyncio` backend when creating processes.

### Steps to QA:
1. Start with (or create) a deployment that runs flows in processes. Either Orion or Prefect Cloud will work.
2. Clone this PR's branch to a Windows machine.
3. Run the tests in `test/infrastructure/test_process.py` and `tests/utilities/test_processutils.py` and verify that they all pass on Windows. 
4. Run an agent that watches the work queue of the deployment from step 1.
5. Manually create a flow run for the deployment. 
6. Once it starts running, verify that pressing Ctrl-C at the console stops both the agent and the flow subprocess. This ensures that the extra handler needed to make Ctrl-C work on subprocesses in different process groups is doing its job.
7. Manually create another flow run for the deployment. 
8. Once it starts running, navigate to the flow run page in the UI and click the red Cancel button. 
9. Watch the output of your agent, verify that it picks up the cancellation, and then verify that it successfully cancels the flow process without terminating the agent process.
10. Note this PR's list of green checkmarks indicating that all tests still pass on Linux.
    * But if you'd like to be thorough in your QA, repeat the steps above on both Linux and MacOS and verify that both Ctrl-C and cancellation behavior remains unchanged from 2.7.1. I did so and everything looked normal. 

### Example
* Install Prefect 2.7.0 or 2.7.1. 
* Run through the steps to QA, but skip steps 6 and 7 and go straight to the step with the big red Cancel button. 
* Click the Cancel button and notice that the flow process stops like you want it to. Then note that unfortunately, the agent also disappeared and left you with a blinking `cmd` or PowerShell prompt.
    * Hopefully, a Prefect agent on Windows in production is run as a service set to auto-restart.
    * But we'd still prefer that the agent not accidentally terminate its own process. 😿 
* Next, install this PR's branch. 
* Run through the same steps.
* Notice that the agent now terminates the flow subprocess without terminating its own process. 😸 

### Checklist


src/prefect/infrastructure/process.py
```diff
@@ -185,7 +206,7 @@ async def kill(self, infrastructure_pid: str, grace_seconds: int = 30):
 
             try:
```
rpeden, line 207: I get a CodeQL warning here, but I got the same warning before making any changes so I left this code in the same place I found it aside from changing the exception type.

rpeden, line 207: I get a CodeQL warning here, but I got the same warning before making any changes so I left this code in the same place I found it aside from changing the exception type.

madkinsz:
Did you mean to close this?

src/prefect/utilities/processutils.py
```diff
@@ -13,6 +26,161 @@
 TextSink = Union[anyio.AsyncFile, TextIO, TextSendStream]
 
 
+if sys.platform == "win32":
+    from ctypes import WINFUNCTYPE, c_int, c_uint, windll
+
+    _windows_process_group_pids = set()
```
rpeden, line 32: I dislike this approach, but haven't found a better one if we want to be able to handle CTRL-C presses after we've started Windows processes in new process groups. 

We could use a Python SIGINT handler instead, but that could get replaced elsewhere in the codebase. In contrast, adding a new Win32 console control handler doesn't remove existing handlers.

 `pywin32` offers a more Pythonic ways of working with the win32 API, but it's a large dependency I don't think we want to add.

madkinsz (approved):
All the logic here is gated to win32 to prevent any regressions on other systems while still delivering this fix ASAP — we are likely to revisit this in the future.

Let's make sure to get a feature request / issue open in AnyIO too.

PR #7802, opened 2022-12-07
Introducing the `PREFECT_API_ENABLE_HTTP2` setting to give users a choice

chrisguidry:
We've seen enough folks having trouble with HTTP/2 connectivity, and we want to give them the option to disable it if they are experiencing difficulty with it with their workloads and environments.

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -390,6 +390,9 @@ PREFECT_API_KEY = Setting(
 )
 """API key used to authenticate against Orion API. Defaults to `None`."""
 
+PREFECT_API_ENABLE_HTTP2 = Setting(bool, default=True)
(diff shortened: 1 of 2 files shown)


src/prefect/settings.py
```diff
@@ -390,6 +390,9 @@ def default_cloud_ui_url(settings, value):
 )
 """API key used to authenticate against Orion API. Defaults to `None`."""
 
+PREFECT_API_ENABLE_HTTP2 = Setting(bool, default=True)
+"""If True, enable support for HTTP/2 for communicating with a remote Orion API"""
```
madkinsz, line 394: Let's note that if the API does not support HTTP/2 this will have no effect.

PR #7810, opened 2022-12-08
Add setting to toggle interpreting square brackets as style

ahuang11:
Sometimes, strings that contain brackets  [  ]  aren't being logged correctly because rich interprets them as styles. This adds a setting to allow toggling and defaults it to False:

"Whether to interpret strings wrapped in square brackets as a style.
This allows styles to be conveniently added to log messages, e.g.
`[red]This is a red message.[/red]`. However, the downside is,
if enabled, strings that contain square brackets may be inaccurately
interpreted and lead to incomplete output, e.g.
`DROP TABLE [dbo].[SomeTable];"` results in `DROP TABLE .[SomeTable];`."

Originally from [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1670455860648239)

### Example

```
from prefect import flow, task, get_run_logger

@task
def execute_sql(sql: str):
    logger = get_run_logger()
    logger.info(f"Executing: {sql}")
    logger.info(sql)


@flow
def bracket_flow():
    execute_sql("DROP TABLE [dbo].[SomeTable];")
    execute_sql("EXEC [dbo].[usp_SomeProcedure];")


if __name__ == "__main__":
    bracket_flow()
```

image

### Checklist

### DIFF
--- a/docs/concepts/logs.md
+++ b/docs/concepts/logs.md
@@ -348,7 +348,7 @@ Then, the following will highlight "fancy" in red.
 from prefect import flow, get_run_logger
 
 @flow
-def log_email_flow():
+def my_flow():
     logger = get_run_logger()
     logger.info("This is [bold red]fancy[/]")
 
--- a/docs/concepts/logs.md
+++ b/docs/concepts/logs.md
@@ -333,6 +333,31 @@ def log_email_flow():
 log_email_flow()
 ```
 
+## Applying markup in logs
+
+To use [Rich's markup](https://rich.readthedocs.io/en/stable/markup.html#console-markup) in Prefect logs, first configure `PREFECT_LOGGING_MARKUP`.
+
+<div class='terminal'>
+```bash
+PREFECT_LOGGING_MARKUP=True
+```
+</div>
+
+Then, the following will highlight "fancy" in red.
+```python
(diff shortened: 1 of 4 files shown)


tpdorsey (approved):
Docs changes look good to me. thx

src/prefect/settings.py
```diff
@@ -562,6 +562,19 @@ def default_cloud_ui_url(settings, value):
 )
 """Whether to style console logs with color."""
 
+PREFECT_LOGGING_MARKUP = Setting(
```
madkinsz, line 565: I think we often use `_ENABLED` for settings like this, but I see you didn't for the logging colors option so I don't feel strongly here.

madkinsz (approved):
Great work!

docs/concepts/logs.md
```diff
@@ -333,6 +333,31 @@ def log_email_flow():
...
+PREFECT_LOGGING_MARKUP=True
+```
+</div>
+
+Then, the following will highlight "fancy" in red.
+```python
+from prefect import flow, get_run_logger
+
+@flow
+def log_email_flow():
```
ahuang11, line 351:
```suggestion
def my_flow():
```

PR #7818, opened 2022-12-08
Update Dockerfile - Decrease number of layers

mohitsaxenaknoldus:
As a best practice, we should have less number of layers to decrease the final image size. Commands like `RUN`, `ADD`, and `COPY` creates layers.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,10 +21,9 @@ RUN apt-get update && \
     apt-get install --no-install-recommends -y \
         # Required for arm64 builds
         chromium \
-    && apt-get clean && rm -rf /var/lib/apt/lists/*
-
-# Install a newer npm to avoid esbuild errors
-RUN npm install -g npm@8
+    && apt-get clean && rm -rf /var/lib/apt/lists/* \
+    # Install a newer npm to avoid esbuild errors
+    && npm install -g npm@8
 
 # Install dependencies separately so they cache
 COPY ./orion-ui/package*.json .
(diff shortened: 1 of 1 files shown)


madkinsz:
I worry this reduces readability — can you share the concrete advantage of reducing the layers here?

mohitsaxenaknoldus:
> Thanks for contributing! Welcome to Prefect :)
> 
> I worry this reduces readability — can you share the concrete advantage of reducing the layers here?

This will result in a lighter final image.

madkinsz:
Can you share some numbers for the change in image size?

jawnsy:
Hey @mohitsaxenaknoldus, thanks so much for taking the time to make these changes and offer this pull request!

While I understand the recommendation ([Minimize the number of layers](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) in the Dockerfile best practices document), we would prefer to explore different approaches to address these issues instead, such as squashing layers in our release builds.

Merging the build instructions as you've done here is a viable way to reduce layers, but also reduces the effectiveness of the build cache, which makes local development slower. As @madkinsz mentioned, there are also tradeoffs with respect to readability and ease of maintenance for others.

For these reasons, I do not think we will accept this pull request. Let us know if you have questions!

mohitsaxenaknoldus:
> Hey @mohitsaxenaknoldus, thanks so much for taking the time to make these changes and offer this pull request!
> 
> While I understand the recommendation ([Minimize the number of layers](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) in the Dockerfile best practices document), we would prefer to explore different approaches to address these issues instead, such as squashing layers in our release builds.
> 
> Merging the build instructions as you've done here is a viable way to reduce layers, but also reduces the effectiveness of the build cache, which makes local development slower. As @madkinsz mentioned, there are also tradeoffs with respect to readability and ease of maintenance for others.
> 
> For these reasons, I do not think we will accept this pull request. Let us know if you have questions!

Thanks for those points! I'll close this now. I can work with anything DevOps related like Docker, CICD, etc so let me know if there is something I can work on :)

PR #7819, opened 2022-12-08
Add experimental utilities

madkinsz:
Adds utilities for marking features as experimental. In brief:

- Use of experimental features will display a warning unless disabled.
- Experimental features can require explicit opt-in or usage will raise an error.
- Adding a new experimental group requires adding a setting to toggle warnings. If the group requires explicit opt-in, another setting must be added.

Regarding the module structure: I intend to begin moving internal code into an `_internal` module so we can be very explicit about what our user-facing API is. This `compatibility` namespace will house a `deprecated` module for upcoming deprecation related utilities and a `backports` module for the code currently in `prefect.utilities.compat`.

### Example

```python
from prefect._internal.compatibility.experimental import experimental


@experimental(group="test", feature="A test function", help="Just a test function :)")
def foo():
    print("foo!")


foo()
```

```
❯ python example.py
/Users/mz/dev/prefect/example.py:69: ExperimentalWarning: A test function is experimental
and the interface or behavior may change without warning. We recommend pinning to a 
specific version to avoid breaking changes. Just a test function :) To disable warnings for 
this group, disable PREFECT_EXPERIMENTAL_WARN_TEST. 
  foo()
foo!
```

### Checklist

### DIFF
--- a/src/prefect/_internal/compatibility/experimental.py
+++ b/src/prefect/_internal/compatibility/experimental.py
@@ -1,0 +1,99 @@
+"""
+Utilities for experimental features.
+
+Experimental features have a group, feature name, and optional help string.
+
+When an experimental feature is used, a warning will be displayed. Warnings may be
+disabled per feature group with the setting `PREFECT_EXPERIMENTAL_WARN_<GROUP>`. 
+Warnings may also be disabled globally with the setting `PREFECT_EXPERIMENTAL_WARN`.
+
+Some experimental features require opt-in to enable any usage. These require the setting
+`PREFECT_EXPERIMENTAL_ENABLE_<GROUP>` to be set or an error will be thrown on use.
+"""
+import functools
+import warnings
+from typing import Callable, TypeVar
+
+from prefect.settings import PREFECT_EXPERIMENTAL_WARN, SETTING_VARIABLES, Setting
+
+T = TypeVar("T", bound=Callable)
+
+EXPERIMENTAL_WARNING = (
+    "{feature} is experimental and the interface or behavior may change without warning. "
+    "We recommend pinning to a specific version to avoid breaking changes. "
+    "{help} "
+    "To disable warnings for this group, disable PREFECT_EXPERIMENTAL_WARN_{group}. "
+)
+
+EXPERIMENTAL_ERROR = (
+    "{feature} is experimental and requires opt-in for usage. "
+    "{help} "
+    "To use this feature, enable PREFECT_EXPERIMENTAL_ENABLE_{group}. "
+)
+
+
+class ExperimentalWarning(Warning):
+    """
+    A warning for use of expirimental code.
+
+    These can be globally disabled by the PREFECT_EXPIRIMENTAL_WARN setting.
+    """
+
+
+class ExperimentalError(Exception):
+    """
+    An exception for use of experimental code.
+    """
+
+
+def _opt_in_setting_for_group(group: str) -> Setting[bool]:
+    group_opt_in_setting_name = f"PREFECT_EXPERIMENTAL_ENABLE_{group.upper()}"
+    group_opt_in = SETTING_VARIABLES.get(group_opt_in_setting_name)
+    if group_opt_in is None:
+        raise ValueError(
+            f"A opt-in setting for experimental feature {group!r} does not exist yet. "
+            f"{group_opt_in_setting_name!r} must be created before the group can be "
+            "used."
+        )
+    return group_opt_in
+
+
+def _warn_setting_for_group(group: str) -> Setting[bool]:
+    group_warn_setting_name = f"PREFECT_EXPERIMENTAL_WARN_{group.upper()}"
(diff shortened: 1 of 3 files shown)


abrookins:
@madkinsz I may not have spent enough time looking at this, but can I use this utility to branch inside of non-decorated code? Assume that I'm adding an experimental change to existing code path.

madkinsz:
@abrookins you can branch on the experimental setting for the group for opt-in behavior e.g. `if PREFECT_EXPERIMENTAL_ENABLE_FOO: ...` — I tried to write a context manage e.g. `with requires_experimental("foo"): ...` but that's actually not supported by Python.

abrookins:
Oh, perfect, branching on a setting works.

abrookins (approved):
Dang, this looks great! I like the "group" concept and the helpers to check if an experiment is enabled. 🔥 👍

bunchesofdonald (approved):
This looks fantastic! Seems like it will do pretty much everything we need it to do. Having a non-decorator utility would be super helpful.

chrisguidry:
One note on the UX:

```
/Users/mz/dev/prefect/example.py:69: ExperimentalWarning: A test function is experimental
and the interface or behavior may change without warning. We recommend pinning to a 
specific version to avoid breaking changes. Just a test function :) To disable warnings for 
this group, disable PREFECT_EXPERIMENTAL_WARN_TEST. 
```

The whole thing is framed as a scary and negative thing.  Can we reframe experiments as positive things that are cool and good, just...experimental?  I think it will require flipping the booleans of the config flags so that `True` opts in to experimental behavior (and they default to `False`).  The warning can be an invitation to explicitly opt into the feature in order to clear the warning if they inadvertently use an experimental function.

```
/Users/mz/dev/prefect/example.py:69: ExperimentalPrefectFeature: A test function is experimental and the interface or behavior may change before we release it to every one. We recommend pinning to a specific version to avoid breaking changes. Just a test function :) To opt in to using this feature and to hide warnings for this group of experiments, set PREFECT_EXPERIMENTAL_TEST=True in your configuration. 
```

src/prefect/_internal/compatibility/experimental.py
```diff
@@ -0,0 +1,106 @@
...
+    "We recommend pinning to a specific version to avoid breaking changes. "
+    "{help} "
+    "To disable warnings for this group, disable PREFECT_EXPERIMENTAL_WARN_{group}. "
+)
+
+EXPERIMENTAL_ERROR = (
+    "{feature} is experimental and requires opt-in for usage. "
+    "{help} "
+    "To use this feature, enable PREFECT_EXPERIMENTAL_ENABLE_{group}. "
+)
```
chrisguidry, line 32: Nit: I forget, do line breaks come through with `Warning` messages?  If so, I think we should break these over lines so they are a little more clean.

chrisguidry (commented):
I'm a big fan of this!  I just thing we need to frame it as a positive thing overall.

chrisguidry:
Also, after reading the PR in detail, I see that you distinguish between features that _require_ `opt-in` (and raise) versus those that only warn if you stumble across them.  That's great!  I still think all the messaging (including the name of the Warning and Exception classes) should be a friendly invitation.

src/prefect/_internal/compatibility/experimental.py
```diff
@@ -0,0 +1,106 @@
...
+        raise ValueError(
+            f"A warn setting for experimental feature {group!r} does not exist yet. "
+            f"{group_warn_setting_name!r} must be created before the group can be used."
+        )
+    return group_warn
+
+
+def experimental(
+    group: str, feature: str, help: str = "", stacklevel: int = 2, opt_in: bool = False
+) -> Callable[[T], T]:
```
chrisguidry, line 74: Decorator typing is already kinda tricky, and decorator _factory_ typing is even more whacky.  I think you need to go beyond just `Callable` and use `ParamSpec` in order for the function types to carry through.  We have some examples in Nebula where we're doing that: https://github.com/PrefectHQ/nebula/blob/main/src/prefect_cloud/utilities/tracing.py#L111-L117

chrisguidry: I _think_ most of this is possible in 3.8 but you'll have to monkey about with it

madkinsz: Really? I think that's only needed if you're moving things around

Screen Shot 2022-12-09 at 12 19 43 PM

Here, we're saying this function returns a callable which given a type T (which is also a callable in this case) returns type T so if you pass it a function with some signature it returns that whole thing back "unchanged".

chrisguidry: Whoa I think you just helped me simplify some things 🤯

madkinsz: Pyright likes it; let's hope mypy does too!

madkinsz:
I haven't found a great way to make the message friendlier, but I did change the exception names. I'm going to get this in so I can continue to move forward but I'm happy to adjust them in the future.

PR #7830, opened 2022-12-09
Allow `reject_transition` to use current state

madkinsz:
Co-authored with @anticorrelator 

Previously a REJECT response required a new state which was then always written to the database. Here, the `OrchestrationRule.reject_transition` utility is updated to accept `state=None`. When this is used, the REJECT orchestration response will contain the current state of the run and a new state will not be appended to the run's states. This allows REJECT semantics to be used to prevent transition to a proposed state without providing an explicit alternative. This is useful when a client is proposing a state with the assumption that they are in a different starting state than the orchestrator has.

Required for #7738 

### Checklist

### DIFF
--- a/src/prefect/orion/orchestration/rules.py
+++ b/src/prefect/orion/orchestration/rules.py
@@ -228,8 +228,7 @@ class FlowOrchestrationContext(OrchestrationContext):
         state. The state on the run is set to the validated state as well.
 
         If the proposed state is `None` when this method is called, no state will be
-        written. If the response type is REJECT, `self.validated_state` will be set to
-        the run's current state.
+        written and `self.validated_state` will be set to the run's current state.
 
         Returns:
             None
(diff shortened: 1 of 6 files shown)


src/prefect/orion/orchestration/rules.py
```diff
@@ -263,7 +265,12 @@ async def _validate_proposed_state(
         self,
         db: OrionDBInterface,
     ):
-        if self.proposed_state is not None:
+        if (
+            self.response_status == SetStateStatus.REJECT
+            and self.proposed_state is None
+        ):
+            validated_orm_state = self.run.state
```
madkinsz, line 272: Could we always do this if the proposed state is null? It seems nice to have it not just apply to REJECT semantics. Or do we perform logic based on `self.validated_state is None` later?

anticorrelator:
this idea is kickass, it's a small change that really allows us to be more expressive in our communication with clients in a backwards compatible way

src/prefect/orion/orchestration/rules.py
anticorrelator: I think we can! If our state machine acts like a markov chain, if orchestration doesn't specify a state only the current state makes sense as information

anticorrelator (approved):

chrisguidry (approved):
This all makes sense to me!

PR #7836, opened 2022-12-09
Defer imports for performance

jlowin:
I used [Tuna](https://github.com/nschloe/tuna) to profile import times and noticed a few easy wins:
- eager imports of `sqlalchemy` in `schemas/filters.py`
- eager import of `docker` in `docker.py` (this looks like an accident; it's lazily loaded a couple lines later)
- eager imports of the entire ephemeral server in `client/orion.py`

I deferred these imports and reduced import times by roughly 25-30%. For anyone using Prefect only for workflow development or as a client, these aren't needed (obviously working on or with the server would require them)

One other low-hanging fruit is the eager import of `apprise` in `blocks/notifications.py`. Unfortunately an Apprise class is used as a base class in the file, so deferring the import isn't possible. It would be straightforward to simply not import `prefect.blocks.notifications` on startup, but it would technically be a breaking change.

------------

Quick instructions for replicating (after `pip install tuna`):
```bash
python -X importtime -c "import prefect" 2> import.log
tuna import.log
```

### DIFF
--- a/src/prefect/client/orion.py
+++ b/src/prefect/client/orion.py
@@ -98,9 +98,9 @@ class OrionClient:
         httpx_settings.setdefault("headers", {})
         if api_version is None:
             # deferred import to avoid importing the entire server unless needed
-            import prefect.orion.api.server
+            from prefect.orion.api.server import ORION_API_VERSION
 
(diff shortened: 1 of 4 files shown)


madkinsz (approved):
Thanks!

PR #7847, opened 2022-12-10
Fix bug where pause expiration and late run services settings were swapped

madkinsz:
Introduced in #7738 — clearly not intentional and not a big deal except the service was not disabled per #7846 so the late runs service is running during tests which causes test failures.


zangell44 (approved):

anticorrelator (approved):
Oops 😢

PR #7850, opened 2022-12-11
Allow disabling SSL verification

dhirschfeld:
### Checklist


dhirschfeld:
Resolves #7850 7596

src/prefect/client/orion.py
```diff
@@ -96,6 +97,17 @@ def __init__(
     ) -> None:
         httpx_settings = httpx_settings.copy() if httpx_settings else {}
         httpx_settings.setdefault("headers", {})
+        verify = os.getenv("PREFECT_API_TLS_INSECURE_SKIP_VERIFY")
```
madkinsz, line 100: You can actually add a setting to `prefect.settings` instead! Settings can be configured with environment variables or via or profiles. It'll also take care of casting input to a bool.

dhirschfeld:
I'm happy for anyone else to pick this up / push changes to do it properly, otherwise I can try to take a look over the weekend...

madkinsz:
@serinamarie interested in looking at moving this to a setting?

dhirschfeld:
Thanks for picking this up - covid & the holiday season caught up with me before having a chance to circle back 😬

src/prefect/client/orion.py
```diff
@@ -96,6 +96,15 @@ def __init__(
...
         httpx_settings = httpx_settings.copy() if httpx_settings else {}
         httpx_settings.setdefault("headers", {})
+        if prefect.settings.PREFECT_API_TLS_INSECURE_SKIP_VERIFY:
+            httpx_settings.setdefault("verify", False)
+            msg = (
+                "SSL checking is disabled due to "
+                "`PREFECT_API_TLS_INSECURE_SKIP_VERIFY` "
+                "environment variable being set!"
+            )
+            warnings.warn(msg, UserWarning)
```
madkinsz, line 106: I'm not sure we need to include a warning, thoughts? If we do keep it, we should say "due to the PREFECT_API_TLS_INSECURE_SKIP_VERIFY setting."

serinamarie: Personally I feel like it would be a nice reminder in case you were to forget to change the setting in prod.

madkinsz: It means you'll get a warning on every flow run and cli command in dev though.

serinamarie: That is quite a lot; and it seems like we don't have other warnings except around deprecation and experimental features. So with that in mind, I'll remove it.

madkinsz:
We'll need a test here as well.

src/prefect/client/orion.py
```diff
@@ -199,6 +200,8 @@ def __init__(
 
         self.logger = get_logger("client")
 
+        self.httpx_settings = httpx_settings
```
serinamarie, line 203: Is there a better way to retrieve the `verify` setting?

tests/client/test_orion_client.py
```diff
@@ -1144,6 +1145,19 @@ async def test_read_filtered_logs(session, orion_client, deployment):
         assert log.flow_run_id not in flow_runs[3:]
 
 
+async def test_default_prefect_api_tls_insecure_skip_verify_setting():
+
+    async with get_client() as client:
+        assert "verify" not in client.httpx_settings
+
+
+async def test_user_configured_prefect_api_tls_insecure_skip_verify_settingis_false():
```
serinamarie, line 1154:
```suggestion
async def test_user_configured_prefect_api_tls_insecure_skip_verify_setting_is_false():
```

src/prefect/client/orion.py
madkinsz: Yeah I would not do it this way. I'd mock the `httpx.Client` and assert that it is called with the correct value.

tests/client/test_orion_client.py
```diff
@@ -1145,17 +1145,45 @@ async def test_read_filtered_logs(session, orion_client, deployment):
...
+        get_client()
 
-    async with get_client() as client:
-        assert "verify" not in client.httpx_settings
+    mock.assert_called_once_with(
+        headers={"X-PREFECT-API-VERSION": ANY},
+        verify=False,
+        app=ANY,
+        base_url=ANY,
+        timeout=httpx.Timeout(timeout=ANY),
```
madkinsz, line 1159: Can this (and similar for `headers`) just be 
```suggestion
        timeout=ANY
```

madkinsz (approved):

Issue #7056, opened 2022-10-03
Add Flag to Force Rerunning Cached Tasks


jeffquinn-msk:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently if a task is defined like so:

```
@task(cache_key_fn=task_input_hash)
def my_task():
  ...
```

The cached value will be returned forever, unless prefect detects a change in the source code of `my_task`, however source code hashing/change detection has corner cases.

### Describe the proposed behavior

I would like to have an environment variable `PREFECT_DISABLE_CACHE_HITS` or similar, that would prevent the cached value for `my_task` from being found, and would instead call the actual function again, and cache its result.

This would be useful in cases where changes are made to source code that cannot be detected by the current task source hashing.

There seem to be many corner cases that are currently not detected by source hashing, this would be a good fallback.

After I update my code, I do a run with PREFECT_DISABLE_CACHE_HITS=1, and then my cache is in the correct state.

### Example Use

Currently if a task is defined like so:

```
def some_other_fn():
  return 1

@task(cache_key_fn=task_input_hash)
def my_task():
  return some_other_fn()

@flow
def my_flow():
  return my_task()
```

I make a change to source code:

```
def some_other_fn():
  return 2
```

I will do a run with `PREFECT_DISABLE_CACHE_HITS=1` and it will force a recall of the function.

toro-berlin:
Hey @billpalombi,

Did you already have the chance to think into this issue or my PR? Can I help you clarify the use case?
Thank you for any general feedback if prefect wants to address the issue in this or another way.

Have a great holiday season, 
Tobias

Issue #7315, opened 2022-10-25
Allow manual cache key expiration


anna-geller:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently, users can set an expiration time on a cache key 

However, there are circumstances when users want to manually clear cache, possibly before the planned expiration date 

### Describe the proposed behavior

- Add a utility function allowing to manually clear cache for a given task run ID (or possibly flow run ID once/if we add flow-level caching)
- Add a route allowing to do the same via an API call
- Add a capability to do that from the UI (most likely a follow-up feature we'll create for the UI team once backend work is done)

tobias-leo:
Hello @anna-geller, 

We are currently exploring caching for our flows and just ran into a similar situation. Great to see that there is already an open issue.

We want to be able to force-execute tasks in different situations: 
- We want to be able to use a boolean Flow parameter to control whether cached results are used (default false, set by the operator via UI).
- If we execute task T from flow A, using the cached results is acceptable. If we execute task T from flow B, it should not use the cache.

What force-execute means: Found this [code snippet](https://github.com/PrefectHQ/prefect/blob/240c8c45a9a4cd46112bc3b90ac931c7e605cd19/src/prefect/orion/orchestration/core_policy.py#L212 ), perhaps it is only about overriding the decision made here.
- Already cached results for a given cache key are **not** used, even if they are not expired.
- After the execution of the task, the newly computed results are cached.

Adding a task parameter/decorator "cache_force_execute" may conveniently solve this.

Advantages over just the API feature request from my perspective: 
- I only invalidate cached results for a subset of inputs, only the ones I force-execute the task for.
- To invalidate cached results for a given task, I do not have to write a utility function called before the task execution or ask the operator to do it manually.


What do you think?

anna-geller:
> We want to be able to use a boolean Flow parameter to control whether cached results are used
it's essentially how caching already works: if your cache key function returns the same key, a cache is used, if it returns a different key, task run is computed without cache

I suggest reexamining your cache key function - your use case is already possible

discdiver:
@billpalombi mentioned we might also want to have the option to clear the cache per deployment or per workspace.

tobias-leo:
[...]

PR #7856, opened 2022-12-12
Introduce task option `bypass_cache` to allow force execution of tasks

toro-berlin:
This pull request adds a new task argument/decorator: `bypass_cache`. The added boolean option allows us to execute tasks even though there is a cached result for the determined cache key. The result returned by the execution is cached again with an identical cache key.

The new decorator allows flow operators to re-execute tasks in a flow. For example, via a flow parameter used in task `with_options(bypass_cache=param)`. Especially useful in case of errors where a re-execution is necessary (storage corruption, triggering something again).

closes https://github.com/PrefectHQ/prefect/issues/7056
point is mentioned in this as well: https://github.com/PrefectHQ/prefect/issues/7315

### Example

```
[~]$ cat example.py 
import time
from prefect.task_runners import SequentialTaskRunner

def static_cache_key(context, parameters):
    return "static constant as a cache key"

@task(cache_key_fn=static_cache_key)
def simple_task():
    print("Task Execution!")
    time.sleep(0.5)
    return time.time()

@flow(log_prints=True, task_runner=SequentialTaskRunner())
def simple_flow():
    print(simple_task.submit().result())
    print(simple_task.submit().result())
    # bypass_cache -> task gets executed and actual time printed + cached
    print(simple_task.with_options(bypass_cache=True).submit().result())
    print(simple_task.submit().result())

simple_flow()

[~]$ python example.py
14:10:06.712 | INFO    | prefect.engine - Created flow run 'amazing-taipan' for flow 'simple-flow'
14:10:06.861 | INFO    | Flow run 'amazing-taipan' - Created task run 'simple_task-5cbbcc6b-0' for task 'simple_task'
14:10:06.864 | INFO    | Flow run 'amazing-taipan' - Executing 'simple_task-5cbbcc6b-0' immediately...
14:10:06.897 | INFO    | Task run 'simple_task-5cbbcc6b-0' - Task Execution!
14:10:07.439 | INFO    | Task run 'simple_task-5cbbcc6b-0' - Finished in state Completed()
14:10:07.440 | INFO    | Flow run 'amazing-taipan' - 1670850607.399133
14:10:07.466 | INFO    | Flow run 'amazing-taipan' - Created task run 'simple_task-5cbbcc6b-1' for task 'simple_task'
14:10:07.467 | INFO    | Flow run 'amazing-taipan' - Executing 'simple_task-5cbbcc6b-1' immediately...
14:10:07.487 | INFO    | Task run 'simple_task-5cbbcc6b-1' - Finished in state Cached(type=COMPLETED)
14:10:07.510 | INFO    | Flow run 'amazing-taipan' - 1670850607.399133
14:10:07.528 | INFO    | Flow run 'amazing-taipan' - Created task run 'simple_task-5cbbcc6b-2' for task 'simple_task'
14:10:07.529 | INFO    | Flow run 'amazing-taipan' - Executing 'simple_task-5cbbcc6b-2' immediately...
14:10:07.557 | INFO    | Task run 'simple_task-5cbbcc6b-2' - Task Execution!
14:10:08.131 | INFO    | Task run 'simple_task-5cbbcc6b-2' - Finished in state Completed()
14:10:08.134 | INFO    | Flow run 'amazing-taipan' - 1670850608.0586176
14:10:08.167 | INFO    | Flow run 'amazing-taipan' - Created task run 'simple_task-5cbbcc6b-3' for task 'simple_task'
14:10:08.167 | INFO    | Flow run 'amazing-taipan' - Executing 'simple_task-5cbbcc6b-3' immediately...
14:10:08.186 | INFO    | Task run 'simple_task-5cbbcc6b-3' - Finished in state Cached(type=COMPLETED)
14:10:08.201 | INFO    | Flow run 'amazing-taipan' - 1670850608.0586176
14:10:08.222 | INFO    | Flow run 'amazing-taipan' - Finished in state Completed('All states completed.')  
```

### Checklist


madkinsz:
This is a great start, nice work! We'll want to get confirmation that this behavior makes sense to our product team since the linked issue is not `status:accepted` yet.

I'm wondering if there's a clearer name out there that emphasizes that the cache read will be skipped but the cache write will occur. Perhaps "overwrite_cache"? May require some brainstorming :)

cc @billpalombi

madkinsz:
We're still debating the best name internally. If you have any other suggestions w/ justification let us know!

toro-berlin:
This is not easy, so here are some ideas from my side.

We should use a name most people can directly link to the functionality. Everyone uses web browsers. This functionality is called "hard reload" or "hard refresh" in the browser world. The website is loaded directly from the server, ignoring the local caching.
- `cache hard reload` has 4.530.000 google search results
- `cache hard refresh` has 38.200.000 google search results

I would not debate not including the `hard` in the arg name. We do not have a soft or regular refresh. Having just a hard one does not make sense. So my suggestion currently is `cache_refresh` or `refresh_cache`.

But I am definitely open to other ideas.

madkinsz:
`refresh_cache` is pretty good!  I'll keep thinking about this — we might not have the bandwidth for review until after the holidays.

One other thing: I think that primarily you'll want to do this for a run without modifying your code. It sounds like a pain to redeploy code just to have some tasks skip the cache. To allow for this, we could have a setting for this i.e. `PREFECT_TASKS_REFRESH_CACHE: bool = False`. Since settings can be set with environment variables, it'll be easy to toggle for a single run. The default for this option at the task level would be `None` which means use the value of the setting. You could opt a specific task in or out at the task level using `True` and `False` and toggle all other tasks globally with the setting. `persist_results` works a bit like this https://docs.prefect.io/concepts/results/#toggling-persistence — What do you think?

toro-berlin:
I see cases/situations where an env var will definitely help. I like the idea. Found your PR regarding the result persistence env var and will have a look at how to adapt it here: https://github.com/PrefectHQ/prefect/pull/7228 

I will be on vacation as of today and till the 4th of January as well. So no worries regarding the Review.

tests/test_tasks.py
```diff
@@ -2365,6 +2432,31 @@ def initial_task():
...
+        def foo(x):
+            return x
+
+        @flow
+        def bar():
+            return (
+                foo._run(1),
+                foo._run(2),
+                foo.with_options(refresh_cache=True)._run(3),
+                foo._run(4),
```
madkinsz, line 2446: Note that `_run` is a holdover from before and we encourage calls with the `return_state` option instead for new tests so `foo(1, return_state=True)`.

src/prefect/tasks.py
```diff
@@ -213,6 +225,11 @@ def __init__(
 
         self.cache_key_fn = cache_key_fn
         self.cache_expiration = cache_expiration
+        self.refresh_cache = (
+            refresh_cache
+            if refresh_cache is not None
+            else get_tasks_refresh_cache_setting()
+        )
```
madkinsz, line 232: I think we want to apply this setting at runtime not init time.

tests/test_tasks.py
```diff
@@ -995,6 +996,64 @@ def bar():
...
+        @flow
+        def bar():
+            foo.submit(0).wait
+            return (
+                foo.submit(1).wait(),
+                refresh_task.submit(2).wait(),
+                not_refresh_task.submit(3).wait(),
+            )
+
+        with temporary_settings({PREFECT_TASKS_REFRESH_CACHE: True}):
```
madkinsz, line 1049: This test should fail per https://github.com/PrefectHQ/prefect/pull/7856/files#r1062934837 — does it?

toro-berlin: You are right; it failed. As I moved the decision task argument vs. Prefect setting to runtime, it is now working as expected and not failing anymore.

src/prefect/settings.py
```diff
@@ -470,6 +470,14 @@ def default_cloud_ui_url(settings, value):
 flow and task results will be persisted unless they opt out.
 """
 
+PREFECT_TASKS_REFRESH_CACHE = Setting(
+    bool,
+    default=False,
+)
+"""
+The task setting enables a refresh of cached results. If enabled, a re-execution of the
```
tpdorsey, line 478:
```suggestion
If `True`, enables a refresh of cached results: re-executing the
```

src/prefect/settings.py
```diff
@@ -470,6 +470,14 @@ def default_cloud_ui_url(settings, value):
...
+tasks will refresh the cached results for all tasks unless a task overwrites the setting.
```
tpdorsey, line 479:
```suggestion
task will refresh the cached results. Defaults to `False`.
```

src/prefect/tasks.py
```diff
@@ -153,6 +153,9 @@ class Task(Generic[P, R]):
         log_prints: If set, `print` statements in the task will be redirected to the
             Prefect logger for the task run. Defaults to `None`, which indicates
             that the value from the flow should be used.
+        refresh_cache: If set, eventually cached results for the cache key are not used.
```
tpdorsey, line 156:
```suggestion
        refresh_cache: If set, cached results for the cache key are not used.
```

src/prefect/tasks.py
```diff
@@ -309,6 +315,9 @@ def with_options(
             persist_result: A new option for enabling or disabling result persistence.
             result_storage: A new storage type to use for results.
             result_serializer: A new serializer to use for results.
+            timeout_seconds: A new maximum number of runtime in seconds.
```
tpdorsey, line 318:
```suggestion
            timeout_seconds: A new maximum time for the task to complete in seconds.
```

src/prefect/tasks.py
```diff
@@ -910,6 +924,9 @@ def task(
         log_prints: If set, `print` statements in the task will be redirected to the
             Prefect logger for the task run. Defaults to `None`, which indicates
             that the value from the flow should be used.
+        refresh_cache: If set, eventually cached results for the cache key are not used.
```
tpdorsey, line 927:
```suggestion
        refresh_cache: If set, cached results for the cache key are not used.
```

tpdorsey (approved):
Just reviewed doc changes and docstrings. Looks good. A few minor suggested edits.

madkinsz:
@anticorrelator could you review as well please?

madkinsz (approved):
Nice work :)

anticorrelator (approved):
This looks great! Thanks so much for contributing @toro-berlin. While I believe this code works as is, I'd like to point out the option of clearing the cache key in the `CacheRetreival` rule if `refresh_cache` is `True`. Two different agents might be orchestrating tasks that hit the same key with different settings. I imagine that if `refresh_cache` is set on one of the agents, we wouldn't want the other agent to pull from the cache if they are being executed concurrently.

madkinsz:
>  I imagine that if refresh_cache is set on one of the agents, we wouldn't want the other agent to pull from the cache if they are being executed concurrently.

I think it's okay if the cache is not yet updated by the task that's refreshing it. I guess ideally we'd put a lock on this cache key until the refresh is complete? Maybe we can track that in an issue.

PR #7864, opened 2022-12-12
Add `PREFECT_KUBERNETES_CLUSTER_UID` to allow bypass of `kube-system` namespace read

madkinsz:
### Checklist

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -1,5 +1,6 @@
 import copy
 import enum
+import os
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


src/prefect/infrastructure/kubernetes.py
```diff
@@ -378,11 +379,21 @@ def _get_cluster_uid(self) -> str:
         There is no real unique identifier for a cluster. However, the `kube-system`
         namespace is immutable and has a persistence UID that we use instead.
 
+        PREFECT_KUBERNETES_CLUSTER_UID can be set in cases where the `kube-system`
+        namespace cannot be read e.g. when a cluster role cannot be created.
+
         See https://github.com/kubernetes/kubernetes/issues/44954
         """
+        # Default to an environment variablle
```
peytonrunyan, line 387:
```suggestion
        # Default to an environment variable
```
^Minor typo, but that's secondary.

The docstring reads like we give preference to `kube-system` namespace and then go to `PREFECT_KUBERNETES_CLUSTER_UID` if that's a no go. It looks like we're giving preference to `PREFECT_KUBERNETES_CLUSTER_UID` in the case where both are provided.

madkinsz: 👍 addressed

jawnsy (approved):
Seems reasonable to me, thank you!

I tagged @jimid27 and @jamiezieziula because they'll be making the Helm-side changes. This env var doesn't require Helm changes since we can use extraEnvVars to pass a setting here.

Jimi/Jamie otherwise look at creating a ClusterRole/ClusterRoleBinding for this, so that the variable is only needed in restrictive environments

Trymzet:
What is "cluster id"? I think you created this concept and it's not defined anywhere what it is and how to create one.

andnig:
@Trymzet The kubernetes cluster id is the metadata.uid field of the kube-system manifest.

madkinsz:
@Trymzet we need to uniquely identify clusters. Kubernetes does not provide a native way to do so. 

> kubernetes cluster id is the metadata.uid field of the kube-system manifest.

is not quite true, but the uid of the kube-system namespace is a stable unique identifier for a given cluster and the best option for us. See https://github.com/kubernetes/kubernetes/issues/44954

PR #7872, opened 2022-12-13
Update experimental utilities to require kwarg for group

madkinsz:
For consistency in the API surface for these, the "feature" blurb comes first then the "group" is specified as a kwarg.


madkinsz:
The netlify build is not showing any logs -.-

peytonrunyan (approved):
lgtm. Pure curiosity - why the change from `TEST` to `test`?

madkinsz:
@peytonrunyan we parse it into upper-case for the setting name, but there's no reason to set the group name that way — figured it'd be good to be consistent and just picked my preference.

PR #7875, opened 2022-12-13
Add `WorkerThreadPool` for running synchronous work in threads

madkinsz:
Follow-up to #7865

Adds a worker pool class for submitting synchronous functions to run in background workers. This will eventually replace our usage of anyio's worker threads, giving us greater control over concurrent work and improved debugging.

Workers will be extended in the near future to support sending asynchronous calls back from workers to the event loop thread and sending asynchronous calls to workers running event loops.

This implementation is inspired largely by the [CPython `ThreadPoolExecutor`](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py) implementation. Handling for several weird edge cases (mostly related to garbage collection) is replicated here. This implementation will diverge further as we capitalize on asynchronous support.

### Example

```python
import asyncio
from prefect._internal.concurrency.workers import WorkerThreadPool

def identity(x):
    return x

async def main():
    pool = WorkerThreadPool()
    future = await pool.submit(identity, 1)
    print(await future.aresult())

asyncio.run(main())
```

### Checklist

### DIFF
--- a/src/prefect/_internal/concurrency/event_loop.py
+++ b/src/prefect/_internal/concurrency/event_loop.py
@@ -25,7 +25,7 @@ def get_running_loop() -> Optional[asyncio.BaseEventLoop]:
         return None
 
 
-def run_in_loop_thread(
+def call_in_loop(
     __loop: asyncio.AbstractEventLoop,
     __fn: Callable[P, T],
     *args: P.args,
@@ -34,6 +34,16 @@ def run_in_loop_thread(
     """
     Run a synchronous call in event loop's thread from another thread.
     """
+    future = call_soon_in_loop(__loop, __fn, *args, **kwargs)
+    return future.result()
+
+
+def call_soon_in_loop(
+    __loop: asyncio.AbstractEventLoop,
+    __fn: Callable[P, T],
+    *args: P.args,
+    **kwargs: P.kwargs
+) -> concurrent.futures.Future:
     future = concurrent.futures.Future()
 
     @functools.wraps(__fn)
@@ -46,4 +56,23 @@ def run_in_loop_thread(
                 raise
 
     __loop.call_soon_threadsafe(wrapper)
-    return future.result()
+    return future
+
+
+def call_soon(
+    __fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
+) -> concurrent.futures.Future:
+    future = concurrent.futures.Future()
+    __loop = asyncio.get_running_loop()
+
+    @functools.wraps(__fn)
+    def wrapper() -> None:
+        try:
+            future.set_result(__fn(*args, **kwargs))
+        except BaseException as exc:
+            future.set_exception(exc)
+            if not isinstance(exc, Exception):
+                raise
+
+    __loop.call_soon(wrapper)
+    return future
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -59,8 +59,8 @@ class _WorkerThread(threading.Thread):
                 self._queue.put_nowait(None)
                 return
 
-            self._idle.release()
             work_item.run()
+            self._idle.release()
 
             del work_item
 
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -43,7 +43,7 @@ class _WorkItem:
 class _WorkerThread(threading.Thread):
     def __init__(
         self,
-        queue: Queue[Union[_WorkItem, None]],
+        queue: "Queue[Union[_WorkItem, None]]",  # Typing only supported in Python 3.9+
         idle: threading.Semaphore,
         name: str = None,
     ):
@@ -67,7 +67,7 @@ class _WorkerThread(threading.Thread):
 
 class WorkerThreadPool:
     def __init__(self, max_workers: int = 40) -> None:
-        self._queue: Queue[Union[_WorkItem, None]] = Queue()
+        self._queue: "Queue[Union[_WorkItem, None]]" = Queue()
         self._workers: Set[_WorkerThread] = set()
         self._max_workers = max_workers
         self._idle = threading.Semaphore(0)
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -135,8 +135,11 @@ class WorkerThreadPool:
 
     def _adjust_worker_count(self):
         """
+        This method should called after work is added to the queue.
+
         If no workers are idle and the maximum worker count is not reached, add a new
-        worker.
+        worker. Otherwise, decrement the idle worker count since work as been added
+        to the queue and a worker will be busy.
 
         Note on cleanup of workers:
             Workers are only removed on shutdown. Workers could be shutdown after a
@@ -145,7 +148,12 @@ class WorkerThreadPool:
             created. As long as the maximum number of workers remains relatively small,
             the overhead of idle workers should be negligable.
         """
-        if not self._idle.acquire(timeout=0) and len(self._workers) < self._max_workers:
+        if (
+            # `acquire` returns false if the idle count is at zero; otherwise, it
+            # decrements the idle count and returns true
+            not self._idle.acquire(blocking=False)
+            and len(self._workers) < self._max_workers
+        ):
             self._add_worker()
 
     def _add_worker(self):
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -118,6 +118,7 @@ class WorkerThreadPool:
         after signalling shutdown to workers.
         """
         async with self._lock:
+            self._shutdown = True
             self._queue.put_nowait(None)
 
             if task_status:
@@ -131,7 +132,6 @@ class WorkerThreadPool:
             await asyncio.gather(*[future.aresult() for future in futures])
 
             self._workers.clear()
-            self._shutdown = True
 
     def _adjust_worker_count(self):
         """
@@ -159,3 +159,9 @@ class WorkerThreadPool:
         )
         self._workers.add(worker)
         worker.start()
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *_):
+        await self.shutdown()
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -145,10 +145,7 @@ class WorkerThreadPool:
             created. As long as the maximum number of workers remains relatively small,
             the overhead of idle workers should be negligable.
         """
-        if (
-            not self._idle.acquire(blocking=False)
-            and len(self._workers) < self._max_workers
-        ):
+        if not self._idle.acquire(timeout=0) and len(self._workers) < self._max_workers:
             self._add_worker()
 
     def _add_worker(self):
--- a/src/prefect/_internal/concurrency/workers.py
+++ b/src/prefect/_internal/concurrency/workers.py
@@ -1,0 +1,161 @@
+import asyncio
+import contextvars
+import dataclasses
+import threading
+import weakref
+from queue import Queue
+from typing import Callable, Dict, Optional, Set, Tuple, TypeVar, Union
+
+import anyio.abc
+from typing_extensions import ParamSpec
+
+from prefect._internal.concurrency.primitives import Future
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+@dataclasses.dataclass
+class _WorkItem:
+    """
+    A representation of work sent to a worker thread.
+    """
+
+    future: Future
+    fn: Callable
+    args: Tuple
+    kwargs: Dict
+    context: contextvars.Context
+
+    def run(self):
+        if not self.future.set_running_or_notify_cancel():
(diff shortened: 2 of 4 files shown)


src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,161 @@
...
+    def run(self) -> None:
+        while True:
+            work_item = self._queue.get()
+            if work_item is None:
+                # Shutdown command received; forward to other workers and exit
+                self._queue.put_nowait(None)
+                return
+
+            self._idle.release()
+            work_item.run()
```
madkinsz, line 63: TODO: We want to mark the worker as idle _after_ running the work.

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+
+            if task_status:
+                task_status.started()
+
+            # Avoid blocking the event loop while waiting for threads to join by
+            # joining in another thread; we use a new instance of ourself to avoid
+            # reimplementing threaded work.
+            pool = WorkerThreadPool(max_workers=1)
+            futures = [await pool.submit(worker.join) for worker in self._workers]
+            await asyncio.gather(*[future.aresult() for future in futures])
```
madkinsz, line 132: This is really meta but it's way easier than spinning a single thread manually just for this purpose.

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+
+    async def submit(
+        self, fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
+    ) -> Future[T]:
+        """
+        Submit a function to run in a worker thread.
+
+        Returns a future which can be used to retrieve the result of the function.
+        """
+        async with self._lock:
```
madkinsz, line 88: Technically, because this function never calls `await` it does not need a lock here. I'm tempted to remove the lock and make a synchronous method, but it feels safer to wait and see what we need in the future.

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
+import asyncio
```
madkinsz, line 1: I am considering renaming this module to `threads.py` to clear the path for process based workers, but think I will also defer that to the future.

tests/_internal/concurrency/test_workers.py
```diff
@@ -0,0 +1,111 @@
...
+    futures = [await pool.submit(identity, 1) for _ in range(5)]
+    await asyncio.gather(*[future.aresult() for future in futures])
+    await pool.shutdown()
+
+
+async def test_shutdown_with_active_worker():
+    pool = WorkerThreadPool()
+    future = await pool.submit(time.sleep, 1)
+    await pool.shutdown()
+    assert await future.aresult() is None
```
abrookins, line 88: I have a mild concern that `aresult()` returning None in the successful case leaves room to mask an unknown failure case that would erroneously lead to the same value. I don't feel strongly about it though.

tests/_internal/concurrency/test_workers.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+async def test_shutdown_exception_during_join():
```
abrookins, line 91: Where does the `join` happen? I'm a little confused about what this test checks.

src/prefect/_internal/concurrency/workers.py
abrookins: The `_internal` path allows us the freedom to do that without worrying about compatibility, right?

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+        super().__init__(name=name)
+        self._queue = queue
+        self._idle = idle
+
+    def run(self) -> None:
+        while True:
+            work_item = self._queue.get()
+            if work_item is None:
+                # Shutdown command received; forward to other workers and exit
+                self._queue.put_nowait(None)
```
abrookins, line 59: I'm wondering about a slightly different approach here, though I haven't thought it through. But what if `WorkerThreadPool` handed a `shutdown_event` (`Event`) to `WorkerThread` on init. The worker checks every iteration of its run loop to see if `shutdown_event` is set, and if it is, `run()` returns. I'm slightly biased toward this approach -- assuming it actually makes sense -- to avoid attaching a semantic value to `None`. Up to you though!

src/prefect/_internal/concurrency/workers.py
abrookins: Hmmm. My instinct would be to remove the lock until we need it, but I'll defer to you here because I'm not sure what keeping it protects us against -- you may have a better idea of that.

src/prefect/_internal/concurrency/workers.py
abrookins: I was confused for a second but then remembered -- yeah, the `ThreadPoolExecutor` interface (and thus ours) is way easier to deal with.

src/prefect/_internal/concurrency/event_loop.py
```diff
@@ -46,4 +56,23 @@ def wrapper() -> None:
                 raise
 
     __loop.call_soon_threadsafe(wrapper)
-    return future.result()
+    return future
```
abrookins, line 59: Can't say I understand what's going on here, but I'm ok with that.

abrookins (approved):
LGTM! Left a few comments, YMMV. 👍

tests/_internal/concurrency/test_workers.py
```diff
@@ -0,0 +1,111 @@
...
+
+async def test_submit_many():
+    async with WorkerThreadPool() as pool:
+        futures = [await pool.submit(identity, i) for i in range(100)]
+        results = await asyncio.gather(*[future.aresult() for future in futures])
+        assert results == list(range(100))
+        assert len(pool._workers) == pool._max_workers
+
+
+async def test_submit_reuses_idle_thread():
```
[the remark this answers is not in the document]
madkinsz: It should never flake in theory :) We can also just sleep for a full second there instead of doing a busy wait, but this is a bit faster. We're just letting Python context switch to call the release method.

tests/_internal/concurrency/test_workers.py
madkinsz: Ah `pool.shutdown` joins all the threads. When called with `start`, it sends the shutdown to signal to works than kills the threads. Here the pattern is a bit like

```
--> Test: Shutdown pool
--> Pool: Sends signal to workers
--> Pool: Awaits on worker join which context switches
--> Test: Raises exception
--> Pool: Shuts down cleanly still
```

I wrote this while dealing with some weird issues with pool shutdown when an exception was raised. It's unclear to me how to clarify it / if it's worth keeping.

src/prefect/_internal/concurrency/workers.py
madkinsz: I've removed the lock in a following pull request — it turns out I need `submit` to be synchronous.

I actually had never considered that if your async function doesn't await anything it doesn't need a lock, this was pointed out to me during review of an httpx PR :)

src/prefect/_internal/concurrency/workers.py
madkinsz: Hm, the issue is that `self._queue.get()` is blocking so the worker will not do anything until it receives something in the queue. We _could_ check an event at the end of each work item but we still need to push something in the queue to wake up all the workers. Both AnyIO's worker threads and the CPython thread pool executor use this model — I trust what they're up to for now :)

src/prefect/_internal/concurrency/workers.py
madkinsz: Yep!

src/prefect/_internal/concurrency/event_loop.py
madkinsz: These utilities are kind of dumb, but basically `AbstractEventLoop.call_soon_threadsafe` returns a `Handle` which is the most useless object around town — you can cancel it and that's it. In cases where we might like, actually want to know what our function returned or wait for the result of our function, we need to return something else. To accomplish this, we wrap the function that we submit to `call_soon_threadsafe` and use a threading `Future` to capture the return value.

madkinsz: I've added some additional documentation for these functions in the next PR

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def run(self):
+        if not self.future.set_running_or_notify_cancel():
+            return
+        try:
+            result = self.context.run(self.fn, *self.args, **self.kwargs)
+        except BaseException as exc:
+            self.future.set_exception(exc)
+            # Prevent reference cycle in `exc`
+            self = None
```
anticorrelator, line 38: I'm confused about how this works

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+                args=args,
+                kwargs=kwargs,
+                context=contextvars.copy_context(),
+            )
+
+            # Place the new work item on the work queue
+            self._queue.put_nowait(work_item)
+
+            # Ensure there are workers available to run the work
+            self._adjust_worker_count()
```
anticorrelator, line 106: would it ever be possible for work to be submitted and error before the worker count is adjusted?

anticorrelator (approved):

src/prefect/_internal/concurrency/workers.py
madkinsz: I don't think the threads will context switch until after submission completes because control isn't yielded, but even if it did I don't think there would be significant effects.

src/prefect/_internal/concurrency/workers.py
madkinsz: https://github.com/python/cpython/issues/80111 may be helpful?

madkinsz: I think the exception captures frame locals so the future references the exception which references the work item which references the future and we have a cycle. If we set `self` to `None` the exception no longer references the work item.

madkinsz:
Thanks for the reviews!

serinamarie:
If we pass in a `timeout` when calling `aresult()` in terms of how long we will wait for a task to complete, would it make sense to have a test around a TimeOut error or would that just never happen?

src/prefect/_internal/concurrency/workers.py
abrookins: Ah, you're right -- I went looking to see how ThreadPoolExecutor handled this and missed its use of None to signal. If it's good enough for them I suppose it's good enough for us! 😂

abrookins:
Good question @serinamarie!

madkinsz:
In `aresult` we pass `timeout=0` to the underlying implementation because we _know_ the future is done (from waiting for the event) and if it's not done for some reason our assumption is wrong and we should crash. If we didn't set the timeout to 0 and waited instead of crashing, we'd sneakily block the event loop.

We could do something like reach into the internals of the `Future` and set the `done_event` without setting a result and we'd get the `TimeOut` error — but generally I try to avoid reaching into the implementation details while writing tests.

src/prefect/_internal/concurrency/workers.py
peytonrunyan: A bit late but I finally feel up to speed with all of the is. Double checking my understanding from https://github.com/python/cpython/blob/1332fdabbab75bc9e4bced064dc4daab2d7acb47/Lib/asyncio/queues.py#L149

`queue.get()` will only ever return `None` after an exception, so we can use it as a signal here?

src/prefect/_internal/concurrency/workers.py
peytonrunyan: Double checking understanding here - this is because we aren't going to be handing execution to another coroutine, right? So as long as this object is only being accessed by by coroutines running within the same thread, and not being accessed by other threads, we don't have to worry about race conditions without the await?

src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,172 @@
...
+            created. As long as the maximum number of workers remains relatively small,
+            the overhead of idle workers should be negligable.
+        """
+        if (
+            # `acquire` returns false if the idle count is at zero; otherwise, it
+            # decrements the idle count and returns true
+            not self._idle.acquire(blocking=False)
+            and len(self._workers) < self._max_workers
+        ):
+            self._add_worker()
```
peytonrunyan, line 157: Double checking understanding here. We've got the semaphore set to `0`. So we go to submit work, `acquire()` will give us back `False`, we check that we have capacity, we spin up a new worker, it does its work, then calls the `release()` incrementing our semaphore, which lets us know we have available workers in the pool, yeah?

src/prefect/_internal/concurrency/workers.py
peytonrunyan: Also double checking - by blocking here, this keeps us burning through CPU on this `while True` loop, yeah?

madkinsz: Yes it stops us from burning CPU. 

It returns `None` when we put `None` in the queue :D  — exceptions are raised. We place `None` in the queue to signal shutdown _and_ to wake up the worker since it is otherwise blocked.

Issue #7800, opened 2022-12-07
Flow run that receives SIGTERM does not go through CRASH detection


madkinsz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Given a flow

```python
@flow
def foo():
    print(get_run_context().flow_run.id)
    print(os.getpid())
    time.sleep(1000)
```

A kill signal

```
kill -TERM 36467 
```

Results in immediate exit of the run

```
10:35:02.131 | INFO    | prefect.engine - Created flow run 'brilliant-chupacabra' for flow 'foo'
7472a4eb-2162-4933-b833-8609d9081a10
36467
zsh: terminated  python example.py
```

### Describe the proposed behavior

This flow should go through our crash handling.

To ensure cancelled flows are not reported as crashed, we should add a rule for transition to CRASHED that checks the parent state for CANCELLED and sets CANCELLED instead

### Example Use

—

madkinsz:
Should help with https://github.com/PrefectHQ/prefect/issues/7732

ddelange:
Hi @madkinsz 👋 

I'll leave some thoughts here. maybe it helps, maybe I can help? either way, here we go:

the use case of a k8s user wanting to (auto)scale up their data pipelines using prefect:
- easiest would be to run `prefect-agent` Deployment behind a HPA (like is currently already possible in the [orion server helm chart](https://github.com/PrefectHQ/prefect-helm/blob/2022.12.22/charts/prefect-orion/values.yaml#L41-L53))
  - different worker Deployments for different queues, depending on average mem usage, cpu-bound/io-bound, etc
- set the `terminationGracePeriodSeconds` of the agent to e.g. ~2x the average duration of a flow run in its queue
  - when a Pod is scaled down in kubernetes, k8s sends a SIGTERM to the container(s) main PID(s). then it waits `terminationGracePeriodSeconds` to send a SIGKILL to get rid of the Pod. this guy [explains it better](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace)
- when the main agent PID receives SIGTERM, allow the flow runs to finish gracefully
- if somehow a flow run is not done after `terminationGracePeriodSeconds`, the main gets a SIGKILL (like when k8s kills the main PID due to OOM).
[...]

PR #7887, opened 2022-12-14
Shutdown engine on SIGTERM

bunchesofdonald:
This adds a hook into the engine's crash handling to handle SIGTERM events. On a SIGTERM a `TerminationSignal` exception is raised, this exception is then caught by the flow run crash handling and not reported as a crash. We also check for this exception when the engine is called directly (`python -m prefect.engine ...`) and ensure that Python's default signal handler is executed.

I began looking at windows support as well, but due to the bug that causes it to crash during a cancellation I was unable to make much progress. 

Related to #7800 

### Checklist


src/prefect/engine.py
```diff
@@ -1592,6 +1600,8 @@ async def report_flow_run_crashes(flow_run: FlowRun, client: OrionClient):
 
         # Reraise the exception
         raise exc from None
+    finally:
+        signal.signal(signal.SIGTERM, original_sigterm_handler)
```
madkinsz, line 1604: Sorry if you covered this already, but what if the SIGTERM is not from Prefect? SIGTERM _should_ be a crash unless the flow is in a CANCELLING state, right?

bunchesofdonald: I elected to have our own handling always be in place, but I could be convinced otherwise. If a true SIGTERM is sent, I didn't see the need to _not_ attempt a graceful shutdown. But if we were to override the SIGKILL handler I think it would be a different story.

src/prefect/engine.py
```diff
@@ -1607,8 +1617,11 @@ async def report_task_run_crashes(task_run: TaskRun, client: OrionClient):
     except (Abort, Pause):
         # Do not capture internal signals as crashes
         raise
+    except Cancel:
+        # Do not capture cancellations as crashes
+        raise
     except BaseException as exc:
-        state = await exception_to_crashed_state(exc)
+        state = await exception_to_final_state(exc)
```
madkinsz, line 1624: Here you ignore `Cancel` but you added handling of `Cancel` to `exception_to_final_state` in https://github.com/PrefectHQ/prefect/pull/7887/files#diff-d96d5698f54cffee300418df379152a03d0f51afedb400bb59849549d95372eeR138-R140 ?

src/prefect/engine.py
madkinsz: I think a graceful shutdown should occur either way, but should we report a CANCELLED state when it was not cancelled by Prefect?

madkinsz: Perhaps we can have an orchestration rule that converts CANCELLED reports to CRASHED if the run is not in a CANCELLING state.

anticorrelator: I remember we discussed at some point in the past whether or not a deliberate SIGTERM should be also considered a form of cancellation, or if CANCELLED should refer explicitly to the prefect functionality—I'm leaning towards the latter because SIGTERM might come from different places?

bunchesofdonald: Is a SIGTERM that causes a graceful shutdown a Crash though?

src/prefect/engine.py
bunchesofdonald: Right, this should probably not capture those.

src/prefect/engine.py
madkinsz: It'll still interrupt the user's work, right? The flow doesn't run to completion. That's a tough question. I am leaning towards the same reasoning as Dustin that a signal causes the flow to exit, when not triggered by Prefect, should not be a CANCELLED state.

bunchesofdonald: :+1: I'll raise a more generic exception if it's not in a CANCELLED state.

madkinsz: I've also reached out to other stakeholders to confirm the behavior here.

madkinsz: I don't think the client should check its state, I think it should just propose the CRASH and the orchestrator can say: Infrastructure teardown during cancellation is expected then REJECT that transition and ensure the run is marked as CANCELLED.

billpalombi: Our states reflect Prefect's view of the world. If a flow run is stopped exogenously, if if we can presume that it was likely intentional, it should be in a CRASHED state. This is similar to Prefect not showing a flow run in a SCHEDULED if it was being triggered by cron, because it doesn't know about that schedule.

bunchesofdonald: Proposing state might be tricky there since sig handlers are sync, but I'll see what we can do.

madkinsz: Crash reporting is already proposing a state — you can just raise an exception using the SIGTERM handler like you are now but not ignore it in crash handling, right?

bunchesofdonald: Yes, I thought you were saying I _shouldn't_ do that. Sorry for the confusion!

madkinsz:
Impressive list of reviewers you have acquired :D

rpeden:
@bunchesofdonald cancellation should work on Windows now. 

Using `SetConcoleCtrlHandler` to catch Ctrl-Break events might be a good way to enable graceful engine shutdown on Windows.

bunchesofdonald:
@madkinsz Have time to take a look at this PR?

tests/test_engine.py
```diff
@@ -1274,6 +1277,27 @@ def my_flow():
...
+        async with report_flow_run_crashes(flow_run=flow_run, client=orion_client):
+            assert signal_receiver.call_args_list[0][0][0] == signal.SIGTERM
+            signal_handler = signal_receiver.call_args_list[0][0][1]
+
+            # Call the signal handler and expect that it raises a
+            # `TerminationSignal` exception.
+            with pytest.raises(TerminationSignal):
+                signal_handler()
+
+        # The original handler should be restorted when the context manager
```
madkinsz, line 1296:
```suggestion
        # The original handler should be restored when the context manager
```

src/prefect/engine.py
```diff
@@ -1906,6 +1921,16 @@ def should_log_prints(flow_or_task: Union[Flow, Task]) -> bool:
...
+    except TerminationSignal as exc:
+        engine_logger.info(
+            f"Engine execution of flow run '{flow_run_id}' cancelled by external signal: {exc}"
+        )
+
+        # Termination signals are swapped out during a flow run to perform a
+        # graceful shutdown and raise this exception. This `os.kill` call
+        # ensures that the previous handler, likely the Python default, gets
+        # called as well.
+        os.kill(os.getpid(), exc.signal)
```
madkinsz, line 1933: This seems like it should be done by `report_flow_run_crashes` — otherwise the original handler is only called for deployed flow runs.

tests/test_engine.py
```diff
@@ -1274,6 +1277,27 @@ def my_flow():
...
+
+        async with report_flow_run_crashes(flow_run=flow_run, client=orion_client):
+            assert signal_receiver.call_args_list[0][0][0] == signal.SIGTERM
+            signal_handler = signal_receiver.call_args_list[0][0][1]
+
+            # Call the signal handler and expect that it raises a
+            # `TerminationSignal` exception.
+            with pytest.raises(TerminationSignal):
+                signal_handler()
+
```
madkinsz, line 1295: I think this test should be updated to register a new signal handler then assert that it is called when the context exits to ensure that the "existing" signal handler is called.

madkinsz (approved):

src/prefect/exceptions.py
```diff
@@ -276,6 +276,21 @@ class Pause(PrefectSignal):
     """
 
 
+class ExternalSignal(BaseException):
```
anticorrelator, line 279: should these be `PrefectSignals`

madkinsz: Maybe? I'm not sure. I think the intent here is to separate this from `PrefectSignal` since we are not the cause of  the signal like for `Abort`.

anticorrelator (approved):
lgtm!

src/prefect/exceptions.py
bunchesofdonald: Yep, that's the intent, just to separate external cause from internal cause.

madkinsz:
Whoa the history for this got really broken... Hopefully this doesn't cause issues with `feature/better-cancellation` into `main`..

madkinsz:
@anticorrelator Did you try to fix this? Should I take a look? I'm not excited about https://github.com/PrefectHQ/prefect/commit/d9591bb322a98f6e19d6e48e556f68a3965e883d including a bunch of changes from `main`, it seems hard to debug if it causes problems in the future.

PR #7896, opened 2022-12-14
Add experimental workers API

jlowin:
This PR introduces the (experimental) Workers API, but only the server-side. A few notes that aren't in the design doc:

- the API must be enabled by setting `PREFECT_EXPERIMENTAL_ENABLE_WORKERS=1`. Otherwise all routes will 404
- routes are behind an `/api/experimental/` endpoint while integration testing with new workers is taking place. Once tested, it will graduate to `/api/beta/`
- (not directly related to this implementation) this introduces a new approach to using Jinja templating for SQL. After we introduced some particularly complex and difficult-to-follow SQLAlchemy code for polling work queues, I took @chrisguidry's feedback to heart and worked to come up with a way to have complex SQL that was comprehensible but still allowed customization. After some discussion, we came to the Jinja template approach used here, which is far more readable than its SQLA equivalent (though granted is not vanilla SQL). See [here](https://github.com/PrefectHQ/prefect/pull/7896/files#diff-cb23528272342beea00109a4adc7de76516ec30a5b3c7e30efe6735cff9ce15a) for an example template and [here](https://github.com/PrefectHQ/prefect/pull/7896/files#diff-83a82323b73c752f97ddf5dee1e0204af7277fafd6503d8111d991bb618c19a8R294-R306) for its use in code.

image


src/prefect/orion/database/migrations/versions/postgresql/2022_11_24_143620_f7587d6c5776_add_worker_tables.py
```diff
@@ -0,0 +1,239 @@
...
+        ),
+        sa.Column(
+            "updated",
+            prefect.orion.utilities.database.Timestamp(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("description", sa.String(), nullable=True),
+        sa.Column("type", sa.String(), nullable=True),
```
cicdw, line 44: I believe we'll need a column for the base job definition (which implicitly will define the `type`)

src/prefect/orion/database/migrations/versions/postgresql/2022_11_24_143620_f7587d6c5776_add_worker_tables.py
```diff
@@ -0,0 +1,239 @@
...
+        ),
+        sa.Column(
+            "updated",
+            prefect.orion.utilities.database.Timestamp(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column(
+            "last_heartbeat_time",
```
cicdw, line 80: The worker table feels a bit heavy to me, especially if we're going to be updating it with every poll, but I understand why it's helpful to have

cicdw (commented):
exciting stuff

desertaxle (approved):
The concepts in the PR and their relationships make a lot of sense to me. I have some questions around worker types, but I'll direct those to the design doc.

anticorrelator (approved):

PR #7897, opened 2022-12-15
Improve `Event` implementation consistency and fix deadlock

madkinsz:
Follow-up to #7865 — extends https://github.com/PrefectHQ/prefect/pull/7875 

I attempted to simplify the `Event` implementation by inheriting from `asyncio.Event`, but it turns out there is significant variation in how they handle binding to event loops across our supported Python versions. I went with a backport of the latest CPython implementation and cleaned up our comments covering the differences.

We no longer bind an event to a single loop at startup. Instead, each `wait()` call creates a future bound to the current event loop. Then, during `set()` calls, we will notify each waiter. This is nearly verbatim the implementation from CPython, with thread safety added in.

The primary motivation for revisiting the `Event` implementation was a hard to find deadlock that occurred when two threaded event loops waited on each other. The offending line 😠 was:

https://github.com/PrefectHQ/prefect/blob/3aa447a5bea28d58907bb1dae30f1c8b3ff98217/src/prefect/_internal/concurrency/primitives.py#L36

This line blocked `Event.set()` calls until the loop got around to running it. For example:
- Thread A creates an event and passes it to Thread B to set when something is done
- Thread A waits for Thread B to finish (without calling `await` and unblocking the event loop)
- Thread B sets the event, _this blocks until Thread A notifies listeners_
- Thread A is busy waiting for Thread B, so we deadlock

`test_events_in_two_loops_do_not_deadlock` reproduces this scenario.

By updating the problematic line to use "call soon" instead, setting an event no longer blocks and there's no deadlock :) This implementation is significantly faster as well.

I've also taken this opportunity to clarify when loop utilities are safe to use by adding documentation.

### DIFF
--- a/src/prefect/_internal/concurrency/event_loop.py
+++ b/src/prefect/_internal/concurrency/event_loop.py
@@ -1,5 +1,5 @@
 """
-Utilities for working with asynchronous event loops.
+Thread-safe utilities for working with asynchronous event loops.
 """
 
 import asyncio
@@ -33,6 +33,10 @@ def call_in_loop(
 ) -> T:
     """
     Run a synchronous call in event loop's thread from another thread.
+
+    This function is blocking and not safe to call from an asynchronous context.
+
+    Returns the result of the call.
     """
     future = call_soon_in_loop(__loop, __fn, *args, **kwargs)
     return future.result()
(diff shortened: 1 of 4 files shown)


madkinsz:
Take that GH Actions!

src/prefect/_internal/concurrency/event_loop.py
```diff
@@ -33,6 +33,10 @@ def call_in_loop(
 ) -> T:
     """
     Run a synchronous call in event loop's thread from another thread.
+
+    This function is blocking and not safe to call from an asynchronous context.
+
+    Returns the result of the call.
     """
     future = call_soon_in_loop(__loop, __fn, *args, **kwargs)
```
serinamarie, line 41: So, is the change here that instead of `call_soon_in_loop` always calling `call_soon_threadsafe`, `call_soon_in_loop` determines whether we're in the main thread and if not then would call `call_soon_threadsafe` to schedule the callback?

madkinsz: This is a bit of a minor change I snuck in here, it doesn't actually change any behavior relevant for this pull request (shame on me). There's basically `loop.call_soon` or `loop.call_soon_threadsafe` — one's for using from the event loop thread and the other let's you schedule things from other threads. By checking if the current event loop is the one you're scheduling something on, we can switch to the threadsafe implementation when needed and use the normal one by default — this avoids whatever synchronization overhead they use to make the call threadsafe when we don't need it.

madkinsz: I'm talking about the change in https://github.com/PrefectHQ/prefect/pull/7897/files#diff-1d13a991aaa8d66f948979d55a7e7dad20ee565c9a94a88e0c80aaa25cfd741cR71-R76 btw.

The fix for the deadlock in this PR is calling `call_soon_in_loop` at https://github.com/PrefectHQ/prefect/pull/7897/files#diff-fb1d7293ffe4010ef6082fa1d4c5235a52a8129e6e33542c808f2edde3332ce7R53 instead of calling `call_in_loop` — which blocks until completion and can fail when there are dependent waiters.

serinamarie: > By checking if the current event loop is the one you're scheduling something on, we can switch to the threadsafe implementation when needed but use the unsafe one otherwise — this avoids whatever synchronization overhead they use to make the call threadsafe when we don't need it.

Got it, that's about what I thought the reasoning was 🧐

PR #7900, opened 2022-12-15
Build multi-arch images for commits on main

ddelange:
On commits to `main`, push:

- `prefecthq/prefect-dev:main-python3.X`
- `prefecthq/prefect-dev:sha-424242-python3.X`
- `prefecthq/prefect-dev:main-python3.X-conda`
- `prefecthq/prefect-dev:sha-424242-python3.X-conda`

### Example

Steps 1-3 ref https://github.com/PrefectHQ/prefect/pull/7740#discussion_r1048803518 for multi arch docker images, towards https://github.com/PrefectHQ/prefect/issues/5388

### Checklist


.github/workflows/docker-builds.yaml
```diff
@@ -0,0 +1,82 @@
...
+          fetch-depth: 0
+
+      - name: Generate image metadata
+        id: metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: prefecthq/prefect-dev
+          # first tag generated here will be used in CACHE_IMAGE below
+          tags: |
+            type=raw,enable=${{ github.event_name == 'release' }},value=${{ github.ref_name }},suffix=-python${{ matrix.python-version }}
```
madkinsz, line 57: This will be skipped since this is on push/dispatch for now, right?

madkinsz:
Thanks for splitting this up! It's much easier to review.

.github/workflows/docker-builds.yaml
ddelange: thanks for pointing this one out, this is not intended. it should be negated, because `ref_name` points to a tag on release event and not a branch. The `git rev-parse` setup from before would return `HEAD` because a release event is a detached branch so is also unusable. it seems to be hard to get out which branch the release was made from.

```suggestion
            type=raw,enable=${{ github.event_name != 'release' }},value=${{ github.ref_name }},suffix=-python${{ matrix.python-version }}
```

this means there will be no cache hit on a release event, because the CACHE_IMAGE would point to the sha pattern instead of the branch pattern, and we dont have `buildcache` for the sha pattern. fix would be to hardcode a buildcache tag in the `cache-from` and `cache-to` based on `github.matrix`. so not getting it from `metadata-action`.

.github/workflows/docker-builds.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
ddelange: we solved this problem with a mount directive: https://github.com/pypa/setuptools_scm#usage-from-docker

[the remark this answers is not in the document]
ddelange: although that does void all subsequent layer caches, so we have to do it as late as possible 😅

.github/workflows/docker-builds.yaml
madkinsz: I don't think we need to know what branch releases are made from, we can just always pull from the cache on `main`?

.github/workflows/docker-builds.yaml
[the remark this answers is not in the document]
madkinsz: Hm we just copy the `.git` folder in during the build step which works fine. The issue is that we cannot retag these images on release, we'll need to rebuild them.

.github/workflows/docker-builds.yaml
```diff
@@ -0,0 +1,82 @@
...
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          build-args: |
+            PYTHON_VERSION=${{ matrix.python-version }}
+            ${{ ( endsWith(matrix.flavor, 'conda') && 'BASE_IMAGE=prefect-conda' ) || '' }}
+          tags: ${{ steps.metadata.outputs.tags }}
+          labels: ${{ steps.metadata.outputs.labels }}
+          push: true
```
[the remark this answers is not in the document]
madkinsz: Makes sense. I've seen that before, but is that best practice?
cc @jawnsy

.github/workflows/docker-builds.yaml
ddelange: yes fine by me, will hardcode :+1:

.github/workflows/docker-builds.yaml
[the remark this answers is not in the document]
ddelange: ah gotcha. yes copying `.git` might be unwanted layer size if you have a lot of history. setuptos_scm also needs `fetch-depth: 0` btw... ref https://github.com/pypa/setuptools_scm/issues/480

they do however provide a 'pretend' environment variable setup you could combine with `${{ github.ref_name}}` to 'solve' the problem

.github/workflows/docker-builds.yaml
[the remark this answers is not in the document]
ddelange: if the statement doesn't have filesystem side-effects, it should be a (near) empty layer.

best practise I don't think so, but it's a good smoke test to have before pushing and afaik the github actions provide no alternative mechanic ;(

.github/workflows/docker-builds.yaml
ddelange: [157dd9b](https://github.com/PrefectHQ/prefect/pull/7900/commits/157dd9bfae12ef20ef673f3f9e000d40e070bf14)

.github/workflows/docker-builds.yaml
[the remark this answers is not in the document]
ddelange: [157dd9b](https://github.com/PrefectHQ/prefect/pull/7900/commits/157dd9bfae12ef20ef673f3f9e000d40e070bf14)

.github/workflows/docker-builds.yaml
```diff
@@ -0,0 +1,75 @@
...
+          # multi-stage cache manifests (mode=max) need a separate tag because they are not usable for runtime ref https://github.com/moby/buildkit#export-cache
+          cache-from: type=registry,ref=prefecthq/prefect-dev:buildcache-python${{ matrix.python-version }}${{ matrix.flavor }}
+          cache-to: type=registry,ref=prefecthq/prefect-dev:buildcache-python${{ matrix.python-version }}${{ matrix.flavor }},mode=max
```
ddelange, line 75: could consider only updating buildcache when the event is push (to `main`):

```suggestion
          cache-to: ${{ ( github.event_name == 'push' && 'type=registry,ref=prefecthq/prefect-dev:buildcache-python${{ matrix.python-version }}${{ matrix.flavor }},mode=max' ) || '' }}
```

madkinsz:
@jawnsy would you be able to take a look at this and investigate Docker credentials that can only push to `prefect-dev`?

ddelange:
fwiw I can see no real security concerns with using the same token for all image pushes without manual approval. the action only pushes to production on a `release` event, and that can only be changed if a malicious PR was merged into `main`

madkinsz:
I've now added DockerHub credentials scoped to `prefect-dev` to the "dev" GitHub Actions Environment.

.github/workflows/docker-builds.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # On each commit merged into main, push sha and branch patterns
+  push:
+    branches: main
+  # On workflow_dispatch, push sha and branch patterns
+  workflow_dispatch:
+
+jobs:
+  publish-docker-images:
+    name: Build Docker images
+    runs-on: ubuntu-latest
```
madkinsz, line 15: I believe you'll need "environment: dev" like we do for the prod environment https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/release.yaml#L186

ddelange: good catch! https://github.com/PrefectHQ/prefect/pull/7900/commits/d6defda08b8d0224e9347ca0aafd9b66044a9364

madkinsz (approved):
Sweet thanks! Let's give this a try :)

ddelange:
fix for failing `main` incoming https://github.com/ddelange/prefect/actions/runs/3875949170/jobs/6609182214#step:7:1

PR #7901, opened 2022-12-15
Build multi-arch images for releases

ddelange:
On releases from `main`, additionally push:

- `prefecthq/prefect:2-python3.X`
- `prefecthq/prefect:2.6-python3.X`
- `prefecthq/prefect:2.6.9-python3.X`
- `prefecthq/prefect:2-latest`
- `prefecthq/prefect:2-python3.X-conda`
- `prefecthq/prefect:2.6-python3.X-conda`
- `prefecthq/prefect:2.6.9-python3.X-conda`
- `prefecthq/prefect:2-latest-conda`

### Example

One commit on top of #7900

Steps 4-6 ref https://github.com/PrefectHQ/prefect/pull/7740#discussion_r1048803518 for multi arch docker images, towards https://github.com/PrefectHQ/prefect/issues/5388

### Checklist

### DIFF
--- a/.github/workflows/docker-builds.yaml
+++ b/.github/workflows/docker-builds.yaml
@@ -1,0 +1,85 @@
+name: Build prefect docker images
+
+# Note: Conda support for 3.11 is pending. See https://github.com/ContinuumIO/anaconda-issues/issues/13082
+
+on:
+  # On each commit merged into main, push sha and branch patterns
+  push:
+    branches: main
+  # On workflow_dispatch, push sha and branch patterns
+  workflow_dispatch:
+
+jobs:
+  publish-docker-images:
+    name: Build Docker images
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flavor:
+          - ""
+          - "-conda"
+        python-version:
+          - "3.7"
+          - "3.8"
+          - "3.9"
+          - "3.10"
+          - "3.11"
+        exclude:
+          # Not yet supported, see note at top
+          - flavor: "-conda"
+            python-version: "3.11"
+
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Export CURRENT_BRANCH env var
+        # can't use github.ref_name because that will be e.g. '2.6.9' on tag push, and we want to cache-from main branch on release
+        run: echo "CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's/\//-/g')" >> $GITHUB_ENV
+
+      - name: Generate image metadata
+        id: metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: prefecthq/prefect-dev
+          # list CURRENT_BRANCH first, to use as CACHE_IMAGE below
+          tags: |
+            type=raw,value=${{ env.CURRENT_BRANCH }},suffix=-python${{ matrix.python-version }}${{ matrix.flavor }}
+            type=sha,suffix=-python${{ matrix.python-version }}${{ matrix.flavor }}
+          flavor: |
+            latest=false
+
+      - name: Export CACHE_IMAGE env var
+        run: echo "CACHE_IMAGE=$(echo '${{ steps.metadata.outputs.json }}' | jq -r '.tags[0]')" >> $GITHUB_ENV
+
+      - name: Build image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          build-args: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-builds.yaml
```diff
@@ -60,6 +63,20 @@ jobs:
             latest=false
             suffix=${{ matrix.flavor }}
 
+      - name: Generate tags for prefecthq/prefect
+        id: metadata-prod
+        uses: docker/metadata-action@v4
+        with:
+          images: prefecthq/prefect,enable=${{ github.event_name == 'release' }}
```
madkinsz, line 70: Might be worth highlighting this is what prevents this from running on development builds.

Should we cover `prereleased` here as well? I'm not sure if we have a use-case for that.

ddelange: both `released` and `prereleased` event types will pass this conditional (checking which one of both it is, is another topic)

how is this?
```suggestion
          # only generate the production tags on release events
          images: prefecthq/prefect,enable=${{ github.event_name == 'release' }}
```

madkinsz: 👍 

Slightly different note: We'll not want to apply the `latest` tag for prereleases.

.github/workflows/docker-builds.yaml
```diff
@@ -60,6 +63,20 @@ jobs:
...
+          tags: |
+            type=raw,enable=${{ matrix.python-version == '3.10' }},value=2-latest
```
ddelange, line 72:
```suggestion
            type=raw,enable=${{ matrix.python-version == '3.10' && github.event.release.prerelease == 'false' }},value=2-latest
```
https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release

ddelange: this relies on the maintainer ticking the 'this is a prerelease' button before publishing the release @madkinsz

madkinsz: That makes sense to me — there's also the `python{x}-latest` tags.

While it's on my mind, we'll eventually want the ability to release patches for old releases which means we'll need to check if the release is marked as the latest too 🤯 there's not an attribute for this so we'll probably need to make an API call for the latest release and see if the tags match. What a pain. I don't think that needs to happen now though.

madkinsz: @ddelange that's okay with me — seems like a pain to parse the tags.

ddelange: > there's also the `python{x}-latest` tags.

did I fail to migrate properly? looks like missing here

ddelange: > release patches for old releases

do you mean upstream patches? like `python:3.9` patched a security hole?

ddelange: which reminds me we might want to set `pull: true` for exactly this reason, otherwise I guess it would eternally be using the base image as included in `buildcache`

.github/workflows/docker-builds.yaml
```diff
@@ -70,8 +87,8 @@ jobs:
           build-args: |
             PYTHON_VERSION=${{ matrix.python-version }}
             ${{ ( endsWith(matrix.flavor, 'conda') && 'BASE_IMAGE=prefect-conda' ) || '' }}
-          tags: ${{ steps.metadata.outputs.tags }}
-          labels: ${{ steps.metadata.outputs.labels }}
+          tags: ${{ join(steps.metadata-dev.outputs.tags) }},${{ join(steps.metadata-prod.outputs.tags) }}
+          labels: ${{ steps.metadata-dev.outputs.labels }}
           push: true
```
ddelange, line 92:
```suggestion
          push: true
          pull: true
```

.github/workflows/docker-builds.yaml
madkinsz: > do you mean upstream patches? like python:3.9 patched a security hole?

I mean if we are on 2.7.7 but I want to release 2.6.10 with a backported fix.

madkinsz: But also yeah that other case is one we care about too! We may want to update the base image periodically for old releases.

madkinsz: > did I fail to migrate properly? looks like missing here

I guess it's `prefect:2-python3.x` i.e. https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/release.yaml#L84

ddelange: ok, this last one is included.

as for the other topic, I'm sure there is a not-too-ugly way of getting this info... maybe the build-push-action `latest` mechanic does these things out of the box 🤔

for now will apply the suggestions, feel free to unresolve of course 👍

ddelange:
I couldn't resist https://stackoverflow.com/a/75079768/5511061

madkinsz:
> I couldn't resist https://stackoverflow.com/a/75079768/5511061

Oooh nice!

So, just so we can move forward with confidence here I'd like to do this in a few steps:

1. Make a slight modification to this so it does not push any images
2. Merge to `main` and verify on the next release that it does everything we intend
3. Remove the existing Docker publishing and enable this one in another pull request

ddelange:
SGTM, will set `push: false` now, and once this merges, will update #7902 and re-enable it there.

.github/workflows/docker-builds.yaml
```diff
@@ -70,9 +100,10 @@ jobs:
...
```
ddelange, line 105:
```suggestion
          push: false
```

madkinsz (approved):

PR #7902, opened 2022-12-15
Consolidate multi-arch docker builds

ddelange:
- Move docker builds from `release.yaml` to `docker-builds.yaml`
- Delete `republish-images.yaml`:
  - there are no longer single-arch artifacts uploaded.
  - re-trigger by flipping the `this is a prerelease` checkbox on the Github Release

### Example

One commit on top of #7901, diff: https://github.com/ddelange/prefect/compare/docker-builds-release...docker-builds-consolidation

Steps 7-8 ref https://github.com/PrefectHQ/prefect/pull/7740#discussion_r1048803518 for multi arch docker images, closes https://github.com/PrefectHQ/prefect/issues/5388

### Checklist


madkinsz:
Hm, there are some warnings on the latest merge: https://github.com/PrefectHQ/prefect/actions/runs/3950856517

madkinsz:
We've got an example run from a release now: https://github.com/PrefectHQ/prefect/actions/runs/3962751282/jobs/6789813811

ddelange:
nice, the build command in L143 LGTM!

sinve recent versions of the action, there is also `--provenance` being pushed unless disabled, an interesting [feature](https://docs.docker.com/build/attestations/slsa-provenance/) that might break some dev/custom setups ref https://github.com/docker/buildx/issues/1509#issuecomment-1378454396 especially ones that use `docker manifest` commands (my comment further down).

I don't think it causes trouble for `docker pull` though, otherwise you'd probably already have seen a bug report of some sort.

.github/workflows/release.yaml
```diff
@@ -35,175 +31,26 @@ jobs:
 
       - name: Install python packages
         run: |
-          python -m pip install --upgrade pip
-          pip install --upgrade --upgrade-strategy eager -e .[dev]
+          python -m pip install -U pip setuptools wheel
+          pip install -U -e .[dev]
```
madkinsz, line 35: Can you revert these changes? I don't think they belong in this pull request.

.github/workflows/release.yaml
```diff
@@ -226,7 +73,7 @@ jobs:
         run: git checkout -b "docs-for-release-${GITHUB_REF#refs/*/}"
 
       - name: Merge with main
-        run: git merge origin/main --no-ff -Xtheirs -m "Merge branch 'main' into temporary docs branch"
+        run: git merge origin/main --no-ff -m "Merge branch 'main' into temporary docs branch"
```
madkinsz, line 76: Can you revert this change? This merge strategy is needed and the change doesn't belong here anyway.

ddelange: ouch something went wrong with merge conflicts. on it

madkinsz: No problem! Thanks :)

madkinsz (approved):
Let's give it a go :) Thanks for your hard work on this!

PR #7905, opened 2022-12-15
eslint config update and warnings fixed

stackoverfloweth:
updated to latest [@prefecthq/eslint-config](https://www.npmjs.com/package/@prefecthq/eslint-config)
applied fixes with `npm run lint:fix`
fixed 1 not auto-fixable error manually (meta/ts string | undefined)

### DIFF
--- a/orion-ui/.eslintrc.js
+++ b/orion-ui/.eslintrc.js
@@ -4,9 +4,12 @@ module.exports = {
     node: true
   },
   extends: ["@prefecthq"],
+  parser: 'vue-eslint-parser',
   parserOptions: {
     ecmaVersion: 2020,
-    project: "./tsconfig.json"
+    project: ["./tsconfig.json"],
+    sourceType: 'module',
+    parser: '@typescript-eslint/parser'
   },
   settings: {
     "import/resolver": {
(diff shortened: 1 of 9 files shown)


orion-ui/.eslintrc.js
```diff
@@ -4,9 +4,12 @@ module.exports = {
     node: true
   },
   extends: ["@prefecthq"],
+  parser: 'vue-eslint-parser',
   parserOptions: {
     ecmaVersion: 2020,
-    project: "./tsconfig.json"
+    project: ["./tsconfig.json"],
+    sourceType: 'module',
+    parser: '@typescript-eslint/parser'
```
pleek91, line 12: Can you give a brief explanation of these changes? Just wanting to understand why these were needed when they weren't before.

pleek91 (approved):
❤️

orion-ui/.eslintrc.js
stackoverfloweth: the important thing here is we're explicitly using the `vue-eslint-parser`, which has a parser option of the underlying parser. 

without this change, running `npm run lint` would error because the rules we specify required "project" set in parser options. With some time on github issues, I discovered this solution.

pleek91: Nice, thanks!

PR #7906, opened 2022-12-15
Add documentation for pausing flow runs

tpdorsey:
Adds a new "Pause a flow run" section to Flows documentation documenting flow run pause and resume.

### Checklist

### DIFF
--- a/docs/concepts/flows.md
+++ b/docs/concepts/flows.md
@@ -608,10 +608,102 @@ Completed(message=None, type=COMPLETED, result='foo', flow_run_id=7240e6f5-f0a8-
 ```
 </div>
 
+## Pause a flow run
 
-## Flow run cancellation
+Prefect enables pausing an in-progress flow run for manual approval. Prefect exposes this functionality via the [`pause_flow_run`](/api-ref/prefect/engine/#prefect.engine.pause_flow_run) and [`resume_flow_run`](/api-ref/prefect/engine/#prefect.engine.resume_flow_run) functions, as well as via the Prefect Orion UI and Prefect Cloud. 
 
-Flow runs can be cancelled from the CLI, UI, REST API, or Python client. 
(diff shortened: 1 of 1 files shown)


docs/concepts/flows.md
```diff
@@ -608,10 +608,102 @@ Completed(message=None, type=COMPLETED, result='foo', flow_run_id=7240e6f5-f0a8-
...
+
+@task
+async def marvin_setup():
+    return "a raft of ducks walk into a bar..."
+@task
+async def marvin_punchline():
+    return "it's a wonder none of them ducked!"
+@flow
+async def inspiring_joke():
+    await marvin_set()
```
anticorrelator, line 628:
```suggestion
    await marvin_setup()
```

docs/concepts/flows.md
```diff
@@ -608,10 +608,102 @@ Completed(message=None, type=COMPLETED, result='foo', flow_run_id=7240e6f5-f0a8-
...
+
+The paused flow run will then finish!
+
+<div class="terminal">
+```
+> "it's a wonder none of them ducked!"
+```
+</div>
+
+Here is an example of a flow that stops executing in a paused state after one task, and will be rescheduled upon resuming. The stored result of the first task is retrieved instead of being rerun.
```
anticorrelator, line 656:
```suggestion
Here is an example of a flow that does not block flow execution while paused. This flow will exit after one task, and will be rescheduled upon resuming. The stored result of the first task is retrieved instead of being rerun.
```

docs/concepts/flows.md
```diff
@@ -608,10 +608,102 @@ Completed(message=None, type=COMPLETED, result='foo', flow_run_id=7240e6f5-f0a8-
...
+
+```python
+from prefect import flow, pause_flow_run, task
+
+@task
+def foo():
+    return 42
+
+@flow
```
anticorrelator, line 665:
```suggestion
@flow(persist_result=True)
```

anticorrelator (changes requested):
Just a few small changes for correctness, but otherwise this looks great!

anticorrelator (approved):

PR #7919, opened 2022-12-16
Ignore vcs directories git, hg (mercurial) in .prefectignore

quassy:
vcs directories tend to be relatively big and make deployment slow, so they should be ignored by default. Added `.git` and `.hg` here as those are the ones I use.

Right now every deployment of a repository would also upload it's version control history although that's generally not needed or expected. This makes uploads quite a bit faster & cheaper.

I'm not aware of any already existing issue I could link here and it's a small fix.

### Example

Deploy a flow e.g. with to observe the time difference and also check the storage to see there is less data saved:

```
time prefect deployment build foobar.py:foobar --name dev -ib kubernetes-job/foobar-dev -sb gcs/foobar-dev -q kubernetes -a
```

### Checklist

### DIFF
--- a/src/prefect/.prefectignore
+++ b/src/prefect/.prefectignore
@@ -35,3 +35,7 @@ dask-worker-space/
 # Editors
 .idea/
 .vscode/
+
+# VCS
+.git/
+.hg/


madkinsz:
Makes sense to me. @cicdw?

madkinsz (approved):

PR #7928, opened 2022-12-16
Cleanup deployment-start-time feature branch, add tests

serinamarie:
Continuation of https://github.com/PrefectHQ/prefect/pull/7772

### Notes
`dateparser.parse` has a default `PREFER_DATES_FROM` setting of `current_period`. This means that when parsing a date, it will go backwards in time if that's closer than a future time, e.g. if it is currently 5pm, then `dateparser.parse("1pm")` will retrieve a date from earlier today (in the past) because this 1pm is closer than tomorrow at 1pm.

By changing the `PREFER_DATES_FROM` setting to `future` like in this PR, when parsing a date, it will always go forward in time regardless of closeness, e.g. if it is currently 5pm, then `dateparser.parse("1pm")` will retrieve the next future instance of 1pm, meaning tomorrow at 1pm even though today at 1pm is closer in proximity.

We reasoned that most of the time, users want to run deployments at a future date, so we should consistently parse for a future date, rather than inconsistently provide a past date or future date based on proximity.

If needed someday, we could add a setting to enable other behavior.

### Checklist


tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
...
+    assert len(flow_runs) == 1
+    flow_run = flow_runs[0]
+
+    assert flow_run.state.is_scheduled()
+    scheduled_time = flow_run.state.state_details.scheduled_time
+
+    assert scheduled_time == expected_start_time
+
+
+@pytest.mark.parametrize(
```
serinamarie, line 113: Wondering about testing `expected_display`, in case pendulum made adjustments to `format_diff`.

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
+import dateparser
```
serinamarie, line 1: Thinking about adding more tests

tests/cli/deployment/test_deployment_run.py
```diff
@@ -81,7 +110,45 @@ async def test_start_in_option_schedules_flow_run_in_future(
...
+        command=[
+            "deployment",
+            "run",
+            deployment_name,
+            "--start-in",
+            start_in,
+        ],
+        expected_output_contains=["Scheduled start time:", expected_display],
+    )
+
```
serinamarie, line 143: lns 144-148 are probably beyond the scope of this test

requirements.txt
```diff
@@ -9,7 +9,7 @@ cloudpickle >= 2.0
 coolname >= 1.0.4
 croniter >= 1.0.12
 cryptography >= 36.0.1
-dateparser >= 1.0
+dateparser >= 1.1.2
```
madkinsz, line 12: Why do we need the higher version?

src/prefect/cli/deployment.py
```diff
@@ -411,31 +412,44 @@ async def run(
...
 
-    if start_time_raw:
+    elif start_in is None and start_at is None:
+        scheduled_start_time = now
+        human_dt_diff = " (now)"
+    else:
+        if start_in:
+            start_time_raw = "in " + start_in
+        else:
+            start_time_raw = "at " + start_at
```
madkinsz, line 427: Since the above exits if both are set you can just do
```python
if start_in: ... 
elif start_at: ...
else: ...
```

src/prefect/cli/deployment.py
```diff
@@ -411,31 +412,44 @@ async def run(
...
+                    start_time_raw,
+                    settings={
+                        "TO_TIMEZONE": "UTC",
+                        "RELATIVE_BASE": datetime.fromtimestamp(now.timestamp()),
+                    },
+                )
+            except Exception as exc:
+                exit_with_error(
+                    f"Exception occurred while parsing '{start_time_raw}': {exc!r}"
+                )
```
madkinsz, line 444: I'd lean towards "Failed to parse ..." for user-facing messages. When we say "Exception occurred" we'll usually want to include a traceback.

I'd also probably use the string for the exception rather than the repr. What kind of exceptions occur here / with what sort of messages?

src/prefect/cli/deployment.py
```diff
@@ -474,21 +488,17 @@ async def run(
...
             ).to_datetime_string()
-            + " ("
-            + pendulum.format_diff(scheduled_start_time.diff(now))
-            + ")"
+            + human_dt_diff
+        )
+    except Exception as exc:
+        exit_with_error(
+            f"Unable to format scheduled start time display from '{scheduled_start_time}': {exc!r}"
         )
```
madkinsz, line 501: We probably shouldn't have a try/except here, we should always be able to format a date time object correctly. Is there a case where this can happen due to weird user input?

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
...
+            start_at,
+        ],
+        expected_output_contains=["Scheduled start time:", expected_start_time],
+    )
+
+    flow_runs = await orion_client.read_flow_runs()
+    assert len(flow_runs) == 1
+    flow_run = flow_runs[0]
+
+    assert flow_run.state.is_scheduled()
```
madkinsz, line 64: There's no assertion here that the flow run is actually scheduled correctly

tests/cli/deployment/test_deployment_run.py
madkinsz: If pendulum changes we can update the test suite, generally we don't guard against possible future changes

tests/cli/deployment/test_deployment_run.py
madkinsz: Yeah I think you could exclude those

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
...
+    await run_sync_in_worker_thread(
+        invoke_and_assert,
+        command=[
+            "deployment",
+            "run",
+            deployment_name,
+            "--start-in",
+            start_in,
+        ],
+        expected_output_contains=f"Scheduled start time: {expected_display}",
```
madkinsz, line 100: It seems like this part could be covered by `test_start_at_displays_scheduled_start_time`

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
...
+    await run_sync_in_worker_thread(
+        invoke_and_assert,
+        command=[
+            "deployment",
+            "run",
+            deployment_name,
+            "--start-at",
+            start_at,
+        ],
+        expected_output_contains=["Scheduled start time:", expected_start_time],
```
madkinsz, line 57: It seems like this part could be covered by `test_start_at_displays_scheduled_start_time`

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,212 @@
...
+            deployment_name,
+            "--start-at",
+            "foobar",
+        ],
+        expected_code=1,
+        expected_output="Unable to parse scheduled start time 'at foobar'.",
+    )
+
+
+def test_start_in_invalid_input(deployment_name: str):
```
madkinsz, line 182: We may want to try covering some more invalid input types here, like malformed timestamps.

src/prefect/cli/deployment.py
serinamarie: Do you mean `else:` being where `start_in` is None and `start_as` in None? Because there is some logic beneath that specific to `start_in`/`start_at` for assigning `start_time_parsed`, `scheduled_start_time`, and `human_dt_diff`.

src/prefect/cli/deployment.py
serinamarie: Sounds good. These are the exceptions that could occur according to their documentation:

> ValueError: Unknown Language, TypeError: Languages argument must be a list, SettingValidationError: A provided setting is not valid.

Based on that, should we expect that `start_time_parsed` will always be set to `None` if it's parsed incorrectly and that these errors are too unlikely to be concerned about?

src/prefect/cli/deployment.py
serinamarie: I'm not sure I've exhausted the full variety of possible user input through testing/to what extent pendulum can handle all of the reformatting.

tests/cli/deployment/test_deployment_run.py
serinamarie: I'll leave them in favor of [your other comment](https://github.com/PrefectHQ/prefect/pull/7928#discussion_r1052522906)

src/prefect/cli/deployment.py
madkinsz: Yeah I think you can do `else` instead of `elif start_in is None and start_at is None:`

src/prefect/cli/deployment.py
madkinsz: We're in control over the language / settings being passed in, so I think that these errors don't need to be captured — there's nothing the user can do if these occur.

src/prefect/cli/deployment.py
```diff
@@ -439,14 +439,18 @@ async def run(
...
-                )
+                exit_with_error(f"Failed to parse '{start_time_raw!r}': {exc!s}")
 
         if start_time_parsed is None:
             exit_with_error(f"Unable to parse scheduled start time {start_time_raw!r}.")
 
-        scheduled_start_time = pendulum.instance(start_time_parsed)
+        if start_time_parsed.tzinfo:
+            exit_with_error(
+                f"Timezone not expected in '{start_time_raw!r}'. Timezone is inferred with pendulum.tz.local_timezone()."
```
madkinsz, line 449: From a UX perspective, we should definitely let them provide a time zone. We may even want to display in that time zone instead of the local timezone if it's explicitly given.

serinamarie: It’s looking like allowing users to specify a time to start-at such as “5pm PST” is going to be difficult because there isn't consistent mapping/meaning for those abbreviations. Alternatively, asking users to specify a locality-based tz seems not to be a good UX (“America/New York”). Another option could be requiring an offset like “5pm UTC-5” which is probably better than nothing. What do you think?

src/prefect/cli/deployment.py
serinamarie: I'm not sure what you mean, since the `else` on ln 423 handles cases where `start_at` or `start_in` is set. If neither are set, we wouldn't want to do anything from ln 423-452.

requirements.txt
serinamarie: Tests were failing: https://github.com/PrefectHQ/prefect/actions/runs/3752658214/jobs/6375066682#step:12:236
I determined the cause to be: https://github.com/scrapinghub/dateparser/issues/1045 and the solution was implemented here: https://github.com/scrapinghub/dateparser/pull/1046.

src/prefect/cli/deployment.py
```diff
@@ -384,7 +384,8 @@ async def run(
     """
     Create a flow run for the given flow and deployment.
 
-    The flow run will be scheduled for now and an agent must execute it.
+    The flow run will be scheduled for now if start-in or start-at are unspecified;
+    and an agent must execute it.
```
peytonrunyan, line 388: This would sentence kind of confuses me. Maybe:

The flow run will be scheduled to run immediately if `start-in` or `start-at` are not specified. The scheduled run can only be executed by an agent.

src/prefect/cli/deployment.py
```diff
@@ -411,31 +412,46 @@ async def run(
...
-    # Parse the start time if provided
-    if start_in:
-        start_time_raw = "in " + start_in
-    elif start_at:
-        start_time_raw = "at " + start_at
-    else:
-        start_time_raw = None
+    if start_in and start_at:
+        exit_with_error(
+            "Expected optional start_in field or start_at field but not both."
```
peytonrunyan, line 417: Maybe put quotes or ticks around `start_in` and `start_at`?

src/prefect/cli/deployment.py
peytonrunyan: I'm kinda of curious what this looks like to without having a helper function for the raw parsing logic.

peytonrunyan (approved):
A couple minor ux nits (not strong feelings) but otherwise lgtm!

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,318 @@
...
+            "1-1-2020",
+            DateTime(2020, 1, 1, 0, 0, 0, tzinfo=pendulum.tz.local_timezone()),
+        ),
+        (
+            "5 June 2015",
+            DateTime(2015, 6, 5, 0, 0, 0, tzinfo=pendulum.tz.local_timezone()),
+        ),
+    ],
+)
+async def test_start_at_option_schedules_flow_run_in_future(
```
madkinsz, line 155: Might want to rename this test since some of these are not scheduled in the future

tests/cli/deployment/test_deployment_run.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def test_start_in_option_invalid_input(deployment_name: str):
+    invoke_and_assert(
+        command=[
+            "deployment",
+            "run",
+            deployment_name,
+            "--start-in",
+            "foobar",
```
madkinsz, line 237: Is this the only invalid input worth covering?

madkinsz: What happens if I give a date like I was using "start at"?

src/prefect/cli/deployment.py
```diff
@@ -384,8 +384,7 @@ async def run(
     """
     Create a flow run for the given flow and deployment.
 
-    The flow run will be scheduled for now and an agent must execute it.
-
+    The flow run will be scheduled to run immediately if start-in or start-at are unspecified.
```
madkinsz, line 387:
```suggestion
    The flow run will be scheduled to run immediately unless `--start-in` or `--start-at` is specified.
```

src/prefect/cli/deployment.py
madkinsz: I think the helper function is a bit harder to read than the initial implementation, especially since it does more than one thing (parses the date time _and_ generates a formatted diff). I don't have strong feelings here though.

madkinsz: (unresolving for visibility; does not mean it needs to be addressed)

src/prefect/cli/deployment.py
```diff
@@ -411,31 +410,57 @@ async def run(
...
+
         scheduled_start_time = pendulum.instance(start_time_parsed)
+        human_dt_diff = (
+            " (" + pendulum.format_diff(scheduled_start_time.diff(now)) + ")"
+        )
+        return scheduled_start_time, human_dt_diff
+
+    if start_in and start_at:
+        exit_with_error(
+            "Expected optional `start_in` field or `start_at` field but not both."
```
madkinsz, line 445: We should use the user-facing options here not our internal variable names
```suggestion
            "Only one of `--start-in` or `--start-at` can be set, not both."
```

tests/cli/deployment/test_deployment_run.py
serinamarie: Passing a date to start_in is successful and I believe has the same result as when you pass a date to start_at.

serinamarie: Which is why it was more difficult to think of other cases, but I can pass it invalid dates, at least.

madkinsz: If passing a start-at date to start-in results in scheduling for that date, we might want to add a test confirming that behavior — you could add a note that says it is not by design

serinamarie: Sounds good! I've added such a test

madkinsz (approved):

PR #7931, opened 2022-12-16
Remove worker pool in favor of reusing standard library executors

madkinsz:
I learned a lot from #7875 and #7880, but I started to get concerned about maintaining our own pool implementation and dove into the CPython implementation to determine if we could build on top of the standard library instead of duplicating it in a simpler form. This let's us use multiprocessing without implementing that executor ourselves, which would truly be a daunting task and is definitely out of scope for us. This will also let us use any `concurrent.Futures` compatible executor when needed, which could be nice :)

I'll still need to use some queue/work item patterns for the upcoming "runtime" pull request (#7942).

### Checklist


peytonrunyan:
LGTM!

peytonrunyan (approved):

madkinsz:
Thank you for reviewing!

PR #7942, opened 2022-12-19
Add `Runtime` for centralized management of concurrent work

madkinsz:
Follow-up to https://github.com/PrefectHQ/prefect/pull/7931

Adds a `Runtime` object. When its context is entered, it creates:
- A pool of worker threads
- A pool of worker processes
- A thread with an event loop

It then provides methods to submit a call to any of the above (`run_in_thread`, `run_in_process`, `run_in_loop`). When a call is sent elsewhere, we create a queue that allows the call to send work _back_ the main thread. The main thread (or the thread that owns the runtime object) watches for work in the queue and executes it. Once the submitted call is complete, the main thread stops watching for more work.

This will allow us to run user flows and tasks in the main thread by default then leverage multithreaded or multiprocess concurrency as needed. The dedicated event loop thread will be used for Prefect orchestration. The worker pools support running asynchronous user code.

### Example

For example, here we submit an orchestration function to run on an event loop thread then the user's function is run on the main thread.

```python
def my_function(x):
    return x

async def orchestrate_run():
    future = runtime.submit_from_thread(my_function, 1)
    result = await asyncio.wrap_future(future)
    return result

with Runtime() as runtime:
    assert runtime.run_in_loop(orchestrate_run) == 1
```

### Checklist


src/prefect/_internal/concurrency/executor.py
```diff
@@ -2,23 +2,16 @@
...
 T = TypeVar("T")
 P = ParamSpec("P")
 WorkerType = Literal["thread", "process"]
 
-threadlocals = threading.local()
-
-
-def _initialize_worker():
-    threadlocals.is_worker = True
-
```
madkinsz, line 21: This has moved to the `runtime` module which will manage worker contexts

src/prefect/_internal/concurrency/executor.py
```diff
@@ -43,28 +36,22 @@ class Executor(BaseExecutor):
     def __init__(
         self,
         worker_type: WorkerType = "thread",
-        max_workers: Optional[int] = None,
+        **worker_pool_kwargs: Any,
```
peytonrunyan, line 39: I'm guessing this is to accommodate the different workers, which seems reasonable enough, but `max_workers` is going to be universal across executors right? It feels like it might be nice to have it documented as an explicit parameter.

madkinsz: The only reason I take "thread"/"process" explicitly (instead of a `BaseExecutor` type) is because we need to know if context vars can be copied over successfully. `BaseExecutor` does not define any `__init__` arguments so for future compatibility with other executor types I'm hesitant to define the parameters here.

serinamarie (approved):

Issue #14, opened 2018-06-30
Add exponential backoff to task retries


jlowin:

evan-burke:
+1, I use this a lot via https://github.com/litl/backoff.

jameslamb:
I saw this has a `needs description`. I just came here because my first thought reading https://docs.prefect.io/core/tutorial/04-handling-failure.html#if-at-first-you-don-t-succeed was "I wonder if `prefect` also supports exponential backoff". I'd like propose a description, hope it helps:

-------------

## Description of  the problem

It's common for tasks that pull data from external systems to fail because of temporary unavailability of those systems. Those systems may, for example, include REST APIs, databases, file systems, or message brokers.

Some of the root causes for that unavailability may be unaffected by consumers putting load on the system, including:

* temporary downtime due to maintenance like OS upgrades
* temporary downtime due to data migrations
* unreliable connection due to poor internet connectivity

In those cases, it makes sense to wait a bit and retry the tasks. `prefect` offers an easy way to do this with arguments `max_retries` and `retry_delay` to `@task()`. These arguments allow you to say "if this task fails, try it `n` more times and wait `s` seconds between each retry".

However, this type of logic could defeat its own purpose if the unavailability is because of a problem which is made worse by new load from consumers, including:

* resource exhaustion
    - example: 100% CPU utilization caused by a large number of requests or a few very-expensive requests (e.g. database queries that result in a full table scan)
* quota limits reached
    - example: many relational databases cap the number of connections which can be open simultaneously
* routing bottleneck
    - example: service sit behind a load balancer and the load balancer is receiving new requests faster than it can route them

In these situations, it's advisable to use a retry strategy which waits longer and longer after each failure. If all consumers do this, it will give the external system a better chance to recover. For one example reference (there are many), see [this AWS blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).

## What it might look like to add this to `prefect`

Implementing exponential backoff means that the amount of time to wait between retries is a function of the number of retries so far.

It might look like this pseudocode:

```python
# what is the shortest time you're willing to wait after a failed attempt?
wait_min = 1

# what is the longest you're willing to wait between retries?
wait_max = 10

# how fast do you want waiting time to scale relative to number of attempts?
wait_base = 2

# what is the most times you're willing to retry before saying a task failed?
max_attempts = 5

keep_trying = True
num_attempts_so_far = 0
while keep_retrying:
    result = task.run()
[...]

PR #7954, opened 2022-12-20
Implement configurable retry delays and exponential backoff

anticorrelator:
Closes https://github.com/PrefectHQ/prefect/issues/14 🙃 

Implements configurable task retry delays and exponential backoff. Also allows clamped jitter to be added to retry delays to avoid cascading correlated failures.

### Example

The `task` constructor can now be passed a list of retry delays to customize retry behavior. The following task will wait for successively increasing intervals before attempting to retry on failure.

```python
import random

@task(retries=3, retry_delay_seconds=[1, 10, 100])
def flaky_function():
    if random.choice([True, False]):
        raise RuntimeError("not this time!")
    return 42
```

Additionally, users can use the `exponential_backoff` utility to automatically configure retry delays that follow the exponential backoff pattern used by [urllib3](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html). Furthermore, an optional `retry_jitter_factor` can be passed. When supplied, the backoff will be jittered up to an amount specified by the jitter factor: a factor of `.5` corresponds to a maximum delay of `1.5 * base_delay`.

The following task will wait for about 10, 20, then 40 seconds with each retry, however those delay times will be jittered up to a maximum of 20, 40 and 80 seconds respectively.

```python
from prefect.tasks import exponential_backoff
import random

@task(retries=3, retry_delay_seconds=exponential_backoff(backoff_factor=10), retry_jitter_factor=1)
def flaky_function():
    if random.choice([True, False]):
        raise RuntimeError("not this time!")
    return 42
```

### Checklist


src/prefect/orion/orchestration/core_policy.py
```diff
@@ -432,9 +445,11 @@ async def before_transition(
         if not scheduled_time:
             return
 
-        # At this moment, we take the floor of the actual delay as the API schema
+        # At this moment, we round delay to the nearest second as the API schema
         # specifies an integer return value.
```
madkinsz, line 449: ! Are we losing an important part of the jitter here? Should we update the API to return a fl.oat? The client can sleep on float timescales.

src/prefect/tasks.py
```diff
@@ -77,6 +77,25 @@ def task_input_hash(
     )
 
 
+def exponential_backoff(backoff_factor: float) -> Callable:
```
madkinsz, line 80:
```suggestion
def exponential_backoff(backoff_factor: float) -> Callable[[int], List[float]]:
```

src/prefect/tasks.py
```diff
@@ -139,7 +164,13 @@ def __init__(
         ] = None,
         cache_expiration: datetime.timedelta = None,
         retries: int = 0,
-        retry_delay_seconds: Union[float, int] = 0,
+        retry_delay_seconds: Union[
+            float,
+            int,
+            List[float],
+            Callable[["TaskRunContext", Dict[str, Any]], List[float]],
```
madkinsz, line 171:
```suggestion
            Callable[[int], List[float]],
```

src/prefect/tasks.py
```diff
@@ -221,7 +253,13 @@ def with_options(
         ] = None,
         cache_expiration: datetime.timedelta = None,
         retries: Optional[int] = NotSet,
-        retry_delay_seconds: Optional[Union[float, int]] = NotSet,
+        retry_delay_seconds: Union[
+            float,
+            int,
+            List[float],
+            Callable[["TaskRunContext", Dict[str, Any]], List[float]],
```
madkinsz, line 260:
```suggestion
            Callable[[int], List[float]],
```

src/prefect/tasks.py
```diff
@@ -756,7 +805,13 @@ def task(
     cache_key_fn: Callable[["TaskRunContext", Dict[str, Any]], Optional[str]] = None,
     cache_expiration: datetime.timedelta = None,
     retries: int = 0,
-    retry_delay_seconds: Union[float, int] = 0,
+    retry_delay_seconds: Union[
+        float,
+        int,
+        List[float],
+        Callable[["TaskRunContext", Dict[str, Any]], List[float]],
```
madkinsz, line 812: Perhaps we should move this full `Union[...]` annotation into a variable e.g. `RetryDelayType = Union[...]` to avoid repeating this big union.

```suggestion
        Callable[[int], List[float]],
```

src/prefect/tasks.py
```diff
@@ -777,7 +832,13 @@ def task(
...
```
madkinsz, line 839: I've made it this far, might as well keep suggesting
```suggestion
        Callable[[int], List[float]],
```

tests/test_engine.py
```diff
@@ -637,6 +639,157 @@ def flaky_function():
...
+            wait_for=None,
+            result_factory=result_factory,
+            interruptible=False,
+            client=orion_client,
+            log_prints=False,
+        )
+
+        assert mock_anyio_sleep.await_count == 10
+        sleeps = [c.args[0] for c in mock_anyio_sleep.await_args_list]
+        assert statistics.variance(sleeps) > 0
```
madkinsz, line 787: ooo.

src/prefect/tasks.py
anticorrelator: whoops

src/prefect/orion/orchestration/core_policy.py
anticorrelator: I think we should but I don't think it's critical if we'd like to discuss this more

billpalombi (approved):
Confirming that this behavior is consistent with expectations. Thanks Dustin!

madkinsz:
But like... how do you keep the average delay fixed if the jitter factor is greater than one?

anticorrelator:
> But like... how do you keep the average delay fixed if the jitter factor is greater than one?

@madkinsz That's a really good question—we're drawing our jittered delays from an asymmetric distribution where for every upper bound greater than the average (up to infinity, in principle), we can find a lower bound greater than zero (but still less than the average) that keeps the average of random values drawn from this distribution fixed. This is a property that symmetric distributions (e.g. Gaussian and uniform) don't share, given that our delay times must be greater than 0. An added benefit to this approach is that we're drawing from an exponential distribution, so for reasonable values of the `retry_jitter_factor` we start to approximate a Poisson process for jittered retry arrival times, giving us the best chance to avoid "thundering herd" scenarios.

src/prefect/orion/schemas/core.py
```diff
@@ -314,11 +318,30 @@ def populate_deprecated_fields(cls, values):
...
+    @root_validator
+    def validate_retry_and_jitter_inputs(cls, values):
+        """
+        Limits the number of configured retry delays to a maximum of 50. And ensures
+        that the relative jitter factor is positive.
+        """
+        if isinstance(values["retry_delay"], list) and (
+            len(values["retry_delay"]) > 50
+        ):
+            values["retry_delay"] = values["retry_delay"][:50]
```
madkinsz, line 339: Do we want to silently ignore the extra items? I feel like we should raise an error?

src/prefect/orion/schemas/core.py
```diff
@@ -314,11 +318,30 @@ def populate_deprecated_fields(cls, values):
...
+
+        configured_relative_jitter = values.get("retry_jitter_factor", None)
+        if configured_relative_jitter is not None and configured_relative_jitter < 0:
+            values["retry_jitter_factor"] = None
```
madkinsz, line 343: Is this enforcing that it should be greater than zero? We should also have this raise an error.

src/prefect/orion/schemas/core.py
```diff
@@ -314,11 +318,30 @@ def populate_deprecated_fields(cls, values):
...
+
         if (
             not values.get("retry_delay", None)
             and values.get("retry_delay_seconds", 0) != 0
         ):
             values["retry_delay"] = values["retry_delay_seconds"]
+
+        return values
+
+    @root_validator
```
madkinsz, line 330: Can you use a `validator` instead of a `root_validator`? Since these validations are for single fields (without interactions) we don't need to use a root validator.

src/prefect/tasks.py
```diff
@@ -805,8 +879,14 @@ def task(
...
             never expire.
         retries: An optional number of times to retry on task run failure
-        retry_delay_seconds: An optional number of seconds to wait before retrying the
-            task after failure. This is only applicable if `retries` is nonzero.
+        retry_delay_seconds: Optionally configures how long to wait before retrying the
+            task after failure. This is only applicable if `retries` is nonzero. This
+            setting can either be a number of seconds, a list of retry delays, or a
+            callable that, given the total number of retries, generates a list of retry
+            delays. If a number of seconds, that delay will be applied to all retries.
+            If a list, each retry will wait for the corresponding delay before retrying.
```
madkinsz, line 887: Let's document the maximum list length here

src/prefect/tasks.py
```diff
@@ -77,6 +77,25 @@ def task_input_hash(
...
+    Arguments:
+        backoff_factor: the base delay for the first retry, subsequent retries will
+            increase the delay time by powers of 2.
+
+    Returns:
+        a callable that can be passed to the task constructor
+    """
+
+    def retry_backoff_callable(retries: int) -> List[float]:
+        return [backoff_factor * max(0, 2**r) for r in range(retries)]
```
madkinsz, line 94: Let's limit this to our maximum length so our own utility doesn't go out of bounds.

src/prefect/orion/schemas/core.py
madkinsz: I could be convinced that we should have this truncate and just document it, but an error seems safer to start.

anticorrelator: I've noticed patterns where we'll "ignore" misconfigured thing in the past on the backend, I don't personally have a principled stance here

src/prefect/orion/schemas/core.py
anticorrelator: this isn't doing the enforcing, but I figured we might want to fully orchestrate otherwise valid task runs. If we remove the check the actual guard in `clamped_poisson_interval` will raise a `ValueError`

src/prefect/orion/schemas/core.py
madkinsz: Hm weird. I think we should avoid that wherever possible, maybe we'll need to audit some of the existing patterns.

bunchesofdonald (approved):
I didn't understand the math at all, but the logic seem solid.

Issue #7927, opened 2022-12-16
Boolean parameter overriden via deployment can't be overriden from UI


hateyouinfinity:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Boolean parameter overriden via deployment can't be overriden from UI.

If you
1. Create a flow that has a boolean parameter with a default value set
2. Override that parameter when creating a deployment from that flow
3. Override that parameter (again) when creating a flow run from UI

the override made from UI will be ignored.

**Example:**

See _Reproduction_ for the deployment source code. <br>
Parameters set via deployment override:
![](https://i.imgur.com/0FXabBB.png) <br>
Creating a flow run with custom parameters:
![](https://i.imgur.com/hhfD5V9.png) <br>
Overrides to some of the parameters are ignored:
![](https://i.imgur.com/TrFeUGa.png) ![](https://i.imgur.com/AARgMya.png) <br>

#### Scope

Affected (AFAIU):
- UI overrides to boolean parameters which have a flow-level default value overriden via deployment

Not affected:
- Non-boolean parameters (at least ints and strings, see `descr` above)
- Boolean parameters w/o a flow-level default value (see `bool_no_default` above)
- Overrides done via API

### Reproduction

```python3
import asyncio

from prefect import flow, get_client, get_run_logger
from prefect.deployments import Deployment


@flow(name="overrides")
def flowfunc(
    bool_no_default: bool,
    bool_first: bool = False,
    bool_second: bool = True,
    descr: str = "no overrides (Flow)",
):
    logger = get_run_logger()
    logger.info(
        f"bool_no_default={bool_no_default}\nbool_first={bool_first}\nbool_second={bool_second}\ndescr={descr}"
    )


async def deploy_overrides():
    depl = await Deployment.build_from_flow(
        flow=flowfunc,
        name="overrides",
        parameters={
            "bool_no_default": True,
            "bool_first": True,
            "bool_second": False,
            "descr": "set bool_no_default=True, bool_first=True, bool_second=False (Deployment)",
        },
        apply=True,
    )


async def create_run_with_double_overrides():
    client = get_client()
    depl = await client.read_deployment_by_name(name="overrides/overrides")
    await client.create_flow_run_from_deployment(
        deployment_id=depl.id,
        parameters={
[...]

PR #7955, opened 2022-12-20
Enhancement: Deployment parameter behavior (RFC)

znicholasbrown:
This PR seeks to address #7927, which demonstrated 2 issues with the current deployments API. The first is that the `parameter_openapi_schema` field reflects deployment build-time parameters of the _flow_ and not the deployment; this means that the schema defaults are not a source of truth and in fact will be wrong when any default parameter is updated on the deployment model. The second issue, related to this, is that the deployment parameter schema is never actually used when creating runs, only the `parameters` blob on the deployment model is used. This means that if a deployment `parameters` blob omits any parameter, the parameter will not be passed to the run, even when a default is defined in the schema. 

This is an issue particularly when a deployment `parameters` field is updated from clients such as the UI, which must use the parameter schema to determine extant parameter values at run time and cannot rely on the `parameters` blob to show forms. In these cases the UI does not send updated parameter values because it has no good way to determine if omitted values should be truly omitted or included as empty e.g. when do we include a boolean parameter set to `False` when it wasn't included in the `parameter` blob but has a default value of `False` in the schema? 

Proposal:
1. Update the `parameter_openapi_schema` whenever deployment parameter defaults are updated; this would mean that the deployment's parameter schema would become a source of truth for how runs are created (implemented in this PR)
2. Capture deployment schema defaults when creating runs from a deployment and use those as the basis for runtime parameters, instead the deployment `parameters` blob (implemented in this PR)
3. Drop the `parameters` blob on the deployments model; if the deployment schema is the new source of truth, the `parameters` blob on the deployment is no longer needed and should be dropped (not implemented)

There are definitely other ways to accomplish this so I'm open to closing this in favor of better solutions. One option is to remove all default values from the schema and treat the `parameters` blob as the source of truth at all times. I was hesitant to do that in this PR because I don't have a good sense of the implications elsewhere. 

cc @pleek91

### DIFF
--- a/src/prefect/orion/api/deployments.py
+++ b/src/prefect/orion/api/deployments.py
@@ -17,6 +17,7 @@ from prefect.orion.database.interface import OrionDBInterface
 from prefect.orion.exceptions import ObjectNotFoundError
 from prefect.orion.utilities.schemas import DateTimeTZ
 from prefect.orion.utilities.server import OrionRouter
+from prefect.utilities.callables import get_default_values_from_parameter_schema_dict
 
 router = OrionRouter(prefix="/deployments", tags=["Deployments"])
 
@@ -304,7 +305,7 @@ async def create_flow_run_from_deployment(
     """
     Create a flow run from a deployment.
 
-    Any parameters not provided will be inferred from the deployment's parameters.
+    Any parameters not provided will be inferred from the deployment's parameter schema.
     If tags are not provided, the deployment's tags will be used.
 
     If no state is provided, the flow run will be created in a PENDING state.
@@ -320,7 +321,7 @@ async def create_flow_run_from_deployment(
                 status_code=status.HTTP_404_NOT_FOUND, detail="Deployment not found"
             )
 
-        parameters = deployment.parameters
+        parameters = get_default_values_from_parameter_schema_dict(deployment.parameter_openapi_schema)
         parameters.update(flow_run.parameters or {})
 
         # hydrate the input model into a full flow run / state model
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -11,17 +11,11 @@ from pydantic import create_model
 import pydantic.schema
 from typing_extensions import Literal
 
-from prefect.logging.loggers import (
-    get_logger,
-)
-
 from prefect.exceptions import (
     ParameterBindError,
     ReservedArgumentError,
     SignatureMismatchError,
 )
-
-callable_logger = get_logger("callables")
 
 
 def get_call_parameters(
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -151,10 +151,10 @@ def parameter_schema(fn: Callable, overrides: dict = {}) -> ParameterSchema:
 
         # If an override is provided, use that, otherwise use the default value
         # from the flow definition
-        parameter_default = ...
+        parameter_default = None
         if overrides.get(param.name) is not None:
             parameter_default = overrides.get(param.name)
-        elif param.default is not None:
+        elif param.default is not param.empty:
             parameter_default = param.default
 
         type_, field = (
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -7,14 +7,21 @@ from typing import Any, Callable, Dict, Iterable, List, Tuple
 
 import cloudpickle
 import pydantic
+from pydantic import create_model
 import pydantic.schema
 from typing_extensions import Literal
+
+from prefect.logging.loggers import (
+    get_logger,
+)
 
 from prefect.exceptions import (
     ParameterBindError,
     ReservedArgumentError,
     SignatureMismatchError,
 )
+
+callable_logger = get_logger("callables")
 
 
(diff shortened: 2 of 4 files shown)


src/prefect/utilities/callables.py
```diff
@@ -119,7 +119,7 @@ def dict(self, *args, **kwargs):
         return super().dict(*args, **kwargs)
 
 
-def parameter_schema(fn: Callable) -> ParameterSchema:
+def parameter_schema(fn: Callable, overrides: dict = {}) -> ParameterSchema:
```
madkinsz, line 122: Let's be explicit that these are `default_overrides`

src/prefect/utilities/callables.py
```diff
@@ -149,10 +149,18 @@ class ModelConfig:
         else:
             name = param.name
 
+        # If an override is provided, use that, otherwise use the default value
+        # from the flow definition
+        parameter_default = None
+        if overrides.get(param.name) is not None:
```
madkinsz, line 155: We should use `if param.name in overrides` instead — otherwise we're not letting you override with a value of `None`

src/prefect/utilities/callables.py
```diff
@@ -149,10 +149,18 @@ class ModelConfig:
         else:
             name = param.name
 
+        # If an override is provided, use that, otherwise use the default value
+        # from the flow definition
+        parameter_default = None
```
madkinsz, line 154: We should still probably represent `empty` with `...` instead of `None` — the meanings are distinct.

src/prefect/utilities/callables.py
```diff
@@ -176,6 +184,16 @@ class ModelConfig:
     schema = model.schema(by_alias=True)
     return ParameterSchema(**schema)
 
+def get_default_values_from_parameter_schema_dict(schema: Dict[str, Any]) -> Dict[str, Any]:
+    """Given a parameter schema, return a dictionary of the default values for each
+    parameter."""
+    values = {}
+    properties = schema.get("properties", {})
+    for key in properties:
+        values[key] = properties.get(key).get('default', None)
```
madkinsz, line 193: Per the earlier note about the semantic deference between `None` and empty — we should not pass values for parameters without a value; this is distinct from passing a value of `None`.

src/prefect/orion/api/deployments.py
```diff
@@ -320,7 +321,7 @@ async def create_flow_run_from_deployment(
                 status_code=status.HTTP_404_NOT_FOUND, detail="Deployment not found"
             )
 
-        parameters = deployment.parameters
+        parameters = get_default_values_from_parameter_schema_dict(deployment.parameter_openapi_schema)
```
madkinsz, line 324: Maybe we should just have a `get_default_parameters()` method on the `Deployment` schema itself.

madkinsz (commented):
Thanks for opening this and investigating the issue! I'm a little confused about a couple points so let's make sure we're on the same page. It's probably not worth addressing any of my review notes until afterwards :)

> parameter_openapi_schema field reflects deployment build-time parameters of the flow and not the deployment; 

This is intentional. This field is meant to encode the parameters of the flow.

> this means that the schema defaults are not a source of truth and in fact will be wrong when any default parameter is updated on the deployment model.

This is also okay, the deployment itself can override the default value of the flow. The flow's schema does not change when the value for a parameter changes. 

> The second issue, related to this, is that the deployment parameter schema is never actually used when creating runs,

Thusfar, this has been intentional. The parameter schema is reflecting our current understanding of the flow's accepted parameters. If the flow has changed but a deployment has not been updated, the schema could be out of date. We only send the parameters set on the deployment to avoid sending stale data.

Here's where I'm confused:

> the UI does not send updated parameter values because it has no good way to determine if omitted values should be truly omitted or included as empty e.g. when do we include a boolean parameter set to False when it wasn't included in the parameter blob but has a default value of False in the schema?

If a user sets a parameter value for a deployment, we should update the deployment's parameters to have that value.

If a user sets a parameter value for a flow run, we should explicitly use that value.

In both of these cases, I'm not sure why the schema's default value would be relevant. The schema is just a hint as to what would happen if the value was omitted entirely. Perhaps the UI needs an option that says "Use flow default" which excludes the parameter entirely?

znicholasbrown:
Ah, sorry, let me add on to that statement:

> Here's where I'm confused:

>> the UI does not send updated parameter values _when they match the schema default_ because it has no good way to determine if omitted values should be truly omitted or included as empty e.g. when do we include a boolean parameter ~set to False~ when it wasn't included in the parameter blob _(no default was included in initial build of the deployment)_ but has a default value of False in the schema?

We discussed this at length around the deployment updates earlier this (last?) year for this exact reason but the UI was deemed a bit confusing. Might be worth revisiting because it could definitely solve this. 

>  Perhaps the UI needs an option that says "Use flow default" which excludes the parameter entirely?

znicholasbrown:
Let me see if I can come up with a minimal UI for this now that we have more visual components to use

znicholasbrown:
The UI has a bit of tech debt around this issue that might make it clearer how we can handle some of the ambiguity around how data in deployment parameter forms (and other open-api schema-based forms) is handled. Closing this at least until that can be addressed

PR #7958, opened 2022-12-21
Add deployment id support to `run_deployment`

### LINKED ISSUE #7259: Ability to run_deployment by providing deployment id instead of deployment name
### First check

### Prefect Version

2.x

Right now the [`run_deployment()`](https://orion-docs.prefect.io/api-ref/prefect/deployments/#prefect.deployments.run_deployment) function is really useful when we want to start a deployed flow run from the Python API. However it requires us to provide the deployment name in the in the form "/".

It would be really helpful if we had the ability to also call [`run_deployment()`](https://orion-docs.prefect.io/api-ref/prefect/deployments/#prefect.deployments.run_deployment) by providing the **deployment id** instead of the deployment name

### Example Use

_No response_

### Additional context

_No response_

devanshdoshi9:
### Updated Overview: 9 Mar 2023
Enhances `run_deployment()` to take a uuid or str uuid of the deployment instead of just string name `flow-name/deployment-name`. This enhancement will allow users to use either the deployment name or deployment id in order to run their flow. Closes #7259.

### Updated Examples: 9 Mar 2023
```python
from prefect.deployments import run_deployment

def example1():
    response = run_deployment(deployment_id=<deployment-id-as-str>)
    print(response)

if __name__ == "__main__":
   example1()
```

```python
from prefect.deployments import run_deployment

def example2():
    response = run_deployment(deployment_id=<deployment-id-uuid>)
    print(response)

if __name__ == "__main__":
   example2()
```
-----------------------
### Overview (outdated)
Added a new `run_deployment_with_id()` function that takes uuid of the deployment as a string instead of `flow-name/deployment-name` used by `run_deployment()`. This enhancement will allow users to use either the deployment name or deployment id in order to run their flow. Closes #7259.

I am yet to write test cases for this change but just wanted to share the current progress to make sure this is what was expected. Will add tests for this upon confirmation.

### Example (outdated)
```python
from prefect.deployments import run_deployment_by_id

def main():
    response = run_deployment_by_id(deployment_id=<id-as-string>)
    print(response)

if __name__ == "__main__":
   main()
```

### Checklist

- [x] This pull request references any related issue by including "closes #7259"


madkinsz (changes requested):
Hi! Thanks for the pull request.

Thanks for contributing!

This makes sense for me functionality wise, but it makes some extra unnecessary API calls to retrieve the name and exposes a second interface to our users. Instead, it'd be nice to have this included in the `run_deployment` interface directly. Let's check if the given "name" is a UUID and perform the lookup from there.

src/prefect/deployments.py
```diff
@@ -41,7 +41,7 @@
 @sync_compatible
 @inject_client
 async def run_deployment(
-    name: str,
+    name_or_id: str,
```
madkinsz, line 44: We cannot change the name of the parameter, as that is a breaking change. Instead we should update the docstring to reflect that this can be an id.

src/prefect/deployments.py
```diff
@@ -80,7 +80,12 @@ async def run_deployment(
 
     parameters = parameters or {}
 
-    deployment = await client.read_deployment_by_name(name)
+    deployment_id = None
+    try:
+        deployment_id = UUID(name_or_id)
+        deployment = await client.read_deployment(deployment_id=deployment_id)
+    except ValueError:
+        deployment = await client.read_deployment_by_name(name_or_id)
```
madkinsz, line 88: It'd be better to call `read_deployment` outside of the `try/except` block. 

1. If `read_deployment` throws a `ValueError` this code will incorrectly try again by name which should be confusing to debug.
2. If `read_deployment_by_name` throws an error we'll get a weird traceback where an exception occurred while handling an exception.

The `try/except` should just be scoped to the attempt to parse the name into a UUID.

madkinsz (changes requested):
Thanks for making those changes!

A couple more requests before this is ready to merge.

devanshdoshi9:
Hi, I've made the changes that you requested. Can you please take a look at it? Also, a few checks are failing and I couldn't figure out the reason for that so I needed some help with that as well.

madkinsz:
@serinamarie could you please take over adding tests for this pull request so we can get it merged?

serinamarie:
> @serinamarie could you please take over adding tests for this pull request so we can get it merged?

Definitely

serinamarie:
I'm wondering if it would be necessary to parametrize the other tests around `run_deployment` with `deployment_id`.

tests/test_deployments.py
```diff
@@ -756,6 +756,22 @@ async def test_run_deployment_with_ephemeral_api(
...
 
+    async def test_run_deployment_with_deployment_id(
+        self,
+        test_deployment,
+        orion_client,
+    ):
+        d, deployment_id = test_deployment
+
+        flow_run = await run_deployment(
+            f"{deployment_id}",
```
madkinsz, line 767: Why a string instead of a UUID? Should we test both?

serinamarie: We can test both if we can take both! I've adjusted the code to allow for a UUID

madkinsz (approved):

PR #7968, opened 2022-12-22
Disable integration tests

pleek91:


znicholasbrown (approved):
Discussed offline - disabling playright tests while we investigate a global scope leak causing node environments to fail

Issue #6990, opened 2022-09-27
Add `prefect server status`


zanieb:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

To perform a healthcheck, manual API requests must be made.

### Describe the proposed behavior

We should provide a CLI command to perform healthchecks against the configured server: `prefect server status`

This command should return a 0 or 1 status code depending on the availability of the server. It can also display information about the server if it is available.

This command should include a `--wait` flag that will attempt to connect until the service responds with a healthy value.

This command should include a `--timeout <seconds>` option that will fail the healthcheck after a total number of seconds.

### Example Use

This is especially useful in combination with `--detach` as proposed in #6989 

```
prefect orion start --detach
prefect orion status --wait --timeout 60
```

Possible output

```
$ prefect orion status
Connecting to <address>...
Server available.
...
```

Also useful when run in a Docker container

### Additional context

This would replace the script at https://github.com/PrefectHQ/prefect/blob/main/scripts/wait-for-server.py

jawnsy:
I'd propose a `prefect orion status` command instead, which would print information about the running instance, but also allow people to suppress output (in case they only care about the status code, such as for readinessProbe checks).

This is how Vault handles their readiness probes: https://github.com/hashicorp/vault-helm/blob/7e21a09ebd9fb7890c5dfe90c4716b50d1662637/templates/server-statefulset.yaml#L160-L161

madkinsz:
👍 sounds better to me!

serinamarie:
Hi @masonmenges, looks like this was assigned to you awhile ago. Shall I unassign?

masonmenges:
@serinamarie That should be fine, if time permits I might take a look at this again but I haven't looked at this in a while.

PR #7970, opened 2022-12-22
Add Orion status cli command 

masonmenges:
This PR introduces a new status command for Orion through the cli. The command runs a health check against the currently configured PREFECT_API_URL based on the users profile. included are an optional "--wait" command that will continue to check the api until it's able to connect, you can also include a "--timeout" argument when waiting that will kill the process when the timeout value is exceeded.

### Example
```
❯ prefect orion status
Healthy
❯ prefect orion status --wait
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
Healthy
prefect orion status --wait --timeout 5
All connection attempts failed retrying
All connection attempts failed retrying
All connection attempts failed retrying
Timeout value exceeded
  File "/Users/masonmenges/Repos/prefect_dev/prefect/src/prefect/cli/orion.py", line 183, in status
    exit_with_error(f"Timeout value exceeded")
  File "/Users/masonmenges/Repos/prefect_dev/prefect/src/prefect/cli/_utilities.py", line 23, in exit_with_error
    raise typer.Exit(code)
click.exceptions.Exit: 1
```

### Checklist

- [x] closes https://github.com/PrefectHQ/prefect/issues/6990


src/prefect/cli/orion.py
```diff
@@ -150,6 +152,42 @@ async def start(
...
+                else:
+                    if not timeout:
+                        print(f"{response} retrying")
+                    else:
+                        timeout_diff = pendulum.now() - start_time
+                        if timeout_diff.seconds > timeout:
+                            print(f"Timeout exceeded, {response}")
+                            status_check = False
+                        else:
+                            print(f"{response} retrying")
```
madkinsz, line 188: I'd use `anyio.move_on_after` to enforce your timeout instead, you can also just break on a healthy response e.g.

```python

import anyio
import random
import contextlib


async def main():
    timeout = None
    with (
        anyio.move_on_after(timeout) if timeout else contextlib.nullcontext()
    ) as timeout_scope:
        while True:
            if not random.choice(range(10)):
                break
            await anyio.sleep(1)

    if timeout and timeout_scope.cancel_called:
        print("Timed out!")
        return -1

    print("Healthy!")
```

src/prefect/cli/orion.py
```diff
@@ -150,6 +152,42 @@ async def start(
...
+        status_check = True
+        start_time = pendulum.now()
+        while status_check == True:
+            response = await client.api_healthcheck()
+            if not response:
+                print("Orion Healthy")
+                status_check = False
+            else:
+                if wait == False:
+                    print(response)
```
madkinsz, line 177: You'll want to use `app.console.print`

madkinsz: See also the `exit_with_success` and `exit_with_failure` utils

src/prefect/cli/orion.py
masonmenges: I updated the code to use move_on_after and it works pretty well, it is generating a runtime error "RuntimeError: The connection pool was closed while 1 HTTP requests/responses were still in-flight." that I'm not sure if we want to suppress since this would likely be expected when it times out while processing a request

src/prefect/cli/orion.py
masonmenges: Updated to use exit_with_failure, the code looks much cleaner now, Thanks for the suggestions 😄

tests/cli/test_orion.py
```diff
@@ -0,0 +1,51 @@
...
+#     async def mock_api_healthcheck(*args):
+#         return Exception("All connection attempts failed")
+
+#     monkeypatch.setattr(
+#         "prefect.client.OrionClient.api_healthcheck", mock_api_healthcheck
+#     )
+#     invoke_and_assert(
+#         ["orion", "status", "--wait", "--timeout", 10],
+#         expected_output_contains="Timeout value exceeded"
+#     )
```
masonmenges, line 51: @madkinsz I ran into some issues with this test since it doesn't seem to be respecting the anyio.move_on_after timeout, if you have any suggestions on this it would be greatly appreciated 😄

madkinsz: I took a look and it should be working! 10 seconds is a while to run in a test. See https://github.com/PrefectHQ/prefect/pull/7970/commits/e5136122d25464cd1f3a90b08ef15be207899f43

masonmenges: Awesome Thanks 😄

src/prefect/cli/orion.py
```diff
@@ -150,6 +152,41 @@ async def start(
...
+        with (
+            anyio.move_on_after(timeout) if timeout else contextlib.nullcontext()
+        ) as timeout_scope:
+            while True:
+                response = await client.api_healthcheck()
+                if not response:
+                    break
+                else:
+                    if not wait:
+                        exit_with_error(response)
```
madkinsz, line 178: What kind of responses do we typically see?

src/prefect/cli/orion.py
```diff
@@ -150,6 +152,41 @@ async def start(
...
+        help="Runs health check loop until successful or optional timeout value is exceeded.",
+    ),
+    timeout: int = typer.Option(
+        None, "--timeout", help="Optional Timeout value in seconds"
+    ),
+):
+    """Verify Connection status with Orion API"""
+    sleep_time = 1  # Initial sleep time when polling
+    async with get_client() as client:
+        with (
```
madkinsz, line 169: Let's do some sort of display about what server we're connecting to?

madkinsz (commented):
Tests are failing so you'll need to make a few tweaks there still :)

Thanks for taking this one on!

src/prefect/cli/orion.py
masonmenges: Would is make sense to display the api URL we're attempting to connect to here?

src/prefect/cli/orion.py
masonmenges: The Health Check either returns None if successful or an Exception - All connection Attempts failed

src/prefect/cli/orion.py
madkinsz: That could work! You'll need to consider the ephemeral case as well but that should always be available (it might fail a database migration or something on startup so a status check is reasonable).

madkinsz:
@masonmenges just want to make sure this one doesn't get lost, do you plan on working on this more / need help?

masonmenges:
@madkinsz Yep still planning on working through this, Just getting caught up on some other work stuff after coming back from vacation, I'll be spending some time on this early next week to hopefully finish tightening everything up 😄

src/prefect/cli/orion.py
masonmenges: Is there anything specific that would be useful to do with the response I'm getting back here?

src/prefect/cli/orion.py
masonmenges: I've made some changes here to display the URL we're attempting to connect to, as for the ephemeral api is there a good way to check for that and would we just want to exit the status check at that point?

madkinsz:
@masonmenges fyi I force-pushed to your branch to fix the history. All I did was check it out and run `git rebase main`.

masonmenges:
@madkinsz thanks, I was just working on fixing that 😅

src/prefect/cli/orion.py
```diff
@@ -150,6 +153,54 @@ async def start(
     app.console.print("Orion stopped!")
 
 
+@orion_app.command()
+async def status(
+    wait: bool = typer.Option(
+        False,
+        "--wait",
+        help="When the wait command is specified the health check loop runs until successful or an optional timeout value is exceeded, otherwise the check is run only once.",
```
tpdorsey, line 161:
```suggestion
        help="When the wait command is specified, the health check loop runs until successful or an optional timeout value is exceeded. Otherwise the check is run only once.",
```

src/prefect/cli/orion.py
```diff
@@ -150,6 +153,54 @@ async def start(
...
+    ),
+    timeout: int = typer.Option(
+        None,
+        "--timeout",
+        help="Optional Timeout value in seconds, must be used with the wait argument",
```
tpdorsey, line 166:
```suggestion
        help="Optional timeout value in seconds. Must be used with the wait argument",
```

src/prefect/cli/orion.py
```diff
@@ -150,6 +153,54 @@ async def start(
...
+    ),
+    api: str = SettingsOption(PREFECT_API_URL),
+):
+    """Verify Connection status with Orion API"""
+
+    if not api:
+        exit_with_success(
+            "PREFECT_API_URL not set for the currently active profile. To check the status of an Orion instance or Cloud API connection the PREFECT_API_URL must be specified for the selected profile."
```
tpdorsey, line 174:
```suggestion
            "PREFECT_API_URL not set for the currently active profile. To check the status of an Orion instance or Cloud API connection, the PREFECT_API_URL must be specified for the selected profile."
```

tpdorsey (commented):
Couple minor suggested edits.

src/prefect/cli/orion.py
masonmenges: Thanks Terry I updated the messages for each of these.

src/prefect/cli/orion.py
masonmenges: I ended up including a check to exit the cli command if it's running against the ephemeral API with a message indicating that the PREFECT_API_URL needs to be set in order to verify the status of an Orion or Cloud api connection

tpdorsey (approved):

masonmenges:
@zhen0 This is mostly done but I'll need to update a few things since there have been some changes since I finished working on it, i.e. reviving Orion references, etc.

zhen0:
> @zhen0 This is mostly done but I'll need to update a few things since there have been some changes since I finished working on it, i.e. reviving Orion references, etc.

Understood!  And thank you for persevering with this when I know you have other competing priorities.

zhen0:
Could we move back to draft status until it's ready for another review?

masonmenges:
Yep, I'll work on finding some time to make the necessary changes here

PR #7982, opened 2022-12-23
feature: get flow run logs from CLI

### LINKED ISSUE #7814: Get flow's logs using command line
### First check

### Prefect Version

2.x

Based on the below conversation,
https://prefect-community.slack.com/archives/CL09KU1K7/p1670480537109379

It would be good if we get flow logs with command line

 In prefect1 we can use the `prefect get` command to get logs. expecting similar feature in prefect2 as well.

### Example Use

_No response_

### Additional context

_No response_

ohadch:
Usage: prefect flow-run logs <flow-run-id>
closes #7814

This PR attempts to resolve issue #7814 .

It adds a new `logs` command to the `flow-runs` command in the CLI.

### Example

Usage is as follows:

```
prefect flow-run logs <flow-run-id>
```

The CLI responds as follows:

```
2022-12-23T14:44:16.847706+00:00  | INFO  | Finished in state Completed()
```

This is my first PR here. Waiting for your review and opinions :)

Ohad

### Checklist

### DIFF
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -149,6 +149,12 @@ async def logs(id: UUID):
     log_filter = LogFilter(flow_run_id={"any_": [id]})
 
     async with get_client() as client:
+        # Get the flow run
+        try:
+            flow_run = await client.read_flow_run(id)
+        except ObjectNotFound as exc:
+            exit_with_error(f"Flow run {str(id)!r} not found!")
+
         while more_logs:
             # Get the next page of logs
             page_logs = await client.read_logs(
@@ -158,13 +164,8 @@ async def logs(id: UUID):
             # Print the logs
             for log in page_logs:
                 app.console.print(
-                    "  | ".join(
-                        [
-                            str(log.timestamp),
-                            logging.getLevelName(log.level),
-                            log.message,
-                        ]
-                    ),
+                    # Print following the flow run format (declared in logging.yml)
+                    f"{pendulum.instance(log.timestamp).to_datetime_string()}.{log.timestamp.microsecond // 1000:03d} | {logging.getLevelName(log.level):7s} | Flow run {flow_run.name!r} - {log.message}",
                     soft_wrap=True,
                 )
 
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -1,6 +1,7 @@
 """
 Command line interface for working with flow runs
 """
+import logging
 from typing import List
 from uuid import UUID
 
@@ -14,9 +15,9 @@ from rich.table import Table
 from prefect.cli._types import PrefectTyper
 from prefect.cli._utilities import exit_with_error, exit_with_success
 from prefect.cli.root import app
-from prefect.client import get_client
+from prefect.client.orion import get_client
 from prefect.exceptions import ObjectNotFound
-from prefect.orion.schemas.filters import FlowFilter, FlowRunFilter
+from prefect.orion.schemas.filters import FlowFilter, FlowRunFilter, LogFilter
 from prefect.orion.schemas.responses import SetStateStatus
 from prefect.orion.schemas.sorting import FlowRunSort
 from prefect.orion.schemas.states import StateType
@@ -135,3 +136,26 @@ async def cancel(id: UUID):
         )
 
     exit_with_success(f"Flow run '{id}' was succcessfully scheduled for cancellation.")
+
+
+@flow_run_app.command()
+async def logs(id: UUID):
+    """
+    View logs for a flow run.
+    """
+    async with get_client() as client:
+        try:
+            log_filter = LogFilter(flow_run_id={"any_": [id]})
+            logs = await client.read_logs(log_filter=log_filter)
+        except httpx.HTTPStatusError as exc:
+            if exc.response.status_code == status.HTTP_404_NOT_FOUND:
+                exit_with_error(f"Flow run {id!r} not found!")
+            else:
+                raise
+
+    for log in logs:
+        app.console.print(
+            "  | ".join(
+                [str(log.timestamp), logging.getLevelName(log.level), log.message]
+            )
+        )
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -143,15 +143,13 @@ async def logs(id: UUID):
     """
     View logs for a flow run.
     """
-    page_size = 50
+    page_size = 200
     offset = 0
     more_logs = True
     log_filter = LogFilter(flow_run_id={"any_": [id]})
 
     async with get_client() as client:
         while more_logs:
-            print()
-
             # Get the next page of logs
             page_logs = await client.read_logs(
                 log_filter=log_filter, limit=page_size, offset=offset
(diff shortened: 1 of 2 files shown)


serinamarie:
Hi @ohadch, thanks so much for contributing to Prefect! When reading the logs we would want to handle paginated queries as well, something like 

```python
offset = 0
limit = 200
more_logs = True
all_logs = []

while more_logs:

    logs = await client.read_logs(
        log_filter=log_filter, limit=limit, offset=offset
        )
    if len(logs) > 0:
        all_logs.extend(logs)
        offset += limit
    else:
        more_logs = False
```
because otherwise we will be limited to 200 logs.

madkinsz:
We'll also probably want to display the logs after each paginated query rather than pushing them all to a list and displaying it in one go.

ohadch:
Hey @serinamarie and @madkinsz ,

Thanks for welcoming me and for reviewing my code. I implemented pagination as requested, the output now looks as follows:

```
2022-12-24T07:56:07.243937+00:00  | INFO  | Log 48 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
2022-12-24T07:56:07.243958+00:00  | INFO  | Log 49 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
Press enter to see more logs...
2022-12-24T07:56:07.243982+00:00  | INFO  | Log 50 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
2022-12-24T07:56:07.244005+00:00  | INFO  | Log 51 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
```

I would like to hear your opinion about the print format. I got the general idea from the way that the logs are printed during the live flow run, but I did not find where the format is declared. Can you maybe either guide me where to find it, or provide me with the desired spec?

Thanks again!
Ohad

madkinsz:
The [logging module](https://github.com/PrefectHQ/prefect/tree/main/src/prefect/logging) may be helpful. The format is defined in the [logging YAML file](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/logging/logging.yml#L16-L21). You may be able to construct a [flow run logger](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/logging/loggers.py#L58) but you'd want to be careful to disable sending those logs to the server ([example](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/engine.py#L706)). It might be more trouble than it's worth to use the actually logger. Maybe you can use a logging formatter without using a logger itself?

madkinsz:
I like the idea of requiring user input to go to the next page, but perhaps we should put that behind a `-i/--interactive` flag and just show all of the logs by default.

ohadch:
@madkinsz 
1. Regarding the logger - I think we should not use the logger there, as we just intend to show logs and not to create new ones. I only want to use the formatter or construct the string according to it in order to keep the UX consistent.
2. I thought you wanted the user input to exist, whoops :) Actually I think this behavior may be counter intuitive and a bit redundant, so maybe we should avoid adding both the flag and the keypress, and just load (and print) the pages each at a time as you suggested?

madkinsz:
@ohadch that's fine with me

tests/cli/test_flow_run.py
```diff
@@ -275,3 +277,98 @@ def test_wrong_id_exits_with_error(self):
...
+            ],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(self.PAGE_SIZE - 1)
+            ],
+        )
+
+    @pytest.mark.parametrize("state", [Completed, Failed, Crashed, Cancelled])
+    async def test_when_more_than_page_size_logs_exists_then_pagination(
```
serinamarie, line 323: Personally, I think that something like `test_when_num_logs_greater_than_page_size_then_pagination` might be clearer.

tests/cli/test_flow_run.py
```diff
@@ -275,3 +277,98 @@ def test_wrong_id_exits_with_error(self):
...
+        self, orion_client, state
+    ):
+        # Given
+        flow_run = await orion_client.create_flow_run(
+            name="scheduled_flow_run", flow=hello_flow, state=state()
+        )
+
+        # Create more than page size flow run logs
```
serinamarie, line 331: Like above, something like "Create enough flow run logs to result in pagination (page_size > 200)" might be clearer.

tests/cli/test_flow_run.py
```diff
@@ -275,3 +277,98 @@ def test_wrong_id_exits_with_error(self):
...
+            invoke_and_assert,
+            command=[
+                "flow-run",
+                "logs",
+                str(flow_run.id),
+            ],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(self.PAGE_SIZE)
```
serinamarie, line 355: If I'm understanding this test correctly, the goal here is to create 201 logs and then return those 201 logs. Would we want this to be `self.PAGE_SIZE + 1` to match with the number of created logs? Regardless, adding `+ 1` still results in a successful test.

serinamarie (commented):
Hi @ohadch, thanks for adding in pagination handling! I left a few nitpick suggestions.

tests/cli/test_flow_run.py
ohadch: That's a good catch! Indeed `self.PAGE_SIZE + 1` is the desired case.

tests/cli/test_flow_run.py
ohadch: Sure 🙏

tests/cli/test_flow_run.py
ohadch: Like above - I agree.

ohadch:
Hey @serinamarie ,

Thank you very much for your review. I fixed the code according to your suggestions.

Happy new year!

src/prefect/cli/flow_run.py
```diff
@@ -135,3 +136,42 @@ async def cancel(id: UUID):
...
+
+            # Print the logs
+            for log in page_logs:
+                app.console.print(
+                    # Print following the flow run format (declared in logging.yml)
+                    f"{pendulum.instance(log.timestamp).to_datetime_string()}.{log.timestamp.microsecond // 1000:03d} | {logging.getLevelName(log.level):7s} | Flow run {flow_run.name!r} - {log.message}",
+                    soft_wrap=True,
+                )
+
+            # Wait for the user to press enter to get the next page
```
serinamarie, line 172: This PR no longer includes user input, though, correct?

ohadch: Correct. As discussed with @madkinsz , we omitted it, and we only load and print the pages chunk by chunk.

serinamarie: I would think we don't need that comment on line 172 then

serinamarie:
```suggestion
```

ohadch:
That's correct, fixed. Thanks!

serinamarie (approved):
@ohadch This looks good - thanks so much for your contribution 🙂

ohadch:
I had the pleasure, thank you @serinamarie !

serinamarie:
Related to this, I have linked to some additional enhancements adding onto this, should you feel so inclined 😀 

Have a happy new year! 🎉

ohadch:
@serinamarie Sure, I already commented on the `--head` one so either of you assign me to it.

Have a happy new year as well!

serinamarie:
@ohadch Amazing, thank you!

PR #8003, opened 2022-12-29
feat: Adds "head" flag to `flow-run logs`

### LINKED ISSUE #7998: Add --head option to the prefect flow-run logs CLI
Follow-up to https://github.com/PrefectHQ/prefect/pull/7982

### First check

### Prefect Version

2.x

The CLI command: `prefect flow-run logs <flow_run_id>` shows all logs.

It would be great to add a `--head` option to show only the first N logs:

```bash
prefect flow-run logs <flow_run_id> [--head [numlines=50]]
```

### Example Use

_No response_

### Additional context

_No response_

ohadch:
closes #7998

Adds a `--head` flag to the `prefect flow-run logs <flow_run_id>` command.

- If head is provided, the first N logs are returned
- Otherwise, all the logs are returned.

### Example

```shell
prefect flow-run logs d6b76008-406c-4f69-a7b7-d1186e47a94d --head 2
```

Returns
```
2022-12-24 07:56:07.242 | INFO    | Flow run 'platinum-bat' - Log 0 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
2022-12-24 07:56:07.243 | INFO    | Flow run 'platinum-bat' - Log 1 from flow_run d6b76008-406c-4f69-a7b7-d1186e47a94d.
```

### Checklist


ohadch:
Although @serinamarie 's spec was 50 lines by default, I chose to return all of them as I found it a little more intuitive and convenient for the user. Please let me know if you think otherwise 🙏

src/prefect/cli/flow_run.py
```diff
@@ -139,13 +139,25 @@ async def cancel(id: UUID):
...
     View logs for a flow run.
     """
+    # Pagination - API returns max 200 lines at a time
     page_size = 200
     offset = 0
     more_logs = True
+
+    # If head is specified, we need to stop after we've retrieved enough lines
+    remaining_logs = head
+    num_logs_returned = 0
```
serinamarie, line 159: I think `num_logs_returned` could go underneath `more_logs`, since it doesn't have to do with whether the head is specified.

src/prefect/cli/flow_run.py
```diff
@@ -156,9 +168,13 @@ async def logs(id: UUID):
             exit_with_error(f"Flow run {str(id)!r} not found!")
 
         while more_logs:
+            num_logs_to_return_from_page = (
+                page_size if remaining_logs is None else min(page_size, remaining_logs)
```
serinamarie, line 172: I think `page_size if remaining_logs is None` might be ambiguous to me since my understanding is that this phrase means the user hasn't specified a head arg, but it doesn't mean there are no remaining logs. I think renaming `remaining_logs` to something like `user_specified_num_logs` might be clearer so that the result is more intuitive, such as `page_size if user_specified_num_logs is None`

serinamarie (commented):
Hi @ohadch, thanks for jumping on this so quickly! 

When you say you return all logs by default when specifying the `head` option, I would expect that 
`prefect flow-run logs <flow_run_id> --head`
would then return all logs, rather than running into the error: `Option '--head' requires an argument.` Also, I think it's likely that if a user specifies the `head` option, it's because they don't want all the logs by default, but I'll defer to @madkinsz on it.

In addition, I wonder if it makes sense to bother with making the [creation of logs a fixture](https://github.com/PrefectHQ/prefect/blob/f91219ec5543e31ceec54a2052bb37b2d03ef992/tests/orion/models/test_logs.py#L33) at some point.

ohadch:
Hey @serinamarie ,

Sure 😄 Let's make it perfect. 
I misunderstood your initial meaning before, but I think I got it now. Maybe let's try to clarify 😅

For now I support the following two cases:

```shell
prefect flow-run logs <uuid>                         // Returns all logs
prefect flow-run logs <uuid> --head N         // Returns the first N logs
```

You would also like the following to work:

```shell
prefect flow-run logs <uuid> --head           // Should return the first 50 logs
```

But unfortunately I can't seem to make it work, because I do not find how to make `typer.Option()` support a flag with a default value. I only succeed with declaring either a boolean flag, and then not being able to pass the optional value (aka 50), or declaring an option with a default value, but then not being able to display all logs when the user does not pass the `--head` parameter.

What do you think? Am I missing something? It is my first time working with `typer`, so maybe I missed something in the docs?

src/prefect/cli/flow_run.py
ohadch: Sure!

src/prefect/cli/flow_run.py
ohadch: I agree

ohadch:
@serinamarie Also, good idea regarding the fixture. Please have a look at that as well.

serinamarie:
> But unfortunately I can't seem to make it work, because I do not find how to make `typer.Option()` support a flag with a default value. I only succeed with declaring either a boolean flag, and then not being able to pass the optional value (aka 50), or declaring an option with a default value, but then not being able to display all logs when the user does not pass the `--head` parameter.
> 
> What do you think? Am I missing something? It is my first time working with `typer`, so maybe I missed something in the docs?

I agree, I don't think it's possible to have a flag support a default value here. We were thinking that an additional `number` param with a default of 20 would be an ideal way to ensure that if the head flag is set, we will receive 20 logs back, unless a user specifies the number themselves.

Example behavior:
`prefect flow-run logs <flow_run_id> --head` return 20 logs
`prefect flow-run logs <flow_run_id> --head -n 210` returns 210 logs

Then we can also use the `-n` param later on with `--tail` 😀 

Also, the new fixture definitely makes the tests look more succinct! 🙏

serinamarie (changes requested):
> But unfortunately I can't seem to make it work, because I do not find how to make `typer.Option()` support a flag with a default value. I only succeed with declaring either a boolean flag, and then not being able to pass the optional value (aka 50), or declaring an option with a default value, but then not being able to display all logs when the user does not pass the `--head` parameter.
> 
> What do you think? Am I missing something? It is my first time working with `typer`, so maybe I missed something in the docs?

I agree, I don't think it's possible to have a flag support a default value here. We were thinking that an additional `number` param with a default of 20 would be an ideal way to ensure that if the `head` flag is set, we will receive 20 logs back, unless a user specifies the number themselves.

Example behavior:
`prefect flow-run logs <flow_run_id> --head` return 20 logs
`prefect flow-run logs <flow_run_id> --head -n 210` returns 210 logs

Then we can also use the `-n` param later on with `--tail` 😀 

Also, the new fixture definitely makes the tests look more succinct! 🙏

ohadch:
@serinamarie Thanks! This idea makes sense to me as well. I implemented it and used 50 lines for the default head value.

src/prefect/cli/flow_run.py
```diff
@@ -139,13 +142,33 @@ async def cancel(id: UUID):
...
 
 @flow_run_app.command()
-async def logs(id: UUID):
+async def logs(
+    id: UUID,
+    head: bool = typer.Option(
+        False,
+        "--head",
+        "-h",
+        help=f"Show the first {LOGS_HEAD_DEFAULT_NUM_LINES} lines of logs instead of the last 10 lines",
```
serinamarie, line 151: Shouldn't this be 
`f"Show the first {LOGS_HEAD_DEFAULT_NUM_LINES} lines of logs instead of all logs"` ?

src/prefect/cli/flow_run.py
```diff
@@ -139,13 +142,33 @@ async def cancel(id: UUID):
...
+        "-n",
+        help=f"Number of lines to show when using the --head flag",
+        min=1,
+    ),
+):
     """
     View logs for a flow run.
     """
-    page_size = 200
+    # Pagination - API returns max 200 lines at a time
```
serinamarie, line 164: Maybe make a note that this is what `LOGS_PAGE_SIZE` refers to?

serinamarie (changes requested):
It's looking good! 😀 One other thing I noticed while testing this is that 
`prefect flow-run logs -n 5` displays all logs but we would like to display the first 5 logs (default sort), e.g. `prefect flow-run logs -n 5` will display the oldest 5 logs in chronological order.

As another example, in the future with `--reverse`, `prefect flow-run logs --reverse -n 5` would display the 5 most recent logs in reverse (newest first)

src/prefect/cli/flow_run.py
ohadch: Good idea

src/prefect/cli/flow_run.py
ohadch: Indeed!

ohadch:
Hey @serinamarie , it's a good catch, I wanted to discuss it but you already resolved my concern :) Now `prefect flow-run logs <uuid> -n 10` for example indeed returns the first 10 results, using `--head` as a default.

tests/cli/test_flow_run.py
```diff
@@ -372,3 +364,143 @@ async def test_when_flow_run_not_found_then_exit_with_error(self, state):
...
+            ],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(self.PAGE_SIZE + 1)
+            ],
+            expected_line_count=self.PAGE_SIZE + 1,
+        )
+
+    async def test_when_head_without_num_lines_then_return_the_default_number_of_lines(
```
serinamarie, line 418:
```suggestion
    async def test_default_head_returns_default_num_logs(
```

tests/cli/test_flow_run.py
```diff
@@ -372,3 +364,143 @@ async def test_when_flow_run_not_found_then_exit_with_error(self, state):
...
+            ],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(10)
+            ],
+            expected_line_count=10,
+        )
+
+    async def test_when_num_lines_passed_without_head_then_return_n_num_lines(
```
serinamarie, line 464:
```suggestion
    async def test_num_logs_passed_standalone_returns_num_logs(
```

tests/cli/test_flow_run.py
```diff
@@ -300,10 +295,24 @@ async def test_when_num_logs_smaller_than_page_size_then_no_pagination(
...
         ]
         await orion_client.create_logs(logs)
 
+        return flow_run
+
+    return create_flow_run
+
+
+class TestFlowRunLogs:
+    PAGE_SIZE = 200
```
serinamarie, line 308: This could be LOGS_DEFAULT_PAGE_SIZE, MAX_PAGE_SIZE or DEFAULT_PAGE_SIZE but I don't feel strongly about it.

serinamarie (changes requested):
Hi @ohadch, just a few small things to wrap this one up 🙂 Since a log can be multiple lines, I think `--num_logs` or `--count` would be a more descriptive long form of the `-n` command. Leaning toward `num_logs`. Sorry I didn't think on this sooner.

Additionally, I'm not sure we need to test Typer functionality, e.g. testing min=1 in `test_when_num_lines_is_smaller_than_one_then_exit_with_error` or the short forms of commands in `test_h_and_n_shortcuts_for_head_and_num_lines`. Is there some particular reason for it?

Thanks so much!

ohadch:
Regarding num-logs - I also think so. I thought that -l --limit may be also a good idea, what do you think?

Regarding the tests - as it is more of an integration compared to a unit test, I do like to test all of the requirements regardless of the implementation method (in this case typer), but I sure can remove it if it is redundant in your opinion :) @serinamarie

serinamarie:
> Regarding num-logs - I also think so. I thought that -l --limit may be also a good idea, what do you think?
> 
> Regarding the tests - as it is more of an integration compared to a unit test, I do like to test all of the requirements regardless of the implementation method (in this case typer), but I sure can remove it if it is redundant in your opinion :) @serinamarie

I can see what you mean regarding the latter point.

I think the `-n` short form is nice, so possibly `-n/--limit`, similar to the [`head` command](https://linux.die.net/man/1/head).

ohadch:
@serinamarie The tests are of course a product of a personal taste, let me know if you decide differently at the end :) 

I do think that it is better to not mix between -n and --limit and stick with one letter, so let's go with your idea of -n/--num-logs and wrap it up?

serinamarie:
@ohadch Sounds great! 🚀

src/prefect/cli/flow_run.py
```diff
@@ -28,6 +28,9 @@
 )
 app.add_typer(flow_run_app, aliases=["flow-runs"])
 
+LOGS_PAGE_SIZE = 200
```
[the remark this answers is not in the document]
ohadch: Agree

tests/cli/test_flow_run.py
ohadch: I agree, your suggestion is much clearer

ohadch:
@serinamarie Done :)

src/prefect/cli/flow_run.py
```diff
@@ -28,8 +28,8 @@
 )
 app.add_typer(flow_run_app, aliases=["flow-runs"])
 
-LOGS_PAGE_SIZE = 200
-LOGS_HEAD_DEFAULT_NUM_LINES = 20
+LOGS_DEFAULT_PAGE_SIZE = 200
+LOGS_DEFAULT_HEAD_NUM_LINES = 20
```
serinamarie, line 32: This will be the constant that is used regardless of whether it is head, tail, or reverse, right? So I don't think having "head" in the name is needed. Maybe `LOGS_FLAG_DEFAULT_NUM_LOGS` or `LOGS_LIMIT_FLAG_DEFAULT_NUM_LOGS` or `LOGS_PREVIEW_FLAG_DEFAULT_NUM_LOGS`?

ohadch: LOGS_WITH_LIMIT_FLAG_DEFAULT_NUM_LOGS is indeed better

src/prefect/cli/flow_run.py
```diff
@@ -139,13 +142,36 @@ async def cancel(id: UUID):
...
-async def logs(id: UUID):
+async def logs(
+    id: UUID,
+    head: bool = typer.Option(
+        False,
+        "--head",
+        "-h",
+        help=f"Show the first {LOGS_WITH_LIMIT_FLAG_DEFAULT_NUM_LOGS} lines of logs instead of all logs.",
+    ),
+    num_lines: int = typer.Option(
```
serinamarie, line 153: I think this should be `num_logs`

ohadch: Good catch. I fixed it an additional few occurrences of 'lines'

madkinsz:
@ohadch looks like this is failing a test case
```
 FAILED tests/cli/test_flow_run.py::TestFlowRunLogs::test_when_num_logs_is_smaller_than_one_then_exit_with_error - AssertionError: Desired contents "Invalid value for '-..." not found in CLI ouput
```

ohadch:
> 

Hey @madkinsz , thanks for reviewing. Apparently this is one of "works on my machine" situations 😅. Not sure what causes it to fail at your end - can you please share the complete test run log?


image

madkinsz:
@ohadch You should be able to see the full logs in GitHub Actions, can you not?

I believe this is do to colored CLI output. @serinamarie and I will look into a fix.

ohadch:
@madkinsz Sure, missed that. Seems like the coloring is a possible lead. On the other hand, as @serinamarie pointed out earlier, this test might be redundant anyway for this feature. What do you think?

madkinsz:
We disable colors with `PREFECT_CLI_COLORS` in normal output but since this validator is on the parameter typer is doing the output and typer does not respect that variable. We'll need to figure out how to disable typer's colored output during tests.

ohadch:
@madkinsz I see the problem. My question is if that is a blocker in your opinion?

tests/cli/test_flow_run.py
```diff
@@ -372,3 +364,139 @@ async def test_when_flow_run_not_found_then_exit_with_error(self, state):
...
+            ],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(10)
+            ],
+            expected_line_count=10,
+        )
+
+    async def test_when_num_logs_is_smaller_than_one_then_exit_with_error(
```
serinamarie, line 484:
```suggestion
    @pytest.mark.skip(reason="we need to disable colors for this test to pass")
    async def test_when_num_logs_is_smaller_than_one_then_exit_with_error(
```

serinamarie (changes requested):
Hi @ohadch Can you skip that test for now?

ohadch:
Hey @serinamarie , committed your suggestion for skipping the test. :)

madkinsz (approved):

serinamarie (approved):

serinamarie:
Thanks a lot, @ohadch! 🎉

PR #8014, opened 2022-12-30
Add experimental use of runtime to engine

madkinsz:
### Checklist

### DIFF
--- a/src/prefect/_internal/concurrency/runtime.py
+++ b/src/prefect/_internal/concurrency/runtime.py
@@ -162,7 +162,11 @@ class Runtime:
         return self._is_async
 
     def run_in_thread(
-        self, __fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
+        self,
+        __fn: Callable[P, T],
+        *args: P.args,
+        __sync__: bool = None,
+        **kwargs: P.kwargs,
     ) -> T:
         """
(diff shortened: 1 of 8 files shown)


src/prefect/_internal/concurrency/runtime.py
```diff
@@ -96,17 +96,16 @@ def __init__(self, name: str = "RuntimeLoopThread"):
...
         """
-        Start the thread and wait until the event loop is ready.
+        Extends the default `Thread.start()` to wait until `run` is ready and reraise
+        any errors encountered during setup.
         """
         super().start()
-        # Extends the typical thread start event to include a readiness event set from
-        # within the event loop.
-        self._ready_event.wait()
+        self._ready_future.result()
```
madkinsz, line 108: Here we use a future instead of an event to raise errors that occur during startup of the thread.

src/prefect/engine.py
```diff
@@ -117,6 +119,14 @@ def enter_flow_run_engine_from_flow_call(
     This function does the heavy lifting of ensuring we can get into an async context
     for flow run execution with minimal overhead.
     """
+    if experiment_enabled("engine_v2"):
+
+        from prefect._internal.engine.flows import enter_engine_from_flow_call
```
madkinsz, line 124: This module is not yet committed.

PR #8037, opened 2023-01-03
Add `namedtuple` to `BaseAnnotation` bases

madkinsz:
Extends https://github.com/PrefectHQ/prefect/pull/7263 allowing annotations to be serialized/deserialized by libraries that support namedtuples.

With https://github.com/dask/distributed/pull/7292 this will close https://github.com/PrefectHQ/prefect-dask/issues/43

### Checklist


ahuang11:
Unsure why Python 3.7 fails but all other versions are fine
```

ERROR    prefect.flow_runs:engine.py:399 Finished in state Failed('Flow run encountered an exception. RecursionError: maximum recursion depth exceeded in comparison\n')
_________________ TestUnmapped.test_always_returns_same_value __________________
[gw0] linux -- Python 3.7.15 /opt/hostedtoolcache/Python/3.7.15/x64/bin/python

self = <tests.utilities.test_annotations.TestUnmapped object at 0x7f2433bedc10>

    def test_always_returns_same_value(self):
        thing = unmapped("hello")
    
        for _ in range(10):
>           assert thing[random.randint(0, 100)] == "hello"

tests/utilities/test_annotations.py:11: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/prefect/utilities/annotations.py:44: in __getitem__
    return self.unwrap()
src/prefect/utilities/annotations.py:20: in unwrap
    return self.value
src/prefect/utilities/annotations.py:44: in __getitem__
    return self.unwrap()
src/prefect/utilities/annotations.py:20: in unwrap
    return self.value
E   RecursionError: maximum recursion depth exceeded in comparison
```

ahuang11:
```
    def __getitem__(self, _) -> T:
        # Internally, this acts as an infinite array where all items are the same value
        return super().__getitem__(_)
```
seems to have fixed the recursion error and returns the expected result. Will see if this breaks other tests.

ahuang11:
Okay I think I found a workaround by converting to dict first; not sure if this should only apply for python 3.7
```
    def unwrap(self) -> T:
        # cannot simply return self.value due to recursion error in Python 3.7
        # also _asdict does not follow convention; it's not an internal method
        # https://stackoverflow.com/a/26180604
        return self._asdict()["value"]
```

src/prefect/utilities/annotations.py
```diff
@@ -41,7 +41,7 @@ class unmapped(BaseAnnotation[T]):
 
     def __getitem__(self, _) -> T:
         # Internally, this acts as an infinite array where all items are the same value
-        return self.unwrap()
+        return super().__getitem__(_)
```
madkinsz: mm this isn't what we want to do  here `__getitem__` should return the value no matter the index. Is this being called by `self.value`?

ahuang11: Right, I think you might have to refresh to see the new changes.

ahuang11:
Yes I've considered it since I'm not sure if using asdict is costly.

tests/utilities/test_collections.py
```diff
@@ -408,6 +408,13 @@ def test_visit_collection_max_depth(self, inp, depth, expected):
         assert result == expected
 
     def test_visit_collection_context(self):
+        foo = [1, 2, [3, 4], [5, [6, 7]], 8, 9]
+
+        def visit(expr, context):
+            if isinstance(expr, list):
+                context["depth"] += 1
+                return expr
+
```
ahuang11, line 417:
```suggestion
```

ahuang11:
Ok, I think this should be ready.

src/prefect/utilities/annotations.py
```diff
@@ -1,20 +1,32 @@
...
+            # https://stackoverflow.com/a/26180604
+            return self._asdict()["value"]
+        else:
+            return self.value
+
+    def rewrap(self, value: T) -> "BaseAnnotation[T]":
+        return type(self)(value)
 
     def rewrap(self, value: T) -> "BaseAnnotation[T]":
         return type(self)(value)
```
madkinsz, line 32: @ahuang11 looks like this got duplicated

ahuang11: Thanks; fixed it.

Issue #8036, opened 2023-01-03
Block names saved with underscore results in a cryptic exception


ahuang11:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

`credentials.save(name='first_aws_creds', overwrite=True)`

results in

```
  raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '422 Unprocessable Entity' for url 
'[https://api.prefect.cloud/api/accounts/../](https://api.prefect.cloud/api/)
```

Originally from https://prefect-community.slack.com/archives/CM28LL405/p1672756772034579?thread_ts=1672663639.940449&cid=CM28LL405

### Reproduction

```python3
from prefect_aws.credentials import AwsCredentials
credentials = AwsCredentials(
        aws_access_key_id="acccess_key_id",
        aws_secret_access_key="secret_access_key"
    )
credentials.save(name='first_aws_creds', overwrite=True)
```

### Error

```python3
raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '422 Unprocessable Entity' for url 
'[https://api.prefect.cloud/api/accounts/../](https://api.prefect.cloud/api/)
```

### Versions

```Text
Version:             2.7.5+12.g2c04fe1ac.dirty
API version:         0.8.4
Python version:      3.9.15
Git commit:          2c04fe1a
Built:               Mon, Jan 2, 2023 11:02 PM
OS/Arch:             darwin/arm64
Profile:             integrations-cloud
Server type:         cloud
```

madkinsz:
@ahuang11 can you fix the formatting?

Is this something you can triage?

ahuang11:
Can you elaborate on what formatting?

Sure, I can take this on.

PR #8038, opened 2023-01-03
Raise a better message when users saves a block with an invalid name

ahuang11:
Previously, the error message was cryptic, when executing with an invalid block name:
`credentials.save(name='first_aws_creds', overwrite=True)`

```
  raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '422 Unprocessable Entity' for url 
'[https://api.prefect.cloud/api/accounts/../](https://api.prefect.cloud/api/)
```

Now raises a more helpful msg:
```
File ~/Applications/python/prefect/src/prefect/client/orion.py:982, in OrionClient.create_block_document(self, block_document, include_secrets)
    980             message = detail.get("msg", "")
    981             if "name must only contain" in message:
--> 982                 raise prefect.exceptions.InvalidNameError(message) from e
    983     raise
    984 return BlockDocument.parse_obj(response.json())

InvalidNameError: name must only contain lowercase letters, numbers, and dashes
```

Closes https://github.com/PrefectHQ/prefect/issues/8036

### Example

```
from prefect_aws.credentials import AwsCredentials
credentials = AwsCredentials(
        aws_access_key_id="acccess_key_id",
        aws_secret_access_key="secret_access_key"
    )
credentials.save(name='first_aws_creds', overwrite=True)
```

### Checklist


src/prefect/client/orion.py
```diff
@@ -975,8 +975,12 @@ async def create_block_document(
         except httpx.HTTPStatusError as e:
             if e.response.status_code == status.HTTP_409_CONFLICT:
                 raise prefect.exceptions.ObjectAlreadyExists(http_exc=e) from e
-            else:
-                raise
+            elif e.response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY:
+                for detail in e.response.json().get("exception_detail", []):
+                    message = detail.get("msg", "")
+                    if "name must only contain" in message:
+                        raise prefect.exceptions.InvalidNameError(message) from e
```
madkinsz, line 982: Instead of this, we should consider not sending this request to the API at all since we know it's invalid. If you add a validator to `BlockDocumentCreate`, then the error will be thrown client-side during construction of the document.

ahuang11: It seems like there already is a validator in BlockDocumentCreate:

```



def validate_block_document_name(value):
    if value is not None and not bool(
        re.match(LOWERCASE_LETTERS_AND_DASHES_ONLY_REGEX, value)
    ):
        raise ValueError(
            "name must only contain lowercase letters, numbers, and dashes"
        )
    return value
...
@copy_model_fields
class BlockDocumentCreate(ActionBaseModel):
    """Data used by the Orion API to create a block document."""

    name: Optional[str] = FieldFrom(schemas.core.BlockDocument)
    data: dict = FieldFrom(schemas.core.BlockDocument)
    block_schema_id: UUID = FieldFrom(schemas.core.BlockDocument)
    block_type_id: UUID = FieldFrom(schemas.core.BlockDocument)
    is_anonymous: bool = FieldFrom(schemas.core.BlockDocument)

    _validate_name_format = validator("name", allow_reuse=True)(
        validate_block_document_name
    )

    @root_validator
    def validate_name_is_present_if_not_anonymous(cls, values):
        # TODO: We should find an elegant way to reuse this logic from the origin model
        if not values.get("is_anonymous") and not values.get("name"):
            raise ValueError("Names must be provided for block documents.")
        return values
```

ahuang11: I guess `_save`'s `_to_block_document` doesn't return `BlockDocumentCreate`:

```

        try:
            block_document = await client.create_block_document(
                block_document=self._to_block_document(name=name)
            )
```

```

    def _to_block_document(
        self,
        name: Optional[str] = None,
        block_schema_id: Optional[UUID] = None,
        block_type_id: Optional[UUID] = None,
        is_anonymous: Optional[bool] = None,
    ) -> BlockDocument:
```

madkinsz: Sounds like we should pass the correct type to the client or the client should coerce it to the expected type.

ahuang11:
I wonder if we should just migrate the validator onto BlockDocument, but I remember Alex mentioning there was a particular reason why he did not do that.

src/prefect/client/orion.py
```diff
@@ -975,12 +985,6 @@ async def create_block_document(
         except httpx.HTTPStatusError as e:
             if e.response.status_code == status.HTTP_409_CONFLICT:
                 raise prefect.exceptions.ObjectAlreadyExists(http_exc=e) from e
-            elif e.response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY:
-                for detail in e.response.json().get("exception_detail", []):
-                    message = detail.get("msg", "")
-                    if "name must only contain" in message:
-                        raise prefect.exceptions.InvalidNameError(message) from e
-            raise
```
madkinsz: Need to retain this raise

src/prefect/client/orion.py
```diff
@@ -962,6 +962,16 @@ async def create_block_document(
...
+        if isinstance(block_document, BlockDocument):
+            block_document_kwargs = block_document.dict(
+                json_compatible=True,
+                include_secrets=include_secrets,
+                exclude_unset=True,
+                exclude={"id", "block_schema", "block_type"},
+            )
+            block_document = schemas.actions.BlockDocumentCreate(
+                **block_document_kwargs
+            )
```
madkinsz, line 974: Can you use `schemas.actions.BlockDocumentCreate.parse_obj(...)`? (actually not sure)

Should probably update the function signature to indicate this is supported as well.

madkinsz (approved):

madkinsz:
Needs a lint fix.

PR #8039, opened 2023-01-03
Refactor App.vue so everything is contained in a suspense wrapper

pleek91:
# Description
We were already doing this for the main page views. This just moves everything inside the suspense so we can user permissions (which are async) inside the context sidebar

@znicholasbrown I'd like you to review this specifically to make sure I didn't mess up the scrolling fix css you added.


znicholasbrown (approved):
lgtm @pleek91

PR #8040, opened 2023-01-03
Add Python dependency scanning for the 1.x branch

madkinsz:
This was removed when we moved the default branch to `main`.

I have not enabled npm scanning since it is only used for legacy documentation, the security alerts are always false positives, and the requirements should stay frozen for consistent builds.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,11 @@ updates:
     schedule:
       interval: "daily"
     labels: ["maintenance"]
+
+  # Check for Python updates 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -24,3 +24,11 @@ updates:
     schedule:
       interval: "daily"
     labels: ["maintenance"]
+
+  # Check for Python updates
```
madkinsz, line 28:
```suggestion
  # Check for Python updates in v1
```

peytonrunyan (approved):

PR #8074, opened 2023-01-05
SecretDict is not JSONSerializable in Deployment

ahuang11:
User reports that: `prefect deployment build sky-pipe/flows/test_flow.py:test_flow -n testaroo -q sky-pipe --skip-upload -a -ib "cloud-run-job/test"` does not work in prefect_gcp==0.2.2, a version that introduces `SecretDict` inside `GcpCredentials`.

However, during my testing, I found this works, with plain blocks.
```
from prefect_gcp import GcpCredentials
from prefect_gcp.cloud_run import CloudRunJob
job = CloudRunJob(image="myimage", region="region", credentials=GcpCredentials.load("gcp-credentials"))
job.json()
```

The reason it works with blocks is because `Block` class has special handling; it does not work in a `Deployment` because `Deployment` inherits from `BaseModel` and not `Block`,  
```
    class Config:
        extra = "allow"

        json_encoders = {SecretDict: lambda v: v.dict()}
```
So this PR copies over the Config to Deployment to make it work.

Closes https://github.com/PrefectHQ/prefect-gcp/issues/102

### Example

```python
from prefect.blocks.fields import SecretDict
from prefect_gcp.cloud_run import CloudRunJob, GcpCredentials

creds = GcpCredentials.load("gcp-credentials")  # must set service_account_info
crj = CloudRunJob(credentials=creds, image="abc", region="abc")
Deployment(name="abc", infrastructure=crj).to_yaml("my.yaml")
```

### Checklist

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -16,6 +16,7 @@ import yaml
 from pydantic import BaseModel, Field, parse_obj_as, validator
 
 from prefect.blocks.core import Block
+from prefect.blocks.fields import SecretDict
 from prefect.client.orion import OrionClient, get_client
 from prefect.client.utilities import inject_client
 from prefect.context import FlowRunContext, PrefectObjectRegistry
@@ -265,6 +266,7 @@ class Deployment(BaseModel):
     """
 
     class Config:
+        json_encoders = {SecretDict: lambda v: v.dict()}
         validate_assignment = True
         extra = "forbid"
 
--- a/tests/test_deployments.py
+++ b/tests/test_deployments.py
@@ -10,6 +10,7 @@ from pydantic.error_wrappers import ValidationError
 
 from prefect import flow, task
 from prefect.blocks.core import Block
+from prefect.blocks.fields import SecretDict
 from prefect.client.orion import OrionClient
 from prefect.deployments import Deployment, run_deployment
 from prefect.exceptions import BlockMissingCapabilities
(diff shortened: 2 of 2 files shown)


madkinsz:
I'm confused — Pydantic does not use nested encoder settings when serializing objects?

ahuang11:
> I'm confused — Pydantic does not use nested encoder settings when serializing objects?

Yes that seems to be the case :/

madkinsz:
Can you write a little MRE demonstrating that and look into opening a bug upstream? That seems like less than ideal behavior. I'm okay with patching in this way in the meantime.

ahuang11:
> Can you write a little MRE demonstrating that and look into opening a bug upstream? That seems like less than ideal behavior. I'm okay with patching in this way in the meantime.

Seems like it's a known issue that will be fixed in v2 (no plans on adding to v1).
https://github.com/pydantic/pydantic/issues/2277

But here's a MRE:
```
from typing import Dict, Any
from pydantic import BaseModel, SecretField
from pydantic.json import timedelta_isoformat

class SecretDict(SecretField):
    def __init__(self, value: Dict[str, Any]):
        self._secret_value = value
    
    def get_secret_value(self) -> Any:
        return self._secret_value


class Credentials(BaseModel):

    secret_dict: SecretDict

    class Config:
        arbitrary_types_allowed = True
        json_encoders = {
            SecretDict: lambda v: v.get_secret_value(),
        }

class Job(BaseModel):

    credentials: Credentials

credentials = Credentials(secret_dict=SecretDict({"key": "val"}))
job = Job(credentials=credentials)

print(credentials.json())

print(job.json())
```

anna-geller (approved):
great work! QA passes

image


thanks so much for fixing it so quickly

desertaxle (approved):

PR #8092, opened 2023-01-06
Add interactive sandbox

ahuang11:
I think it'd be cool if users get to click run on example code!

image

### Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -144,7 +144,7 @@ This code fetches data about GitHub stars for a few repositories. Add the three 
 from prefect import flow, task
 import httpx
 
-@task(retries=3)
+@task(retries=3, log_prints=True)
 def get_stars(repo):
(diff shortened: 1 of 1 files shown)


madkinsz:
I'm a little against this — a big part of our value add is you can run flows then see them in a UI but that won't work here.

ahuang11:
I think it's cool that users can edit the code and see new output.

I was also exploring replit as well and was able to start Orion and have results show up in UI.
image

tpdorsey:
If it works and we can run Orion or, say, link to a Cloud account, I'd like to see this as part of a more holistic first run experience. My suggestion is to continue exploring what is possible here, and engage Reuter and team for input on making this robust and well aligned with a good first/early experience. A tutorial around this would be great.

ahuang11:
Ok will close this for now.

PR #8106, opened 2023-01-09
maintenance: Add github token to remove labels action

cicdw:
Our GitHub Action for auto-removing labels appears to be failing; [concrete example here](https://github.com/PrefectHQ/prefect/actions/runs/3869458533/jobs/6595565332) and [long list here](https://github.com/PrefectHQ/prefect/actions/workflows/remove-issue-labels.yaml).

This PR adds the relevant environment variable that should resolve the issue

### Checklist

### DIFF
--- a/.github/workflows/remove-issue-labels.yaml
+++ b/.github/workflows/remove-issue-labels.yaml
@@ -10,3 +10,5 @@ jobs:
     steps:
       - name: Remove label
         run: gh issue edit ${{ github.event.issue.number }} --remove-label "status:triage"
+        env:
+          GH_TOKEN: ${{ github.token }}


.github/workflows/remove-issue-labels.yaml
```diff
@@ -1,12 +1,14 @@
...
+  issues:
+    types: [closed]
+
+jobs:
+  remove_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove label
+        run: gh issue edit ${{ github.event.issue.number }} --remove-label "status:triage"
+        env:
```
cicdw, line 13: these last two lines are the only lines I actually changed; github is rendering the whole file though because it contained carriage returns and I removed them all in favor of standard newlines

madkinsz:
Good catch! We do this a little differently in the release workflow

https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/release.yaml#L234-L243

Curious if both work?

cicdw:
@madkinsz interesting! I _think_ they both work, but the big difference is that the release workflow example requires an environment secret to be set and managed, and for the workflow to [run within that environment](https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/release.yaml#L207).  Honestly not sure if we should prefer one or the other for the label-removal action

cicdw:
Updated to use `secrets.GITHUB_TOKEN` for consistency with other workflows

madkinsz (approved):

PR #8125, opened 2023-01-10
Update Postgres tests to use Docker instead of GHA services

madkinsz:
### Checklist


.github/workflows/python-tests.yaml
```diff
@@ -161,45 +161,46 @@ jobs:
...
       matrix:
+        postgres-version:
+          - "13"
+          - "14"
         python-version:
-          - "3.7"
-          - "3.8"
-          - "3.9"
           - "3.10"
-          - "3.11"
```
madkinsz, line 169: Probably shouldn't drop all these

madkinsz: Do we expect different database behavior depending on Python version though?

anticorrelator (approved):

PR #8134, opened 2023-01-10
Add additional links to live Orion and Cloud REST API docs

ahuang11:
Took this snippet from https://docs.prefect.io/api-ref/rest-api/ and additionally copied here because a user asked "Thoughts around making REST API endpoints available so users can configure workspaces and permissions without using the UI? e.g workspace creation, roles, service accounts"

I can understand why he missed it because if a user is already familiar with REST concepts, he may skip reading the overview and jump right into the API reference.

### Checklist

### DIFF
--- a/docs/api-ref/overview.md
+++ b/docs/api-ref/overview.md
@@ -15,4 +15,11 @@ Prefect 2 provides a number of programmatic workflow interfaces. Each API is doc
 
 - The Prefect Python API is used to build, test, and execute workflows against the Orion workflow engine. This is the primary user-facing API.
 - The Prefect Orion Python API is used by the Orion server to work with workflow metadata and enforce orchestration logic. This API is primarily used by Orion developers.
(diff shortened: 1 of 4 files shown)


tpdorsey:
@billpalombi what do you think about this. We've been trying to get this to be less of a page-in-a-page experience. I think this gets us back where we started.

tpdorsey:
But maybe this confirms my suspicion that a lack of navigation is going to be confusing to users.

madkinsz:
I struggled to find the Cloud API reference docs as well and looked here expecting a link. I don't think this indicates a problem with lack of navigation — I think this highlights that people want to see our Cloud API reference (which is a way better experience since it's _not_ a page-in-page) which includes many Cloud specific routes but it's not discoverable from our docs (only from the UI in the settings menu).

tpdorsey:
My initial thought: copy the lines from the REST API overview pointing to the server API docs and put them on the API reference overview page (/api-ref/overview/) in an admonition box. This is the first page in the docs people are going to see when they navigate to API Reference

tpdorsey:
Pushed some additional changes to make sure REST API is visible from overview pages of Orion, Cloud and API docs.

madkinsz (approved):

Issue #8124, opened 2023-01-10
`allow_failure()` doesn't work as expected with mapped tasks in 2.7.7


anna-geller:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Running the flow below should result in the `cleanup_task` being always executed. The example below works in Prefect 2.7.6 (and earlier), but not in 2.7.7, which could be related to changes made to annotations 

cc @madkinsz 

### Reproduction

```python3
from prefect import flow, task, get_run_logger, allow_failure


@task
def extract():
    return [1, 2, 3]


@task
def add_one(x):
    if x == 2:
        raise Exception("Something is not right")
    return x + 1


@task
def add_two(x):
    return x + 2


@task
def cleanup_task():
    get_run_logger().info("Cleaning up e.g. removing temp Ray cluster")


@flow
def map_with_cleanup_task():
    a = extract()
    b = add_one.map(a)
    c = add_two.map(b)
    cleanup_task.submit(wait_for=[allow_failure(c)])


if __name__ == "__main__":
    map_with_cleanup_task()
```

### Versions

```Text
Version:             2.7.7
API version:         0.8.4
Python version:      3.10.8
Git commit:          e8ca30b8
Built:               Fri, Jan 6, 2023 4:25 PM
OS/Arch:             darwin/arm64
Profile:             stg
Server type:         cloud
```

madkinsz:
I'm not sure this isn't as-designed yet, but I'll investigate today.

madkinsz:
@anna-geller I've addressed the first bug with https://github.com/PrefectHQ/prefect/pull/8135

I believe https://github.com/PrefectHQ/prefect/pull/8136 will address the use-case here.

PR #8136, opened 2023-01-10
Update `allow_failure` handling to include PENDING tasks for chained failures

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/8124

### Example

```python
from prefect import flow, task, allow_failure


@task
def fails_on_two(x):
    if x == 2:
        raise ValueError("Fail task")
    return x


@task
def identity(y):
    return y


@flow
def test_flow():
    f = fails_on_two.map([1, 2, 3])  # [Completed, Failed, Completed]
    b = identity.map(f)  # [Completed, NotReady, Completed]
    c = identity.submit(1, wait_for=allow_failure(b))
    return c.wait().type


print(test_flow())
```

Prior to this change, `c` would be `NotReady` since an upstream was `NotReady`.
With this change, `c` will reach a `Completed` state.

### Checklist


anna-geller:
I QAed and can confirm that this fixes the issue (the fix from https://github.com/PrefectHQ/prefect/pull/8135 wasn't enough, the cleanup task was still in NotReady state)

closes #8124

anna-geller (approved):

madkinsz:
Looking into this further, I'm not sure this makes sense. `wait_for=allow_failure(...)` doesn't seem like it should allow the task to run when given PENDING tasks. That sounds like it defeats the purpose of... waiting for them. I think we'll want to investigate another way to solve this.

PR #8156, opened 2023-01-12
Add documentation for cache refreshes

madkinsz:
Adds documentation for https://github.com/PrefectHQ/prefect/pull/7856

### Example

Preview at https://deploy-preview-8156--prefect-orion.netlify.app/concepts/tasks/#refreshing-the-cache

### Checklist

### DIFF
--- a/docs/concepts/tasks.md
+++ b/docs/concepts/tasks.md
@@ -267,9 +267,6 @@ def caching_task():
     return random.random()
 ```
 
-
-
-
 ## Timeouts
 
 Task timeouts are used to prevent unintentional long-running tasks. When the duration of execution for a task exceeds the duration specified in the timeout, a timeout exception will be raised and the task will be marked as failed. In the UI, the task will be visibly designated as `TimedOut`. From the perspective of the flow, the timed-out task will be treated like any other failed task. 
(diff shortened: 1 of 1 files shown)


docs/concepts/tasks.md
```diff
@@ -235,6 +235,38 @@ def cached_task():
...
+@task(cache_key_fn=static_cache_key, refresh_cache=True)
+def caching_task():
+    return random.random()
+```
+
+When this task runs, it will _always_ update the cache key instead of using the cached value. This is particularly useful when you have a flow that is responsible for updating the cache.
+
+If you want to refresh the cache for all tasks, you can use the `PREFECT_TASKS_REFRESH_CACHE` setting. Setting `PREFECT_TASKS_REFRESH_CACHE=true` will change the default behavior of all tasks to refresh. This is particularly useful if you want to rerun a flow without cached results.
+
+If you have tasks that should not refresh when this setting is enabled, you may explicitly set `refresh_cache` to `False`. These tasks will never refresh the cache — if a cache key exists it will be read not updated. Note that if a cache key does _not_ exist yet, these tasks can still write to the cache.
```
tpdorsey, line 262:
```suggestion
If you have tasks that should not refresh when this setting is enabled, you may explicitly set `refresh_cache` to `False`. These tasks will never refresh the cache &mdash; if a cache key exists it will be read, not updated. Note that, if a cache key does _not_ exist yet, these tasks can still write to the cache.
```

tpdorsey (approved):
Minor edit suggestion. Looks good!

PR #8164, opened 2023-01-13
Refactor results into a standalone artifact table

anticorrelator:
This PR implements a backend-only refactor of results, moving them from the `state` table to a new `artifact` table, enabling a new results-first way to interact with and visualize workflows. These changes should have no user-facing impact and are backwards compatible. The data migration is also separated from the schema migration, allowing the data migration to happen separately for large databases without disrupting functionality.

This branch should pass tests with only the schema migration: 
- test PR: https://github.com/PrefectHQ/prefect/pull/8199
- diff: https://github.com/PrefectHQ/prefect/pull/8199/files

This branch should pass tests with only the schema migration and results still being written to the state table:
- test PR: https://github.com/PrefectHQ/prefect/pull/8200
- diff: https://github.com/PrefectHQ/prefect/pull/8200/files

### Checklist


src/prefect/orion/orchestration/rules.py
```diff
@@ -268,18 +250,32 @@ async def _validate_proposed_state(
...
+
+        self.session.add(validated_orm_state)
+        self.run.set_state(validated_orm_state)
 
         await self.session.flush()
-        self.validated_state = (
+        validated_state = (
             validated_orm_state.as_state() if validated_orm_state else None
         )
+        validated_state.data = state_data
```
madkinsz, line 277: Won't this fail when the `validated_state` is `None`?

src/prefect/orion/schemas/states.py
```diff
@@ -52,11 +54,37 @@ class StateDetails(PrefectBaseModel):
...
+
+class StateGetter(GetterDict):
+    def get(self, key: str, default: Any) -> Any:
+        # During orchestration, ORM states can be instantiated prior to inserting
+        # results into the artifact table, and the `data` field will fail to load
+        if key == "data":
+            try:
+                return super().get(key, default)
+            except MissingGreenlet:
+                return None
```
madkinsz, line 77: This seems dangerous — can we expand on why we would swallow this error?

src/prefect/orion/schemas/states.py
```diff
@@ -52,11 +54,37 @@ class StateDetails(PrefectBaseModel):
...
+
+        return super().get(key, default)
+
+
+class State(StateBaseModel, Generic[R]):
```
madkinsz, line 82: We should be able to drop the generic — that's only necessary client-side now.

src/prefect/orion/orchestration/rules.py
anticorrelator: you're right, I think it will but I also thought `validated_state` will never be `None` anymore except for `None -> None` transitions?

madkinsz: It can be set to `None` in the line just above though so we need to handle it

src/prefect/orion/schemas/states.py
anticorrelator: there's a comment above and we plan to add tests, do we need more clarity than that?

```
# During orchestration, ORM states can be instantiated prior to inserting
# results into the artifact table, and the `data` field will fail to load
```

anticorrelator: I've updated the comment to the following:

```
# During orchestration, ORM states can be instantiated prior to inserting
# results into the artifact table and the `data` field will not be eagerly
# loaded. In these cases, sqlalchemy will attept to lazily load the the
# relationship, which will fail when called within a synchronous pydantic
 # method
```

src/prefect/orion/schemas/states.py
anticorrelator: It seems to break everything when I remove the generic so I'm leaving it for now

src/prefect/orion/schemas/states.py
madkinsz: That's better, I think. It does not necessarily cover why it doesn't matter that we can't load the field though.

That this is necessary at all worries me and could hide a bug in the future, is there a way we can avoid making the call that causes the problem in the first place?

anticorrelator: I think it will be brittle; we'd have to avoid using the `from_orm` method on the pydantic model, which means we'll have to stitch the object together by hand. I think this is prone to falling out of sync with our state ORM/schema implementations

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045501_2882cd2df464_migrate_artifact_data.py
```diff
@@ -0,0 +1,125 @@
...
+def upgrade():
+    ### START DATA MIGRATION
+
+    # insert nontrivial task run state results into the artifact table
+    def update_task_run_artifact_data_in_batches(batch_size, offset):
+        return f"""
+            INSERT INTO artifact (task_run_state_id, task_run_id, artifact_data)
+            SELECT id, task_run_id, _data
+            FROM task_run_state
+            WHERE _data != 'null' AND _data IS NOT NULL
```
zangell44, line 27: `_data` is not indexed, correct?

I think we'll have to get clever with our query, a temporary index, or maybe just visit all row by id?

src/prefect/orion/database/migrations/versions/sqlite/2023_01_12_000042_f92143d30c24_implement_artifact_table.py
```diff
@@ -0,0 +1,183 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "f92143d30c24"
+down_revision = "bb38729c471a"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("PRAGMA foreign_keys=OFF")
```
zangell44, line 21: ❤️

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045501_2882cd2df464_migrate_artifact_data.py
anticorrelator: we already have temporary indexes on `artifact.task_run_state_id` and `artifact._flow_run_state_id` to improve the data migration, we can add one here as well and remove it in the cleanup migration

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
```diff
@@ -0,0 +1,180 @@
...
+        sa.Column(
+            "updated",
+            prefect.orion.utilities.database.Timestamp(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.Column("key", sa.String(), nullable=True),
+        sa.Column("artifact_type", sa.String(), nullable=True),
+        sa.Column("artifact_data", sa.JSON(), nullable=True),
+        sa.Column("artifact_metadata", sa.JSON(), nullable=True),
```
cicdw, line 44: Any particular reason you chose to prefix these columns with the table name?

src/prefect/orion/models/artifacts.py
```diff
@@ -0,0 +1,50 @@
...
+
+@inject_db
+async def create_artifact(
+    session: sa.orm.Session, artifact: Artifact, db: OrionDBInterface, key: str = None
+):
+    now = pendulum.now("UTC")
+    artifact_id = artifact.id
+    insert_stmt = (await db.insert(db.Artifact)).values(
+        created=now,
+        updated=now,
```
cicdw, line 19: don't we normally rely on a trigger for managing the `updated` column, or am i off base?

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
anticorrelator: I don't think any serious ones, I think of these names were reserved by pydantic or sqlalchemy or something so I just added a prefix. I think there is a way to get around that pretty easily if we want it

src/prefect/orion/models/artifacts.py
anticorrelator: I think we use different strategies in various places, this is the pattern I believe we use if we have separate upsert logic if we're trying to overwrite an existing row

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
serinamarie: Renamed cols `artifact_type` to `type`, `artifact_data` to `data`, `artifact_metadata` to `metadata`.

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045501_2882cd2df464_migrate_artifact_data.py
serinamarie: A temporary index on `data` has been added into the `implement_artifact_table` schema migration and then removed in the `cleanup_artifact_migration` versions.

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045502_2882cd2df465_cleanup_artifact_migration.py
```diff
@@ -0,0 +1,60 @@
...
+                prefect.orion.utilities.database.UUID(),
+                nullable=True,
+            ),
+        )
+        batch_op.create_index(
+            batch_op.f("ix_artifact__task_run_state_id"),
+            ["task_run_state_id"],
+            unique=False,
+        )
+        # batch_op.create_index(
```
zangell44, line 56: can we remove this?

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045502_2882cd2df465_cleanup_artifact_migration.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def upgrade():
+    # drop state id columns after data migration
+    with op.batch_alter_table("artifact", schema=None) as batch_op:
+        batch_op.drop_index(batch_op.f("ix_artifact__task_run_state_id"))
+        batch_op.drop_column("task_run_state_id")
+        batch_op.drop_index(batch_op.f("ix_artifact__flow_run_state_id"))
+        batch_op.drop_column("flow_run_state_id")
+        # batch_op.drop_index(batch_op.f("ix_artifact__data"))
```
zangell44, line 27: can we remove this?

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
serinamarie: Changed col name to `metadata_`.

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
```diff
@@ -0,0 +1,207 @@
...
+
+    with op.batch_alter_table("flow_run_state", schema=None) as batch_op:
+        batch_op.alter_column("data", new_column_name="_data")
+        batch_op.add_column(sa.Column("has_data", sa.Boolean))
+
+    with op.batch_alter_table("task_run_state", schema=None) as batch_op:
+        batch_op.alter_column("data", new_column_name="_data")
+        batch_op.add_column(sa.Column("has_data", sa.Boolean))
+
+    op.execute(
```
zangell44, line 151: 1. This is a large update, we should do it in batches.
2. Is Postgres smart enough to use this index?

anticorrelator: I think this should be moved to its own migration and batched, probably

anticorrelator: The index is not necessary for backwards compatibility but required before the data migration

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045500_2882cd2df463_implement_artifact_table.py
```diff
@@ -0,0 +1,181 @@
...
+        op.f("fk_task_run_state__result_artifact_id__artifact"),
+        "task_run_state",
+        "artifact",
+        ["result_artifact_id"],
+        ["id"],
+        ondelete="SET NULL",
+        use_alter=True,
+    )
+
+    with op.batch_alter_table("flow_run_state", schema=None) as batch_op:
```
zangell44, line 143: does the code in `main` and this PR work with either the column named `_data` or `data`?

src/prefect/orion/database/migrations/versions/postgresql/2023_01_13_045501_2882cd2df464_create_migration_index.py
```diff
@@ -0,0 +1,81 @@
...
+            while True:
+                # execute until we've updated task_run_state_id and artifact_data
+                # autocommit mode will commit each time `execute` is called
+                sql_stmt = sa.text(query(batch_size))
+                result = conn.execute(sql_stmt)
+
+                if result.rowcount < batch_size:
+                    break
+
+    with op.batch_alter_table("flow_run_state", schema=None) as batch_op:
```
zangell44, line 59: Since we're using this index in the data migration, we should create these indexes before migrating the data

src/prefect/orion/database/orm_models.py
```diff
@@ -212,6 +270,37 @@ def __table_args__(cls):
...
+class ORMArtifact:
+    """
+    SQLAlchemy model of artifacts.
+    """
+
+    key = sa.Column(
+        sa.String,
+        nullable=True,
+        index=True,
+        unique=True,
```
zangell44, line 283: Can we move unique constraint to `__table_args__`?

src/prefect/orion/database/orm_models.py
```diff
@@ -133,7 +133,36 @@ def flow_run_id(cls):
...
+    def _result_artifact(cls):
+        return sa.orm.relationship(
+            "Artifact",
+            lazy="joined",
+            foreign_keys=[cls.result_artifact_id],
+            primaryjoin="Artifact.id==%s.result_artifact_id" % cls.__name__,
+        )
+
+    @hybrid_property
+    def data(self):
```
madkinsz, line 160: Can you add a docstring here about why this property is needed?

src/prefect/orion/database/orm_models.py
```diff
@@ -187,7 +216,36 @@ def task_run_id(cls):
...
```
madkinsz, line 243: Same as previous comment; a docstring seems helpful for understanding why we've done this.

src/prefect/orion/database/orm_models.py
```diff
@@ -187,7 +216,36 @@ def task_run_id(cls):
...
+        if self._data:
+            return self._data
+        if not self.result_artifact_id:
+            return None
+        return getattr(self._result_artifact, "data", None)
```
madkinsz, line 248: When can this attribute be missing given that `data` is a field on `ORMArtifact`?

src/prefect/orion/schemas/states.py
```diff
@@ -52,7 +52,19 @@ class StateDetails(PrefectBaseModel):
...
-class State(IDBaseModel, Generic[R]):
+class StateBaseModel(IDBaseModel):
+    def orm_dict(
+        self, *args, shallow: bool = False, json_compatible: bool = False, **kwargs
+    ) -> dict:
+        schema_dict = self.dict(
+            *args, shallow=shallow, json_compatible=json_compatible, **kwargs
+        )
+        # remove the data field in order to construct a state ORM model
+        schema_dict.pop("data", None)
```
madkinsz, line 63: Hm so where does the data go?

anticorrelator: it disappears, which is why this odd behavior is scoped to just this method; we're only calling `orm_dict` in test fixtures and the like, and I added this method so we aren't automatically dropping `data` whenever we're constructing states in the application

src/prefect/orion/orchestration/rules.py
```diff
@@ -233,33 +233,15 @@ async def validate_proposed_state(
...
-        self.proposed_state = None
-        reason = f"Error validating state: {validation_errors!r}"
-        self.response_status = SetStateStatus.ABORT
-        self.response_details = StateAbortDetails(reason=reason)
+        except Exception as exc:
+            logger.exception("Encountered error during state validation")
+            self.proposed_state = None
+            reason = f"Error validating state: {exc!r}"
+            self.response_status = SetStateStatus.ABORT
+            self.response_details = StateAbortDetails(reason=reason)
```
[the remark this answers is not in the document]
anticorrelator: made a tracking issue here: https://github.com/PrefectHQ/prefect/issues/8275

src/prefect/orion/database/orm_models.py
anticorrelator: It shouldn't be, this is leftover from an earlier implementation, so I'm removing the `getattr` call

src/prefect/orion/database/migrations/versions/postgresql/2023_01_26_045501_2882cd2df465_migrate_artifact_data.py
```diff
@@ -0,0 +1,125 @@
...
+    ### START DATA MIGRATION
+
+    # insert nontrivial task run state results into the artifact table
+    def update_task_run_artifact_data_in_batches(batch_size, offset):
+        return f"""
+            INSERT INTO artifact (task_run_state_id, task_run_id, data)
+            SELECT id, task_run_id, data
+            FROM task_run_state
+            WHERE has_data IS TRUE
+            LIMIT {batch_size} OFFSET {offset};
```
zangell44, line 28: Do we need a stable `ORDER BY` clause here too?

Same question for `update_flow_run_artifact_data_in_batches`

anticorrelator: yes good catch

zangell44 (approved):
One more question on the data migration, looks good to me!

PR #8165, opened 2023-01-13
Expose more options for DockerContainer

idantene:
Expode the `user` keyword argument to `DockerContainer` so that it can be used with the `DockerClient` SDK.

### Example

```python3
infra = DockerContainer(user=1000, image="foo:latest")
```

### Checklist

### DIFF
--- a/src/prefect/infrastructure/docker.py
+++ b/src/prefect/infrastructure/docker.py
@@ -211,9 +211,9 @@ class DockerContainer(Infrastructure):
         default_factory=get_prefect_image_name,
         description="Tag of a Docker image to use. Defaults to the Prefect image.",
     )
-    extra_kwargs: Optional[Dict[str, Any]] = Field(
+    extra_options: Optional[Dict[str, Any]] = Field(
         default=None,
-        description='Any additional keyword arguments to pass to the Docker SDK Client',
+        description='Any additional options to pass to the Docker SDK Client',
     )
     image_pull_policy: Optional[ImagePullPolicy] = Field(
         default=None, description="Specifies if the image should be pulled."
@@ -372,7 +372,7 @@ class DockerContainer(Infrastructure):
         docker_client: "DockerClient",
     ) -> Dict:
         network_mode = self._get_network_mode()
-        return {**self.extra_kwargs or dict(), **dict(
+        options = dict(
             image=self.image,
             network=self.networks[0] if self.networks else None,
             network_mode=network_mode,
@@ -386,7 +386,14 @@ class DockerContainer(Infrastructure):
             mem_limit=self.mem_limit,
             memswap_limit=self.memswap_limit,
             privileged=self.privileged,
-        )}
+        )
+        extra_options = self.extra_options or dict()
+        common_keys = set(extra_options).intersection(set(options))
+        if common_keys:
+            warnings.warn("The following keys are defined in 'extra_options' and "
+                          "explicitly in the container settings:\n\t{list(common_keys)}"
+                          "\nUsing the ones defined in 'extra_options'.")
+        return {**options, **extra_options}
 
     def _create_and_start_container(self) -> "Container":
         if self.image_registry:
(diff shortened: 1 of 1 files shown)


peytonrunyan:
Hey @idantene - thanks for the PR! Can you describe the use case here a little more? What are you trying to do with the additional information?

idantene:
In our specific use case, we have a network mount that allows write access only from a specific user.
Our workaround now is to manipulate the Dockerfile such that it uses this specific user ID.
But the docker SDK offers this functionalist anyway, so I don't see why not expose it to the Prefect user:)

peytonrunyan:
That sounds quite reasonable to me! Can you add tests and then we can get this into code review?

madkinsz:
> Our workaround now is to manipulate the Dockerfile such that it uses this specific user ID.

With `USER ...`? What's the downside to that approach for you?

> But the docker SDK offers this functionalist anyway, so I don't see why not expose it to the Prefect user:)

We integrate with a lot of SDKs/APIs and cannot offer 1:1 mappings for configuration — the maintenance cost is too high. Our philosophy is to expose the most common settings with first-class support, then allow users to provide additional settings. For example, here we could add a `extra_options: Dict[str, Any]` field instead and then we'd solve for _any_ option in the Docker SDK without adding more fields to our UI representation of this object. 

Since you note that there's a way to do this in the Dockerfile (I thought you were changing the user from Python as a workaround), I'm worried that this is a bit too niche for a dedicated setting. That's not to say we shouldn't add it, but I'm curious what you think about an alternative approach.

idantene:
@madkinsz I can very much relate and 100% agree. An `extra_options` dictionary sounds much better.

The `USER ...` raises other incompatibility issues, since some of our images can then be used as base images for another build, and having the different user IDs hardcoded in `Dockerfile` makes things difficult.

idantene:
@madkinsz How's this latest patch? I'm not sure which should take precedence, the `extra_kwargs` or the built-in parameters?

src/prefect/infrastructure/docker.py
```diff
@@ -368,7 +372,7 @@ def _build_container_settings(
         docker_client: "DockerClient",
     ) -> Dict:
         network_mode = self._get_network_mode()
-        return dict(
+        return {**self.extra_kwargs or dict(), **dict(
```
madkinsz, line 375: I think we'll want `extra_kwargs` to take precedence — that will allow users to override our inference behavior if they really want to. I think we should warn if the keys collide though, so I'd do something roughly like:

```
# Construct our dict first
options = dict(...)

extra_kwargs = self.extra_kwargs or {}
duplicate_keys = set(options.keys()).intersection(extra_kwargs.keys())
if duplicate_keys:
   # Throw a warning, use `list_repr` to format keys that overlap, note which will take precedence
    warnings.warn(...)
    
options.update(extra_kwargs)
return options
```

src/prefect/infrastructure/docker.py
```diff
@@ -211,6 +211,10 @@ class DockerContainer(Infrastructure):
         default_factory=get_prefect_image_name,
         description="Tag of a Docker image to use. Defaults to the Prefect image.",
     )
+    extra_kwargs: Optional[Dict[str, Any]] = Field(
```
madkinsz, line 214: I have a preference for `extra_options` as we may use these blocks across multiple languages in the future and kwargs is Python-specific.

madkinsz:
@idantene Sweet thanks! I made a couple comments. Can you update the title too? Note we'll want test coverage before it can be merged.

idantene:
@madkinsz Sorry, forgot about this for a while.
I've addressed your suggestions, though I'm not sure what you meant by `list_repr` 🤔

madkinsz:
@idantene sorry I didn't get back to you :)

We have a `list_repr` utility that is helpful for displaying a list of keys like you do in your warning; not a big deal though.

Can you add tests that cover this? Probably:
- One asserting with a mock that the options are passed to Docker
- One asserting a warning is raised when the keys

PR #8166, opened 2023-01-13
Enable black string formatting

madkinsz:
Exploring if this breaks anything


madkinsz:
We will need https://github.com/psf/black/pull/3463 to be released

PR #8168, opened 2023-01-13
Use uvloop during test runs

madkinsz:


madkinsz:
Not a worthwhile change

PR #8171, opened 2023-01-13
Update Python tests to use a single test matrix for both databases

madkinsz:
Following #8125 we can consolidate the Postgres and SQLite tests into a single matrix for maintainability :)

This also creates an avenue for splitting tests further for speedups (as we have recently done in Cloud)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -361,63 +361,68 @@ async def generate_test_database_connection_url(
         return
 
     print(f"Generating test database connection URL from {original_url!r}")
+    try:
+        scheme, netloc, database, query, fragment = urlsplit(original_url)
+        if scheme == "sqlite+aiosqlite":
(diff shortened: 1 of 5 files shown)


madkinsz:
Failure log at https://github.com/PrefectHQ/prefect/actions/runs/3913490878

tests/conftest.py
```diff
@@ -410,6 +417,10 @@ async def generate_test_database_connection_url(
             pass
         finally:
             await connection.close()
+    else:
+        raise ValueError(
+            f"Unknown scheme {scheme!r} parsed from database url {original_url!r}."
+        )
```
madkinsz, line 423: Without this change, if the schema cannot be parsed pytest raises a `StopIteration` error since the async generator never yields and what you've done wrong is utterly inscrutable.

serinamarie (approved):

PR #8188, opened 2023-01-17
Add support for variadic keyword arguments to `Task.map`

madkinsz:
Prompted by https://prefect-community.slack.com/archives/CL09KU1K7/p1673994869243719

Currently, if you have a task with variadic keyword arguments (i.e. `**kwargs`) it cannot be used with mapping: a mapping length mismatch error will be raised. This occurs because when we cast the call arguments and keyword arguments to _parameters_, we capture the variadic options in a "kwargs" key. 

To resolve this, we need to expand the parameter key (i.e. "kwargs") during length checks. This pull request adds utilities to explode and collapse variadic parameters which are used during mapping to ensure that we are performing length checks on the actual arguments that would be passed to the function.

Instead of checking the lengths of `{'x': [1, 2], 'kwargs': {}}` we check `{'x': [1, 2]}` and  instead of `{'x': [1, 2], 'kwargs': {'y': 3}}` we check `{'x': [1, 2], 'y': 3}`.

### Example

```python

@task
def add(x: int, **kwargs):
    return x + kwargs.get("y", 0)


@flow
def my_flow():
    results = add.map(x=[1, 2])

my_flow()

# Previously:
# MappingLengthMismatch: Received iterable parameters with different lengths.
# Parameters for map must all be the same length. Got lengths: {'x': 2, 'kwargs': 0}

# In this implementation
# `y` can be excluded, a noniterable value, or an iterable of the same length as `x`
```

### Checklist

### DIFF
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -132,7 +132,8 @@ def collapse_variadic_parameters(
         )
 
     new_parameters = parameters.copy()
-    new_parameters[variadic_key] = {}
+    if variadic_key:
+        new_parameters[variadic_key] = {}
 
     for key in missing_parameters:
         new_parameters[variadic_key][key] = new_parameters.pop(key)
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -64,6 +64,82 @@ def get_parameter_defaults(
     return parameter_defaults
 
 
+def explode_variadic_parameter(
+    fn: Callable, parameters: Dict[str, Any]
+) -> Dict[str, Any]:
+    """
+    Given a parameter dictionary, move any parameters stored in a variadic keyword
+    argument parameter (i.e. **kwargs) into the top level.
+
+    Example:
+
+        ```python
+        def foo(a, b, **kwargs):
+            pass
+
+        parameters = {"a": 1, "b": 2, "kwargs": {"c": 3, "d": 4}}
+        explode_variadic_parameter(foo, parameters)
+        # {"a": 1, "b": 2, "c": 3, "d": 4}
+        ```
+    """
+    variadic_key = None
+    for key, parameter in inspect.signature(fn).parameters.items():
+        if parameter.kind == parameter.VAR_KEYWORD:
+            variadic_key = key
+            break
+
+    if not variadic_key:
+        return parameters
+
+    new_parameters = parameters.copy()
+    for key, value in new_parameters.pop(variadic_key).items():
+        new_parameters[key] = value
+
+    return new_parameters
+
+
+def collapse_variadic_parameters(
+    fn: Callable, parameters: Dict[str, Any]
+) -> Dict[str, Any]:
+    """
+    Given a parameter dictionary, move any parameters stored not present in the
+    signature into the variadic keyword argument.
+
+    Example:
+
+        ```python
+        def foo(a, b, **kwargs):
+            pass
+
+        parameters = {"a": 1, "b": 2, "c": 3, "d": 4}
+        collapse_variadic_parameters(foo, parameters)
+        # {"a": 1, "b": 2, "kwargs": {"c": 3, "d": 4}}
+        ```
+    """
+    signature_parameters = inspect.signature(fn).parameters
+    variadic_key = None
+    for key, parameter in signature_parameters.items():
+        if parameter.kind == parameter.VAR_KEYWORD:
+            variadic_key = key
+            break
+
+    missing_parameters = set(parameters.keys()) - set(signature_parameters.keys())
+
+    if not variadic_key and missing_parameters:
+        raise ValueError(
+            f"Signature for {fn} does not include any variadic keyword argument "
+            "but parameters were given that are not present in the signature."
+        )
+
+    new_parameters = parameters.copy()
+    new_parameters[variadic_key] = {}
+
+    for key in missing_parameters:
+        new_parameters[variadic_key][key] = new_parameters.pop(key)
+
+    return new_parameters
+
+
 def parameters_to_args_kwargs(
     fn: Callable,
     parameters: Dict[str, Any],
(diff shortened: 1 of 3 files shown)


tests/test_tasks.py
```diff
@@ -2952,6 +2952,32 @@ def my_flow():
...
+
+    def test_with_variadic_keywords_and_noniterable(self):
+        @task
+        def add_some(x, **kwargs):
+            return x + kwargs["y"]
+
+        @flow
+        def my_flow():
+            numbers = [1, 2, 3]
+            return add_some.map(numbers, y=1)
```
bayersglassey-zesty, line 2976: I think this may need to be:
```python
            return add_some.map(numbers, y=unmapped(1))
```


...where `unmapped` can be imported directly from the `prefect` module.
I'm not familiar enough with the codebase to know how to test that quickly, though.

src/prefect/utilities/callables.py
```diff
@@ -47,6 +47,82 @@ def get_call_parameters(
...
+
+        parameters = {"a": 1, "b": 2, "c": 3, "d": 4}
+        collapse_variadic_parameters(foo, parameters)
+        # {"a": 1, "b": 2, "kwargs": {"c": 3, "d": 4}}
+        ```
+    """
+    signature_parameters = inspect.signature(fn).parameters
+    variadic_key = None
+    for key, parameter in signature_parameters.items():
+        if parameter.kind == parameter.VAR_KEYWORD:
```
bayersglassey-zesty, line 105: I'm not sure whether `*args` works correctly at the moment, but if so, this function could also look for `parameter.kind == parameter.VAR_POSITIONAL`.

tests/test_tasks.py
madkinsz: Nope! We detect `unmapped` automatically now. It can be used for iterable types that you don't want to map over, but isn't needed otherwise.

src/prefect/utilities/callables.py
madkinsz: I was thinking something similar. I presume the same issue will occur with `*args` but I'd rather tackle that after we've agreed on a design for this.

Thanks for the review!

src/prefect/utilities/callables.py
```diff
@@ -47,6 +47,82 @@ def get_call_parameters(
     return dict(bound_signature.arguments)
 
 
+def explode_variadic_parameter(
```
madkinsz, line 50: Simple dedicated tests for these utilities should probably be added before merging.

tests/test_tasks.py
bayersglassey-zesty: Oh nice!

src/prefect/utilities/callables.py
bayersglassey-zesty: You basically have doctests written there already, but I guess you guys don't have doctests turned on in your pytest.ini?..
I highly recommend doctests. :)
https://docs.pytest.org/en/7.1.x/how-to/doctest.html

Basically just replace your triple-backtick blocks with `>>>`. That is, change this:
```python
    Example:

        ```python
        def foo(a, b, **kwargs):
            pass

        parameters = {"a": 1, "b": 2, "kwargs": {"c": 3, "d": 4}}
        explode_variadic_parameter(foo, parameters)
        # {"a": 1, "b": 2, "c": 3, "d": 4}
        ```
```

...into this:
```python
    Example:

        >>> def foo(a, b, **kwargs):
        ...     pass

        >>> parameters = {"a": 1, "b": 2, "kwargs": {"c": 3, "d": 4}}
        >>> explode_variadic_parameter(foo, parameters)
        {'a': 1, 'b': 2, 'c': 3, 'd': 4}
```

madkinsz: I used to write my examples like that but I don't our docs renderer can handle those so we've been switching. It'd be great to enable doctests though if we can get that sorted out :)

bayersglassey-zesty: > our docs renderer can handle those

Awwww. :)

It seems to work fine for me... I cloned the prefect repo, did `pip install -e ".[dev]"` as instructed by the [Contributing page](https://docs.prefect.io/contributing/overview/), and then ran `mkdocs serve`.

I added both styles of Python snippet to this docstring:
```python
class EnforceMinimumAPIVersion:
    """
    FastAPI Dependency used to check compatibility between the version of the api
    and a given request.

    Looks for the header 'X-PREFECT-API-VERSION' in the request and compares it
    to the api's version. Rejects requests that are lower than the minimum version.

    Examples:

        >>> x, y = 1, 2
        >>> z = x + y
        >>> x, y, z
        (1, 2, 3)

        ```python
        x, y = 1, 2
        z = x + y
        x, y, z
        # (1, 2, 3)
        ```

    """
```

...and the resulting page at http://localhost:8000/api-ref/orion/api/dependencies/ looks like this:
![Screenshot from 2023-01-17 19-23-35](https://user-images.githubusercontent.com/101958884/213076078-c0f3b873-46a8-4fd2-83a0-c8b191a19c73.png)

...so maybe there's something I don't know about, but AFAICT, doctests and mkdocs should work just fine together!

madkinsz: hm interesting! I'd be happy to go back to that :) thanks for checking! cc @tpdorsey

tpdorsey: These kinds of examples in the docstrings will always render fine _in the source code view_. There is a small hitch with the new mkdocstrings Python handler and griffe -- it will render certain examples as an "Example" callout box. In that view, `>>>` renders as nested quoted text. But there's a subtlety about which examples get pulled out that I need to look into.

@madkinsz if you see value in pursuing doctests, I'll spend some cycles looking how we can differentiate between the examples used for doctests and examples we want to display in callouts.

bunchesofdonald (approved):
Makes sense to me!

src/prefect/utilities/callables.py
madkinsz: It'd be good to have test coverage of our examples, but I think we probably have some higher priority docs tooling work right now.

Issue #8080, opened 2023-01-05
Flow runs randomly fail with "Crash detected!" with an underlying httpx.HTTPStatusError: Server error '502 Bad Gateway


imsurat:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Randomly the flow runs ( we have seen this for different deployments at different times) fail to start with the logs throwing an exception 

```
Crash detected! Execution was interrupted by an unexpected exception: httpx.HTTPStatusError: Server error '502 Bad Gateway' for url 'http://<orion-server>/api/block_types/slug/local-file-system'
For more information check: https://httpstatuses.com/502
```

Since it is not happening for the same deployment, and the stack trace does not show our code being executed before the exception occurs, we believe it is an issue in the orchestration layer. We have also checked and found that orion server was up and running at this time and are fairly confident that this is not a server issue, but an issue related to the client making the API Call.

See attached [502_Flowrun_fails_to_start.csv](https://github.com/PrefectHQ/prefect/files/10356183/502_Flowrun_fails_to_start.csv) for some more details on deployment id, time it happened and the 502 error message .



### Reproduction

```python3
There is no certain way to reproduce this, it happens randomly.
```


### Error

```python3
"deployment_id","flow_run_id","start_time","end_time","created","message"
"a35c6f3f-7101-4385-a39b-49594e585238","a325ef7c-35e7-4263-9ca3-16af8f085b5d",NULL,NULL,"2023-01-05 21:01:56.793546+00","Crash detected! Execution was interrupted by an unexpected exception: httpx.HTTPStatusError: Server error '502 Bad Gateway' for url 'http://<orion-server>/api/block_types/slug/local-file-system'
For more information check: https://httpstatuses.com/502
"
"1d77831b-97c5-433c-ac2a-f3f4018d00ce","3a745c61-e01b-4cbd-bc55-b21b9a81891c",NULL,NULL,"2023-01-04 21:42:50.371361+00","Crash detected! Execution was interrupted by an unexpected exception: httpx.HTTPStatusError: Server error '502 Bad Gateway' for url 'http://<orion-server>/api/block_types/slug/local-file-system'
For more information check: https://httpstatuses.com/502
"
[...]

PR #8190, opened 2023-01-18
Adding Setting to configure Uvicorn timeout-keep-alive

imsurat:
### Example
Setting PREFECT_ORION_API_KEEPALIVE is useful for avoiding 502 Gateway timeouts when the Orion API is behind an AWS Load Balancer. It should be set to 65 so that AWS Load Balancer timeout closes the socket before uvicorn closes the socket.
closes  https://github.com/PrefectHQ/prefect/issues/8080
### Checklist


madkinsz:
Hi! So this CLI command actually happened not to have any tests yet. Consequently, I pushed https://github.com/PrefectHQ/prefect/pull/8190/commits/3ca312d0783cc12b673dc563d0cbf6592e3bf26f with some scaffolding and an example test checking that changing the setting will update the value. Can you add a test for providing the value via the CLI?

tests/test_settings.py
```diff
@@ -390,6 +391,10 @@ def test_ui_url_set_directly(self):
         with temporary_settings({PREFECT_UI_URL: "test"}):
             assert PREFECT_UI_URL.value() == "test"
 
+    def test_keep_alive_timeout_on_set(self):
+        with temporary_settings({PREFECT_ORION_API_KEEPALIVE: 30}):
+            assert PREFECT_ORION_API_KEEPALIVE.value() == 30
+
```
madkinsz, line 397: I appreciate that you added coverage, but we don't need this test. These tests are only to cover settings with more complex behavior and this one is just a raw value and should be covered by tests of our `Setting` abstraction.

src/prefect/settings.py
```diff
@@ -835,6 +835,12 @@ def default_cloud_ui_url(settings, value):
 )
 """The API's port address (defaults to `4200`)."""
 
+PREFECT_ORION_API_KEEPALIVE = Setting(
+    int,
+    default=5,
+)
+"""The API's keep alive timeout (defaults to `5`)."""
```
madkinsz, line 842: It may be good to explain what this is for and when it is useful. Perhaps link to the Uvicorn documentation.

src/prefect/settings.py
```diff
@@ -835,6 +835,12 @@ def default_cloud_ui_url(settings, value):
 )
 """The API's port address (defaults to `4200`)."""
 
+PREFECT_ORION_API_KEEPALIVE = Setting(
+    int,
+    default=5,
```
madkinsz, line 840: Is this the Uvicorn default value?

imsurat: Yes, based on https://www.uvicorn.org/settings/

imsurat:
> Hi! So this CLI command actually happened not to have any tests yet. Consequently, I pushed [3ca312d](https://github.com/PrefectHQ/prefect/commit/3ca312d0783cc12b673dc563d0cbf6592e3bf26f) with some scaffolding and an example test checking that changing the setting will update the value. Can you add a test for providing the value via the CLI?

Thanks Absolutely. I will try to get this done next week.

src/prefect/settings.py
imsurat: Will do. I was not sure if it was okay to add an external link. I added it and then removed it before submitting.

tests/cli/test_orion.py
```diff
@@ -25,7 +25,7 @@ def test_start_no_options(mock_run_process: AsyncMock):
         ["orion", "start"],
         expected_output_contains=[
             "prefect config set PREFECT_API_URL=http://127.0.0.1:4200/api",
-            "Check out the dashboard at http://127.0.0.1:4200",
```
madkinsz: Ah this won't display if the UI isn't built — nice fix.

tests/cli/test_orion.py
```diff
@@ -56,3 +56,38 @@ def test_start_with_keep_alive_from_setting(mock_run_process: AsyncMock):
...
+    )
+
+
+def test_start_help_from_cli_has_keep_alive(mock_run_process: AsyncMock):
+    invoke_and_assert(
+        ["orion", "start", "--help"],
+        expected_output_contains=[
+            "--keep-alive-timeout",
+            "PREFECT_ORION_API_KEEPA",
+        ],
```
madkinsz, line 92: Is there a clearer assertion we can do here? This doesn't seem like it will prevent the help menu from becoming malformed in the future. We could probably skip this test entirely if it's hard to make an assertion about the help options — that kind of thing is covered by the `typer` tests.

tests/cli/test_orion.py
```diff
@@ -56,3 +56,38 @@ def test_start_with_keep_alive_from_setting(mock_run_process: AsyncMock):
...
+
+
+def test_start_from_cli_with_invalid_keep_alive(mock_run_process: AsyncMock):
+    invoke_and_assert(
+        ["orion", "start", "--timeout-keep-alive", "45"],
+        expected_code=2,
+        expected_output_contains=[
+            "No such option: --timeout-keep-alive Did you mean --keep-alive-timeout?",
+        ],
+    )
```
madkinsz, line 83: We could probably skip this test entirely — this kind of thing is covered by the `typer` tests. Maybe write a test about what happens if an invalid value is passed instead? Like "-10" or "foo"?

tests/cli/test_orion.py
imsurat: The intent is to make sure that this setting is not removed accidentally and tbh I was not thinking about the malformed help menu, with the assumption that is covered elsewhere.

tests/cli/test_orion.py
imsurat: @madkinsz It looks like prefect Setting allows to specify a negative value and passes it along to uvicorn and uvicorn does not complain and starts. What is unclear is that what would be the effective value in this case.

Tested this with uvicorn directly and it lets a negative value through and starts normally. I did take a quick look at uvicorn repo but did not find any checks being made for a non negative value. I have not looked at the code for the transport layer to see if it has any check or what the behavior is with a value of < 1.

Because the Setting type is declared to be int, it does complain when using a non-numeric value. 

I can add an exit_with_error with a check for value > 5 in src/prefect/cli/orion.py around L99 ?

madkinsz: Thanks for investigating! I don't think we need to validate a positive integer if Uvicorn doesn't — let's see if anyone has problem first.

tests/cli/test_orion.py
madkinsz: Ah accidental removal of the setting should be covered by the other tests (like the one I wrote initially).

tests/cli/test_orion.py
imsurat: I have removed this test

tests/cli/test_orion.py
imsurat: Yes of ocurse I can see it now that you mention it. I have removed this test

tests/test_settings.py
imsurat: I have removed this test

src/prefect/settings.py
imsurat: I have added additional text. Hopefully, that is more helpful.

madkinsz (approved):
Thanks!

Issue #7664, opened 2022-11-28
GitHub storage block downloads read-only `.git` files, causing permission errors for `shutil.copytree`


tibuch:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Running a workflow from GitHub storage changes the working directory permissions of the flow from `rwxrwxr-x` to `rwx------`. When the flow is executed a second time the code can not be copied into this directory anymore. 

### Reproduction
Configure a GitHub storage block and a Process infrastructure block with a configured `working_dir`.

Deploy a flow and run it twice. After the first run the permissions of the working directory are changed. The second flow run will fail during flow retrieval from GitHub storage.

### Error

```python3
Flow could not be retrieved from deployment.
Traceback (most recent call last):
  File "/tungstenfs/scratch/gmicro_share/_prefect/miniconda3/envs/test_gfriedri-em-alignment-flows/lib/python3.9/site-packages/prefect/engine.py", line 255, in retrieve_flow_then_begin_flow_run
    flow = await load_flow_from_flow_run(flow_run, client=client)
  File "/tungstenfs/scratch/gmicro_share/_prefect/miniconda3/envs/test_gfriedri-em-alignment-flows/lib/python3.9/site-packages/prefect/client/utilities.py", line 47, in with_injected_client
    return await fn(*args, **kwargs)
  File "/tungstenfs/scratch/gmicro_share/_prefect/miniconda3/envs/test_gfriedri-em-alignment-flows/lib/python3.9/site-packages/prefect/deployments.py", line 163, in load_flow_from_flow_run
    await storage_block.get_directory(from_path=deployment.path, local_path=".")
  File "/tungstenfs/scratch/gmicro_share/_prefect/miniconda3/envs/test_gfriedri-em-alignment-flows/lib/python3.9/site-packages/prefect/filesystems.py", line 917, in get_directory
    copytree(src=content_source, dst=content_destination, dirs_exist_ok=True)
  File "/tungstenfs/scratch/gmicro_share/_prefect/miniconda3/envs/test_gfriedri-em-alignment-flows/lib/python3.9/shutil.py", line 568, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
[...]

PR #8193, opened 2023-01-18
Add option to filter git files with GitHub block

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/7664

Confirmed to work for the user here: https://github.com/PrefectHQ/prefect/issues/7664#issuecomment-1396589690

The GitHub storage block currently clones a repo from GitHub and then copies that repo into a directory. The default behavior is to copy the repo into a random directory, but a directory can also be specified on a process infrastructure block using `working_dir`. 

Using `working_dir` causes an issue when attempting to run a deployment multiple times, because a cloned repo includes `.git` files, which are read only. When shutil attempts to copy files into this directory, it raises a permission error when attempting to overwrite the git objects.

This pull request add a parameter `include_git_objects` to the block, which says whether or not to include objects in `.git`. It  defaults to `True` to preserve current behavior. If it's set to false, then `copytree` will ignore anything matching the glob `'.git'`.

```python
g = GitHub(
    repository="https://github.com/PrefectHQ/prefect.git",
    include_git_objects=False,
)
```

### Checklist

### DIFF
--- a/src/prefect/filesystems.py
+++ b/src/prefect/filesystems.py
@@ -924,13 +924,13 @@ class GitHub(ReadableDeploymentStorage):
                 dst_dir=local_path, src_dir=tmp_dir, sub_directory=from_path
             )
 
-            ignore = None
+            ignore_func = None
             if not self.include_git_objects:
-                ignore = ignore_patterns(".git")
+                ignore_func = ignore_patterns(".git")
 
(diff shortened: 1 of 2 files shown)


cicdw (approved):
seems like a clean solution!

PR #8197, opened 2023-01-18
Stop blocks from referencing themselves

jakekaplan:
closes: https://github.com/PrefectHQ/nebula/issues/3190

### Example
It is currently possible to get a block to create a `BlockDocumentReference` where the `parent_block_document_id` and `reference_block_document_id` are the same. This will lead to an infinite loop when trying to query the database for blocks using `WITH RECURSIVE ...` which looks to extract nested blocks. This pr will guard against this self referential loop from occurring. 

Note: There currently is not an MRE for actually getting into this state using the block interface. Once that exists, more changes and better validation might need to be added. Until then this ensures that the server cannot get into an undesired state.

```bash
# Before
>>> same_id = uuid4()
>>> BlockDocumentReference(parent_block_document_id=same_id, reference_block_document_id=same_id, name="name")

# Now
>>> same_id = uuid4()
>>> BlockDocumentReference(parent_block_document_id=same_id, reference_block_document_id=same_id, name="name")
  File "/Users/jakekaplan/PycharmProjects/prefect/jake-test2.py", line 5, in <module>
    BlockDocumentReference(parent_block_document_id=same_id, reference_block_document_id=same_id, name="name")
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for BlockDocumentReference
__root__
  `parent_block_document_id` and `reference_block_document_id` cannot be the same (type=value_error)

```

### Checklist

### DIFF
--- a/src/prefect/orion/schemas/core.py
+++ b/src/prefect/orion/schemas/core.py
@@ -750,6 +750,16 @@ class BlockDocumentReference(ORMBaseModel):
         default=..., description="The name that the reference is nested under"
     )
 
+    @root_validator
+    def validate_parent_and_ref_are_different(cls, values):
+        parent_id = values.get("parent_block_document_id")
+        ref_id = values.get("reference_block_document_id")
+        if parent_id and ref_id and parent_id == ref_id:
+            raise ValueError(
+                "`parent_block_document_id` and `reference_block_document_id` cannot be the same"
+            )
+        return values
+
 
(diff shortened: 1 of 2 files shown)


peytonrunyan (approved):
This seems reasonable to me.

madkinsz (approved):

PR #8224, opened 2023-01-22
Add `project_urls` to `setup.py`

Laerte:
Today we only use `url` parameter. This PR add relevant project links to `setup.py` as well. 

Today:
image

After this change:
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ setup(
     author_email="help@prefect.io",
     url="https://www.prefect.io",
     project_urls={
+        "Changelog": "https://github.com/PrefectHQ/prefect/blob/main/RELEASE-NOTES.md",
         "Documentation": "https://docs.prefect.io",
         "Source": "https://github.com/PrefectHQ/prefect",
         "Tracker": "https://github.com/PrefectHQ/prefect/issues",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,7 +11,12 @@
     description="Workflow orchestration and management.",
     author="Prefect Technologies, Inc.",
     author_email="help@prefect.io",
-    url="https://github.com/PrefectHQ/prefect/",
+    url="https://www.prefect.io",
+    project_urls={
+        "Documentation": "https://docs.prefect.io",
+        "Source": "https://github.com/PrefectHQ/prefect",
```
madkinsz, line 17: Should we just call this one "GitHub" or is "Source" more common?

madkinsz (approved):
Thank you! These look nice :)

tpdorsey (approved):

setup.py
Laerte: Seems source is more used: https://github.com/pypi/warehouse/issues/5947#issuecomment-699660629

But i can switch to Github if you want. We also have the `Changelog` that render like this on PyPI:

image

Laerte: I just need the  Changelog url, if we want to add too.

madkinsz: "Source" seems fine, thanks for the reference!

There's a changelog at https://github.com/PrefectHQ/prefect/blob/main/RELEASE-NOTES.md

PR #8266, opened 2023-01-26
Add a hint to the Kubernetes block description for imageref

jawnsy:
The OCI spec calls refers to the fully-qualified image path (optional registry, tags, digest, etc) as an "image reference", so we can do the same here.

---

References:

* https://kubernetes.io/docs/concepts/containers/images/#image-names
* https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

In doing research for this, it seems that the name "image reference" isn't very well-defined. I think the example should help illustrate what users should put here, and in particular that they may need to include a tag, if they're not using `latest`

### Checklist


serinamarie (approved):
This seems very reasonable to me

Issue #8281, opened 2023-01-26
Ability to add kubernetes annotations to prefect KubernetesJobs


Wirick:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Right now, there is no way for annotations to be added to a KubernetesJob manifest from my understanding. This goes next to labels as I'm sure you're all aware. 

### Describe the proposed behavior

I would like to be able to specify annotations in my job template manifest and have them populated to my kubernetes job

### Example Use

I need to put an annotation on my kubernetes job to tell an admission controller to populate secrets into the job environment

### Additional context

Annotations are pretty straightforward I think. I implemented a quick fix that I would be happy to iterate on :) it's a little more subtle than labels because annotations aren't in the default JobManifest, so the json application of shortcuts errors out when it is trying to apply an annotation to a doc that doesn't exist. I made a little intermediate function to inject noop defaults into kubernetes job manifests, but I'm not sure if I'm just missing this functionality somewhere else so feel free to point me!

serinamarie:
Hi @Wirick, welcome and thanks so much for reaching out and bringing this issue to our attention! We also appreciate your contribution to work to resolve this issue and will take a look at your pull request 🙂

Wirick:
Thanks so much!

rdenham:
This is something I'd like to see as well. In my manifest, I need to have something like:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: examplepod
  annotations:
    io.kubernetes.cri-o.userns-mode: "auto:size=65536"
  namespace: prefectns
```

but while I can add the `annotations` section to the base job template, it doesn't populate the actual manifest. I'd even be ok modifying the python code to do it in a hacky way, but I got confused about where to put it.

We are running a self hosted server, and I submit a job from a separate client machine.

I thought it would be as simple as editing 'src/prefect/infrastructure/kubernetes.py' but my changes didn't seem to make a difference. Can anyone point to the simplest way of adding a custom annotations section to the metadata of the pod manifest?

BMelekAlan:
Is there any progress for this issue? I  tried using the work pool job template.
I added the following configuration to the prefect-worker Helm chart values.yaml, but it doesn't seem to apply the annotation to the job pods:
```
baseJobTemplate:
  configuration: |
    {
[...]

PR #8284, opened 2023-01-27
Annotations in KubernetesJob

Wirick:
Resolves the issue that I created https://github.com/PrefectHQ/prefect/issues/8281

Adds the ability to add an optional annotations dictionary to the KubenetesJob class. This allows users to specify annotations to be added to their flows. An empty annotations dictionary will be injected if none is specified by the user, but it does not have to be included in the minimal manifest.

Kubernetes annotations are important parts of many application lifecycles, and many things like admission controllers use annotations to apply certain functionality to objects. Right now there is no way to customize annotations within the KubernetesJob objects, and this has become an issue for me and my organization. 

I want to be able to specify a dictionary of annotations in the definition of kubernetes jobs (from my tests):
```
KubernetesJob(command=["echo", "hello"], annotations={"foo": "foo", "bar": "bar"})
```
or
```
manifest = KubernetesJob(
            command=["echo", "hello"],
            job={
                "apiVersion": "batch/v1",
                "kind": "Job",
                "metadata": {"labels": {}, "annotations": {"my-custom-label": "sweet"}},
                "spec": {
                    ...
                },
            },
        ).build_job()
```

I have written tests using the labels tests as a guide. Annotations don't have to conform to the same standards as labels so I didn't put anything through the slugify function. In this vein I would like guidance on two things if this might be something I can upstream:
1. Are there any formatting checks that you want me to implement for either annotation keys or values. I'm happy just kinda trusting users to use them correctly but also totally happy to do some checking.
2. I have made a _add_noop_defaults function. I haven't written any testing for it because I want to ask for more guidance on if it's a viable way of resolving this issue, or if you want to add annotations to the base manifest, or want something else to inject the default value. Right now there seems to be the ability to make this optional in the KubernetesJob class, but if the base manifest doesn't have the default value the json shortcut application errors out.

### Checklist

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -70,6 +70,7 @@ class KubernetesJob(Infrastructure):
             timing out. Defaults to `None`, which means no timeout will be enforced
             while waiting for each event in the job lifecycle.
         labels: An optional dictionary of labels to add to the job.
+        annotations: An optional dictionary of annotations to add to the job.
         name: An optional name for the job.
         namespace: An optional string signifying the Kubernetes namespace to use.
         pod_watch_timeout_seconds: Number of seconds to watch for pod creation before timing out (default 60).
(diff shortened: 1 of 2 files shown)


madkinsz:
Hi @Wirick — this looks pretty good! 

re. your questions
1. I'm happy to avoid any formatting checks for now. 
2. It seems better to just add an empty annotations blurb to the base job definition rather than injecting it with a method.

Wirick:
Ok I will add it to the base job definition!

PR #8287, opened 2023-01-27
add display_name field to graphs endpoint

brandonreid:
Resolves PrefectHQ/graphs/issues/55

This PR adds a `display_name` field to the graphs endpoint, so that on the Flow Run Timeline, we can show the task name in a more human readable way, without the unique id's that make the name super long:

image
See `display_name` compared to `name`.

This was achieved by grabbing the end of the `task_key`, which consistently uses the original task name (either the default, or user specified one) at the end, according to a scanning of the latest 1000 records in staging.

### Checklist

### DIFF
--- a/src/prefect/orion/models/flow_runs.py
+++ b/src/prefect/orion/models/flow_runs.py
@@ -305,6 +305,7 @@ async def read_flow_runs(
 class DependencyResult(PrefectBaseModel):
     id: UUID
     name: str
+    display_name: str
     upstream_dependencies: List[TaskRunResult]
     state: State
     expected_start_time: Optional[datetime.datetime]
@@ -351,6 +352,7 @@ async def read_task_run_dependencies(
                 "state": task_run.state,
                 "expected_start_time": task_run.expected_start_time,
                 "name": task_run.name,
+                "display_name": task_run.task_key.rpartition(".")[-1],
                 "start_time": task_run.start_time,
                 "end_time": task_run.end_time,
                 "total_run_time": task_run.total_run_time,


zangell44 (approved):

src/prefect/orion/models/flow_runs.py
```diff
@@ -351,6 +352,7 @@ async def read_task_run_dependencies(
                 "state": task_run.state,
                 "expected_start_time": task_run.expected_start_time,
                 "name": task_run.name,
+                "display_name": task_run.task_key.rpartition(".")[-1],
```
madkinsz, line 355: Why do we partition on a period?

madkinsz (commented):
This might be weird once we start letting users define their own names.

Can you add test coverage for this?

src/prefect/orion/models/flow_runs.py
madkinsz: Also, if anything, I think the display name should be a portion of the task run name not the task key?

brandonreid: The `task_key` is delimited by a `.`, in instances where there is no period, like when a user has defined a key manually, `rpartition` will still result in an array where the last value is the whole word.

madkinsz: I see. The `task_key` defaults to the qualified import name of the task, a right-partition on a period does make sense here. However, the task key doesn't really make sense as the display name for a task run since task runs won't be distinguishable? A right partition of the `task_key` is the _same_ as the default value for the task name.

brandonreid: That's correct. If there are more than one instance of a task run in a flow, it will show up twice with the same name, but the intent is for the UI to surface the unique name on click. The unique string at the end is not very human readable on it's own and clutters charts, making it more harmful than helpful when simply fallowing the series of events. 

You can see that demonstrated here:
https://user-images.githubusercontent.com/6776415/215178374-c839420b-5bbd-461b-9501-2a602846e9e9.mov

The individual task run name will be available when exploring the task details.

brandonreid:
Closing this as we explore ways of making the task_run name itself be somewhat simplified and more future proofed for when the user is able to create their own name values.

PR #8292, opened 2023-01-27
Update default task run name to exclude hash of task key

madkinsz:
Supersedes #8287 

The task run name isn't very user friendly by default and since we're showing task run names in lots of places we'd like the default to be focused on human readability. The task key will still be available in details. This will allow us to improve readability of task runs in the UI without short-term patches that would complicate design once we allow users to set their own task run names.

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -101,7 +101,6 @@ from prefect.utilities.asyncutils import (
 )
 from prefect.utilities.callables import parameters_to_args_kwargs
 from prefect.utilities.collections import isiterable, visit_collection
-from prefect.utilities.hashing import stable_hash
 from prefect.utilities.pydantic import PartialModel
 
 R = TypeVar("R")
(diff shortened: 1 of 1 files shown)


madkinsz:
Apparently, this is currently untested! I'm not sure I feel like we need to make assertions about the default task run names but it could be kind of nice to add coverage now?

serinamarie (approved):

zangell44 (approved):

madkinsz:
@ahuang11 this won't break the Dask task key right?

ahuang11:
> @ahuang11 this won't break the Dask task key right?

It's separate but I wonder how this shows up in the graph.
https://github.com/PrefectHQ/prefect-dask/blob/main/prefect_dask/task_runners.py#L227-L234

madkinsz:
@ahuang11 it should display the same since Dask just eats up to the first hyphen.

Issue #7547, opened 2022-11-16
Update to task input tab on the task run page


zhen0:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

Currently the task run page task input tab shows the correct keys but all values are empty.  There's a ticket to track api work for that here: https://github.com/PrefectHQ/prefect/issues/6627 Until that work is done, we should remove the values shown in the tab and only show the keys so that we're not giving the misleading suggestion that all task inputs are empty arrays. 

<img width="1293" alt="image" src="https://user-images.githubusercontent.com/40272060/202204368-c56d9506-1246-4e78-84f9-ea9755858e3c.png">

### Reproduction

Create a flow with tasks and run it - go to the task run page and look at the task inputs tab

### Error

<img width="1293" alt="image" src="https://user-images.githubusercontent.com/40272060/202204629-ac8d0e9c-5fdf-4e49-89be-19e7dd238eb6.png">

The [] section is misleading

### Browers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

madkinsz:
Note clarification at https://github.com/PrefectHQ/prefect/issues/6627#issuecomment-1372886667 — I do not believe this issue is following correct assumptions.

jmesterh:
Just want to note as a new user -- this page _is_ misleading, I thought this was a bug or I had misconfigured Orion and my inputs were being lost. I'm testing an ETL and was hoping I could see what table the task was working on based on the inputs shown on this screen.

aeisenbarth:
I came here because this issue describes what I am observing now, although it is marked as resolved and the solutions are merged.

I had to follow all merge requests links _and_ read between the lines to come to the conclusion that for now, only the keys are meaningful and the empty arrays are placeholders (?). This is absolutely not clear from the info modal. It mentions that it can include constants, parameters or other tasks, but no word about whether it means only the keys or whether we should also expect values. In that regard, https://github.com/PrefectHQ/prefect/pull/8295 does some clarification for other issues, but I don't feel it resolves this issue here.

amgcc:
[...]

Issue #6627, opened 2022-08-30
UI task run input page is misleading


madkinsz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the proposed behavior

The task run inputs page should include a description of what contents are expected to appear. For example, clarifying that only relationships between tasks are tracked here.

In the future, Prefect should do some sort of minimal tracking of the actual data passed between tasks so it can be displayed in the UI. However, that feature should be a separate ticket.

Until then, we should explain what a user should expect to see on this page.

### Describe the current behavior

The task run page added in https://github.com/PrefectHQ/nebula-ui/pull/1538 / https://github.com/PrefectHQ/prefect/pull/6570 includes a task inputs tab. This tab does not include the input values to my task.

For a simple flow:
```python
import prefect
from prefect import flow, task, get_run_logger

@task
def log_task(name):
    logger = get_run_logger()
    logger.info("Hello %s!", name)
    logger.info("Prefect Version = %s 🚀", prefect.__version__)

@flow()
def log_flow(name: str):
    log_task(name)

log_flow("Marvin")
```

My task receives the value "Marvin" for the`name` parameter but this is not captured by the API.

<img width="932" alt="Screen Shot 2022-08-30 at 5 01 31 PM" src="https://user-images.githubusercontent.com/2586601/187551995-48c81e8c-1a47-4535-8c69-122ccacb32ed.png">

zhen0:
I worry this might get overlooked as it has a ui label but the issue is actually coming from the api.  So updating the labels.  Let me know if you think it needs any ui action @madkinsz

deklund-xilis:
Can someone clarify what the expected behavior is here? Is it expected in general that the Task Inputs should show the actual values that were passed into a task? Similarly, if a task has an output, is its value expected to show up in the interface?

kelvindecosta:
The expected behavior would be for the API to account for the `dynamicKey` for the task.
AFAIK, this value identifies each mapped task, and as such could index the input values passed to the task

madkinsz:
@zhen0 I feel like tracking the other types of inputs is a separate ticket. This is explicitly about the lack of context for what this is expected to contain. Notably https://github.com/PrefectHQ/prefect/issues/7547 incorrectly says

> Currently the task run page task input tab shows the correct keys but all values are empty. 

[...]

PR #8295, opened 2023-01-28
BugFix: Add information about task inputs

zhen0:
image
Screenshot 2023-02-01 at 12 00 33 PM

The current task run page has a misleading tab called "task inputs".  This adds a modal to explain its purpose. 
Closes https://github.com/PrefectHQ/prefect/issues/7547 and #6627 

@madkinsz - for comment here.  I'd welcome your input on the explanation.  I'd prefer not to remove the tab completely in case it was useful to some users. 

### Example

 
 The UI task run page has a Task Input tab that causes confusion 

### Checklist


madkinsz:
@zhen0 we'll also track relationships that aren't from futures, i.e. if you have a task that returns some object then a later task that uses it (and it's a uniquely identifiable object) we'll show the relationship here.

zhen0:
Partner to https://github.com/PrefectHQ/nebula-ui/pull/2308

orion-ui/src/pages/TaskRun.vue
```diff
@@ -13,6 +13,17 @@
         <TaskRunLogs :task-run="taskRun" />
       </template>
 
+      <template v-if="taskRun" #task-inputs-heading>
+        Task inputs
+        <ExtraInfoModal title="Task Inputs">
```
pleek91, line 18: I think we need a little spacing here. Icon is a little close

orion-ui/src/pages/TaskRun.vue
```diff
@@ -13,6 +13,17 @@
...
+          <template #default>
+            <p>
+              Task inputs are passed to task runs and can
+              include constants, parameter keys, and other task runs.
```
madkinsz, line 22: Hm except right now we are not including constants. It's always a mapping of _parameter keys_ to something the API has been told is to track as an input. Right now, the API is only tracking task run relationships that the client reports when it sees _results from other task runs_.

orion-ui/src/pages/TaskRun.vue
zhen0: Sorry!  The screenshot I provided was an old one - we added padding as part of the component in orion design. It now looks like this:

image

pleek91: Nice!

orion-ui/src/pages/TaskRun.vue
zhen0: OK.  I may need your input on this one.  Any good suggestions for the exact wording here?

zhen0: Updated and using localization for this to keep updates across oss and cloud simpler. Linked to https://github.com/PrefectHQ/orion-design/pull/1136

zhen0:
Converting to draft until https://github.com/PrefectHQ/orion-design/pull/1136 is released

PR #8334, opened 2023-01-31
Update issue label workflow

serinamarie:
Remove issue label `status:in-progress` and `status:accepted` when issue closed as completed.

### Checklist

### DIFF
--- a/.github/workflows/issue-bot.yaml
+++ b/.github/workflows/issue-bot.yaml
@@ -16,13 +16,13 @@ jobs:
 
       - name: Update issue status when assigned
         if: github.event.action == 'assigned'
-        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:accepted" --add-label "status:in-progress"
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --add-label "status:in-progress"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Update issue status when unassigned
         if: github.event.action == 'unassigned'
-        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress" --add-label "status:accepted"
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -34,7 +34,7 @@ jobs:
 
       - name: Update issue status if the issue is closed as completed
         if: github.event.action == 'closed' && contains(github.event.issue.body, 'closed as completed')
-        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress" --add-label "status:new-release"
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress"  --remove-label "status:in-progress"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-bot.yaml
```diff
@@ -32,3 +32,9 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Update issue status if the issue is closed as completed
+        if: github.event.action == 'closed' && contains(github.event.issue.body, 'closed as completed')
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress" --add-label "status:new-release"
```
madkinsz, line 37: Honestly I'm not sure if we should be using this label, I'm not sure it makes sense.

.github/workflows/issue-bot.yaml
```diff
@@ -16,13 +16,13 @@ jobs:
...
         if: github.event.action == 'assigned'
-        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --add-label "status:in-progress"
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:accepted" --add-label "status:in-progress"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Update issue status when unassigned
         if: github.event.action == 'unassigned'
-        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress"
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress" --add-label "status:accepted"
```
madkinsz, line 25: I was hesitant to do this because of roadmap issues. Maybe the amount of those that we'd unassign are rare enough that it doesn't matter. What if we unassign someone because we've just changed the status to deferred or denied? hm.

.github/workflows/issue-bot.yaml
serinamarie: Should `status:in-progress` just be removed then when an issue is closed on completion, leaving it with no `status` label?

.github/workflows/issue-bot.yaml
serinamarie: Ah I see, I just wonder if it gets lost in the shuffle without a `status` label after being unassigned.

.github/workflows/issue-bot.yaml
madkinsz: Yeah that'd make sense, `status:accepted` could be removed too?

.github/workflows/issue-bot.yaml
madkinsz: I was thinking it'd probably be okay if it had both `status:accepted` and `status:in-progress` at the same time.

serinamarie: Got it, so `status:accepted` is never removed until completed, basically.

.github/workflows/issue-bot.yaml
```diff
@@ -32,3 +32,9 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Update issue status if the issue is closed as completed
+        if: github.event.action == 'closed' && contains(github.event.issue.body, 'closed as completed')
+        run: gh issue edit --repo prefecthq/prefect ${{ github.event.issue.number }} --remove-label "status:in-progress"  --remove-label "status:accepted"
```
madkinsz, line 37: Sorry about the back and forth here, but maybe we should just remove these labels in https://github.com/PrefectHQ/prefect/pull/8334/files#diff-4f76a181788cd19641d356309d984a1f4d19a2cfaa81eae2ac271ff63d92a19dR13?

serinamarie: Oh yeah that makes sense. There's nothing special about being 'closed as completed' anymore 😞

serinamarie: I don't know that "Remove status labels on close is the most descriptive" name or if we want to call them out individually

madkinsz: Good enough for me!

madkinsz (approved):

Issue #8345, opened 2023-02-01
KubernetesJob fails due to timeout even when job_watch_timeout_secs is set to None


david-elliott-deliveroo:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

`KubernetesJob` flows terminate early and end up as `Crashed` when executing long-running tasks (e.g. long SQL scripts) because the `kubernetes.watch.Watch().stream()` has retries disabled and thus the stream exits after a period of job inactivity. This is happening even when `job_watch_timeout_secs` is set to None.

The reason this happens is that the kwarg `timeout_seconds` is **always** passed into `watch.stream()` [code ref here](https://github.com/PrefectHQ/prefect/blob/2.7.10/src/prefect/infrastructure/kubernetes.py#LL630) and the kubernetes package uses the **presence** of this kwarg (not its value) to determine whether or not to disable retries [code ref here](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#LL160).

As such, even when `job_watch_timeout_secs` is set to `None`, `disable_retries` is always set to False purely because of the presence of the kwarg (even though the value of it is None). 

The result is that flows with long-running tasks on k8s end up exiting early, because the event stream goes quiet, retries are disabled and thus it exits. Note - it hits [this else clause](https://github.com/PrefectHQ/prefect/blob/2.7.10/src/prefect/infrastructure/kubernetes.py#L639-L644) when this happens, resulting in the following Agent log output:
```shell
prefect.infrastructure.kubernetes-job - Job 'sapphire-beluga-n6t7d': Job did not complete within timeout of Nones.
```

Suggested simple fix might be passing the `timeout_seconds` arg via `**`, ie only passing the timeout if it's actually set, something like...
 
[EXISTING CODE](https://github.com/PrefectHQ/prefect/blob/2.7.10/src/prefect/infrastructure/kubernetes.py#L616-L632)
```python
            watch = kubernetes.watch.Watch()
            with self.get_batch_client() as batch_client:
                remaining_timeout = (
                    (  # subtract previous watch time
                        self.job_watch_timeout_seconds - elapsed
                    )
                    if self.job_watch_timeout_seconds
                    else None
                )

                for event in watch.stream(
[...]

PR #8350, opened 2023-02-01
Fix Kubernetes job watch exit when no timeout given

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/8345

The issue is a good summary of the details here, but briefly:

1. We should not pass `timeout_seconds` to K8s unless it is set, they use the presence of the kwarg to determine if retries should be enabled.
2. We should not exit early if the watch exits unless a timeout is given

### Checklist

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -1,11 +1,11 @@
 import copy
 import enum
 import os
+import time
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Tuple, Union
 
 import anyio.abc
-import pendulum
 import yaml
 from pydantic import Field, root_validator, validator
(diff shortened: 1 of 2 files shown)


zzstoatzz (approved):
makes sense to me - my only other question is whether the k8s watch (with retries enabled) would nullify our `while` loop

david-elliott-deliveroo:
Following up from my above comment - I've now tested the change in this PR myself (branched off 2.7.10 and made the above changes manually in my local copy of prefect) and I can confirm they fix the original issue I raised 🎉  Ran a 1hr sql wait query with no problem 👍 . 

So yeah, this PR fixes the reported issue, however there is a bug (see above comment) in the current `main` branch (included in this branch) which borks the k8s agent in some other way that will need resolving before releasing 2.7.11 CC @madkinsz fyi

zzstoatzz (approved):
👍🏼

PR #8357, opened 2023-02-01
Introduces the Prefect Cloud Events schema and clients.

chrisguidry:
This includes the client-side schema for Prefect Cloud events, as well as the
initial set of client implementations.  The primary client streams events over
WebSockets to Prefect Cloud, including reliable delivery through unreliable
network connections.  Prefect Cloud limits connections to ~30 seconds, so we
need to have highly reliable handling of disconnects.  This implementation has
been tested emperically at high rates (several hundreds of events per second)
and tested with Prefect Cloud for accuracy in their delivery.

Also included are the `NullEventsClient` which will be the default for non-Cloud
users and an `AssertingEventsClient` that will assist in future unit testing of
event emission from various Prefect components.

### Example

```
async with PrefectCloudEventsClient(API_URL, API_KEY) as client:
    await client.emit(
       Event(
           event="marvelous.things.happened",
           resource={
               "prefect.resource.id": "amazement"
           }
       )
    )
```

### Checklist

### DIFF
--- a/src/prefect/events/clients.py
+++ b/src/prefect/events/clients.py
@@ -12,16 +12,18 @@ from prefect.events import Event
 class EventsClient(abc.ABC):
     """The abstract interface for all Prefect Events clients"""
 
-    @abc.abstractmethod
-    async def emit(self, event: Event) -> None:  # pragma: no cover
+    async def emit(self, event: Event) -> None:
         """Emit a single event"""
-
-    def _require_context(self) -> None:
         if not hasattr(self, "_in_context"):
             raise TypeError(
                 "Events may only be emitted while this client is being used as a "
                 "context manager"
             )
+        return await self._emit(event)
+
+    @abc.abstractmethod
+    async def _emit(self, event: Event) -> None:  # pragma: no cover
+        ...
 
     async def __aenter__(self) -> "EventsClient":
         self._in_context = True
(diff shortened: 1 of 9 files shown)


src/prefect/events/clients.py
```diff
@@ -0,0 +1,191 @@
...
+    async def _reconnect(self) -> None:
+        if self._websocket:
+            self._websocket = None
+            await self._connect.__aexit__(None, None, None)
+
+        self._websocket = await self._connect.__aenter__()
+
+        # make sure we have actually connected
+        pong = await self._websocket.ping()
+        await pong
```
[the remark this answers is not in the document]
chrisguidry: I'm not sure why CodeQL thinks these have no effect, they are awaiting a future returned by the (async) function `ping()`.  The effect is to wait until the `pong` returns.

src/prefect/events/clients.py
```diff
@@ -0,0 +1,191 @@
...
+        del self._in_context
+        return None
+
+
+class NullEventsClient(EventsClient):
+    """A Prefect Events client implementation that does nothing"""
+
+    async def emit(self, event: Event) -> None:
+        """Emit a single event"""
+        self._require_context()
```
zangell44, line 45: perhaps it's worth standardizing this handling somewhere? either a decorator or allowing a separate `_emit` method for implementation?

chrisguidry: Yeah this thing got a little ugly, I like what you're saying about having `emit` be concrete with an abstract `_emit`.  That would be much cleaner

zangell44 (approved):
All the events!

PR #8380, opened 2023-02-02
BugFix: Use recentFlowRunFilter instead of flowRunFilter

zhen0:
Uses the recentFlowRunFilter instead of runFilter to match the date filter that's applied on work pools and the empty state message. 

### Checklist


zhen0:
@marichka-offen - need your 👀 here! I'm not certain if the difference was deliberate but it feels confusing to see more runs on the queue page than its parent pool page.

marichka-offen (approved):
Good catch!!

PR #8383, opened 2023-02-02
Clearer output CLI message when switching profiles

cicdw:
If you store the following two profiles in `~/.prefect/profiles.toml`:

```yaml
active = "one"

[profiles.one]
PREFECT_ORION_DATABASE_CONNECTION_URL = "one"

[profiles.two]
PREFECT_ORION_DATABASE_CONNECTION_URL = "two"
```
and then run the following CLI:
```
prefect profile use two
```
you will find the following misleading output:
```
No Prefect Orion instance specified using profile 'two'. 
Flow run metadata will be stored at the locally configured database: one
```
This is because the output messages were created prior to the profile switch.  In addition, I believe we should disambiguate between the API and the persistent DB layer where possible, as this separation is key for certain use cases.  

This PR updates the CLI message as can be seen in the example below.

### Example
With the same profile setup as above, we now get:
```
prefect profile use two
No Prefect Orion instance specified using profile 'two' - the API will run in ephemeral mode.
```

### Checklist

### DIFF
--- a/src/prefect/cli/profile.py
+++ b/src/prefect/cli/profile.py
@@ -124,7 +124,7 @@ async def use(name: str):
         ),
         ConnectionStatus.EPHEMERAL: (
             exit_with_success,
-            f"No Prefect Orion instance specified using profile {name!r}. Flow run metadata will be stored at the locally configured database: {prefect.settings.PREFECT_ORION_DATABASE_CONNECTION_URL.value()}",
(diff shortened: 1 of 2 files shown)


serinamarie (approved):

madkinsz (approved):
Wonderful thanks!

PR #8390, opened 2023-02-02
Remove type requirement when creating a work pool via CLI

cicdw:
Users can't meaningfully interact with work pool types at this time; therefore, we should remove references of `type` especially from all creation logic to avoid confusing users or misleading them into thinking there are options for the value here.

### Checklist


desertaxle (approved):
LGTM!

peytonrunyan (approved):
Yeah, that seems reasonable. LGTM!

PR #8402, opened 2023-02-03
Add utilities for deprecation

madkinsz:
Similar to our "experimental" utilities, this adds tooling for marking functions, parameters, fields, and settings as deprecated.

Additional work will be done to facilitate automatic renames of items that are _moved_ rather than _removed_.

### Example

```python
from prefect._internal.compatibility.deprecated import deprecated_callable


@deprecated_callable(start_date="Feb 2023", help="Use 'foo' instead.")
def my_function():
    pass


my_function()
```

```
example.py:9: PrefectDeprecationWarning: __main__.my_function has been deprecated. 
It will not be available after Aug 2023. Use 'foo' instead.
  my_function()
```

```python
from prefect._internal.compatibility.deprecated import deprecated_parameter


@deprecated_parameter(
    name="y",
    start_date="Feb 2023",
    help="Use 'x' instead or provide a smaller value.",
    when=lambda y: y > 10,
    when_message="y is greater than 10",
)
def my_function(x: int = 1, y: int = 5):
    pass


# does not warn
my_function()
my_function(y=5)

# warns
my_function(y=15)
```

```
example.py:20: PrefectDeprecationWarning: The parameter 'y' for 'my_function' has been deprecated when y is greater than 10. 
It will not be available after Aug 2023. Use 'x' instead or provide a smaller value.
  my_function(y=15)
```

### Checklist

### DIFF
--- a/src/prefect/_internal/compatibility/deprecated.py
+++ b/src/prefect/_internal/compatibility/deprecated.py
@@ -4,6 +4,10 @@ Utilities for deprecated items.
 When an deprecated item is used, a warning will be displayed. Warnings may not be
 disabled with Prefect settings. Instead, the standard Python warnings filters can be
 used.
+
+Deprecated items require a start or end date. If a start date is given, the end date 
+will be calculated 6 months later. Start and end dates are always in the format MMM YYYY
+e.g. Jan 2023.
 """
 import functools
 import warnings
--- a/src/prefect/_internal/compatibility/deprecated.py
+++ b/src/prefect/_internal/compatibility/deprecated.py
@@ -55,9 +55,12 @@ def generate_deprecation_message(
         parsed_start_date = pendulum.from_format(start_date, DEPRECATED_DATEFMT)
         parsed_end_date = parsed_start_date.add(months=6)
         end_date = parsed_end_date.format(DEPRECATED_DATEFMT)
+    else:
+        # Validate format
+        pendulum.from_format(end_date, DEPRECATED_DATEFMT)
 
     message = DEPRECATED_WARNING.format(name=name, end_date=end_date, help=help)
-    return message
+    return message.rstrip()
 
 
 def deprecated_callable(
--- a/src/prefect/_internal/compatibility/deprecated.py
+++ b/src/prefect/_internal/compatibility/deprecated.py
@@ -1,7 +1,7 @@
 """
 Utilities for deprecated items.
 
-When an deprecated item is used, a warning will be displayed. Warnings may not be
+When a deprecated item is used, a warning will be displayed. Warnings may not be
 disabled with Prefect settings. Instead, the standard Python warnings filters can be
 used.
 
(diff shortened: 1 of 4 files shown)


src/prefect/_internal/compatibility/deprecated.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+DEPRECATED_WARNING = (
+    "{name} has been deprecated. It will not be available after {end_date}. {help}"
+)
+DEPRECATED_MOVED_WARNING = (
+    "{name} has moved to {new_location}. It will not be available at the old import "
+    "path after {end_date}. {help}"
+)
+DEPRECATED_DATEFMT = "MMM YYYY"
```
ahuang11, line 31: I was initially really confused on how MMM could be 3 digits, e.g. 02/2023. Later discovered this is the abbreviated month (Feb) from looking at the tests. Maybe a short comment to indicate this?
```suggestion
DEPRECATED_DATEFMT = "MMM YYYY"  # e.g. Feb 2023
```

src/prefect/_internal/compatibility/deprecated.py
```diff
@@ -0,0 +1,196 @@
+"""
+Utilities for deprecated items.
+
+When an deprecated item is used, a warning will be displayed. Warnings may not be
```
serinamarie, line 4:
```suggestion
When a deprecated item is used, a warning will be displayed. Warnings may not be
```

serinamarie (approved):
Cool! A `deprecator decorator`. Looks good to me :)

src/prefect/_internal/compatibility/deprecated.py
```diff
@@ -0,0 +1,204 @@
...
+
+        ```python
+
+        @deprecated_parameter("y", when=lambda y: y is not None)
+        def foo(x, y = None):
+            return x + 1 + (y or 0)
+        ```
+    """
+
+    when = when or (lambda _: True)
```
ahuang11, line 118: If `when` is False, would this trigger `or`?

peytonrunyan: The deprecation warning will trigger if `when` evaluates to `True`. In this example, passing any value for `y` other than `None` will result in the deprecation warning.

peytonrunyan (approved):
lgtm!

src/prefect/_internal/compatibility/deprecated.py
ahuang11: Thanks, I think I got tripped up on: `Optional[Callable[[Any], bool]]` I misread this as `Optional[Union[Callable, bool]]`

madkinsz: Indeed `when` should be a callable and this sets the default callable to be a noop that always evaluates to true to simplify the downstream logic

PR #8404, opened 2023-02-03
Add experimental artifacts API

serinamarie:
### Overview

Exposes Artifacts API (server-side) atop new `artifact` table behind a `PREFECT_EXPERIMENTAL_ENABLE_ARTIFACTS` flag. Cloud implementation: https://github.com/PrefectHQ/nebula/pull/3513

Information related to 
1. the data and schema migration
2. related work this builds off of 

can be found in https://github.com/PrefectHQ/prefect/pull/8164

### Checklist


src/prefect/server/models/artifacts.py
```diff
@@ -48,3 +50,127 @@ async def read_artifact(
 
     result = await session.execute(query)
     return result.scalar()
+
+
+@inject_db
+async def _apply_artifact_filters(
```
zangell44, line 56: Hmmm this doesn't quite look right to me.

zangell44: Can we also add a test that uses filtering?

zangell44 (changes requested):
Looks good other than the comment on applying filters!

src/prefect/server/models/artifacts.py
serinamarie: Yeah, it wasn't right, good catch!

serinamarie: I'll need add this into the `nebula` as well

src/prefect/server/models/artifacts.py
```diff
@@ -48,3 +50,127 @@ async def read_artifact(
...
+        exists_clause = select(db.FlowRun).where(
+            db.Artifact.flow_run_id == db.FlowRun.id
+        )
+
+        if flow_run_filter:
+            exists_clause = exists_clause.where(flow_run_filter.as_sql_filter(db))
+        if task_run_filter:
+            exists_clause = exists_clause.join(
+                db.TaskRun,
+                db.TaskRun.flow_run_id == db.FlowRun.id,
```
zangell44, line 77: Should we be joining task run and artifact instead here?

serinamarie: Ah, yes 👍

zangell44 (approved):
Excellent work!

Issue #8194, opened 2023-01-18
Cancelling a flow run in a SCHEDULED state leaves it in a CANCELLING state


madkinsz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When runs are cancelled, they transition from CURRENT STATE -> CANCELLING -> CANCELLED where CANCELLED is proposed after successful cancellation of the run.

However, if the flow run is in a SCHEDULED state it cannot be CANCELLED by the agent because the run has not started and has no infrastructure PID. There are a couple solutions here:

1. The agent should report flow runs without PIDs as CANCELLED (maybe only if the prior state was scheduled?)
2. The API should reject transitions to CANCELLING from SCHEDULED and place the run directly in a CANCELLED state

### Reproduction

```python3
prefect deployment build flows/hello_world.py:hello --name 'test' --apply
prefect deployment run hello/test --start-in 1hr 
prefect flow-run cancel 1c5907ac-1057-4db3-9d91-6ad379f39e0c
prefect agent start -q default
```

### Error

```python3
10:39:17.693 | DEBUG   | prefect.agent - Checking for cancelled flow runs...
10:39:17.836 | INFO    | prefect.agent - Found 1 flow runs awaiting cancellation.
10:39:17.836 | ERROR   | prefect.agent - Flow run '1c5907ac-1057-4db3-9d91-6ad379f39e0c' does not have an infrastructure pid attached. Cancellation cannot be guaranteed.
```

### Versions

```Text
Version:             2.7.8+11.g3ca312d078
API version:         0.8.4
Python version:      3.9.12
Git commit:          3ca312d0
Built:               Tue, Jan 17, 2023 11:12 PM
OS/Arch:             darwin/x86_64
Profile:             test
Server type:         cloud
```

PR #8414, opened 2023-02-06
Transition scheduled flow runs immediately to cancelled instead of cancelling

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/8194

This PR adds a flow orchestration rule that rejects attempts to transition scheduled flow runs to a `CANCELLING` state, and instead forces them into a `CANCELLED` state. This keeps the agent from attempting to shutdown infrastructure that does not exist for the scheduled flow run (which is what happens for flow runs in a `CANCELLING` state). 

I went with this approach over the other approach that @madkinsz offered because it keeps orchestration logic on the API.

### Example

Before bug fix:
https://www.loom.com/share/b708257a5fe74433853de084e96faf8c

After Bug Fix:
https://www.loom.com/share/60e60d54de47437b8ad72076cfa4e443
### Checklist


src/prefect/orion/orchestration/core_policy.py
```diff
@@ -811,3 +812,27 @@ async def before_transition(
             ),
         )
         return
+
+
+class HandleCancellingScheduledFlowRuns(BaseOrchestrationRule):
+    """
+    Rejects transitions from Scheduled to Cancelling and set the state to Cancelled.
```
peytonrunyan, line 819:
```suggestion
    Rejects transitions from Scheduled to Cancelling and instead sets the state to Cancelled.
```

peytonrunyan:
Hmmmm, this may be slightly complicated by `AwaitingRetry` being a `SCHEDULED` state

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -811,3 +812,31 @@ async def before_transition(
             ),
         )
         return
+
+
+class HandleCancellingScheduledFlowRuns(BaseOrchestrationRule):
+    """Rejects transitions from Scheduled and Late to Cancelling and set the state to Cancelled.
```
peytonrunyan, line 818:
```suggestion
    """Rejects transitions from Scheduled and Late to Cancelling and sets the state to Cancelled.
```

peytonrunyan:
```suggestion
    """Rejects transitions from Scheduled to Cancelling and sets the state to Cancelled if the flow run has no associated infrastructure process ID.
```

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -42,6 +42,7 @@ def priority():
         return [
             HandleFlowTerminalStateTransitions,
             HandleCancellingStateTransitions,
+            HandleCancellingScheduledFlowRuns,
```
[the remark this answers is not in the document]
anticorrelator: We probably can put it into `HandleCancellingStateTransitions` for brevity

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -811,3 +812,32 @@ async def before_transition(
...
         )
         return
+
+
+class HandleCancellingScheduledFlowRuns(BaseOrchestrationRule):
+    """Rejects transitions from Scheduled to Cancelling, and sets the state to Cancelled, if
+    the flow run has no associated infrastructure process ID.
+
+    The `Cancelling` state is used to clean up infrastructure. If there is not infrastructure
+    to clean up, we can transition directly to `Cancelled`. Runs that are `AwaitingRetry` are
```
anticorrelator, line 822: Does this comment indicate we want to exclude `AwaitingRetry` named states from this short-circuit?

tests/cli/test_flow_run.py
```diff
@@ -221,15 +221,15 @@ class TestCancelFlowRun:
     @pytest.mark.parametrize(
         "state",
         [
-            Scheduled,
             Running,
-            Late,
             Pending,
-            AwaitingRetry,
```
anticorrelator, line 228: As above, should `AwaitingRetry` named states still be put in a `Cancelling` state?

src/prefect/orion/orchestration/core_policy.py
peytonrunyan: It's currently just a context comment. In theory if there is somehow an AwaitingRetry flow run without an infra PID, we should jump it straight to cancelled too

tests/cli/test_flow_run.py
peytonrunyan: As long as they don't have an infra PID

tests/orion/orchestration/test_core_policy.py
```diff
@@ -2599,3 +2600,92 @@ async def test_rejects_cancelled_cancelling_to_anything_but_cancelled(
...
+            session,
+            "flow",
+            *intended_transition,
+        )
+
+        async with HandleCancellingScheduledFlowRuns(ctx, *intended_transition) as ctx:
+            await ctx.validate_proposed_state()
+
+        assert ctx.response_status == SetStateStatus.REJECT
+        assert ctx.validated_state_type == states.StateType.CANCELLED
```
serinamarie, line 2676: Are lines 2663-2676 needed? How is this different than lines 2617-2627?

tests/orion/orchestration/test_core_policy.py
```diff
@@ -2599,3 +2600,92 @@ async def test_rejects_cancelled_cancelling_to_anything_but_cancelled(
...
+            *intended_transition,
+        )
+
+        async with HandleCancellingScheduledFlowRuns(ctx, *intended_transition) as ctx:
+            await ctx.validate_proposed_state()
+
+        assert ctx.response_status == SetStateStatus.ACCEPT
+        assert ctx.validated_state_type == states.StateType.CANCELLING
+
+    async def test_accepts_cancelling_flow_run_with_pid(
```
serinamarie, line 2655:
```suggestion
    async def test_accepts_cancelling_scheduled_flow_run_and_sets_to_cancelling(
```

Would this match better with the other test name?

tests/orion/orchestration/test_core_policy.py
```diff
@@ -2599,3 +2600,92 @@ async def test_rejects_cancelled_cancelling_to_anything_but_cancelled(
...
+        self,
+        session,
+        initialize_orchestration,
+    ):
+        """Flow runs awaiting retry should still go into a cancelling state even though they.
+        are technically "Scheduled".
```
serinamarie, line 2661:
```suggestion
        """Flow runs awaiting retry should still go into a cancelling state as they have an associated pid, even though they are technically "Scheduled".
```

peytonrunyan:
Spoke with Dustin about it and we decided to keep the rules separate and rename them for clarity. I'm going to keep this as a draft until things are running in Cloud.

tests/orion/orchestration/test_core_policy.py
peytonrunyan: This was mostly a behavior sanity check. "This is behavior at point X" and then "This is behavior at point X + <change>". It is 100% duplicating the actual test, and I'm never quite sure how correct it is to do something like. I'll ask in the engineering channel.

src/prefect/orion/orchestration/core_policy.py
```diff
@@ -826,13 +829,14 @@ async def before_transition(
...
-            ),
-        )
-        return
+
+        if not context.run.infrastructure_pid:
+            await self.reject_transition(
+                state=states.Cancelled(),
+                reason=(
+                    "Scheduled flows are transitioned immediately to Cancelled because there is no "
+                    "infrastructure to shut down."
```
madkinsz, line 841: This should probably just read: "Scheduled flow run has no infrastructure to terminate"

madkinsz:
@serinamarie Can you take this over?

cumulus-nrussell:
It would also be great if automations that are designed to cancel a flow would immediately switch to "Cancelled". I have one such automation where the flows just get stuck in "Cancelling" and take up my concurrency limits.

madkinsz:
@cumulus-nrussell we cannot immediately transition them to "Cancelled" if they're not yet cancelled. It is by-design that a flow run does not release its concurrency slot until it is cancelled, otherwise, it your limits may be exceeded. This pull request is specifically for cancelling runs in a SCHEDULED state though and is likely not relevant to your issue.

serinamarie (approved):

PR #8416, opened 2023-02-06
Avoid sending internal server error message on state validation failure

madkinsz:
We should not be including the exception in the response to the user as it could share sensitive details from the server.

### Checklist

### DIFF
--- a/src/prefect/orion/orchestration/rules.py
+++ b/src/prefect/orion/orchestration/rules.py
@@ -238,7 +238,7 @@ class FlowOrchestrationContext(OrchestrationContext):
             await self._validate_proposed_state()
             return
         except Exception as exc:
-            error_key = hashlib.md5(str(exc))
+            error_key = hashlib.md5(str(exc).encode())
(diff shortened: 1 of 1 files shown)


anticorrelator (commented):
This looks reasonable to me

src/prefect/orion/orchestration/rules.py
```diff
@@ -237,11 +238,12 @@ async def validate_proposed_state(
             await self._validate_proposed_state()
             return
         except Exception as exc:
-            logger.exception("Encountered error during state validation")
+            error_key = hashlib.md5(str(exc).encode())
+            error_msg = f"Internal server error while validating state ({error_key}"
```
peytonrunyan, line 242: Possibly a dumb question, but what are we doing with this error key?

madkinsz: The user could provide the key to a server admin to lookup the internal error.

madkinsz:
This is more critical in Cloud than the open source.

Issue #8417, opened 2023-02-06
Upgrading to 2.7.11 breaks older sqlite databases


hateyouinfinity:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

After upgrading to 2.7.11 orion started crashlooping, logs showed `sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) duplicate column name: has_data`.
Rolled back the migrations, downgraded to 2.7.10, tried to investigate.

Alembic upgrade never terminates (see `Error`).
The culpable revision is `f92143d30c25`, introduced in PR #8164.
It prepares `flow_run_state` and `task_run_state` tables for the subsequent migration of artifact data (in `f92143d30c26`)
The relevant part is that it calculates and sets the (boolean) value for `has_data` column based on the value of `data` column.
`data` can either be a `NULL`, a string `'null'`, or a string containing a json value.
The logic in [populate_flow_has_data_in_batches](https://github.com/PrefectHQ/prefect/blob/6b27b476703c84397d751565f8e1ab31654ac0a5/src/prefect/orion/database/migrations/versions/sqlite/2023_01_12_000043_f92143d30c25_create_migration_index.py#L37) is, I believe, intended to set `has_data` to 1 if `data` contains a json value and to 0 otherwise.
The current logic does not achieve that goal.

#### Problems
1. Comparing anything with a NULL via an (in)equality operator results in a NULL ([see 2.](https://www.sqlite.org/lang_expr.html))
2. Since the two parts of the condition are exclusive, disjunction leads to the condition evaluating to 1 for `data='null'` (same would happen for `data=NULL` if not for problem 1)

|        | Goal | data IS NOT NULL | data != 'null' | data IS NOT NULL or data != 'null' |
| ------ | ---- | ---------------- | -------------- | ---------------------------------- |
| NULL   | 0    | 0                | NULL           | NULL                               |
| 'null' | 0    | 1                | 0              | 1                                  |
| JSON   | 1    | 1                | 1              | 1                                  |

Problem 2 results in false positives, migrating `data` from every row to `artifact` table. This affects runs created on every version, though I'm not sure what the impact is besides bloat.
Problem 1 results in the upgrade never terminating if >500 rows with `data=NULL` are present in a table.
[...]

PR #8420, opened 2023-02-06
Properly limit artifact migration to states that have non-null data

anticorrelator:
Thanks to @hateyouinfinity for the excellent bug report.

Addresses the bug raised in https://github.com/PrefectHQ/prefect/issues/8417.

Our artifact migration improperly marked too many states to copy over to the artifact table, causing table bloat and, at times, migrations that might not end. We've implemented the changes suggested in #8417.

Users that have already made it past this migration should don't need to change anything, this change makes it so that users who have not yet upgraded are less likely to need to run a too-large migration.


serinamarie:
Looks good to me 👍

zangell44 (approved):

PR #8422, opened 2023-02-06
Update return type for get_run_logger

carderne:
### Description
Some of the code paths return `PrefectLogAdapter`, so updated the main `get_run_logger()` return type to include `logging.LoggerAdapter`.

### Checklist

### DIFF
--- a/src/prefect/logging/loggers.py
+++ b/src/prefect/logging/loggers.py
@@ -3,7 +3,7 @@ import logging
 from builtins import print
 from contextlib import contextmanager
 from functools import lru_cache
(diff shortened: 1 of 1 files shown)


madkinsz (commented):
Thanks for the pull request! We can't use the `|` syntax since PEP-604 is only supported by 3.10+ and we support 3.7+

carderne:
> Thanks for the pull request! We can't use the `|` syntax since PEP-604 is only supported by 3.10+ and we support 3.7+

Sorry, fixed now to use `Union[]`!

madkinsz (approved):

PR #8425, opened 2023-02-06
Add handling for statement timeouts to state validation

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/8435

Database query timeouts are resulting in an ABORT response from the API:

> Task run '80dc471b-bd97-47d2-9f54-fb8707551b39' received abort during orchestration: Error validating state: DBAPIError("(sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.QueryCanceledError'>: canceling statement due to statement timeout") Task run is in RUNNING state.

Previously, these errors resulted in a 503 response allowing the client to retry. Since an aborted task run can crash the flow and these timeouts are happening frequently, customers are blocked by the lack of retry. It seems likely that this regression is due to a change from #8164 where error handling in state validation was _fixed_.

Here, we explicitly do not capture statement timeouts and instead reraise them. We should probably we doing cleanup instead of just raising an exception, but unfortunately this seems quite complicated and we should first attempt to achieve the previous behavior.

This pull request extends #8423 which adds baseline test coverage for this area.

### Example

Discussion at https://prefecthq.slack.com/archives/C04F1PTMUCQ/p1675455197515479
User report at https://prefect-community.slack.com/archives/CL09KU1K7/p1675449149527989

### Checklist

### DIFF
--- a/src/prefect/orion/orchestration/rules.py
+++ b/src/prefect/orion/orchestration/rules.py
@@ -392,6 +392,12 @@ class TaskOrchestrationContext(OrchestrationContext):
         db: OrionDBInterface,
     ):
         if self.proposed_state is None:
+            # If the run has no state and needs to stay in the same state, there is
+            # nothing to do here
+            if self.run.state is None:
+                self.validated_state = None
+                return
+
             validated_orm_state = self.run.state
             state_data = None
         else:
@@ -414,12 +420,10 @@ class TaskOrchestrationContext(OrchestrationContext):
         self.run.set_state(validated_orm_state)
 
         await self.session.flush()
-        if validated_orm_state:
-            self.validated_state = states.State.from_orm_without_result(
-                validated_orm_state, with_data=state_data
-            )
-        else:
-            self.validated_state = None
+
+        self.validated_state = states.State.from_orm_without_result(
+            validated_orm_state, with_data=state_data
+        )
 
     def safe_copy(self):
         """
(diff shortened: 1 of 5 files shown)


src/prefect/orion/orchestration/rules.py
```diff
@@ -384,6 +392,12 @@ async def _validate_proposed_state(
         db: OrionDBInterface,
     ):
         if self.proposed_state is None:
+            # If the run has no state and needs to stay in the same state, there is
+            # nothing to do here
+            if self.run.state is None:
+                self.validated_state = None
+                return
```
madkinsz, line 399: This was previous behavior but avoids calling `self.session.add(None)` which throws an exception

madkinsz:
Superseded by https://github.com/PrefectHQ/nebula/pull/3482

ccueto36:
> Superseded by https://github.com/PrefectHQ/nebula/pull/3482

Getting a 404 upon clicking this link.

madkinsz:
@ccueto36 that's our internal repository — we're fixing this in Cloud but it does not need a fix in the open source.

ccueto36:
> @ccueto36 that's our internal repository — we're fixing this in Cloud but it does not need a fix in the open source.

Got it. Thanks for the clarification. Looking forward to the fix.

Issue #8426, opened 2023-02-06
Client Error 418 when running prefect work-queue ls


prefectcboyd:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Running `prefect work-queue ls` returns : 
httpx.HTTPStatusError: Client error '418 I'm a teapot' for url 'https://api.prefect.cloud/api/accounts//workspaces//experimental/work_pools/filter'
For more information check: https://httpstatuses.com/418

### Reproduction

```python3
pip install -U prefect .
$ prefect version
Version:             2.7.11
API version:         0.8.4
Python version:      3.9.14
Git commit:          6b27b476
Built:               Thu, Feb 2, 2023 7:22 PM
OS/Arch:             darwin/arm64
Profile:             internal_kube
Server type:         cloud

prefect work-queue ls
```

### Error

```python3
prefect work-queue ls
Traceback (most recent call last):
  File "/Users/christopherboyd/virtualenvs/prefect2/lib/python3.9/site-packages/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/Users/christopherboyd/virtualenvs/prefect2/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 230, in coroutine_wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/Users/christopherboyd/virtualenvs/prefect2/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 181, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/Users/christopherboyd/virtualenvs/prefect2/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/Users/christopherboyd/virtualenvs/prefect2/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
[...]

PR #8427, opened 2023-02-06
Fix error when using `prefect work-queue ls` without enabling work pools

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/8426

The `prefect work-queue ls` command was using a route to lookup work pools associated with with each of the work queues. This doesn't work without enabling a flag. This wasn't caught by my tests because I was using a fixture that auto-enabled the flag. This is no longer the case. 

Updated video of the changes:
https://www.loom.com/share/c697692aa93e4c2f90c185f23e160fed

Note, it shows two default queues because I have a default-agent-pool and a separate pool with a default queue. I could see this potentially causing confusion. 

### Checklist

### DIFF
--- a/src/prefect/cli/work_queue.py
+++ b/src/prefect/cli/work_queue.py
@@ -17,7 +17,6 @@ from prefect.cli._utilities import exit_with_error, exit_with_success
 from prefect.cli.root import app
 from prefect.exceptions import ObjectAlreadyExists, ObjectNotFound
 from prefect.orion.models.workers import DEFAULT_AGENT_WORK_POOL_NAME
-from prefect.orion.schemas.filters import WorkPoolFilter, WorkPoolFilterId
 
 work_app = PrefectTyper(
     name="work-queue", help="Commands for working with work queues."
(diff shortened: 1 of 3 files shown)


desertaxle (approved):
LGTM!

I don't think the double default will be a large issue since users won't have the ability to create another work pool unless they choose to toggle on the experimental flag.

src/prefect/cli/work_queue.py
```diff
@@ -365,7 +366,39 @@ async def ls(
     """
     View all work queues.
     """
-    if not pool:
+    if not pool and PREFECT_EXPERIMENTAL_ENABLE_WORK_POOLS.value() is False:
```
madkinsz, line 369: You should use the `experiment_enabled` helper.

src/prefect/cli/work_queue.py
```diff
@@ -365,7 +366,39 @@ async def ls(
...
+        table = Table(
+            title="Work Queues",
+            caption="(**) denotes a paused queue",
+            caption_style="red",
+        )
+        table.add_column("Name", style="green", no_wrap=True)
```
madkinsz, line 375: Is all that follows here just a replication of some old code that was deleted before it should have been ?

Issue #6774, opened 2022-09-10
Prefect agent fails to stream logs from k8s job pod with multiple containers, causing agent restart


jefflaporte:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When using a job pod spec with multiple containers (eg: with a sidecar), the prefect agent fails to start a watch on the job pod because it doesn't specify which container in the pod to watch, and as a result the k8s API responds with:

```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"a container name must be specified for pod prefect-jobtrbn6-gmxm7, choose one of: [kubexit prefect-job cloud-sql-proxy]","reason":"BadRequest","code":400}
```

When the watch request fails, the prefect agent exits due to the resulting exception.

The solution appears to be to provide the `container` kwarg in the call to `read_namespaced_pod_log` in [`KuberrnetesJob._watch_job`](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/infrastructure/kubernetes.py#L394).

### Reproduction

```python
# Use any Kubernetes job spec with a second container defined in the pod spec
```

### Error response from k8s API
```
HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"a container name must be specified for pod prefect-jobtrbn6-gmxm7, choose one of: [kubexit prefect-job cloud-sql-proxy]","reason":"BadRequest","code":400}\n'
```

### Stack trace

```
Error
2022-09-09 15:48:58.791 PDT Traceback (most recent call last): File "/usr/local/lib/python3.9/site-packages/prefect/cli/_utilities.py", line 41,
 in wrapper return fn(*args, **kwargs) File "/usr/local/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 212,
 in wrapper return run_async_in_new_loop(async_fn, *args, **kwargs) File "/usr/local/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 141,
 in run_async_in_new_loop return anyio.run(partial(__fn, *args, **kwargs)) File "/usr/local/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70,
[...]

Issue #6744, opened 2022-09-08
Mismatched icon in the Create Notification section


khuyentran1401:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

I found that the Failed icon in Review your notification under the Create Notifications section doesn’t match with a typical Failed icon. 

### Reproduction

Click Create Notification -> Click + -> Add Failed as a State

### Error

![2022-09-07_17-38](https://user-images.githubusercontent.com/49108771/189168130-6e24a728-a93c-439a-8aa4-79d332adcdbc.png)

### Browers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```
Version:             2.3.2
API version:         0.8.0
Python version:      3.9.12
Git commit:          6e931ee9
Built:               Tue, Sep 6, 2022 12:36 PM
OS/Arch:             darwin/arm64
Profile:             2.0
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.38.3
```

PR #8430, opened 2023-02-06
Fixes PrefectHQ/prefect#6774 - k8s log stream when using multiple containers in job

jefflaporte:
This fix closes [PrefectHQ/prefect #6744](https://github.com/PrefectHQ/prefect/issues/6774).
It fixes the call to k8s API `read_namespaced_pod_log`, which requires specifying the container name when there are multiple containers in the job pod. 

### Example
For example, when running a `cloud_sql_proxy` sidecar, this will resolve a failed call to `read_namespaced_pod_log`.

### Checklist


jefflaporte:
Note we have tested this PR as working in our k8s cluster, using a job spec with 3 containers in the pod (prefect-job, sidecar, and an exit watchdog).

chrisguidry (approved):
Thanks @jefflaporte, great call!  This makes sense to me, because we've defined a loose "contract" for prefect K8s jobs that they always include a container called `prefect-job` as the first container.

Issue #8439, opened 2023-02-07
Can't delete flow runs


malinkinsa:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When we try to delete the flow from _flow runs_ with status Completed for example, we get an error: _Failed to delete Flow Run_
And in log:
```
sqlalchemy.exc.IntegrityError: (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.ForeignKeyViolationError'>: update or delete on table "task_run" violates foreign key constraint "fk_artifact__task_run_id__task_run" on table "artifact
DETAIL:  Key (id)=(4e03e94b-55bf-40a7-ba13-f776e30efa7a) is still referenced from table "artifact".
[SQL: DELETE FROM flow_run WHERE flow_run.id = %s]
[parameters: (UUID('80f665b5-1731-4a47-92db-a4c8e221fe0c'),)]
(Background on this error at: https://sqlalche.me/e/14/gkpj)
```

### Reproduction

```python3
Run flow, after finish try to delete.
```

### Error

```python3
Traceback (most recent call last):
  File "/opt/prefect/venv/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 64, in __call__
    await self.app(scope, receive, sender)
  File "/opt/prefect/venv/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/opt/prefect/venv/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/opt/prefect/venv/lib/python3.11/site-packages/starlette/routing.py", line 680, in __call__
    await route.handle(scope, receive, send)
  File "/opt/prefect/venv/lib/python3.11/site-packages/starlette/routing.py", line 275, in handle
    await self.app(scope, receive, send)
  File "/opt/prefect/venv/lib/python3.11/site-packages/starlette/routing.py", line 65, in app
    response = await func(request)
               ^^^^^^^^^^^^^^^^^^^
  File "/opt/prefect/venv/lib/python3.11/site-packages/prefect/orion/utilities/server.py", line 101, in handle_response_scoped_depends
[...]

PR #8457, opened 2023-02-08
Remove `Artifact` foreign key constraints

anticorrelator:
resolves https://github.com/PrefectHQ/prefect/issues/8439

After the addition of artifacts, we can no longer delete flow runs because any persisted artifacts will violate their foreign key constraint. These constraints aren't strictly necessary, and we're simply going to drop them.

### Checklist


madkinsz (commented):
Can you add a test that runs a flow then deletes it so we catch this in the future?

madkinsz (approved):

PR #8454, opened 2023-02-08
Rename "Orion"

madkinsz:
## Pull requests

- https://github.com/PrefectHQ/prefect/pull/8434
- https://github.com/PrefectHQ/prefect/pull/8460
- https://github.com/PrefectHQ/prefect/pull/8461
- https://github.com/PrefectHQ/prefect/pull/8463
- https://github.com/PrefectHQ/prefect/pull/8464
- https://github.com/PrefectHQ/prefect/pull/8482
- https://github.com/PrefectHQ/prefect/pull/8483
- https://github.com/PrefectHQ/prefect/pull/8482
- #8469 
- #8486 
- #8487 
- #8490 
- #8488 
- #8489 
- https://github.com/PrefectHQ/prefect/pull/8492
- #8491 
- #8499
- #8496
- #8500 

## Changes

- Default SQLite database name changed from `orion.db` to `prefect.db`
- Logger `prefect.orion` renamed to `prefect.server`
- Constant `ORION_API_VERSION` renamed to `SERVER_API_VERSION`
- Kubernetes deployment template application name changed from `prefect-orion` to `prefect-server`
- Command `prefect kubernetes manifest orion` renamed to `prefect kubernetes manifest server`
- Log config  handler `orion` renamed to `api`
- Class `OrionLogWorker` renamed to `APILogWorker`
- Class `OrionHandler` renamed to `APILogHandler`
- Directory `orion-ui` renamed to `ui`
- Class `OrionRouter` renamed to `PrefectRouter`
- Class `OrionAPIRoute` renamed to `PrefectAPIRoute`
- Class `OrionDBInterface` renamed to `PrefectDBInterface`
- Class `OrionClient` renamed to `PrefectClient`
- Module `prefect.client.orion` renamed to `prefect.client.orchestration`
- Command group `prefect orion` renamed to `prefect server`
- Module `prefect.orion` renamed to `prefect.server`
- The following settings have been renamed:
    - `PREFECT_LOGGING_ORION_ENABLED` → `PREFECT_LOGGING_TO_API_ENABLED`
    - `PREFECT_LOGGING_ORION_BATCH_INTERVAL` → `PREFECT_LOGGING_TO_API_BATCH_INTERVAL`
    - `PREFECT_LOGGING_ORION_BATCH_SIZE` → `PREFECT_LOGGING_TO_API_BATCH_SIZE`
[...]

PR #8463, opened 2023-02-08
Rename `prefect orion` CLI to `prefect server`

madkinsz:
Part of #8454 

Adds a duplicate command group `prefect server` that does everything `prefect orion` does and adds deprecation notices to all of the `prefect orion` commands. The deprecation notices are implemented in the `PrefectTyper` class to deprecate command groups in the future. The `orion` command is hidden from the `prefect --help` menu. The Typer builtin `deprecated` functionality isn't good enough for me, I tried doing that first and it was harder and less coherent with the rest of our interface.

```
❯ prefect orion start       
WARNING: The 'prefect orion' command group has been deprecated. It will not be available after Aug 2023. Use 'prefect server' instead.

 ___ ___ ___ ___ ___ ___ _____ 
| _ \ _ \ __| __| __/ __|_   _| 
|  _/   / _|| _|| _| (__  | |  
|_| |_|_\___|_| |___\___| |_|  
```

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Start server
         run: |
-          prefect orion start&
+          prefect server start&
           PREFECT_API_URL="http://127.0.0.1:4200/api" ./scripts/wait-for-server.py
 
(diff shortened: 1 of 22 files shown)


billpalombi (approved):
Thank you!

PR #8464, opened 2023-02-09
Rename `prefect.orion` module to `prefect.server`

madkinsz:
Part of #8454 

- Renames the `orion` module to `server` and refactors all imports.
- Updates manifests and ignore files
- Updates api references in documentation
- Update codeowners

Adds a `prefect.orion` module stub with some backwards compatibility mechanisms to redirect to `prefect.server`; extended in #8469.


cicdw (approved):
it's really hard to know from a review whether this is stable or not, but looking through it makes sense to me.  Are there any setting access patterns for vars like `PREFECT_ORION_DATABASE_CONNECTION_URL` that might have accidentally been affected?  I don't believe so but wanted to call it out just in case

madkinsz:
@cicdw thanks for looking through anyway.

>  Are there any setting access patterns for vars like PREFECT_ORION_DATABASE_CONNECTION_URL that might have accidentally been affected? 

No those shouldn't be affected yet and since the tests pass I feel good about the access patterns. I'm going to rename those settings in the next pull request though.

PR #8469, opened 2023-02-09
Update rename of the `prefect.orion` module to use more robust alias utilities

madkinsz:
Extends https://github.com/PrefectHQ/prefect/pull/8464 to use actual `importtools` machinery for more reliable module aliasing. This was required to get tests with the old imports to pass but is a separate pull request for reviewability.


serinamarie (approved):
LGTM, also did a little QA: Ran tests locally and in IPython. Created and ran a flow around this: 
```python
import prefect.server
from prefect.orion import models
assert models is prefect.server.models
```
Made a deployment and submitted flow run to agent. Flow run was successful.

madkinsz:
Thank you!

Issue #7763, opened 2022-12-05
Tasks not run sequentially with `.map` and sequential task runner


vdytyniak-exos:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

We see issue with SequentialTaskRunner.

### Reproduction

```python3
@task()
def copy_file(n: int):
    get_run_logger().info('Start file copying')
    time.sleep(n)
    get_run_logger().info('Completed file copying')


@flow(task_runner=SequentialTaskRunner())
def sample() -> None:
    copy_file.map([3, 4])


if __name__ == '__main__':
    sample()

```

### Error

18:05:29.909 | INFO    | Flow run 'devious-aardwolf' - Created task run 'copy_file-6cc8b0a3-0' for task 'copy_file'
18:05:29.910 | INFO    | Flow run 'devious-aardwolf' - Executing 'copy_file-6cc8b0a3-0' immediately...
18:05:29.916 | INFO    | Flow run 'devious-aardwolf' - Created task run 'copy_file-6cc8b0a3-1' for task 'copy_file'
18:05:29.917 | INFO    | Flow run 'devious-aardwolf' - Executing 'copy_file-6cc8b0a3-1' immediately...
18:05:30.468 | INFO    | Task run 'copy_file-6cc8b0a3-1' - Start file copying
18:05:30.534 | INFO    | Task run 'copy_file-6cc8b0a3-0' - Start file copying
18:05:33.536 | INFO    | Task run 'copy_file-6cc8b0a3-0' - Completed file copying
18:05:33.726 | INFO    | Task run 'copy_file-6cc8b0a3-0' - Finished in state Completed()
18:05:34.469 | INFO    | Task run 'copy_file-6cc8b0a3-1' - Completed file copying
18:05:34.658 | INFO    | Task run 'copy_file-6cc8b0a3-1' - Finished in state Completed()

### Versions

```Text
2.6.9
```

PR #8473, opened 2023-02-09
Fix sequential execution with mapped tasks using the SequentialTaskRunner

peytonrunyan:
Closes https://github.com/PrefectHQ/prefect/issues/7763

Mapped tasks were being executed asynchronously even when using the SequentialTaskRunner. This was because inside of the `begin_task_map` function we get a list of partial `get_task_call_return_value` functions, which are async, and then we call ` return await gather(*task_runs)`.

My solution sidesteps this by doing a check right before we call gather to see if the runner is sequential, and if so, executes the coroutines without the gather. 

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1059,6 +1059,11 @@ async def begin_task_map(
             )
         )
 
+    # Maintain the order of the task runs when using the sequential task runner
+    runner = task_runner if task_runner else flow_run_context.task_runner
+    if runner.concurrency_type == TaskConcurrencyType.SEQUENTIAL:
+        return [await task_run() for task_run in task_runs]
+
     return await gather(*task_runs)
 
 
@@ -1145,7 +1150,7 @@ async def create_task_run_future(
     # Default to the flow run's task runner
     task_runner = task_runner or flow_run_context.task_runner
 
-    # Generate a name for the future
+    # Generate a name for th future
     dynamic_key = _dynamic_key_for_task_run(flow_run_context, task)
     task_run_name = f"{task.name}-{dynamic_key}"
 
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1150,7 +1150,7 @@ async def create_task_run_future(
     # Default to the flow run's task runner
     task_runner = task_runner or flow_run_context.task_runner
 
-    # Generate a name for th future
+    # Generate a name for the future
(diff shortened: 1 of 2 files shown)


tests/test_tasks.py
```diff
@@ -2978,6 +2979,34 @@ def my_flow():
...
+        def sleepy_task(n, mock_item):
+            time.sleep(n)
+            mock_item(n)
+            return n
+
+        @flow
+        def my_flow(mock_item):
+            sleepy_task.map([3, 0], unmapped(mock_item))
+
+        mock_item = MagicMock()
```
madkinsz, line 2998: Note you could just be appending integers to a list instead of calling a mock.

tests/test_tasks.py
```diff
@@ -2978,6 +2979,34 @@ def my_flow():
         task_states = my_flow()
         assert [state.result() for state in task_states] == [6, 7, 8]
 
+    def test_map_with_sequential_runner_is_sequential(self):
```
madkinsz, line 2982: As far as making this more performant of a test, you could map over `range(10)` or something then have each task sleep for `1 / (n * 100)` to do much smaller sleeps but still have the first one sleep the longest?

src/prefect/engine.py
```diff
@@ -1021,6 +1021,11 @@ async def begin_task_map(
             )
         )
 
+    # Maintain the order of the task runs when using the sequential task runner
+    runner = task_runner if task_runner else flow_run_context.task_runner
+    if runner.concurrency_type == TaskConcurrencyType.SEQUENTIAL:
+        return [await task_run() for task_run in task_runs]
```
madkinsz, line 1027: This is a surprisingly simple solution. I'm into it for now. 💯

tests/test_tasks.py
peytonrunyan: I tried this and couldn't make it work for some reason

madkinsz:
@peytonrunyan can you get this passing CI?

src/prefect/engine.py
```diff
@@ -1102,7 +1107,7 @@ async def create_task_run_future(
     # Default to the flow run's task runner
     task_runner = task_runner or flow_run_context.task_runner
 
-    # Generate a name for the future
+    # Generate a name for th future
```
peytonrunyan, line 1110:
```suggestion
    # Generate a name for the future
```

madkinsz (approved):
Thanks @peytonrunyan <3

PR #8496, opened 2023-02-13
Update docstrings and messages

tpdorsey:
Remove Orion from docstrings, comments, and messages

### Checklist


src/prefect/cli/kubernetes.py
```diff
@@ -1,5 +1,5 @@
 """
-Command line interface for working with Orion on Kubernetes
+Command line interface for working with Prefect on Kubernetes
```
peytonrunyan, line 2: I gave the replacement guidance a look and it doesn't appear to be 100% clear on cases like this. Did we decided that we can use a generic `Prefect` or should this be `the Prefect server` or `Prefect server`?

madkinsz: Not all of these commands are related to the server, e.g. installing an agent.

madkinsz:
@tpdorsey I've rebased this and resolve a bunch of conflicts where I made the same edits :D

PR #8500, opened 2023-02-13
Update default database to be `prefect.db` instead of `orion.db`

madkinsz:
If `PREFECT_API_DATABASE_CONNECTION_URL` is set, nothing changes here.

Otherwise, the default value is now a SQLite database called `prefect.db`. For backwards compatibility, if the `orion.db` file exists and `prefect.db` does not, we will use that as the default instead. 

No warnings are displayed here, users can keep using the old database name. I'm very hesitant to move the database for the user as it could be in-use.


serinamarie (commented):
LGTM :)

PR #8501, opened 2023-02-13
Populate default result metadata

anticorrelator:
Adds default artifact metadata to `LiteralResults` and `PersistedResults`. Currently these changes are inferred, but `WritableBlock` functionality will be extended to provide a richer and more robust default behavior.

These changes should be fully backwards compatible with old clients. Old clients will write `Artifacts` with no metadata, but new clients will automatically populate an artifact `type` and `metadata_["description"]`. States will be read from the server with these metadata fields omitted, ensuring that results *with* metadata created by newer clients can still be handled by old clients.

Serverside, a currently unnamed concept of "state results" is being introduced, which then get converted to an `Artifact` in the backend. In time, we'll want to formalize this `result` interface as a pydantic schema, but for now I'm leaving this flexible until we're confident about the structure.

### Checklist


src/prefect/orion/schemas/core.py
```diff
@@ -1069,14 +1069,14 @@ class Artifact(ORMBaseModel):
         default=None,
         description="An identifier for how this artifact is persisted.",
     )
-    data: Optional[Any] = Field(
+    data: Dict[str, Any] = Field(
```
madkinsz, line 1072: I think this is a breaking change

madkinsz: Oh sorry this `Artifact.data` not `State.data` — still breaking but perhaps not impactful

anticorrelator: Good note, reverted

zangell44 (approved):

PR #8515, opened 2023-02-14
DEBUG: Add prints to deadlocking event test

madkinsz:


madkinsz:
Success. Failure looks like:

```
two: waiting for event_two to be created
two: setting event_one
one: waiting for event_two to be created
one: setting event_two
one: waiting for event_one
```

PR #8520, opened 2023-02-14
Fix test bug where `two` did not wait for creation of `event_one`

madkinsz:
Enables the test that was flaking and solves the issue which was actually just a mistake in the test itself not the implementation.

Verified with flake-finder in https://github.com/PrefectHQ/prefect/pull/8515


serinamarie (approved):

chrisguidry (approved):
:D

PR #8522, opened 2023-02-14
Move server type detection into PrefectClient

### LINKED ISSUE #8521: Move ephemeral/hosted/cloud detection to PrefectClient
### First check

### Description

`PrefectClient` should know what type of server it's talking to.

### Impact

As part of introducing events to Prefect core we need to have a way to easily detect if we're currently connected to Cloud or to a local/ephemeral server. 

### Additional context

_No response_

bunchesofdonald:
This moves the server type detection from the version CLI to the PrefectClient itself and then updates the profile CLI to also use this new attribute. Having this attribute in the client itself will allow us to easily detect which events client to use.

Closes #8521 

### Checklist

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -49,13 +49,21 @@ from prefect.settings import (
     PREFECT_API_REQUEST_TIMEOUT,
     PREFECT_API_TLS_INSECURE_SKIP_VERIFY,
     PREFECT_API_URL,
+    PREFECT_CLOUD_API_URL,
 )
+from prefect.utilities.collections import AutoEnum
 
 if TYPE_CHECKING:
     from prefect.flows import Flow
     from prefect.tasks import Task
 
 from prefect.client.base import PrefectHttpxClient, app_lifespan_context
+
+
+class ServerType(AutoEnum):
+    EPHEMERAL = AutoEnum.auto()
+    HOSTED = AutoEnum.auto()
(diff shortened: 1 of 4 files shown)


src/prefect/client/orchestration.py
```diff
@@ -58,6 +60,12 @@
 from prefect.client.base import PrefectHttpxClient, app_lifespan_context
 
 
+class ServerType(AutoEnum):
+    EPHEMERAL = AutoEnum.auto()
+    HOSTED = AutoEnum.auto()
```
madkinsz, line 65: Probably could/should use SERVER now?

madkinsz (approved):

tests/cli/test_version.py
```diff
@@ -62,7 +62,7 @@ def test_correct_output_ephemeral_sqlite(monkeypatch):
 Built:               {built.to_day_datetime_string()}
 OS/Arch:             {sys.platform}/{platform.machine()}
 Profile:             {profile.name}
-Server type:         ephemeral
+Server type:         EPHEMERAL
```
chrisguidry, line 65: @madkinsz We don't have any reason to believe that the lower-cased string values were in some way load-bearing do we?

madkinsz: I can't think of any! Although it does seem nicer lowercase in this command output :)

bunchesofdonald: I can make it lowercase.

PR #8523, opened 2023-02-14
Fix runtime bug where async context vars were not propagated

madkinsz:
When we called `await self.context.run(self.fn, *self.args, **self.kwargs)` the coroutine would be _created_ within the context but not _run_ within the context. Instead, we must create a new _async task_ with the desired context. This is a little slower, but we need context variables to propagate.


chrisguidry (approved):
Thanks for the clear tests!

PR #8530, opened 2023-02-14
Add re-raise of exceptions during runtime startup

madkinsz:
Switches from an event to signal readiness to a future, allowing us to attach an exception. Otherwise, exceptions are hidden in the runtime thread and do not result in an exception in the main thread.

### DIFF
--- a/tests/_internal/concurrency/test_runtime.py
+++ b/tests/_internal/concurrency/test_runtime.py
@@ -1,5 +1,6 @@
 import asyncio
 import contextvars
+from unittest.mock import MagicMock
 
 import anyio
 import pytest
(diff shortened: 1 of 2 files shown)


tests/_internal/concurrency/test_runtime.py
```diff
@@ -24,6 +25,20 @@ def test_runtime_context_manager():
         pass
 
 
+def test_runtime_with_failure_in_loop_thread_start():
```
madkinsz, line 28: Previously, this would hang for 60s then throw a `pytest.PytestUnhandledThreadExceptionWarning: Exception in thread RuntimeLoopThread`

chrisguidry (approved):
:heart: Thank you for these very focused PRs!!

PR #8546, opened 2023-02-15
Hook events client into PrefectClient and ensure that its torn-down when needed.

bunchesofdonald:
This adds an `events` method to `PrefectClient` that will initialize and enter the context of an EventClient. The type of event client depends on the `server_type` being used by the PrefectClient. Only `cloud` will get a 'working' EventClient, all others will get a NullEventClient that is a no-op. Also added to `PrefectClient` is a `disconnect_events` method that will ensure that the events client context is exited and that the PrefectClient no longer maintains a reference to the EventsClient. The engine also calls `disconnect_events` whenever we're exiting orchestration of a task/flow so that we ensure that any open connections to Cloud's websocket endpoint are closed.

### Checklist

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -269,15 +269,9 @@ class PrefectClient:
                 self._events_client = NullEventsClient()
 
             await self._events_client.__aenter__()
+            self._exit_stack.push_async_exit(self._events_client)
 
         return self._events_client
-
-    async def disconnect_events(self):
-        if self._events_client:
-            await self._events_client.__aexit__(
-                exc_tb=None, exc_type=None, exc_val=None
-            )
-            self._events_client = None
 
     # API methods ----------------------------------------------------------------------
 
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -17,8 +17,14 @@ import prefect.server.schemas as schemas
 import prefect.settings
 import prefect.states
 from prefect._internal.compatibility.deprecated import deprecated_callable
+from prefect._internal.compatibility.experimental import experiment_enabled
 from prefect.client.schemas import FlowRun, OrchestrationResult, TaskRun
 from prefect.deprecated.data_documents import DataDocument
+from prefect.events.clients import (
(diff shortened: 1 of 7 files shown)


bunchesofdonald:
@madkinsz and @chrisguidry mind taking a look at this? I _think_ this is the right track, but testing the disconnect part is proving difficult so I just wanted to make sure before I go down that road.

src/prefect/engine.py
```diff
@@ -1340,6 +1343,7 @@ async def begin_task_run(
                 # When a a task run finishes on a remote worker flush logs to prevent
                 # loss if the process exits
                 APILogHandler.flush(block=True)
+                await client.disconnect_events()
```
chrisguidry, line 1346: This lines up with my (limited) understanding of how we make sure to flush all our run logs to the server, and is where I was expecting we'd tie into the current lifecycle of tasks and flows.

src/prefect/client/orchestration.py
```diff
@@ -247,6 +255,30 @@ def api_url(self) -> httpx.URL:
...
+                self._events_client = PrefectCloudEventsClient(
+                    api_url=PREFECT_CLOUD_API_URL.value(),
+                    api_key=prefect.settings.PREFECT_API_KEY.value(),
+                )
+            else:
+                self._events_client = NullEventsClient()
+
+            await self._events_client.__aenter__()
+
+        return self._events_client
```
chrisguidry, line 273: I notice that the `PrefectClient` has an `_exit_stack`, is this something we could join in on as well?

src/prefect/engine.py
madkinsz: 👍 these make sense to me

src/prefect/client/orchestration.py
bunchesofdonald: Possibly, I'll look at it.

chrisguidry (approved):
:hot_pepper:

madkinsz:
@bunchesofdonald fyi there's an `experimental` label that would categorize this correctly in the changelog :)

Issue #6676, opened 2022-09-01
`prefect block delete` does not handle missing `/` well


zanieb:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If you call `prefect block delete` with a value without a `/` it throws an exception. Instead, it should have a nice error message suggesting the proper format.

### Reproduction

```python
$ prefect block delete foobar
```

### Error

```
Traceback (most recent call last):
  File "/Users/mz/dev/prefect/src/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/Users/mz/dev/prefect/src/prefect/utilities/asyncutils.py", line 212, in wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/Users/mz/dev/prefect/src/prefect/utilities/asyncutils.py", line 141, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/mz/dev/prefect/src/prefect/cli/block.py", line 204, in block_delete
    block_type_slug, block_document_name = slug.split("/")
ValueError: not enough values to unpack (expected 2, got 1)
An exception occurred.
```

### Versions

```
[...]

PR #8555, opened 2023-02-16
Resolving #6676 missing slash in block command

westford14:
Resolves issue #[6676](https://github.com/PrefectHQ/prefect/issues/6676) where an exception is raised instead of an error being raised with the `prefect block` command is run with a slug missing a `/`.

### Example

An example of the new behavior is:

```
(.venv) (base) ➜  prefect git:(main) prefect block delete foobar
slug must contain '/' -- foobar is not valid
```

and 

```
(.venv) (base) ➜  prefect git:(main) ✗ prefect block inspect footbar
Slug must contain '/' -- footbar is not valid
```

### Checklist

- [X] This pull request references any related issue by including "closes https://github.com/PrefectHQ/prefect/issues/6676"
- [X] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`

### DIFF
--- a/src/prefect/cli/block.py
+++ b/src/prefect/cli/block.py
@@ -219,6 +219,8 @@ async def block_delete(
             except ObjectNotFound:
                 exit_with_error(f"Deployment {block_id!r} not found!")
         elif slug is not None:
+            if "/" not in slug:
+                exit_with_error(f"Slug must contain '/' -- {slug} is not valid")
             block_type_slug, block_document_name = slug.split("/")
             try:
                 block_document = await client.read_block_document_by_name(
@@ -286,6 +288,8 @@ async def block_inspect(
             except ObjectNotFound:
                 exit_with_error(f"Deployment {block_id!r} not found!")
         elif slug is not None:
+            if "/" not in slug:
+                exit_with_error(f"Slug must contain '/' -- {slug} is not valid")
             block_type_slug, block_document_name = slug.split("/")
             try:
                 block_document = await client.read_block_document_by_name(


serinamarie:
Hi @westford14, thanks for contributing! A few things to get squared away:

I think it would be nice to give a little example in the error message, something like:

```bash
Please include the block slug followed by a '/' and then the block name. \
e.g. 'prefect block delete string/my-string-block'
```

I think the error message you provided fulfills the case where the user tries to delete a valid block and provides no slug:

```bash
❯ prefect block delete test-string-block
Slug must contain '/' -- test-string-block is not valid
```

Without a clearer error message, though, a user may then attempt to prefix or suffix a `/` and feel confused if they don't know what slug refers to:
```
❯ prefect block delete /test-string-block
Block '/test-string-block' not found!
❯ prefect block delete test-string-block/
Block 'test-string-block/' not found!
```

Finally, let's say they format it correctly but add the extra slash to be safe; they do receive the same mysterious error as before. 
```bash
❯ prefect block delete string/test-string-block/
  File "/Users/bean/code/prefect/src/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/Users/bean/code/prefect/src/prefect/utilities/asyncutils.py", line 230, in coroutine_wrapper
    return run_async_in_new_loop(async_fn, *args, **kwargs)
  File "/Users/bean/code/prefect/src/prefect/utilities/asyncutils.py", line 181, in run_async_in_new_loop
    return anyio.run(partial(__fn, *args, **kwargs))
  File "/Users/bean/code/prefect/venv-feb/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/Users/bean/code/prefect/venv-feb/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/Users/bean/.pyenv/versions/3.9.13/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/bean/.pyenv/versions/3.9.13/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/Users/bean/code/prefect/venv-feb/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/bean/code/prefect/src/prefect/cli/block.py", line 224, in block_delete
    block_type_slug, block_document_name = slug.split("/")
ValueError: too many values to unpack (expected 2)
An exception occurred.
```

So perhaps we would want to handle `block_id` the same way. 

Corresponding tests for these would also help to ensure the behavior is as expected 🙂

PR #8559, opened 2023-02-16
Implement watcher pattern at the futures level

madkinsz:


src/prefect/_internal/concurrency/from_sync.py
```diff
@@ -0,0 +1,64 @@
...
+    `call_soon_in_runtime_thread` or the main thread will not be watching for work.
+
+    Returns a future.
+    """
+    current_future = get_current_future()
+    if current_future is None:
+        raise RuntimeError("No watching future found.")
+
+    if current_future.owner_thread_ident != threading.main_thread().ident:
+        raise RuntimeError("Watching future is not owned by the main thread.")
```
madkinsz, line 61: We could add a `call_soon_in_watcher_thread` that doesn't have this limitation

src/prefect/_internal/concurrency/futures.py
```diff
@@ -0,0 +1,201 @@
...
+        logger.debug("Finished work item %r", self)
+
+
+class WatchingFuture(abc.ABC):
+    """
+    A watching future allows work to be sent back to the thread that owns the future by
+    the call the future represents.
+
+    Work sent back to the thread will be executed when the owner waits for the result
+    of the future.
```
chrisguidry, line 129: This docstring is intellectually load-bearing and maybe should be incorporated into a future module doc for this.  This is the clearest explanation for it that we've got so far.  Still think "watching" is the wrong word and I'll keep thinking on it...

src/prefect/_internal/concurrency/from_async.py
```diff
@@ -0,0 +1,65 @@
+import asyncio
+import threading
+
+from prefect._internal.concurrency.futures import (
```
chrisguidry, line 4: I very much appreciate the two-module structure and the separate `futures` module

src/prefect/_internal/concurrency/futures.py
madkinsz: Yes I agree an overview at the module level would make a lot of sense, perhaps even in `concurrency/__init__.py`

src/prefect/_internal/concurrency/futures.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+
+    def __init__(self) -> None:
+        self.owner_thread_ident = threading.get_ident()
+        logger.debug("Created future %r", self)
+
+    def wrap_future(self: Self, future) -> Self:
+        asyncio.futures._chain_future(future, self)
```
madkinsz, line 137: Please don't judge this one — it's not worth maintaining ourselves. If they remove it, we'll deal with it then.

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -0,0 +1,212 @@
...
+            self.future.set_result(result)
+
+        logger.debug("Finished work item %r", self)
+
+
+class Supervisor(abc.ABC):
+    """
+    A supervisor allows work to be sent back to the thread that owns the supervisor.
+
+    Work sent to the supervisor will be executed when the owner waits for the result.
```
chrisguidry, line 126: :boom:

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+
+    def __init__(self) -> None:
+        self.owner_thread_ident = threading.get_ident()
+        self._future = None
+        logger.debug("Created supervisor %r", self)
+
+    def send_call(self, __fn, *args, **kwargs) -> concurrent.futures.Future:
+        work_item = WorkItem.from_call(__fn, *args, **kwargs)
+        self._put_work_item(work_item)
```
chrisguidry, line 136: We should probably declare the `abc.abstractmethod` for this

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -0,0 +1,212 @@
+"""
+Implementations of supervisors for futures, which allow work to be sent back to the
+thread waiting for the result of the future.
+"""
```
chrisguidry, line 4: This whole module is clutch, :a: :heavy_plus_sign: 

These primitives really get to the heart of what client-side orchestration is all about.  It's like a micro-scale version of the hybrid architecture.  It's _your_ thread/event_loop, and you're asking `prefect` to advise you what to run on it.  `prefect` is a passenger navigating, not driving the car.  Love it!

chrisguidry (approved):

PR #8562, opened 2023-02-16
Add note about `prefect config validate` to release notes

madkinsz:
I forgot to mention this previously.

I've also moved the full list of changes to the "Deprecated" section so it flows a little better.


billpalombi (approved):
lgtm

PR #8564, opened 2023-02-16
pulling in api (and other) patterns from cloud

stackoverfloweth:
This PR applies changes from [latest prefect-ui-library](https://github.com/PrefectHQ/prefect-ui-library/pull/1161) api paradigm.

- removes dupe apis
- removes references to long deprecated base api class
- moves AdminApi and ServerSettings out of prefect-ui-library since it's only used in oss
- follows nebula in defining `usePrefectApi`, which includes oss specific apis along with everything from `useWorkspaceApi`
- also follows nebula in using mirrored compositions to export some logic out of App.vue


ui/src/compositions/useMobileMenuOpen.ts
```diff
@@ -0,0 +1,31 @@
+import { Ref, ref } from 'vue'
```
pleek91, line 1: side quest? ❤️

ui/src/pages/AppRouterView.vue
```diff
@@ -22,42 +22,35 @@
...
   provide(workspaceRoutesKey, routes)
 
-  const mobileMenuOpen = ref(false)
-  const showMenu = computed(() => media.lg || mobileMenuOpen.value)
-
-  function toggle(): void {
-    mobileMenuOpen.value = !mobileMenuOpen.value
+  const healthy = await api.health.checkHealth()
+  if (!healthy) {
+    showToast(`Can't connect to Server API at ${config.baseUrl}. Check that it's accessible from your machine.`, 'error', { timeout: false })
```
pleek91, line 49: This needs to be reworked a little. It was intentionally in a utility before so that it wouldn't block the app rendering. As is nothing in the app will start rendering until this health check is done. We want it to be fire and forget. This might apply to the nebula version as well.

ui/src/pages/Settings.vue
```diff
@@ -26,13 +26,14 @@
...
+  import { usePrefectApi } from '@/compositions/usePrefectApi'
   import { activeColorMode } from '@/utilities/colorMode'
 
   const crumbs = [{ text: 'Settings' }]
 
-  const engineSettings = await adminApi.getSettings()
-  const version = await adminApi.getVersion()
+  const api = usePrefectApi()
+  const engineSettings = await api.admin.getSettings()
+  const version = await api.admin.getVersion()
```
pleek91, line 36: Unrelated but seeing this I noticed we could do better

```suggestion
  const [engineSettings, version] = await Promise.all([
    api.admin.getSettings(),
    api.admin.getVersion(),
  ])
```

ui/src/compositions/useMobileMenuOpen.ts
stackoverfloweth: you know me

ui/src/pages/AppRouterView.vue
stackoverfloweth: hows this?
```ts
  api.health.isHealthy().then(healthy => {
    if (!healthy) {
      showToast(`Can't connect to Server API at ${config.baseUrl}. Check that it's accessible from your machine.`, 'error', { timeout: false })
    }
  })
```

pleek91: Yep that works. Cannot believe you're using promise.then 😱 lol

stackoverfloweth: Lol. They have their place

pleek91 (approved):

PR #8576, opened 2023-02-17
Update `quote` handling in input resolution to skip descending into the quoted expression

madkinsz:
Adds a `StopVisiting` signal that can be raised in a visit function to prevent further recursion then adds usage of it to input parameter resolution. Previously, we would recurse into objects and just not modify them if they were quoted. Quoting is much more useful if we avoid traversing the object though.


serinamarie (approved):
LGTM :)

Issue #8542, opened 2023-02-15
Unable to handle Pydantic BaseModel subclasses with custom init functions


masonmenges:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Attempting to instantiate Pydantic Subclasses with custom init functions within a task or referencing the model within a task results in the following error `model_instance = typ(
TypeError: MyModel.__init__() missing 1 required keyword-only argument: 'bar'` running the same code in the example below with both tasks as functions completes successfully.

### Reproduction

```python3
from pydantic import BaseModel
from prefect import flow, task

class MyModel(BaseModel):
    id: int
    info: str = "Foo"

    def __init__(self, id: int = 1, *, bar: str, **data) -> None:
        """My custom init!"""
        super().__init__(id=id, bar=bar, **data)

@task
def return_model(x) -> MyModel:
    return MyModel(id=x, bar="foo")


@task()
def use_model(model: MyModel):
    print(model.id)


@flow
def pydantic_test(x: int = 10):
    model = return_model(x)
    use_model(model)


if __name__ == "__main__":
    print(pydantic_test(2))
```

### Error

```python3
Traceback (most recent call last):
  File "/Users/masonmenges/Repos/support-directory/prefect-support/cloud2/masonm/test/test13.py", line 37, in <module>
    print(pydantic_test(2))
  File "/opt/homebrew/Caskroom/miniconda/base/envs/cloud2/lib/python3.10/site-packages/prefect/flows.py", line 456, in __call__
    return enter_flow_run_engine_from_flow_call(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/cloud2/lib/python3.10/site-packages/prefect/engine.py", line 170, in enter_flow_run_engine_from_flow_call
    return anyio.run(begin_run)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/cloud2/lib/python3.10/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/cloud2/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
[...]

PR #8584, opened 2023-02-17
Add wrapper for exceptions encountered while resolving parameter inputs

madkinsz:
Improve error messaging for cases like #8542 

e.g.

```python
13:23:00.102 | ERROR   | Task run 'use_model-0' - Crash detected! Execution was interrupted by an unexpected exception: Traceback (most recent call last):
  File "/Users/mz/dev/prefect/src/prefect/engine.py", line 1765, in resolve_inputs
    resolved_parameters[parameter] = await run_sync_in_worker_thread(
  File "/Users/mz/dev/prefect/src/prefect/utilities/asyncutils.py", line 91, in run_sync_in_worker_thread
    return await anyio.to_thread.run_sync(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/Users/mz/dev/prefect/src/prefect/utilities/collections.py", line 344, in visit_collection
    model_instance = typ(
TypeError: __init__() missing 1 required keyword-only argument: 'bar'

The above exception was the direct cause of the following exception:

PrefectException: Failed to resolve inputs in parameter 'model'. If your parameter type is not supported, consider using the `quote` annotation to skip resolution of inputs.
```


tharwan:
Hi, as the reason for https://github.com/PrefectHQ/prefect/issues/8542, I think this is very much an improvement. However, I don't think an error message is the right way to ultimately resolve this. 

From my point of view the core problem here is an assumption about Pydantic BaseModel that does not hold up.

I think it would be fair to have a warning about automatically treating a BaseModel as "quoted" and no Futures being resolved. However I lack understanding why creating a new instance is necessary in the first place.

src/prefect/engine.py
```diff
@@ -1723,6 +1724,8 @@ async def resolve_inputs(
     Raises:
         UpstreamTaskError: If any of the upstream states are not `COMPLETED`
     """
+    if not parameters:
+        return parameters
```
madkinsz, line 1728:
```suggestion
        return {}
```

peytonrunyan (approved):
In the near term this improves interpretability, and it doesn't close any doors for us long term, so it feels like a good move. LGTM!

src/prefect/engine.py
```diff
@@ -1861,14 +1861,35 @@ def resolve_input(expr, context):
...
 
-    return visit_collection(
-        parameters,
-        visit_fn=resolve_input,
-        return_data=return_data,
-        max_depth=max_depth,
-        remove_annotations=True,
-        context={},
-    )
+    if isinstance(parameters, dict):
```
jakekaplan, line 1864: @madkinsz `parameters` sometimes seems to be called like `resolve_inputs(parameters=wait_for, ...)` where `wait_for` is a `Iterable[PrefectFuture]`. Is this an okay solution to conditionally do this?

madkinsz: Thanks for looking into this!

We should probably do `resolve_inputs(parameters={"wait_for": wait_for})` instead since that key is reserved anyway it'll be helpful for understanding where failures come from in that case as well.

jakekaplan: ah got it! thanks will update and hopefully CI should clear up 🤞

src/prefect/engine.py
```diff
@@ -1861,14 +1865,28 @@ def resolve_input(expr, context):
...
-    )
+    resolved_parameters = {}
+    for parameter, value in parameters.items():
+        try:
+            resolved_parameters[parameter] = visit_collection(
+                value,
+                visit_fn=resolve_input,
+                return_data=return_data,
+                # we're manually going 1 layer deeper here
+                max_depth=max_depth - 1,
```
jakekaplan, line 1876: @madkinsz 

the remaining errors seem to be caused by:
https://github.com/PrefectHQ/prefect/blob/main/src/prefect/engine.py#L1027
```
    parameters = await resolve_inputs(parameters, max_depth=1)
```
This will will lead errors due the `max_depth`. I'm having trouble reasoning if the `max_depth` should change in that line OR changed here because we're manually going 1 layer deep into the parameters?

jakekaplan: I may be.... outside of my depth...

jakekaplan (approved):

PR #8588, opened 2023-02-18
Add outlier Docker version name format function and tests

### LINKED ISSUE #8587: Docker Version Checker: outlier Docker version format throws exception
### First check

### Bug summary

(I also submitted a PR for this: [#8588](https://github.com/PrefectHQ/prefect/pull/8588))

We use SLES linux distributions with Docker version name formats that fail the Prefect version checker: "20.10.17-ce" ("-ce" being the issue)
- State: Prefect uses `packaging.verison.parse` to convert Docker version names into a `Version` object that is used to check for a minimum version (20.10.0 or higher).
- Issue: Our Docker version names - "20.10.17-ce" - is seen as a invalid version string by `packaging.version.parse` (the Enterprise Edition in our linux distress would fail as well, since they end with "-ee"). Because our version isn't failing the intended validation (i.e., lower than what's required) it took some time to figure out why it was failing.

The resolution to this issue in my own environment was pretty straightforward: replace "-ce" or "-ee" with "" for the version name string before passing it into the `parse` method. This wouldn't account for other possible version name outliers, but I did some research and couldn't find any other variants (obviously doesn't mean they don't exist).

I have a PR that resolves this. It puts that logic into a separate function so it can be tested, and so it's easier to add additional outlier version name cases in the future.

### Reproduction

```python3
# pulled from prefect.infrastructure.docker.py, line 676

# `docker_client.version()["Version"]` replaced with relevant "invalid" version
user_version = packaging.version.parse("20.10.17-ce")
```

### Error

```python3
  File "/Users/ben/Documents/github/prefect/_delete.py", line 20, in <module>
    packaging.version.parse("20.10.17-ce")
  File "/Users/ben/Documents/github/prefect/venv/lib/python3.10/site-packages/packaging/version.py", line 52, in parse
    return Version(version)
  File "/Users/ben/Documents/github/prefect/venv/lib/python3.10/site-packages/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '20.10.17-ce'
```

### Versions

```Text
# This is from the computer I'm using currently, which isn't really relevant, since the version is only checked with "linux"
# - I get the error results when running on SLES linux, or Windows. The issue is specifically the string format failing to be parsed (which I did run on this computer)
Version:             2.6.6. # <- is this supposed to be different than the 2.8.2 that "pip freeze" shows?
API version:         0.8.3
Python version:      3.9.6
Git commit:          87767cda
Built:               Thu, Nov 3, 2022 1:15 PM
Profile:             darrida
Server type:         cloud
```

### Additional context

I cloned the Prefect repo at work, adjusted the line in question to the following, built a wheel, deployed a new agent to our test server using that wheel, and the problem with "20.10.17-ce" was resolved. Everything worked fine.

```python
user_version = packaging.version.parse(
    docker_client.version()["Version"].replace("-ce", "").replace("-ee", "")
)
```

darrida:
This PR potentially closes #8587

The current Docker version checker in `prefect.infrastructure.docker.py` doesn't account for a Docker version name format we use that seems to be an outlier. Our current version is "20.10.17-ce". This causes `packaging.version.parse` to through an `InvalidVersion` exception, rather than failing the check by being an insufficient version.

This PR suggests a few lines of code that resolves this issue. It simply removes "-ce" or "-ee" (community edition; enterprise edition) from the end of the version name string. I starting creating more complication regex type logic, but later realized the `packaging.version.parse` method does quite a bit of that itself, and I also failed to find examples of other outlier name formats that might also throw the same exception. The two simply find/replace statements may be all that is needed for any current case that could occur.

Prior to creating a change that was a little more formal (this PR), I simply put `.replace("-ce", "")` on the end of `docker_client.version()["Version"]`, built a local Prefect wheel, deployed an agent with it on a test server that uses the same "20.10.17-ce" name format, and confirmed it resolved the issue.

For more details, see the information I included in the issue (#8587).

Obviously open to changes -- whether it's simplifying the change, or if it's determined a more robust solution is needed (i.e., more inclusive of other hypothetical version name outliers).

### Example
- Current Behavior: Prefect is unable to launch a Docker container with our Docker version - "20.10.17-ce".
- Behavior after PR change: Prefect successfully launches a Docker container with the version name.

### Checklist


darrida:
I can also see that there might be a want for the extra format operations to not run every single time for Docker version names that pass the `parse` validation without issue. In that case, perhaps the the outlier logic could be moved into an exception handler, and only used an `InvalidVersion` exception is thrown.

madkinsz (approved):
Thanks!

PR #8600, opened 2023-02-20
Expose idempotency key in orion rest api

### LINKED ISSUE #8321: Expose idempotency_key in orion rest api
### First check

### Prefect Version

2.x

The only way to find by the Orios rest API if a flow_run with a given idempotency_key exists is to create following request:

```
import requests
request.post("xxx/api/flow_runs", json={"idempotency_key":"aabb", "flow_id": "12345678-abcd-abcd-abcd-123456781234"})
```

Nevertheless there are situations where we just want to know if there already was a flow_run with a given idempotency key without triggering new one.

Possible solutions:

1. Add idempotency_key to `https://docs.prefect.io/api/flow_runs/filter` for the `flow_runs`.
2. Add a new get method with possibility to filter by idempotency_key and flow_id.

### Example Use

_No response_

### Additional context

This issue came as discussion from https://github.com/PrefectHQ/prefect/issues/5534

jcorrado76:
Closes #8321

### The Change
This change exposes the idempotency key attribute of flow runs to be used for filter operations in the Orion REST API. 
Since the implementation to expose other attributes of flow run objects was already in place, this change consisted of replicating the code to expose other attributes, such as `FlowRunFilterId`. 

### Example
In addition to filtering flow runs based on the flow run ID:
```python
flow_run_filter = dict(
    flow_runs=schemas.filters.FlowRunFilter(
        id=schemas.filters.FlowRunFilterId(any_=[flow_runs[0].id])
    ).dict(json_compatible=True)
)
response = await client.post("/flow_runs/filter", json=flow_run_filter)
```
We can also filter on the idempotency key:
```python
flow_run_idempotency_key_filter = dict(
    flow_runs=schemas.filters.FlowRunFilter(
        idempotency_key=schemas.filters.FlowRunFilterIdempotencyKey(
            any_=[idempotency_key_of_flow_run_we_want_to_retrieve]
        )
    ).dict(json_compatible=True)
)
response = await client.post(
    "/flow_runs/filter", json=flow_run_idempotency_key_filter
)
```
### Checklist


jcorrado76:
Hi @madkinsz ! Thanks for approving my workflows. 
I see that a few of the workflows failed. 
The first one failed because of the time the test took to run changed:
```
FAILED tests/server/models/test_orm.py::TestTotalRunTimeEstimate::test_flow_run_estimated_run_time_includes_current_run - AssertionError: assert datetime.timedelta(seconds=60, microseconds=2152) < datetime.timedelta(seconds=60)
```

I was wondering what to do about this one? Do I increase the limit on the test that checks the runtime? Or do I have to update something else with the test? 

Then the integration tests with Prefect 2.6 and 2.7 failed, but all the other integration tests passed. 
This failed with a 422 Unprocessable Entity error:
```
prefect.exceptions.PrefectHTTPStatusError: Client error '422 Unprocessable Entity' for url 'http://127.0.0.1:4200/api/flow_runs/filter'
Response: {'exception_message': 'Invalid request received.', 'exception_detail': [{'loc': ['body', 'flow_runs', 'idempotency_key'], 'msg': 'extra fields not permitted', 'type': 'value_error.extra'}], 'request_body': {'flows': None, 'flow_runs': {'operator': 'and_', 'id': None, 'name': None, 'tags': None, 'deployment_id': None, 'work_queue_name': None, 'state': None, 'flow_version': None, 'start_time': None, 'expected_start_time': None, 'next_scheduled_start_time': None, 'parent_task_run_id': {'operator': 'and_', 'any_': ['ca54001e-5ab6-4c51-8538-889e63ea7baa'], 'is_null_': None}, 'idempotency_key': None}, 'task_runs': None, 'deployments': None, 'work_pools': None, 'work_pool_queues': None, 'sort': 'EXPECTED_START_TIME_ASC', 'limit': None, 'offset': 0}}
```
Is this one related to the `cloud-pr` tag you attached? So there's another PR that needs to be made in Prefect Cloud to make these tests pass? 

Would you be able to give me guidance as to how to move forward in resolving these issues? 
Thanks.

madkinsz:
`tests/server/models/test_orm.py::TestTotalRunTimeEstimate::test_flow_run_estimated_run_time_includes_current_run` looks like a rare flake that is unrelated too your change, we can tackle that separately. 

The compatibility tests failed because this client version is now always sending an `idempotency_key` filter (even if it is null) and prior server versions do not know about the field and consequently throw an error. The solution to this is likely to include the `exclude_unset` option when the client makes the request so that fields that are not explicitly set are not sent to the server.

jcorrado76:
Hi @madkinsz and @rpeden, apologies for being MIA. I was traveling for work the past few weeks. I'd like to work on the Cloud PR that needs to happen to get this over the finish line as well. I was wondering if you could point me in the right direction so I can get started? Thanks

madkinsz:
@jcorrado76 the Cloud repository is closed source, sorry. We can leverage most of the implementation here there though! The next step here is to fix the compatibility with old servers.

jcorrado76:
Hi @madkinsz @rpeden, apologies for the late response. 

Was this change as simple as adding `exclude_unset=True` in all the locations where the `FlowRunFilter` is used in client/orchestration.py?

I'm only getting an error on one of the usages of `FlowRunFilter` (in `read_flow_runs`), but since my change adds a new kwarg to the FlowRunFilter, and that FlowRunFilter is used in several of the methods (like also `read_deployments`), I would need to pass `exclude_unset=True` in all these locations, right? 

 Should `exclude_unset=True` just be a default everywhere, if it excludes response variables that haven't been set to something in the client request? 

For the backtesting on previous versions of Prefect, are the local Docker unit tests sufficient to know if they'll pass locally (for Prefect 2.6 and 2.7), (before I push code and trigger a remote CI/CD run), or do I need to go through the remote integration tests to get feedback on my changes? 

Thanks

madkinsz:
Yeah `exclude_unset=True` is generally the key to backward compatibility in these cases. I don't think we're generally doing it by default right now, I'm not sure what the implications of changing the default behavior would be. Perhaps in our `actions` schemas we should default to exclude unset values.

> Are the local Docker unit tests sufficient to know if they'll pass locally (for Prefect 2.6 and 2.7)

I don't think there's coverage across versions in the unit tests right now. You can run a server on a different version with Docker and try running one of the failed integration flows against it. You're also welcome to push the change here to test.

jcorrado76:
Hi, I added the `exclude_unset=True` argument everywhere `flow_run_filter.dict` is used in `orchestration.py`.

However, all the tests with `--exclude-services` are failing now. 

But now I'm getting tests that are failing on some sort of environment setup failure. (`test_pip.py` is failing; some tests are failing to raise an error when installing prefect in editable mode). 

I'm not sure what would be causing these issues, or how to fix them. Do you have any ideas as to what these are testing for, and why adding `exclude_unset=True` is causing them? 

I've added `exclude_unset=True` in `src/prefect/client/orchestration.py`, in `PrefectClient` in the methods:
* `read_flows`
* `read_deployments`
* `read_flow_runs`
* `read_task_runs`

Basically, I'm not sure what would even cause these environment-related errors to pop up. I know it has something to do with the `--exclude-services` flag, but I don't see why that'd cause the errors.. 

Any direction would be helpful! 
Thanks

madkinsz:
@jcorrado76 that `pip` test error is resolved on `main` —  it's caused by a change in the latest version of `pip`.

@jakekaplan can you please take over review and the implementation in Cloud here?

jakekaplan:
@madkinsz yes happy to.

@jcorrado76 I'll try and do a quick review later today and can do a cloud implementation tomorrow to pair this with.

jakekaplan:
cloud pr added here: https://github.com/PrefectHQ/nebula/pull/4414

@jcorrado76 code looks good to me. If you're able to merge main and tests pass happy to merge this

tests/server/api/test_flow_runs.py
```diff
@@ -356,6 +389,32 @@ async def test_read_flow_runs_applies_flow_run_filter(
...
+        self, flow_runs_with_idempotency_key, client
+    ):
+        idempotency_key_of_flow_run_we_want_to_retrieve = (
+            flow_runs_with_idempotency_key[0].idempotency_key
+        )
+        flow_run_idempotency_key_filter = dict(
+            flow_runs=schemas.filters.FlowRunFilter(
+                idempotency_key=schemas.filters.FlowRunFilterIdempotencyKey(
+                    any_=[idempotency_key_of_flow_run_we_want_to_retrieve]
+                )
```
jakekaplan, line 402: Could you add case for `not_any` test as well?

jakekaplan (approved):
LGTM! going to merge this.

The test failures seem to be related to a known failing test on main fixed here: https://github.com/PrefectHQ/prefect/pull/9274

PR #8612, opened 2023-02-22
Troubleshooting-Pytest-timout

sidharth1805:
This PR attempts to include troubleshooting procedures while running pytest locally

### Example
Currently, the timeout for the pytest is set to 60 seconds which seems to not work for some tests. There is no mention in the documentation to increase the timeout in the configuration file. Hence created this PR to add troubleshooting procedures while encountering the Pytest Timeout error.

### Checklist

### DIFF
--- a/docs/contributing/common-mistakes.md
+++ b/docs/contributing/common-mistakes.md
@@ -74,3 +74,37 @@ it.
 
 It may help to re-run the specific test with `PYTHONTRACEMALLOC=25 pytest ...` so that
 Python can display more of the stack trace where the connection was opened.
+
+## Pytest Timeout Error
+
+  
+
+**Summary:** Troubleshooting Pytest Timeout error.
+
+  
+
+If you are facing a timeout error when running pytest it is most likely that you need to change the timeout seconds at configuration file.
+
+You'll see a failure like:
+
+```bash
+pytest tests/test_context.py
+++++++++++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++++++++
+```
+  
+To resolve this, go to the config file
+  
+```
+ prefect/setup.cfg
+
+```
+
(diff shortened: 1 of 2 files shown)


sidharth1805:
### Below are the changes implemented

https://deploy-preview-8612--prefect-docs.netlify.app/contributing/common-mistakes/#pytest-timeout-error

image

https://deploy-preview-8612--prefect-docs.netlify.app/contributing/common-mistakes/#pytest-timeout-error

image

madkinsz:
Hey @sidharth1805 :) I appreciate that you took the time to improve the documentation. However, we do not recommend adjusting the timeout. If a test takes longer than 60s to run, it is not an acceptable test for our suite because it's too slow. Why did you need to tweak your timeout?

Also just so you know, you can bump the timeout per call to pytest with the `--timeout` option.

sidharth1805:
@madkinsz oh I see while running Pytest locally I faced this timeout issue and couldn't find it in the official documentation. Hence thought to add it here so people might find it helpful. I ran the pytest on the windows system. I have another question using --timeout can it override the timeout mentioned in cfg file?

sidharth1805:
Below is the Error

image

sidharth1805:
it seems the --timeout option overrides the cfg file. Do you want to add this to the docs or shall I close this PR.

madkinsz:
Since it needs to run with a timeout of 60s in CI it generally won't help to change it locally. Which test times out in that screenshot?

sidharth1805:
@madkinsz  test_logging.py

madkinsz:
This sounds like an issue with your development setup, maybe open an issue with the traceback and we can look into a fix? We don't officially support development on Windows though.

sidharth1805:
This sounds good thanks for looking into this using pytest tests/...py --timeout 300 seems to solve the issue for me.

PR #8614, opened 2023-02-22
Update `call_soon_in_main_thread` to `call_soon_in_supervising_thread`

madkinsz:
In practice, we do not actually care if it is the main thread we just want to be able to pass work to the supervisor.


chrisguidry (approved):
Excellent point!

PR #8615, opened 2023-02-22
Removes public-only indication in GitHub doc

EmilRex:
Changes the `GitHub` storage block docstring to indicate that repos can be both public or private.


billpalombi (approved):
lgtm - lets get this merged

Issue #8563, opened 2023-02-16
Flow Run end up in 'Crashed' state following job watch timeout.


efranksrecroom:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

We've seen several instances lately where a Flow Run ends in a 'Crashed' state even though the actual Flow completes successfully. I am able to verify the success by looking at the Job pod logs. What appears to be happening is that the Agent raises an error and then shortly afterward will log another error stating that it hit a timeout. In our Flow deployments, we have the job_watch_timeout_seconds set to 60s. In every recent instance where a Flow Run ended in 'Crashed' I seem something similar to the following in the Agent logs:

```
2023-02-16T03:51:22-08:00 11:51:22.627 | WARNING | prefect.infrastructure.kubernetes-job - Error occurred while streaming logs - Job will continue to run but logs will no longer be streamed to stdout.
2023-02-16T03:51:22-08:00 Traceback (most recent call last):
2023-02-16T03:51:22-08:00   File "/usr/local/lib/python3.10/site-packages/urllib3/response.py", line 761, in _update_chunk_length
2023-02-16T03:51:22-08:00     self.chunk_left = int(line, 16)
2023-02-16T03:51:22-08:00 ValueError: invalid literal for int() with base 16: b''
2023-02-16T03:51:22-08:00 
2023-02-16T03:51:22-08:00 During handling of the above exception, another exception occurred:
2023-02-16T03:51:22-08:00 
2023-02-16T03:51:22-08:00 Traceback (most recent call last):
2023-02-16T03:51:22-08:00   File "/usr/local/lib/python3.10/site-packages/urllib3/response.py", line 444, in _error_catcher
2023-02-16T03:51:22-08:00     yield
2023-02-16T03:51:22-08:00   File "/usr/local/lib/python3.10/site-packages/urllib3/response.py", line 828, in read_chunked
2023-02-16T03:51:22-08:00     self._update_chunk_length()
2023-02-16T03:51:22-08:00   File "/usr/local/lib/python3.10/site-packages/urllib3/response.py", line 765, in _update_chunk_length
2023-02-16T03:51:22-08:00     raise InvalidChunkLength(self, line)
[...]

PR #8618, opened 2023-02-22
Fix Kubernetes job watch timeout when streaming logs

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/8563

See discussion in issue for details. The Kubernetes job timeout was only being started _after_ streaming logs which meant that timeout was not applying to the full runtime of the job. Now if the timeout is exceeded while streaming logs, we will stop streaming logs and exit. We will also correctly enforce the timeout if we stream logs for some period of time then begin watching for job completion.

### DIFF
--- a/src/prefect/infrastructure/kubernetes.py
+++ b/src/prefect/infrastructure/kubernetes.py
@@ -595,6 +595,8 @@ class KubernetesJob(Infrastructure):
 
         Return the final status code of the first container.
         """
+        self.logger.debug(f"Job {job_name!r}: Monitoring job...")
+
         job = self._get_job(job_name)
         if not job:
             return -1
@@ -602,6 +604,13 @@ class KubernetesJob(Infrastructure):
         pod = self._get_job_pod(job_name)
         if not pod:
             return -1
+
+        # Calculate the deadline before streaming output
+        deadline = (
+            (time.time() + self.job_watch_timeout_seconds)
+            if self.job_watch_timeout_seconds is not None
+            else None
+        )
 
         if self.stream_output:
             with self.get_client() as client:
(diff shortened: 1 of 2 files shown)


src/prefect/infrastructure/kubernetes.py
```diff
@@ -625,24 +641,23 @@ def _watch_job(self, job_name: str) -> int:
...
-        completed = False
-        while not completed:
-            remaining_time = deadline - time.time() if deadline else None
-            if deadline and remaining_time <= 0:
-                self.logger.error(
-                    f"Job {job_name!r}: Job did not complete within "
-                    f"timeout of {self.job_watch_timeout_seconds}s."
-                )
-                return -1
+        with self.get_batch_client() as batch_client:
```
madkinsz, line 644: This section perhaps best read without whitespace changes https://github.com/PrefectHQ/prefect/pull/8618/files?diff=unified&w=1

src/prefect/infrastructure/kubernetes.py
```diff
@@ -625,24 +641,23 @@ def _watch_job(self, job_name: str) -> int:
...
+            # Check if the job is completed before beginning a watch
+            job = batch_client.read_namespaced_job(
+                name=job_name, namespace=self.namespace
+            )
+            completed = job.status.completion_time is not None
```
madkinsz, line 649: Here we check the status _before_ starting the watch so that if a deadline is exceeded during log streaming but the job completes then we do not fail the job as timed out.

zzstoatzz (approved):

rpeden (approved):
Looks good, makes sense, and worked when I QA'ed it.

PR #8631, opened 2023-02-23
Refactor supervisors to manage submission

madkinsz:
This allows the supervisor to track the thread the supervised future is executing on, which is essential for cancellation handling. This also has the benefit of clarifying the supervisor interface and reducing the responsibilities of users of the supervisor.

### DIFF
--- a/src/prefect/_internal/concurrency/from_async.py
+++ b/src/prefect/_internal/concurrency/from_async.py
@@ -4,11 +4,7 @@ from typing import Callable, TypeVar
 from typing_extensions import ParamSpec
 
 from prefect._internal.concurrency.runtime import get_runtime_thread
-from prefect._internal.concurrency.supervisors import (
-    AsyncSupervisor,
-    get_supervisor,
-    set_supervisor,
-)
+from prefect._internal.concurrency.supervisors import AsyncSupervisor, get_supervisor
 
 P = ParamSpec("P")
 T = TypeVar("T")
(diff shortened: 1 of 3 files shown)


src/prefect/_internal/concurrency/from_async.py
```diff
@@ -45,10 +43,8 @@ def call_soon_in_worker_thread(
     Returns a supervisor.
     """
     runtime = get_runtime_thread()
-    supervisor = AsyncSupervisor()
-    with set_supervisor(supervisor):
-        future = runtime.submit_to_worker_thread(__fn, *args, **kwargs)
-    supervisor.set_future(future)
+    supervisor = AsyncSupervisor(runtime.submit_to_worker_thread)
+    supervisor.submit(__fn, *args, **kwargs)
```
madkinsz, line 47: This diff shows how much easier it is to use the supervisor correctly

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -144,16 +145,38 @@ class Supervisor(abc.ABC, Generic[T]):
...
-        self._future = future
+        if self._future:
+            raise RuntimeError("A supervisor can only monitor a single future")
+
+        @functools.wraps(__fn)
+        def _call_in_supervised_thread():
+            # Capture the worker thread before running the function
+            thread = threading.current_thread()
+            self._future_thread.set_result(thread)
+            return __fn(*args, **kwargs)
```
madkinsz, line 172: Here, we add tracking of the thread executing the future. We use a future to store this data so the supervisor can wait for it to be populated if it needs to.

src/prefect/_internal/concurrency/from_async.py
chrisguidry: Very nice!

src/prefect/_internal/concurrency/supervisors.py
chrisguidry: Cool trick :D

chrisguidry (approved):

PR #8641, opened 2023-02-23
Handle `PREFECT_API_KEY` set as environment variable in `cloud login` command

### LINKED ISSUE #7797: Having PREFECT_API_KEY as env var prevent from login to cloud
### First check

### Bug summary

When having `PREFECT_API_KEY` set as an environment variable, the `prefect cloud login` command will exit with `This profile is already authenticated with that key.` without being actually logged in.

### Reproduction

```python3
PREFECT_API_KEY=foo prefect cloud login -k foo
```

### Error

```python3
We are not actually logged in.
```

### Versions

```Text
Version:             2.7.0
API version:         0.8.3
Python version:      3.10.5
Git commit:          c2833339
Built:               Thu, Dec 1, 2022 4:03 PM
Profile:             default
Server:
  Database:          sqlite
  SQLite version:    3.38.5
```

### Additional context

_No response_

serinamarie:
### Overview
Closes #7797 and https://github.com/PrefectHQ/prefect/issues/10510

Previously one would **exit with false success** when running this command:
```bash
PREFECT_API_KEY=foo prefect cloud login -k foo
```
despite **not** being logged in. Just because we have a `PREFECT_API_KEY` env var set and it's equal to the `key` passed in to `prefect cloud login -k pnu_xyz`, and we are authenticated does not mean that we are using that particular profile, which I think is the user goal with the `login` command.

With this PR, we are directed to fix our issues (or what will become an issue down the line) rather than exiting with success.

Additionally, we're checking that the key is valid **before** exiting with success, which should close the issue of expired `PREFECT_API_KEY`s exiting with success.

## Examples
A key in an unexpected format, hinting to users that something could be very off about what they pasted as the key:
```bash
❯ prefect cloud login -k bar
Unable to authenticate with Prefect Cloud. Your key is not in our expected format ('pnu').
```
```bash
❯ PREFECT_API_KEY=foo prefect cloud login -k foo
Unable to authenticate with Prefect Cloud. Your key is not in our expected format ('pnu').
```

A key in the correct format, but invalid:
```bash
❯ prefect cloud login -k pnu_invalid_key
Unable to authenticate with Prefect Cloud. Please ensure your credentials are correct.
```

A `PREFECT_API_KEY` env var is set but it is different from the user-provided key:
```bash
❯ PREFECT_API_KEY=foo prefect cloud login -k bar
Cannot log in with a key when a different PREFECT_API_KEY is present as an environment variable 
that will override it.
```
```bash
❯ PREFECT_API_KEY=pnu_invalid1 prefect cloud login -k pnu_invalid2
Cannot log in with a key when a different PREFECT_API_KEY is present as an environment variable 
that will override it.
```
A `PREFECT_API_KEY` env var is set and it is the same as a user-provided key but the key is invalid:
```bash
❯ PREFECT_API_KEY=pnu_invalid prefect cloud login -k pnu_invalid
Unable to authenticate with Prefect Cloud. Please ensure your credentials are correct.
```
A `PREFECT_API_KEY` env var is set and it is the same as a user-provided key and the key is valid:
```bash
❯ PREFECT_API_KEY=pnu_xyz prefect cloud login -k pnu_xyz
It looks like you're already authenticated with another profile.
? Would you like to reauthenticate with this profile? [y/N]: y
? Which workspace would you like to use? [Use arrows to move; enter to select]                                  
  my-workspace1//my-tests                                      
  my-workspace1/my-pond                                                                                      
> my-workspace2/my-account                                                                                           
Authenticated with Prefect Cloud! Using workspace 'my-workspace2/my-account'.
```

### Checklist


madkinsz (changes requested):
We shouldn't respect the current profile over the environment variable, I think. That breaks our standard behavior for settings. I think what we want to do here is check `check_key_is_valid_for_login` before exiting with an "already logged in" message.

serinamarie:
> We shouldn't respect the current profile over the environment variable, I think. That breaks our standard behavior for settings. I think what we want to do here is check `check_key_is_valid_for_login` before exiting with an "already logged in" message.

I believe you mean this?
```python
if key and PREFECT_API_KEY.value() == key:
  if await check_key_is_valid_for_login(key):
    exit_with_success("This profile is already authenticated with that key.")
```
I'm not sure I understand how that could work. The key may be valid for login but that doesn't mean that it's already logged into that profile.

For instance, with the above logic, if I am logged into `profile-1`, and I run
```bash
PREFECT_API_KEY=<api-key-for-profile-2> prefect cloud login -k <api-key-for-profile-2>
``` 
it will wrongly exit with 
```bash
This profile is already authenticated with that key.
```

when I am still logged into `profile-1`.

madkinsz:
Ah I see, confusing. What does it mean to do a `prefect cloud login` when you have an environment variable set? We can log the profile in but the key in the environment is going to override your profile so you won't actually be logged in. Perhaps if we see an environment variable API key we should just exit saying that we can't login because you've set an environment var?

serinamarie:
> Ah I see, confusing. What does it mean to do a `prefect cloud login` when you have an environment variable set? We can log the profile in but the key in the environment is going to override your profile so you won't actually be logged in. Perhaps if we see an environment variable API key we should just exit saying that we can't login because you've set an environment var?

I think it's probably only an issue if the active profile has a different API key than the env var. If it does differ, we could exit with error. If they're the same, we could tell them they're already authenticated.

madkinsz:
I'm not sure it's worth complicating it seems like a weird case to have the same key set in both locations. I guess we can check if it works and maybe allow it to be copied into the profile with a prompt? Then they'd be more likely to be in a situation where it matches. It seems like more work than it's worth though and it may be best to simply check if it works and exit with a helpful message.

madkinsz:
@serinamarie let's get this merged or closed

src/prefect/cli/cloud.py
```diff
@@ -319,10 +319,17 @@ async def login(
 
     profiles = load_profiles()
     current_profile = get_settings_context().profile
-
-    if key and PREFECT_API_KEY.value() == key:
+    current_profile_api_key = current_profile.settings.get(PREFECT_API_KEY)
```
madkinsz, line 322: This will bypass callbacks, you should use `settings.value_of(SETTING)` or `SETTING.value_from(settings)

serinamarie:
Closing for now until I or someone gets a chance to work on it

src/prefect/cli/cloud/__init__.py
```diff
@@ -318,9 +318,24 @@ async def login(
 
     profiles = load_profiles()
     current_profile = get_settings_context().profile
+    env_var_api_key = PREFECT_API_KEY.value()
 
-    if key and PREFECT_API_KEY.value() == key:
```
serinamarie, line 322: Just because we have a `PREFECT_API_KEY` env var set and it's equal to the `key` passed in to `prefect cloud login -k pnu_xyz`, and we are authenticated does not mean that we are using that particular profile, which I think is the prime of the `login` command.

serinamarie:
> It seems like this UX should be:
> 
> If there is a key in the environment that does not match the profile, say we cannot login if a key is present as an environment variable.
> If the key matches the profile or a current key is found, check if the key works before exiting with an already authenticated message.

I believe my changes address Zanie's comments.

zangell44 (approved):
🚀

PR #8645, opened 2023-02-23
Add ArtifactCollection table for query optimization of artifact history

serinamarie:
### Update March 7 2023
https://github.com/PrefectHQ/nebula/pull/3822

### Overview
Cloud implementation: https://github.com/PrefectHQ/nebula/pull/3822
1. Adds a new table, `artifact_collection`, with 2 columns, `key` and `latest_id` (a column name up for debate). The `latest_id` column stores the most recently created version of the artifact.

   For example, if you have an `artifact` table like so:
   #### artifact
   | id        | key         | created  | updated | 
   | ------------- |:-------------:| -----:| ----:|
   | 1      | voltaic | 01-01-23 | 01-31-23
   | 2      | voltaic      | 01-28-23 | 01-28-23
   | 3 |      |   01-05-23 | 01-05-23
   | 4 | zebra | 01-10-23 | 01-10-23

   Your `artifact_collection` table will store the the latest version of each artifact with a `key` (optional):
   #### artifact_collection
   | latest_id        | key         |
   | ------------- |:-------------:|  
   | 2      | voltaic      | 01-28-23 |
   | 4 | zebra | 01-10-23

   This is useful because your artifact page will display the latest version of each artifact (it's your artifact collection!). When you drill into a particular artifact, the `artifact` table lends its data to a history-centered view, providing associated flow/task runs of each artifact version, when they were updated, associated data and metadata.

   Alongside the new `artifact_collection` table comes updated `create_artifact` and `delete_artifact` model methods. 
   - In `create_artifact`, we'll first upsert to the `artifact_collection` table, ensuring we keep track of the latest artifact version id.
   - In `delete_artifact`, we now need to check if the artifact is the latest version. If it is, we'll need to update our reference to it in the `artifact_collection` table prior to deletion from the `artifact` table. If there are no prior versions of the artifact (there's just 1 version left), then the row itself related to the artifact will be deleted from the `artifact_collection` table.

2. Since we will now allow artifacts to have the same key, we remove the `artifact` table's `key` column unique constraint. Instead we index on `key` in the `artifact` table.  Additionally, we only allow updates to artifact metadata rather than data as well.

3.  Tests around the updated methods have been added.
4. Migrations added for removing the unique constraint from the `artifact` table, adding the new `artifact_collection` table, and adding the necessary indexes.

### Notes
A separate PR will add an index for querying artifacts with a given key, as well as a corresponding API route. This PR is in lockstep with its Cloud counterpart.

### Checklist

### DIFF
--- a/src/prefect/server/models/artifacts.py
+++ b/src/prefect/server/models/artifacts.py
@@ -298,12 +298,12 @@ async def delete_artifact(
         ).scalar_one_or_none()
 
         if next_latest_version is not None:
-            update_latest_version = (
+            set_next_latest_version = (
                 sa.update(db.ArtifactCollection)
                 .where(db.ArtifactCollection.key == artifact.key)
                 .values(latest_id=next_latest_version.id)
             )
-            await session.execute(update_latest_version)
+            await session.execute(set_next_latest_version)
 
         else:
             await session.execute(


src/prefect/server/models/artifacts.py
```diff
@@ -172,6 +184,44 @@ async def delete_artifact(
...
+
+    is_latest_artifact_version = (
+        await session.execute(
+            sa.select(db.ArtifactCollection)
+            .where(db.ArtifactCollection.key == current_artifact.key)
+            .where(db.ArtifactCollection.latest_id == artifact_id)
+        )
+    ).scalar_one_or_none()
+
+    if is_latest_artifact_version is not None:
```
madkinsz, line 199: If this is `is_latest_...` maybe we should do the `is not None` check above so this is a bool? Then this if statement reads clearer.

src/prefect/server/models/artifacts.py
```diff
@@ -172,6 +184,44 @@ async def delete_artifact(
     Returns:
         bool: True if the delete was successful, False otherwise
     """
+    current_artifact = await session.get(db.Artifact, artifact_id)
```
madkinsz, line 187: This might not be the latest artifact though? "current" seems like a weird prefix here. Should this just be `artifact`?

src/prefect/server/models/artifacts.py
```diff
@@ -172,6 +184,44 @@ async def delete_artifact(
...
+    is_latest_artifact_version = (
+        await session.execute(
+            sa.select(db.ArtifactCollection)
+            .where(db.ArtifactCollection.key == current_artifact.key)
+            .where(db.ArtifactCollection.latest_id == artifact_id)
+        )
+    ).scalar_one_or_none()
+
+    if is_latest_artifact_version is not None:
+        previous_artifact = (
```
madkinsz, line 200: Maybe this should be `next_latest_artifact_version` or similar?

src/prefect/server/models/artifacts.py
madkinsz: I found the logic here a little challenging to follow, I think mostly because of the variable names so I have a few suggestions here. A comment describing that we update the latest artifact to point to the next latest version if it exists would be helpful too.

src/prefect/server/schemas/actions.py
```diff
@@ -548,17 +548,41 @@ class FlowRunNotificationPolicyUpdate(ActionBaseModel):
 class ArtifactCreate(ActionBaseModel):
     """Data used by the Prefect REST API to create an artifact."""
 
-    key: Optional[str] = FieldFrom(schemas.core.Artifact)
-    type: Optional[str] = FieldFrom(schemas.core.Artifact)
-    data: Optional[Union[Dict[str, Any], Any]] = FieldFrom(schemas.core.Artifact)
-    metadata_: Optional[Dict[str, str]] = FieldFrom(schemas.core.Artifact)
-    flow_run_id: Optional[UUID] = FieldFrom(schemas.core.Artifact)
-    task_run_id: Optional[UUID] = FieldFrom(schemas.core.Artifact)
```
madkinsz, line 556: Why do none of these use `FieldFrom` anymore?

serinamarie: Good catch, this file should be left unchanged except for removing the `data` field in `ArtifactUpdate`

src/prefect/server/models/artifacts.py
serinamarie: That sounds better for sure

src/prefect/server/models/artifacts.py
serinamarie: Yes, that sounds clearer 👍

serinamarie:
Closing in favor of smaller PRs
https://github.com/PrefectHQ/prefect/pull/8860
https://github.com/PrefectHQ/prefect/pull/8859
https://github.com/PrefectHQ/prefect/pull/8861
https://github.com/PrefectHQ/prefect/pull/8855

PR #8649, opened 2023-02-24
Add timeout support to supervisors

madkinsz:
Adds timeout handling to supervisors.

Supervisors can be created with a `timeout`. The timeout will start when the future submitted to the supervisor begins running. The timeout will be enforced for both synchronous and asynchronous work that the future represents. The timeout will also be enforced on work sent _back_ to the supervising thread by the supervised call, ensuring that the supervising thread never waits for a result longer than the timeout.

Timeouts are implemented in three ways:
- Asynchronous cancel scopes (async only)
- Alarm signal (main thread only — can notably interrupt sys calls)
- Watcher threads

### Example

```python
import concurrent.futures
import time

from prefect._internal.concurrency.supervisors import SyncSupervisor


def slow_worker():
    for _ in range(100):
        time.sleep(0.1)


executor = concurrent.futures.ThreadPoolExecutor()

supervisor = SyncSupervisor(submit_fn=executor.submit, timeout=0.1)
supervisor.submit(slow_worker)
supervisor.result()
```

In this example, you'll see two cancel contexts get created. One, alarm based is for the main thread which is watching for work returned by the future. This prevents the future from blocking the main thread by sending work back. The other, watcher based, is for the future executing in the worker thread.

```
13:52:42.257 | DEBUG   | prefect._internal.concurrency.supervisors - Created supervisor <SyncSupervisor submit_fn='submit', owner='MainThread'>
13:52:42.259 | DEBUG   | prefect._internal.concurrency.supervisors - Call to 'slow_worker' submitted to supervisor <SyncSupervisor submit_fn='submit', owner='MainThread'> tracked by future <Future at 0x7f8281e5d5b0 state=running>
13:52:42.259 | DEBUG   | prefect._internal.concurrency.timeouts - Entered synchronous cancel context with 0.1s alarm based timeout
13:52:42.260 | DEBUG   | prefect._internal.concurrency.supervisors - Watching for work sent to supervisor <SyncSupervisor submit_fn='submit', submitted='slow_worker', owner='MainThread'>
13:52:42.260 | DEBUG   | prefect._internal.concurrency.timeouts - Entered synchronous cancel context with 0.1s watcher based timeout
13:52:42.364 | DEBUG   | prefect._internal.concurrency.timeouts - Cancel fired for watched based timeout of thread 'ThreadPoolExecutor-0_0'
13:52:42.471 | DEBUG   | prefect._internal.concurrency.supervisors - Supervisor <SyncSupervisor submit_fn='submit', submitted='slow_worker', owner='MainThread'> retrieving result of future <Future at 0x7f8281e5d5b0 state=finished raised TimeoutError>
  File "example.py", line 15, in <module>
    supervisor.result()
  File "/Users/mz/dev/prefect/src/prefect/_internal/concurrency/supervisors.py", line 314, in result
    return self._future.result()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/mz/dev/prefect/src/prefect/_internal/concurrency/supervisors.py", line 228, in _call_in_supervised_thread
    retval = fn(*args, **kwargs)
  File "example.py", line 8, in slow_worker
    time.sleep(0.1)
TimeoutError
```
### Checklist

### DIFF
--- a/src/prefect/_internal/concurrency/supervisors.py
+++ b/src/prefect/_internal/concurrency/supervisors.py
@@ -265,8 +265,11 @@ class SyncSupervisor(Supervisor[T]):
             cancel_sync_after(timeout)
             if timeout is not None
             else contextlib.nullcontext()
-        ):
+        ) as ctx:
             self._watch_for_work_items()
+
+        if timeout is not None and ctx.cancelled:
+            raise TimeoutError()
 
         logger.debug("Retrieving result for %r", self)
         return self._future.result(
@@ -320,8 +323,11 @@ class AsyncSupervisor(Supervisor[T]):
             cancel_async_after(timeout)
             if timeout is not None
             else contextlib.nullcontext()
-        ):
+        ) as ctx:
             await self._watch_for_work_items()
+
+        if timeout is not None and ctx.cancelled:
+            raise TimeoutError()
 
         logger.debug("Retrieving result for %r", self)
         return await self._future
--- a/src/prefect/_internal/concurrency/supervisors.py
+++ b/src/prefect/_internal/concurrency/supervisors.py
@@ -261,14 +261,10 @@ class SyncSupervisor(Supervisor[T]):
         # deadline and enforce the timeout when retrieving the result as well
         deadline = (time.time() + timeout) if timeout is not None else None
 
-        with (
-            cancel_sync_after(timeout)
-            if timeout is not None
-            else contextlib.nullcontext()
-        ) as ctx:
+        with cancel_sync_after(timeout) as ctx:
             self._watch_for_work_items()
 
-        if timeout is not None and ctx.cancelled:
+        if ctx.cancelled:
             raise TimeoutError()
 
         logger.debug("Retrieving result for %r", self)
@@ -319,14 +315,10 @@ class AsyncSupervisor(Supervisor[T]):
         if not self._future:
             raise ValueError("No future being supervised.")
 
-        with (
-            cancel_async_after(timeout)
-            if timeout is not None
-            else contextlib.nullcontext()
-        ) as ctx:
+        with cancel_async_after(timeout) as ctx:
             await self._watch_for_work_items()
 
-        if timeout is not None and ctx.cancelled:
+        if ctx.cancelled:
             raise TimeoutError()
 
         logger.debug("Retrieving result for %r", self)
--- a/src/prefect/_internal/concurrency/supervisors.py
+++ b/src/prefect/_internal/concurrency/supervisors.py
@@ -13,7 +13,6 @@ import functools
 import inspect
 import queue
 import threading
-import time
 from typing import (
     Any,
     Awaitable,
@@ -27,10 +26,17 @@ from typing import (
 )
 
 from prefect._internal.concurrency.event_loop import call_soon_in_loop, get_running_loop
-from prefect._internal.concurrency.timeouts import cancel_async_after, cancel_sync_after
+from prefect._internal.concurrency.timeouts import (
+    cancel_async_after,
+    cancel_async_at,
+    cancel_sync_after,
+    cancel_sync_at,
+    get_deadline,
+)
 from prefect.logging import get_logger
 
 T = TypeVar("T")
+Fn = TypeVar("Fn", bound=Callable)
 
 # Python uses duck typing for futures; asyncio/threaded futures do not share a base
 AnyFuture = Any
@@ -111,12 +117,12 @@ class WorkItem:
             result = self.context.run(self.fn, *self.args, **self.kwargs)
         except BaseException as exc:
             self.future.set_exception(exc)
+            logger.debug("Encountered exception in work item %r", self)
             # Prevent reference cycle in `exc`
-            self = None
+            del self
         else:
             self.future.set_result(result)
-
-        logger.debug("Finished work item %r", self)
+            logger.debug("Finished work item %r", self)
 
     async def _run_async(self):
         loop = asyncio.get_running_loop()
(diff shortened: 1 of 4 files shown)


tests/_internal/concurrency/test_timeouts.py
```diff
@@ -0,0 +1,46 @@
+import asyncio
+import concurrent.futures
+import time
+
+import pytest
+
+from prefect._internal.concurrency.timeouts import cancel_async_after, cancel_sync_after
+
```
madkinsz, line 8: Need to add to these tests here since I have since extended the timeout module

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -145,12 +153,21 @@ class Supervisor(abc.ABC, Generic[T]):
     Work sent to the supervisor will be executed when the owner waits for the result.
     """
 
-    def __init__(self, submit_fn: Callable[..., concurrent.futures.Future]) -> None:
+    def __init__(
+        self,
+        submit_fn: Callable[..., concurrent.futures.Future],
+        timeout: Optional[float] = None,
```
madkinsz, line 159: I'm going to need to find a clean way to expose this in the `from_sync` and `from_async` interfaces.

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -109,12 +117,12 @@ def _run_sync(self):
             result = self.context.run(self.fn, *self.args, **self.kwargs)
         except BaseException as exc:
             self.future.set_exception(exc)
+            logger.debug("Encountered exception in work item %r", self)
             # Prevent reference cycle in `exc`
-            self = None
+            del self
```
chrisguidry, line 122: I have never seen this done before :exploding_head:  Pretty wild, but I _think_ it makes sense in this context.  What was the problem you're anticipating here, the `WorkItem` never getting cleaned up and thus never freeing up the `future` it holds onto?

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -0,0 +1,249 @@
...
+import sys
+import threading
+import time
+from typing import Optional, Type
+
+import anyio
+
+from prefect.logging import get_logger
+
+# TODO: We should update the format for this logger to include the current thread
```
chrisguidry, line 14: :+1:

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Yields a `CancelContext`.
+    """
+    ctx = CancelContext(timeout=timeout)
+    if timeout is None:
+        yield ctx
+        return
+
+    if sys.platform.startswith("win"):
+        # Timeouts cannot be enforced on Windows
```
chrisguidry, line 150: Maybe a `logger.warning` to that effect here?

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -0,0 +1,249 @@
...
+def get_deadline(timeout: Optional[float]):
+    """
+    Compute an deadline given a timeout.
+
+    Uses a monotonic clock.
+    """
+    if timeout is None:
+        return None
+
+    return time.monotonic() + timeout
```
chrisguidry, line 79: :nail_care:

chrisguidry (approved):
Looks awesome!  I started with the test suite and read upwards, that was a nice way to do it before diving into the guts.  Great tests

src/prefect/_internal/concurrency/supervisors.py
madkinsz: This is common in the standard library implementations where reference cycles are a big deal. I think basically the stack frame attached to the exception will have a reference to the work item creating a reference cycle. I've switched from `= None` to `del` because I think it's more readable.

src/prefect/_internal/concurrency/timeouts.py
madkinsz: Seems wise in the meantime, I'm guessing we can have some sort of support eventually.

src/prefect/_internal/concurrency/supervisors.py
madkinsz: e.g. https://github.com/python/cpython/blob/main/Lib/concurrent/futures/_base.py#L404

madkinsz: or, even more relevant, https://github.com/python/cpython/blob/2db23d10bf64bf7c061fd95c6a8079ddc5c9aa4b/Lib/concurrent/futures/thread.py#L59-L62

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @property
+    def cancelled(self):
+        with self._lock:
+            return self._cancelled
+
+    @cancelled.setter
+    def cancelled(self, value: bool):
+        with self._lock:
+            self._cancelled = value
```
madkinsz, line 42: @chrisguidry do you know if this is necessary? I feel like you actually don't need a lock for threadsafe get/set of a single value and it'd only be needed if there was more going on in the `with`.

chrisguidry: I was puzzling over that but I thought you knew something I didn't HAHA.  I don't think you need a lock around either one of those.  It's definitely possible that they are multiple instructions under the hood, but that doesn't really matter, because the caller will either see one or the other value when they call it.  The real thing you would need a lock around is if you were trying to make a decision _with_ the value of the property:

```
with lock:
  if not thing.cancelled:
    thing.cancel()
```

madkinsz: Makes sense; removed this.

PR #8656, opened 2023-02-25
Override analytics block instead of partial

jawnsy:
We previously overrode the Analytics partial shipped with mkdocs, which configures Google Analytics. However, the documentation suggests instead overriding a block as done here, and this pattern means that we preserve the default behavior of mkdocs (which differs slightly from the official Google snippet).

From the [Material for Mkdocs documentation](https://squidfunk.github.io/mkdocs-material/customization/#overriding-blocks):

>Besides overriding partials, it's also possible to override (and extend) template blocks, which are defined inside the templates and wrap specific features

This allows us to preserve the [default mkdocs-material behavior](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/integrations/analytics/google.html) and also means we can use features like the "give feedback" option

### Checklist


madkinsz (approved):
Can you update the PR body to describe what the purpose of the change is?

PR #8668, opened 2023-02-27
Refactor supervisor API to allow configuration

madkinsz:
Removes the `from_sync` and `from_async` modules 😱 and moves methods for submitting calls to the supervisor class. This means instead of `supervisor = call_soon_in_thread(__fn, ...)` we do `new_supervisor().call_soon_in_thread(__fn, ...)` which allows the supervisor to be configured in the `new_supervisor` call which is essential for exposing things like timeouts.

This is a little bit of a mix of concerns, because now the supervisor requires knowledge of the runtime to provide a simple API for submission to the runtime. Maybe we should push that elsewhere.

### Checklist


src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -139,26 +176,29 @@ async def _run_async(self):
...
     """
 
     def __init__(
         self,
-        submit_fn: Callable[..., concurrent.futures.Future],
+        allow_callbacks: bool = True,
         timeout: Optional[float] = None,
     ) -> None:
-        self._submit_fn = submit_fn
+        self._callbacks_allowed = allow_callbacks
```
madkinsz, line 201: Not yet used.

Issue #8672, opened 2023-02-28
Instrument blocks methods with events


bunchesofdonald:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Prefect blocks currently do not emit any events.

### Describe the proposed behavior

Blocks should emit events whenever their public methods are called. The events should include the name of the block, the method that was called and whether or not it was successful (ie an exception was not raised).

PR #8673, opened 2023-02-28
Create `EventsWorker` to manage client lifecycle and abstract async nature

bunchesofdonald:
In a previous pull request we added an `events` method to `PrefectClient` that managed the events client lifecycle. However this proved to be difficult to use for our _current_ primary purpose of block instrumentation for a couple reasons:

1. The events client itself is purely async, but many block methods that we want to instrument are sync.
2. Ensuring `PrefectClient` was injected into instrumentation decorators without also overriding any client injection done by the methods themselves proved to be difficult.

This creates an `EventsWorker` which:

1. Spins up a thread.
2. Initializes a client with the given kwargs.
3. Ensures that the client's context is properly managed.
4. Watches for `Event` objects in a queue and emits them via the client as they are added.
5. Provides an `emit` method of its own which adds items to the queue.

Related to #8672 

### Example
```python
import uuid
from prefect import flow
from prefect.events import Event
from prefect.events.worker import get_events_worker


@flow
def emitter():
    with get_events_worker() as events:
        event = Event(
            event="vogon.poetry.read",
            resource={"prefect.resource.id": f"poem.{uuid.uuid4()}"},
        )

        events.emit(event)
```

### Checklist


src/prefect/events/worker.py
```diff
@@ -0,0 +1,128 @@
...
+    async def _main_loop(self):
+        """Orchestrate the events client and emit any queued events."""
+        self._client = self._client_type(**self._client_kwargs)
+        async with self._client:
+            self._ready_event.set()
+            while not self._stop_event.is_set():
+                try:
+                    event = self._queue.get_nowait()
+                except queue.Empty:
+                    await asyncio.sleep(0.1)
```
madkinsz, line 51: This is not an performant way to poll. You should probably use an `asyncio.Queue` so you can `await` a `get` call properly then you can put items in the queue like https://github.com/PrefectHQ/prefect/blob/main/src/prefect/_internal/concurrency/supervisors.py#L330-L331

bunchesofdonald: Awesome, thanks Zanie! I'll refactor that bit.

src/prefect/events/worker.py
```diff
@@ -0,0 +1,130 @@
...
+        self._ready_event = threading.Event()
+        self._lock = threading.Lock()
+
+        atexit.register(self.stop)
+
+    def _start_loop(self):
+        """Entrypoint for the thread."""
+        loop = asyncio.new_event_loop()
+        loop.run_until_complete(self._main_loop())
+        loop.close()
```
madkinsz, line 40: Managing an event loop this way is not recommend as it will not shutdown async generators correctly. You should just use `asyncio.run`.

src/prefect/events/worker.py
```diff
@@ -0,0 +1,130 @@
...
+
+    async def _main_loop(self):
+        """Orchestrate the events client and emit any queued events."""
+        self._client = self._client_type(**self._client_kwargs)
+        async with self._client:
+            self._ready_event.set()
```
madkinsz, line 46: You'll want to use a future for this, so you can raise an exception if there's a failure starting the worker thread. I would recommend reviewing the implementation in https://github.com/PrefectHQ/prefect/blob/main/src/prefect/_internal/concurrency/runtime.py#L44-L74

src/prefect/events/worker.py
bunchesofdonald: Actually I explicitly don't want to use an async queue, the _main_ point of this whole thing is to ensure that we can communicate with the events stuff synchronously.  Is there a better way to poll a `SimpleQueue` ?

chrisguidry: I don't think you need to use a worker thread to `put_nowait` into an `asyncio.Queue`, it's a sync function:

```
$ python
Python 3.10.6 (main, Dec 15 2022, 11:19:56) [GCC 11.3.0] on linux
>>> import asyncio
>>> q = asyncio.Queue()
>>> q.put_nowait('hi')
>>> q
<Queue at 0x7f72e95d1cc0 maxsize=0 _queue=['hi'] tasks=1>
```

chrisguidry: Without using a `get_nowait` and a `sleep`, how would it be possible to stop this thread?

tests/events/test_events_worker.py
```diff
@@ -0,0 +1,160 @@
...
+    worker.stop()
+    assert not hasattr(worker._client, "_in_context")
+    assert not worker._thread.is_alive()
+    assert not worker._ready_event.is_set()
+
+
+def test_start_is_idempontent(worker: EventsWorker):
+    worker.start()
+    worker.start()  # Would throw a RuntimeError if the thread was started again.
+    worker.stop()
```
chrisguidry, line 57: Nice!

chrisguidry (approved):
Super excited about this!

src/prefect/events/worker.py
```diff
@@ -0,0 +1,127 @@
...
+        if self._thread.is_alive():
+            with self._lock:
+                self._ready_event.clear()
+                self._stop_event.set()
+                self._thread.join()
+
+    def emit(self, event: Event) -> None:
+        """Emit `event` and ensure that the worker thread is running."""
+        if not self._thread.is_alive():
+            self.start()
```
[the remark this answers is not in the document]
bunchesofdonald: Sure

src/prefect/events/worker.py
```diff
@@ -0,0 +1,127 @@
...
+                except queue.Empty:
+                    await asyncio.sleep(0.1)
+                else:
+                    await self._client.emit(event)
+
+    def start(self):
+        """Start worker thread if not already started."""
+        if not self._thread.is_alive():
+            with self._lock:
+                self._thread.start()
```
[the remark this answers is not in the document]
bunchesofdonald: Not as far as I can tell. It works without a lock and no documentation I can find says that we need one before checking.

src/prefect/events/worker.py
bunchesofdonald: @madkinsz explained this to me. Essentially the idea here is that it blocks on start until something goes into the queue and then to stop it you enqueue some sentinel value (like `None`) that signals the thread to unblock and shutdown.

chrisguidry: Ahh yes, a good old sentinel, that works!

madkinsz: @chrisguidry if you put items in the `asyncio.Queue` from outside the event loop that the queue belongs to, it won't wake up the queue when an item is placed in it because the implementation is not threadsafe. That's why you need to wrap `put_nowait` with `call_soon_threadsafe` even if it's just a sync function.

src/prefect/events/worker.py
```diff
@@ -0,0 +1,138 @@
...
+                self._ready_future.set_result(True)
+                while True:
+                    event = await self._queue.get()
+
+                    if event is None:
+                        break
+
+                    await self._client.emit(event)
+        except Exception as exc:
+            self._ready_future.set_exception(exc)
```
madkinsz, line 58: You could probably do this in `_start_loop` if you wanted to reduce the indentation

src/prefect/events/worker.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _enqueue(self, value: Optional[Event]):
+        call_soon_in_loop(self._loop, self._queue.put_nowait, value)
+
+    def start(self):
+        """Start worker thread if not already started."""
+        if not self._thread.is_alive():
+            with self._lock:
+                self._thread.start()
```
madkinsz, line 67: I think threads can't be started multiple times, is the lock doing anything here?

madkinsz: https://docs.python.org/3/library/threading.html#threading.Thread.start

> It must be called at most once per thread object. It arranges for the object’s [run()](https://docs.python.org/3/library/threading.html#threading.Thread.run) method to be invoked in a separate thread of control.
>
> This method will raise a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) if called more than once on the same thread object.

madkinsz (approved):
LGTM! Only minor comments.

src/prefect/events/worker.py
bunchesofdonald: Good question and probably not.

src/prefect/events/worker.py
chrisguidry: Ohhh that makes sense, thank you for clarifying that.  I guess I've almost always been operating in a one-loop situation and not fretting too much about the context switch between sync and async.

src/prefect/events/worker.py
```diff
@@ -0,0 +1,137 @@
...
+        """Start worker thread if not already started."""
+        if not self._thread.is_alive():
+            self._thread.start()
+
+        self._ready_future.result(timeout=5)
+
+    def stop(self) -> None:
+        """Stop worker thread if started."""
+        if self._thread.is_alive():
+            with self._lock:
```
madkinsz, line 73: Sorry to come back again, but now this is your only call to the lock so it's definitely not doing anything.

bunchesofdonald: Thanks Zanie!

Issue #7016, opened 2022-09-29
BUG: Orion version header is misaligned


znicholasbrown:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

The version tag is misaligned (should be on the right) on the settings page:

![Screen Shot 2022-09-29 at 12 35 09 PM](https://user-images.githubusercontent.com/27291717/193088705-5c07da04-0107-4569-ba2c-5397b45ae880.png)

### Reproduction

```python3
Visit the settings page
```

### Error

```python3
N/A
```

### Versions

```Text
2.4.3
```

pleek91:
@space-age-pete wanna take this on?

PR #8676, opened 2023-02-28
Fix version misaligned on the settings page in the UI

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/7016
image

### DIFF
--- a/ui/src/pages/Settings.vue
+++ b/ui/src/pages/Settings.vue
@@ -3,7 +3,7 @@
     <template #header>
       <PageHeading :crumbs="crumbs">
         <template #actions>
-          <p-key-value label="Version" :value="version" alternate />
+          <p-key-value class="settings__version" label="Version" :value="version" alternate />
         </template>
       </PageHeading>
     </template>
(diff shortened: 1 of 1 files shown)


ui/src/pages/Settings.vue
```diff
@@ -3,7 +3,7 @@
     <template #header>
       <PageHeading :crumbs="crumbs">
         <template #actions>
-          <p-key-value label="Version" :value="version" alternate />
+          <p-key-value class="settings__version" label="Version" :value="version" alternate />
```
znicholasbrown, line 6: I'm surprised this is the solution - I'd expect `PageHeading` to be responsible for floating the `actions` slot to the right

pleek91: So page has that as `fit-content`. I also tried `min-content` but that didn't help. p-key-value has `width: 100%` which is the culprit here. I'm not sure why it does tbh. But we're not having any issues what that component elsewhere so adding `width: auto` here seemed like the simplest fix.

znicholasbrown: Fair enough, this is certainly the least-risky change 👍🏻

znicholasbrown (approved):

PR #8684, opened 2023-03-01
remove deployment flowrun fk

jakekaplan:
Remove the `FlowRun.deployment_id` -> `Deployment.id` foreign key. This is for deployment performance purposes and keeps consistent UX behavior with Cloud

### Checklist


src/prefect/server/database/migrations/versions/sqlite/2023_03_01_165551_f3df94dca3cc_remove_flowrun_deployment_fk.py
```diff
@@ -0,0 +1,35 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "f3df94dca3cc"
+down_revision = "8d148e44e669"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("PRAGMA foreign_keys=OFF")
```
cicdw, line 18: do these need to be turned back on after?  I usually do but maybe it's not necessary

cicdw (commented):
qq but otherwise I think this is ready for broader review - @znicholasbrown my UI PRs essentially are so that this change has a little bit cleaner UX

madkinsz (approved):

src/prefect/server/database/migrations/versions/sqlite/2023_03_01_165551_f3df94dca3cc_remove_flowrun_deployment_fk.py
madkinsz: I think it might be necessary or they'll remain off for that database connection?

jakekaplan: woops good call!

cicdw (approved):

PR #8686, opened 2023-03-01
Add basic event instrumentation to blocks

bunchesofdonald:
This adds basic support for sending events when a methods on a block are called. It's supported via decorators that decorate each method of a class and use the `EventsWorker` to emit events when one of those methods are called. The class itself supplies the main `kind` as well as the bulk of the resource and related resources.

Related to #8672 

### Example
```python
from pydantic import SecretStr


secret = Secret(value=SecretStr("I'm hidden!"))
await secret.save("top-secret", overwrite=True)
secret = await Secret.load("top-secret")
secret.get()
```

Would result in three events being emitted, one for each method call to the
Secret block:

```
Event(
    occurred=...,
    event='prefect.block.secret.save.called',
    resource=Resource(__root__={'prefect.resource.id': 'prefect.block-document.top-secret'}),
    related=[RelatedResource(__root__={'prefect.resource.id': 'prefect.block-type.secret', 'prefect.resource.role': 'block-type'})],
    payload={},
    id=UUID(...)
)

Event(
    occurred=...,
    event='prefect.block.secret.load.called',
    resource=Resource(__root__={'prefect.resource.id': 'prefect.block-document.top-secret'}),
    related=[RelatedResource(__root__={'prefect.resource.id': 'prefect.block-type.secret', 'prefect.resource.role': 'block-type'})],
    payload={},
    id=UUID(...)
)

Event(
    occurred=...,
    event='prefect.block.secret.get.called',
    resource=Resource(__root__={'prefect.resource.id': 'prefect.block-document.top-secret', 'prefect.result': 'successful'}),
    related=[RelatedResource(__root__={'prefect.resource.id': 'prefect.block-type.secret', 'prefect.resource.role': 'block-type'})],
    payload={},
    id=UUID(...)
)
```

### Checklist

### DIFF
--- a/src/prefect/blocks/core.py
+++ b/src/prefect/blocks/core.py
@@ -28,6 +28,11 @@ import prefect
 import prefect.exceptions
 from prefect.blocks.fields import SecretDict
 from prefect.client.utilities import inject_client
+from prefect.events.instrument import (
+    ResourceTuple,
+    emit_instance_method_called_event,
+    instrument_method_calls_on_class_instances,
+)
 from prefect.logging.loggers import disable_logger
 from prefect.server.schemas.core import (
     DEFAULT_BLOCK_SCHEMA_VERSION,
@@ -134,6 +139,7 @@ def _collect_secret_fields(name: str, type_: Type, secrets: List[str]) -> None:
 
 
 @register_base_type
+@instrument_method_calls_on_class_instances(exclude_methods=["block_initialization"])
 class Block(BaseModel, ABC):
     """
         A base class for implementing a block that wraps an external service.
@@ -575,6 +581,10 @@ class Block(BaseModel, ABC):
             else cls.get_block_class_from_schema(block_document.block_schema)
         )
 
+        block_cls = instrument_method_calls_on_class_instances(
+            exclude_methods=["block_initialization"],
+        )(block_cls)
+
         block = block_cls.parse_obj(block_document.data)
         block._block_document_id = block_document.id
         block.__class__._block_schema_id = block_document.block_schema_id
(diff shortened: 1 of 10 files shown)


src/prefect/blocks/core.py
```diff
@@ -585,8 +595,34 @@ def _from_block_document(cls, block_document: BlockDocument):
...
 
+    def _event_kind(self) -> str:
+        return "prefect.block"
+
+    def _event_method_called_resources(self, method_name: str) -> ResourceTuple:
+        name = self._block_document_name if self._block_document_name else "anonymous"
+        return (
+            {
+                "prefect.resource.id": f"prefect.block-document.{name}.{method_name}",
+            },
```
chrisguidry, line 615: So in this case, the resource is a block-document _method_ with a related resource of the block type, but I'm thinking that the method name should be in the event name so and then the resource is just the block(-document) itself:

```suggestion
            {
                "prefect.resource.id": f"prefect.block-document.{name}",
            },
```

This way the block can stand is as the remote resource, like a database or a bucket.

pleek91:
Looking at the emitted events I'm curious about `prefect.block-document.anonymous.save` as a resource id. I see the slug for any non anonymous blocks which makes sense. But without a guid we have no way of displaying the block in the UI (or even knowing much about it at all). 

So I want to question using slugs over ids for two reasons
1. Anonymous blocks
2. Work queue's are generally known by their name rather than their id (similar to block slugs vs ids) but the work queue events use the id in the resource id rather than the name. Treating blocks the same way brings some consistency.

chrisguidry (commented):
I didn't see any assertions on what the `event.event` comes out as, but I think it should be something like:

```
prefect.block.{block-type}.{method-name}.(called|failed)
```

So you might get:

```
event="prefect.block.s3-bucket.write_path.called",
resource={
  "prefect.resource.id": "prefect.block-document.my-bucket"
}
```

src/prefect/blocks/core.py
pleek91: Aren't automation executions similar to this? The execution is part of the resource id. Curious what the difference in thought is between block method calls and automation action executions.

chrisguidry: These are the _event_ names (not the resource IDs) for automations:

image

chrisguidry: So what I'm suggesting is block methods look like that.  Also, I wonder if we should change the automations to look like:

```
prefect-cloud.automation.action.run-deployment.executed
prefect-cloud.automation.action.send-notification.failed
```

pleek91: But don't the id's have the action number?

chrisguidry: Those are the resource IDs:

image

chrisguidry: The `event` should be "what happened", and the `prefect.resource.id` should be "what it happened to"

bunchesofdonald:
> Looking at the emitted events I'm curious about `prefect.block-document.anonymous.save` as a resource id. I see the slug for any non anonymous blocks which makes sense. But without a guid for an anonymous block we have no way of displaying the block in the UI (or even knowing much about it at all).
> 
> So I want to question using slugs over ids for two reasons
> 
> 1. Anonymous blocks
> 2. Work queue's are generally known by their name rather than their id (similar to block slugs vs ids) but the work queue events use the id in the resource id rather than the name. Treating blocks the same way brings some consistency.
> 
> EDIT Looking at this closer is that even an anonymous block? Do we just not have the id when the event is emitted?

Good catch @pleek91, I'll take a look at that. In my testing it was coming back with the right block name and not `anonymous`, maybe something weird about it being in a test.

pleek91:
@bunchesofdonald im also curious what events for anonymous blocks look like. Will they all appear to be the same resource?

bunchesofdonald:
At the moment `anonymous` blocks won't emit events. I should mention this in the channel, but my thinking was that they'd all appear to be the same resource and wouldn't be particularly useful.

chrisguidry (approved):
:heart:

PR #8695, opened 2023-03-02
Refactor supervisor API to allow configuration

madkinsz:
Replaces https://github.com/PrefectHQ/prefect/pull/8668

- Move `from_sync` and `from_async` implementations to single file
- Update API to consume `Call` objects instead of bare calls so kwargs can be used for options

```python
import time
from prefect._internal.concurrency.api import create_call, from_sync

supervisor = from_sync.supervise_call_in_worker_thread(
    create_call(time.sleep, 2),
    timeout=1,
)

supervisor.result()  # TimeoutError
```


src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -63,6 +66,33 @@
...
+class Call(Generic[T]):
+    fn: Callable[..., T]
+    args: Tuple
+    kwargs: Dict[str, Any]
+
+    def __call__(self) -> Any:
+        return self.fn(*self.args, **self.kwargs)
+
+    @classmethod
+    def new(self, __fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> "Call[T]":
```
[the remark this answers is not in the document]
madkinsz: Fixing this in the next pull request

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -63,6 +66,37 @@ def set_supervisor(supervisor: "Supervisor"):
         current_supervisor.reset(token)
 
 
+@dataclasses.dataclass
+class Call(Generic[T]):
```
madkinsz, line 70: Going to combine this and `WorkItem` in a follow-up

src/prefect/_internal/concurrency/api.py
```diff
@@ -0,0 +1,141 @@
...
+    get_supervisor,
+)
+
+P = ParamSpec("P")
+T = TypeVar("T")
+Future = Union[concurrent.futures.Future, asyncio.Future]
+
+
+def create_call(__fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> Call[T]:
+    return Call(fn=__fn, args=args, kwargs=kwargs)
```
chrisguidry, line 23: Is there an advantage of the `create_call` factory versus folks just creating `Call` objects or using `Call.new`?  Maybe a layer of indirection we don't need?

chrisguidry (approved):
:boom: I dig it

src/prefect/_internal/concurrency/api.py
madkinsz: Good question. With `Call.new` around it feels less necessary. The best reason I can think of to keep it here is that I think `Call` should be a private implementation detail of the `supervisors` module and this module should contain the full "public API" for managing concurrency.

PR #8697, opened 2023-03-02
Consolidate `WorkItem` and `Call` classes

madkinsz:
All changes are internal to the `supervisors` module.

- Combine `WorkItem` and `Call` classes
- Update `Call.__call__` to use `Call.run` behind the scenes
- Improve `Call.run` handling of async

### DIFF
--- a/src/prefect/_internal/concurrency/supervisors.py
+++ b/src/prefect/_internal/concurrency/supervisors.py
@@ -69,19 +69,56 @@ def set_supervisor(supervisor: "Supervisor"):
 @dataclasses.dataclass
 class Call(Generic[T]):
     """
-    A deferred function call; a convenience for transport.
+    A deferred function call.
     """
 
+    future: concurrent.futures.Future
     fn: Callable[..., T]
     args: Tuple
     kwargs: Dict[str, Any]
-
-    def __call__(self) -> Any:
-        return self.fn(*self.args, **self.kwargs)
+    context: contextvars.Context
 
     @classmethod
-    def new(self, __fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> "Call[T]":
-        return Call(fn=__fn, args=args, kwargs=kwargs)
+    def new(cls, __fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> "Call[T]":
+        return cls(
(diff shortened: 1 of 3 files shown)


src/prefect/cli/deployment.py
```diff
@@ -787,8 +787,8 @@ async def build(
         "--description",
         "-d",
         help=(
-            "The description to give the deployment. "
-            "If not provided, the description will be populated from the flow's description."
+            "The description to give the deployment. If not provided, the description"
+            " will be populated from the flow's description."
```
madkinsz, line 791: idk why this is passing in the base but upset here, w/e

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -119,65 +88,97 @@ def from_call(cls, __fn, *args, **kwargs) -> "WorkItem":
...
-                return self._run_async()
+                return self._run_async(coro)
             else:
                 # Otherwise, execute the function here
-                return asyncio.run(self._run_async())
-        else:
-            return self._run_sync()
+                return asyncio.run(self._run_async(coro))
+
+        return None
```
madkinsz, line 113: This is actually kind of the juicy part.

The engine wraps async functions with a synchronous function that returns a coroutine. However, since it's not `async def` anymore the `inspect.iscoroutinefunction` call returns `False`! So here we always call `_run_sync` then check if what was returned was a coroutine and if so we'll return `_run_async`.

src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -119,65 +88,97 @@ def from_call(cls, __fn, *args, **kwargs) -> "WorkItem":
...
 
     def _run_sync(self):
         try:
             result = self.context.run(self.fn, *self.args, **self.kwargs)
+
+            # Return the coroutine for async execution
+            if inspect.isawaitable(result):
+                return result
```
madkinsz, line 121: Here we don't set the result on the future because we're going to hand that off to `_run_async`

chrisguidry (approved):
I love it when a plan comes together!  `WorkItem` was kinda "meh" even when we first came up with it :D  `Call` is much better.

PR #8700, opened 2023-03-02
Chore: Update CSP script-src to include bundled mermaid script

znicholasbrown:
Should fix an issue where the deployment docs were throwing an error.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -21,147 +21,182 @@ publish = "site"
 # moved page redirects #
 
 [[redirects]]
-  from = "/collections/overview/"
-  to = "/collections/catalog/"
-  status = 301
+from = "/collections/overview/"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -21,147 +21,182 @@ publish = "site"
...
+      object-src \
+        'none' \
+        ; \
+      script-src \
+        'self' \
+        'unsafe-inline' \
+        https://cdn.segment.com/ \
+        https://cdn.segment.prefect.io \
+        https://www.googletagmanager.com \
+        https://unpkg.com/mermaid \
```
znicholasbrown, line 174: This is the only actual new content - everything else is a difference in TOML formatter between me and @jawnsy

jawnsy (approved):

PR #8705, opened 2023-03-03
Add carriage return to `prompt_select_from_list`

madkinsz:
Otherwise, when I attempted to use the `prefect cloud login` CLI command it would not let me move forward.

`ENTER` on my machine actually sent `CR` then `CTRL_M`; no idea what's up with that maybe it's a readchar bug?

```
❯ pip show readchar          
Name: readchar
Version: 4.0.3
```


serinamarie (approved):
LGTM 🐴

PR #8714, opened 2023-03-06
Refactor supervisors to interact directly with "Worker" threads

madkinsz:
Instead of supporting a "submission function" which returned a future, supervisors now take a "worker" which allows submission of a `Call` object. Use of `Call` for deferred function calls is used everywhere now; previously, the call was unpacked during submission. The runtime thread and executor modules are merged into a `concurrency.workers` module which defines a simple worker thread. This improves visibility into which thread is being supervised and reduces the number of concepts involved. There is no change in the `concurrency.api` interface. We will always spawn a new worker thread now though, instead of using a pool with a fixed limit. The "runtime" thread has been renamed to the "global worker".


src/prefect/_internal/concurrency/workers.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+
+        # Enforce a maximum length
+        if len(call_args) > 100:
+            call_args = call_args[:100] + "..."
+
+        return f"{name}({call_args})"
+
+
+class Worker:
```
chrisguidry, line 137: Are we asking for trouble naming this `Worker`?  I know it's private, but searching the code for `Worker` is going to be sampling from two disjoint worlds (`Worker` the product feature and `Worker` the concurrency implementation detail).  Perhaps `Minion`? :D

chrisguidry (approved):

src/prefect/_internal/concurrency/workers.py
madkinsz: I was thinking about that too. I'll try to think of a name that doesn't overlap.

Issue #8712, opened 2023-03-06
Prefect UI fails to display deployment page (Cannot access 't' before initialization)


dejapong:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

Attempting to display a deployment page (i.e. the /deployments/deployment/<uuid> route in the UI) fails. In the developer console, I receive an error about a variable referenced before assignment. 

### Reproduction

Create a deployment in Prefect v2.8.4, go to the deployment detail page and it fails to load. Here's an example deployment that will cause the issue for me:
[example.txt](https://github.com/PrefectHQ/prefect/files/10901008/example.txt) 
[environment.txt](https://github.com/PrefectHQ/prefect/files/10901038/environment.txt)

### Error

![Screen Shot 2023-03-06 at 9 26 45 AM](https://user-images.githubusercontent.com/633205/223189705-209304f8-31fe-4a12-a00a-2ddbada09708.png)

### Browers

- [X] Chrome
- [X] Firefox
- [X] Safari
- [ ] Edge

### Prefect version

```Text
Prefect 2.8.4 running both in M1 mac, and x86 Linux. See environment.txt for full python environment
```

PR #8716, opened 2023-03-06
Bugfix: Deployment and ConcurrencyLimit pages tabs

znicholasbrown:
Updates the Deployment and ConcurrencyLimit pages to use the new `useTabs` composition return value properly instead of trying to map on an object. 

Resolves: #8712

### DIFF
--- a/ui/src/pages/Deployment.vue
+++ b/ui/src/pages/Deployment.vue
@@ -42,12 +42,7 @@
     </p-tabs>
 
     <template #well>
-      <DeploymentDetails
-        v-if="deployment"
-        :deployment="deployment"
-        alternate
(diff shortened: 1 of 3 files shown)


ui/src/pages/Deployment.vue
```diff
@@ -72,17 +67,17 @@
     interval: 300000,
   }
 
+  const deploymentSubscription = useSubscription(api.deployments.getDeployment, [deploymentId.value], subscriptionOptions)
+  const deployment = computed(() => deploymentSubscription.response)
```
znicholasbrown, line 71: Technically we were trying to access these before initialization in `computedTabs`

pleek91 (approved):

PR #8717, opened 2023-03-06
Add portal abstract base for worker threads and supervisors

madkinsz:
Extends #8714 renaming `Worker` to `Portal` to avoid conflating with our "Workers" concept. Moves the previous worker thread implementation from `Worker` to `WorkerThreadPortal`. Updates `Supervisor` to conform to a `Portal` interface as well. We now have a `Supervisor` which monitors a call occurring on a worker thread while also allowing work to be submitted to its internal queue; or more generally: a supervisor monitors a call running in another portal while providing a portal for the call to send work back to.

I admit, this is a little awkward right now but I expect things to become clearer in the future.


chrisguidry (approved):
I can definitely vibe with the term `Portal`!

PR #8702, opened 2023-03-02
Use new concurrency tooling in engine

madkinsz:
Updates the flow and task run engine entrypoints to use waiters and uses new worker thread tooling for flow and task execution.

Some changes in behavior:

- All engine functions are executed in a Prefect managed thread and dedicated event loop instead of the main thread
- Synchronous flows are run on the main thread instead of worker threads
- Synchronous flows timeouts are now enforced with signals, which interrupts long-running system calls like `sleep`

Fixes some known bugs:

- Async flows can now be called from sync flows
- Async tasks can now be used as upstream dependencies for sync tasks in async flows
- Waiting for many tasks that sleep no longer causes a deadlock as long as there are less upstreams than the thread limit
- Closes https://github.com/PrefectHQ/prefect/issues/7393

Additional changes:

- Adds more test coverage for timeouts and crashes
- Adds benchmarks with more task runs
- Fixes benchmark comparisons with the main branch
- Changes to waiter/call concurrency tooling to meet requirements

Replaces https://github.com/PrefectHQ/prefect/pull/8014

PR #8718, opened 2023-03-06
Fix bugs in supervisors implementation

madkinsz:
Cherry-picking fixes from https://github.com/PrefectHQ/prefect/pull/8702 onto #8717


src/prefect/_internal/concurrency/supervisors.py
```diff
@@ -151,9 +151,7 @@ def _call_in_supervised_thread(*args, **kwargs):
             if inspect.isawaitable(retval):
 
                 async def _call_in_supervised_coro():
-                    # TODO: Technnically, this should use the deadline but it is not
-                    #       clear how to mix sync/async deadlines yet
-                    with cancel_async_after(self._timeout):
+                    with cancel_async_at(deadline):
```
madkinsz, line 154: Always using a monotonic clock for deadlines now

PR #8719, opened 2023-03-06
Add chaining to cancel contexts

madkinsz:
Adds a `chain` method to `CancelContexts` allowing cancelled marks to be propagated upstream. This will be required for adding a `cancelled` attribute to `Call` objects in supervisors since they can be cancelled by multiple implementations (sync / async); this is, in turn, required for only reporting cancellations caused by us as flow run timeouts in #8702.


chrisguidry (approved):
Slick!

PR #8729, opened 2023-03-07
Use 8-core machines for GitHub Actions runs of test suite

madkinsz:
As seen in https://github.com/PrefectHQ/nebula/pull/3953/

Not sure if we have these available in this repository.


madkinsz:
Not supported; tracking at https://github.com/PrefectHQ/platform/issues/2785

PR #8730, opened 2023-03-07
Update block event resource IDs to use block-document id instead of name.

bunchesofdonald:
While testing out the block events we found that while the `name` of the block document was nicely human-readable it didn't provide an easy way to hook back into the API to get the block document itself. This updates the block events to emit the document id as the resource id and include the name as a label.

### Checklist

### DIFF
--- a/src/prefect/blocks/core.py
+++ b/src/prefect/blocks/core.py
@@ -611,11 +611,16 @@ class Block(BaseModel, ABC):
     def _event_kind(self) -> str:
         return f"prefect.block.{self.get_block_type_slug()}"
 
-    def _event_method_called_resources(self) -> ResourceTuple:
-        name = self._block_document_name if self._block_document_name else "anonymous"
+    def _event_method_called_resources(self) -> Optional[ResourceTuple]:
+        if not self._block_document_id:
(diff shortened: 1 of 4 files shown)


pleek91:
Thanks @bunchesofdonald!

zangell44 (approved):

src/prefect/blocks/core.py
```diff
@@ -611,11 +611,16 @@ def _from_block_document(cls, block_document: BlockDocument):
...
+        if not self._block_document_id:
+            return None
+
         return (
             {
-                "prefect.resource.id": f"prefect.block-document.{name}",
+                "prefect.resource.id": (
+                    f"prefect.block-document.{self._block_document_id}"
+                ),
+                "prefect.block-document.name": self._block_document_name,
```
chrisguidry, line 623: I think we've used `prefect.name` in other cases...I think it's better if we have fewer kinds of labels and use the same label to represent the same concept (like a `prefect.name` or a `prefect.slug`)

madkinsz (approved):

chrisguidry (approved):
Woot!

PR #8734, opened 2023-03-07
Add tests for base exceptions and calls

madkinsz:
Adds test coverage for calls and capturing base exceptions during waits


tests/_internal/concurrency/test_waiters.py
```diff
@@ -180,37 +180,91 @@ def sync_then_async_sleep():
     assert call.cancelled()
 
 
-@pytest.mark.parametrize("fn", [identity, aidentity])
-def test_sync_call(fn):
-    call = Call.new(fn, 1)
-    assert call() == 1
```
madkinsz, line 186: These tests moved to `test_calls.py`

PR #8737, opened 2023-03-07
Add support for float timeouts using alarms

madkinsz:
After much annoyance adjusting tests that used subsecond timeouts, I figured out how to make float values work with alarm based timeouts.

See https://docs.python.org/3/library/signal.html#signal.ITIMER_REAL and https://docs.python.org/3/library/signal.html#signal.setitimer


serinamarie (approved):

chrisguidry (approved):
Neat find!

PR #8745, opened 2023-03-08
Improve CI time for migration tests

madkinsz:
- Skip backfill test on SQLite (takes ~60s vs ~6s on Postgres)
- Add service test mark to database migration test to run in separate job (takes ~60s)


jakekaplan (approved):
⚡

PR #8756, opened 2023-03-09
Added task restarts to client and schema

rpeden:
This PR contains the client-side changes necessary to enable task restarts:
* Add an `allow_restarts` field to `TaskRunPolicy`
* Add an `allow_restarts` parameter to the `Task` class and `task` decorator
* Set `allow_restarts` for each new task upon creation in `PrefectClient.create_task_run`

Further description updates to be added shortly.

### Checklist

### DIFF
--- a/src/prefect/server/schemas/core.py
+++ b/src/prefect/server/schemas/core.py
@@ -339,6 +339,13 @@ class TaskRunPolicy(PrefectBaseModel):
     retry_jitter_factor: Optional[float] = Field(
         default=None, description="Determines the amount a retry should jitter"
     )
+    allow_restarts: bool = Field(
+        default=False,
+        description=(
+            "Determines whether retries are allowed after the task has"
+            "reached the RUNNING state."
+        ),
+    )
 
     @root_validator
     def populate_deprecated_fields(cls, values):
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -1799,6 +1799,7 @@ class PrefectClient:
                 retries=task.retries,
                 retry_delay=task.retry_delay_seconds,
                 retry_jitter_factor=task.retry_jitter_factor,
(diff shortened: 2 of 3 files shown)


src/prefect/server/schemas/core.py
```diff
@@ -339,6 +339,13 @@ class TaskRunPolicy(PrefectBaseModel):
     retry_jitter_factor: Optional[float] = Field(
         default=None, description="Determines the amount a retry should jitter"
     )
+    allow_restarts: bool = Field(
+        default=False,
+        description=(
+            "Determines whether retries are allowed after the task has"
+            "reached the RUNNING state."
```
madkinsz, line 346: This doesn't seem quite accurate. Retries are controlled by the retries option. This allows "backwards" or "same-to-same" transitions?

cicdw: I actually think this should default to `True`

madkinsz: We had a lot of problems with this being the default behavior in V1 and the intent of this option was to have people _opt in to possible duplicate runs_ which means they then need to ensure their code is safe for re-execution. I feel like not allowing duplicate runs by default makes more sense since it's clearer orchestrator behavior when you don't have an unstable execution environment.

madkinsz: Note this option doesn't affect like... restarting the flow run manually or retries. It's just allowing a run to be orchestrated again when we did not think it should be.

cicdw: I don't think I'm remembering what the v1 problems with this were - what were they?

madkinsz: I think primarily user complaints about us not preventing duplicate execution of runs but honestly my memory is vague and I don't have any concrete examples to point to right now 🤷‍♀️

Issue #8710, opened 2023-03-06
Flow could not be retrieved from deployment with s3fs==2023.3.0


LeeMendelowitz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I started to get a "Flow could not be retrieved from deployment" error for a flow run deployment which had been previously working.

After several hours of investigation, I tracked it down a trailing slash required for `S3.bucket_path` with dependency `s3fs==2023.3.0`.  A trailing slash was not required under `s3fs==2023.1.0`. 

By simply adding a trailing slash to the `S3.bucket_path`, everything seems to work again.

### Reproduction

```python3
from prefect import flow, get_run_logger
from prefect.deployments import Deployment
from prefect.filesystems import S3
from prefect.infrastructure import Process
from pathlib import Path

@flow
def my_flow():
    logger = get_run_logger()
    logger.info("Hello")
    return True

# The deployment flow run fails without a trailing slash in the `bucket_path` on s3fs==2023.3.0
# The deployment flow run succeeds without a trailing slash in the `bucket_path` on s3fs==2023.1.0
storage = S3(bucket_path = 'my-bucket/test-flow')

working_dir = Path('./working_dir').resolve()
working_dir.mkdir(exist_ok = True)
infrastructure = Process(working_dir=working_dir)

my_deployment = Deployment.build_from_flow(
    flow = my_flow,
    name = 'my-deployment',
    output = 'deployment.yaml',
    storage = storage,
    infrastructure = infrastructure,
    skip_upload = True,
    apply = False
)

if __name__ == '__main__':
    my_deployment.apply(upload = True)
```

### Error

```python3
Downloading flow code from storage at None

Flow could not be retrieved from deployment.
Traceback (most recent call last):
  File "<frozen importlib._bootstrap_external>", line 879, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1016, in get_code
  File "<frozen importlib._bootstrap_external>", line 1073, in get_data
FileNotFoundError: [Errno 2] No such file or directory: '/Users/lmendelowitz/playground/prefect/storage_issue/working_dir/deployment.py'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #8757, opened 2023-03-09
Ignore fsspec `2023.3.0` during requirements installation

madkinsz:
This appears to have a breaking change and we can try to avoid releasing it to our users https://github.com/PrefectHQ/prefect/issues/8710


serinamarie (approved):
Nice! 🗄️

PR #8759, opened 2023-03-09
Add artifacts for unpersisted results

madkinsz:
Previously, when results were not persisted no artifact would be sent to the API unless the return value was `None`. This can cause confusing behavior i.e. the presence of a return value of the task or flow can affect restart behavior. It also leads to a less than ideal experience with the upcoming result and artifact display in the UI. In general, result persistence needed to be enabled to see any information about the results in the UI; unless your task or flow returned `None`, then we'd show a result in the UI. This behavior made it challenging to set expectations about what user action was required to see results in the UI.

Here, instead of leaning on `LiteralResult(None)` to distinguish between unpersisted results and null return values, we introduce an `UnpersistedResult` class. This type is used whenever result persistence is disabled. The result value itself is cached in memory, but never sent to the API. The result artifact _is_ sent to the API though which allows metadata for unpersisted results to be tracked and eventually displayed in the UI.

Note we include the _type_ of the result _value_ as the default artifact description for unpersisted results. We should expose ways to customize this description and allow opt-in of increased exposure of the value (e.g. a `repr` of the result value) or opt-out (e.g. no description or no artifact).

Now, `LiteralResult` is used to represent _persisted results_ whose data was deemed unworthy of persistence outside of the Prefect database. We store bools and nulls in the database directly to reduce the overhead of writing to external locations. We'll likely support additional types and data here in the future.

There should be no user-facing changes here yet. 

### Checklist

### DIFF
--- a/src/prefect/results.py
+++ b/src/prefect/results.py
@@ -8,7 +8,7 @@ from typing_extensions import Self
 import prefect
 from prefect.blocks.core import Block
 from prefect.client.utilities import inject_client
-from prefect.exceptions import MissingContextError
+from prefect.exceptions import MissingContextError, MissingResult
 from prefect.filesystems import LocalFileSystem, ReadableFileSystem, WritableFileSystem
 from prefect.logging import get_logger
 from prefect.serializers import Serializer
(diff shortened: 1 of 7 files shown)


serinamarie (approved):
LGTM! 🚀

PR #8762, opened 2023-03-09
Use file locks for database migrations to prevent multiprocessing conflicts

madkinsz:
Instead of using a thread-lock which is safe for single-process migrations, we use a file lock which is safe for multiprocess migrations. This is notably not safe for multimachine migrations which as possible when using a remote postgres server. In that case, we ought to be using a lock on a table in the database.


madkinsz:
Unclear that this is the issue.

PR #8763, opened 2023-03-09
Add setting to exclude traversal of Pydantic models in parameters

madkinsz:
Addresses concerns raised in #8542 by allowing Pydantic model traversal to be disabled so they are not copied during visits. We could introduce a setting like this for dataclasses as well.

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -594,6 +594,16 @@ The default setting for persisting results when not otherwise specified. If enab
 flow and task results will be persisted unless they opt out.
 """
 
+PREFECT_PARAMETERS_TRAVERSE_PYDANTIC_MODELS = Setting(
+    bool,
+    default=True,
+)
+"""
+Determines if Pydantic models should be traversed in order to resolve Prefect futures
+and states into results.
+"""
+
+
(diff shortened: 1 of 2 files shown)


src/prefect/utilities/collections.py
```diff
@@ -29,6 +29,8 @@
 
 import pydantic
 
+from prefect.settings import PREFECT_PARAMETERS_TRAVERSE_PYDANTIC_MODELS
```
madkinsz, line 32: This seems like it ought to live in the engine and `visit_collection` should take a set of types to exclude traversal of. However, it's simpler to put it here and I'm not sure how much the purity matters.

src/prefect/settings.py
```diff
@@ -594,6 +594,16 @@ def default_cloud_ui_url(settings, value):
 flow and task results will be persisted unless they opt out.
 """
 
+PREFECT_PARAMETERS_TRAVERSE_PYDANTIC_MODELS = Setting(
```
madkinsz, line 597: I'm not sure I love this name.

madkinsz:
Deferring this in favor of explicitly using quote for now.

PR #8772, opened 2023-03-10
Update default result descriptions

madkinsz:


src/prefect/results.py
```diff
@@ -504,7 +492,12 @@ async def create(
         key = uuid.uuid4().hex
         await storage_block.write_path(key, content=blob.to_bytes())
 
-        description = cls._infer_description(obj, storage_block, key)
+        description = f"Result of type {type(obj).__name__!r}"
+        uri = cls._infer_path(storage_block, key)
+        if uri:
+            description += f" persisted to {uri}."
+        else:
+            description += f" persisted with storage block '{storage_block_id}'."
```
billpalombi, line 500: 👍 I think including the block if we don't have the URI, the block is a good fallback.

src/prefect/results.py
```diff
@@ -504,7 +492,12 @@ async def create(
         key = uuid.uuid4().hex
         await storage_block.write_path(key, content=blob.to_bytes())
 
-        description = cls._infer_description(obj, storage_block, key)
+        description = f"Result of type {type(obj).__name__!r}"
+        uri = cls._infer_path(storage_block, key)
+        if uri:
+            description += f" persisted to {uri}."
```
billpalombi, line 498:
```suggestion
            description += f" Persisted to {uri}."
```

src/prefect/results.py
billpalombi, line 500:
```suggestion
            description += f" Persisted with storage block '{storage_block_id}'."
```

billpalombi (approved):
Thanks for jumping on this so quickly. I think this is the right place to start.

src/prefect/results.py
madkinsz: this gets appended to the previous string so we don't want to capitalize it

serinamarie (approved):
LGTM! ⚡

PR #8775, opened 2023-03-10
Add webhook block to docs API reference

madkinsz:


madkinsz:
Dupe of https://github.com/PrefectHQ/prefect/pull/8773

Issue #6519, opened 2022-08-23
Intermittent fatal error running deployment with docker infra


bjorhn:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

We run 40-50 deployments per day and one or two of them will usually crash within the first 5 seconds with a docker error. Re-running the deployment always works. All of our deployments are docker deployments. No logs are sent to the server GUI, but they can be retrieved from the agent.

Let me know if there's any additional information you need.

### Reproduction

Since the problem is intermittent it's difficult to put together a minimal example.

### Error

```
08:11:55.728 | INFO    | prefect.agent - Submitting flow run '00dc9668-fb5b-4935-9bfb-0892a865d982'
08:11:56.201 | ERROR   | prefect.agent - Failed to submit flow run '00dc9668-fb5b-4935-9bfb-0892a865d982' to infrastructure.
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/prefect/infrastructure/docker.py", line 443, in _get_client
    docker_client = docker.from_env()
AttributeError: module 'docker' has no attribute 'from_env'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/prefect/agent.py", line 199, in submit_run
    await self.task_group.start(submit_flow_run, flow_run, infrastructure)
  File "/usr/local/lib/python3.8/dist-packages/anyio/_backends/_asyncio.py", line 807, in start
    return await future
  File "/usr/local/lib/python3.8/dist-packages/prefect/infrastructure/submission.py", line 48, in submit_flow_run
    return await infrastructure.run(task_status=task_status)
  File "/usr/local/lib/python3.8/dist-packages/prefect/infrastructure/docker.py", line 209, in run
    container_id = await run_sync_in_worker_thread(self._create_and_start_container)
  File "/usr/local/lib/python3.8/dist-packages/prefect/utilities/asyncutils.py", line 56, in run_sync_in_worker_thread
    return await anyio.to_thread.run_sync(call, cancellable=True)
  File "/usr/local/lib/python3.8/dist-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
[...]

PR #8788, opened 2023-03-13
Rename `docker` modules to avoid conflict

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/6519


serinamarie (approved):
Nice! 🤠

Pierre-VF:
Do you already know when this will be merged? (we're having many problems with this in our production setup at the moment)
Or is there a plan to solve it in another manner? (I am unsure about what the "Deprecated" tag means, given that there isn't extra information about it in the thread)

madkinsz:
@Pierre-VF this needs more work to be completed.

The old module path (that is causing problems) is deprecated. This retains backwards compatibility for other users, but we've moved it to a new path to resolve the issues.

The Docker worker also should resolve this problem https://prefecthq.github.io/prefect-docker/worker/

Issue #8794, opened 2023-03-14
Run_deployment subflow kicked off from inside of a wrapper task are not associated with their wrapper tasks. 


taylor-curran:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

Users often call run_deployment from inside of tasks. Is there a way that the parent task can be associated with the subflow it has triggered?

All childflows are on the inner ring even though their wrapper flows are properly situated along the radar chart.

![image](https://user-images.githubusercontent.com/68969861/225083479-9039b99f-4b92-49c3-a2a0-234e3a2b40e3.png)

### Reproduction
```python
from prefect import flow, task
from datetime import date, timedelta
from prefect_dask import DaskTaskRunner
from prefect.deployments import run_deployment
from prefect.task_runners import ConcurrentTaskRunner
from pydantic import BaseModel
from datetime import timedelta
from prefect.tasks import task_input_hash

# prefect deployment build concurrent_task_runner.py:child_flow_a -n child-deployment-a -t 2506 -a
@flow(persist_result=True)
def child_flow_a(sim_failure_child_flow_a):
    if sim_failure_child_flow_a:
        raise Exception("This is a test exception")
    else:
        return {'a': 'child flow a'}

# prefect deployment build concurrent_task_runner.py:child_flow_b -n child-deployment-b -t 2506 -a
@flow(persist_result=True)
def child_flow_b(sim_failure_child_flow_b):
    if sim_failure_child_flow_b:
        raise Exception("This is a test exception")
    else:
        return {'b': 'child flow b'}

# prefect deployment build concurrent_task_runner.py:child_flow_c -n child-deployment-c -t 2506 -a
@flow(persist_result=True)
def child_flow_c():
    return {'c': 'child flow c'}

# prefect deployment build concurrent_task_runner.py:child_flow_d -n child-deployment-d -t 2506 -a
@flow(persist_result=True)
def child_flow_d():
    return {'d': 'child flow d'}

# ---

@task(cache_key_fn=task_input_hash, cache_expiration=timedelta(minutes=1))
def upstream_task_h():
    print('upstream task')
    return {'h': 'upstream task'}

@task(cache_key_fn=task_input_hash, cache_expiration=timedelta(minutes=1))
def upstream_task_i():
[...]

PR #8796, opened 2023-03-14
Link flow runs created by `run_deployment` to the task they are called from

zanieb:
Closes https://github.com/PrefectHQ/prefect/issues/8794

We use a dummy input parameter instead of replacing the representative task. Replacing the representative task has orchestration implications and we hide representative tasks from the UI so it'd be complex to do so.

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -21,7 +21,7 @@ from prefect.blocks.core import Block
 from prefect.blocks.fields import SecretDict
 from prefect.client.orchestration import PrefectClient, get_client
 from prefect.client.utilities import inject_client
-from prefect.context import FlowRunContext, PrefectObjectRegistry
+from prefect.context import FlowRunContext, PrefectObjectRegistry, TaskRunContext
 from prefect.exceptions import (
     BlockMissingCapabilities,
     ObjectAlreadyExists,
@@ -102,6 +102,7 @@ async def run_deployment(
         deployment = await client.read_deployment_by_name(name)
 
     flow_run_ctx = FlowRunContext.get()
+    task_run_ctx = TaskRunContext.get()
     if flow_run_ctx:
         # This was called from a flow. Link the flow run as a subflow.
         from prefect.engine import (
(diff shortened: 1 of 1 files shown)


madkinsz:
@brandonreid would you be interested in trying this out?

brandonreid:
Ah thanks @madkinsz !

I have this on my todo list for very soon (need to get my local backend spruced up again). This isn't a blocking requirement for getting subflow's showing, but could be handy in circumstances where I happen to be seeing a subflow somehow and want to traverse to the parent, etc.

**How I'm currently getting subflows, if you're curious:**
Currently the graphs endpoint does provide a `child_flow_run_id` on those dummy tasks from the `state_details`. So basically when I request the initial graph for a flow run, I can determine that tasks with `state.state_details.child_flow_run_id` are actually subflows, then request the subflow and I'm able to associate it with the parent task from that moment.

madkinsz:
@brandonreid Just to make sure we're on the same page: The goal here is when you run a deployment from a task within a flow we should be able to show that relationship. We'll still have a separate dummy task in the parent flow. We'll be adding an edge from the calling task to the dummy task. Right now, we only show the relationship to the parent flow not the calling task.

billpalombi:
@madkinsz do "representative task" and "dummy task" mean in this context? Is that a task that represents the flow run created by `run_deployment`?

madkinsz:
@billpalombi and I talked about this offline I think?

@brandonreid should I be prioritizing this work?

billpalombi:
Yes, we want to do this.

serinamarie (approved):

PiotrSiejda:
I would say this change would not work properly when idempotency_key is used on the subflow which was called by two different flows. As on the subflow there is possibility to set the "parent_task_run_id" only once there would be race condition. This problem was described in:
https://github.com/PrefectHQ/prefect/issues/9483

madkinsz:
@billpalombi / @zhen0 

This needs two things
1. A client-side test asserting that this information is provided when `run_deployment` is called from a task
2. A clear path to use of this information in the timeline view

I've mostly been waiting for the latter as it would be confusing to make this change unless it is reflected in the graph.

PR #8798, opened 2023-03-14
Fix blocking file read in `Deployment.load_from_yaml`

madkinsz:


src/prefect/deployments.py
```diff
@@ -499,25 +499,24 @@ def handle_openapi_schema(cls, value):
...
-                    data["storage"] = block
-
-            if data["infrastructure"]:
-                block_doc_name = data["infrastructure"].get("_block_document_name")
-                # if no doc name, this block is not stored on the server
-                if block_doc_name:
-                    block_slug = data["infrastructure"]["_block_type_slug"]
-                    block = await Block.load(f"{block_slug}/{block_doc_name}")
-                    data["infrastructure"] = block
+        data = yaml.safe_load(await anyio.Path(path).read_bytes())
```
madkinsz, line 502: Not only should we not be performing blocking IO here, we should also not keep the file open while loading blocks from the API.

serinamarie (approved):

PR #8806, opened 2023-03-15
Cleanup fixture for hosted API server

madkinsz:
Renames to exclude `orion` and uses `unused_tcp_port` fixture to find a free port.

A factory is used (https://pytest-asyncio.readthedocs.io/en/latest/reference/fixtures.html#unused-tcp-port-factory) since we're using a session scoped fixture.


desertaxle (approved):
LGTM!

PR #8808, opened 2023-03-16
Enhancement: Call `str` instead of `repr` on UnpersistedResult descriptions

znicholasbrown:
### Example

This PR changes how we stringify unpersisted result types by calling `str` instead of `repr` in the description and adding backticks (`). This creates a markdown-enabled string and removes the single quotes that python automatically wraps the type with otherwise.

**Before:**
 
Screen Shot 2023-03-15 at 10 08 13 PM

**After:**

Screen Shot 2023-03-15 at 10 08 08 PM

### Checklist

### DIFF
--- a/src/prefect/results.py
+++ b/src/prefect/results.py
@@ -369,7 +369,7 @@ class UnpersistedResult(BaseResult):
         obj: R,
         cache_object: bool = True,
     ) -> "UnpersistedResult[R]":
-        description = f"Unpersisted result of type `{type(obj).__name__!s}`"
+        description = f"Unpersisted result of type `{type(obj).__name__}`"
         result = cls(
             artifact_type="result",
             artifact_description=description,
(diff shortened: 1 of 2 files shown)


src/prefect/results.py
```diff
@@ -369,9 +369,10 @@ async def create(
         obj: R,
         cache_object: bool = True,
     ) -> "UnpersistedResult[R]":
+        description = f"Unpersisted result of type `{type(obj).__name__!s}`"
```
madkinsz, line 372: We'll want to do this for persisted results too. `!s` is the default btw, we can omit that.

znicholasbrown: TIL!

znicholasbrown: Good callout @madkinsz, I hadn't looked at that class. Updated that and also added a markdown link to the description instead of just the bare url

madkinsz (approved):

Issue #8825, opened 2023-03-16
`propose_state` can reach recursion depth on `Wait`


jakekaplan:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When receiving `SetStateStatus.WAIT` from the server it is possible to reach max recursion depth in `propose_state`. https://github.com/PrefectHQ/prefect/blob/main/src/prefect/engine.py#L1884-L1895

A possible case is reproduced below. Where a lot of tasks are being sent but a concurrency limit is keeping the tasks from being executed in a timely manner. The server will send a `Wait` response if a concurrency slot is not available and tell the task to wait 30 seconds before trying again. If the concurrency limit is low but the number of tasks is high, a task can eventually hit the recursion depth.

- We should look to change this to a loop as opposed to a recursive function
- It may be helpful to add an additional exponential backoff on top of whatever amount of time the server is instructing to wait

### Reproduction

```bash
$ prefect concurrency-limit create "my-tag" 1
```

```python
from prefect import flow, task


@task(tags=["my-tag"])
def print_num_and_sleep(num):
    print(num)

@flow
def map_flow(nums):
    print_num_and_sleep.map(nums)

map_flow([i for i in range(100000)]) # or some other high number
```

### Error

```python3
Crash details:

Traceback (most recent call last):

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1722, in report_task_run_crashes

    yield

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1351, in begin_task_run

    state = await orchestrate_task_run(

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1471, in orchestrate_task_run

    state = await propose_state(

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1890, in propose_state

    return await propose_state(

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1890, in propose_state

    return await propose_state(

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1890, in propose_state

    return await propose_state(

  [Previous line repeated 2947 more times]

  File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 1857, in propose_state

[...]

PR #8827, opened 2023-03-16
remove recursion from `propose_state`

jakekaplan:
closes: https://github.com/PrefectHQ/prefect/issues/8825

Previously `propose_state` made a recursive call to itself when it received a `SetStateStatus.WAIT` as response. In extreme cases where a flow ran with a lot of tasks and a very low concurrency limit, the server could instruct a task to keep waiting and making new recursive `propose_state` calls until the maximum recursion depth is exceeded.

This pr moves the recursive call into a loop instead.

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -32,7 +32,7 @@ from typing_extensions import Literal
 import prefect
 import prefect.context
 from prefect.client.orchestration import PrefectClient, get_client
-from prefect.client.schemas import FlowRun, TaskRun
+from prefect.client.schemas import FlowRun, OrchestrationResult, TaskRun
 from prefect.client.utilities import inject_client
 from prefect.context import (
     FlowRunContext,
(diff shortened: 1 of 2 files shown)


src/prefect/engine.py
```diff
@@ -1852,19 +1852,24 @@ async def propose_state(
 
         link_state_to_result(state, result)
 
+    async def set_state_and_wait(set_state_func) -> OrchestrationResult:
```
madkinsz, line 1855: Maybe `set_state_and_handle_waits`? Can you add a docstring here that says we're doing this to prevent recursion depth

madkinsz: Is it feasible to add a regression test that mocks sleep and has the server return a WAIT instruction 3000 times or w/e?

jakekaplan: !! I did not commit my test file. but that sounds a lot better than what I had, which was just checking it could handle a couple waits

jakekaplan: Updated and added tests that will WAIT recursion depth + buffer number of times

madkinsz (approved):

PR #8830, opened 2023-03-17
Improve concurrency tooling API clarity

madkinsz:


src/prefect/_internal/concurrency/api.py
```diff
@@ -34,51 +36,55 @@ def call_soon_in_global_thread(
...
+        runner.submit(call)
+        return call
 
     @staticmethod
-    def send_callback(call: Call, timeout: Optional[float] = None) -> Call:
+    def call_soon_in_waiter_thread(
+        call: Call[T], timeout: Optional[float] = None
+    ) -> Call[T]:
+        """
+        Schedule a call for execution the thread that is waiting for the current call.
```
jakekaplan, line 84:
```suggestion
        Schedule a call for execution in the thread that is waiting for the current call.
```

src/prefect/_internal/concurrency/threads.py
```diff
@@ -38,15 +39,17 @@ def start(self):
         """
         Start the worker thread.
         """
-        self.thread.start()
+        with self._lock:
+            if not self._started:
```
jakekaplan, line 43: If these are going to get re-used might be worthwhile to do "check, lock, check", to avoid the overhead of acquiring the lock.
```suggestion
    if not self._started:
        with self._lock:
            if not self._started:
```

jakekaplan (approved):
Looks great to me!

src/prefect/_internal/concurrency/threads.py
madkinsz: I don't think it should be called often. We do a guard like this in `submit` which is the most likely place for it to be called repeatedly https://github.com/PrefectHQ/prefect/pull/8830/files#diff-18ed573092916f8235b2a00fa32dc36b8aa5ca168e695d1e80038c561e59ae8dL50-R54

PR #8832, opened 2023-03-17
Add interface for artifacts

serinamarie:
### Overview
We start with a few types of artifacts that we'd like to be able to render in the UI. We can create link and markdown artifacts, both of which will be rendered as markdown. We also have a table type, if for example, a data analyst wanted to quickly scan for changes in their data distribution, or if a data scientist wanted to view their model's accuracy over time. 

### Usage
Enable this experimental feature by adding this setting to your profile like so: 
```bash
prefect config set PREFECT_EXPERIMENTAL_ENABLE_ARTIFACTS=True
```

#### Creating a table artifact
```python
from prefect.artifacts import create_table_artifact

my_table = [{"a": 1, "b": 2}, {"a": 3, "b": 4}]

artifact_id = create_table_artifact(
    key="my-table-artifact",
    table=my_table,
    description="My very important k/v store",
)
```
#### Create a link artifact in a task
```python
from prefect.artifacts import create_link_artifact

@task
def my_special_task():
    artifact_id = create_link_artifact(
        key="my-important-link",
        link="s3://my-bucket/my-key",
        link_text="s3-bucket"
        description="This creates a link to my important bucket.",
    )
    return artifact_id

@flow
def my_flow():
    my_special_task()

if __name__ == "__main__":
    my_flow()
```
#### Create a markdown artifact in a flow
```python
from prefect import flow
from prefect.artifacts import create_markdown_artifact

@flow
def my_flow():
    create_markdown_artifact(
        key="important-markdown",
        markdown="# List of important ducks: 1. Marvin",
        description="Everything you need to know",
    )


if __name__ == "__main__":
    my_flow()
```

### Checklist


src/prefect/artifacts.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    if isinstance(artifact_id, str):
+        artifact_id = UUID(artifact_id)
+
+    return await client.read_artifact(artifact_id=artifact_id)
+
+
+@sync_compatible
+async def read_markdown(
+    artifact_id: Union[str, UUID],
```
madkinsz, line 166: Shouldn't users be fetching artifacts by key not id?

madkinsz: It probably doesn't make sense to expose different `read_` methods. We can parse the artifact based on the type that we read from the API.

src/prefect/artifacts.py
```diff
@@ -0,0 +1,201 @@
...
+        description=description,
+        data=formatted_link,
+        metadata=metadata,
+    )
+
+    return artifact.id
+
+
+@sync_compatible
+async def create_markdown(
```
madkinsz, line 95: We may want to be explicit i.e. `create_markdown_artifact`, `create_link_artifact`, `create_table_artifact`

src/prefect/artifacts.py
serinamarie: There will be many artifacts with the same key, so we could always return the latest artifact on that key? It seems less specific. There could be multiple ways to retrieve artifacts?

src/prefect/artifacts.py
```diff
@@ -0,0 +1,201 @@
...
+
+    return artifact.id
+
+
+@sync_compatible
+async def create_table(
+    table: Union[Dict[str, List[Any]], List[Dict[str, Any]]],
+    name: Optional[str] = None,
+    description: Optional[str] = None,
+    metadata: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
serinamarie: Oh yes, that's right 👍

src/prefect/artifacts.py
serinamarie: That would definitely keep the v1 parity 👍

src/prefect/artifacts.py
serinamarie: I believe we would need a new model method "read_artifact_by_key" which would require adding the ArtifactCollection table which doesn't exist in oss in order to to optimally retrieve the (key, id) we want. Perhaps this PR should just include creating artifacts, and a follow-up PR would get oss on par with nebula before a third PR would add the read interface.

madkinsz: We don't need to have retrieval be the optimized version to add it here.

billpalombi (commented):
The create methods look good! Looks like reading is implemented as well. I'm not sure if thats the primary focus here, but the primary way to read an artifact should be its key.

src/prefect/artifacts.py
billpalombi: I'm agnostic to the order in which the functionality is merged, but I agree with @madkinsz that the key should be the primary way to retrieve the artifact.

madkinsz (commented):
Can you write documentation for this too please?

billpalombi:
> > The create methods look good! Looks like reading is implemented as well. I'm not sure if thats the primary focus here, but the primary way to read an artifact should be its key.
> 
> @billpalombi, reading is just implemented here to make it easier to test. However, I have now removed all references to reading artifacts from the client and changed the tests accordingly.

Sounds good, thank you!

src/prefect/artifacts.py
```diff
@@ -0,0 +1,136 @@
...
+    if data is not None:
+        artifact_args["data"] = data
+
+    artifact = ArtifactCreate(**artifact_args)
+
+    return await client.create_artifact(artifact=artifact)
+
+
+@sync_compatible
+async def create_link_artifact(
```
madkinsz, line 61: Are these missing an experimental flag?

serinamarie: Right now you would need adjust your `PREFECT_EXPERIMENTAL_ENABLE_ARTIFACTS` setting to True to use these.

serinamarie: Or do you mean something else?

serinamarie:
> Can you write documentation for this too please?

I think they can be in a followup PR

madkinsz:
@serinamarie I have a strong preference for including documentation alongside adding features — an undocumented feature is not visible to the users. Is there a strong motivation for adding it later?

billpalombi (approved):
This is good to go from a docs and UX perspective. Thank you @serinamarie!

zhen0 (approved):
Since this is a combination of two previously reviewed PRs it looks good to me.

zhen0 (changes requested):
Sorry! One docs change incoming

docs/concepts/artifacts.md
```diff
@@ -0,0 +1,227 @@
...
+        'customer_id': ['123456789', '987654321', '246810121', '135791113', '864208046'],
+        'name': ['John Smith', 'Jane Doe', 'Bob Johnson', 'Sarah Jones', 'Tom Wilson'],
+        'churn_probability': [0.85, 0.79, 0.67, 0.61, 0.57]
+    }
+
+    create_table_artifact(
+        key="personalized-reachout",
+        table=highest_churn_possibility,
+        description= "# Marvin, please reach out to these customers today!"
+    )
```
serinamarie, line 168:
```suggestion
def my_fn():
    highest_churn_possibility = [
       {'customer_id':'12345', 'name': 'John Smith', 'churn_probability': 0.85 }, 
       {'customer_id':'56789', 'name': 'Jane Jones', 'churn_probability': 0.65 } 
    ]

    create_table_artifact(
        key="personalized-reachout",
        table=highest_churn_possibility,
        description= "# Marvin, please reach out to these customers today!"
    )
```

PR #8838, opened 2023-03-17
Check if block type needs updating first on register

jakekaplan:
closes: https://github.com/PrefectHQ/nebula/issues/4064

Every time a block is registered it does so by calling `Block.register_type_and_schema`

For example this call would create a `DateTime` `BlockType` on the server if it did not exist.
```python
block = DateTime(value=datetime(2022, 11, 5))
block.save("my-date-block") # <--invokes `register_type_and_schema`
```
However any subsequent calls would also update the block type, even if none of the parameters of `BlockType` were different. 
```python
block = DateTime(value=datetime(2022, 11, 5))
block.save("my-date-block") # <-- initial `BlockType` creation if `DateTime` didn't exist server side
other_block = DateTime(value=datetime(2022, 11, 5))
other_block.save("my-other-date-block") # <-- would update values `DateTime` even though they haven't changed. 
```

Server side this creates a problem when creating a lot of `Block` of the same type. Imagine creating 1000 blocks of the same type concurrently. That's 1000 individual queries that are attempting to access 1 row. This leads to a lot of the updates timing out trying to contend for the same single row.

Most of the time there is no need to touch the `BlockType` on the server if it already exists. This pr adds a method `_should_update_block_type` that will only update the `BlockType` if any of the updatable values are different.

**NOTE:** I am not positive we should be updating the `BlockType` at all here, but I don't want to break expectations of block's auto registering/updating by changing it.

### Checklist


jakekaplan:
Moving this back to draft for now, as we may take a different approach

desertaxle (commented):
We also discussed adding a server-side check to account for older clients. Is the plan to implement that as a separate PR?

jakekaplan:
@desertaxle good call. I was thinking about adding to Cloud as the server, but added here for self hosted cases

PR #8839, opened 2023-03-17
Add jitter to Prefect client request retries

madkinsz:
Adds jitter to the Prefect base client when retrying requests.

Uses a Poisson interval with a shifted average for responses with "Retry-After" headers to ensure we never retry sooner than requested. Otherwise, we use a Poisson interval with an average matching our previous exponential back-off. A setting is added to control the amount of jitter (can be disabled with a value of 0) which defaults to +/- 20%.

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -516,6 +516,15 @@ If the API does not support HTTP/2, this will have no effect and connections wil
 made via HTTP/1.1.
 """
 
+PREFECT_CLIENT_RETRY_JITTER_FACTOR = Setting(float, default=0.2)
+"""
+A value greater than or equal to zero to control the amount of jitter added to retried
+client requests. Higher values introduce larger amounts of jitter.
+
+Set to 0 to disable jitter. See `clamped_poisson_interval` for details on the how jitter
+can affect retry lengths.
+"""
+
 PREFECT_CLOUD_API_URL = Setting(
     str,
     default="https://api.prefect.cloud/api",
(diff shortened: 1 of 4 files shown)


src/prefect/settings.py
```diff
@@ -516,6 +516,15 @@ def default_cloud_ui_url(settings, value):
 made via HTTP/1.1.
 """
 
+PREFECT_CLIENT_RETRY_JITTER_FACTOR = Setting(float, default=0.2)
```
madkinsz, line 519: I'm certain I'll need to toggle this off for some of our existing tests to pass

src/prefect/utilities/math.py
```diff
@@ -55,3 +55,18 @@ def clamped_poisson_interval(average_interval, clamping_factor=0.3):
...
+
+def bounded_poisson_interval(lower_bound, upper_bound):
+    """
+    Bounds Poisson "inter-arrival times" to a range.
+
+    Unlike `clamped_poisson_interval` this does not take a target average interval.
+    Instead, the interval is predetermined and the average is calculated as their
+    midpoint. This allows Poisson intervals to be used in cases where a lower bound
+    must be enforced.
+    """
```
madkinsz, line 68: I must admit: I'm not the mathematician that authored these utilities. This does what I want it to, but I'm not sure if it's mathematically correct. cc @cicdw ?

billpalombi (approved):
Looks like this mirrors the implementation in #7954. No concerns with that experience so the consistency is good.

PR #8840, opened 2023-03-17
Improve handling of concurrent cancellation and timeouts

madkinsz:
Extends https://github.com/PrefectHQ/prefect/pull/8830 cleaning up cancellation tooling added in #8702 to reduce complexity and risk of problems when timeouts are nested.

- Update alarm-based timeouts to avoid overriding existing alarms
- Add more test cases for nested sync timeouts
- Display exception information on debug log
- Update waiters to avoid creating extra cancel scopes
- Use `os.kill` instead of `signal.raise_signal` because otherwise it seems to wait to raise
- Log exceptions encountered during chained cancellation
- Add debug log format with thread names
- Fix async cancellation from across threads
- Add name to sync timeout threads
- Shutdown timeout watcher threads on completion of context
- Add thread leak detection fixtures to tests
- Fix logging errors when shutting down services at interpreter exit
- Only use alarms when a timeout is specified

### Checklist

### DIFF
--- a/src/prefect/_internal/concurrency/timeouts.py
+++ b/src/prefect/_internal/concurrency/timeouts.py
@@ -9,7 +9,6 @@ import os
 import signal
 import sys
 import threading
-import logging
 import time
 from typing import Callable, List, Optional, Type
 
@@ -20,8 +19,7 @@ from prefect.logging import get_logger
 # TODO: We should update the format for this logger to include the current thread
 logger = get_logger("prefect._internal.concurrency.timeouts")
 
-
-logging.logThreads = 0
+_ENFORCERS = {}
 
 
 class CancelledError(asyncio.CancelledError):
@@ -396,6 +394,7 @@ def _watcher_thread_based_timeout(timeout: Optional[float], name: Optional[str] 
             target=timeout_enforcer,
             name=f"timeout-watcher {name or '<unnamed>'} {timeout:.2f}",
         )
+        _ENFORCERS[supervised_thread] = enforcer
         enforcer.start()
 
     try:
@@ -405,7 +404,6 @@ def _watcher_thread_based_timeout(timeout: Optional[float], name: Optional[str] 
         ctx.mark_completed()
         if enforcer:
             enforcer.join()
-            del enforcer
 
 
 def _send_exception_to_thread(thread: threading.Thread, exc_type: Type[BaseException]):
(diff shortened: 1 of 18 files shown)


madkinsz:
There's an issue with `test_async_waiter_timeout_in_main_thread` that needs to be addressed before this is ready for review.

jakekaplan (approved):

madkinsz:
Oh.. apparently something is different on Linux because this hangs now. Will investigate.

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -154,10 +163,15 @@ def __init__(self, deadline: float) -> None:
...
+        self._loop = asyncio.get_running_loop()
 
     def cancel(self):
         self.context.mark_cancelled()
-        return super().cancel()
+        # Cancellation must be called from the event loop that owns the scope
+        if get_running_loop() != self._loop:
+            return call_soon_in_loop(self._loop, super().cancel).result()
+        else:
+            return super().cancel()
```
madkinsz, line 174: Fixes bug where cancellation failed due to call from another loop

src/prefect/_internal/concurrency/timeouts.py
```diff
@@ -297,7 +322,7 @@ def sigalarm_to_error(*args):
             else CancelledError()
         )
 
-    if previous_alarm_handler:
+    if previous_alarm_handler != signal.SIG_DFL:
```
madkinsz, line 325: Fixes bug where a previous alarm handler was always detected

src/prefect/logging/logging.yml
```diff
@@ -20,6 +20,10 @@ formatters:
         task_run_fmt: "%(asctime)s.%(msecs)03d | %(levelname)-7s | Task run %(task_run_name)r - %(message)s"
         datefmt: "%H:%M:%S"
 
+    debug:
+        format: "%(asctime)s.%(msecs)03d | %(levelname)-7s | %(threadName)-12s | %(name)s - %(message)s"
+        datefmt: "%H:%M:%S"
+
```
madkinsz, line 26: Adds a debug handler/format for internal logs that include the thread name!

zhen0:
@madkinsz - is this one waiting for re-review or are there still changes to be made?

madkinsz:
Struggling to reproduce hangs on Python 3.11 locally :(

madkinsz:
Replaced by #9835

PR #8849, opened 2023-03-20
Allow opt-out of server dependencies during installation

madkinsz:
Will require substantial updates to prevent importing required libraries when the server dependencies are not installed.

I'd prefer to use "extras" instead of an environment variable, but there's not support for this case upstream https://discuss.python.org/t/adding-a-default-extra-require-environment/4898

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -1,0 +1,7 @@
+aiosqlite >= 0.17.0
+alembic >= 1.7.5
+asgi-lifespan >= 1.0
+asyncpg >= 0.23
+fastapi >= 0.70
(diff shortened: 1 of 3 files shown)


madkinsz:
This was a POC.

carderne:
Would love to give this idea a huge +1. It seems like there are at least three projects (core, server, cli...) bundled into one. Ref: https://github.com/PrefectHQ/prefect/issues/8052

A less magical alternative would just be to update all the docs to say `pip install prefect[all]` and make a note that once people move to production, they can drop the extra deps...

PR #8851, opened 2023-03-20
Disable coverage display during local pytest runs by default

madkinsz:
This drives me crazy and is not a metric we are using


serinamarie (approved):
sgtm

Issue #7774, opened 2022-12-06
Flow descriptions are not saved into orion database


J-F-Far:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

deploying locally, the flow description attribute does not appear to be saved in the Orion database. This results in the flow details in the UI 'flow' tab always containing 'None' for the description value, despite having both the decorator description attribute set, and a docstring set for the function.

![Screenshot from 2022-12-05 16-33-14](https://user-images.githubusercontent.com/7216850/205757310-d765dcee-fad5-47cd-84f1-aac91a894097.png)

### Reproduction

```python3
# Simple script with one flow, one task that is sufficient to reproduce the issue. 

from prefect import flow, task


@task(name="My Test Task")
def my_task():
    print("In the task")


@flow(name="My Test Flow.",
      description="A minimal flow for verifying description problem updated."
      )
def my_flow():
    """Flow description from docstring."""
    print("In the flow")
    my_task()


if __name__ == '__main__':
    my_flow()
```

### Versions

```Text
2.7.0
```

### Additional context

I have created a branch that could serve as a starting point for fixing this issue. I'll include a bit more detail in a draft pull request.

desertaxle:
Flow descriptions are intentionally stored on deployments rather than on the flow record itself. When you call `flow.deploy()` or `flow.serve()`, the flow's description is automatically passed to the deployment, which is where it's visible in the UI. The `Flow` table in the database is a lightweight reference (name + tags), while deployments carry the rich metadata. For ad-hoc flow runs (calling a flow directly without a deployment), descriptions intentionally aren't persisted to the server. If you have a use case for flow-level descriptions in the UI on a current version of Prefect, please open a new issue.

PR #8862, opened 2023-03-21
Add flow descriptions

madkinsz:
Rebase of https://github.com/PrefectHQ/prefect/pull/7775; see that pull request for discussion and context.

Authored by @J-F-Far

Closes https://github.com/PrefectHQ/prefect/issues/7774

### Checklist

### DIFF
--- a/src/prefect/server/database/migrations/versions/postgresql/2022_12_05_185028_adabad235795_add_description_to_flow.py
+++ b/src/prefect/server/database/migrations/versions/postgresql/2022_12_05_185028_adabad235795_add_description_to_flow.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: adabad235795
+Revises: 4a1a0e4f89de
+Create Date: 2022-12-05 18:50:28.800545
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "adabad235795"
+down_revision = "4a1a0e4f89de"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("flow", sa.Column("description", sa.TEXT(), nullable=True))
+
+
+def downgrade():
+    with op.batch_alter_table("flow", schema=None) as batch_op:
+        batch_op.drop_column("description")
--- a/src/prefect/server/database/migrations/versions/sqlite/2022_12_05_184519_dc729e5c0448_add_description_to_flow.py
+++ b/src/prefect/server/database/migrations/versions/sqlite/2022_12_05_184519_dc729e5c0448_add_description_to_flow.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: dc729e5c0448
+Revises: f3df94dca3cc
+Create Date: 2022-12-05 18:45:19.631375
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "dc729e5c0448"
+down_revision = "f3df94dca3cc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("flow", sa.Column("description", sa.TEXT(), nullable=True))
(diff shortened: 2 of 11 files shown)


src/prefect/server/database/migrations/versions/postgresql/2022_12_05_185028_adabad235795_add_description_to_flow.py
```diff
@@ -0,0 +1,24 @@
+"""empty message
```
serinamarie, line 1: 🕳️

src/prefect/server/database/migrations/versions/sqlite/2022_12_05_184519_dc729e5c0448_add_description_to_flow.py
```diff
@@ -0,0 +1,24 @@
+"""empty message
```
serinamarie, line 1:
```suggestion
"""Add flow description
```

serinamarie (approved):
LGTM! We'd probably just want to add the changes to the migration notes.

cicdw (changes requested):
I don't think we should merge this - flows are emergent entities only, they don't have any meaningful persistence in the backend and this risks a frequently updating field that changes from run-to-run.

zhen0:
Do you mind if I close this one @madkinsz?

madkinsz:
We do set default deployment descriptions to the flow description https://github.com/PrefectHQ/prefect/blob/9c0440eb356483bbc938d9231fef0998c9c0e905/src/prefect/deployments.py#L794 but perhaps the description should only be exposed on the deployment if we're not going to store the flow description. We can still parse the flow's docstring for default descriptions, just in the logic for deployments.

znicholasbrown:
I have no plans for the flow description atm but I _could_ if we want to support it.

billpalombi:
I just discussed this with @cicdw. Flows should have descriptions. Its useful for users. However, we don't want to encourage users to create/update flow descriptions with each flow run, so lets drop the `description` parameter from the `@flow` decorator and only allow flow descriptions to be created/updated via the UI/API.

madkinsz:
> I have no plans for the flow description atm but I could if we want to support it.

Interestingly, the deployment description is probably always an exact duplicate right now. How are we going to handle the presence of both? If we deprecate `Flow.description` client-side then I guess it'd be empty until a user updates it in the UI.

> However, we don't want to encourage users to create/update flow descriptions with each flow run, so lets drop the description parameter from the @flow decorator and only allow flow descriptions to be created/updated via the UI/API.

Without this pull request, there is no flow description in the backend. We can deprecate the `Flow.description` attribute and `@flow(description=...)` here but keep the backend changes. I'd like to understand how we intend to expose / display this in the UI before putting that work in though.

billpalombi:
> the deployment description is probably always an exact duplicate right now.

I'm sorry, I didn't follow, what is it a duplicate of?

> If we deprecate Flow.description client-side then I guess it'd be empty until a user updates it in the UI.

Yes, exactly.

> I'd like to understand how we intend to expose / display this in the UI before putting that work in though.

That's totally reasonable. I can mock that up.

I think it makes sense to break this into at least two seperate steps:
1. Deprecate Flow.description client-side
2. Add a description to the UI/server-side

madkinsz:
> I'm sorry, I didn't follow, what is it a duplicate of?

The deployment description currently defaults to the flow description which defaults to the flow function's docstring; so if we turned on flow descriptions naively we'd have the same description (the flow docstring) displayed twice. The problem with updating deployment descriptions to read the flow function's docstring by default and removing a default flow description is that then each deployment (for a given flow) will have a duplicate default description. The default value (the docstring of a flow function) is a very reasonable flow description but it makes much less sense for deployments.

billpalombi:
Ah I see, thank you for the explainer. Yeah I agree that it doesn't make sense to use the docstring of a flow function for the deployment description. It sounds like our current setup is a bit of a hack.

PR #8867, opened 2023-03-21
Create `emit_event` helper that takes args for an `Event` and emits it via a worker

bunchesofdonald:
This creates the `emit_event` helper which takes args for an `Event` and emits it via a events worker.

### Example
```python
from prefect.events import emit_event

emit_event(
    event="vogon.poetry.read",
    resource={"prefect.resource.id": "vogon.poem.oh-freddled-gruntbuggly"},
)
```

### Checklist

### DIFF
--- a/src/prefect/events/utilities.py
+++ b/src/prefect/events/utilities.py
@@ -1,0 +1,51 @@
+from typing import Any, Dict, List, Optional
+from uuid import UUID
+
+from prefect.server.utilities.schemas import DateTimeTZ
+
+from .schemas import Event
+from .worker import get_events_worker
+
+
+def emit_event(
+    event: str,
+    resource: Dict[str, str],
+    occurred: Optional[DateTimeTZ] = None,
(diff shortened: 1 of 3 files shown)


madkinsz:
Is this waiting on anything?

src/prefect/events/utilities.py
```diff
@@ -0,0 +1,51 @@
...
+    if payload is not None:
+        event_kwargs["payload"] = payload
+
+    if id is not None:
+        event_kwargs["id"] = id
+
+    event_obj = Event(**event_kwargs)
+
+    with get_events_worker() as worker:
+        worker.emit(event_obj)
```
chrisguidry, line 51: I believe this is the case: if `get_events_worker` finds an existing worker, it will reuse it and not close it when the context exits, right?

chrisguidry (approved):
Looks great!  Just clarifying that lifecycle bit...

PR #8872, opened 2023-03-21
Add preparation steps even to empty infra block setup

cicdw:
This PR updates agents to prepare their own default infrastructure in the case where the deployment does not have an infrastructure block configured.  I came across this while working on backwards compatibility issues with projects.

I believe I could also adjust this code to apply infra overrides, but wanted to break it down into smaller pieces for review.

### Checklist

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -383,12 +383,18 @@ class PrefectAgent:
         # overrides only apply when configuring known infra blocks
         if not deployment.infrastructure_document_id:
             if self.default_infrastructure:
-                return self.default_infrastructure
+                infra_block = self.default_infrastructure
             else:
                 infra_document = await self.client.read_block_document(
                     self.default_infrastructure_document_id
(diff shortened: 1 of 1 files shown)


src/prefect/agent.py
```diff
@@ -383,12 +383,18 @@ async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
...
                     self.default_infrastructure_document_id
                 )
-                return Block._from_block_document(infra_document)
+                infra_block = Block._from_block_document(infra_document)
+
+            # Add flow run metadata to the infrastructure
+            prepared_infrastructure = infra_block.prepare_for_flow_run(
+                flow_run, deployment=deployment, flow=flow
+            )
+            return prepared_infrastructure
```
madkinsz, line 397: Should we bypassing infrastructure overrides still?

madkinsz (approved):

PR #8874, opened 2023-03-21
Update create & delete methods, add read_latest_artifact method

serinamarie:
OSS implementation of: https://github.com/PrefectHQ/nebula/pull/3975

Alongside the new `artifact_collection` table comes updated `create_artifact` and `delete_artifact` model methods to ensure upserting and deleting extends to this table (where the latest of each artifact is stored).

In `create_artifact`, we'll first upsert to the `artifact_collection` table, ensuring we keep track of the latest artifact version id. In `delete_artifact`, we now need to check if the artifact is the latest version. If it is, we'll need to update our reference to it in the `artifact_collection` table prior to deletion from the `artifact` table. If there are no prior versions of the artifact (there's just 1 version in the `artifact` table), then the row itself related to the artifact will be deleted from the `artifact_collection` table.

Also adding a `read_latest_artifact` method, so that we can add a new route for this to optimally retrieve the latest of each artifact on a given key!

### Checklist

### DIFF
--- a/src/prefect/server/models/artifacts.py
+++ b/src/prefect/server/models/artifacts.py
@@ -11,15 +11,75 @@ from prefect.server.schemas.core import Artifact
 
 
 @inject_db
-async def create_artifact(
-    session: sa.orm.Session, artifact: Artifact, db: PrefectDBInterface, key: str = None
+async def _insert_into_artifact_collection(
+    session: sa.orm.Session,
+    key: str,
+    artifact_id: UUID,
+    db: PrefectDBInterface,
+    now: pendulum.DateTime = None,
 ):
-    now = pendulum.now("UTC")
+    """
+    Inserts a new artifact into the artifact_collection table or updates it.
+    """
+    upsert_new_latest_id = (
+        (await db.insert(db.ArtifactCollection))
+        .values(
+            key=key,
+            latest_id=artifact_id,
+            created=now,
+            updated=now,
+        )
+        .on_conflict_do_update(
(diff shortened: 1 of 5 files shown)


jakekaplan (approved):
LGTM! most of the similar comments on https://github.com/PrefectHQ/nebula/pull/3973 apply here as well

src/prefect/server/models/artifacts.py
```diff
@@ -36,6 +96,57 @@ async def create_artifact(
...
+        .where(db.ArtifactCollection.key == key)
+        .limit(1)
+    )
+    latest_id = await session.execute(query)
+    latest_id_scalar = latest_id.scalar()
+
+    query = sa.select(db.Artifact).where(db.Artifact.id == latest_id_scalar)
+    result = await session.execute(query)
+
+    return result.scalar()
```
serinamarie, line 147: @jakekaplan Separated into 2 queries here 👍

src/prefect/server/models/artifacts.py
```diff
@@ -36,6 +96,59 @@ async def create_artifact(
...
+    )
+    latest_id = await session.execute(latest_id_query)
+    latest_id_scalar = latest_id.scalar()
+
+    latest_artifact_query = sa.select(db.Artifact).where(
+        db.Artifact.id == latest_id_scalar
+    )
+    result = await session.execute(latest_artifact_query)
+
+    return result.scalar()
```
serinamarie, line 149: @jakekaplan Separated into 2 queries here 👍

Issue #8878, opened 2023-03-22
Pause/Resume doesn't work post 2.8.6


EmilRex:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Pausing a flow with `pause_flow_run` crashes when installing from main (`g1897139b7`). I'm guessing this is related to the #8702.

### Reproduction

```python3
from prefect import flow, task, pause_flow_run

@task
def my_task():
    return 2 * 2

@flow
def pause_test():
    my_task()
    pause_flow_run()
    my_task()

if __name__ == "__main__":
    pause_test()
```

### Error

```python3
(default) ➜  examples python pause.py                                              
10:49:19.181 | INFO    | prefect.engine - Created flow run 'jovial-orca' for flow 'pause-test'
10:49:20.201 | INFO    | Flow run 'jovial-orca' - Created task run 'my_task-0' for task 'my_task'
10:49:20.202 | INFO    | Flow run 'jovial-orca' - Executing 'my_task-0' immediately...
10:49:20.771 | INFO    | Task run 'my_task-0' - Finished in state Completed()
10:49:20.774 | INFO    | Flow run 'jovial-orca' - Pausing flow, execution will continue when this flow run is resumed.
10:49:20.779 | ERROR   | Flow run 'jovial-orca' - Encountered exception during execution:
Traceback (most recent call last):
  File "/Users/emilchristensen/.virtualenvs/default/lib/python3.9/site-packages/anyio/streams/tls.py", line 130, in _call_sslobject_method
    result = func(*args)
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ssl.py", line 889, in read
    v = self._sslobj.read(len)
ssl.SSLWantReadError: The operation did not complete (read) (_ssl.c:2633)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/emilchristensen/.virtualenvs/default/lib/python3.9/site-packages/prefect/engine.py", line 671, in orchestrate_flow_run
    result = await flow_call.aresult()
  File "/Users/emilchristensen/.virtualenvs/default/lib/python3.9/site-packages/prefect/_internal/concurrency/calls.py", line 181, in aresult
[...]

PR #8879, opened 2023-03-22
Fix flow run pauses

madkinsz:
The client was not being injected into these methods correctly which caused it to be pulled from the context when it should not have been. An integration test for this feature has been added.

Closes https://github.com/PrefectHQ/prefect/issues/8878


serinamarie (approved):
LGTM! 🔥 QA was successful

Edit: Not including Emil's issue above after resumption of flow run.

PR #8882, opened 2023-03-22
Fix sleep time assertion

madkinsz:
This assertion was still < 1s instead of the variable sleep time constant which caused it to flake in CI


abrookins (approved):
LGTM 👍

PR #8885, opened 2023-03-22
Show the full run history in ui

pleek91:
# Description

Updates everywhere we're displaying lists of flow runs to use the full history rather than just the last 7 days. 

Needs https://github.com/PrefectHQ/prefect-ui-library/pull/1245

### DIFF
--- a/ui/src/pages/Deployment.vue
+++ b/ui/src/pages/Deployment.vue
@@ -37,11 +37,7 @@
       </template>
 
       <template #runs>
-        <FlowRunFilteredList :flow-run-filter="deploymentFilter">
-          <template #empty-message>
-            No runs found
-          </template>
-        </FlowRunFilteredList>
+        <FlowRunFilteredList :flow-run-filter="deploymentFilter" />
       </template>
     </p-tabs>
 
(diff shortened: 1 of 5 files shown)


madkinsz:
Will this be ready for release on the 23rd?

ui/src/pages/Deployment.vue
```diff
@@ -37,11 +37,7 @@
       </template>
 
       <template #runs>
-        <FlowRunFilteredList :flow-run-filter="deploymentFilter">
-          <template #empty-message>
-            No runs found
-          </template>
-        </FlowRunFilteredList>
+        <FlowRunFilteredList :flow-run-filter="deploymentFilter" />
```
znicholasbrown, line 40: Can you make one additional change here and use this same component on the `FlowRuns.vue` page as well? I think we're potentially using the wrong one there and it's actually not loading more data like this one does.

pleek91: So I agree we need to address that. But I don't want to do that in this PR. Mainly because the flow runs page has a refresh interval. Which we haven't implemented in the `FlowRunFilteredList` or anywhere we use list paging that I'm aware of.

znicholasbrown (approved):

PR #8886, opened 2023-03-22
Fix race condition in worker thread start

madkinsz:
Fixes race condition where thread.start() would return but not set the thread as started until a bit later. See 'https://github.com/PrefectHQ/prefect/actions/runs/4494685885/jobs/7907472976?pr=8882'


abrookins (approved):
LGTM 👍

PR #8896, opened 2023-03-23
Idea: Add optional server-only requirements

carderne:
Prompted by: https://github.com/PrefectHQ/prefect/issues/8052

The very beginnings of a way to have more optional requirements.
Obviously this would require many changes to documentation, and probably more libs should be made optional, and likely another set of requirements (such as `-cli` would be needed).

Would be great for production flows to only need to bare minimum requirements, as this would make our Docker images lighter too.

### Example
```bash
pip install prefect[server]
```

### Checklist

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -1,0 +1,1 @@
+sqlalchemy[asyncio] >= 1.4.22, != 1.4.33, < 2.0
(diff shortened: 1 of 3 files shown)


madkinsz:
Hi! We cannot make these requirements optional like this as they are required for new users when we run the API in memory. See #8849 instead.

Issue #8894, opened 2023-03-23
Allow overriding the default `PersistedResult` key passed to `storage_block.write`


limx0:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently, the path that is passed to a storage block for writing is simply a uuid: 

```python
# PersistedResult ~ L493

key = uuid.uuid4().hex
await storage_block.write_path(key, content=blob.to_bytes())
```

### Describe the proposed behavior

While prefect takes care of mapping task names to these keys in the database, it would be great if this default value could be overridden for the purposes of some organisation on the backend storage block.

Either some sort of callable or (prefect1) target-like str formatting would be great.

### Example Use

```python
@task(
    persist_result=True, 
    cache_key_fn=task_input_hash, 
    result_storage=WritableFileSystem(target="{task_name}/{n}")
)
def get(n):
    return n
```

billpalombi:
Thanks for submitting this! We agree. We have some ideas about how to address this need.

PR #8580, opened 2023-02-17
POC: Expose custom persisted result types

madkinsz:
Demonstrates the addition of a `result_cls` option to tasks, allowing users to use a custom subclass of `PersistedResult` to perform arbitrary persistence behavior.

### Example

```python
from prefect import flow, task
from prefect.results import (
    PersistedResult,
    WritableFileSystem,
    Serializer,
    R,
)
from typing import Type
import uuid
from prefect.serializers import JSONSerializer


class CustomResult(PersistedResult):
    """
    This differs from the default `PersistedResult` in that the data written does not
    have the serializer inlined; instead the serializer config is stored in the Prefect
    database.
    """

    serializer: Serializer
    type = "custom-result"

    @classmethod
    async def create(
        cls: "Type[CustomResult]",
        obj: R,
        storage_block: WritableFileSystem,
        storage_block_id: uuid.UUID,
        serializer: Serializer,
        cache_object: bool = True,
    ) -> "CustomResult[R]":
        """
        Create a new result reference from a user's object.
        """
        data = serializer.dumps(obj)

        key = uuid.uuid4().hex

        await storage_block.write_path(key, content=data)
        print(f"Wrote to {storage_block.basepath}/{key}")

        description = cls._infer_description(obj, storage_block, key)

        result = cls(
            serializer=serializer,
            serializer_type=serializer.type,
            storage_block_id=storage_block_id,
            storage_key=key,
            artifact_type="result",
            artifact_description=description,
        )

        if cache_object:
            # Attach the object to the result so it's available without deserialization
            result._cache_object(obj)

        object.__setattr__(result, "_should_cache_object", cache_object)

        return result


@flow
def example():
    my_task()


@task(persist_result=True, result_cls=CustomResult, result_serializer=JSONSerializer())
def my_task():
    return 1


example()
```
```
11:52:26.002 | DEBUG   | prefect.client - Using ephemeral application with database at sqlite+aiosqlite:////Users/mz/.prefect/prefect.db
11:52:26.066 | INFO    | prefect.engine - Created flow run 'lean-cassowary' for flow 'example'
11:52:26.067 | DEBUG   | Flow run 'lean-cassowary' - Starting 'ConcurrentTaskRunner'; submitted tasks will be run concurrently...
11:52:26.067 | DEBUG   | prefect.task_runner.concurrent - Starting task runner...
11:52:26.203 | DEBUG   | Flow run 'lean-cassowary' - Executing flow 'example' for flow run 'lean-cassowary'...
[...]

PR #8924, opened 2023-03-24
Add support for custom static task storage keys

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/8894

New documentation at https://deploy-preview-8924--prefect-docs.netlify.app/concepts/results/#result-storage-key

Only allows static keys at this time. A follow-up can add support for templating keys with task run parameters by default. An additional follow-up can add support for providing a callable directly instead of a string.

Could also be accomplished by #8580 but I think that's a bit too low level to do first.

### Example

```python
from prefect.serializers import JSONSerializer

@task(persist_result=True, result_storage_key="foo.json")
def hello_world():
    return "hello world"


@flow(result_serializer=JSONSerializer())
def my_flow():
    hello_world()


my_flow()
```

```
❯ cat ~/.prefect/storage/foo.json | jq
{
  "serializer": {
    "type": "json",
    "jsonlib": "json",
    "object_encoder": "prefect.serializers.prefect_json_object_encoder",
    "object_decoder": "prefect.serializers.prefect_json_object_decoder",
    "dumps_kwargs": {},
    "loads_kwargs": {}
  },
  "data": "\"hello world\"",
  "prefect_version": "2.8.7+11.gfb8822153"
}
```

### Checklist

### DIFF
--- a/docs/concepts/results.md
+++ b/docs/concepts/results.md
@@ -399,6 +399,30 @@ You can configure this to use a specific storage using one of the following:
 - A storage instance, e.g. `LocalFileSystem(basepath=".my-results")`
 - A storage slug, e.g. `'s3/dev-s3-block'`
 
+#### Result storage key
+
+The path of the result file in the result storage can be configured with the `result_storage_key`. The `result_storage_key` option defaults to a null value, which generates a unique identifier for each result.
+
+
+```python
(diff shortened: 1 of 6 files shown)


desertaxle (approved):
LGTM!

PR #8928, opened 2023-03-24
Add ability to filter artifacts by whether key exists

serinamarie:
### Checklist

### DIFF
--- a/src/prefect/server/schemas/filters.py
+++ b/src/prefect/server/schemas/filters.py
@@ -1524,6 +1524,25 @@ class ArtifactFilterTaskRunId(PrefectFilterBaseModel):
         return filters
 
 
+class ArtifactFilterType(PrefectFilterBaseModel):
+    """Filter by `Artifact.type`."""
+
+    any_: Optional[List[str]] = Field(
+        default=None, description="A list of artifact types to include"
+    )
+    not_any_: Optional[List[str]] = Field(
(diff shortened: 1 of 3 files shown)


zhen0:
I think the counterpart for this actually https://github.com/PrefectHQ/nebula/pull/4208

src/prefect/server/schemas/filters.py
```diff
@@ -1487,12 +1487,22 @@ class ArtifactFilterKey(PrefectFilterBaseModel):
...
+        ),
+    )
+
     def _get_filter_list(self, db: "PrefectDBInterface") -> List:
         filters = []
         if self.any_ is not None:
             filters.append(db.Artifact.key.in_(self.any_))
         if self.like_ is not None:
             filters.append(db.Artifact.key.ilike(f"%{self.like_}%"))
+        if self.exists_ is True:
```
zangell44, line 1504: would we ever want to explicitly query for artifacts without a key set?

zangell44 (approved):

src/prefect/server/schemas/filters.py
serinamarie: I'm pretty sure no, but @znicholasbrown do you disagree? We also have a `type` filter to get results (which don't have a key).

serinamarie: Both are now allowed 👍

PR #8931, opened 2023-03-24
Feature: Gzip middleware API and UI responses

znicholasbrown:
This PR adds Gzip middleware to the UI and API FastAPI apps for compressing responses.

### Example
This script runs a flow that accepts a string input and passes a string of ~20mb:

```py
from prefect import flow

def create_string_of_size_x_bytes(size_in_bytes):
    return "*" * size_in_bytes

@flow
def massive_input(input_str: str):
    pass

if __name__ == "__main__":
    input_str = create_string_of_size_x_bytes(20000000)
    massive_input(input_str=input_str)
```

Testing these changes with and without an `Accept-Encoding: gzip` header:
Screenshot 2023-03-24 at 6 42 27 PM

The gzipped response has a compressed size of ~20kb, compared to the uncompressed ~19mb.

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -16,6 +16,7 @@ from fastapi import APIRouter, Depends, FastAPI, Request, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.middleware.gzip import GZipMiddleware
 from fastapi.openapi.utils import get_openapi
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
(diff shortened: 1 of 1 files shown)


src/prefect/server/api/server.py
```diff
@@ -186,6 +187,7 @@ def create_orion_api(
     """
     fast_api_app_kwargs = fast_api_app_kwargs or {}
     api_app = FastAPI(title=API_TITLE, **fast_api_app_kwargs)
+    api_app.add_middleware(GZipMiddleware)
```
madkinsz, line 190: Is this something we just want to add to all API routes? I don't know if it'll get us much of a benefit for many cases. Do we do this in Cloud?

znicholasbrown: It doesn't hurt - clients have full control over how data is returned so clients that can't/don't want to handle compression can just ignore the `Accept-Encoding` header. This is a boon for browser clients especially because they send that header and handle compression automatically. 

To your second question: no, not yet. But soon 😄

znicholasbrown: Oh also it should be noted that the `GZipMiddleware` has a min compression size of 500 bytes so it'll ignore payloads smaller than that. We could bump that if we think it's not worth it for a larger response body, I'm just not sure what that magic number is

madkinsz (approved):

Issue #8895, opened 2023-03-23
[Review] Accept `functions` and `objects` to `flow_run_name` and `task_run_name`


rsampaths16:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently we can set a `flow_run_name` for a flow run and a `task_run_name` for a task run via template strings. The template string is limited to the parameters passed.

### Describe the proposed behavior

Accept a `Callable[Any] -> str` for `flow_run_name` and `task_run_name`. It will enable us to provide some dynamisim to naming the runs. Additionally apart from the parameters, even flow run context and task run context can be passed to them.

Even a type / subtype can be accepted whose constructor accept context and the generator function accept the parameters

### Example Use

```python
def flow_run_name_generator(flow_run_context, **kwargs):
    ... fetch useful context from flow run context
    ... fetch useful information from parameters
    ... fetch other useful context
    ... generate a run name and return string


@flow(flow_run_name=flow_run_name_generator)
def my_flow(arg1):
    pass
```

or

```python
# FlowRunNameGenerator base provided by prefect
class CustomFlowRunNameGenerator(FlowRunNameGenerator):
    def __init__(self, flow_run_context):
        super().__init__(flow_run_context)
    
    # overrides base method
    def generate(self, **kwargs) -> str:
        ... fetch useful context from flow run context
        ... fetch useful information from parameters
        ... fetch other useful context
        ... generate a run name and return string

@flow(flow_run_name=CustomFlowRunNameGenerator)
def my_flow(arg1):
    pass
```

similar support for `task_run_name`

### Additional context

I believe a function can be monkey-patched if we send an object with `format(**kwargs)` method, but an official support will be more appreciated. It should be possible to support both a function and subclasses for the same argument by using isinstance checks `e.g. isinstance(flow_run_name, str), isinstance(flow_run_name, Callable), isinstance(flow_run_name, type)`

billpalombi:
Thanks for submitting this enhancement request! We would definitely like to do this, though it may be some time before we get to it.

rsampaths16:
Hi @billpalombi it'd be really helpful if the proposal is reviewed to know if it is good enough for implementation, or if some modifications are needed.

madkinsz:
[...]

PR #8933, opened 2023-03-25
Accept functions to flow_run_name and task_run_name settings

rsampaths16:
Currently flow_run_name and task_run_name settings only accept template strings to generate new name. This PR enhances this feature by accept a function that expects input parameters dict and returns a string. This should ideally allow injecting flow run context and task run context too.

#### Flow run name setting
 ```python
import datetime
from prefect import flow

def generate_flow_run_name(params):
    date = datetime.datetime.utcnow()
    name = "anon" if "name" not in params else str(params["name"])

    return f"{name}-on-{date:%A}"

@flow(flow_run_name=generate_flow_run_name)
def my_flow(name: str):
    pass

# creates a flow run called 'marvin-on-Thursday'
my_flow(name="marvin")
```

#### Task run name setting
```python
import datetime

def generate_task_name(params):
    name = "anon" if "name" not in params else params["name"]
    date = datetime.datetime.utcnow()

    return f"hello-{name}-on-{date:%A}")

@task(name="My Example Task", 
      description="An example task for a tutorial.",
      task_run_name=generate_task_name)
def my_task(name):
    pass

@flow
def my_flow():
    # creates a run with a name like "hello-marvin-on-Thursday"
    my_task(name="marvin")
```

### Checklist

closes https://github.com/PrefectHQ/prefect/issues/8895

### DIFF
--- a/docs/tutorials/flow-task-config.md
+++ b/docs/tutorials/flow-task-config.md
@@ -190,6 +190,57 @@ def my_flow():
     my_task(name="marvin", date=datetime.datetime.utcnow())
 ```
 
+Additionally this setting also accepts a function that returns a string to be used for the task run name:
+
+```python
+import datetime
+from prefect import flow, task
+
+def generate_task_name():
+    date = datetime.datetime.utcnow()
+    return f"{date:%A}-is-a-lovely-day"
+
+@task(name="My Example Task",
+      description="An example task for a tutorial.",
+      task_run_name=generate_task_name)
+def my_task(name):
+    pass
+
+@flow
+def my_flow():
+    # creates a run with a name like "Thursday-is-a-lovely-day"
+    my_task(name="marvin")
+```
+
+If you need access to information about the task, use the `prefect.runtime` module. For example:
+
+```python
+from prefect import flow
+from prefect.runtime import flow_run, task_run
+
+def generate_task_name():
+    flow_name = flow_run.flow_name
+    task_name = task_run.task_name
+
+    parameters = task_run.parameters
+    name = parameters["name"]
+    limit = parameters["limit"]
+
+    return f"{flow_name}-{task_name}-with-{name}-and-{limit}"
+
+@task(name="my-example-task",
+      description="An example task for a tutorial.",
+      task_run_name=generate_task_name)
+def my_task(name: str, limit: int = 100):
+    pass
+
+@flow
+def my_flow(name: str):
+    # creates a run with a name like "my-flow-my-example-task-with-marvin-and-100"
+    my_task(name="marvin")
+```
+
+
 ## Flow and task retries
 
 Prefect includes built-in support for both flow and [task retries](/concepts/tasks/#retries), which you configure on the flow or task. This enables flows and tasks to automatically retry on failure. You can specify how many retries you want to attempt and, optionally, a delay between retry attempts:
--- a/docs/tutorials/flow-task-config.md
+++ b/docs/tutorials/flow-task-config.md
@@ -94,6 +94,49 @@ def my_flow(name: str, date: datetime.datetime):
 my_flow(name="marvin", date=datetime.datetime.utcnow())
 ```
 
+Additionally this setting also accepts a function that returns a string for the flow run name:
+
+```python
+import datetime
+from prefect import flow
+
+def generate_flow_run_name():
+    date = datetime.datetime.utcnow()
+
+    return f"{date:%A}-is-a-nice-day"
+
+@flow(flow_run_name=generate_flow_run_name)
+def my_flow(name: str):
+    pass
+
+# creates a flow run called 'Thursday-is-a-nice-day'
+my_flow(name="marvin")
+```
+
+If you need access to information about the flow, use the `prefect.runtime` module. For example:
+
+```python
+from prefect import flow
+from prefect.runtime import flow_run
+
+def generate_flow_run_name():
+    flow_name = flow_run.flow_name
+
+    parameters = flow_run.parameters
+    name = parameters["name"]
+    limit = parameters["limit"]
+
+    return f"{flow_name}-with-{name}-and-{limit}"
+
+@flow(flow_run_name=generate_flow_run_name)
+def my_flow(name: str, limit: int = 100):
+    pass
+
+# creates a flow run called 'my-flow-with-marvin-and-100'
+my_flow(name="marvin")
+```
+
+
 ## Basic task configuration
 
 By design, tasks follow a very similar model to flows: you can independently assign tasks their own name and description.
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1533,13 +1533,11 @@ async def orchestrate_task_run(
                     args, kwargs = parameters_to_args_kwargs(
                         task.fn, resolved_parameters
                     )
-
                     # update task run name
                     if not run_name_set and task.task_run_name:
-                        if not isinstance(task.task_run_name, str):
-                            raise RuntimeError("'task_run_name' is not a string")
-
-                        task_run_name = task.task_run_name.format(**resolved_parameters)
+                        task_run_name = _resolve_custom_task_run_name(
+                            task=task, parameters=resolved_parameters
+                        )
                         await client.set_task_run_name(
                             task_run_id=task_run.id, name=task_run_name
                         )
@@ -2085,6 +2083,25 @@ def _resolve_custom_flow_run_name(flow: Flow, parameters: Dict[str, Any]) -> str
         )
 
     return flow_run_name
+
+
+def _resolve_custom_task_run_name(task: Task, parameters: Dict[str, Any]) -> str:
+    if callable(task.task_run_name):
+        task_run_name = task.task_run_name()
+        if not isinstance(task_run_name, str):
+            raise TypeError(
+                f"Callable {task.task_run_name} for 'task_run_name' returned type"
+                f" {type(task_run_name).__name__} but a string is required."
+            )
+    elif isinstance(task.task_run_name, str):
+        task_run_name = task.task_run_name.format(**parameters)
+    else:
+        raise TypeError(
+            "Expected string or callable for 'task_run_name'; got"
+            f" {type(task.task_run_name).__name__} instead."
+        )
+
+    return task_run_name
 
 
 async def _run_task_hooks(task: Task, task_run: TaskRun, state: State) -> None:
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -612,14 +612,6 @@ async def orchestrate_flow_run(
             force=flow_run.state.is_pending(),
         )
 
-    if flow.flow_run_name:
-        flow_run_name = _resolve_custom_flow_run_name(flow, parameters)
-
-        await client.update_flow_run(flow_run_id=flow_run.id, name=flow_run_name)
-        logger.extra["flow_run_name"] = flow_run_name
-        logger.debug(f"Renamed flow run {flow_run.name!r} to {flow_run_name!r}")
-        flow_run.name = flow_run_name
-
     state = await propose_state(client, Running(), flow_run_id=flow_run.id)
 
     while state.is_running():
@@ -634,6 +626,21 @@ async def orchestrate_flow_run(
                 client=client,
                 parameters=parameters,
             ) as flow_run_context:
+                # update flow run name
+                if flow.flow_run_name:
+                    flow_run_name = _resolve_custom_flow_run_name(
+                        flow=flow, parameters=parameters
+                    )
+
+                    await client.update_flow_run(
+                        flow_run_id=flow_run.id, name=flow_run_name
+                    )
+                    logger.extra["flow_run_name"] = flow_run_name
+                    logger.debug(
+                        f"Renamed flow run {flow_run.name!r} to {flow_run_name!r}"
+                    )
+                    flow_run.name = flow_run_name
+
                 args, kwargs = parameters_to_args_kwargs(flow.fn, parameters)
                 logger.debug(
                     f"Executing flow {flow.name!r} for flow run {flow_run.name!r}..."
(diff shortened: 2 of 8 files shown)


src/prefect/engine.py
```diff
@@ -618,7 +618,14 @@ async def orchestrate_flow_run(
         )
 
     if flow.flow_run_name:
-        flow_run_name = flow.flow_run_name.format(**parameters)
+        if callable(flow.flow_run_name):
+            flow_run_name = flow.flow_run_name(parameters)
```
rsampaths16, line 622: **WARNING**: The parameters is sent directly to the `flow_run_name` function which could leave it vulnerable to mutation. We could try to send a shallow copy of the parameters dict, but still the individual args themselves could be mutated ( similar to how it's vulnerable in `str.format(**parameters)` ).

I'll leave the call on what to do here to the Prefect team.

src/prefect/engine.py
```diff
@@ -618,7 +618,14 @@ async def orchestrate_flow_run(
...
+        elif isinstance(flow.flow_run_name, str):
+            flow_run_name = flow.flow_run_name.format(**parameters)
+        else:
+            raise RuntimeError(
+                "flow_run_name parameter is neither a str nor a callable"
+            )
```
rsampaths16, line 628: In both `flow_run_name` and `task_run_name` I'm raising exceptions when they are neither `str` nor `callable`. Should this assertion be removed? If not should an assertion be added in the actual `Flow` and `Task` classes too?

src/prefect/engine.py
madkinsz: I think that's okay with me for now 👍

src/prefect/engine.py
madkinsz: Raising errors at the flow/task class level is better — then we will help people do the right thing when writing their task / flow rather than at runtime.

rsampaths16:
> Thanks for contributing!
> 
> Before we dive into a review of the implementation, let's chat about the interface a little. I'm thinking that since I've added `parameters` to the `prefect.runtime` module perhaps this should be `Callable[[], str]` and you should just pull data from the runtime as needed. What do you think?

I believe this would be a good approach. We won't force a people to accept parameters if they don't want to use it, and since they'd be available in `prefect.runtime` those who want them are not blocked.

src/prefect/engine.py
rsampaths16: I'll make changes to raise `TypeError`'s in `Flow` & `Task` class level. I'm also thinking to leave the `RuntimeError`'s here. Please confirm if the same if acceptable.

madkinsz:
@rsampaths16 sounds good to me!

rsampaths16:
Hey @madkinsz during testing I've realized that `prefect.runtime.flow_run` tries to access the parent's `flow_run_context`, making the current flows parameters inaccessible to it. I think we'll have to make the function accept parameters as a kwargs dictionary. What are your thoughts on this?

P.S: PR seems to be accidentally closed as I've force pushed a rebase. Will re-open again when I've completed the implementation.

rsampaths16:
Hey @madkinsz - regarding https://github.com/PrefectHQ/prefect/issues/8895#issuecomment-1483060367 what would be the plan to address this? Since the `run_name` functions are being called before we are inside the context of the `flow/task`? I'm assuming other functions that are accepted wouldn't have context too.

src/prefect/engine.py
madkinsz: It's okay here as well, but we'll probably want to move this into a separate function since it's kind of long now.

rsampaths16:
Hi @madkinsz !!

https://github.com/PrefectHQ/prefect/blob/c0828fa478f242610c4fe8dac87c05e364cb5094/src/prefect/engine.py#L1520-L1530


Upon going through the code above, is my understanding correct that since the `task_run_name` function will be called within `task_run_context` context, it'll be able to access the `prefect.runtime` parameters?

If yes, I'll try out accessing `prefect.runtime` from the `task_run_name` and check if a similar pattern works for flow too. Maybe that'd be a separate pull-request.

Do let me know if I'm looking at the wrong pattern here.

rsampaths16:
Awesome @madkinsz! I'll test this in the morning see if I can raise a new revision where functions don't accept parameters as discussed earlier. Thanks.

rsampaths16:
Hey @madkinsz I've made the discussed changes.
The summary is:
1. refactored few tests in flows and tasks by grouping them together
2. add flow_run_name and task_run_name function support: they don't accept any arguments.
3. move flow_run_name generation within the flow run context: so it can access parameters from `prefect.runtime.flow_run.parameters`

For more detailed info you can check-out the commit messages. :)

rsampaths16:
Hey @madkinsz - bumping again incase you've missed this;

docs/tutorials/flow-task-config.md
```diff
@@ -94,6 +94,26 @@ def my_flow(name: str, date: datetime.datetime):
 my_flow(name="marvin", date=datetime.datetime.utcnow())
 ```
 
+Additionally this setting also accepts a function that returns a string for the flow_run_name
```
madkinsz, line 97:
```suggestion
Additionally this setting also accepts a function that returns a string for the flow run name:
```

docs/tutorials/flow-task-config.md
```diff
@@ -94,6 +94,26 @@ def my_flow(name: str, date: datetime.datetime):
...
+    return f"{date:%A}-is-a-nice-day"
+
+@flow(flow_run_name=generate_flow_run_name)
+def my_flow(name: str):
+    pass
+
+# creates a flow run called 'Thursday-is-a-nice-day'
+my_flow(name="marvin")
+```
+
```
madkinsz, line 115: Can we add a blurb here that says something like

> If you need access to information about the flow, use the `prefect.runtime` module. For example:

Then include an example that pulls the name of the flow and a parameter from the runtime?

docs/tutorials/flow-task-config.md
```diff
@@ -147,6 +167,29 @@ def my_flow():
...
+      task_run_name=generate_task_name)
+def my_task(name):
+    pass
+
+@flow
+def my_flow():
+    # creates a run with a name like "Thursday-is-a-lovely-day"
+    my_task(name="marvin")
+```
+
```
madkinsz, line 191: Same as above re an additional example but with the task run runtime.

src/prefect/engine.py
```diff
@@ -633,6 +626,21 @@ async def orchestrate_flow_run(
...
+                    )
+
+                    await client.update_flow_run(
+                        flow_run_id=flow_run.id, name=flow_run_name
+                    )
+                    logger.extra["flow_run_name"] = flow_run_name
+                    logger.debug(
+                        f"Renamed flow run {flow_run.name!r} to {flow_run_name!r}"
+                    )
+                    flow_run.name = flow_run_name
```
madkinsz, line 642: This will update the name on each retry, right? We'll want to avoid setting the name more than once.

src/prefect/flows.py
```diff
@@ -148,7 +149,12 @@ def __init__(
             raise_on_invalid_name(name)
 
         self.name = name or fn.__name__.replace("_", "-")
+
+        if flow_run_name is not None:
+            if not isinstance(flow_run_name, str) and not callable(flow_run_name):
+                raise TypeError("'flow_run_name' is neither a string nor a callable")
```
madkinsz, line 155: Generally we want to give more information to help debug
```suggestion
                raise TypeError(f"Expected string or callable for 'flow_run_name'; got {type(flow_run_name).__name__}.")
```

src/prefect/tasks.py
```diff
@@ -214,7 +215,11 @@ def __init__(
         else:
             self.name = name
 
+        if task_run_name is not None:
+            if not isinstance(task_run_name, str) and not callable(task_run_name):
+                raise TypeError("'task_run_name' is neither a string nor a callable")
```
madkinsz, line 220: See flow run name error comment

src/prefect/engine.py
```diff
@@ -2056,6 +2065,32 @@ def should_log_prints(flow_or_task: Union[Flow, Task]) -> bool:
...
+        raise RuntimeError("'flow_run_name' is neither a string nor a callable")
+
+    return flow_run_name
+
+
+def _resolve_custom_task_run_name(task: Task, parameters: Dict[str, Any]) -> str:
+    if callable(task.task_run_name):
+        task_run_name = task.task_run_name()
+        if not isinstance(task_run_name, str):
+            raise RuntimeError("callable 'task_run_name' did not return a string")
```
madkinsz, line 2085: We probably want something more verbose for debugging

```suggestion
            raise RuntimeError(f"Callable {task.task_run_name} for 'task_run_name' returned type {type(task_run_name).__name__} but a string is required.")
```

Similar for the other error messages here

src/prefect/engine.py
```diff
@@ -2056,6 +2065,32 @@ def should_log_prints(flow_or_task: Union[Flow, Task]) -> bool:
     return flow_or_task.log_prints
 
 
+def _resolve_custom_flow_run_name(flow: Flow, parameters: Dict[str, Any]) -> str:
+    if callable(flow.flow_run_name):
+        flow_run_name = flow.flow_run_name()
+        if not isinstance(flow_run_name, str):
+            raise RuntimeError("callable 'flow_run_name' did not return a string")
```
madkinsz, line 2072: Note all of these `RuntimeError`s can be `TypeError`s

madkinsz:
Thanks again for contributing @rsampaths16! This looks great. I have some minor comments regarding UX when throwing exceptions.

src/prefect/engine.py
rsampaths16: Hey, I'll convert `RuntimeError`s to `TypeError`s where we check the type of the generating function/string. I'm not sure `TypeError` should be raised for the return value of the function.

rsampaths16: Just checked that `ValueError` is being raised on invalid return values in the code-base. Since the practice is established, I'll raise `TypeError` for the non string return value.

src/prefect/engine.py
rsampaths16: This is updating on each retry. Made changes to avoid setting it more than once. Used the same pattern as implemented in task_run_name. Have also added relevant tests.

docs/tutorials/flow-task-config.md
rsampaths16: done

docs/tutorials/flow-task-config.md
rsampaths16: done

docs/tutorials/flow-task-config.md
```diff
@@ -147,6 +167,29 @@ def my_flow():
     my_task(name="marvin", date=datetime.datetime.utcnow())
 ```
 
+Additionally this setting also accepts a function that returns a string to be used for the task_run_name
```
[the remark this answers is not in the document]
rsampaths16: done

docs/tutorials/flow-task-config.md
rsampaths16: done

src/prefect/engine.py
rsampaths16: done

src/prefect/engine.py
rsampaths16: done

src/prefect/engine.py
rsampaths16: done

src/prefect/flows.py
rsampaths16: done

src/prefect/tasks.py
rsampaths16: done

tests/test_flows.py
```diff
@@ -156,11 +173,36 @@ def my_flow():
         assert my_flow.fn() == "bar"
         assert my_flow.flow_run_name == "hi"
 
+    def test_flow_decorator_initializes_with_callable_flow_run_name(self):
+        # TODO: We should cover initialization with a task runner once introduced
```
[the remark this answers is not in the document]
rsampaths16: removed

tests/test_flows.py
```diff
@@ -156,11 +173,36 @@ def my_flow():
...
+        @flow(name="foo", version="B", flow_run_name=lambda: "hi")
```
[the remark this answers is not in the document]
rsampaths16: dropped

rsampaths16:
Hey @madkinsz - here are the changes since the last review:
* added documentation for using `prefect.runtime` within `flow_run_name` and `task_run_name` callable
* addressed nitpicks in documentation
* made exception messages more verbose
* changed `RuntimeError`s to `TypeError`s in `prefect.engine`
* addressed nitpicks in tests
* fixed issue where `flow_run_name` is reset on retries. ( followed the same pattern used in `task_run_name`. added missing tests too ).

do let me know your thoughts on it

rsampaths16:
Hi @madkinsz - I've verified that the tests failing aren't related to the current pr. Could you please check and unblock any workflows that could've been blocked by this. Thanks.

ref: https://github.com/PrefectHQ/prefect/pull/9239

rsampaths16:
Hi @madkinsz - thanks for unblocking the tests via https://github.com/PrefectHQ/prefect/pull/9241!!

madkinsz (approved):
Thanks for your patience!

rsampaths16:
Awesome :)

batuhanfaik:
Current implementation :)

```python
import datetime
from prefect import flow

def generate_flow_run_name():
    flow_run = prefect.runtime.flow_run
    date = datetime.datetime.utcnow()
    name = "anon" if not flow_run.name else str(flow_run.name)
    return f"{name}-on-{date:%A}"

@flow(flow_run_name=generate_flow_run_name)
def my_flow(name: str):
   pass

# creates a flow run called 'marvin-on-Thursday'
my_flow(name="marvin")
```

Issue #8239, opened 2023-01-23
Increase visibility of task cache runs, Allow users to manage their caches better


wendytang:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.7.7+

### Describe the current behavior

Currently users can implement a cache key and input a cache expiration in order to cache task run results and reuse them for later task runs. 

However there is not much visibility into when a task cache gets used and also not much functionality in managing the cache itself. Improving the **cache management functionality** would be incredibly helpful for long-running data pulling tasks which can be cached and skipped. 

Cache management mechanisms we are interested in include (in order of priority):
1. Viewing list of cache keys, which task + task run they trace back to
2. Deleting caches
3. Detecting when cache was last used, eg various cache usage metadata

Currently detecting task cache runs is not ideal:

```
@task(cache_key_fn=task_input_hash, cache_expiration=timedelta(days=1))
def hello_task(name_input):
    return "hello " + name_input

@flow
def hello_flow(name_input):
    # Flow Level Work
    logger = get_run_logger()
    rc = context.get_run_context()

    rs = hello_task(name_input, return_state=True)
    logger.info(f'RETURNED STATE: {rs}')
    logger.info(f'RETURNED STATE RESULT: {rs.result}')
    logger.info(f'RETURNED STATE DETAILS 1: {rs.state_details.cache_key}')

    logger.info("TASK RUN: ")
    if rs.result is not None and ("Cached" in str(rs)):
        logger.info("Task returned cached result")
    elif rs.result is not None:
        logger.info(f'Task caching result to {rs.state_details.cache_key}')
    return 0, 0

```
### Describe the proposed behavior

I'm not a stickler to how the solution has to be - any form of improvement + increased reliability to the current way of detecting cached result retrieval would be great.

I am wondering if the product roadmap would consider this direction:
<img width="665" alt="image" src="https://user-images.githubusercontent.com/7952060/214177309-001cf8a4-9e0a-4a1d-b177-7090f0072ed1.png">

### Additional context
For ++visibility and discussion

peytonrunyan:
Thanks for submitting the issue!

In the short term, there's definitely an easier way to access whether or not a task was run cached. I'll include a code snippet below. 

For the UI updates, we have access to the information about whether or not a task was completed and cached in the `task_run_state` table. The UI and product teams will have to give that look.

[...]

PR #8949, opened 2023-03-27
Add formatting of task storage keys and `prefect.runtime.task_run` module

madkinsz:
Extends #8924 
Closes https://github.com/PrefectHQ/prefect/issues/8239

Adds dynamic task storage keys with formatting from parameters and the `prefect.runtime` modules. Instead of explicitly passing format variables through to the result factory, it adds a `prefect.runtime.task_run` module to pull from and adjusts the scope of the task run context so that it is available during result creation 

Read the docs https://deploy-preview-8949--prefect-docs.netlify.app/concepts/results/#result-storage-key or this example:

```python
from prefect.serializers import JSONSerializer


@task(
    persist_result=True,
    result_storage_key="{flow_run.flow_name}__hello-{parameters[name]}.json",
)
def hello_world(name: str = "world"):
    return f"hello {name}"


@flow(result_serializer=JSONSerializer())
def my_flow():
    hello_world()
    hello_world(name="foo")
    hello_world(name="bar")


my_flow()
```

```
❯ ls ~/.prefect/storage | grep "my-flow"
my-flow__hello-bar.json
my-flow__hello-foo.json
my-flow__hello-world.json
```

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1471,7 +1471,10 @@ async def orchestrate_task_run(
     # Generate the cache key to attach to proposed states
     # The cache key uses a TaskRunContext that does not include a `timeout_context``
     cache_key = (
-        task.cache_key_fn(partial_task_run_context.finalize(), resolved_parameters)
+        task.cache_key_fn(
+            partial_task_run_context.finalize(parameters=resolved_parameters),
+            resolved_parameters,
+        )
         if task.cache_key_fn
         else None
     )
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1476,6 +1476,8 @@ async def orchestrate_task_run(
         else None
     )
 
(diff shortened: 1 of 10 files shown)


src/prefect/engine.py
```diff
@@ -1509,105 +1511,107 @@ async def orchestrate_task_run(
...
-                    logger.extra["task_run_name"] = task_run_name
-                    logger.debug(
-                        f"Renamed task run {task_run.name!r} to {task_run_name!r}"
+        with task_run_context.copy(
+            update={"task_run": task_run, "start_time": pendulum.now("UTC")}
+        ):
+            try:
+                with timeout_context as timeout_scope:
+                    args, kwargs = parameters_to_args_kwargs(
+                        task.fn, resolved_parameters
```
madkinsz, line 1520: This all just indented

billpalombi (approved):
Docs look good. Thank you!

jakekaplan (approved):

PR #8950, opened 2023-03-27
Retry on sqlite operational errors

jakekaplan:
Return a 503 error if the sqlite db is locked due to other connections accessing it. This will let the client retry with exponential backoff so the connections will all eventually resolve instead of erroring and returning 500s

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -145,11 +145,26 @@ async def integrity_exception_handler(request: Request, exc: Exception):
     )
 
 
-async def retry_exception_handler(request: Request, exc: Exception):
-    """Return a 503 so the client can try again."""
+async def db_locked_exception_handler(
+    request: Request, exc: sqlalchemy.exc.OperationalError
+):
+    """
+    Catch all sqlalchemy.exc.OperationalError. Return a 503 if it's a db locked error
+    to retry, otherwise log the error and return 500.
+    """
+    if (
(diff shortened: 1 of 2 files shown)


src/prefect/server/api/server.py
```diff
@@ -56,6 +58,17 @@
     logger=logger,
 )
 
+SERVICE_UNAVAILABLE_RETRY_AFTER = (
+    2  # The number of seconds after which the client should retry on 503s
+)
+SERVICE_UNAVAILABLE_RETRY_AFTER_JITTER = 1  # Retry jitter for 503s
```
madkinsz, line 64: Do we need jitter? The client adds jitter to retries now

src/prefect/server/api/server.py
```diff
@@ -144,6 +157,15 @@ async def integrity_exception_handler(request: Request, exc: Exception):
     )
 
 
+async def retry_exception_handler(request: Request, exc: Exception):
+    """Return a 503 so the client can try again."""
+    return JSONResponse(
+        content={"exception_message": "Service Unavailable"},
+        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+        headers={"Retry-After": str(get_retry_after_value())},
```
madkinsz, line 165: Do we need a retry after header? This will prevent the client from doing an exponential backoff

jakekaplan: same as above 😅

madkinsz (approved):

PR #8951, opened 2023-03-27
Add retrieval of flow run id in `prefect.runtime.flow_run` from task run contexts

madkinsz:
The flow run id can be retrieved from the task run object in a task run context if a flow run context is not available. This is important for runtime information to be accessible when running tasks on distributed workers and will ensure storage keys in https://github.com/PrefectHQ/prefect/pull/8949 are formatted correctly in that case.

### DIFF
--- a/tests/runtime/test_flow_run.py
+++ b/tests/runtime/test_flow_run.py
@@ -4,8 +4,8 @@ import pendulum
 import pytest
 
 from prefect import flow, states, tags
-from prefect.client.schemas import FlowRun
-from prefect.context import FlowRunContext
+from prefect.client.schemas import FlowRun, TaskRun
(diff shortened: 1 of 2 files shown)


tests/runtime/test_flow_run.py
```diff
@@ -57,6 +57,10 @@ def flow_with_new_id():
         assert flow_with_new_id() != "foo"
         assert flow_run.id == "foo"
 
+    async def test_id_can_be_retreived_from_task_run_context(self):
```
serinamarie, line 60:
```suggestion
    async def test_id_can_be_retrieved_from_task_run_context(self):
```

serinamarie (approved):

PR #8955, opened 2023-03-27
Bump prefect-ui-library and implement breaking composition changes

pleek91:
# Description
Implementation of https://github.com/PrefectHQ/prefect-ui-library/pull/1264
Also added the validate:types npm task we use in other projects and fixed a bad import

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -9,7 +9,7 @@
       "version": "2.8.0",
       "dependencies": {
         "@prefecthq/prefect-design": "1.5.0",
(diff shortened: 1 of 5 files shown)


znicholasbrown (commented):
I think this update will break the work queues playwright tests

pleek91:
> I think this update will break the work queues playwright tests

@znicholasbrown how so? The tests do have a similar naming convention for fixtures but they don't use these compositions.

znicholasbrown (approved):
🤦🏻 nevermind, I didn't realize they were named the same but not the same import at all. ignore me

PR #8971, opened 2023-03-28
Enhance experimental artifact CLI `ls`

serinamarie:
- Rename `--latest` option to `-a/--all` to match with https://man7.org/linux/man-pages/man1/ls.1.html
- Update the artifact filter to ensure it handles all possible `is_latest` scenarios.
- Add 2 more tests around the `--all` option to ensure correct output.

- Add clarification around a formerly less helpful error message.

### Checklist


desertaxle (approved):
LGTM!

src/prefect/server/schemas/filters.py
```diff
@@ -1567,7 +1567,7 @@ class ArtifactFilterLatest(PrefectFilterBaseModel):
 
     def _get_filter_list(self, db: "PrefectDBInterface") -> List:
         filters = []
-        if self.is_latest is not None:
+        if self.is_latest is not None and self.is_latest:
```
madkinsz, line 1570: Why is the field an `Optional[bool] = None` instead of `bool = False`? `None` has no special meaning for `is_latest` so we should just have it be a boolean with the correct default value.

serinamarie: Ah, thanks, I've applied your suggestion 👍

PR #8977, opened 2023-03-29
Fix `KubernetesJob.job_watch_timeout_seconds` docstring

madkinsz:
Addresses 'https://github.com/PrefectHQ/prefect/issues/8563#issuecomment-1487610535'


jakekaplan (approved):
seems risky

PR #8979, opened 2023-03-29
Enable stale pull requests and increase ops/hr

madkinsz:
- Enable stale pull requests after 60 days
- Bump stale bot operations to 500/hr; previously this was 60/hr to avoid reaching our GitHub API request limit but that actually appears to be 15,000/hr so we should be safe with this and I want to work through the backlog


bunchesofdonald (approved):
Thank you!

PR #9006, opened 2023-03-30
Add internal `QueueService` abstract class

madkinsz:
Adds a queue service class that runs on the global event loop. A queue service runs on a loop handling items as they are received. Items can be sent from any thread. The service can be "drained" which will ensure all remaining items are handled.

This will be used for the event and logs workers and we also hope to use this to implement a singleton client instance that runs on the global loop.

```python
from prefect._internal.concurrency.services import QueueService, drain_on_exit


class ExampleService(QueueService[int]):
    async def _handle(self, item: int):
        print(hex(id(self)), item)


with drain_on_exit(ExampleService):
    # These will all be handled by the same instance
    ExampleService.instance().send(1)
    ExampleService.instance().send(2)
    ExampleService.instance().send(3)


# Services are drained on interpreter exit, so 4 will display
# Since the service was already drained, it should be handled by a new instance though

ExampleService.instance().send(4)

```

```
❯ python example.py
0x1004e2610 1
0x1004e2610 2
0x1004e2610 3
0x106516550 4
```

### DIFF
--- a/src/prefect/_internal/concurrency/services.py
+++ b/src/prefect/_internal/concurrency/services.py
@@ -67,7 +67,7 @@ class QueueService(abc.ABC, Generic[T]):
 
     def send(self, item: T):
         """
-        Send an item to the instance of the service.
+        Send an item to this instance of the service.
         """
         if self._stopped:
             raise RuntimeError("Cannot put items in a stopped service.")
@@ -94,7 +94,7 @@ class QueueService(abc.ABC, Generic[T]):
         Process an item sent to the service.
         """
 
-    def drain_one(self) -> Union[Awaitable, None]:
+    def drain(self) -> Union[Awaitable, None]:
         """
         Stop this instance of the service and wait for remaining work to be completed.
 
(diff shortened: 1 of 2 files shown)


src/prefect/_internal/concurrency/services.py
```diff
@@ -0,0 +1,139 @@
...
+        # Put all early submissions in the queue
+        while self._early_items:
+            self.send(self._early_items.pop())
+
+        # Stop at interpreter exit by default
+        atexit.register(self.drain)
+
+    def stop(self):
+        # Stop sending work to this instance
+        self._clear_instance()
```
madkinsz, line 47: This should send `self` so that it only clears the instance _if_ this was the current one :)

src/prefect/_internal/concurrency/services.py
```diff
@@ -0,0 +1,139 @@
...
+            self._early_items.append(item)
+        else:
+            call_soon_in_loop(self._loop, self._queue.put_nowait, item)
+
+    async def __aenter__(self):
+        self.start()
+        return self
+
+    async def __aexit__(self, *_):
+        await self.drain()
```
madkinsz, line 73:
```suggestion
        await self.adrain()
```

jakekaplan (approved):

Issue #7999, opened 2022-12-28
Add --tail option to the prefect flow-run logs CLI


serinamarie:
Follow-up to https://github.com/PrefectHQ/prefect/pull/7982

### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

The CLI command: `prefect flow-run logs <flow_run_id>` shows all logs.

### Describe the proposed behavior

It would be great to add a `--tail` option to show only the last N logs:

```bash
prefect flow-run logs <flow_run_id> [--tail [numlines=50]]
```

Gairick52:
@serinamarie Hello mam ,can you please assign this issue to me,i would like to work on this

serinamarie:
Hi @Gairick52, that sounds great, thanks!

SlackJordan:
Does this issue still need to be worked on?

serinamarie:
Hi, @SlackJordan, it does 🙂

Shubhamparashar:
Has this issue been resolved and if not I would like to work on this

serinamarie:
@Shubhamparashar I don't believe anyone has picked up this issue. It hasn't been assigned to anyone.

Shubhamparashar:
@serinamarie please assign this issue to me, I would like to work on this

Shubhamparashar:
@serinamarie here is the PR link https://github.com/PrefectHQ/prefect/pull/9028

PR #9007, opened 2023-03-30
[Feature] --tail option added the prefect flow-run logs CLI #7999

Shubhamparashar:
Closes https://github.com/PrefectHQ/prefect/issues/7999

Added an option to see the last N logs for a flow run 

### Example

prefect flow-run logs <flow_run_id> --tail 20

it will show last 20 logs for flow_run_id

### Checklist

- [x] This pull request references any related issue by including closes


madkinsz (commented):
Thanks for contributing and welcome to the community!

I'm curious about this implementation. I was imagining just getting the last N logs using the sort-order we use for `--reverse` then displaying them in the right order to the user. What do you think about that? It's one less API call but it won't support pagination well.

We'll need to make sure we can add the `logs/count` route to Prefect Cloud in a performant way before we can accept the current implementation.

Shubhamparashar:
Hi @madkinsz ,

Yes, That was my first approach but I thought that if we are fetching a large number of logs first we are sorting it in DB then we will be reversing them and also the pagination, and compare to that counting logs seems cheaper to me. But I don't know the actual indexing.

And it's my first open-source contribution, so if you have any other feedback or advice, they are most welcome.

madkinsz:
@Shubhamparashar that makes sense to me! If you're interested in poking around at the database layer, all of the migrations are in this repository and we can add an index for SQLite / Postgres to improve the efficiency of the count query if it is not sufficient. We'll also have one of our engineers that works on Cloud take a look at the performance implications.

Shubhamparashar:
@madkinsz sure I will check out the migration to get more insight on this and add the tests for the CLI changes.

zangell44:
Thanks for contributing @Shubhamparashar!

> Yes, That was my first approach but I thought that if we are fetching a large number of logs first we are sorting it in DB then we will be reversing them and also the pagination, and compare to that counting logs seems cheaper to me. But I don't know the actual indexing.

Love that you're thinking about performance in detail!
As it turns out, our logs table is indexed to make querying the last N logs by timestamp a fast query. (And if heavy use reveals I'm wrong, we can always add the appropriate index!)
For cases where many logs match the filter criteria, counting all logs first will definitely be more expensive to do.

For this PR, I would use the reverse sort order instead of adding a new `/count` route.

--
For even faster performance when using Cloud in the future, we could use the `events` websocket endpoint. I would recommend doing that in a subsequent PR.

Shubhamparashar:
@zangell44 sure I will do it using the reverse sort on timestamp

Shubhamparashar:
@zangell44 @madkinsz I created a new PR for tail functionality added using reverse and offset
here is the link for the same - https://github.com/PrefectHQ/prefect/pull/9028

PR #9019, opened 2023-03-31
events queue service

madkinsz:
- Update `drain` implementation to wait
- Avoid deadlocks when errors are encountered
- Update events workers to run on the global event loop


jakekaplan (approved):

bunchesofdonald (approved):
🥳

Issue #9020, opened 2023-03-31
map in prefect2 does not handle default parameters very well


tharwan:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

in prefect 2 map an not handle default parameters that are a list correctly. 
I think this is problematic for multiple reasons:
* it worked in prefect 1 (see example)
* it is not documented
* it makes converting existing functions to a task (and mapping them) impossible in some cases

### Reproduction

in prefect 1 this works fine:
```python3
from typing import List

from prefect import Flow, task


@task
def task_a(one: List[int] = [1, 2, 3], two: List[int] = [4, 5, 6]):
    print(sum(one + two))


with Flow("main") as main:
    a = [
        [1, 1, 1],
        [2, 2, 2],
    ]
    task_a.map(a)


if __name__ == "__main__":
    main.run()
```

while in prefect 2 this fails:
```python3
from typing import List

from prefect import flow, task


@task
def task_a(one: List[int] = [1, 2, 3], two: List[int] = [4, 5, 6]):
    print(sum(one + two))


@flow()
def main():
    a = [
        [1, 1, 1],
        [2, 2, 2],
    ]
    task_a.map(a)


if __name__ == "__main__":
    main()

```

### Error

```python3
Traceback (most recent call last):
  File "/Users/fkuehnlenz/Documents/e2m/e2m.prometheus.afrractivationanalysisapp/prefect_test.py", line 21, in <module>
    main()
  File "/Users/fkuehnlenz/Documents/e2m/e2m.prometheus.afrractivationanalysisapp/.venv/lib/python3.8/site-packages/prefect/flows.py", line 468, in __call__
    return enter_flow_run_engine_from_flow_call(
  File "/Users/fkuehnlenz/Documents/e2m/e2m.prometheus.afrractivationanalysisapp/.venv/lib/python3.8/site-packages/prefect/engine.py", line 182, in enter_flow_run_engine_from_flow_call
    retval = from_sync.wait_for_call_in_loop_thread(
  File "/Users/fkuehnlenz/Documents/e2m/e2m.prometheus.afrractivationanalysisapp/.venv/lib/python3.8/site-packages/prefect/_internal/concurrency/api.py", line 137, in wait_for_call_in_loop_thread
    return call.result()
[...]

PR #9021, opened 2023-03-31
Fix bug where iterable defaults were treated as mapped parameters

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9020

Previously we treated iterable defaults as mapped parameters. However, default values should never be mapped over and should always be treated as `unmapped`. 

```python
@task
def add_some(x, y=[1, 4]):
    return x + sum(y)

@flow
def my_flow():
    numbers = [1, 2, 3]
    return add_some.map(numbers)
```

Example previous failure:

```
prefect.exceptions.MappingLengthMismatch: Received iterable parameters with different lengths.
Parameters for map must all be the same length. Got lengths: {'x': 3, 'y': 2}
```

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1039,6 +1039,11 @@ async def begin_task_map(
         call_parameters = {key: value[i] for key, value in iterable_parameters.items()}
         call_parameters.update({key: value for key, value in static_parameters.items()})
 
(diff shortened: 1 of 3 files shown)


desertaxle (approved):
LGTM!

PR #9028, opened 2023-04-01
[enhancement] --tail option added the prefect flow-run logs CLI #7999

Shubhamparashar:
Closes https://github.com/PrefectHQ/prefect/issues/7999
Added an option to see the last N logs for a flow run
### Example

prefect flow-run logs <flow_run_id> --tail 20

it will show last 20 logs for flow_run_id

### Checklist

- [x] This pull request references any related issue by including "closes https://github.com/PrefectHQ/prefect/issues/7999"


serinamarie (changes requested):
Hi @Shubhamparashar, this is looking great! Please see my comment regarding implementation. As for the tests, it would be great if you could add `test_passing_head_and_tail_raises` and `test_default_tail_returns_default_num_logs`. Thanks again!

src/prefect/cli/flow_run.py
```diff
@@ -171,6 +171,13 @@ async def logs(
         "-r",
         help="Reverse the logs order to print the most recent logs first",
     ),
+    tail: int = typer.Option(
+        None,
+        "--tail",
+        "-t",
+        help="Show the last N logs",
+        min=1,
+    ),
```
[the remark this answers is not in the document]
Shubhamparashar: @serinamarie I have updated the PR according to your feedback

tests/cli/test_flow_run.py
```diff
@@ -563,3 +563,69 @@ async def test_when_num_logs_passed_with_reverse_param_and_num_logs(
...
+        # When/Then
+        await run_sync_in_worker_thread(
+            invoke_and_assert,
+            command=["flow-run", "logs", str(flow_run.id), "-t"],
+            expected_code=0,
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(
+                    self.LOGS_DEFAULT_PAGE_SIZE, self.LOGS_DEFAULT_PAGE_SIZE - 20, -1
+                )
```
serinamarie, line 602:
```suggestion
                for i in range(
                    self.LOGS_DEFAULT_PAGE_SIZE - 9, self.LOGS_DEFAULT_PAGE_SIZE
                )
```

Since we aren't specifying the `reverse` flag (which reverses "the logs order to print the most recent logs first"), we would expect the logs to be printed TIMESTAMP ASC. I believe the above suggestion would test that.

src/prefect/cli/flow_run.py
```diff
@@ -209,13 +231,12 @@ async def logs(
                 offset=offset,
                 sort=(
                     schemas.sorting.LogSort.TIMESTAMP_DESC
-                    if reverse
+                    if reverse or tail
                     else schemas.sorting.LogSort.TIMESTAMP_ASC
                 ),
```
serinamarie, line 236: We want offset to be 0 as we iterate through the first time with `tail` because we are retrieving the logs in TIMESTAMP DESC order. If user specified `flow-run logs --tail --num_logs 250` for example, your logic below would change offset to 50:
offset = max(0, user_specified_num_logs - LOGS_DEFAULT_PAGE_SIZE).

In other words, the `if reverse or tail` condition you added should be sufficient and you should be able to remove the code that sets the tail to a specific limit or offset as I believe the existing code was made to be extensible to new use cases like this one.

src/prefect/cli/flow_run.py
```diff
@@ -229,8 +250,19 @@ async def logs(
...
+                        offset = offset - LOGS_DEFAULT_PAGE_SIZE
+                else:
+                    more_logs = False
             else:
-                # No more logs to show, exit
-                more_logs = False
+                if len(page_logs) == LOGS_DEFAULT_PAGE_SIZE:
+                    offset += LOGS_DEFAULT_PAGE_SIZE
+                else:
+                    # No more logs to show, exit
```
serinamarie, line 267: I don't think you need these changes. See note above.

serinamarie (commented):
Hi @Shubhamparashar! Thanks for making those updates so quickly. We're getting really close here. I have a few suggestions to make the code a little simpler as I think some of the existing code can be utilized more for the tail option so that you don't need to make as many changes. I would read through all the comments and please let me know if you have any clarifying questions to ask! Thanks again and looking forward to getting this one in :)

src/prefect/cli/flow_run.py
Shubhamparashar: I believe there is a bug in the previous logic. 
`            # Update the number of logs retrieved
            num_logs_returned += num_logs_to_return_from_page

            if len(page_logs) == LOGS_DEFAULT_PAGE_SIZE:
                offset += LOGS_DEFAULT_PAGE_SIZE
            else:
                # No more logs to show, exit
                more_logs = False
`

When using the command `flow-run logs --head --num_logs 250 `to retrieve logs and assuming there are 500 logs available, the code will retrieve all 500 logs because it does not utilize the `num_logs_returned` variable. The code only stops when there are no more logs left, as indicated by the `more_logs` flag. Similarly, if the previous logic were to be used for the `--tail` command, it would first return the last 200 logs, followed by the rest of the logs in the same manner, with logs 301 to 500, 101 to 300, and finally 1 to 100.

madkinsz: Not up to date on this but there's some changes to this logic happening in #9130 — would that resolve problems here?

Shubhamparashar: Hi @madkinsz Based on my understanding of #9130 it seems that the code has been modified to eliminate hard-coded variables. Additionally, a new API has been incorporated that retrieves all the logs to determine their length. However, it appears that the issue of limiting the logs has not been resolved.

serinamarie: I've put up a PR for that bug: https://github.com/PrefectHQ/prefect/pull/9314

Shubhamparashar: Once #9314 is merged I will make changes accordingly

Shubhamparashar: Hi @serinamarie  I've updated the PR, but we still require the offset for the tail flag. To clarify, if there are 500 logs and I want to retrieve the last 250 logs using the command flow-run logs --tail --num_logs 250, the offset should not start from 0. If it does, the command will return the last 200 logs (logs 301-500) followed by logs 251-300. Although the logs themselves are correct, the sequence will be inaccurate.

tests/cli/test_flow_run.py
```diff
@@ -587,3 +587,69 @@ async def test_when_num_logs_passed_with_reverse_param_and_num_logs(
...
+            expected_output_contains=[
+                f"Flow run '{flow_run.name}' - Log {i} from flow_run {flow_run.id}."
+                for i in range(
+                    self.LOGS_DEFAULT_PAGE_SIZE - 9, self.LOGS_DEFAULT_PAGE_SIZE
+                )
+            ],
+            expected_line_count=20,
+        )
+
+    async def test_reverse_tail_with_num_logs(self, flow_run_factory):
```
serinamarie, line 631: Ah, could we add 2 more tests: 
- one called `test_reverse_tail_returns_default_num_logs` (in TIMESTAMP DESC order)
- and one called `test_when_num_logs_greater_than_page_size_with_tail_outputs_correct_num_logs` (in TIMESTAMP ASC order)

src/prefect/cli/flow_run.py
serinamarie: Ah, you're right. The code looks clearer now, thanks 👍

src/prefect/cli/flow_run.py
```diff
@@ -180,11 +189,19 @@ async def logs(
...
+    if head and tail:
+        exit_with_error("Please provide either a `head` or `tail` option but not both.")
+
+    user_specified_num_logs = (
+        num_logs or LOGS_WITH_LIMIT_FLAG_DEFAULT_NUM_LOGS
+        if num_logs or tail or num_logs
+        else None
+    )
+
+    # if using tail update offset and tail_limit according to LOGS_DEFAULT_PAGE_SIZE
```
serinamarie, line 202:
```suggestion
    # if using tail update offset according to LOGS_DEFAULT_PAGE_SIZE
```

src/prefect/cli/flow_run.py
```diff
@@ -231,8 +247,21 @@ async def logs(
...
-                # No more logs to show, exit
-                more_logs = False
+                if (
+                    num_logs_returned < user_specified_num_logs
+                    and len(page_logs) == LOGS_DEFAULT_PAGE_SIZE
+                ):
+                    offset += LOGS_DEFAULT_PAGE_SIZE
+                else:
+                    # No more logs to show, exit
+                    more_logs = False
```
serinamarie, line 267: Some comments here explaining what is happening will be useful for future implementers :)

serinamarie (changes requested):
Hey @Shubhamparashar, I think we just need a few more tests and some clarifying comments :)

Shubhamparashar:
Hi @serinamarie I updated PR with the requested changes. Hope this sums it up and if there is anything else required from my end please let me know

serinamarie (approved):
The tests passed on the first try-- Great work on this and congrats on your first contribution! 🎉

PR #9045, opened 2023-04-03
Add `BatchedQueueService` abstract class

madkinsz:
In preparation for moving the log worker to run on the global event loop, introduces a batched queue service so we can send logs in batches even when they're emitted one at a time.


jakekaplan (approved):
LGTM! This is cool to see in action

PR #9048, opened 2023-04-03
Update log worker to run on the global event loop

madkinsz:
Uses the abstraction introduced in #9045 to reimplement the log worker and update it to run on the global event loop.

Drops a great deal of worker management complexity as well as handling of retries (we rely on retries in the client layer instead).

### DIFF
--- a/src/prefect/_internal/concurrency/services.py
+++ b/src/prefect/_internal/concurrency/services.py
@@ -3,6 +3,7 @@ import asyncio
 import atexit
 import concurrent.futures
 import contextlib
+import sys
 import threading
 from collections import deque
 from typing import Awaitable, Dict, Generic, List, Optional, Type, TypeVar, Union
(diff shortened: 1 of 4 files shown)


src/prefect/_internal/concurrency/services.py
```diff
@@ -46,7 +47,18 @@ def start(self):
             self.send(self._early_items.popleft())
 
         # Stop at interpreter exit by default
-        atexit.register(self.drain_all)
+        if sys.version_info < (3, 9):
+            atexit.register(self.drain)
+        else:
+            # See related issue at https://bugs.python.org/issue42647
```
jakekaplan, line 53: 🤯

src/prefect/_internal/concurrency/services.py
```diff
@@ -122,6 +134,11 @@ def _drain(self) -> concurrent.futures.Future:
         Internal implementation for `drain`. Returns a future for sync/async interfaces.
         """
         self._stop()
+        if self._done_event.is_set():
```
jakekaplan, line 137: When does this case occur? If we've already called `_drain` before?

jakekaplan (approved):
`logs-queue-service-4real` 👏

src/prefect/_internal/concurrency/services.py
madkinsz: Yeah basically repeated calls to `drain` i.e. the `drain` call trigger by atexit after a manual drain

src/prefect/_internal/concurrency/services.py
madkinsz: This copied over from the old log worker. I had a great time with that one originally :D

PR #9056, opened 2023-04-04
[DRAFT] Prefect client retry error codes

### LINKED ISSUE #6748: Ability to customize retry status codes for client

EmilRex:
Closes #6748 

### Checklist

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -312,6 +312,25 @@ def max_log_size_smaller_than_batch_size(values):
     return values
 
 
+def status_codes_as_integers_in_range(value):
+    """
+    Validator for settings asserting the status codes are integers in the range 100-599
+    """
+    if value == "":
+        return set()
+
+    values = {v.strip() for v in value.split(",")}
+
+    if any(not v.isdigit() or int(v) < 100 or int(v) > 599 for v in values):
+        raise ValueError(
+            "PREFECT_SERVER_HEALTHY_STATUS_CODES must be a comma separated list of "
+            "integers between 100 and 599."
+        )
+
+    values = {int(v) for v in values}
+    return values
+
+
 def warn_on_database_password_value_without_usage(values):
     """
     Validator for settings warning if the database password is set but not used.
@@ -523,6 +542,14 @@ client requests. Higher values introduce larger amounts of jitter.
 
 Set to 0 to disable jitter. See `clamped_poisson_interval` for details on the how jitter
 can affect retry lengths.
+"""
+
+PREFECT_CLIENT_RETRY_EXTRA_CODES = Setting(
+    str, default="", value_callback=status_codes_as_integers_in_range
+)
+"""
+A comma-separated list of extra HTTP status codes to retry on. Defaults to an empty string.
+429 and 503 are always retried.
 """
 
 PREFECT_CLOUD_API_URL = Setting(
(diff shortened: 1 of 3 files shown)


src/prefect/settings.py
```diff
@@ -525,6 +544,14 @@ def default_cloud_ui_url(settings, value):
 can affect retry lengths.
 """
 
+PREFECT_CLIENT_RETRY_EXTRA_CODES = Setting(
```
madkinsz, line 547: Hm should we do extra codes or just `CODES` and have it default to our current list?

src/prefect/settings.py
```diff
@@ -312,6 +312,25 @@ def max_log_size_smaller_than_batch_size(values):
...
+    Validator for settings asserting the status codes are integers in the range 100-599
+    """
+    if value == "":
+        return set()
+
+    values = {v.strip() for v in value.split(",")}
+
+    if any(not v.isdigit() or int(v) < 100 or int(v) > 599 for v in values):
+        raise ValueError(
+            "PREFECT_SERVER_HEALTHY_STATUS_CODES must be a comma separated list of "
```
madkinsz, line 326: Looks like this refers to the wrong setting

EmilRex: Whoops... yep, need to clean up the name.

src/prefect/settings.py
EmilRex: I was thinking the same thing at first, but then I worry users might accidentally override the safe defaults. On the other hand, it's more optionality.

madkinsz: Yeah I'd be a little concerned about that too. And like.. who doesn't want to retry on a 429?

EmilRex: Exactly! I think it's reasonable to assume that users will always want to retry 429s and 503s.

src/prefect/settings.py
```diff
@@ -525,6 +545,14 @@ def default_cloud_ui_url(settings, value):
 can affect retry lengths.
 """
 
+PREFECT_CLIENT_RETRY_EXTRA_CODES = Setting(
+    str, default="", value_callback=status_codes_as_integers_in_range
+)
+"""
+A comma-separated list of extra HTTP status codes to retry on. Defaults to an empty string.
+429 and 503 are always retried.
+"""
```
madkinsz, line 554: Let's add a warning here about retrying on other error codes — might cause unexpected behavior if the route is not idempotent.

madkinsz (approved):
Nice work! Thanks!

Issue #7642, opened 2022-11-23
Avoid throwing `MissingResult` errors for `NotReady` tasks


j-tr:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

when running a series of dependend tasks and one but not the last tasks fails, the remaining dependent tasks go into state NotReady and have None data in the state. when collecting results for these tasks, this fails with the exception:
```
raise MissingResult(
                "State data is missing. "
                "Typically, this occurs when result persistence is disabled and the "
                "state has been retrieved from the API."
                )
```

The exception message suggests that this would be typically due to result persistence, however in the above mentioned scenario, it is only about failed upstream tasks. This is potentially missleading.

### Describe the proposed behavior

Raise an exception with a vaguer message or treat the NotReady case individually and provide a sensible error message.

### Example Use

Users would not be sent down the rabbit hole of result persistence if not needed.

j-tr:
Please find below an example flow and traceback.

Version:             2.6.9
API version:         0.8.3
Python version:      3.8.15
Git commit:          014d093e
Built:               Tue, Nov 22, 2022 2:14 PM
OS/Arch:             linux/x86_64
Server type:         cloud


```
from prefect import flow, task


@task()
def task_1():
    raise Exception


@task()
def task_2(x):
    return


@flow()
def example_flow():
    future_1 = task_1.submit()
    future_2 = task_2.submit(future_1)
    
    future_2.result()


if __name__ == "__main__":
    example_flow()
```

Traceback

```
10:06:19.113 | INFO    | prefect.engine - Created flow run 'real-warthog' for flow 'example-flow'
10:06:20.676 | INFO    | Flow run 'real-warthog' - Created task run 'task_1-1619c4ff-0' for task 'task_1'
10:06:20.678 | INFO    | Flow run 'real-warthog' - Submitted task run 'task_1-1619c4ff-0' for execution.
10:06:20.860 | INFO    | Flow run 'real-warthog' - Created task run 'task_2-49f2f021-0' for task 'task_2'
10:06:20.862 | INFO    | Flow run 'real-warthog' - Submitted task run 'task_2-49f2f021-0' for execution.
10:06:21.127 | ERROR   | Task run 'task_1-1619c4ff-0' - Encountered exception during execution:
Traceback (most recent call last):
[...]

PR #9062, opened 2023-04-04
Update task runs without completed upstreams to be placed in FAILED state instead of PENDING

zanieb:
Closes https://github.com/PrefectHQ/prefect/issues/8124 
Closes https://github.com/PrefectHQ/prefect/issues/7642

Changes our "not ready" behavior to use a FAILED state type instead of a PENDING state type. This means that `allow_failure` can be used to ensure a downstream runs if an upstream has not run when chaining tasks.

This has the possibility of being a breaking change for some users.
I've also changed the state name from "NotReady" to "UpstreamNotReady".

Following #9152 this should not have negative effects on reruns.


madkinsz:
@billpalombi / @cicdw

Let me know if you want to discuss this. In short, this is necessary for good behavior with chained mapped tasks and `allow_failure`.

There are just a few failing tests to resolve here.

madkinsz:
cc @zhen0 I'm not sure who is best suited to take this up

Issue #9065, opened 2023-04-04
add model functions for `Variables`


jakekaplan:
Adds `Variable` model functions for the OSS implementation of:
- ` count_variables`
- ` create_variable`
- ` delete_variable`
- ` delete_variable_by_name`
- ` read_variable`
- ` read_variable_by_name`
- ` read_variables`
- ` update_variable`
- ` update_variable_by_name`

jakekaplan:
done in https://github.com/PrefectHQ/prefect/pull/9066

PR #9066, opened 2023-04-04
Add `Variables` model functions

jakekaplan:
Closes: https://github.com/PrefectHQ/prefect/issues/9065

Adds `Variable` model functions for the OSS implementation of:
- `count_variables`
- `create_variable`
- `delete_variable`
- `delete_variable_by_name`
- `read_variable`
- `read_variable_by_name`
- `read_variables`
- `update_variable`
- `update_variable_by_name`

### DIFF
--- a/src/prefect/server/schemas/core.py
+++ b/src/prefect/server/schemas/core.py
@@ -1221,7 +1221,7 @@ class Variable(ORMBaseModel):
         example="my-value",
         max_length=MAX_VARIABLE_VALUE_LENGTH,
     )
-    tags: list[str] = Field(
+    tags: List[str] = Field(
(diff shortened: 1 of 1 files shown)


src/prefect/server/schemas/core.py
```diff
@@ -37,6 +37,9 @@
 
 DEFAULT_BLOCK_SCHEMA_VERSION = "non-versioned"
 
+MAX_VARIABLE_NAME_LENGTH = 255
```
jakekaplan, line 40: I think these are still reasonable defaults to keep in place

madkinsz (approved):

PR #9069, opened 2023-04-04
Add a 10 minute cache to API healthchecks

madkinsz:
Caching applies per API URL reducing the number of redundant API healthchecks we make. Only applies to explicit checks in the engine, this is not applied at the client level.

### DIFF
--- a/src/prefect/testing/fixtures.py
+++ b/src/prefect/testing/fixtures.py
@@ -74,7 +74,7 @@ async def hosted_api_server(unused_tcp_port_factory):
             with anyio.move_on_after(20):
                 while True:
                     try:
-                        response = await client.get(api_url + "health")
+                        response = await client.get(api_url + "/health")
                     except httpx.ConnectError:
                         pass
                     else:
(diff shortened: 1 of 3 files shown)


madkinsz:
Benchmarks show a significant and consistent performance improvement, roughly 30% ❗

madkinsz:
I didn't add tests for this yet

billpalombi:
> Benchmarks show a significant and consistent performance improvement, roughly 30% ❗

Thats amazing!

cicdw (approved):

tests/test_engine.py
```diff
@@ -2030,10 +2030,12 @@ async def test_healthcheck_for_remote_client(self, hosted_api_server):
         async with get_client() as client:
             await check_api_reachable(client, fail_message="test")
 
-        assert hosted_api_server in API_HEALTHCHECKS
-        assert isinstance(API_HEALTHCHECKS[hosted_api_server], float)
+        expected_url = hosted_api_server + "/"  # httpx client appends trailing /
```
cicdw, line 2033: you could use posixpath manipulation to guard against this

src/prefect/testing/fixtures.py
```diff
@@ -74,7 +74,7 @@ async def hosted_api_server(unused_tcp_port_factory):
             with anyio.move_on_after(20):
                 while True:
                     try:
-                        response = await client.get(api_url + "/health")
+                        response = await client.get(api_url + "health")
```
madkinsz, line 77:
```suggestion
                        response = await client.get(api_url + "/health")
```

Issue #8986, opened 2023-03-30
Race Condition Between Task Submission and Waiting


haddon-korinek-ent:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Starting in 2.8.7, task runners are sometimes having 'wait' called before the 'submit' call finishes if you perform an 'await' call during submission. Swapping back to 2.8.6 fixes the issue.

We use a slightly modified version of the DaskTaskRunner in production that performs an 'await' during the task submission method which is how we noticed this issue.

Task submission happens in the background https://github.com/PrefectHQ/prefect/blob/main/src/prefect/engine.py#L1159

There's some checks to "wait for submission" https://github.com/PrefectHQ/prefect/blob/main/src/prefect/futures.py#L174, but that isn't sufficient to safely call 'wait' on task runners because it only checks that the task run was created, not that the task runner actually finished submitting it.

The DaskTaskRunner has to wait for it's client to submit a dask task to the cluster in order to populate _dask_futures. When you wait on that task runner, it does a lookup in that map based on the task run ID.

So this ends up in a situation where results are awaited before their keys are ready to be awaited by the DaskTaskRunner https://github.com/PrefectHQ/prefect/blob/main/src/prefect/engine.py#L1131

### Reproduction

```python3
import asyncio
import sys

from prefect import flow, task
from prefect.futures import resolve_futures_to_data
from prefect.server.schemas.states import Completed
from prefect.task_runners import BaseTaskRunner, TaskConcurrencyType


@task()
async def task1():
    return "foo"


class SleepyTaskRunner(BaseTaskRunner):
    def __init__(self):
        super().__init__()
        self.submitted = []

    @property
    def concurrency_type(self) -> TaskConcurrencyType:
        return TaskConcurrencyType.PARALLEL

    async def wait(self, key, timeout: float = None):
        print(f"waiting on {key}")
        if key not in self.submitted:
            print(f"tried to wait on key {key} but it didn't finish submitting")
            sys.exit()
        return Completed(
            message="finished",
        )

    async def submit(self, key, call):
        await asyncio.sleep(60)
        print(f"submitting task {key}")
        self.submitted.append(key)


@flow(task_runner=SleepyTaskRunner())
async def test_slow_submitter():
    res = []
    for i in range(100):
[...]

Issue #8982, opened 2023-03-29
Submitting many concurrent tasks causes the flow to be stuck


lucawork:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If is submit many tasks at the same time the tasks and thus the flow never finishes. The limit when this starts to happen seems to depend on the machine.

* on windows with WSL 10 submits is enough to lock up the flow
* on most other machines more tasks are needed (the 1000 in the example does the trick)
* on our production machine the flows gets stuck every so often (~20 concurrent tasks)

### Reproduction

```python3
import time
from prefect import task, flow, get_run_logger

@task
def my_task(i):
    logger = get_run_logger()
    logger.info(f"Task start {i}")
    time.sleep(0.5)

@flow
def my_flow():
    for i in range(1000):
        my_task.submit(i)

my_flow()
```

### Error

```python3
The execution get stuck, the status of the flow is running but it persist in that state without proceeding.
```

### Versions

```Text
prefect 2.8.7
```

### Additional context

The issue on the production machine only showed up after upgrading from Prefect 2.7.9 (might be a coincidence tho)

madkinsz:
What is the last log message the flow run emits before hanging?

felix-ht:
@madkinsz we are using prefect cloud if that is what you are asking

```
19:07:05.413 | INFO    | Flow run 'mellow-groundhog' - Created task run 'my_task-998' for task 'my_task'
19:07:05.413 | INFO    | Flow run 'mellow-groundhog' - Submitted task run 'my_task-998' for execution.
19:07:05.598 | INFO    | Flow run 'mellow-groundhog' - Created task run 'my_task-993' for task 'my_task'
19:07:05.599 | INFO    | Flow run 'mellow-groundhog' - Submitted task run 'my_task-993' for execution.
19:07:05.654 | INFO    | Flow run 'mellow-groundhog' - Created task run 'my_task-994' for task 'my_task'
19:07:05.654 | INFO    | Flow run 'mellow-groundhog' - Submitted task run 'my_task-994' for execution.
19:07:05.867 | INFO    | Flow run 'mellow-groundhog' - Created task run 'my_task-999' for task 'my_task'
[...]

PR #9072, opened 2023-04-04
Avoid using spin locks while waiting for task submission

madkinsz:
Supersedes #9070 
Closes https://github.com/PrefectHQ/prefect/issues/8986
May resolve https://github.com/PrefectHQ/prefect/issues/8982

Removes use of a spin-lock for cross-event loop waits in favor of nice future-based scheduling :) should be a significant performance improvement for some task submission cases and may resolve submission deadlocks.


madkinsz:
Benchmarks suggest a performance regression although I find that unlikely. We could check if the event is set before sending a wait to another loop.

jakekaplan (approved):
I trust you 😄

PR #9073, opened 2023-04-04
Update concurrent task runner to use efficient wait patterns

madkinsz:
Similar to https://github.com/PrefectHQ/prefect/pull/9072 removes spin-locks from the concurrent task runner which should improve stability and performance.

### DIFF
--- a/src/prefect/task_runners.py
+++ b/src/prefect/task_runners.py
@@ -52,7 +52,6 @@ Example:
 For usage details, see the [Task Runners](/concepts/task-runners/) documentation.
 """
 import abc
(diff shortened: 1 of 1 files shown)


madkinsz:
Benchmarks suggest a 20% improvement when using submitted tasks.

madkinsz:
Futures caused some interesting test failures when timeouts were used to retrieve results. See comment in https://github.com/PrefectHQ/prefect/pull/9073/commits/f332c4a580716b445c48b102ee258b71cd3cb5d3

jakekaplan (approved):

PR #9077, opened 2023-04-05
Update Dockerfile

mianos:
My new docker install( 20.10.23)  complains with the following self explanatory message:
```
Step 10/47 : COPY ./ui/package*.json .
When using COPY with more than one source file, the destination must be a directory and end with a /
```
To fix this I need to change it to:
```
COPY ./ui/package*.json ./
```

### Example
With this change the docker image is built.

### Checklist
Trivial fix. May not need a proper PR?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN apt-get update && \
 RUN npm install -g npm@8
 
 # Install dependencies separately so they cache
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
Thanks for contributing!

Weird that this is fine in our CI process. Is this a new change in Docker? cc @chrisguidry

PR #9104, opened 2023-04-05
`Variables` module

jakekaplan:
Adds `Variables` to be used in and outside of flows.

You can access variables in 3 ways:

```python
from prefect import variables

@flow
async def my_flow():
    variables.get("my_var")
    variables["my_var"]
    await variables.get("my_var")
```
   
closes: https://github.com/PrefectHQ/prefect/issues/9108

### DIFF
--- a/src/prefect/variables.py
+++ b/src/prefect/variables.py
@@ -3,13 +3,15 @@ from types import ModuleType
 from typing import Optional
 
 from prefect._internal.concurrency.api import create_call, from_sync
-from prefect.client.orchestration import get_client
+from prefect.client.orchestration import PrefectClient
+from prefect.client.utilities import inject_client
+from prefect.utilities.asyncutils import sync_compatible
 
 
-def get(name: str, default: str = None) -> Optional[str]:
+@sync_compatible
+async def get(name: str, default: str = None) -> Optional[str]:
     """
-    Get a variable by name from a sync context.
-
+    Get a variable by name. If doesn't exist return the default.
     ```
         from prefect import variables
 
@@ -17,31 +19,26 @@ def get(name: str, default: str = None) -> Optional[str]:
         def my_flow():
             var = variables.get("my_var")
     ```
-    """
-    variable = get_variable_by_name(name)
-    return variable.value if variable else default
-
-
-async def aget(name: str, default: str = None) -> Optional[str]:
-    """
-    Get a variable by name from an async context.
-
+    or
     ```
         from prefect import variables
 
         @flow
         async def my_flow():
-            var = await variables.aget("my_var")
+            var = await variables.get("my_var")
     ```
     """
-    variable = await _get_variable_by_name(name)
+    variable = get_variable_by_name(name)
     return variable.value if variable else default
 
 
-async def _get_variable_by_name(name):
-    async with get_client() as client:
-        variable = await client.read_variable_by_name(name)
-        return variable
+@inject_client
+async def _get_variable_by_name(
+    name: str,
+    client: PrefectClient,
+):
+    variable = await client.read_variable_by_name(name)
+    return variable
 
 
 def get_variable_by_name(name: str):
@@ -50,6 +47,11 @@ def get_variable_by_name(name: str):
     ).result()
 
     return variable
+
+
+def __getitem__(name: str) -> Optional[str]:
+    # This is a stub for VariablesModule.__getitem__
+    raise NotImplementedError
 
 
 class VariablesModule(ModuleType):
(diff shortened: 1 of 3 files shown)


tests/test_variables.py
```diff
@@ -0,0 +1,66 @@
...
+):
+    model = await create_variable(
+        session,
+        VariableCreate(name="my_variable", value="my-value", tags=["123", "456"]),
+    )
+    await session.commit()
+
+    return model
+
+
```
jakekaplan, line 21: I am not totally sure what else would make sense to test here. Suggestions welcome!

src/prefect/variables.py
```diff
@@ -0,0 +1,41 @@
...
+from prefect.client.orchestration import get_client
+
+
+def get(name: str, default: str = None) -> Optional[str]:
+    """Get a variable by name from a sync context"""
+    variable = get_variable_by_name(name)
+    return variable.value if variable else default
+
+
+async def aget(name: str, default: str = None) -> Optional[str]:
```
jakekaplan, line 15: @madkinsz is this what you had in mind when you said offering something that users could await?

madkinsz: Technically it should probably use `sync_compatible` — we don't really publish two interfaces as a general rule.

src/prefect/variables.py
```diff
@@ -0,0 +1,72 @@
...
+        async def my_flow():
+            var = await variables.aget("my_var")
+    ```
+    """
+    variable = await _get_variable_by_name(name)
+    return variable.value if variable else default
+
+
+async def _get_variable_by_name(name):
+    async with get_client() as client:
```
madkinsz, line 42: Should probably use `inject_client` so it receives optimizations to client retrieval

src/prefect/variables.py
jakekaplan: duh 🤦‍♂️, good call

billpalombi:
Access patterns lgtm!

jakekaplan:
Does this need to get added to `__init__` under the user facing api?

src/prefect/variables.py
```diff
@@ -0,0 +1,74 @@
...
+def get_variable_by_name(name: str):
+    variable = from_sync.call_soon_in_loop_thread(
+        create_call(_get_variable_by_name, name)
+    ).result()
+
+    return variable
+
+
+def __getitem__(name: str) -> Optional[str]:
+    # This is a stub for VariablesModule.__getitem__
```
jakekaplan, line 53: this makes IDE's happy but not sure if this is kosher

src/prefect/variables.py
```diff
@@ -0,0 +1,74 @@
...
+            @flow
+            def my_flow():
+                var = variables["my_var"]
+        ```
+        """
+        variable = get_variable_by_name(name)
+        return variable.value if variable else None
+
+
+sys.modules[__name__].__class__ = VariablesModule
```
madkinsz, line 74: This is a little terrifying. Why not just implement this as a `__getitem__` without injecting it into the modules and defining a class?

madkinsz: Do modules not allow `__getitem__` just `__getattr__` otherwise?

src/prefect/variables.py
```diff
@@ -0,0 +1,74 @@
...
+@inject_client
+async def _get_variable_by_name(
+    name: str,
+    client: PrefectClient,
+):
+    variable = await client.read_variable_by_name(name)
+    return variable
+
+
+def get_variable_by_name(name: str):
```
madkinsz, line 44: Should this be private too for now?

src/prefect/variables.py
madkinsz: I see... https://stackoverflow.com/questions/29064125/add-getitem-to-module

This is quite the hack. Do we care enough about the `["name"]` interface to do this?

src/prefect/variables.py
madkinsz: Cool that this works for type checkers :D nice

src/prefect/variables.py
madkinsz: For the record, this shouldn't hold up the PR. I think we should probably expose this kind of interface in a follow-up so we can figure out if it's worth the possible complications but I'm okay with whatever

madkinsz:
> Does this need to get added to `__init__` under the user facing api?

Might as well add it there. We don't have a clear public/private split as things are setup right now.

src/prefect/variables.py
jakekaplan: @billpalombi are you okay if we remove accessing variables like `variables["my_var"]` for now? I don't think this it's too bad but @madkinsz is right it's probably the safer route to add later vs now and then have to remove it

billpalombi: Yup, I'm totally fine with that.

Issue #9087, opened 2023-04-05
UI - Add artifacts tab (OSS)


znicholasbrown:
Artifacts tab should be added to flow run and task run pages

PR #9109, opened 2023-04-05
Enhancement: Add artifacts tab to flow run page

znicholasbrown:
### Example
Adds an artifacts tab to the flow run page and updates the query we're using to fetch keyed artifacts. Bumps prefect-design and prefect-ui-library packages to latest. 

Related / Example here: https://github.com/PrefectHQ/prefect-ui-library/pull/1316

### DIFF
--- a/ui/src/pages/ArtifactKey.vue
+++ b/ui/src/pages/ArtifactKey.vue
@@ -18,7 +18,6 @@
     ArtifactDescription,
     ArtifactTimeline,
     localization,
-    capitalize,
     useWorkspaceApi
   } from '@prefecthq/prefect-ui-library'
   import { useSubscription, useRouteParam } from '@prefecthq/vue-compositions'
@@ -36,7 +35,7 @@
       return localization.info.artifact
     }
 
-    return `${localization.info.artifact}: ${artifact.value.key ?? capitalize(artifact.value.type)}`
+    return `${localization.info.artifact}: ${artifact.value.key}`
   })
 
   usePageTitle(pageTitle)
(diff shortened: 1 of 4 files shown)


ui/src/pages/ArtifactKey.vue
```diff
@@ -18,25 +18,24 @@
...
   import { computed } from 'vue'
   import { usePageTitle } from '@/compositions/usePageTitle'
 
   const api = useWorkspaceApi()
-  const artifactId = useRouteParam('artifactId')
+  const artifactKey = useRouteParam('artifactKey')
 
-  const artifactSubscription = useSubscription(api.artifacts.getArtifact, [artifactId])
+  const artifactSubscription = useSubscription(api.artifacts.getArtifactCollection, [artifactKey])
   const artifact = computed(() => artifactSubscription.response)
```
pleek91, line 31: Adding a `useArtifact` composition is a nice pattern we've used in other places (`useFlow`, `useFlowRun`, etc) that you might want to adopt for artifacts.

pleek91 (approved):
Comment is unrelated to this PR.

ui/src/pages/ArtifactKey.vue
znicholasbrown: that composition exists, I created it and used it in exactly one place lol 🤦🏻. It won't be usable in this particular instance but there are other places it can be. I'll add a ticket to my backlog to update those and add a collection composition as well

znicholasbrown:
Closes: #9087

Issue #9120, opened 2023-04-06
CancellationCleanup service locking up DB


robfreedy:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Database locking occurs due to the CancellationCleanup loop service. When the PREFECT_API_SERVICES_CANCELLATION_CLEANUP_ENABLED env variable is set to false, the issue goes away. 

![image](https://user-images.githubusercontent.com/104510333/230415335-f5217a37-bf57-4934-8c62-7f6f7ff0b1c0.png)
![image](https://user-images.githubusercontent.com/104510333/230415374-be018a3a-2bfc-41b9-b43d-482f1bb27dff.png)
![image](https://user-images.githubusercontent.com/104510333/230415417-47f123a3-3914-4846-ba08-67bd7f725188.png)
![image](https://user-images.githubusercontent.com/104510333/230416037-265692c6-6d33-41da-afc2-6f2af27b3219.png)

### Reproduction

```python3
Running prefect server with a high number of flow/subflow runs in the Database. Once a large flow run is cancelled, the CancellationCleanup service attempts to do the cleanup on a single connection to the DB. This connection could last a while due to the large number of objects that need to be set to Cancelled state. 

Even when PREFECT_API_SERVICES_CANCELLATION_CLEANUP_LOOP_SECONDS is set to a very high number, the service still runs on server start and ends up locking the DB.
```

### Error

```python3
Error message in last photo of bug summary section (sqlalchemy Invalid Request Error: Can't operate on closed transaction inside of context manager. Please complete the context manager before emitting further commands)
```

### Versions

```Text
2.9.0 (unsure on this but will confirm)
```

### Additional context

Raised by Haddon Korinek (@haddon-korinek-ent)

haddon-korinek-ent:
Sure thing, here's some additional logs:

```
20:43:42.608 | ERROR   | prefect.server - Encountered error during state validation
```

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/prefect/server/orchestration/rules.py", line 237, in validate_proposed_state
    await self._validate_proposed_state()
  File "/usr/local/lib/python3.10/dist-packages/prefect/server/database/dependencies.py", line 119, in async_wrapper
[...]

PR #9124, opened 2023-04-06
Use new sessions and transactions for each query during `CancellationCleanup`

madkinsz:
Attempts to resolve https://github.com/PrefectHQ/prefect/issues/9120 by zealously opening and closing sessions and avoiding transactions for reads.


jakekaplan (approved):
LGTM

Ishankoradia:
Hi @zanieb which version of prefect this is. I can't figure it out. 
I see this has been released under the tag `prefect-sqlalchemy-0.5.0rc2`. What does this mean ? If we upgrade prefect, do the packages need to handled separately

PR #9129, opened 2023-04-06
Include flow run and flow as related resources when emitting events via the events worker

### LINKED ISSUE #9127: Include flow run and flow as related resources when emitting events via the events worker
### First check

### Prefect Version

2.x

Currently when the events worker emits an event that event only includes whatever information was given when the Event was created.

Since we have the flow and task run contexts we should include information from these contexts, when they're available, in the event that's emitted.

### Example Use

_No response_

### Additional context

_No response_

bunchesofdonald:
This changes the EventWorker to implicitly add flow run and flow objects from the run context as related objects when emitting events. This allows for events, such as block instrumentation events, to automatically get the correct related resources without needing to get this information directly.

Closes #9127 

### Checklist

### DIFF
--- a/src/prefect/_internal/concurrency/services.py
+++ b/src/prefect/_internal/concurrency/services.py
@@ -6,8 +6,7 @@ import contextlib
 import sys
 import threading
 from collections import deque
-from contextvars import Context, ContextVar, Token, copy_context
-from typing import Awaitable, Dict, Generic, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Awaitable, Dict, Generic, List, Optional, Type, TypeVar, Union
 
 import anyio
 from typing_extensions import Self
@@ -27,7 +26,7 @@ class QueueService(abc.ABC, Generic[T]):
     _instances: Dict[int, Self] = {}
 
     def __init__(self, *args) -> None:
-        self._queue: Optional[asyncio.Queue[Optional[Tuple[T, Context]]]] = None
+        self._queue: Optional[asyncio.Queue] = None
         self._loop: Optional[asyncio.AbstractEventLoop] = None
         self._done_event: Optional[asyncio.Event] = None
         self._task: Optional[asyncio.Task] = None
@@ -96,8 +95,18 @@ class QueueService(abc.ABC, Generic[T]):
             if not self._started:
                 self._early_items.append(item)
             else:
-                ctx = copy_context()
-                call_soon_in_loop(self._loop, self._queue.put_nowait, (item, ctx))
+                call_soon_in_loop(
+                    self._loop, self._queue.put_nowait, self._prepare_item(item)
+                )
+
+    def _prepare_item(self, item: T) -> T:
+        """
+        Prepare an item for submission to the service. This is called before
+        the item is sent to the service.
+
+        The default implementation returns the item unchanged.
+        """
+        return item
 
     async def _run(self):
         try:
@@ -115,17 +124,14 @@ class QueueService(abc.ABC, Generic[T]):
 
     async def _main_loop(self):
         while True:
-            thing = await self._queue.get()
+            item: T = await self._queue.get()
 
-            if thing is None:
+            if item is None:
                 break
-
-            item, context = thing
 
             try:
                 logger.debug("Service %r handling item %r", self, item)
-                with temporary_context(context=context):
-                    await self._handle(item)
+                await self._handle(item)
             except Exception:
                 logger.exception(
                     "Service %r failed to process item %r", type(self).__name__, item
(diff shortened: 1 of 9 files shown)


chrisguidry (approved):
This is great!  I'm not very familiar with how our context works, but the events look great.  Is the current task run available in the context as well? (not for this PR)

src/prefect/_internal/concurrency/services.py
```diff
@@ -113,18 +115,28 @@ async def _run(self):
 
     async def _main_loop(self):
         while True:
-            item: T = await self._queue.get()
+            thing = await self._queue.get()
```
madkinsz, line 118: lol why `thing`?

src/prefect/_internal/concurrency/services.py
```diff
@@ -113,18 +115,28 @@ async def _run(self):
...
-            if item is None:
+            if thing is None:
                 break
 
+            item, context = thing
+
+            tokens: Dict[ContextVar, Token] = {}
+            for key, value in context.items():
+                token = key.set(value)
+                tokens[key] = token
```
madkinsz, line 128: I was imagining this as a separate utility — okay with just using it here though

tests/_internal/concurrency/test_services.py
```diff
@@ -308,3 +309,22 @@ class IntervalMockBatchedService(MockBatchedService):
...
+
+        async def _handle(self, item):
+            context = copy_context()
+            for key, value in context.items():
+                if key.name == "the-answer":
+                    self.mock(value)
+
+    ContextAwareService.instance().send(1)
+
+    ContextVar("the-answer").set(42)
```
madkinsz, line 326: You're not supposed to declare context vars outside the top-level scope for some reason — I'd declare a constant at the top of the test file.

madkinsz (approved):

src/prefect/events/related.py
```diff
@@ -0,0 +1,94 @@
...
+        task_run_context = TaskRunContext.get()
+        if task_run_context is None:
+            return []
+
+        flow_run_id = task_run_context.task_run.flow_run_id
+
+        if flow_run_id in related_resource_cache:
+            flow_run, flow = related_resource_cache[flow_run_id]
+        else:
+            client = task_run_context.client
```
madkinsz, line 37: It's risky to grab the client off the context like this fyi it'll only work if you're on the same event loop as the client was created in. It's okay in this case but may not be in the future.

src/prefect/events/related.py
```diff
@@ -0,0 +1,94 @@
...
+        exclude = set()
+
+    flow_run_id: Optional[UUID] = None
+    flow_run: Optional["FlowRun"] = None
+    flow: Optional["Flow"] = None
+
+    flow_run_context = FlowRunContext.get()
+
+    if flow_run_context is None:
+        task_run_context = TaskRunContext.get()
```
madkinsz, line 28: Generally we have the task run context take precedence over the flow run context — not the other way around. If you emit an event from a task in a flow without a remote task runner then this looks like it will get skipped?

src/prefect/events/related.py
```diff
@@ -0,0 +1,94 @@
...
+        related.append(
+            RelatedResource(
+                __root__={
+                    "prefect.resource.id": f"prefect.flow-run.{flow_run.id}",
+                    "prefect.resource.role": "flow-run",
+                    "prefect.name": flow_run.name,
+                }
+            )
+        )
+        tags |= set(flow_run.tags)
```
madkinsz, line 69: Why aren't tags getting pulled from the tags context?

bunchesofdonald: Because I forgot it existed, thanks for the reminder!

bunchesofdonald:
> This is great! I'm not very familiar with how our context works, but the events look great. Is the current task run available in the context as well? (not for this PR)

The `TaskRunContext` has the task run attached, but I was just going for the MVP here. Adding that in should be pretty simple.

src/prefect/_internal/concurrency/services.py
bunchesofdonald: :+1: I can make it a context manager.

src/prefect/events/related.py
madkinsz: I'm not sure if we update it with the run's tags — we might just use it to assign the run's tags. Might be an additional datasource.

src/prefect/_internal/concurrency/services.py
madkinsz: I think it's funny but I don't have a better idea right now so 🤷‍♀️

src/prefect/events/related.py
bunchesofdonald: Cool, I'm not going to worry about it right now, but definitely could be something to use in the future.

PR #9136, opened 2023-04-07
Rename `__root_path__` to `__development_base_path__`

madkinsz:
To clarify that this path is not available in production installs (i.e. non-editable) to avoid future problems like https://github.com/PrefectHQ/prefect/pull/9132 — this has bit me before as well.

### DIFF
--- a/src/prefect/cli/project.py
+++ b/src/prefect/cli/project.py
@@ -7,6 +7,7 @@ import typer
 import yaml
 from rich.table import Table
 
+import prefect
 from prefect.cli._types import PrefectTyper
 from prefect.cli._utilities import exit_with_error
 from prefect.cli.root import app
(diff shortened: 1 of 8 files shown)


src/prefect/cli/project.py
```diff
@@ -33,7 +34,7 @@ async def ls():
     List available recipes.
     """
 
-    recipe_paths = Path(__file__).parent / ".." / "projects" / "recipes"
+    recipe_paths = prefect.__module_path__ / "projects" / "recipes"
```
madkinsz, line 37: I don't have strong feelings about this change but it's consistent with the purpose of having the module path in the init.

cicdw: Makes sense to me!

cicdw (approved):
Nice

PR #9145, opened 2023-04-07
Cloud storage recipe improvements

cicdw:
While projects are in beta, we are looking for ways to improve the default experience of using recipes; in this case, when using either of the cloud based recipes, it's important that the push and pull step stay in sync with their bucket and folder references.  Thanks to pull step templating, we can achieve this by default by referencing `push` step outputs in our `pull` step via templated variables.

This PR updates the cloud storage pull steps in our recipes to default to using a templated bucket and folder placeholder

### Checklist


desertaxle (approved):
LGTM!

PR #9149, opened 2023-04-07
Stream `git` output during `git_clone_project`

madkinsz:
Displays git output during clones which is helpful for debugging when this fails or for monitoring progress

e.g.

```
In [1]: from prefect.projects.steps.pull import git_clone_project

In [2]: git_clone_project("prefecthq/prefect")
fatal: repository 'prefecthq/prefect' does not exist
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
Cell In[2], line 1
----> 1 git_clone_project("prefecthq/prefect")

File ~/dev/prefect/src/prefect/projects/steps/pull.py:38, in git_clone_project(repository, branch, access_token)
     35 # Limit git history
     36 cmd += ["--depth", "1"]
---> 38 subprocess.check_call(
     39     cmd, shell=sys.platform == "win32", stderr=sys.stderr, stdout=sys.stdout
     40 )
     41 directory = "/".join(repository.strip().split("/")[-1:]).replace(".git", "")
     42 return {"directory": directory}

File ~/.pyenv/versions/3.11.2/lib/python3.11/subprocess.py:413, in check_call(*popenargs, **kwargs)
    411     if cmd is None:
    412         cmd = popenargs[0]
--> 413     raise CalledProcessError(retcode, cmd)
    414 return 0

CalledProcessError: Command '['git', 'clone', 'prefecthq/prefect', '--depth', '1']' returned non-zero exit status 128.

In [3]: git_clone_project("http://github.com/prefecthq/prefect")
Cloning into 'prefect'...
warning: redirecting to https://github.com/prefecthq/prefect/
remote: Enumerating objects: 1432, done.
remote: Counting objects: 100% (1432/1432), done.
remote: Compressing objects: 100% (1274/1274), done.
remote: Total 1432 (delta 182), reused 699 (delta 93), pack-reused 0
Receiving objects: 100% (1432/1432), 14.65 MiB | 16.43 MiB/s, done.
Resolving deltas: 100% (182/182), done.
Out[3]: {'directory': 'prefect'}
```

PR #9150, opened 2023-04-07
Hide command when access token is provided and `git_clone_project` fails

madkinsz:
Hides the called process error displaying the used command when an access token is provided since the access token is in the command.


cicdw:
Probably worth merging with the steaming output implementation in https://github.com/PrefectHQ/prefect/pull/9149

madkinsz:
@cicdw I figure I'll just resolve the merge conflict after that one is merged then just add tests here.

desertaxle (approved):
LGTM!

Issue #9012, opened 2023-03-31
Caching shouldn't be used when a task has retires and in-non completed state


BitTheByte:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

This issue applies when using `prefect_dask` I'm not sure if it's occurring somewhere else when a task is submitted in a dask cluster and has retires configured if that task fails for any reason e.g. dask just killed the task or even the worker is killed if it has the time to change state to "Crashed" a cache entry of this task will be created and when its resubmitted for execution again e.g. flow restarts it will use the cached "Crashed" state even if it still has some retires.

### Reproduction

```python3
-
```

### Versions

```Text
Version:             2.9.0
API version:         0.8.4
Python version:      3.11.2
Git commit:          69f57bd5
Built:               Thu, Mar 30, 2023 1:08 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         cloud
```

BitTheByte:
https://github.com/PrefectHQ/prefect/issues/8821 is affected by this issue

madkinsz:
Hey @BitTheByte I think this is probably Dask caching the result of the task run not us.

Can you please provide MREs with your issues? We cannot prioritize resolving issues without minimal examples.

BitTheByte:
I think it's prefect as I could see
```
15:13:23.080 | INFO    | Task run 'dummy_task-0' - Task run '6ff789a3-7f56-42e7-aed6-b918414487c3' already finished.
```
being logged to the could, I believe this means dask has executed the future and prefect decided to use its cache. I'm still working on the minimal example, this might take a while as the behavior is unpredictable

madkinsz:
@BitTheByte Ah I see what you mean — this is not a "cache" persay, it's just the state of the run itself. 

@billpalombi / @rpeden perhaps similar to our recent changes around restart transitions we should attempt transitions from CRASHED -> RUNNING if the client finds itself being asked to run a crashed task run.

billpalombi:
Yes, I agree. That's consistent with the choice we made in #8802.

Is it just a matter or removing the rule, or are there other implementation considerations? If its just the rule, this could be a quick win.

BitTheByte:
@billpalombi 

[...]

PR #9152, opened 2023-04-07
Update orchestration rules to allow transitions from terminal states

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9012
Closes https://github.com/PrefectHQ/prefect/issues/9292

Previously, transitions out of terminal states were allowed in very specific cases:
- A task run could move from a failed/crashed/cancelled state to running if the flow run was retrying
- A flow run could move to a scheduled (awaiting retry) state

Now, we allow transitions out of terminal states unless the run is completed _and_ has a persisted result. 

This enables the following behaviors:
- A task run that fails and is orchestrated again will run instead of aborting
- A task run that completes but does not persist its result will run again on flow run retry
- A flow run may be rescheduled without using the "awaiting retry" name
- A flow run that fails and is orchestrated again will run instead of aborting

Note above when I say "fails" I am referring to FAILED, CRASHED, and CANCELLED state types.

We do allow some terminal state transitions still:
- We do not allow runs to transition from a terminal state to CANCELLING, CRASHED, or PAUSED

Additionally, we do not abort the transition if a run is completed and has a persisted result. Instead, we reject the transition and return the terminal state. This matches our semantics for rejections and means in the future the client does not need to capture aborts to retrieve the result of finished runs.

### DIFF
--- a/src/prefect/server/orchestration/core_policy.py
+++ b/src/prefect/server/orchestration/core_policy.py
@@ -610,17 +610,12 @@ class UpdateFlowRunTrackerOnTasks(BaseOrchestrationRule):
 
 class HandleTaskTerminalStateTransitions(BaseOrchestrationRule):
     """
-    Prevents transitions from terminal states.
+    We do not allow tasks to leave terminal states if:
+    - The task is completed and has a persisted result
+    - The task is going to CANCELLING / PAUSED / CRASHED
 
-    Orchestration logic in Prefect REST API assumes that once runs enter a terminal state, no
-    further action will be taken on them. This rule prevents unintended transitions out
-    of terminal states and sents an instruction to the client to abort any execution.
-
-    While rerunning a flow, the client will attempt to re-orchestrate tasks that may
-    have previously failed. This rule will permit transitions back into a running state
-    if the parent flow run is either currently restarting or retrying. The task run's
-    run count will also be reset so task-level retries can still fire and tracking
-    metadata is updated.
+    We reset the run count when a task leaves a terminal state for a non-terminal state
+    which resets task run retries; this is particularly relevant for flow run retries.
     """
 
     FROM_STATES = TERMINAL_STATES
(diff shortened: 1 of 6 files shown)


src/prefect/server/orchestration/core_policy.py
```diff
@@ -631,24 +631,26 @@ async def before_transition(
...
         ):
-            self.original_run_count = context.run.run_count
-            self.original_retry_attempt = context.run.flow_run_run_count
+            await self.reject_transition(None, "This run is already completed.")
+
+        self.original_run_count = context.run.run_count
+        self.original_retry_attempt = context.run.flow_run_run_count
+
+        # Reset run count to reset retries
+        context.run.run_count = 0
```
BitTheByte, line 646: Maybe we could decrease the run count by 1?

Task A(retires=100) executed 50 times and failed if Task A canceled and executed again later it will start counting from 0 again giving it the same 100 retires.

madkinsz: We reset it to zero because we want to reset the task run retries during retry of a flow run. I'm not sure we should complicate it by decrementing the count in other cases. Retries are currently the amount of times a RUNNING -> FAILED proposal should be repeated. If you go from a terminal state back to RUNNING it feels like retries should be reset as this is a "new attempt" to orchestrate the task.

BitTheByte:
Is there an ETA For this?

madkinsz:
@BitTheByte we don't share ETAs. I need to get this passing tests and have other high priority work too.

BitTheByte:
Hey madkinsz, Appreciate the fast reply, I was asking if it will be included in the next release rather than a specific date

madkinsz:
Previously running a previously orchestrated run again would abort

```
❯ PREFECT__FLOW_RUN_ID="e60bcdff-8e4e-44f3-af43-9b152a1cb387" python -m prefect.engine
...
14:05:33.188 | INFO    | prefect.engine - Engine execution of flow run 'e60bcdff-8e4e-44f3-af43-9b152a1cb387' aborted by orchestrator: This run has already terminated.
```

Now the flow run and its tasks will all run again unless their results are persisted.

jakekaplan (approved):
🚀

PR #9154, opened 2023-04-07
Fixes worker start CLI args

EmilRex:
Fixes CLI arguments in `prefect worker start`. Currently `work_queues` and `worker_type` are both assigned to `-t` / `--type`.


madkinsz (approved):
Can we make sure there's test coverage for starting a worker with a queue?

cicdw:
@madkinsz do you want Emil to do this or can we merge this and write a test separately?

PR #9156, opened 2023-04-07
Mock runtime via environment variable

cicdw:
This PR introduces a mechanism for mocking / manually setting runtime attributes via environment variable.  This was requested by a user looking to test their code and provide explicit known values for the attributes.

### Example
To do this, a user only needs to set an appropriately prefixed environment variable, e.g., `PREFECT__RUNTIME__TASK_RUN__NEW_KEY="foobar"` and now `prefect.runtime.task_run.new_key` evaluates to `"foobar"`.

### Checklist

### DIFF
--- a/docs/concepts/runtime-context.md
+++ b/docs/concepts/runtime-context.md
@@ -15,7 +15,7 @@ Prefect tracks information about the current flow or task run with a run context
 The run context itself is full of internal objects used by Prefect to manage execution of your run and is only available in specific situtations. For this reason, we expose a simple interface that only includes the items you care about and dynamically retrieves additional information when necessary. We call this the "runtime context" as it contains information that can only be accessed when a run is happening.
 
 !!! tip "Mock values via environment variable"
-    Oftentimes you may want to mock certain values for testing purposes, for example by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3` you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
+    Oftentimes, you may want to mock certain values for testing purposes.  For example, by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3`, you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
     <div class="terminal">
     ```bash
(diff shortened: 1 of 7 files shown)


docs/concepts/runtime-context.md
```diff
@@ -14,6 +14,16 @@ Prefect tracks information about the current flow or task run with a run context
 
 The run context itself is full of internal objects used by Prefect to manage execution of your run and is only available in specific situtations. For this reason, we expose a simple interface that only includes the items you care about and dynamically retrieves additional information when necessary. We call this the "runtime context" as it contains information that can only be accessed when a run is happening.
 
+!!! tip "Mock values via environment variable"
+    Oftentimes you may want to mock certain values for testing purposes, for example by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3` you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
```
billpalombi, line 18:
```suggestion
    Oftentimes, you may want to mock certain values for testing purposes.  For example, by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3`, you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
```

billpalombi (approved):
This is cool. Suggesting minor copyedits to the tip.

madkinsz:
Just wondering, does `monkeypatch.setattr` not work? That seems more straight-forward than setting an environment variable.

cicdw:
@madkinsz it probably does, but requires users to use `pytest` and be familiar with `monkeypatch` (both reasonable I admit, but this isn't hard to support either and allows us to document it so that users can discover it easier).

madkinsz (approved):

jcozar87:
I've been using this feature for a while and it works like charm! However I discovered a problem with datetime variables: existing datetime context variables like `flow_run.scheduled_start time` are of type `pendulum.datetime.DateTime`. However, if I set a value for `PREFECT__RUNTIME__FLOW_RUN__SCHEDULED_START_TIME` environment variable it must be a string, and it is not parsed as a `pendulum.datetime.DateTime` internally. Maybe for existing content variables it is required to cast environment variables to the expected type?

madkinsz:
@jcozar87 that seems reasonable; feel free to open a PR!

jcozar87:
Thank you for the feedback, I will do it!

Issue #9174, opened 2023-04-10
Pauses on remote infrastructure block (ECS) crash in 2.10.2


jbnitorum:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

We recently upgraded the image our flow is running on from 2.8.6 to 2.10.2.  Using the new image the flow now crashes when an out of process pause is called.  Initially the flow will pause but a few seconds after the state will move to 'crashed' as the infrastructure exits.

We are running an agent on ECS Fargate and the flow is running inside a separate ECS Task as an infrastructure block.  This issue seems to stem from the ECS Task running the flow and not the agent.  

- Using 2.8.6 on both the agent and the flow, the flow will run successfully.
- Using 2.8.6 on the agent and 2.10.2 on the flow, the flow will crash.  
- Using 2.10.2 on the agent and 2.8.6 on the flow, the flow will run successfully.  
- Using 2.10.2 on both the agent and the flow, the flow will crash.

### Reproduction

```python3
`pause_flow_run(flow_run_id = <<current flow run id>>)`
```

### Error

```python3
2023-04-10T17:03:31.880-04:00	21:03:24.212 | INFO | Flow run 'friendly-kingfisher' - Currently paused and suspending execution. Resume before 2023-04-11T15:03:22.626751+00:00 to finish execution.
2023-04-10T17:03:31.880-04:00	21:03:26.234 | ERROR | prefect.engine - Engine execution of flow run '9a6864ef-a946-4275-9e09-775086069d86' exited with unexpected exception
2023-04-10T17:03:31.880-04:00	Traceback (most recent call last):
2023-04-10T17:03:31.880-04:00	File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 2158, in <module>
2023-04-10T17:03:31.880-04:00	enter_flow_run_engine_from_subprocess(flow_run_id)
2023-04-10T17:03:31.880-04:00	File "/usr/local/lib/python3.10/site-packages/prefect/engine.py", line 204, in enter_flow_run_engine_from_subprocess
2023-04-10T17:03:31.880-04:00	).result()
2023-04-10T17:03:31.880-04:00	File "/usr/local/lib/python3.10/site-packages/prefect/client/schemas.py", line 107, in result
[...]

PR #9175, opened 2023-04-10
Fix bug where non-zero status codes would be reported when deployed flow runs paused or failed

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9174


jakekaplan (approved):
Makes sense to me!

PR #9195, opened 2023-04-12
refresh FlowRunTimeline when navigating

brandonreid:
Currently, when viewing a flow run that includes sub-flows, directly navigating to the sub-flow's flow run page doesn't update the graph and you end up viewing stale data. This refreshes the graph so that it's redrawn.

### Checklist


pleek91:
@brandonreid this works but it doesn't fix the root cause. The timeline component needs to react to prop changes without being torn down. Is there something that can be update inside the component to fix that?

Manually updating keys is discouraged. In fact they almost removed the ability to manually set keys in vue3.

Issue #9220, opened 2023-04-14
Prefect.Deployment doesn't have a property pull_steps


davzucky:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

The new deployment using [Project](https://docs.prefect.io/latest/tutorials/projects/) was released with Prefect 2.10.x. This adds the ability not to have to pull flow definition from external storage like it was the case before.

However, to enable that, when we do deployment using code, ```prefect.Deployment``` doesn't have the property ```pull_steps``` that we could set.

### Reproduction

[Prefect Deployment model](https://github.com/PrefectHQ/prefect/blob/b6d0433a15b8172c0f897a6eac2b4d747350fd1f/src/prefect/deployments.py#L249) does not have an argument. However, the function in the same module [load_flow_from_flow_run](https://github.com/PrefectHQ/prefect/blob/b6d0433a15b8172c0f897a6eac2b4d747350fd1f/src/prefect/deployments.py#L171) is using this property.
This is working because this property exists on the ORM layer.
The call to [client.create_deployment](https://github.com/PrefectHQ/prefect/blob/b6d0433a15b8172c0f897a6eac2b4d747350fd1f/src/prefect/deployments.py#L691) will need to be updated as well.

If you are happy with adding that, I will create a pull request

### Error

I cannot set the property  in the deployment object

### Versions

```Text
(venv) dzucker@x1clinux ~/git $ prefect version
Version:             2.10.4
API version:         0.8.4
Python version:      3.10.10
Git commit:          b6d0433a
Built:               Thu, Apr 13, 2023 5:34 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.41.2
```

### Additional context

We need that because we control and run all our deployments using code.

serinamarie:
Hi @davzucky, thanks for the issue! What I'm hearing is that prior to running the `prefect deploy` command, you want to edit the pull step? When running `prefect project init`, the prefect.yaml file is created in your directory with an automatically populated pull step specific to your project. This is then used when creating a deployment:

[...]

PR #9233, opened 2023-04-17
[enhancement]: Add pull_steps to the base Deployment model

davzucky:
Following an issue related to remote storage in Kubernetes @cicdw pointed me to the new concept of ```project```. Looking at the code project introduces a new concept to the deployment model, which is ```pull_steps```. This exists on the ORM model but is not exposed to the ```prefect.Deployment``` model. This change a  the ```pull_steps``` to the base mdel.
Another issue about that topic is open [#9220](https://github.com/PrefectHQ/prefect/issues/9220)

### Example

This will allow to create deployment using ```build_from_flow```

``` python

from my_project.flows import my_flow

      flow=my_flow,
      name="example",
      version="1",
      tags=["demo"],
      pull_steps=[{"prefect.project.steps.pull.set_working_directory": {"directory": "/prefect/workdir"}}]
      )

deployment.apply()
```

### Checklist

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -466,7 +466,7 @@ class Deployment(BaseModel):
     timestamp: datetime = Field(default_factory=partial(pendulum.now, "UTC"))
 
     pull_steps: Optional[List[dict]] = Field(
-        default_factory=list,
+        default=None,
         description="Pull steps for cloning and running this deployment.",
     )
 
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -271,6 +271,7 @@ class Deployment(BaseModel):
             if stored on a local filesystem, an absolute path
         entrypoint: The path to the entrypoint for the workflow, always relative to the `path`
         parameter_openapi_schema: The parameter schema of the flow, including defaults.
+        pull_steps: Pull steps for cloning and running this deployment.
 
     Examples:
 
@@ -324,6 +325,7 @@ class Deployment(BaseModel):
             "schedule",
             "is_schedule_active",
             "infra_overrides",
+            "pull_steps",
         ]
 
         # if infrastructure is baked as a pre-saved block, then
@@ -462,6 +464,11 @@ class Deployment(BaseModel):
         description="The parameter schema of the flow, including defaults.",
     )
     timestamp: datetime = Field(default_factory=partial(pendulum.now, "UTC"))
+
+    pull_steps: Optional[List[dict]] = Field(
+        default_factory=list,
+        description="Pull steps for cloning and running this deployment.",
+    )
 
     @validator("infrastructure", pre=True)
     def infrastructure_must_have_capabilities(cls, value):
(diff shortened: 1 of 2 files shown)


cicdw:
My only hesitation here is that `pull_steps` are not fully compatible with blocks and the logic in `build_from_flow` will currently allow for setting all 3 fields simultaneously which could lead to unexpected outcomes - could you introduce some early error logic within that method to avoid this confusion?

davzucky:
Indeed that makes sense that ```build_from_flow``` can be confused. However, as the project still relies on the deployment object, I can see three options:

1. add pre-check to ensure ```build_from_flow```  that is (```infrastructure```, ``` storage```) are not set when  (```worker_pool```, ```pull_steps```) are set
2. Create a new function ```builld_from_project``` that would limit the field expose and make it clear to be built for the new way to deploy
3. Add a validator for ```pull_steps``` and ```worker_pool``` 

Am I missing something related to the change you are making with ```Project``` that I need to be aware of?
What would be your preference on the implementation side?

cicdw:
To be totally honest, we wanted to entirely avoid a Python SDK for projects so I'm not sure the best way to proceed here - in the same way that you run testing suites from a CLI, not from within one of the test files, it really complicates things to expose ways of deploying projects from within a file (or from within a REPL, which people often try but it fails because there's no natural filesystem to base things on).

davzucky:
I can see where you come from, and keeping the API level small help a lot
Our process today is using prefect flow to set up prefect flow and blocks (like 'Inception") This makes our deployment testable, versioned, and reproducible since everything is embedded in our code.

Today, with Prefect, we had the choice between using Prefect command line,Rest API, or Python Api. If you create a new paradigm, it is essential that we can still have access at all these levels

Looking more at Project, It needs more capabilities today to be helpful.

- We want to deploy once, and this deploys all the flows we need
- Block and deployment need to be able to be deployed together
- Multiple docker registries (Dev to push), and we can only promote to prod one.
- We need to be able to set up which deployment we enable in different environments

Happy to have a chat about how we are doing that.

cicdw:
Returning to this -- I still think https://github.com/PrefectHQ/prefect/pull/9233#issuecomment-1511893127 articulates why we can't merge the PR as-is, but we _are_ actively experimenting with streamlined deployment UX (including Python-first deployments) that will take this feedback into account!

davzucky:
Thank you, @cicdw for the update. Do you have an issue we can follow regarding your current work on that side?

cicdw:
Absolutely @davzucky - this PR is the first major milestone in that Python-first direction: https://github.com/PrefectHQ/prefect/pull/10534

It probably isn't obvious from the PR that it's related, but we are filling in the incremental gaps that lead up to a full worker / work pool setup and hardening the Python SDK for deployments as we go:
- the linked PR introduces a mechanism that allows for Python-based workflows to be deployed without workers or work pools (this is intended to be the easiest way to get a deployment setup for situations that don't need dynamically provisioned infrastructure, so it's all pure Python)
- The next milestone will be to expose a way to handle workflows defined in remote filesystems, but still in the "no dynamic infra" world (i.e., no workers and work pools).  This is when Python configuration for `pull_steps` will show up
- the last step will be to provide mechanisms for transitioning deployments defined in this highly Python-centric way to the dynamic and decoupled worker / work pool world (the current world)

We are targeting end of September for completion of the core components of this work (so roughly one month from now)

PR #9239, opened 2023-04-17
Debugging test_pip.py failures.

bunchesofdonald:
A PR to debug the failures we're seeing with `test_pip.py`.

PR #9241, opened 2023-04-17
Pin pip version in python tests to 23.0.1

madkinsz:
Debugging like https://github.com/PrefectHQ/prefect/pull/9239


madkinsz:
Well let's do this until we get a real fix

madkinsz:
Well let's do this until we get a real fix

bunchesofdonald (approved):

PR #9274, opened 2023-04-20
Fix broken worker events test and include related from configuration

bunchesofdonald:
This fixes the broken `test_worker_emits_cancelled_event` test. It also adds some extra resources to the event now that the `configuration` is available when cancelling flow runs.


madkinsz (approved):

bunchesofdonald:
The tests failures here are in two categories:

1. pip issues which are being addressed in a different PR.
2. A _different_ flaky events test which I'll address in a follow-up PR.

Issue #9242, opened 2023-04-17
Upgrading to pip 23.1 causes tests to fail in CI


bunchesofdonald:
### First check

- [X] I am a contributor to the Prefect codebase

### Description

A new version of pip was released and it appears to cause the following test failures, but only in CI, we've been unable to reproduce locally with Mac or Linux:

- `FAILED tests/software/test_pip.py::test_current_environment_requirements - AssertionError: assert 'prefect' not in ['zipp', 'xmltodict', 'wheel', 'Werkzeug', 'websockets', 'websocket-client', ...]`
- `FAILED tests/software/test_pip.py::test_current_environment_requirements_warns_about_editable_prefect - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.`
- `FAILED tests/software/test_pip.py::test_current_environment_requirements_raises_on_editable_prefect - Failed: DID NOT RAISE <class 'ValueError'>`
- `FAILED tests/software/test_pip.py::test_current_environment_requirements_raises_on_bad_mode - Failed: DID NOT RAISE <class 'ValueError'>`

It appears that somehow the editable install of prefect is making its way into site-packages.

### Impact

This is causing us to need to pin pip to an older version.

PR #9278, opened 2023-04-20
Fix pip requirement inference compatibility with Python 3.11+ and pip 23.1+

madkinsz:
Upgrades from the deprecated `pkg_resources` to `importlib_metadata` for Python 3.11+ compatibility.
Resolves issues inferring if modules are editable for pip 23.1+ compatibility.

Closes https://github.com/PrefectHQ/prefect/issues/9242


bunchesofdonald (approved):
🤞

PR #9309, opened 2023-04-24
bring back logs to flow run page

### LINKED ISSUE #9300: UI: Flow run log tab not shown while flow run is in state pending.
### First check

### Bug summary

Since very recently in Prefect Cloud UI, the log tab for a flow run is only shown after the run transitions into running. No logs tab is shown while in pending.
However, various logs can occur before the task goes into the running state (e.g. logs from a custom infrastructure block, debug logs from prefect, etc.) and it is very convenient to see them right away and not only after the run transitions to running.
This was the case in previous versions of the UI.

### Reproduction

```python3
Run any deployment.
See log lines  flow code from storage at ... (if using storage blocks) and ```Importing flow code from ...``` appear only after the flow run transitions to running.
However, these lines are emitted before the flow run transitions to running.
```

### Error

_No response_

### Versions

```Text
Version:             2.9.0
API version:         0.8.4
Python version:      3.8.15
Git commit:          69f57bd5
Built:               Thu, Mar 30, 2023 1:08 PM
Server type:         cloud
```

### Additional context

While the reproduction example seems a bit pointless, displaying the logs as early as possible is relevant in certain situations.
For example, when using a custom Infrastructure Block that spins up infra for a flow run, this can take quite a while and it is very convenient to see log messages from this infrastructure block as early as possible and not only after all of the infrastructure finished deploying.

brandonreid:
This brings back the "Parent Flow Run" by bumping the `prefect-ui-library` dependency to latest and makes the Logs tab visible to pending flow runs again (pending flow runs can actually have logs in certain instances).

Resolves #9300 

### Checklist

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -9,7 +9,7 @@
       "version": "2.8.0",
       "dependencies": {
         "@prefecthq/prefect-design": "1.7.2",
-        "@prefecthq/prefect-ui-library": "1.5.2",
+        "@prefecthq/prefect-ui-library": "1.5.4",
         "@prefecthq/vue-compositions": "1.4.3",
         "@types/lodash.debounce": "4.0.7",
         "axios": "0.27.2",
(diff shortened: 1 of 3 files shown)


pleek91 (approved):

zhen0 (changes requested):
The logs page change looks good but there's a lot of changes in the package-lock.json. I'm not sure if they're deliberate - if they are let's put them in a separate PR. 

If you want to only update the Prefect-ui-library version, you can use dependabot or npm i @prefecthq/prefect-ui-library --save --save-exact

PR #9130, opened 2023-04-06
Removed Hard-Coded Default Log's Pagination Configuration, instead va…

N3XT14:
## Overview

The proposed change in this pull request involves removing the hard-coded default log pagination configuration and instead fetching values from settings.py. This change will make the code more flexible and easier to maintain, as users can easily modify the log pagination configuration without needing to modify any code.

### Example
```python
from prefect.settings import (
    PREFECT_API_DEFAULT_LOGS_PAGE_SIZE,
    PREFECT_API_DEFAULT_NUM_LOGS,
)

LOGS_DEFAULT_PAGE_SIZE = PREFECT_API_DEFAULT_LOGS_PAGE_SIZE.value()
LOGS_WITH_LIMIT_FLAG_DEFAULT_NUM_LOGS = PREFECT_API_DEFAULT_NUM_LOGS.value()
```
```sh
$ prefect flow-run logs c1b58559-cae1-400e-bb87-6b3827934e95
2023-04-06 16:27:02.656 | INFO    | Flow run 'mellow-antelope' - Created task run 'call_api-0' for task 'call_api'
2023-04-06 16:27:02.657 | INFO    | Flow run 'mellow-antelope' - Executing 'call_api-0' immediately...
2023-04-06 16:27:03.158 | INFO    | Flow run 'mellow-antelope' - Finished in state Completed()
2023-04-06 16:27:03.209 | INFO    | Flow run 'mellow-antelope' - Created task run 'parse_fact-0' for task 'parse_fact'
2023-04-06 16:27:03.210 | INFO    | Flow run 'mellow-antelope' - Executing 'parse_fact-0' immediately...
2023-04-06 16:27:03.349 | INFO    | Flow run 'mellow-antelope' - Finished in state Completed()
2023-04-06 16:27:03.410 | INFO    | Flow run 'mellow-antelope' - Finished in state Completed()

$ prefect flow-run logs c1b58559-cae1-400e-bb87-6b3827934e95 --head
2023-04-06 16:27:02.656 | INFO    | Flow run 'mellow-antelope' - Created task run 'call_api-0' for task 'call_api'
2023-04-06 16:27:02.657 | INFO    | Flow run 'mellow-antelope' - Executing 'call_api-0' immediately...
2023-04-06 16:27:03.158 | INFO    | Flow run 'mellow-antelope' - Finished in state Completed()
[...]

PR #9314, opened 2023-04-24
Fix `flow-run logs` limit

serinamarie:
When using the CLI to retrieve flow run logs, if the `num_logs` to return was >= `LOGS_DEFAULT_PAGE_SIZE`, the limit wasn't being recalculated correctly in the loop. Instead, all logs were being returned if the specified `num_logs` was 200+. With this fix, specifying 205 logs, for example, should return 205 logs.

### Checklist

### DIFF
--- a/src/prefect/cli/flow_run.py
+++ b/src/prefect/cli/flow_run.py
@@ -199,7 +199,9 @@ async def logs(
             num_logs_to_return_from_page = (
                 LOGS_DEFAULT_PAGE_SIZE
                 if user_specified_num_logs is None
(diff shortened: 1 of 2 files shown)


madkinsz:
Does this conflict with https://github.com/PrefectHQ/prefect/pull/9130

serinamarie:
> Does this conflict with #9130

I don't believe so. I think that PR operates on the assumption that there isn't a bug. It does nothing with `num_logs_returned`, which mean `num_logs_to_return_from_page` remains a constant value throughout the while loop.

madkinsz (approved):

PR #9316, opened 2023-04-24
Fix `future.result()` and `future.wait()` calls from async contexts

madkinsz:
Ensures that futures retrieve results from the same event loop as the task runner is managed in. Otherwise, mixed event loops can cause bugs — i.e. the Dask client does not detect that it is being used in asynchronous context because the event loops do not match.

Closes https://github.com/PrefectHQ/prefect-dask/issues/91

### DIFF
--- a/src/prefect/futures.py
+++ b/src/prefect/futures.py
@@ -22,7 +22,7 @@ from uuid import UUID
 
 import anyio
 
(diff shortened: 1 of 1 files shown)


madkinsz:
Flaking tests resolved in #9318

jakekaplan (approved):

PR #9318, opened 2023-04-24
Update `QueueService.send` to wait for the item to be placed in the queue before returning

madkinsz:
Fixes race conditions where items may not be fully sent before the service is drained.

Example flake at https://github.com/PrefectHQ/prefect/actions/runs/4789808505/jobs/8518118439?pr=9316


bunchesofdonald (approved):

madkinsz:
Looks like I introduced a deadlock — probably when `send` is called from the event loop itself.

Issue #9296, opened 2023-04-23
Running two different flows in parallel using multi-threading will raise exception


soleilchang:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Following the previously opened issue [Workflow cannot run in parallel in multi-threading context #9228](https://github.com/PrefectHQ/prefect/issues/9228), I was informed that currently Prefect doesn't support running multiple concurrent runs of a **single** flow instance. 

For testing purposes, two different flows with the same functionality were created. The flow runs are completing successfully as shown in the image below. 
![image](https://user-images.githubusercontent.com/6514675/233832661-501183ba-064c-490a-9639-31c41f6b7411.png)

However, when the script is executed, only the first return value ("workflow1 completed") is printed and an exception is raised, resulting in an unexpected exit, as shown in the error message.

By the way, I am curious if Prefect has plans to support running multiple flow runs of a same flow in parallel (using multi-threading) in the future? We envision a scenario where we create a Web API to trigger a flow. Users should be able to trigger the flow simultaneously, and we don't want the flow run block other requests. Sometimes the optimal solution is to put the flow run in a thread pool, because it's too heavy to use multi-processing and the flow cannot be converted into an asynchronous operation. 

### Reproduction

```python3
from prefect import flow, task
from prefect.task_runners import ConcurrentTaskRunner
import time
from concurrent.futures import ThreadPoolExecutor


@flow(task_runner=ConcurrentTaskRunner())
def test_workflow1():
    time.sleep(5)
    
    return "workflow1 completed"

@flow(task_runner=ConcurrentTaskRunner())
def test_workflow2():
    time.sleep(5)
    
    return "workflow2 completed"

with ThreadPoolExecutor(max_workers=5) as executor:
    t1 = executor.submit(test_workflow1)
    t2 = executor.submit(test_workflow2)
    print(t1.result())
    print(t2.result())
```

### Error

```python3
16:47:55.463 | INFO    | prefect.engine - Created flow run 'colorful-kudu' for flow 'test-workflow1'
16:47:55.489 | INFO    | prefect.engine - Created flow run 'spirited-caiman' for flow 'test-workflow2'
16:48:00.649 | INFO    | Flow run 'colorful-kudu' - Finished in state Completed()
[...]

PR #9343, opened 2023-04-26
Fix race condition in event loop thread start

madkinsz:
`EventLoopThread.start` was not safe to calls from multiple threads.

Reported in https://github.com/PrefectHQ/prefect/issues/9296

### DIFF
--- a/src/prefect/_internal/concurrency/threads.py
+++ b/src/prefect/_internal/concurrency/threads.py
@@ -132,6 +132,7 @@ class EventLoopThread(Portal):
         self._run_once: bool = run_once
         self._submitted_count: int = 0
         self._on_shutdown: List[Call] = []
+        self._lock = threading.Lock()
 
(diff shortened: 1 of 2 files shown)


src/prefect/_internal/concurrency/threads.py
```diff
@@ -140,42 +141,46 @@ def start(self):
...
             self.start()
 
-        if self._shutdown_event.is_set():
-            raise RuntimeError("Worker is shutdown.")
+        with self._lock:
+            if self._submitted_count > 0 and self._run_once:
+                raise RuntimeError(
+                    "Worker configured to only run once. A call has already been"
+                    " submitted."
+                )
```
madkinsz, line 158: Also fixes race conditions with submission / shutdown

serinamarie (approved):
Way to get this code on lock! 🔐

Issue #9345, opened 2023-04-27
Deadlock when running deployments that reconfigure the python logger


j-tr:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Dependencies like mlflow might contain code that directly manipulates the python logger which causes all previously existing handlers to shut down. While this is probably not a good idea when using prefect, these dependencies can be essential and immutable for the developer and it seemed to work in prefect <=2.9.0 when using the RayTaskRunner.

However, in prefect >=2.10.0, the reworked logging introduces a deadlock on the shutdown of the APILogHandler.

This only happens when prefect and prefect_ray are imported before the other dependencies but reordering dependencies in complex systems is not always possible.
This also only affects the execution of deployments. Running the flow as python script works fine.

See reproduction example for expected behavious on 2.9.0 and deadlock on 2.10.0

### Reproduction

```python3
logger_deadlock.py:

import logging.config
from prefect import flow
from prefect import get_run_logger, task
from prefect_ray import RayTaskRunner

# reconfigure the python logger like it is done during
# import mlflow
# see https://github.com/mlflow/mlflow/blob/master/mlflow/utils/logging_utils.py#L61
# this causes all existing handlers to shut down
# see https://github.com/python/cpython/blob/52f96d3ea39fea4c16e26b7b10bd2db09726bd7c/Lib/logging/config.py#L544
logging.config.dictConfig({
    'version': 1,
    'formatters': {
        'standard': {
            'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
        },
    },
    'handlers': {
        'default': {
            'level': 'INFO',
            'formatter': 'standard',
            'class': 'logging.StreamHandler',
            'stream': 'ext://sys.stdout',
        },
    },
})


@task
def task():
    logger = get_run_logger()
    logger.info("info from task")
    # create file to verify that the task was executed when logging isn't working
    open("/home/user/dummy.txt", "w")


@flow(task_runner=RayTaskRunner)
def main() -> None:
    task.submit().result()


if __name__ == "__main__":
    main()

==========================================

$ pip install prefect==2.9.0
$ prefect deployment build logger_deadlock.py:main --name DeadlockTest --apply
[...]

PR #9354, opened 2023-04-27
Add `timeout` to `QueueService.drain_all` and limit `APILogHandler.flush` time to 5s

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9345

I'd prefer a solution that resolves the deadlock entirely, but let's not wait forever when it occurs as a starting point.


madkinsz:
@serinamarie fyi I just added some tests and it required some changes

serinamarie:
> @serinamarie fyi I just added some tests and it required some changes

LGTM! 
In the tests, why is the upper bound on the expected time 2 and 1 instead of the same number of secs?

madkinsz:
Great question; debugging weird problems and forgot to change it. I'll make it a couple seconds to satisfy GHA.

madkinsz:
Closing in favor of smaller change in #9399

desertaxle (approved):

Issue #9251, opened 2023-04-18
Verbose logs due to log_prints capturing non-std prints


aeisenbarth:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Libraries may use print for other purposes than printing to stdout/stderr, for example building strings or writing to files.

Our situation is that we use an upstream library which uses the very common `numba`, and `numba` uses `print` to build a string, causing our Prefect logs to be polluted with _thousands_ of lines, and affected tasks occasionally take a multiple of the expected run time.

The cause is that Prefect's [`print_as_log`](https://github.com/PrefectHQ/prefect/blob/f1cf1109e2db279f16f8257405c80482e21359c0/src/prefect/logging/loggers.py#L203) does not handle an existing `file=` keyword argument but overrides it. When other Python code uses the `file` keyword argument like in `print(something, file=some_file_handle)`, it may expect the prints neither to end up on stdout/stderr nor in any logs.

### Reproduction

```python3
import logging
from io import StringIO

from prefect import flow, task


@task(log_prints=True)
def get_string() -> str:
    with StringIO() as sb:
        print(
            "What is not printed to stdout/stderr should not be logged by Prefect",
            file=sb,
        )
        print("Returning task result, this should be logged")
        return sb.getvalue()


@flow
def mcve():
    return get_string()


def test_log_prints_should_exclude_printing_to_file():
    logger = logging.getLogger("prefect.task_runs")
    with StringIO() as logger_stream:
        handler = logging.StreamHandler(stream=logger_stream)
        logger.addHandler(handler)
        # Run the flow
        mcve()
        log = logger_stream.getvalue()
    assert "should be logged" in log
    assert "should not be logged" not in log
```

### Error

```python3
17:20:05.481 | INFO    | prefect.engine - Created flow run 'amazing-heron' for flow 'mcve'
17:20:05.585 | INFO    | Flow run 'amazing-heron' - Created task run 'get_string-0' for task 'get_string'
17:20:05.586 | INFO    | Flow run 'amazing-heron' - Executing 'get_string-0' immediately...
17:20:05.627 | INFO    | Task run 'get_string-0' - What is not printed to stdout/stderr should not be logged by Prefect
[...]

PR #9358, opened 2023-04-27
Update `log_prints` to ignore prints where a custom `file` is used

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9251

## Example

"bye" is not captured

```python
from prefect import flow
import sys
import io


@flow(log_prints=True)
def example():
    print("hi")
    print("hi", file=None)
    print("hi", file=sys.stdout)
    print("hi", file=sys.stderr)
    print("bye", file=io.StringIO())


example()
```

### DIFF
--- a/src/prefect/logging/loggers.py
+++ b/src/prefect/logging/loggers.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import sys
 from builtins import print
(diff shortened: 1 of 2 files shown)


serinamarie (approved):
LGTM :)

Issue #9193, opened 2023-04-12
Calling flow marked successful even when some of the called subflows have failed.


OptimeeringBigya:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If the input parameter validation of a sub flow fails, the subflow is marked as a failed flow run. 
This failure of sub-flow does not necessarily result in failure of the calling flow.

### Reproduction

```python3
import prefect
from prefect import flow
from datetime import datetime


@flow
def sub_flow(i: datetime):
    print(i)


@flow
def main_flow():
    for i in range(9, 11):
        inp = f"2022-01-0{i} 01:00:00+0000"
        prefect.get_run_logger().info(inp)
        sub_flow(inp, return_state=True)


if __name__ == "__main__":
    main_flow()
```

### Error

```python3
10:02:11.893 | INFO    | prefect.engine - Created flow run 'pastel-squid' for flow 'main-flow'
10:02:19.351 | INFO    | Flow run 'pastel-squid' - 2022-01-09 01:00:00+0000
10:02:24.624 | INFO    | Flow run 'pastel-squid' - Created subflow run 'secret-aardwark' for flow 'sub-flow'
10:02:26.214 | INFO    | Flow run 'secret-aardwark' - Finished in state Completed()
10:02:26.215 | INFO    | Flow run 'pastel-squid' - 2022-01-010 01:00:00+0000
10:02:37.566 | INFO    | Flow run 'pastel-squid' - Created subflow run 'quaint-caterpillar' for flow 'sub-flow'
10:02:37.569 | ERROR   | Flow run 'quaint-caterpillar' - Validation of flow parameters failed with error:
Traceback (most recent call last):
  File "/home/user/source_code/test_project/.venv/lib/python3.10/site-packages/prefect/engine.py", line 509, in create_and_begin_subflow_run
    parameters = flow.validate_parameters(parameters)
  File "/home/user/source_code/test_project/.venv/lib/python3.10/site-packages/prefect/flows.py", line 351, in validate_parameters
    raise ParameterTypeError.from_validation_error(exc) from None
prefect.exceptions.ParameterTypeError: Flow run received invalid parameters:
 - i: invalid datetime format
[...]

PR #9386, opened 2023-05-01
Fix bug where child flows would not fail the parent when they received invalid paremeters

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9193


serinamarie (approved):
LGTM :)

PR #9387, opened 2023-05-01
resolve_futures_to_data should raise on failure by default 

Sage-2001:
### Checklist


src/prefect/futures.py
```diff
@@ -336,11 +336,17 @@ async def resolve_futures_to_data(
...
             ).aresult()
             for future in futures
         ]
     )
+    return visit_collection(
+        expr,
+        visit_fn=partial(_replace_futures_with_results, results),
+        return_data=True,
+        context={},
+    )
```
madkinsz, line 349: We should not return early here, why this change?

src/prefect/futures.py
```diff
@@ -336,11 +336,17 @@ async def resolve_futures_to_data(
         *[
             # We must wait for the future in the thread it was created in
             from_async.call_soon_in_loop_thread(
-                create_call(future._result, raise_on_failure=False)
+                create_call(future._result, raise_on_failure=raise_on_failure)
```
madkinsz, line 339: 👍 but you'll need to add `raise_on_failure` as a parameter to the `resolve_futures_to_data` as well

madkinsz (changes requested):
Thanks for the PR! You'll need to make some more changes and write a test in `test_futures.py` that checks that it does / does not raise on failure as intended.

Sage-2001:
> Please try to write a descriptive title in the future. Can you link to the issue this will close as well as described in the PR template?

Ok i will from next time.

src/prefect/futures.py
Sage-2001: The change was made because the visit_collection function was being called inside the if statement, which could potentially result in the function returning early and not executing the desired behavior of replacing futures with their results. By moving the function call outside of the if statement, the visit_collection function will always be called and replace futures with their results, even if maybe_expr is None.

In other words, by removing the early return statement, we ensure that the function always performs the desired behavior of replacing futures with their results, rather than potentially returning early and not performing the desired behavior.

madkinsz: This doesn't make sense, there's not an `if` statement here and this returns before L356 -> L373 which are essential for this function.

src/prefect/futures.py
```diff
@@ -310,13 +310,16 @@ def _collect_futures(futures, expr, context):
...
-    expr: Union[PrefectFuture[R, Any], Any]
+    expr: Union[PrefectFuture[R, Any], Any],
+    raise_on_failure: bool = True
 ) -> Union[R, Any]:
     """
-    Given a Python built-in collection, recursively find `PrefectFutures` and build a
-    new collection with the same structure with futures resolved to their results.
-    Resolving futures to their results may wait for execution to complete and require
-    communication with the API.
+    Recursively walks an expression tree and replaces any PrefectFutures with their results.
```
madkinsz, line 317: Please don't change this docstring.

Sage-2001: Ok

madkinsz:
This looks broken still and you'll need to add a test covering the change in behavior. Let us know if you have questions!

PR #9399, opened 2023-05-02
Limit `APILogHandler.flush` time to 5s when called from synchronous contexts

madkinsz:
Alternative to https://github.com/PrefectHQ/prefect/pull/9354 with reduced change scope.

Closes https://github.com/PrefectHQ/prefect/issues/9345

I'd prefer a solution that resolves the deadlock entirely, but let's not wait forever when it occurs as a starting point.


serinamarie (approved):

madkinsz:
Returning to the larger changes inhttps://github.com/PrefectHQ/prefect/pull/9354 :(

PR #9426, opened 2023-05-03
Fix send/drain race conditions in queue services

madkinsz:
Better implementation of #9318
Effectively reverts https://github.com/PrefectHQ/prefect/pull/9398
Fixes race conditions seen in work like https://github.com/PrefectHQ/prefect/pull/9354

- Adds a lock that is taken on `drain_all` to ensure it captures all of the instances
- Change from an `asyncio.Queue` with calls to put items into it scheduled on the event loop to a `queue.Queue` with calls to retrieve items from it run in worker threads. This ensures the item enters the queue before `send` returns without risking deadlock at the cost of creating a worker thread on each retrieval from the queue.

### DIFF
--- a/src/prefect/_internal/concurrency/services.py
+++ b/src/prefect/_internal/concurrency/services.py
@@ -1,18 +1,20 @@
 import abc
+import queue
 import asyncio
 import atexit
 import concurrent.futures
 import contextlib
 import sys
 import threading
-from collections import deque
+from functools import partial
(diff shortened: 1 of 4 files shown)


jakekaplan (approved):
Makes sense to me!

madkinsz:
Hm I cannot replicate these event worker failures with flake finder locally.

bunchesofdonald (approved):
That looks like a much better solution.

PR #9434, opened 2023-05-04
Deprecate `prefect kubernetes manifest` command

enriched-h:
### Checklist

### DIFF
--- a/src/prefect/cli/kubernetes.py
+++ b/src/prefect/cli/kubernetes.py
@@ -24,9 +24,14 @@ kubernetes_app = PrefectTyper(
 )
 app.add_typer(kubernetes_app)
 
+# Deprecate kubernetes maniest
 manifest_app = PrefectTyper(
     name="manifest",
-    help="Commands for generating Kubernetes manifests.",
+    help="Deprecated. Use the 'Helm charts' instead.",
+    deprecated=True,
+    deprecated_name="prefect kubernetes manifest",
+    deprecated_start_date="4th May 2023",
+    deprecated_help="Use the 'Helm charts' instead.",
 )
 kubernetes_app.add_typer(manifest_app)
 
@@ -184,3 +189,6 @@ async def manifest_flow_run_job():
     )
 
     print(output)
+"""
+
+"""


src/prefect/cli/kubernetes.py
```diff
@@ -24,9 +24,14 @@
...
 app.add_typer(kubernetes_app)
 
+# Deprecate kubernetes maniest
 manifest_app = PrefectTyper(
     name="manifest",
-    help="Commands for generating Kubernetes manifests.",
+    help="Deprecated. Use the 'Helm charts' instead.",
+    deprecated=True,
+    deprecated_name="prefect kubernetes manifest",
+    deprecated_start_date="4th May 2023",
```
madkinsz, line 33: I believe this expects month and year but not day

src/prefect/cli/kubernetes.py
```diff
@@ -184,3 +189,6 @@ async def manifest_flow_run_job():
     )
 
     print(output)
+"""
+
+"""
```
madkinsz, line 194: Looks unintentional

src/prefect/cli/kubernetes.py
```diff
@@ -24,9 +24,14 @@
 )
 app.add_typer(kubernetes_app)
 
+# Deprecate kubernetes maniest
 manifest_app = PrefectTyper(
     name="manifest",
-    help="Commands for generating Kubernetes manifests.",
+    help="Deprecated. Use the 'Helm charts' instead.",
```
madkinsz, line 30: I think that we can retain the previous help as this should be covered by the `deprecated_help`. Maybe 
```suggestion
    help="Commands for generating Kubernetes manifests. Deprecated.",
```

src/prefect/cli/kubernetes.py
```diff
@@ -24,9 +24,14 @@
...
+    deprecated=True,
+    deprecated_name="prefect kubernetes manifest",
+    deprecated_start_date="4th May 2023",
+    deprecated_help="Use the 'Helm charts' instead.",
```
madkinsz, line 34: I think we can give a better suggestion here — maybe a link to the Helm charts?

cc @jawnsy

madkinsz:
Can you update the body of the PR to:
- Describe the change
- Link to the issue this closes
- Include an example of running the command and the deprecation warning

chrisguidry:
Thank you for your contribution, @enriched-h, we have two PRs open for the same issue, so we're going to unify around #9479.

PR #9460, opened 2023-05-05
Only touch the ignorefile if it doesn't exist

### LINKED ISSUE #9332: Prefect 2.10.3 introduces write access requirement for deployment creation
### First check

### Bug summary

After upgrading to `2.10.3`, some users have found that their CI pipelines are breaking. The issue appears to be caused by an attempt to write a file during the creation of a deployment via Python. Specifically, the problematic line of code is here: https://github.com/PrefectHQ/prefect/blob/490b7ab55b6667e45ed845c87326c9bb23fbd1c8/src/prefect/deployments.py#L788-L789

In the flow registration step, if the environment doesn't have write access, creating the `.prefectignore` file fails.

### Reproduction

```python3
prefect deployment build path/to/file.py:entrypoint
```

### Error

_No response_

### Versions

```Text
2.10.3
```

### Additional context

A potential solution would be to improve the default handling for cases where the `.prefectignore` file does not exist. Instead of attempting to create the file and causing an error in environments with restricted write access, the code could proceed with a default behavior, such as using an empty ignore list or a pre-defined set of ignore patterns.

EmilRex:
Touching the `.prefectignore` file modifies it even if the file already exists which breaks some read-only CI environments. This change simply performs an existence check before touching the file.

Closes #9332

### DIFF
--- a/src/prefect/deployments.py
+++ b/src/prefect/deployments.py
@@ -786,7 +786,8 @@ class Deployment(BaseModel):
         deployment.parameter_openapi_schema = parameter_schema(flow)
 
         # ensure the ignore file exists
-        Path(ignore_file).touch()
+        if not Path(ignore_file).exists():
+            Path(ignore_file).touch()
 
         if not deployment.version:
             deployment.version = flow.version


madkinsz:
Would it be easy to add a test with a read-only directory to prevent further regressions here?

EmilRex:
> Would it be easy to add a test with a read-only directory to prevent further regressions here?

Not sure about that, but mind if I open up an issue to do it as a follow up? Would love to get this merged sooner rather than later.

PR #9464, opened 2023-05-05
Handle deleted deployments gracefully

cicdw:
Right now if a user deletes a deployment associated with a new work pool, workers trying to cancel its runs will error out with a 404.  This PR updates workers to handle this situation gracefully and mark the flow run as failed.

### Checklist

### DIFF
--- a/tests/workers/test_base_worker.py
+++ b/tests/workers/test_base_worker.py
@@ -1490,6 +1490,32 @@ class TestCancellation:
     @pytest.mark.parametrize(
         "cancelling_constructor", [legacy_named_cancelling_state, Cancelling]
     )
+    async def test_worker_cancel_run_handles_missing_deployment(
+        self,
+        orion_client: PrefectClient,
+        worker_deployment_wq1,
+        cancelling_constructor,
+        work_pool,
+    ):
+        flow_run = await orion_client.create_flow_run_from_deployment(
+            worker_deployment_wq1.id,
+            state=cancelling_constructor(),
+        )
+
+        await orion_client.delete_deployment(worker_deployment_wq1.id)
+
+        async with WorkerTestImpl(
+            work_pool_name=work_pool.name, prefetch_seconds=10
+        ) as worker:
+            await worker.sync_with_backend()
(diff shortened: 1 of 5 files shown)


serinamarie:
LGTM :) Would it make sense to handle a test for this?

cicdw:
@serinamarie test added 👍

tests/workers/test_base_worker.py
```diff
@@ -1487,6 +1487,32 @@ async def test_worker_cancel_run_updates_state_type(
...
+
+        async with WorkerTestImpl(
+            work_pool_name=work_pool.name, prefetch_seconds=10
+        ) as worker:
+            await worker.sync_with_backend()
+            await worker.check_for_cancelled_flow_runs()
+
+        post_flow_run = await orion_client.read_flow_run(flow_run.id)
+        assert post_flow_run.state.type == StateType.CANCELLED
+
```
serinamarie, line 1515: You mention in the description that this "updates workers to handle this situation gracefully and mark the flow run as failed." Maybe I'm not understanding the test correctly but shouldn't the flow run have a `FAILED` state then?

cicdw: @serinamarie ah good catch - yea I updated the logic after opening the PR; because the flow run actually _might_ be a real run that is still running, I wanted to allow the rest of the cancellation logic to proceed which results in a CANCELLED state

serinamarie: @cicdw Got it! Is the logger warning still accurate then about the flow run being unable to be cancelled by the worker?

serinamarie (approved):
Hey @cicdw this looks good to me. My question about the logger warning is non-blocking.

madkinsz:
@cicdw Is this ready for merge?

cicdw:
Sorry y'all! Yes this is ready if you're good with it 👍

tests/workers/test_base_worker.py
cicdw: Yea I think it is - somehow the worker was told to cancel this, but can't fulfill that

Issue #9459, opened 2023-05-05
Deployment CLI Command not returning


toro-berlin:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

With the most recent Prefect version (2.10.7), the deployment command is not returning anymore. This means our CI Pipeline times out and does not execute the consecutive jobs.

### Reproduction

```python3
# Flow Code
from prefect import flow
@flow
def main() -> None:
    pass

if __name__ == "__main__":
    main()

# command to deploy flow
prefect deployment build ./example.py:main --name tobias-test-tobias-test --work-queue work-queue:default --infra-block kubernetes-job/tobias-test-tobias-test --storage-block s3/tobias-test-tobias-test --apply
```

### Error

```python3
# command output 2.10.7 (not returning)
Found flow 'main'
17:00:36.165 | INFO    | prefect._internal.concurrency.threads - Exiting worker thread 'WorkerThread-0'
17:00:36.167 | INFO    | prefect._internal.concurrency.threads - Exiting worker thread 'WorkerThread-1'
17:00:40.203 | INFO    | prefect._internal.concurrency.threads - Exiting worker thread 'WorkerThread-2'
17:00:40.212 | INFO    | prefect._internal.concurrency.threads - Exiting worker thread 'WorkerThread-3'
Deployment YAML created at 
'abc/main-deployment.yaml'.
17:00:42.524 | INFO    | prefect._internal.concurrency.threads - Exiting worker thread 'WorkerThread-4'
Successfully uploaded 17 files to s3://abc-prefect-storage/tobias-test/tobias-test/
Deployment 'main/tobias-test-tobias-test' successfully created with id 
'99cd2518-22b5-493b-8a5c-16b5cb700c9f'.

To execute flow runs from this deployment, start an agent that pulls work from the 'default-agent-pool' 
work pool:
$ prefect agent start -p 'default-agent-pool'


# command output 2.10.6 (returning)
Found flow 'main'
Default '.prefectignore' file written to 
abc/.prefectignore
Deployment YAML created at 
'abc/main-deployment.yaml'.
Successfully uploaded 17 files to s3://abc-prefect-storage/tobias-test/tobias-test/
Deployment 'main/tobias-test-tobias-test' successfully created with id 
[...]

PR #9481, opened 2023-05-08
Update `QueueService` to use a dedicated worker thread

madkinsz:
This reduces the overhead of spawning a new thread on each `queue.get()` call and allows us to set the worker thread as a daemon resolving a deadlock on interpreter exit.

Closes https://github.com/PrefectHQ/prefect/issues/9459


madkinsz:
Tested resolution of bug with `prefect deployment build flows/hello_world.py:hello --name 'test' --apply` which previously hung >50% of the time

PR #9484, opened 2023-05-08
Add release notes for 2.10.8

madkinsz:
[Rendered](https://github.com/PrefectHQ/prefect/blob/release-notes-2.10.8/RELEASE-NOTES.md#release-2108)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,38 @@
 # Prefect Release Notes
 
+## Release 2.10.8
+
+### Flow run orchestration rule updates
+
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,38 @@
 # Prefect Release Notes
 
+## Release 2.10.8
+
+### Flow run orchestration rule updates
+
+A flow run orchestration rule which was previously intended to prevent backwards transitions is updated in this release to allow most transitions. Now, it only prevents some transitions to `PENDING` states to prevent race conditions during handling of runs by multiple agents or workers. This improves orchestration behavior during infrastructure restarts. For example, when a Kubernetes pod is interrupted the flow run can be rescheduled on a new pod by Kubernetes. Previously, Prefect would abort the run as it attempted to transition from a `RUNNING` -> `RUNNING` state. Now, we will allow this transition and your flow run will continue.
```
billpalombi, line 7:
```suggestion
A flow run orchestration rule which was previously intended to prevent backwards transitions is updated in this release to allow most transitions. Now, it only prevents some transitions to `PENDING` states to prevent race conditions during handling of runs by multiple agents or workers. This improves orchestration behavior during infrastructure restarts. For example, when a Kubernetes pod is interrupted, the flow run can be rescheduled on a new pod by Kubernetes. Previously, Prefect would abort the run as it attempted to transition from a `RUNNING` to a `RUNNING` state. Now, Prefect will allow this transition and your flow run will continue.
```

billpalombi (approved):
lgtm, minor suggestions

PR #9486, opened 2023-05-08
Use Python 3.10 to build docs

abrookins:
Our docs build is broken because mkdocs-material uses syntax that's only compatible with Python 3.10. As a quick solution, I've updated the version of Python that we use for the build.

### Example
NA

### Checklist


madkinsz:
#9499 / #9501

PR #9502, opened 2023-05-09
Update `clear_db` to be an opt-in fixture

madkinsz:


madkinsz:
@serinamarie / @chrisguidry 

This is a possible option for faster test runs. Another could be splitting the test suite into directories that need a fresh database and ones that don't. Right now, running a single test that does not touch the database can take >15s just to create the database.

PR #9521, opened 2023-05-10
Correctly pass optional parameters to API in create_work_queue

andrewbrannan:
I noticed that when using the `create_work_queue` method in the python client that the work queue would get created but without any of the parameters that were set to non-default values. I guess this was an oversight or rebase issue when the params were [first added here](https://github.com/PrefectHQ/prefect/commit/839cc3e#diff-ed60480b777e77bdeeced26ed733ea8d0da54e3da5a6dce53054258a1625b004R696).

This PR just builds the API request from the object that I assume should have been used all along and adds a test to check the extra params were used.

### Example

```python
# this was succeeding but the work queue was being created without the concurrency limit
async with get_client() as prefect_client:
    await prefect_client.create_work_queue(
        name="my-queue",
        concurrency_limit=1,
    )
```

### Checklist

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -804,7 +804,8 @@ class PrefectClient:
             create_model.concurrency_limit = concurrency_limit
         if priority is not None:
             create_model.priority = priority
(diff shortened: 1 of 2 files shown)


madkinsz (approved):
Thank you!

PR #9531, opened 2023-05-11
Update worker to be able to include itself as a related resource

bunchesofdonald:
This adds an `include_self` kwarg to the Worker's `_event_related_resources` method that will make the worker include itself as a related resource when this method is called. This should be a more elegant solution for the downstream libraries instead of the [3 line dance](https://github.com/PrefectHQ/prefect-kubernetes/blob/main/prefect_kubernetes/events.py#L51-L53) they do now.


chrisguidry (approved):
Aha makes good sense!

PR #9498, opened 2023-05-09
Fix bug where SIGTERM was not properly captured as a flow run crash

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9403

When we moved the engine to a dedicated thread, we lost the ability to register signal handlers. Here we add signal handler registration to the main thread before entering the engine which is important for crash detection of SIGTERM and consequently reports of cancellation.

I implemented this by adding support for context managers to the concurrency API because async flow calls return a coroutine and we need to enter the contexts inside that coroutine or it'll exit before it actually runs.

## Example

```
12:34:14.097 | DEBUG   | prefect.profiles - Using profile 'default'
12:34:14.576 | DEBUG   | prefect.client - Using ephemeral application with database at sqlite+aiosqlite:////Users/mz/.prefect/prefect.db
12:34:14.744 | INFO    | prefect.engine - Created flow run 'nifty-avocet' for flow 'test-flow'
12:34:14.744 | DEBUG   | Flow run 'nifty-avocet' - Starting 'ConcurrentTaskRunner'; submitted tasks will be run concurrently...
12:34:14.744 | DEBUG   | prefect.task_runner.concurrent - Starting task runner...
12:34:14.745 | DEBUG   | prefect.client - Using ephemeral application with database at sqlite+aiosqlite:////Users/mz/.prefect/prefect.db
12:34:14.853 | DEBUG   | Flow run 'nifty-avocet' - Executing flow 'test-flow' for flow run 'nifty-avocet'...
12:34:14.853 | DEBUG   | Flow run 'nifty-avocet' - Beginning execution...
12:34:21.704 | DEBUG   | prefect.task_runner.concurrent - Shutting down task runner...
12:34:21.705 | ERROR   | Flow run 'nifty-avocet' - Crash detected! Execution was interrupted by an unexpected exception: TerminationSignal

12:34:21.707 | DEBUG   | Flow run 'nifty-avocet' - Crash details:
Traceback (most recent call last):
  File "/Users/mz/dev/prefect/src/prefect/engine.py", line 1759, in report_flow_run_crashes
    yield
  File "/Users/mz/dev/prefect/src/prefect/engine.py", line 415, in begin_flow_run
    interruptible=flow.timeout_seconds is not None,
  File "/Users/mz/dev/prefect/src/prefect/engine.py", line 708, in orchestrate_flow_run
    result = await flow_call.aresult()
[...]

PR #9543, opened 2023-05-11
Fix bug where `SIGTERM` as not reported as a flow run crash for deployed runs

madkinsz:
Follow-up to #9498 that adds the same context management to the engine entrypoint for deployed flow runs.

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -208,7 +208,8 @@ def enter_flow_run_engine_from_subprocess(flow_run_id: UUID) -> State:
     setup_logging()
 
     state = from_sync.wait_for_call_in_loop_thread(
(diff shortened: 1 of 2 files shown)


src/prefect/engine.py
```diff
@@ -208,7 +208,8 @@ def enter_flow_run_engine_from_subprocess(flow_run_id: UUID) -> State:
     setup_logging()
 
     state = from_sync.wait_for_call_in_loop_thread(
-        create_call(retrieve_flow_then_begin_flow_run, flow_run_id)
+        create_call(retrieve_flow_then_begin_flow_run, flow_run_id),
+        contexts=[capture_sigterm()],
```
madkinsz, line 212: We added this to `enter_flow_run_engine_from_flow_call` but not here 😬

serinamarie (approved):

Issue #9535, opened 2023-05-11
CronSchedule validation fails when provided with a timezone


mferryRV:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Version ==2.10.8

Supplying a timezone to CronSchedule causes a validation failure.

### Reproduction

```python3
>>> from prefect.server.schemas.schedules import CronSchedule
>>> schedule = CronSchedule(cron="0 7 * * *", timezone="America/New_York", day_or=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CronSchedule
timezone
  argument of type 'function' is not iterable (type=type_error)
```

### Error

```python3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CronSchedule
timezone
  argument of type 'function' is not iterable (type=type_error)
```

### Versions

```Text
2.10.8
```

### Additional context

This appears to originate in the validation function:
```
    @validator("timezone")
    def valid_timezone(cls, v):
        if v and v not in pendulum.tz.timezones:
            raise ValueError(f'Invalid timezone: "{v}"')
        return v
```

This appears to be an error in how pendulum.tz.timezones is referenced without making a function call:
```
>>> import pendulum
>>> pendulum.tz.timezones
<function timezones at 0x10d923370>
>>> pendulum.tz.timezones()
('Africa/Algiers', 'Atlantic/Cape_Verde',
```

madkinsz:
@mferryRV want to submit a fix? Looks pretty quick!

ajwadjaved:
Hi, this was my first open source contribution. Please let me know if I did anything wrong or if my code didn't fix the error. I wasn't getting the error anymore when I ran the code locally.

rkscodes:
Should I work on it, seems stale at this point

jawnsy:
I'm not sure if this was a change in pendulum's API, but creating a CronSchedule works for me.

[...]

PR #9546, opened 2023-05-11
Fix Issue #9535 - CronSchedule validation fails when provided with a …

ajwadjaved:
Fixes simple bug where providing timezone to CronSchedule caused an error
  

### Checklist

### DIFF
--- a/src/prefect/server/schemas/schedules.py
+++ b/src/prefect/server/schemas/schedules.py
@@ -230,7 +230,7 @@ class CronSchedule(PrefectBaseModel):
         extra = "forbid"
 
     cron: str = Field(default=..., example="0 0 * * *")
-    timezone: Optional[str] = Field(default=None, example="America/New_York")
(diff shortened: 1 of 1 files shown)


madkinsz (changes requested):
Thanks for contributing!

This field should be a string not a list. Isn't the bug that we're not calling `timezones` on L243?

Please add a test that demonstrates the bug is fixed.

ajwadjaved:
Hey sorry, what do you mean by L234?
  
I can try working on it if you specify and then write a test for it. Changing the field name to list had solved the error on my end, I originally thought it was just a type hinting error which was resulting in the value not being iterable

jawnsy:
Thanks so much for your contribution!

We cannot reproduce the original issue, so I'm closing this out. See https://github.com/PrefectHQ/prefect/issues/9535#issuecomment-1601388508 for details

Issue #8627, opened 2023-02-23
Add link to dashboard on flow run creation


seajhawk:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

When my code runs, I see no indication of which profile is being used. 
![image](https://user-images.githubusercontent.com/18316676/220934816-2bef6b91-0d92-43fc-ae07-7e80290f2039.png)

### Describe the proposed behavior

As I'm testing, and switching back and forth between local and cloud, it would be extremely helpful to see the name of the profile and have a link I can click to open the dashboard. 

### Example Use

**06:15:59.830 | INFO    | Profile - using profile 'local' - view dashboard at http://127.0.0.1:4200**
06:15:59.830 | INFO    | prefect.engine - Created flow run 'scrupulous-lion' for flow 'log-flow'
06:16:01.830 | INFO    | Flow run 'scrupulous-lion' - Created task run 'log_task-0' for task 'log_task'
06:16:01.834 | INFO    | Flow run 'scrupulous-lion' - Executing 'log_task-0' immediately...
06:16:02.365 | INFO    | Task run 'log_task-0' - Hello ID!
06:16:02.366 | INFO    | Task run 'log_task-0' - Prefect Version = 2.8.2 🚀
06:16:02.367 | WARNING | Task run 'log_task-0' - You better hurry! You're going to be late!
06:16:02.582 | INFO    | Task run 'log_task-0' - Finished in state Completed()
06:16:02.767 | INFO    | Flow run 'scrupulous-lion' - Finished in state Completed('All states completed.')

madkinsz:
@seajhawk would it suffice to have a link to the flow run on creation?

charris-msft:
@madkinsz - I believe such a link would be helpful, but not satisfy my personal need unless that link contains the profile name in a form that is easy to spot.

Thanks!

Sam-ScottLogic:
May I work on this?

Sam-ScottLogic:
Hi there! is the dashboard URL the same every time the code is run (http://127.0.0.1:4200/) @madkinsz

Sam-ScottLogic:
May I be assigned to this?

madkinsz:
Hey @Sam-ScottLogic the dashboard URL is not constant. Here's an example of constructing a URL to the dashboard:

[...]

PR #9551, opened 2023-05-12
Add link to dashboard on flow run

Sam-ScottLogic:
### Overview
In this branch I have added debug level logging of the URL to the dashboard as described within #8627. I extended the work completed within #8856 as I wasnt too sure where exactly the logging needed to be added but seemed to be needed to be logged on the same line.

### Example (Screenshot)
image

In the run above PREFECT_UI_URL is not set I believe cause  ephemeral mode is true which means PREFECT_UI_URL is set to none.

### Checklist

Closes - #8627


src/prefect/__init__.py
```diff
@@ -83,8 +85,15 @@
...
+if PREFECT_UI_URL:
+    dashboard_url = PREFECT_UI_URL.value()
+else:
+    dashboard_url = "<no dashboard available>"
+
 prefect.logging.get_logger("profiles").debug(
-    f"Using profile {prefect.context.get_settings_context().profile.name!r}"
+    f"Using profile {prefect.context.get_settings_context().profile.name!r} - View"
+    f" dashboard at {dashboard_url!r}"
 )
```
madkinsz, line 97: This is the right idea but I don't think we should do this on import. Instead:

- We want to do this in `engine.py` after creating a flow run
- We'll want to include the link to the running flow run in the dashboard
- We'll want to use an INFO log level

madkinsz:
Implemented at https://github.com/PrefectHQ/prefect/pull/9740

PR #9561, opened 2023-05-14
Feature/cast mocked runtime attributes

jcozar87:
# Cast environment variable for known mocked runtime attributes
    
The PR #9156 introduces a mechanism for mocking / manually setting runtime attributes via environment variable.
In commit https://github.com/asmundo/prefect/commit/a56c7a1457581501c78ba250193bbbe58dc1c276 this functionality is extended to deployment and task_run runtime attributes.
    
The environment variables substitute the real attribute values in case that environment variable names matches with the name of some flow_run, task_run or deployment attribute.

However, if the type of the attribute is not str, that would lead to a type insconsistency (i.e. flow_run.scheduled_start_time is of type pendulum.DateTime, but environment variable PREFECT__RUNTIME__FLOW_RUN__SCHEDULED_START_TIME is of type str). Now cast environment variable types to its real attribute type if exists. The functionality has been extended to flow_run, task_run and deployments, however only flow_run.scheduled_start_time is of a type different than str. Code is ready to adapt for any of these types: bool, int, float, str and pendulum.DateTime. It would be easy to extend to more attribute types.

A new test has been developed in `tests/runtime/test_flow_run.py::TestAttributeAccessPatterns::test_scheduled_start_time_override_via_env_var`, to check that the new functionality correctly cast types.

### Example

```python
import os
from prefect import flow
os.environ["PREFECT__RUNTIME__FLOW_RUN__SCHEDULED_START_TIME"] = "2023-05-13 20:00:00"

@flow(log_prints=True)
def test_schedule_start_time_flow():
    print(f"The type of flow_run.scheduled_start_time attribute ({prefect.runtime.flow_run.scheduled_start_time}) is {type(prefect.runtime.flow_run.scheduled_start_time)}")

test_schedule_start_time_flow()  # The type of flow_run.scheduled_start_time attribute (2023-05-13 20:00:00) is <class 'pendulum.datetime.DateTime'>
```

### Checklist


madkinsz:
@jcozar87 We don't use a `dev` branch — feel free to change the base to `master`. Ping me if you have difficulty getting the history correct. We can't review it with all the extra commits.

jcozar87:
Perfect! I changed the base branch to main. Now you can review just my single commit. Tell me if you need me to change anything! Thank you for your time!

src/prefect/runtime/flow_run.py
```diff
@@ -29,14 +30,36 @@ def __getattr__(name: str) -> Any:
...
+                return bool(mocked_value)
+            elif isinstance(real_value, int):
+                return int(mocked_value)
+            elif isinstance(real_value, float):
+                return float(mocked_value)
+            elif isinstance(real_value, pendulum.DateTime):
+                return pendulum.instance(dateparser.parse(mocked_value))
+            else:
+                # default str
+                return mocked_value
```
serinamarie, line 55: Instead of every type being an if statement, we could create a type casting dictionary to make it more readable:

```python
type_cast = {
            bool: bool,
            int: int,
            ...
}
```

and then we can cast it in the end

```python
cast_func = type_cast.get(type(real_value), str)
        return cast_func(mocked_value)
```

tests/runtime/test_flow_run.py
```diff
@@ -28,6 +29,15 @@ async def test_attribute_override_via_env_var(self, monkeypatch):
...
 
+    async def test_scheduled_start_time_override_via_env_var(self, monkeypatch):
+        # this env var must be converted to pendulum.DateTime
+        dt = "2023-05-13 20:00:00"
+        monkeypatch.setenv(
+            name="PREFECT__RUNTIME__FLOW_RUN__SCHEDULED_START_TIME", value=dt
+        )
+        assert isinstance(flow_run.scheduled_start_time, pendulum.DateTime)
+        assert flow_run.scheduled_start_time == pendulum.instance(dateparser.parse(dt))
+
```
serinamarie, line 40: It would be ideal to have tests in:
`test_task_run.py` and `test_deployment.py` as well!



It makes sense to test all different types, perhaps using pytest's parametrize like so:

https://github.com/PrefectHQ/prefect/blob/d709b49c587053423eb61d05ad0c261d0adbadc7/benches/bench_flows.py#L30-L33

src/prefect/runtime/flow_run.py
```diff
@@ -29,14 +30,36 @@ def __getattr__(name: str) -> Any:
...
+        else:
+            return real_value
     else:
-        return func()
+        if env_key in os.environ:
+            return os.environ[env_key]
+        else:
+            raise AttributeError(f"{__name__} has no attribute {name!r}")
```
serinamarie, line 62: It might be more readable to check `if func is None` first-- that way we can quickly `return` or `raise`. With that change, we can then make `if env_key in os.environ` (where `func is not None`) a top-level/outermost condition.

serinamarie (commented):
Hi @jcozar87, we're excited for your contribution!

I left comments in the `flow_run.py` file, but the comments equally apply to `deployment.py` and `task_run.py`. Please let me know if you have any questions! 🙂

src/prefect/runtime/flow_run.py
jcozar87: That's a great idea! I implemented it in that way, but without the default str. The reason is that an attribute is of type complex (i.e. a List, or Dict) then it is difficult to implement the casting from a string (we should define a language to do that, think about a dict of parameters, each value of a different type). With that in mind, I think it is better for the end user to write in the docs that complex types (lists and dicts for now) cannot be mocked using environment variables, and suggesting to do that using monkeypatch or a similar tool.

tests/runtime/test_flow_run.py
jcozar87: That's also a good idea! I implemented the tests for all the known attributes, including those whose cannot be mocked (and checking that it raises a ValueError).

I left two attributes from be tested in `test_deployment.py`, because they are Optional[Dict] types, and by default they are None during the test executions.

src/prefect/runtime/flow_run.py
jcozar87: Totally agree!

jcozar87:
The documentation should be updated. In the file `docs/concepts/runtime-context.md` the text would be as follows (tell me if you think it is right so I can do the commit with the doc change). Probably we should mention the patch version where this functionality is released.

```
!!! tip "Mock values via environment variable"
    Oftentimes, you may want to mock certain values for testing purposes.  For example, by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3`, you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
    <div class="terminal">
    ```bash
    $ export PREFECT__RUNTIME__TASK_RUN__FAKE_KEY='foo'
    $ python -c 'from prefect.runtime import task_run; print(task_run.fake_key)' # "foo"
    ```
    </div>
    If the environment variable mock an existing runtime attribute, the value is casted to the same type. It only works for runtime attributes of basic types, for those of type `list` or `dict` they can be mocked using [monkeypatch](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) or a similar tool.
```

tests/runtime/test_deployment.py
```diff
@@ -29,9 +29,45 @@ async def test_known_attributes_autocomplete(self):
...
+        # check the type of the flow_run attribute
+        assert isinstance(deployment_attr, type(expected_value))
+        # check the flow_run attribute value is expected_value
+        assert deployment_attr == expected_value
+
+    @pytest.mark.parametrize(
+        "env_name",
+        [
+            # complex types (list and dict) not allowed to be mocked using environment variables
+            "parameters",
```
serinamarie, line 60: This tests only `dict`, correct? If so, we would ideally have one for `list`.

src/prefect/runtime/flow_run.py
serinamarie: Handling the built-in types for now makes sense 👍

serinamarie:
> The documentation should be updated. In the file `docs/concepts/runtime-context.md` the text would be as follows (tell me if you think it is right so I can do the commit with the doc change). Probably we should mention the patch version where this functionality is released.
> 
> ```
> !!! tip "Mock values via environment variable"
>     Oftentimes, you may want to mock certain values for testing purposes.  For example, by manually setting an ID or a scheduled start time to ensure your code is functioning properly.  Starting in version `2.10.3`, you can mock values in runtime via environment variable using the schema `PREFECT__RUNTIME__{SUBMODULE}__{KEY_NAME}=value`:
>     <div class="terminal">
>     ```bash
>     $ export PREFECT__RUNTIME__TASK_RUN__FAKE_KEY='foo'
>     $ python -c 'from prefect.runtime import task_run; print(task_run.fake_key)' # "foo"
>     ```
>     </div>
>     If the environment variable mock an existing runtime attribute, the value is casted to the same type. It only works for runtime attributes of basic types, for those of type `list` or `dict` they can be mocked using [monkeypatch](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) or a similar tool.
> ```

That would be great if you could open a PR for it 🙂

serinamarie:
> 

Yep, a new commit makes sense!

tests/runtime/test_deployment.py
```diff
@@ -29,9 +29,45 @@ async def test_known_attributes_autocomplete(self):
         assert "id" in dir(deployment)
         assert "foo" not in dir(deployment)
 
-    async def test_attribute_override_via_env_var(self, monkeypatch):
-        monkeypatch.setenv(name="PREFECT__RUNTIME__DEPLOYMENT__NEW_KEY", value="foobar")
-        assert deployment.new_key == "foobar"
+    @pytest.mark.parametrize(
+        "env_name, env_value, expected_value",
```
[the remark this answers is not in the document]
jcozar87: That's makes more sense

tests/runtime/test_deployment.py
jcozar87: Yeah, the problem is that it only can be tested with existing attributes. In the case of deployments attributes are:
- id (Optional[str])
- flow_run_id (Optional[str])
- parameters (Dict)
- name (Optional[Dict])
- version (Optional[Dict])

As any of the attributes are of type List, it can be tested. It can be mocked if you think it would improve the readability, but the code that raises the exception is already tested when the type is Dict.

serinamarie: 👍

docs/concepts/runtime-context.md
```diff
@@ -22,6 +22,7 @@ The run context itself contains many internal objects used by Prefect to manage
     $ python -c 'from prefect.runtime import task_run; print(task_run.fake_key)' # "foo"
     ```
     </div>
+    If the environment variable mock an existing runtime attribute, the value is casted to the same type. It only works for runtime attributes of basic types (`bool`, `int`, `float` and `str`) and `pendulum.DateTime`, for those of type `list` or `dict` they can be mocked using [monkeypatch](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) or a similar tool.
```
serinamarie, line 25:
```suggestion
    If the environment variable mocks an existing runtime attribute, the value is cast to the same type. This works for runtime attributes of basic types (`bool`, `int`, `float` and `str`) and `pendulum.DateTime`. For complex types like `list` or `dict`, we suggest mocking them using [monkeypatch](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) or a similar tool.
```

tests/runtime/test_task_run.py
```diff
@@ -19,9 +19,49 @@ async def test_known_attributes_autocomplete(self):
         assert "id" in dir(task_run)
         assert "foo" not in dir(task_run)
 
-    async def test_attribute_override_via_env_var(self, monkeypatch):
-        monkeypatch.setenv(name="PREFECT__RUNTIME__TASK_RUN__NEW_KEY", value="foobar")
-        assert task_run.new_key == "foobar"
+    @pytest.mark.parametrize(
+        "attribute_name, attribute_value, expected_value",
+        [
+            # new user defined attribute
```
serinamarie, line 25:
```suggestion
            # new user-defined attribute
```

tests/runtime/test_deployment.py
serinamarie: @jcozar87 Looking through the tests a bit more, yes I do think we should mock types that don't exist, e.g.
```python 
           ...  # other parameters
           ("pretend_number", "123", 123),
           ("pretend_bool", "True", True),
        ],
    )
    async def test_attribute_override_via_env_var(
        self, monkeypatch, attribute_name, attribute_value, expected_value
    ):
        flow_run.FIELDS["pretend_number"] = int
        flow_run.FIELDS["pretend_bool"] = bool
```

tests/runtime/test_task_run.py
```diff
@@ -19,9 +19,49 @@ async def test_known_attributes_autocomplete(self):
...
+            ("new_key", "foobar", "foobar"),
+            # KNOWN FIELDS
+            # id is of type str
+            ("id", "fake-id", "fake-id"),
+            # name is of type str
+            ("name", "fake-name", "fake-name"),
+            # flow_name is of type str
```
serinamarie, line 32:
```suggestion
            # task_name is of type str
```

tests/runtime/test_task_run.py
```diff
@@ -19,9 +19,49 @@ async def test_known_attributes_autocomplete(self):
...
+    )
+    async def test_attribute_override_via_env_var(
+        self, monkeypatch, attribute_name, attribute_value, expected_value
+    ):
+        monkeypatch.setenv(
+            name=f"PREFECT__RUNTIME__TASK_RUN__{attribute_name.upper()}",
+            value=attribute_value,
+        )
+        tasks_run_attr = getattr(task_run, attribute_name)
+        # check the type of the flow_run attribute
```
serinamarie, line 44:
```suggestion
        # check the type of the task_run attribute
```

tests/runtime/test_task_run.py
```diff
@@ -19,9 +19,49 @@ async def test_known_attributes_autocomplete(self):
...
+        assert isinstance(tasks_run_attr, type(expected_value))
+        # check the flow_run attribute value is expected_value
```
serinamarie, line 46:
```suggestion
        # check the task_run attribute value is expected_value
```

tests/runtime/test_flow_run.py
```diff
@@ -24,9 +24,55 @@ async def test_known_attributes_autocomplete(self):
         assert "id" in dir(flow_run)
         assert "foo" not in dir(flow_run)
 
-    async def test_attribute_override_via_env_var(self, monkeypatch):
-        monkeypatch.setenv(name="PREFECT__RUNTIME__FLOW_RUN__NEW_KEY", value="foobar")
-        assert flow_run.new_key == "foobar"
+    @pytest.mark.parametrize(
+        "attribute_name, attribute_value, expected_value",
+        [
+            # new user defined attribute
```
serinamarie, line 30:
```suggestion
            # new user-defined attribute
```

tests/runtime/test_deployment.py
```diff
@@ -29,9 +29,46 @@ async def test_known_attributes_autocomplete(self):
         assert "id" in dir(deployment)
         assert "foo" not in dir(deployment)
 
-    async def test_attribute_override_via_env_var(self, monkeypatch):
-        monkeypatch.setenv(name="PREFECT__RUNTIME__DEPLOYMENT__NEW_KEY", value="foobar")
-        assert deployment.new_key == "foobar"
+    @pytest.mark.parametrize(
+        "attribute_name, attribute_value, expected_value",
+        [
+            # new user defined attribute
```
serinamarie, line 35:
```suggestion
            # new user-defined attribute
```

tests/runtime/test_deployment.py
```diff
@@ -29,9 +29,46 @@ async def test_known_attributes_autocomplete(self):
...
+    )
+    async def test_attribute_override_via_env_var(
+        self, monkeypatch, attribute_name, attribute_value, expected_value
+    ):
+        monkeypatch.setenv(
+            name=f"PREFECT__RUNTIME__DEPLOYMENT__{attribute_name.upper()}",
+            value=attribute_value,
+        )
+        deployment_attr = getattr(deployment, attribute_name)
+        # check the type of the flow_run attribute
```
serinamarie, line 52:
```suggestion
        # check the type of the deployment attribute
```

tests/runtime/test_deployment.py
```diff
@@ -29,9 +29,46 @@ async def test_known_attributes_autocomplete(self):
...
+        assert isinstance(deployment_attr, type(expected_value))
+        # check the flow_run attribute value is expected_value
```
serinamarie, line 54:
```suggestion
        # check the deployment attribute value is expected_value
```

tests/runtime/test_deployment.py
serinamarie: Hi @jcozar87, I am confused by the addition of the new test. Using the code above, it won't raise and you can ensure that a user is able to cast to a bool, int, float, etc., which would be ideal to ensure that works in case new runtime attribute types are added in future. I don't believe any new tests need to be added, just new parameters for each of the 3 test files. After that, we should be good to go!

serinamarie: Hi @jcozar87, if you want to add a new test type to each file that is fine with me. As long as we test that all of the basic types can be correctly cast (bool, int, float, etc.) so that we can be sure we can provide that functionality to the user 🙂 Whether you want to add them in as a new test or new parameters is fine either way with me!

serinamarie (approved):

serinamarie:
Thanks so much for your contribution to Prefect, @jcozar87! 🎉

PR #9567, opened 2023-05-15
Configurable timeout for client startup and shutdown

asmundo:
I apologize for a second posting of the same PR. The first PR was by a mistake automatically closed by me resetting the branch before adding the new change on top. Still new to GitHub workflow! 

In our CI system we sometimes have very high load. In these cases the client may sometimes hit the timeout of 30s without it being a actual error. With this change it is now possible to configure the timeouts of starting up and terminating the client. In our investigation and development of this change we have been able to consistently reproduce the issue. By setting this value higher we have found that that our tests passes 100% of the time.

Email from CI stats that it is not compatible with:

Check compatibility with Prefect 2.6
Process completed with exit code 1.
Check compatibility with Prefect 2.8
Process completed with exit code 1.
Check compatibility with Prefect 2.7
Process completed with exit code 1.
Check compatibility with Prefect 2.10
Process completed with exit code 1.
Check compatibility with Prefect 2.9
Process completed with exit code 1.

The stack traces do not look related to my change. If they are I apologize and I need some hints on how to fix them.  
### Checklist


asmundo:
I did spend quite a lot of time to be able to reproduce the issue. At the end I had to start 3 process that each copied a 1G file. While I a small prefect job was executed.  Without load the test took about 4-10s. With IO load it took about 2min and 30s. 

I did first try to trigger this with CPU load but did never manage this. It was IO load that triggered the issue.

asmundo:
Hi @madkinsz,

You mention that you want to make a more optimized solution that will not suffer from the issue we hit. Is there an issue or pull request I can follow?  
I can do some testing on my machine if you would like too.

PR #9575, opened 2023-05-15
Additional tweaks to task retry docs

abrookins:
Clean up the docs for task retries.

### Checklist


docs/concepts/tasks.md
```diff
@@ -204,29 +204,80 @@ def my_flow():
 
 ## Retries
 
-Prefect tasks can automatically retry on failure. To enable retries, pass `retries` and `retry_delay_seconds` parameters to your task.
+Prefect can automatically retry tasks on failure. In Prefect, a task _fails_ 
+if it returns a value that is not a `State` object, or if it raises an
```
billpalombi, line 208:
```suggestion
if its python function raises an exception or, uncommonly, returns a value that is not a `State` object.
```

docs/concepts/tasks.md
```diff
@@ -204,29 +204,80 @@ def my_flow():
...
+exception.
```
billpalombi, line 209:
```suggestion
```

docs/concepts/tasks.md
```diff
@@ -204,29 +204,80 @@ def my_flow():
...
+### A simple example: dividing by zero
+
+For example, this task divides a numerator by a random denominator without 
+checking to ensure that `denominator` is a non-zero number:
+
+```python
+from random import randint
+from prefect import task
+
+@task(retries=3)
```
billpalombi, line 229: This doesn't work, right? A task must be in a flow.

billpalombi:
```RuntimeError: Tasks cannot be run outside of a flow. To call the underlying task function outside of a flow use task.fn()
```

billpalombi (changes requested):
Tasks must

docs/concepts/tasks.md
abrookins: @billpalombi Isn't a task a function? 😅

docs/concepts/tasks.md
abrookins: This is the definition of a task. You have to call or submit tasks from a flow.

billpalombi: Ah I see, you didn't intend for this to be a runnable example. Most (all?) of the other code examples that we include in the docs are runnable, so including a code snippet that can't be run in isolation is a departure from the norm. If we are going to include an example that isn't runnable in isolation, I think we should remove the `import` statements, so it doesn't look like its code meant to be copied, pasted, and run in isolation.

billpalombi: We also might want to explicitly call out thats its just a definition, and must be called from a flow.

abrookins: Ah, gotcha. I see. The other examples in this section aren't runnable, so I didn't even notice! I'll adapt this to have a flow that calls it.

docs/concepts/tasks.md
abrookins: Actually I see what you mean now

abrookins: I'm going to leave out the `State` part for now because it doesn't make sense and I don't understand what would cause that. (It can't be that the user-defined function returned an object other than a `State`...).

billpalombi (approved):
This looks great, thank you!

Issue #9524, opened 2023-05-11
JSON serializing for `create_table_artifact` should not allow NaNs


tmthyln:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

A table saved with `create_table_artifact` that contains NaNs of any sort will be serialized via [`json.dumps`](https://github.com/PrefectHQ/prefect/blob/93c41a0a656944881115c5616a6b004a1d0d333d/src/prefect/artifacts.py#LL127C28-L127C33), but this is not properly displayed in the Prefect UI. Tables will appear empty showing "No results" although the raw data shows the full JSON. (JSON parsing fails since `JSON.parse` in the browser does not accept NaNs whereas `json.dumps` in Python will default to encoding them.)

Without knowing this, it looks like there was some issue with the table, but it's really an issue of incompatible encodings.

Possible fixes?:
- Maybe it would be nice to at least throw an error on the Python side (e.g. set `allow_nan=False` for `json.dumps`) so the user is aware that these saved table artifacts are not visible in the UI. However, I see how this could cause issues for those who don't care about viewing certain artifacts in the UI and just want to persist results.
  - Relatedly, could also automatically replace NaNs with e.g. nulls, but changing the value of the user's artifact is probably worse.
- Custom/different parser for JSON in the UI: I'm not a prefect developer, but I imagine this is more work than necessary or desired.
- No code change, but a note is put in the docs for `create_table_artifact` that tables with NaNs can be saved as artifacts, but will not display properly in the UI. That way users can do their own processing beforehand (just like we should be doing for non-JSON serializable objects in tables too?).

### Reproduction

```python3
from prefect.artifacts import create_table_artifact

sample_table = [
    {'key': 'ae165d', 'numeric': 0.85},
    {'key': '55dy7e', 'numeric': float('nan')},
]

create_table_artifact(
    key='some-table',
    table=sample_table,
    description='This table contains NaN',
)

# then see artifacts page on Prefect UI
```

### Error

```python3
No stack trace; table artifact is shown as "No results" in Prefect UI.
```

### Versions

```Text
Version:  2.10.0
API version: 0.8.4
Python version: 3.9.16
Git commit: 6115fd2c
Built: Thu, Apr 6, 2023 3:12 PM
[...]

PR #9585, opened 2023-05-15
Add ability to view table artifacts with NaN values in the Prefect UI

serinamarie:
Closes https://github.com/PrefectHQ/prefect/issues/9524

### Checklist

### DIFF
--- a/tests/test_artifacts.py
+++ b/tests/test_artifacts.py
@@ -416,7 +416,7 @@ class TestCreateArtifacts:
             description="my-artifact-description",
         )
 
-    async def test_create_dict_table_artifact_with_nan_raises(self, client):
+    async def test_create_dict_table_artifact_with_nan_succeeds(self, client):
         my_table = {"a": [1, 3], "b": [2, float("nan")]}
 
         artifact_id = await create_table_artifact(


src/prefect/artifacts.py
```diff
@@ -124,7 +124,28 @@ async def create_table_artifact(
...
+            if isinstance(v, list):
+                if any(isinstance(x, float) and x != x for x in v):
+                    raise ValueError(
+                        "`create_table_artifact` does not support NaN values in `table`"
+                        " argument."
+                    )
+        formatted_table = table
+    elif isinstance(table, list):
+        try:
+            formatted_table = json.dumps(table, allow_nan=False)
```
madkinsz, line 138: Why can't we dump the table in the `dict` case? Don't we need it to be JSON eventually too?

src/prefect/artifacts.py
```diff
@@ -124,7 +124,28 @@
     Returns:
         - The table artifact ID.
     """
-    formatted_table = json.dumps(table) if isinstance(table, list) else table
+    if isinstance(table, dict):
+        for _, v in table.items():
+            if isinstance(v, list):
+                if any(isinstance(x, float) and x != x for x in v):
```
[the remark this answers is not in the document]
madkinsz: This seems more reasonable than the equality check

src/prefect/artifacts.py
```diff
@@ -124,7 +124,28 @@ async def create_table_artifact(
...
+                    raise ValueError(
+                        "`create_table_artifact` does not support NaN values in `table`"
+                        " argument."
+                    )
+        formatted_table = table
+    elif isinstance(table, list):
+        try:
+            formatted_table = json.dumps(table, allow_nan=False)
+        except ValueError:
```
madkinsz, line 139: Can we just cast NaN to a reasonable value instead of throwing an exception?

src/prefect/artifacts.py
```diff
@@ -124,7 +124,28 @@ async def create_table_artifact(
...
-    formatted_table = json.dumps(table) if isinstance(table, list) else table
+    if isinstance(table, dict):
+        for _, v in table.items():
+            if isinstance(v, list):
+                if any(isinstance(x, float) and x != x for x in v):
+                    raise ValueError(
+                        "`create_table_artifact` does not support NaN values in `table`"
+                        " argument."
+                    )
+        formatted_table = table
```
serinamarie, line 135: I'm wondering if it would make sense to warn the user past a certain length. For instance, creating a table artifact like so: 

{"a": [1,2,...1000000], "b": [5]} 

would not be ideal.

madkinsz: Don't we limit the size of the JSON payload?

src/prefect/artifacts.py
madkinsz: The user's not giving us JSON though, they're giving us a table for display purposes in the UI. That it's invalid JSON is a detail of our implementation. We could easily make it valid JSON by replacing it with "NaN" and then it'd be displayed well in the UI.

serinamarie: I agree

serinamarie: @billpalombi do you have any thoughts on whether NaNs in table artifacts should be converted so that they can be rendered properly in the UI or whether we should raise an exception?

billpalombi: I agree with you and @madkinsz, we should convert NaN's to something we can render rather than throwing an exception.

jakekaplan (approved):
👏

Issue #9493, opened 2023-05-09
`prefect server start` fails in arm64 container images


jannekeskitalo:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Starting the prefect server in a fresh docker container fails. Last version where the following command works is 2.7.10.

```bash
docker run \
    -it \
    --rm \
    -p 4200:4200 \
    -e PREFECT_ORION_API_HOST="0.0.0.0" \
    -e PREFECT_API_URL="http://localhost:4200/api" \
    prefecthq/prefect:2.10.8-python3.10 /bin/sh -c "prefect server start"
```

Interestingly changing the command to delete and reset the sqlite database works, but it's a bit ugly.

```bash
docker run \
    -it \
    --rm \
    -p 4200:4200 \
    -e PREFECT_ORION_API_HOST="0.0.0.0" \
    -e PREFECT_API_URL="http://localhost:4200/api" \
    prefecthq/prefect:2.10.8-python3.10 /bin/sh -c "rm /root/.prefect/prefect.db && prefect server database reset -y && prefect server start"
```

### Reproduction

```python3
docker run \
    -it \
    --rm \
    -p 4200:4200 \
    -e PREFECT_ORION_API_HOST="0.0.0.0" \
    -e PREFECT_API_URL="http://localhost:4200/api" \
    prefecthq/prefect:2.10.8-python3.10 /bin/sh -c "prefect server start"
```

### Error

```python3
___ ___ ___ ___ ___ ___ _____ 
| _ \ _ \ __| __| __/ __|_   _| 
|  _/   / _|| _|| _| (__  | |  
|_| |_|_\___|_| |___\___| |_|  

Configure Prefect to communicate with the server with:

    prefect config set PREFECT_API_URL=http://0.0.0.0:4200/api

View the API reference documentation at http://0.0.0.0:4200/docs

Check out the dashboard at http://0.0.0.0:4200



Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1905, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
[...]

PR #9586, opened 2023-05-16
Update `prefect version` command to avoid creating the database if it does not exist

madkinsz:
When using the ephemeral app, database migrations are run during a `prefect version` call. There's no reason to be running migrations when retrieving version information and it has caused a corrupted database in https://github.com/PrefectHQ/prefect/issues/9493

Closes https://github.com/PrefectHQ/prefect/issues/9493

### DIFF
--- a/src/prefect/cli/root.py
+++ b/src/prefect/cli/root.py
@@ -118,8 +118,10 @@ async def version():
     server_type: str
 
     try:
-        async with prefect.get_client() as client:
-            server_type = client.server_type.value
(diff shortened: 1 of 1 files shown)


jawnsy (approved):

madkinsz:
We could add a unit test for this with a temporary database url and an ephemeral application if we wanted?

I tested with

```
rm -rf ~/.prefect
prefect version
ls ~/.prefect
```

PR #9587, opened 2023-05-16
Remove image layer caching during builds

jawnsy:
Avoid caching during image builds to ensure freshness at the cost of slower build times.

### Checklist

### DIFF
--- a/.github/workflows/docker-images.yaml
+++ b/.github/workflows/docker-images.yaml
@@ -94,8 +94,6 @@ jobs:
 
       - name: Build and push image
         uses: docker/build-push-action@v4
-        env:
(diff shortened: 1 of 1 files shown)


madkinsz:
Are we disabling caching for _all_ builds here? Should we just do so for releases and retain the cache for the dev images?

jawnsy:
@madkinsz Yea, I removed it for everything (dev + release builds). I went back and forth on keeping it for dev images, but I think having more consistency between our builds is preferable, unless it becomes a problem otherwise.

I figured that the performance cost wouldn't be too significant, and if I'm wrong, we can always add caching back for dev builds

madkinsz:
I think it _was_ significant in the original testing. We'll know in a second :)

madkinsz (approved):

jawnsy:
Related to https://github.com/PrefectHQ/prefect/issues/9493 - we seemed to have a caching issue related to the `prefect.db` database file

PR #9594, opened 2023-05-16
Improve SQLite database transaction behavior

madkinsz:


src/prefect/server/database/configurations.py
```diff
@@ -246,32 +247,51 @@ async def dispose_engine(cache_key):
...
 
         # write to a write-ahead-log instead and regularly
         # commit the changes
         # this allows multiple concurrent readers even during
         # a write transaction
-        conn.execute(sa.text("PRAGMA journal_mode = WAL;"))
+        cursor.execute("PRAGMA journal_mode = WAL;")
+
+        # enable foreign keys
+        # cursor.execute("PRAGMA foreign_keys = ON;")
```
madkinsz, line 266: This breaks database migrations

src/prefect/server/database/migrations/versions/sqlite/2022_04_23_114831_fd966d4ad99c_rename_block_to_blockbasis_and_.py
```diff
@@ -15,10 +15,7 @@
 
 
 def upgrade():
-    op.rename_table("block_spec", "block_schema")
-    op.rename_table("block", "block_document")
-
-    with op.batch_alter_table("block_document", schema=None) as batch_op:
+    with op.batch_alter_table("block", schema=None) as batch_op:
```
madkinsz, line 18: Renaming the table before doing batch alter does not work with the new transactional model. It fails complaining about a missing table.

docs/contributing/overview.md
```diff
@@ -262,26 +262,6 @@ The `--autogenerate` flag will automatically generate a migration file based on
...
-When adding a migration for SQLite, it's important to include the following `PRAGMA` statements for both upgrade and downgrade:
-
-```python
-def upgrade():
-    op.execute("PRAGMA foreign_keys=OFF") # <-- add this line
-    
-    # migration code here
-    
-    op.execute("PRAGMA foreign_keys=ON") # <-- add this line
-
```
jakekaplan, line 274: probably will need to remove from the migration template as well:

https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/database/migrations/script.py.mako

src/prefect/server/api/flow_runs.py
```diff
@@ -335,7 +335,9 @@ async def set_flow_run_state(
     now = pendulum.now()
 
     # create the state
-    async with db.session_context(begin_transaction=True) as session:
+    async with db.session_context(
+        begin_transaction=True, locking_transaction=True
+    ) as session:
```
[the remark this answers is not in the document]
madkinsz: What do you mean by two transactions?

src/prefect/server/database/migrations/env.py
```diff
@@ -126,7 +127,26 @@ def do_run_migrations(connection: AsyncEngine) -> None:
...
+
+@contextlib.contextmanager
+def disable_sqlite_foreign_keys(context):
+    """
+    Disable foreign key constraints on sqlite.
+    """
+    if dialect.name == "sqlite":
+        context.execute("COMMIT")
+        context.execute("PRAGMA foreign_keys=OFF")
+        context.execute("BEGIN")
```
madkinsz, line 142: Foreign key behavior cannot be changed inside transactions so we cannot just disable them in migrations like we were before.

madkinsz: Maybe this should be moved outside the `begin_transaction` block and we can skip the commit/begin 🤔

src/prefect/server/database/configurations.py
```diff
@@ -246,32 +266,76 @@ async def dispose_engine(cache_key):
...
-        # write to a write-ahead-log instead and regularly
-        # commit the changes
-        # this allows multiple concurrent readers even during
-        # a write transaction
-        conn.execute(sa.text("PRAGMA journal_mode = WAL;"))
+        # disable legacy alter table behavior as it will cause problems during
+        # migrations when tables are renamed as references would otherwise be retained
+        # in some locations
+        # https://www.sqlite.org/pragma.html#pragma_legacy_alter_table
+        cursor.execute("PRAGMA legacy_alter_table=OFF")
```
madkinsz, line 294: It's kind of wild, but this was ON by default and causing errors during migrations. I'm not entirely sure why it was not a problem before.

src/prefect/server/api/flow_runs.py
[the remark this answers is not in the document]
jakekaplan: Wow that is word salad, sorry.

To make sure all state transitions are locked, I was wondering if it made sense to nest the transaction inside of `models.flow_runs.set_flow_run_state`? I know there are few other places where it's called outside of this route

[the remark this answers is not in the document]
madkinsz: Ah interesting. If nested transactions like that are supported, that seems reasonable.

src/prefect/server/database/configurations.py
```diff
@@ -246,32 +268,78 @@ async def dispose_engine(cache_key):
...
+        # emit our own BEGIN
+        # requires `begin_sqlite_conn`
+        # see https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#serializable-isolation-savepoints-transactional-ddl
+        mode = SQLITE_BEGIN_MODE.get()
+        if mode is not None:
+            conn.exec_driver_sql(f"BEGIN {mode}")
+
+        # Note this is intentionally a no-op if there is no BEGIN MODE set
+        # This allows us to use SQLite's default behavior for reads which do not need
+        # to be wrapped in a long-running transaction
```
madkinsz, line 330: This is a little dangerous because if someone begins a transaction on a session manually we will now ignore it. This matches the previous behavior of pysqlite so it shouldn't necessarily be a regression but I'm worried about the possibility of future footguns.

tests/server/models/test_flow_run_states.py
```diff
@@ -49,14 +47,8 @@ async def test_locks_row_for_update(self, db, flow_run):
             - If not locked correctly, session 2 will see SCHEDULED -> PENDING
 
         """
-        connection_url = PREFECT_API_DATABASE_CONNECTION_URL.value()
-        dialect = get_dialect(connection_url)
-
-        if dialect.name == "sqlite":
-            pytest.skip("SQLite does not support row-level locking")
```
madkinsz, line 56: When confirming this actually failed on SQLite before my changes it turns out that the way the events were set up ensured this test would pass even without the locking. I've updated the test to cover the actual regression.

tests/server/models/test_flow_run_states.py
```diff
@@ -65,9 +57,15 @@ class TestRule1(BaseOrchestrationRule):
             async def before_transition(
                 self, initial_state, proposed_state, context: OrchestrationContext
             ):
+                orchestrating_1.set()
+
+                # If we implement this correctly, this will actually deadlock as 2
+                # will not be set until after 1 exits its transaction so we will not
+                # wait forever. Some wait here is necessary for a reproduction.
+                with anyio.move_on_after(1):
+                    await orchestrating_2.wait()
```
madkinsz, line 66: I'd love if someone has a better suggestion for how to do this.

src/prefect/server/api/flow_runs.py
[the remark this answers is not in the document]
madkinsz: There's not a good way for us to open a nested transaction in the current paradigm (the transaction is created when you get a session)

madkinsz:
We are likely to see an increased incidence of BUSY database errors as users encounter bugs that were previously allowed. We should probably capture those and send a 503 to the client for automatic retry to reduce brittleness.

jakekaplan (approved):
🔥 🔥 🔥

PR #9621, opened 2023-05-17
Add `Block.get_block_class_from_key` and replace external uses of `lookup_type`

madkinsz:
Follow-up to #9571 ensuring that some additional cases where we rely on dispatch go through the code path that loads collections


desertaxle (approved):
LGTM!

PR #9632, opened 2023-05-18
Allow client retries when server SQLite database is busy

madkinsz:
Requires https://github.com/PrefectHQ/prefect/pull/9637
Required for https://github.com/PrefectHQ/prefect/pull/9594

Previously, if a BUSY was encountered during orchestration the rules context would eat the exception so it would never be raised. A client would receive an ABORT signal. Now, we send a 503 instead.


jakekaplan (approved):
🙌

PR #9633, opened 2023-05-18
Allow client retries when database errors are encountered

madkinsz:
Extends https://github.com/PrefectHQ/prefect/pull/9632 to include all of the retryable database exceptions that we use in Cloud.

This improves resilience for all open source server users.

### DIFF
--- a/tests/server/api/test_server.py
+++ b/tests/server/api/test_server.py
@@ -87,10 +87,8 @@ async def test_sqlite_database_locked_handler(errorname, ephemeral):
         response = await client.get("/api/raise_busy_error")
         assert response.status_code == 503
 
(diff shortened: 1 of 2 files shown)


tests/server/api/test_server.py
```diff
@@ -74,7 +76,7 @@ async def raise_other_error():
             Exception,
         )
 
-    app = create_app(ephemeral=ephemeral)
+    app = create_app(ephemeral=ephemeral, ignore_cache=True)
```
madkinsz, line 79: Uh yeah if you don't do this the test routes get added to the cached app and other tests see it :D

jakekaplan (approved):
🌶️ 🌶️ 🌶️

PR #9637, opened 2023-05-18
Fix bug where ephemeral server raised exceptions client-side

madkinsz:
While working on #9632 we noticed that the database error was being raised client-side.

@jakekaplan and I discovered that FastAPI treats [exception handlers that capture `Exception` as `ServerErrorMiddleware` instead of `ExceptionMiddleware`](https://github.com/tiangolo/fastapi/blob/8cc967a7605d3883bd04ceb5d25cc94ae079612f/fastapi/applications.py#L163-L164). The key difference here is that [`ServerErrorMiddleware` will always raise an encountered exception after returning the response](https://github.com/encode/starlette/blob/d3a11205ed35f8e5a58a711db0ff59c86fa7bb31/starlette/middleware/errors.py#L184). When using an ephemeral server, this causes server-side exceptions to be raised client-side breaking all of our response error code handling. To work around this, we can either insert our exception handler directly as middleware to bypass FastAPI's logic server-side (https://github.com/PrefectHQ/prefect/pull/9637/commits/37a1c4b817941e8cb6b2bf34465466fb2f89918c) or define a special transport client-side (https://github.com/PrefectHQ/prefect/pull/9637/commits/50c03fdca386051bc452fdd8c5a7dab18c08fef3). Here, we opt for the transport with error raising disabled client-side.

We don't change this behavior for non-ephemeral servers because (in theory) it's nice to have the exceptions bubble up in the server to be reported by external error handlers (e.g. GCP will track the internal server errors).

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -563,6 +563,7 @@ def create_app(
         ),
         name="static",
     )
+    app.api_app = api_app
     app.mount("/api", app=api_app, name="api")
     app.mount("/", app=ui_app, name="ui")
 
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -22,6 +22,7 @@ from fastapi.openapi.utils import get_openapi
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
 from starlette.exceptions import HTTPException
+from starlette.middleware.exceptions import ExceptionMiddleware
 
 import prefect
 import prefect.server.api as api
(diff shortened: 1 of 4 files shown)


src/prefect/server/api/server.py
```diff
@@ -543,6 +545,7 @@ def on_service_exit(service, task):
         ),
         name="static",
     )
+    app.api_app = api_app
```
madkinsz, line 548: Not needed anymore
```suggestion
```

jakekaplan (approved):

PR #9656, opened 2023-05-19
Add support for SQLAlchemy 2

madkinsz:
- Update block select clauses for SA2
- Remove ignore for SA2 removed warnings
- Update database migrations for SA2
- Remove upper bound for sqlalchemy requirement
- Resolve additional deprecated `select` usage
- Update migration to handle row key change
- Revert change to `from_select`
- Update run history and notifications to handle row mapping changes
- Fix bare use of `execute` without `sa.text()`

### Checklist


src/prefect/server/database/orm_models.py
```diff
@@ -1304,9 +1308,9 @@ class ORMFlowRunNotificationQueue:
 
 @declarative_mixin
 class ORMVariable:
-    name: str = sa.Column(sa.String, nullable=False)
-    value: str = sa.Column(sa.String, nullable=False)
-    tags: List[str] = sa.Column(JSON, server_default="[]", default=list, nullable=False)
+    name = sa.Column(sa.String, nullable=False)
```
jakekaplan, line 1311: Can we not suggest types here anymore?

jakekaplan (approved):

desertaxle (approved):
🚀

src/prefect/server/database/orm_models.py
madkinsz: We cannot as far as I can tell. It causes all sorts of weird problems. I removed them all in Cloud and mypy is quite upset — the SQLAlchemy types are all new in v2 though.

jakekaplan: ahhh gotcha

PR #9684, opened 2023-05-22
Emit task run state change events when orchestrating a task run

### LINKED ISSUE #9651: Emit events when the state of a task run changes.
### First check

### Prefect Version

2.x

No events are currently emitted when a task run changes states.

Whenever a task run changes states an event should be emitted.

### Example Use

_No response_

### Additional context

_No response_

bunchesofdonald:
This updates the engine to emit an event when a task run changes states. 

Closes #9651 

### Example
Screenshot 2023-05-22 at 4 50 01 PM
 -->

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -60,6 +60,7 @@ from prefect.context import (
     TaskRunContext,
 )
 from prefect.deployments import load_flow_from_flow_run
+from prefect.events import Event, emit_event
 from prefect.exceptions import (
     Abort,
     FlowPauseTimeout,
@@ -121,6 +122,7 @@ from prefect.utilities.callables import (
 )
 from prefect.utilities.collections import StopVisiting, isiterable, visit_collection
 from prefect.utilities.pydantic import PartialModel
+from prefect.utilities.text import truncated_to
 
 R = TypeVar("R")
 EngineReturnType = Literal["future", "state", "result"]
@@ -1568,6 +1570,12 @@ async def orchestrate_task_run(
         else PREFECT_TASKS_REFRESH_CACHE.value()
     )
 
+    # Emit an event to capture that the task run was in the `PENDING` state.
+    last_event = _emit_task_run_state_change_event(
+        task_run=task_run, initial_state=None, validated_state=task_run.state
+    )
+    last_state = task_run.state
+
     # Transition from `PENDING` -> `RUNNING`
     state = await propose_state(
         client,
@@ -1576,6 +1584,15 @@ async def orchestrate_task_run(
         ),
         task_run_id=task_run.id,
     )
+
+    # Emit an event to capture the result of proposing a `RUNNING` state.
+    last_event = _emit_task_run_state_change_event(
+        task_run=task_run,
+        initial_state=last_state,
+        validated_state=state,
+        follows=last_event,
+    )
+    last_state = state
 
     # flag to ensure we only update the task run name once
     run_name_set = False
@@ -1668,6 +1685,13 @@ async def orchestrate_task_run(
                     terminal_state.state_details.cache_key = cache_key
 
             state = await propose_state(client, terminal_state, task_run_id=task_run.id)
+            last_event = _emit_task_run_state_change_event(
+                task_run=task_run,
+                initial_state=last_state,
+                validated_state=state,
+                follows=last_event,
+            )
+            last_state = state
 
             await _run_task_hooks(
                 task=task,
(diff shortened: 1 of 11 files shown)


tests/events/test_events_related_from_context.py
```diff
@@ -186,26 +186,6 @@ async def test_flow():
     spy_client.client.read_flow.assert_called_once()
 
 
-async def test_caches_from_task_run_context(spy_client):
```
bunchesofdonald, line 189: This test started failing because the engine is emitting events which is warming the cache. I elected to remove it because it's not really testing anything different than the `test_caches_related_objects` test above. The run context has no influence on the cache.

src/prefect/engine.py
```diff
@@ -2021,6 +2052,8 @@ async def set_state_and_handle_waits(set_state_func) -> OrchestrationResult:
     # Parse the response to return the new state
     if response.status == SetStateStatus.ACCEPT:
         # Update the state with the details if provided
+        state.id = response.state.id
```
bunchesofdonald, line 2055: This part feels a bit... hacky. I considered changing it so that the `response.state` is returned, but we use the `state` for linking to results and that immediately started failing. I decided this was likely the least obtrusive change, but happy to consider alternatives.

chrisguidry (approved):
I'm excited about this!  I'm not an expert on where you've got the hooks, but it looks legit to me!

src/prefect/engine.py
madkinsz: Yeah we want the client-side state so that we retain in-memory results. This lgtm.

madkinsz (approved):

madkinsz:
Looks like this is a major performance regression such that some of the benchmark tests are actually timing out


```
----------------------------------------- benchmark 'bench_flow_with_called_tasks': 6 tests ------------------------------------------
Name (time in s)                                        Mean            StdDev                Min                Max            Rounds
--------------------------------------------------------------------------------------------------------------------------------------
bench_flow_with_called_tasks[10] (0246_main-d2)       1.3821 (1.0)      0.0000 (1.0)       1.3821 (1.0)       1.3821 (1.0)           1
bench_flow_with_called_tasks[10] (NOW)                4.2169 (3.05)     0.0000 (1.0)       4.2169 (3.05)      4.2169 (3.05)          1
bench_flow_with_called_tasks[50] (0246_main-d2)       5.5301 (4.00)     0.0000 (1.0)       5.5301 (4.00)      5.5301 (4.00)          1
bench_flow_with_called_tasks[100] (0246_main-d2)     11.3085 (8.18)     0.0000 (1.0)      11.3085 (8.18)     11.3085 (8.18)          1
bench_flow_with_called_tasks[50] (NOW)               21.4015 (15.48)    0.0000 (1.0)      21.4015 (15.48)    21.4015 (15.48)         1
bench_flow_with_called_tasks[100] (NOW)              40.7442 (29.48)    0.0000 (1.0)      40.7442 (29.48)    40.7442 (29.48)         1
```

madkinsz (changes requested):
We'll need to assess if the performance changes here are acceptable. There should be no performance change when not using Cloud (as in the benchmarks) so I'm surprised there is such a significant difference. Given these results, I'd also like to see benchmarks when using Cloud as a backend. The higher latency of communicating with Cloud can have a significant impact and I'm not sure how much of that is mitigated by using websockets.

bunchesofdonald:
Same, I'm surprised that there's any performance degradation, _especially_ in the non-cloud scenario given that the events client it should be using is a no-op and being run within a thread. I'll see what I can do to track down the issue.

madkinsz:
@bunchesofdonald the queue service still has to process all of the emitted events. I feel like perhaps we got that bit of the design wrong and if there is not a Cloud client we should avoid putting the event on the queue at all.

bunchesofdonald:
This is interesting (to me at least), running these benchmarks locally against main and this branch (on a mac), I'm not seeing the same performance degradation: ```---------------------------------------- benchmark 'bench_task_submit': 4 tests ---------------------------------------
Name (time in ms)                        Mean            StdDev               Min                 Max            Rounds
-----------------------------------------------------------------------------------------------------------------------
bench_task_submit[100] (NOW)           1.0450 (1.0)      0.8096 (1.0)      0.3101 (1.15)       7.3885 (1.0)         100
bench_task_submit[100] (0005_main)     1.0641 (1.02)     0.8614 (1.06)     0.2803 (1.04)       7.5927 (1.03)        100
bench_task_submit[250] (0005_main)     1.7055 (1.63)     7.9131 (9.77)     0.2694 (1.0)      125.6337 (17.00)       250
bench_task_submit[250] (NOW)           1.7233 (1.65)     7.8017 (9.64)     0.2932 (1.09)     123.6695 (16.74)       250
-----------------------------------------------------------------------------------------------------------------------```

bunchesofdonald:
@madkinsz I've tracked the issue down to the calls we make to [get the related objects from the run context](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/events/worker.py#L38-L39), which is completely unnecessary when we're not actually emitting the event anyway. I've gone ahead and updated `emit_event` with your suggestion of not queueing the event if we're not using an operational client which has fixed these base benchmarks. 

cc: @chrisguidry

madkinsz:
👍 thanks for checking!

For future reference, you can pass pytest flags so `python benches --timeout=600` should get you past the default timeout.

madkinsz (approved):

PR #9692, opened 2023-05-23
Throwing exception if single callable are not passed as List type 

rkscodes:
### Checklist

- [X] This pull request references any related issue by including "closes #9686"
- [X] This pull request includes a label categorizing the change `fix`.

### DIFF
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -62,7 +62,8 @@ from prefect.utilities.collections import listrepr
 from prefect.utilities.hashing import file_hash
 from prefect.utilities.importtools import import_object
 
-T = TypeVar("T")  # Generic type var for capturing the inner return type of async funcs
+# Generic type var for capturing the inner return type of async funcs
+T = TypeVar("T")
 R = TypeVar("R")  # The return type of the user's function
 P = ParamSpec("P")  # The parameters of the flow
 
@@ -158,6 +159,15 @@ class Flow(Generic[P, R]):
             List[Callable[[schemas.core.Flow, FlowRun, State], None]]
         ] = None,
     ):
+        # Validate states is passed as list
+        states = [on_completion, on_failure, on_cancellation, on_crashed]
+        for state in states:
+            if state is not None and not isinstance(state, list):
+                raise TypeError(
+                    f"Expected list of callable for {state}; got"
+                    f" {type(state).__name__} instead."
+                )
+
         if not callable(fn):
             raise TypeError("'fn' must be callable")
 
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -51,7 +51,8 @@ if TYPE_CHECKING:
     from prefect.context import TaskRunContext
 
 
-T = TypeVar("T")  # Generic type var for capturing the inner return type of async funcs
+# Generic type var for capturing the inner return type of async funcs
+T = TypeVar("T")
 R = TypeVar("R")  # The return type of the user's function
 P = ParamSpec("P")  # The parameters of the task
 
@@ -205,8 +206,14 @@ class Task(Generic[P, R]):
         on_completion: Optional[List[Callable[["Task", TaskRun, State], None]]] = None,
         on_failure: Optional[List[Callable[["Task", TaskRun, State], None]]] = None,
     ):
-        if not callable(fn):
-            raise TypeError("'fn' must be callable")
+        # Validate states is passed as list
+        states = [on_completion, on_failure]
+        for state in states:
+            if state is not None and not isinstance(state, list):
+                raise TypeError(
+                    f"Expected list of callable for {state}; got"
+                    f" {type(state).__name__} instead."
+                )
 
         self.description = description or inspect.getdoc(fn)
         update_wrapper(self, fn)
--- a/src/prefect/tasks.py
+++ b/src/prefect/tasks.py
@@ -215,6 +215,9 @@ class Task(Generic[P, R]):
                     f" {type(state).__name__} instead."
                 )
 
+        if not callable(fn):
+            raise TypeError("'fn' must be callable")
+
         self.description = description or inspect.getdoc(fn)
         update_wrapper(self, fn)
         self.fn = fn


src/prefect/flows.py
```diff
@@ -158,6 +159,15 @@ def __init__(
...
         ] = None,
     ):
+        # Validate states is passed as list
+        states = [on_completion, on_failure, on_cancellation, on_crashed]
+        for state in states:
+            if state is not None and not isinstance(state, list):
+                raise TypeError(
+                    f"Expected list of callable for {state}; got"
+                    f" {type(state).__name__} instead."
+                )
```
madkinsz, line 169: I'd call these `hook_types`/`hooks` instead of `states`/`state`. 

We probably want to support other iterables like tuples and sets here. Perhaps we can try to cast it to a list? e.g. `on_completion = list(on_completion)` with a try/except that captures the `TypeError` that is raised if a non-iterable type is given — we'd then raise it again with a nicer error?

This message you wrote also made me notice we could be checking if the items in the list are callables.

src/prefect/flows.py
```diff
@@ -62,7 +62,8 @@
 from prefect.utilities.hashing import file_hash
 from prefect.utilities.importtools import import_object
 
-T = TypeVar("T")  # Generic type var for capturing the inner return type of async funcs
+# Generic type var for capturing the inner return type of async funcs
+T = TypeVar("T")
```
madkinsz, line 66: Ideally we'd retain this to match the other type variables here

src/prefect/flows.py
rkscodes: Yeah sure, makes sense.

src/prefect/flows.py
rkscodes: I am sorry, I don't quite get this.

madkinsz: Ah. You changed the position of this comment so it doesn't match the comments below it for the other type variables. This change isn't necessary for this pull request and you should undo it.

madkinsz (commented):
Nice! This looks good to me. We'll want to add tests next in `test_flows.py` and `test_tasks.py` that use `with pytest.raises(...)` to ensure that the correct errors are raised.

src/prefect/flows.py
```diff
@@ -158,6 +159,25 @@ def __init__(
             List[Callable[[schemas.core.Flow, FlowRun, State], None]]
         ] = None,
     ):
+        # Validate if hooks type is list and hook_type is callable
+        hooks = [on_completion, on_failure, on_cancellation, on_crashed]
+        for hook_types in hooks:
```
madkinsz, line 164: Should probably swap use of `hooks` and `hook_types` here for clarity.

src/prefect/flows.py
```diff
@@ -158,6 +159,25 @@ def __init__(
...
+            if hook_types is not None:
+                try:
+                    hook_types = list(hook_types)
+                except TypeError:
+                    raise TypeError(
+                        f"Expected list of callable for {hook_types}; got"
```
madkinsz, line 170: This could probably say "Expected iterable for..." since we're covering callables in the next message

madkinsz: We'll want to make sure this also points the user to the correct parameter i.e. "Expected iterable for 'on_cancellation'; got..." — right now it looks like it will display the value they passed.

madkinsz:
It looks like you linked the wrong issue in the PR body as well.

src/prefect/flows.py
rkscodes: Hmm, I have no idea how it got changed, I think somehow black did it, anyways I am reverting it back.

src/prefect/flows.py
rkscodes: Sure thing, the names are now confusing me too.

rkscodes:
Updated with suggested changes, have a look.

src/prefect/flows.py
```diff
@@ -158,6 +159,26 @@ def __init__(
             List[Callable[[schemas.core.Flow, FlowRun, State], None]]
         ] = None,
     ):
+        # Validate if hook passed is list and contains callables
+        hook_types = [on_completion, on_failure, on_cancellation, on_crashed]
```
serinamarie, line 163: I hate to be a stickler *but* since we're doing duck typing below it seems a tad confusing to have `_types` in this variable (to me). I might suggest `hook_categories` or even `hook_kinds`🤷‍♀️

src/prefect/flows.py
```diff
@@ -158,6 +159,26 @@ def __init__(
...
+                except TypeError:
+                    raise TypeError(
+                        f"Expected iterable for '{hook_name}'; got"
+                        f" {type(hook).__name__} instead."
+                    )
+
+                for hook_item in hook:
+                    if not callable(hook_item):
+                        raise TypeError(
+                            f"Expected callables in {hook}; got"
```
serinamarie, line 178: I believe it would be more informative to include `hook_name` in this error message, to identify which hook category contains the non-callable item.

src/prefect/flows.py
```diff
@@ -158,6 +159,26 @@ def __init__(
...
+            if hook is not None:
+                try:
+                    hook = list(hook)
+                except TypeError:
+                    raise TypeError(
+                        f"Expected iterable for '{hook_name}'; got"
+                        f" {type(hook).__name__} instead."
+                    )
+
+                for hook_item in hook:
```
serinamarie, line 175: `for hook in hooks`: 
might be more intuitive (since right now `hook` is a list)

src/prefect/tasks.py
```diff
@@ -51,7 +51,8 @@
     from prefect.context import TaskRunContext
 
 
-T = TypeVar("T")  # Generic type var for capturing the inner return type of async funcs
+# Generic type var for capturing the inner return type of async funcs
+T = TypeVar("T")
```
serinamarie, line 55: [As Zanie mentioned](https://github.com/PrefectHQ/prefect/pull/9692#discussion_r1213192612), this might have been accidentally modified?

serinamarie (changes requested):
Hi @rkscodes, this is looking good! [As Zanie mentioned](https://github.com/PrefectHQ/prefect/pull/9692#pullrequestreview-1455539125), we'll want tests for this.

src/prefect/flows.py
rkscodes: Yeah sure, changed it to `hook_categories`

src/prefect/flows.py
rkscodes: Sure, `hook` is list, so I think I would have to rename it to `hooks` and then I can make it `for hook in hooks`

src/prefect/tasks.py
rkscodes: Yup, reverting this too

src/prefect/flows.py
rkscodes: Noted

rkscodes:
Updated with all suggested changes. I am really new to tests never written one should I attempt it ?

serinamarie:
> Updated with all suggested changes. I am really new to tests never written one should I attempt it ?

Yeah, go ahead with what Zanie suggested regarding tests!

rkscodes:
Added the test-cases, have a look!

serinamarie (commented):
These first few tests are looking good! A few things:

1. Renaming these `test()` functions to `<hook_category>_hook` would be more comprehendible, e.g. `def completion_hook():`.

2. Adding tests for an empty iterable would make sense, e.g. `@flow(on_completion=[])`.

3. Adding tests for mixed iterables would also be ideal, e.g. `@flow(on_completion=[failure_hook, "non-callable"])`.

Let me know if you've got any questions! 🚀

rkscodes:
I am bit unsure about point 1. So this fun  `def test_on_completion_hooks_is_iterable()` should be renamed to  
`def completion_hook_is_iterable()` ?

rkscodes:
Yeah I am not able to get point 1. Could you give me an example let's say for `def test_on_completion_hooks_is_iterable()` what should I rename it to 

My guess : `def completion_hook_is_iterable()` 

Or are you suggesting to merge `... is_iterable` and `... is_callable` functions into single fun named `def completion_hook()`

rkscodes:
Weird my last commit is not showing in this PR

madkinsz:
@rkscodes I think GitHub is having a service disruption

rkscodes:
Last commit is now visible, please have a look and let me know changes required if any.

tests/test_flows.py
```diff
@@ -2364,6 +2403,44 @@ def my_flow():
...
+
+            @flow(on_failure=failure_hook)
+            def flow1():
+                pass
+
+        @flow(on_failure=[])
+        def flow2():
+            pass
+
+        assert flow2.on_failure == []
```
serinamarie, line 2422: I would expect this to raise. Also, it's probably best to move this to a separate test.

tests/test_flows.py
```diff
@@ -2364,6 +2403,44 @@ def my_flow():
 
 
 class TestFlowHooksOnFailure:
+    def test_on_failure_hooks_is_iterable(self):
```
serinamarie, line 2406: I think it might make sense to name the tests like so:
`test_noniterable_hook_raises()`
`test_empty_hook_list_raises()`
`test_noncallable_hook_raises()`

tests/test_flows.py
```diff
@@ -2364,6 +2403,44 @@ def my_flow():
...
+            def flow1():
+                pass
+
+        with pytest.raises(
+            TypeError, match="Expected callables in 'on_failure'; got str instead."
+        ):
+
+            @flow(on_failure=[failure_hook, "test"])
+            def flow2():
+                pass
```
serinamarie, line 2442: I would move this to a separate test

tests/test_flows.py
rkscodes: I think default value for `on_failure` is `[ ]`, so isn't is allowed to have empty list, please help me understand.

rkscodes: Got it, I assumed default value to be empty list which is not true it is `None`

rkscodes: In `flow.py` and `tasks.py` we are not concerned about empty list as values, I think I need to update code there first.

serinamarie: Let's check for the empty list first:

```python
if hooks is not None:
  if not hooks:  # This is equivalent to checking if len(hooks) == 0 for lists
      raise ValueError(f"Empty list passed for '{hook_name}'")
  try:
      hooks = list(hooks)
```

rkscodes: Sure, it's done.

src/prefect/flows.py
```diff
@@ -158,6 +159,28 @@ def __init__(
...
+        # Validate if hook passed is list and contains callables
+        hook_categories = [on_completion, on_failure, on_cancellation, on_crashed]
+        hook_names = ["on_completion", "on_failure", "on_cancellation", "on_crashed"]
+        for hooks, hook_name in zip(hook_categories, hook_names):
+            if hooks is not None:
+                if not hooks:
+                    raise ValueError(f"Empty list passed for '{hook_name}'")
+                try:
+                    hooks = list(hooks)
+                except TypeError:
```
serinamarie, line 171: @rkscodes, this is looking great! The only thing I can think of is to try to avoid a dead end when raising this `TypeError.` It would be more helpful to show the user what is expected:

```python
            raise TypeError(
                f"Expected iterable for '{hook_name}'; got"
                f" {type(hooks).__name__} instead. Please provide a list of hooks to 
                f"{hook_name}. For example:"
                f"\n@flow(on_completion=[hook1, hook2])\ndef flow1():\npass"
            )
```

rkscodes: Yeah that would be more helpful for the user. I will adding the same msg looks good to me.

src/prefect/flows.py
```diff
@@ -158,6 +159,34 @@ def __init__(
...
+                    raise TypeError(
+                        f"Expected iterable for '{hook_name}'; got"
+                        f" {type(hooks).__name__} instead. Please provide a list of"
+                        f" hooks to '{hook_name}'. For"
+                        f" example:\n@flow({hook_name}=[hook1, hook2])\ndef"
+                        " flow1():\npass"
+                    )
```
serinamarie, line 178:
```suggestion
                    raise TypeError(
                        f"Expected iterable for '{hook_name}'; got"
                        f" {type(hooks).__name__} instead. Please provide a list of"
                        f" hooks to '{hook_name}':\n\n"
                        f"@flow({hook_name}=[hook1, hook2])\ndef"
                        " my_flow():\n\tpass"
                    )
```
Let's add indentation here, it will look nicer when it prints out:

```bash
  File "/Users/bean/code/prefect/src/prefect/flows.py", line 172, in __init__
    raise TypeError(
TypeError: Expected iterable for 'on_completion'; got function instead. Please provide a list of hooks to 'on_completion':

@flow(on_completion=[hook1, hook2])
def my_flow():
        pass
```

rkscodes: Yeah it looks great, on it.

serinamarie:
@rito-sixt This looks good, but there are changes to 2 additional files from [4893a52](https://github.com/PrefectHQ/prefect/pull/9692/commits/4893a524f52ecd8b433982d393a748c9c056cfdb)

rito-sixt:
> @rito-sixt This looks good, but there are changes to 2 additional files from [4893a52](https://github.com/PrefectHQ/prefect/pull/9692/commits/4893a524f52ecd8b433982d393a748c9c056cfdb)

Me? 🤔

rkscodes:
Ooops no idea those got added, reverting it right now.

rkscodes:
Alright, I am looking into it serinamarie.

rkscodes:
@serinamarie  Earlier `[]` was allowed but now we have to change it, I think `None` makes sense now over there.

serinamarie (approved):
Thanks a lot for your contribution and for going through several rounds of review, @rkscodes!

rkscodes:
It was really amazing, I learned a lot thanks for guiding 🎉

Issue #8940, opened 2023-03-27
Add ping and health endpoints on the Server


davzucky:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

At the moment, the deployment of the Prefect server on Kubernetes doesn't have any liveness or readiness probe enabled. An issue has been opened here to get them added after the server supports new endpoints
https://github.com/PrefectHQ/prefect-helm/issues/132

### Describe the proposed behavior

Prefect Server should have two extras endpoints
- ```/ping```, which allows liveness to be checked
- ```/health``` which would allow the readiness state to be checked, ensuring the db was updated

 ### Example Use

At the moment, the health endpoint exists under ```/api/health``` and the ping doesn't exist, however the ```/api/hello``` could be used as a proxy

_No response_

parkedwards:
thanks for this suggestion, @davzucky.  after #9701 + the corresponding server release, we'll now have a `GET /api/ready` endpoint that will perform a db check; you can use this endpoint for any readiness probes (which we will also add to the helm chart shortly)

in terms of the liveness check, please continue to use `GET /api/hello` (or `GET /api/health`).  the `/metrics` endpoint is one we're currently discussing internally, but agree that it's something we'll want to do

davzucky:
Thank you, @parkedwards, for all the updates and work. I will update Prefect in the coming weeks

PR #9701, opened 2023-05-23
feat: add an /api/ready check to prefect server

parkedwards:
relates to https://github.com/PrefectHQ/prefect-helm/issues/132
relates to https://github.com/PrefectHQ/prefect/issues/8940

the idea is that we can continue using `GET /api/health` (or `GET /api/hello`) as a liveness check, and this new `GET /api/ready` endpoint as a readiness check, which probes to see if the database is connect-able

### Checklist

### DIFF
--- a/src/prefect/server/api/root.py
+++ b/src/prefect/server/api/root.py
@@ -1,7 +1,9 @@
 """
 Contains the `hello` route for testing and healthcheck purposes.
 """
-from fastapi import Depends, Response, status
+from fastapi import Depends, status
+from fastapi.encoders import jsonable_encoder
+from fastapi.responses import JSONResponse
 from prefect.server.utilities.server import PrefectRouter
 from prefect.server.database.dependencies import provide_database_interface
 from prefect.server.database.interface import PrefectDBInterface
@@ -18,13 +20,16 @@ async def hello():
 @router.get("/ready")
 async def perform_readiness_check(
     db: PrefectDBInterface = Depends(provide_database_interface),
-    response: Response = None,
 ):
     is_db_connectable = await db.is_db_connectable()
 
     if is_db_connectable:
-        response.status_code = status.HTTP_200_OK
-        return is_db_connectable
+        return JSONResponse(
+            status_code=status.HTTP_200_OK,
+            content=jsonable_encoder({"status": "OK"}),
+        )
 
-    response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
-    return "Database is not available"
+    return JSONResponse(
+        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+        content=jsonable_encoder({"status": "Database is not available"}),
+    )
--- a/src/prefect/server/api/root.py
+++ b/src/prefect/server/api/root.py
@@ -27,4 +27,4 @@ async def peform_readiness_check(
         return is_db_connectable
 
     response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
-    return is_db_connectable
+    return "Database is not available"
--- a/src/prefect/server/api/root.py
+++ b/src/prefect/server/api/root.py
@@ -2,7 +2,6 @@
 Contains the `hello` route for testing and healthcheck purposes.
 """
 from fastapi import Depends, status
-from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
 from prefect.server.utilities.server import PrefectRouter
 from prefect.server.database.dependencies import provide_database_interface
@@ -26,10 +25,10 @@ async def perform_readiness_check(
     if is_db_connectable:
         return JSONResponse(
             status_code=status.HTTP_200_OK,
-            content=jsonable_encoder({"status": "OK"}),
+            content={"message": "OK"},
         )
 
     return JSONResponse(
         status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-        content=jsonable_encoder({"status": "Database is not available"}),
+        content={"message": "Database is not available"},
     )
--- a/src/prefect/server/api/root.py
+++ b/src/prefect/server/api/root.py
@@ -1,8 +1,10 @@
 """
 Contains the `hello` route for testing and healthcheck purposes.
 """
-
+from fastapi import Depends, Response, status
 from prefect.server.utilities.server import PrefectRouter
+from prefect.server.database.dependencies import provide_database_interface
+from prefect.server.database.interface import PrefectDBInterface
 
 router = PrefectRouter(prefix="", tags=["Root"])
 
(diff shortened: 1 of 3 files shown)


src/prefect/server/api/root.py
```diff
@@ -11,3 +13,18 @@
 async def hello():
     """Say hello!"""
     return "👋"
+
+
+@router.get("/ready")
```
parkedwards, line 18: `GET /api/ready`

i toyed with adding this to the base router, but i think that `/api/ready` is the correct endpoint namespace bc we're really checking the API server's ability to draw data and respond to requests; we dont really care about the UI server in this situation

src/prefect/server/api/root.py
```diff
@@ -11,3 +13,18 @@
...
+    response: Response = None,
+):
+    is_db_connectable = await db.is_db_connectable()
+
+    if is_db_connectable:
+        response.status_code = status.HTTP_200_OK
+        return is_db_connectable
+
+    response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
+    return "Database is not available"
```
parkedwards, line 30: opted not to return a JSONResponse instance, since this is just information for an operator to know why a readiness check might have failed

src/prefect/server/api/root.py
```diff
@@ -11,3 +13,18 @@
 async def hello():
     """Say hello!"""
     return "👋"
+
+
+@router.get("/ready")
+async def peform_readiness_check(
```
jawnsy, line 19: typo in "perform"
```suggestion
async def perform_readiness_check(
```

src/prefect/server/api/root.py
jawnsy: A lot of other services in Kubernetes use the `z` suffix convention, maybe something we can consider (or maybe not, since we already have health) https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f

src/prefect/server/api/root.py
jawnsy: IMO, JSON responses are preferable because we can extend them without breaking API compatibility. Sadly we don't do this consistently.

src/prefect/server/api/root.py
abrookins: Yeah, I'd just stick with the existing (in our app) convention.

src/prefect/server/api/root.py
abrookins: @parkedwards Can you return a 503 with this and a JSON response like `{"reason": "Database is not available"}`?

abrookins (dismissed):
Can you return a 503 with this and a JSON response like {"reason": "Database is not available"}?

src/prefect/server/api/root.py
parkedwards: i knew that the `z` was something from google, but interesting to know more about the origin details

parkedwards:
okey doke:

```
$ curl -i localhost:4200/api/ready
HTTP/1.1 200 OK
date: Wed, 24 May 2023 16:57:09 GMT
server: uvicorn
content-length: 15
content-type: application/json

{"status":"OK"}%
```

and when the db is unavailable:

```
$ docker stop 71e
71e

$ curl -i localhost:4200/api/ready
HTTP/1.1 503 Service Unavailable
date: Wed, 24 May 2023 16:57:29 GMT
server: uvicorn
content-length: 38
content-type: application/json

{"status":"Database is not available"}%
```

zangell44 (approved):

src/prefect/server/api/root.py
```diff
@@ -11,3 +15,21 @@
...
+
+    if is_db_connectable:
+        return JSONResponse(
+            status_code=status.HTTP_200_OK,
+            content=jsonable_encoder({"status": "OK"}),
+        )
+
+    return JSONResponse(
+        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+        content=jsonable_encoder({"status": "Database is not available"}),
```
madkinsz, line 34: Do we really need the `jsonable_encoder` call? I'm pretty sure we do not use this elsewhere

Nit: Can we do "message" or "details" instead of "status"? Status feels like an enum to me.

src/prefect/server/database/interface.py
```diff
@@ -66,6 +66,18 @@ async def run_migrations_downgrade(self):
...
         await run_sync_in_worker_thread(alembic_downgrade)
 
+    async def is_db_connectable(self):
+        """
+        Returns boolean indicating if the database is connectable.
+        This method is used to determine if the server is ready to accept requests.
+        """
+        engine = await self.engine()
+        try:
+            async with engine.connect():
```
madkinsz, line 76: What kind of timeout is there on this? Is the `/ready` endpoint going to take 60s if it's not available?

parkedwards: i want to say this is dependent on the underlying engine, i couldn't find any exact literature on the a default timeout value here.  during my local testing, this timed out a 5s - but assuming this is used in k8s or similar, usually those probes have their own timeouts, after which a probe is considered to be failed

src/prefect/server/api/root.py
parkedwards: errrm maybe we don't need `jsonable_encoder`, i did pull this out from examples in this repo, but taking a look at the fastapi docs, `jsonable_encoder` is more for pydantic models that first need to be converted to a dict. in our case, the responses are hardcoded dicts so dont think we need em

>Nit: Can we do "message" or "details" instead of "status"? Status feels like an enum to me.
works for me

src/prefect/server/database/interface.py
madkinsz: Hm okay. I know we'll wait quite a while for a SQLite connection. We may want to add an `with anyio.fail_after(...)` timeout in the future but this seems okay for now :)

madkinsz (approved):

src/prefect/server/api/root.py
```diff
@@ -11,3 +14,21 @@
 async def hello():
     """Say hello!"""
     return "👋"
+
+
+@router.get("/ready")
```
madkinsz, line 19: Technically we should have unit tests for this route. I'm okay with addressing separately though.

src/prefect/server/database/interface.py
abrookins: I assumed this would use the value of `PREFECT_API_DATABASE_CONNECTION_TIMEOUT`: https://github.com/PrefectHQ/prefect/blob/91105c475c5688d520b4b4de1e548989d533f581/src/prefect/settings.py#L922

abrookins (approved):
LGTM! 👍

src/prefect/server/database/interface.py
parkedwards: heh, that's probably it

src/prefect/server/api/root.py
parkedwards: grrr. yes this does need a test

src/prefect/server/database/interface.py
madkinsz: Sounds right lol — great!

src/prefect/server/api/root.py
parkedwards: ok added - simple enough, lots of great examples in the repo already

Issue #9709, opened 2023-05-24
Incorrect `Variables.value` max len on update 


jakekaplan:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When calling `PATCH - .../api/variables/{id}` `Variables.value` can't be updated to more than 256 characters. That is the correct max length for `Variables.name`, but `Variables.value`.

### Reproduction

```python3
from prefect.client.orchestration import get_client
import asyncio


async def main():

    async with get_client() as client:
        res = await client._client.post(f"/variables/", json={"name": "foo", "value": "bar"})
        assert res.status_code == 201

        res = await client._client.patch(f"/variables/name/foo", json={"value": "b" * 256})
        assert res.status_code == 422
        print(res.json())

asyncio.run(main())
```

### Error

```python3
Traceback (most recent call last):
  File "/Users/jakekaplan/PycharmProjects/prefect/jake-test2.py", line 15, in <module>
    asyncio.run(main())
  File "/Users/jakekaplan/opt/anaconda3/envs/py311/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/jakekaplan/opt/anaconda3/envs/py311/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jakekaplan/opt/anaconda3/envs/py311/lib/python3.11/asyncio/base_events.py", line 650, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/jakekaplan/opt/anaconda3/envs/py311/lib/python3.11/contextlib.py", line 222, in __aexit__
    await self.gen.athrow(typ, value, traceback)
  File "/Users/jakekaplan/PycharmProjects/prefect/src/prefect/client/base.py", line 100, in app_lifespan_context
    yield
  File "/Users/jakekaplan/PycharmProjects/prefect/jake-test2.py", line 11, in main
    res = await client._client.patch(f"/variables/name/foo", json={"value": "b" * 257})
[...]

PR #9710, opened 2023-05-24
fix variable max value update

jakekaplan:
closes: https://github.com/PrefectHQ/prefect/issues/9709

When calling `PATCH - .../api/variables/{id}` `Variables.value` can't be updated to more than 255 characters. That is the correct max length for `Variables.name`, but `Variables.value` should be max length of 5000.

### Example
The update would return a 422 before saying the max length was 255. Now this will correctly update up to 5000 characters.
```python
import asyncio


async def main():

        res = await client._client.post(f"/variables/", json={"name": "foo", "value": "bar"})
        assert res.status_code == 201

        res = await client._client.patch(f"/variables/name/foo", json={"value": "b" * 257})
        assert res.status_code == 422
        print(res.json())

asyncio.run(main())
```

### Checklist

### DIFF
--- a/src/prefect/server/schemas/actions.py
+++ b/src/prefect/server/schemas/actions.py
@@ -644,7 +644,7 @@ class VariableUpdate(ActionBaseModel):
         default=None,
         description="The value of the variable",
         example="my-value",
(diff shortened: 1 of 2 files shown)


serinamarie (approved):
LGTM :)

PR #9715, opened 2023-05-24
Improve error message for common `TypeError` exception in state change hooks

serinamarie:
### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -2198,6 +2198,15 @@ async def _run_task_hooks(task: Task, task_run: TaskRun, state: State) -> None:
                     await run_sync_in_worker_thread(
                         hook, task=task, task_run=task_run, state=state
                     )
+            except TypeError:
+                logger.error(
+                    (
+                        "An error was encountered while running hook"
+                        f" {hook.__name__!r}. Ensure you are passing a task run state"
+                        " change hook rather than a flow run state change hook."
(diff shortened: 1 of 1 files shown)


src/prefect/engine.py
```diff
@@ -2198,6 +2198,15 @@ async def _run_task_hooks(task: Task, task_run: TaskRun, state: State) -> None:
                     await run_sync_in_worker_thread(
                         hook, task=task, task_run=task_run, state=state
                     )
+            except TypeError:
+                logger.error(
```
madkinsz, line 2202: Use `logger.exception` to automatically set `exc_info=True`

src/prefect/engine.py
```diff
@@ -2198,6 +2198,15 @@ async def _run_task_hooks(task: Task, task_run: TaskRun, state: State) -> None:
...
+                    (
+                        "An error was encountered while running hook"
+                        f" {hook.__name__!r}. Ensure you are passing a task run state"
+                        " change hook rather than a flow run state change hook."
```
madkinsz, line 2206: I think this suggestion is too specific. There are tons of reasons a type error could be thrown during execution of the hook. If we want to provide information about this we could inspect the signature and see if it matches the expected pattern? This would be better done at task / flow init time as an extension of https://github.com/PrefectHQ/prefect/pull/9692

PR #9720, opened 2023-05-24
Add Amplitude to Docs

dylanbhughes:
This PR updates our analytics on docs.prefect.io to use Amplitude instead of Google Analytics.

### Checklist


docs/overrides/partials/integrations/analytics/custom.html
```diff
@@ -0,0 +1,77 @@
+<script src="https://cdn.amplitude.com/libs/analytics-browser-1.9.1-min.js.gz"></script>
+<script type="text/javascript">
+
+const amplitudeKey = "8ab9ae57b486baa5d4ab626559fd8106"
```
dylanbhughes, line 4: If anyone knows a good way for me to switch this between stg/prd via configuration of some kind, I'd love suggestions

jawnsy: I don't know if mkdocs or material-mkdocs supports custom variables, this is probably fine for now

mkdocs.insiders.yml
```diff
@@ -1,6 +1,9 @@
 INHERIT: mkdocs.yml
 plugins:
     meta:
+    privacy:
+        assets_include:
+          - cdn.amplitude.com/*
```
jawnsy, line 6: sweet, glad you got the bundler thing working!

I think it's acceptable but fwiw if we build without insiders (such as for a local development build), then we might have a CSP issue if we don't allow the Amplitude CDN. Probably not worth fixing, just thought I'd mention it.

docs/overrides/partials/integrations/analytics/custom.html
```diff
@@ -0,0 +1,77 @@
...
+    element.addEventListener('blur', removeUrl)
+    element.addEventListener('touchstart', addUrl)
+    element.addEventListener('touchend', removeUrl)
+  })
+}
+
+
+const init = () => {
+  amplitude.init(amplitudeKey, undefined, {
+    useBatch: false,
```
jawnsy, line 46: how come we're turning off batching? won't performance be much worse?

jawnsy (commented):
Thanks for doing this, LGTM!

Marking this as a comment instead of approval as I'd like Nicholas to take a look, since I'm much less experienced with frontend stuff

sarahmk125:
Testing:
- I'm seeing `[Amplitude] Start Session` events, can these be suppressed?
- I'm only seeing a docs page view on the first page load (homepage), and then no other pages
- I think there's extra work needed to add deviceid. Will ping with some examples.

mkdocs.insiders.yml
dylanbhughes: Yeah I think that's fine for now. We'll be building w/ insiders for the production build for the foreseeable future

docs/overrides/partials/integrations/analytics/custom.html
dylanbhughes: I was turning it off for testing, we can definitely enable

madkinsz:
Why do we need this here instead of just over in `prefecthq/docs`?

dylanbhughes:
> Why do we need this here instead of just over in `prefecthq/docs`?

Zanie and I discussed in Slack. For those following, this is because we update the MKDocs config.

docs/overrides/partials/integrations/analytics/custom.html
```diff
@@ -0,0 +1,93 @@
...
+  const url = new URL(href)
+  url.searchParams.delete('deviceId')
+  event.target.href = url.toString()
+}
+
+const urls = [
+  'https://app.prefect.cloud',
+  'https://app.stg.prefect.dev',
+  'https://app.prefect.dev',
+  'http://app.prefect.cloud',
```
jawnsy, line 30: fyi, it's not possible to load any of these sites over http, we have HSTS enabled. I would suggest removing these, as it's a bug if we ever use non-HTTPS URLs.

mkdocs.yml
```diff
@@ -299,8 +299,12 @@ extra:
     version:
         provider: mike
     analytics:
-        provider: google
-        property: G-3M31G9B0QJ
+        provider: custom
+        property: !ENV [AMPLITUDE_API_KEY, "8ab9ae57b486baa5d4ab626559fd8106"]
```
znicholasbrown, line 303: Is this key something we can inject from the netlify runner instead? Weird to see this hard coded

mkdocs.yml
```diff
@@ -299,8 +299,12 @@ extra:
...
+        url: !ENV [AMPLITUDE_URL, "https://api2.amplitude.com/2/httpapi"]
+        # disabled by default for local development
+        # to enable, set AMPLITUDE_ENABLED=enabled in your environment
+        enabled: !ENV [AMPLITUDE_ENABLED, "disabled"]
```
jawnsy, line 307: I would suggest enabling this if the API Key is set, so that you don't need a separate enabled/disabled setting. It doesn't make sense to enable Amplitude without an API Key, anyways.

It does mean that for local development we shouldn't have a default API key set in our config file or envrc, but I think that should be acceptable. Most of the time we'll just be testing this in staging or in previews

mkdocs.yml
```diff
@@ -299,8 +299,12 @@ extra:
     version:
         provider: mike
     analytics:
-        provider: google
-        property: G-3M31G9B0QJ
+        provider: custom
+        property: !ENV [AMPLITUDE_API_KEY, "8ab9ae57b486baa5d4ab626559fd8106"]
+        url: !ENV [AMPLITUDE_URL, "https://api2.amplitude.com/2/httpapi"]
```
jawnsy, line 304: This should be able to use the Amplitude proxy? Maybe needs to be copied into the docs though

mkdocs.yml
znicholasbrown: Agreed with this - if we're able to set the API key in the runner we won't need to include it anywhere in source

docs/overrides/partials/integrations/analytics/custom.html
dylanbhughes: There is at least one http url currently in the docs and since we have open-source contributions I figured it would be easier to cover the edge-case here

mkdocs.yml
dylanbhughes: We are, it's purely here for testing locally but it makes sense to remove it. Will drop this now

mkdocs.yml
dylanbhughes: Set via environment variable in Netlify 👍🏻 for local testing you need the full url

znicholasbrown (approved):

Issue #9723, opened 2023-05-25
PrefectHTTPStatusError: Client error '429 Too Many Requests' for url


BitTheByte:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

While using `prefect` with `prefect-dask` I encountered a rate limit error. this shouldn't be happening as prefect client base should retry on those. I'm not sure why this is happening but this has risen at `2.10.10` and did not exist before

### Reproduction

```python3
Any Flow with prefect-dask
```

### Error

```python3
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/distributed/client.py", line 1697, in _close
    await self.scheduler_comm.close()
asyncio.exceptions.CancelledError
01:00:08.452 | ERROR   | Flow run 'psi5-alastria-x' - Crash detected! Execution was interrupted by an unexpected exception: PrefectHTTPStatusError: Client error '429 Too Many Requests' for url 'https://cloud-url/task_runs/'
Response: {'detail': 'Orchestration API rate limit reached'}
For more information check: https://httpstatuses.com/429
```

### Versions

```Text
Version:             2.10.10
API version:         0.8.4
Python version:      3.11.2
Git commit:          8159450b
Built:               Thu, May 18, 2023 3:43 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         server
```

BitTheByte:
For some reason, prefect cloud is not using Prefect base client ?

https://github.com/PrefectHQ/prefect/blob/main/src/prefect/client/cloud.py#L59

vp-walter:
We've also encountered this issue today with many of our flows failing. I realize now, thanks to @BitTheByte 's comments, that we inadvertently upgraded to prefect 2.10.10

BitTheByte:
Turns out this was the problem, I've updated our pipelines to use my patched version and everything seems fine

madkinsz:
Hm I'm a bit confused — the Cloud client is only used for authentication. The normal client is used for all orchestration behavior including the `task_runs` route you shared logs for. It seems unlikely that the lack of retries on the Cloud client is the issue. It seems more likely that you encountered actual rate limits? We'll only retry so many times on 429.

BitTheByte:
[...]

PR #9724, opened 2023-05-25
Use PrefectHttpxClient in cloud communication

BitTheByte:
closes https://github.com/PrefectHQ/prefect/issues/9723


madkinsz:
I do not think this will effect https://github.com/PrefectHQ/prefect/issues/9723 but seems like a good change nonetheless.

PR #9736, opened 2023-05-25
fix: use counts for worker events instrumentation tests

parkedwards:
After adding `poll.*` events from #9702, this uncovered that some of our event test assertions expected consistent ordering, when they are not - this led to some test flakiness

Here, we'll use `len()` checks instead, to assert the count of events we'd expect.  In some cases, we'll still use positional identifiers to check single-member sub-lists, but I think this is OK

### Checklist


bunchesofdonald (approved):
This all looks great. It reminds me of Nebula's AssertingAuditor where it can assert that certain audit events happened (https://github.com/PrefectHQ/nebula/blob/main/src/prefect_cloud/nebula/audit.py#L419), maybe something we'll want to replicate for the this asserting events client, I'll write up a ticket.

bunchesofdonald:
I created issue #9741 for adding some `assert_` methods to the client.

Issue #7709, opened 2022-12-01
Investigate issues with async engine cleanup


madkinsz:
### First check

- [X] I am a contributor to the Prefect codebase

### Description

Suspicious logs associated with `TimeoutError` flakes in Python 3.7 CI https://github.com/PrefectHQ/prefect/actions/runs/3589617486/jobs/6042222320

```
 Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x7f8aeb59e560>
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/asyncio/base_events.py", line 482, in _asyncgen_finalizer_hook
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/_weakrefset.py", line 114, in discard
TypeError: 'NoneType' object is not callable
Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x7f8aa84db5f0>
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/asyncio/base_events.py", line 482, in _asyncgen_finalizer_hook
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/_weakrefset.py", line 114, in discard
TypeError: 'NoneType' object is not callable
```

### Impact

Not only does this worry me for flaky tests, in cases where we are spinning up clients connected to an ephemeral application we may not clean up engines correctly.

anticorrelator:
potentially an instance of bad engine cleanup in [this CI run](https://github.com/PrefectHQ/prefect/actions/runs/3680825302/jobs/6226846180)

madkinsz:
Unfortunately I still see these errors, so we'll want to investigate it when we have the chance.

rmorshea:
I'm seeing something that looks similar:

```
Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x16fe264c0>
Traceback (most recent call last):
  File "/Users/ryanmorshead/.mambaforge/envs/***/lib/python3.10/site-packages/prefect/utilities/asyncutils.py", line 326, in on_shutdown
AttributeError: 'NoneType' object has no attribute 'pop'
Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x28fcfb640>
[...]

PR #9748, opened 2023-05-25
Fix bug where event loop shutdown hooks could fail due to early garbage collection

rmorshea:
# Summary

It seems like it's possible for `EVENT_LOOP_GC_REFS` to be garbage collected earlier than `add_event_loop_shutdown_callback` expects. As a result, you can get the following error (as seen in [this issue](https://github.com/PrefectHQ/prefect/issues/7709#issuecomment-1560021109)):

```
Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x16fe264c0>
  File "/Users/ryanmorshead/.mambaforge/envs/***/lib/python3.10/site-packages/prefect/utilities/asyncutils.py", line 326, in on_shutdown
AttributeError: 'NoneType' object has no attribute 'pop'
Exception ignored in: <async_generator object add_event_loop_shutdown_callback.<locals>.on_shutdown at 0x28fcfb640>
  File "/Users/ryanmorshead/.mambaforge/envs/***/lib/python3.10/site-packages/prefect/utilities/asyncutils.py", line 326, in on_shutdown
AttributeError: 'NoneType' object has no attribute 'pop'
```

This change creates a reference to `EVENT_LOOP_GC_REFS` inside `on_shutdown` so that it survives for at least the lifetime of the coroutine.

I have no idea how to test this unfortunately.

### DIFF
--- a/src/prefect/utilities/asyncutils.py
+++ b/src/prefect/utilities/asyncutils.py
@@ -316,11 +316,13 @@ async def add_event_loop_shutdown_callback(coroutine_fn: Callable[[], Awaitable]
     asyncio does not provided _any_ other way to clean up a resource when the event
     loop is about to close.
     """
-    # EVENT_LOOP_GC_REFS can randomly become `None` (no idea why)
-    # See: https://github.com/PrefectHQ/prefect/issues/7709#issuecomment-1560021109
-    local_event_loop_gc_refs = EVENT_LOOP_GC_REFS
 
     async def on_shutdown(key):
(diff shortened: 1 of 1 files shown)


madkinsz (approved):
How are you testing this?

rmorshea:
I'm running Prefect manually. As mentioned, I have no idea how to write a unit test that replicates this garbage collection issue.

rmorshea:
Thanks!

PR #9732, opened 2023-05-25
Allow table artifacts as list of lists

serinamarie:
### Overview
We're able to render table artifacts in the UI where the input is a `list[list]`, and we should allow it.

### Example

```python
@task
def my_table_task():
    create_table_artifact(
        key="sales-data-table",
        table=[
            ["Region", "Revenue"],
            ["North America", 500000],
            ["Europe", 250000],
            ["Asia", 150000],
            ["South America", 75000],
            ["Africa", 25000],
        ],
        description="## Sales by Region",
    )
```

### Checklist

- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [x] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`, `docs`.

PR #9751, opened 2023-05-26
fixed issue

shindodkar:


madkinsz (changes requested):
Hi thanks for contributing!

- Those commands are how you can test this; not what you should change the description to
- This is the wrong field. `State` is not the one we're deprecating.
- You'll want to set the deprecated boolean as done on the example I linked
- You'll want to update the description to suggest using the other route to create a flow run
- This pull request doesn't meet our standards for description / title — see [#9732](https://github.com/PrefectHQ/prefect/pull/9732) for example

Feel free to let us know if you have questions.

shindodkar:
@madkinsz  since I am new to open source. I am little off here and there. I would be grateful if you could guide me on my improvements.

madkinsz:
@shindodkar we're happy to help out. I think it would be helpful for you to review the issue and the linked examples again.

PR #9771, opened 2023-05-29
Fix default value comment in docs/concepts/variables

marco-buttu:
This PR only affects documentation.  Here is the current example in [concepts/variables](https://docs.prefect.io/latest/concepts/variables/):

```python
# with a default value
answer = variables.get('not_the_answer')
print(answer)
# None

# with a default value
answer = variables.get('not_the_answer', default='42')
print(answer)
# 42
```

The first comment should be: `#without a default value`.

### DIFF
--- a/docs/concepts/variables.md
+++ b/docs/concepts/variables.md
@@ -67,7 +67,7 @@ answer = await variables.get('the_answer')
 print(answer)
 # 42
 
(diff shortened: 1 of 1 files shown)


billpalombi (approved):
Thank you!

PR #9772, opened 2023-05-29
The link to Ray page was not properly formatted

marco-buttu:
As you can see [in the doc](https://docs.prefect.io/latest/concepts/task-runners/#running-tasks-on-ray), speaking about *Ray's support for Python 3.11*, the link is not properly formatted and there is even a wrong URL in the middle.


billpalombi (approved):
Thank you!

PR #9792, opened 2023-05-31
Remove outdated announcement

jawnsy:
We previously had announcement text, which is not visible on the page but appears in the generated HTML and thus would appear in search engines indexes.

### Checklist


billpalombi (approved):
Thank you!

Issue #9794, opened 2023-05-31
Parameter 'priority' is ignored when creating work queues


jozo:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I wrote a function to create 3 work queues with specified priority (10, 20 and 30). Work queues are created but priority is the default one (2,3,4).

![Screenshot from 2023-05-31 13-46-29](https://github.com/PrefectHQ/prefect/assets/4247916/2600f273-cffe-428f-8cd6-6653c913b88b)

### Reproduction

```python3
async def create_work_queues():
    async with get_client() as client:
        print("Creating work queues ...")
        queues = await client.read_work_queues("default-agent-pool")
        names = {queue.name for queue in queues}
        print(f"Existing work queues: {names}")
        tasks = []
        if "high" not in names:
            tasks.append(client.create_work_queue("high", priority=10))
        if "medium" not in names:
            tasks.append(client.create_work_queue("medium", priority=20))
        if "low" not in names:
            tasks.append(client.create_work_queue("low", priority=30))

        await asyncio.gather(*tasks)
        print("Finished creating work queues")
```

### Versions

```Text
Version:             2.10.11
API version:         0.8.4
Python version:      3.11.3
Git commit:          8c651ffc
Built:               Thu, May 25, 2023 2:59 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         server
```

PR #9795, opened 2023-05-31
Fix bug where requested work queue priority was not respected

madkinsz:
Closes https://github.com/PrefectHQ/prefect/issues/9794


madkinsz:
We've implemented a bit more complex logic at https://github.com/PrefectHQ/nebula/pull/3778 which should probably be duplicated in full here.

madkinsz:
See https://github.com/PrefectHQ/prefect/pull/9798 instead

PR #9813, opened 2023-06-01
Add `prefect.runtime.deployment.schedule`

madkinsz:
Adds support to retrieve the deployment's schedule from the runtime.

Prompted by https://prefecthq.slack.com/archives/C046WGGKF4P/p1685643356397549

### Example

This would be useful for getting the next scheduled start time e.g. `runtime.deployment.schedule.get_dates(n=1)` but the client-side schemas do not support `get_dates` yet. We could duplicate this functionality from server -> client if desired.

We could add follow-up runtime attributes such as `next_scheduled_time` / `prev_scheduled_time` that use the flow run scheduled time to calculate the values for the user.

### Checklist

### DIFF
--- a/src/prefect/runtime/deployment.py
+++ b/src/prefect/runtime/deployment.py
@@ -33,7 +33,7 @@ from prefect.context import FlowRunContext
 
 from .flow_run import _get_flow_run
 
-__all__ = ["id", "flow_run_id", "name", "parameters", "version"]
+__all__ = ["id", "flow_run_id", "name", "parameters", "version", "schedule"]
 
 CACHED_DEPLOYMENT = {}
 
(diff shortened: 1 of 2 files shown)


serinamarie (approved):

billpalombi:
I definitely see the use case for  `next_scheduled_time` / `prev_scheduled_time` and would love to add those. I'm haven't trouble thinking of the use case for getting the entire schedule. Do you have one in mind?

madkinsz:
@billpalombi simple example is if you want the next two dates instead of the next one.

I defer to @taylor-curran / @biancaines who were looking for this sort of functionality.

madkinsz:
@taylor-curran if you want this then someone else can continue this work, cc @cicdw / @serinamarie

PR #9817, opened 2023-06-02
Fixes name equals main help string

EmilRex:
Adds missing underscores to name == main help string.


serinamarie (approved):
Thanks @EmilRex!

PR #9854, opened 2023-06-07
Boost non api docs pages

### LINKED ISSUE #9553: Documentation search shows API reference results first
### First check

- [X] I used GitHub search to find a similar request and didn't find it 😇

### Describe the issue

When searching in the docs, the API reference results are often shown before our actual documentation e.g. search for `return_state`

### Describe the proposed change

The API reference search results should _always_ be shown after other documentation results even if it is a worse match.

### Additional context

_No response_

discdiver:
In docs search results, this change boosts non-API pages.

Closes #9553.

### Checklist

### DIFF
--- a/docs/cloud/index.md
+++ b/docs/cloud/index.md
@@ -11,6 +11,8 @@ tags:
     - workspaces
     - PaaS
 title: Prefect Cloud
+search:
+  boost: 2
 ---
 
 # Welcome to Prefect Cloud <span class="badge cloud"></span>
--- a/docs/api-ref/rest-api-reference.md
+++ b/docs/api-ref/rest-api-reference.md
@@ -2,7 +2,7 @@
 description: Prefect REST API interactive reference.
 tags:
     - REST API
-    - Prefect Cloud
+    - Prefect server
 hide:
     - toc
 title: Prefect Server REST API reference
--- a/docs/cloud/automations.md
+++ b/docs/cloud/automations.md
@@ -7,6 +7,8 @@ tags:
     - events
     - triggers
     - Prefect Cloud
+search:
+  boost: 2
 ---
 
 # Automations <span class="badge cloud"></span>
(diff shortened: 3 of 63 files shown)


discdiver:
The non-API pages are elevated in some searches.

See Screenshot below for a comparison of before and after search results when searching for "web" (existing docs on the left, preview docs on the right). The top two results are API pages in the existing docs and only one is an API page in the PR preview.

Screenshot 2023-06-07 at 12 51 11 PM

docs/cloud/index.md
```diff
@@ -11,6 +11,8 @@ tags:
     - workspaces
     - PaaS
 title: Prefect Cloud
+search:
+  boost: 2
```
madkinsz, line 15: Do we really need to remember to specify this on every file we write?

madkinsz: Is there a way to set a default boost then turn it _down_ in the api reference?

discdiver: Yeah, no parent unfortunately.

I'm assuming that when someone creates a new page they copy the meta-data at the top over from another page (at least I do) - this includes tags and other information. As all of the pages one would copy from would include these two lines, I don't believe that there would be an increased burden. 

The downside if the boost isn't included is relatively low, as the boost value chosen (2) isn't that high.

If the boost value were desired to be changed, the text editor search/replace functionality should make it a low lift.

chrisguidry: @discdiver what if we downrank the API reference with a fractional boost:

https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/#rank-down

```
---
search:
  boost: 0.5
---
```

?

discdiver: We could do that @chrisguidry, but there are even more API reference docs, so went this route.

chrisguidry: Ohhh right right, gotcha

chrisguidry (approved):
This is going to be great for users!

Issue #9412, opened 2023-05-03
Async context manager in flow causes RuntimeError: Event loop is closed


TWeatherston:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I have a flow that makes numerous requests to the same API. I use a httpx client to make these requests and each request has its own task, with the client being passed as a parameter to the task.

This was **working fine on Prefect version v2.8.4** but we recently upgraded to the latest version and this is now throwing a RuntimeError: Event loop is closed exception. I seem to **get this error on anything >= v2.8.7.**

If I change the task to a regular method then the flow completes just fine.

### Reproduction

```python3
from prefect import flow, task
import asyncio
import httpx


@task()
async def query_api(client: httpx.AsyncClient):
    resp = await client.get("data")
    return resp


@flow()
async def my_flow():
    async with httpx.AsyncClient(base_url="https://example.com") as client:
        print(await query_api(client))


if __name__ == "__main__":
    asyncio.run(my_flow())
```

### Error

```python3
Traceback (most recent call last):
  File "/home/project/atheon_prefect/flows/collection/my_flow.py", line 19, in <module>
    asyncio.run(my_flow())
  File "/home/.pyenv/versions/3.10.3/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/.pyenv/versions/3.10.3/lib/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File "/home/project/.venv/lib/python3.10/site-packages/prefect/_internal/concurrency/api.py", line 109, in wait_for_call_in_loop_thread
    return call.result()
  File "/home/project/.venv/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 173, in result
    return self.future.result(timeout=timeout)
  File "/home/.pyenv/versions/3.10.3/lib/python3.10/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/home/.pyenv/versions/3.10.3/lib/python3.10/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/home/project/.venv/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 218, in _run_async
    result = await coro
[...]

PR #9855, opened 2023-06-07
Run tasks on the main thread

zanieb:
When feasible, task runs are now executed on the main thread instead of a worker thread. Previously, all task runs were run in a new worker thread. This should significantly improve performance in asynchronous and sequential use cases. This allows objects to be passed to and from tasks without worrying about thread safety unless you have opted into concurrency. For example, an HTTP client or database connection can be shared between a flow and its tasks now (unless synchronous concurrency is being used).

Tasks are run in the main thread when:
- The task and flow are both asynchronous
- The flow is synchronous and the task run is sequential

In a kind of funny way, this causes _deadlocks_ when asynchronous and synchronous tasks are mixed in an asynchronous flow because if a synchronous task depends on an asynchronous task it will block the event loop waiting for the upstream to finish but the upstream cannot run since it is now on the same event loop. For example:

```python
@flow
async def foo():
    upstream = await async_task.submit(42)  # Running in the background on the main thread event loop
    sync_task(upstream) # Blocks the event loop because there's no `await`
    # If the upstream is not done when `sync_task` is called, we deadlock

await foo()
```

To resolve this, I've added `greenback` 🤮 as a dependency. We no longer block the event loop when synchronous tasks are called from asynchronous flows which resolves this issue _and_ addresses general performance concerns when calling synchronous tasks from asynchronous flows.

Closes https://github.com/PrefectHQ/prefect/issues/9412

## Example

```python
import asyncio
import threading

@task
def sync_task():
    print(threading.current_thread())


@task
async def async_task():
    print(threading.current_thread())


@flow(log_prints=True)
def sync_flow():
    print(threading.current_thread())

    sync_task()  # Runs on main thread
    async_task()  # Runs on main thread
    sync_task.submit()  # Runs on worker thread
    async_task.submit()  # Runs on worker thread


@flow(log_prints=True)
async def async_flow():
    print(threading.current_thread())

    sync_task()  # Runs on worker thread
    sync_task.submit()  # Runs on worker thread
    await async_task()  # Runs on main thread
    await async_task.submit()  # Runs on main thread


sync_flow()
print("---")
asyncio.run(async_flow())
```

```
❯ python example.py
12:34:50.735 | INFO    | prefect.engine - Created flow run 'organic-wallaby' for flow 'sync-flow'
12:34:50.799 | INFO    | Flow run 'organic-wallaby' - <_MainThread(MainThread, started 8725699392)>
12:34:50.814 | INFO    | Flow run 'organic-wallaby' - Created task run 'sync_task-0' for task 'sync_task'
12:34:50.814 | INFO    | Flow run 'organic-wallaby' - Executing 'sync_task-0' immediately...
12:34:50.843 | INFO    | Task run 'sync_task-0' - <_MainThread(MainThread, started 8725699392)>
12:34:50.856 | INFO    | Task run 'sync_task-0' - Finished in state Completed()
12:34:50.870 | INFO    | Flow run 'organic-wallaby' - Created task run 'async_task-0' for task 'async_task'
12:34:50.871 | INFO    | Flow run 'organic-wallaby' - Executing 'async_task-0' immediately...
12:34:50.895 | INFO    | Task run 'async_task-0' - <_MainThread(MainThread, started 8725699392)>
12:34:50.911 | INFO    | Task run 'async_task-0' - Finished in state Completed()
12:34:50.924 | INFO    | Flow run 'organic-wallaby' - Created task run 'sync_task-1' for task 'sync_task'
12:34:50.925 | INFO    | Flow run 'organic-wallaby' - Submitted task run 'sync_task-1' for execution.
12:34:50.935 | INFO    | Flow run 'organic-wallaby' - Created task run 'async_task-1' for task 'async_task'
12:34:50.935 | INFO    | Flow run 'organic-wallaby' - Submitted task run 'async_task-1' for execution.
12:34:50.958 | INFO    | Task run 'sync_task-1' - <Thread(WorkerThread-0, started 6246838272)>
12:34:50.973 | INFO    | Task run 'async_task-1' - <Thread(WorkerThread-1, started 10804703232)>
12:34:50.977 | INFO    | Task run 'sync_task-1' - Finished in state Completed()
12:34:50.988 | INFO    | Task run 'async_task-1' - Finished in state Completed()
12:34:51.003 | INFO    | Flow run 'organic-wallaby' - Finished in state Completed('All states completed.')
---
12:34:51.071 | INFO    | prefect.engine - Created flow run 'melodic-bullfinch' for flow 'async-flow'
12:34:51.115 | INFO    | Flow run 'melodic-bullfinch' - <_MainThread(MainThread, started 8725699392)>
12:34:51.130 | INFO    | Flow run 'melodic-bullfinch' - Created task run 'sync_task-0' for task 'sync_task'
12:34:51.130 | INFO    | Flow run 'melodic-bullfinch' - Executing 'sync_task-0' immediately...
12:34:51.154 | INFO    | Task run 'sync_task-0' - <Thread(WorkerThread-2, started 10771050496)>
12:34:51.174 | INFO    | Task run 'sync_task-0' - Finished in state Completed()
12:34:51.186 | INFO    | Flow run 'melodic-bullfinch' - Created task run 'sync_task-1' for task 'sync_task'
12:34:51.187 | INFO    | Flow run 'melodic-bullfinch' - Submitted task run 'sync_task-1' for execution.
12:34:51.193 | INFO    | Flow run 'melodic-bullfinch' - Created task run 'async_task-0' for task 'async_task'
12:34:51.193 | INFO    | Flow run 'melodic-bullfinch' - Executing 'async_task-0' immediately...
12:34:51.219 | INFO    | Task run 'sync_task-1' - <Thread(WorkerThread-3, started 6263664640)>
12:34:51.231 | INFO    | Task run 'async_task-0' - <_MainThread(MainThread, started 8725699392)>
12:34:51.239 | INFO    | Task run 'sync_task-1' - Finished in state Completed()
12:34:51.248 | INFO    | Task run 'async_task-0' - Finished in state Completed()
12:34:51.259 | INFO    | Flow run 'melodic-bullfinch' - Created task run 'async_task-1' for task 'async_task'
12:34:51.259 | INFO    | Flow run 'melodic-bullfinch' - Submitted task run 'async_task-1' for execution.
12:34:51.282 | INFO    | Task run 'async_task-1' - <_MainThread(MainThread, started 8725699392)>
12:34:51.295 | INFO    | Task run 'async_task-1' - Finished in state Completed()
12:34:51.309 | INFO    | Flow run 'melodic-bullfinch' - Finished in state Completed('All states completed.')
```

Now possible, previously would fail with event loop errors

```python
import asyncio
import httpx


@task()
async def query_api(client: httpx.AsyncClient):
    resp = await client.get("data")
    return resp


@flow()
async def my_flow():
    async with httpx.AsyncClient(base_url="https://example.com") as client:
        print(await query_api(client))


if __name__ == "__main__":
    asyncio.run(my_flow())
```

### DIFF
--- a/src/prefect/_internal/concurrency/api.py
+++ b/src/prefect/_internal/concurrency/api.py
@@ -31,6 +31,7 @@ from prefect._internal.concurrency.waiters import (
     get_waiter_for_thread,
 )
 from prefect._internal.concurrency.event_loop import get_running_loop
+from prefect._internal.concurrency import logger
 
 P = ParamSpec("P")
 T = TypeVar("T")
@@ -235,21 +236,22 @@ class from_sync(_base):
     ) -> Awaitable[T]:
         # Check for a running event loop to prevent blocking
         if get_running_loop():
-            if not greenback.has_portal():
-                raise RuntimeError(
+            if greenback.has_portal():
+                # Use greenback to avoid blocking the event loop while waiting
+                return greenback.await_(
+                    from_async.wait_for_call_in_loop_thread(
+                        __call,
+                        timeout=timeout,
+                        done_callbacks=done_callbacks,
+                        contexts=contexts,
+                    )
+                )
+            else:
+                logger.error(
                     "Detected unsafe call to `from_sync` from thread with event loop. "
-                    "Call `await greenback.ensure_portal()` to allow functionality."
+                    "Use `await greenback.ensure_portal()` to allow call to run "
+                    "without blocking the event loop."
                 )
-
-            # Use greenback to avoid blocking the event loop while waiting
-            return greenback.await_(
-                from_async.wait_for_call_in_loop_thread(
-                    __call,
-                    timeout=timeout,
-                    done_callbacks=done_callbacks,
-                    contexts=contexts,
-                )
-            )
 
         call = _cast_to_call(__call)
         waiter = SyncWaiter(call)
--- a/src/prefect/_internal/concurrency/api.py
+++ b/src/prefect/_internal/concurrency/api.py
@@ -6,6 +6,7 @@ import asyncio
 import concurrent.futures
 import contextlib
 import threading
+import greenback
 from typing import (
     Awaitable,
     Callable,
@@ -29,6 +30,7 @@ from prefect._internal.concurrency.waiters import (
     SyncWaiter,
     get_waiter_for_thread,
 )
+from prefect._internal.concurrency.event_loop import get_running_loop
 
 P = ParamSpec("P")
 T = TypeVar("T")
@@ -179,7 +181,7 @@ class from_async(_base):
             for context in contexts or []:
                 stack.enter_context(context)
             await waiter.wait()
-            return call.result()
+            return await call.aresult()
 
     @staticmethod
     async def wait_for_call_in_new_thread(
@@ -193,7 +195,7 @@ class from_async(_base):
             waiter.add_done_callback(callback)
         _base.call_soon_in_new_thread(call, timeout=timeout)
         await waiter.wait()
-        return call.result()
+        return await call.aresult()
 
     @staticmethod
     def call_in_waiting_thread(
@@ -231,6 +233,24 @@ class from_sync(_base):
         done_callbacks: Optional[Iterable[Call]] = None,
         contexts: Optional[Iterable[ContextManager]] = None,
     ) -> Awaitable[T]:
+        # Check for a running event loop to prevent blocking
+        if get_running_loop():
+            if not greenback.has_portal():
+                raise RuntimeError(
+                    "Detected unsafe call to `from_sync` from thread with event loop. "
+                    "Call `await greenback.ensure_portal()` to allow functionality."
+                )
+
+            # Use greenback to avoid blocking the event loop while waiting
+            return greenback.await_(
+                from_async.wait_for_call_in_loop_thread(
+                    __call,
+                    timeout=timeout,
+                    done_callbacks=done_callbacks,
+                    contexts=contexts,
+                )
+            )
+
         call = _cast_to_call(__call)
         waiter = SyncWaiter(call)
         _base.call_soon_in_loop_thread(call, timeout=timeout)
@@ -262,6 +282,10 @@ class from_sync(_base):
         thread: threading.Thread,
         timeout: Optional[float] = None,
     ) -> T:
+        if get_running_loop():
+            raise RuntimeError(
+                "Detected unsafe call to `from_sync` from thread with event loop."
+            )
         call = _base.call_soon_in_waiting_thread(__call, thread, timeout=timeout)
         return call.result()
 
@@ -269,6 +293,10 @@ class from_sync(_base):
     def call_in_new_thread(
         __call: Union[Callable[[], T], Call[T]], timeout: Optional[float] = None
     ) -> T:
+        if get_running_loop():
+            raise RuntimeError(
+                "Detected unsafe call to `from_sync` from thread with event loop."
+            )
         call = _base.call_soon_in_new_thread(__call, timeout=timeout)
         return call.result()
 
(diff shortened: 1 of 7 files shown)


src/prefect/_internal/concurrency/api.py
```diff
@@ -179,7 +181,7 @@ async def wait_for_call_in_loop_thread(
             for context in contexts or []:
                 stack.enter_context(context)
             await waiter.wait()
-            return call.result()
+            return await call.aresult()
```
madkinsz, line 184: No change in behavior here, just clearer to call `aresult` even though `result` is safe since we just finished waiting.

src/prefect/_internal/concurrency/api.py
```diff
@@ -231,6 +233,24 @@ def wait_for_call_in_loop_thread(
...
+
+            # Use greenback to avoid blocking the event loop while waiting
+            return greenback.await_(
+                from_async.wait_for_call_in_loop_thread(
+                    __call,
+                    timeout=timeout,
+                    done_callbacks=done_callbacks,
+                    contexts=contexts,
+                )
+            )
```
madkinsz, line 252: Wow isn't this magical!

Here, we avoid blocking the event loop in a `from_sync` call from a thread with an event loop by using `greenback` to schedule the call as a `from_async` waiter instead!

This error message is not intended to be user-facing.

src/prefect/_internal/concurrency/calls.py
```diff
@@ -339,6 +340,14 @@ def _run_sync(self):
...
 
     async def _run_async(self, coro):
+        from prefect._internal.concurrency.threads import in_global_loop
+
+        # Ensure the greenback portal is shimmed for this task so we can safely call
+        # back into async code from sync code if the user does something silly; avoid
+        # this overhead for our internal event loop
+        if not in_global_loop():
+            await greenback.ensure_portal()
+
```
madkinsz, line 350: I had a tough time figuring out just _where_ to inject `ensure_portal`.

This is the only place that works without having the user do it themselves.

If we call this in the global event loop we actually can cause deadlocks.

src/prefect/engine.py
```diff
@@ -1387,6 +1402,8 @@ async def submit_task_run(
             ),
             log_prints=should_log_prints(task),
             settings=prefect.context.SettingsContext.get().copy(),
+            user_thread=user_thread,
+            concurrency_type=task_runner.concurrency_type,
```
madkinsz, line 1406: We need the concurrency type so we can check if the task run is sequential for scheduling on the main thread. We can't use `flow_run_context.task_runner.concurrency_type` because we use a different task runner for sequential task calls.

madkinsz:
Unfortunately using `greenback` alongside `sqlalchemy` is not possible at this time. If the ephemeral application is running in the same thread as one in which we need to use greenback, then greenback and sqlalchemy will accidentally "mix" messages and crash each other.

There are a few options here:
1. Open an issue on `greenback` and/or `sqlalchemy` and determine the best path towards isolation of their use of greenlets
2. Run the ephemeral application in a separate thread (i.e. #9870); greenlets are isolated by thread boundaries
3. Avoid use of greenlet entirely and just throw an exception if a user passes an asynchronous upstream to a blocking synchronous downstream task
4. Learn a bunch about greenlets and vendor the relevant chunk of greenback with support for SQLAlchemy messages
5. Make users `await` calls to synchronous tasks in asynchronous flows to avoid blocking the event loop in the first place; this would remove the need for greenback

@jakekaplan has the most context on these problems

Additionally, we could defer handling of tasks on the main thread for asynchronous flows and immediately release support for this feature in synchronous flows.

cpdeethree:
I would also absolutely love to see this get merged. It's blocking adoption of using Prefect for us because the current implementation creates incompatibilities with other, existing async libraries that we are using.

jakekaplan:
Looking into the best path forward to get this merged

cpdeethree:
> Looking into the best path forward to get this merged

Thank you jake! looking forward to seeing it get in!

milonimrod:
Any update on this? This is a blocker for us. It prevents us from using aiohttp, aiokafka and any shared async package

cpdeethree:
@jakekaplan any chance we could get a new pr going? Don't want this to fall though the cracks as a missed opportunity for a great improvement!

PR #9867, opened 2023-06-08
Adds note explaining multiple match_related syntax

EmilRex:
Adds note to Automations docs explaining multiple `match_related` syntax.


chrisguidry (approved):
That's spot-on, @EmilRex. This also applies to the `match` as well.

WillRaphaelson (approved):
awesome thanks

Issue #9695, opened 2023-05-23
Allow Prefect client to be used from synchronous contexts


MattDelac:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

`get_client()` from the python SDK is async and using a context manager.
So it means that all the functions using it would have to be async too (whether or not the program runs (a)synchronously)

Using the following does not work with a context manager
```python
asyncio.run(with get_client()) as client:
    ...
```

### Describe the proposed behavior

```python
client = asyncio.run(get_client())
client.another_method()
# OR
asyncio.run(client.another_method())
```

And ideally I could get the client by using `asyncio.run()` so that after I don't need to do any async operation. And my script could remain totally synchronous.

### Additional context

As per our convo with the support

> Is there a way to run the client SDK without async? this just adds complexity in our codebase for no good reason. Also I would have to make most of our functions async too ...
> For us there is no gain is using the SDK async. Our scripts' performance is unimportant as we use the SDK for light operations only.
>
> I don't think I can use the asyncio.run() function with a context manager.
Basically the following does not work
>
> ```
> asyncio.run(with get_client()) as client:
>    ...
> ```
> 
> And as soon as I make anything "async" or "await", I would need to mark the calling function also async ... Which is like a big refactor of our codebase and how our users were using our internal SDK

madkinsz:
This is on our roadmap.

In the meantime, you can use asynchronous tasks from synchronous flows without changing the rest of your code. This would allow you to use the asynchronous client without changing the rest of your code.

>  this just adds complexity in our codebase for no good reason.

There is a good reason — since our engine is asynchronous using the client asynchronously will be more performant. That said, we understand that a synchronous interface is still very valuable and convenient.

MattDelac:
> There is a good reason — since our engine is asynchronous using the client asynchronously will be more performant

I get the design decision 👍 
just that it adds complexity for users with simpler use cases like "Have a CLI that uses the python SDK to perform some operations". Here speed is unimportant to some extent.


> This is not the case. The following pattern is okay to use:

Yeah I would need to wrap most of the Prefect logic in async functions and then call them synchronously. Realistically this adds complexity.

Right now, we play with the REST endpoints directly.

MattDelac:
> This is not the case. The following pattern is okay to use:
> 
> ```python
[...]

PR #9870, opened 2023-06-08
Run clients with ephemeral applications on the global event loop thread

zanieb:
When a client is hooked to an ephemeral application, its moved to the global event loop instead of the event loop it is created on. We do this by sending the underlying httpx client's context manager to the global event loop and patch the `client.request` method to send coroutines to the other thread. This is a good stepping stone towards providing a singleton client that runs on the global event loop as well as towards providing a synchronous client.

Enabling this only for the ephemeral application is motivated by #9855 which requires SQLAlchemy to be managed on a separate thread than user code.

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -224,7 +224,7 @@ jobs:
           # Parallelize tests by scope to reduce expensive service fixture duplication
           # Do not allow the test suite to build images, as we want the prebuilt images to be tested
           # Do not run Kubernetes service tests, we do not have a cluster available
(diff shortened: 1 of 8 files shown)


madkinsz:
This is:

1. A possible way to fix issues in #9855 
2. A path to a singleton client with a synchronous interface  #9695 

This pull request appears to:

1. Introduce a large performance penalty in tests; this would likely not be seen by users but we are spawning a lot of extra threads in tests since we use a lot of ephemeral client instances.
2. Cause intermittent deadlocks

It's unclear what the best path forward is, but a more holistic approach will need to be taken.

@jakekaplan is best suited to take over this work.

zanieb:
Closing this as I am not working on it and a new pull request will need to be opened if someone else takes it up.

PR #9872, opened 2023-06-08
Provide a default `DTSTART` to anchor `RRULE` schedules

chrisguidry:
We discovered that `RRULE` schedules that didn't include their own embedded
`DTSTART` were prone to producing a ton of extraneous schedules.  Each time
you asked one of these `RRULE` schedules to generate dates, it used the current
time as the anchor, so some types of rules would produce duplicates over and
over again because they weren't idempotent.

Adding a fixed `DTSTART` (which is only used if the rule doesn't have an
embedded one) makes sure that we always produce a stable set of dates/times no
matter when we ask.

### DIFF
--- a/src/prefect/client/schemas/schedules.py
+++ b/src/prefect/client/schemas/schedules.py
@@ -144,6 +144,9 @@ class CronSchedule(PrefectBaseModel):
         return v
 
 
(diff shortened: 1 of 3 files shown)


parkedwards (approved):

madkinsz (approved):
Nice!

PR #9874, opened 2023-06-09
feat: Cloud Webhook CLI

parkedwards:
relates to https://github.com/PrefectHQ/nebula/issues/4817

Adds the `prefect cloud webhook` command group, docs, and tests

### Checklist


src/prefect/cli/__init__.py
```diff
@@ -8,7 +8,8 @@
 import prefect.cli.agent
 import prefect.cli.artifact
 import prefect.cli.block
-import prefect.cli.cloud
+import prefect.cli.cloud.cloud
+import prefect.cli.cloud.webhook
```
parkedwards, line 12: not entirely sure if this is the right move, but since we're expanding the `prefect.cli.cloud` module to include `webhooks` now, I thought it might be time to nest `cloud` commands into a `/cloud` subdirectory

docs/cloud/webhooks.md
```diff
@@ -37,7 +37,48 @@ Cloud](/cloud/connecting#manually-configure-prefect-api-settings).
 
 ### Via the Prefect CLI
 
-_More information about using the Prefect CLI to manage webhooks coming soon!_
+Webhooks can be managed and interacted with via the `prefect cloud webhook` command group.
```
chrisguidry, line 40: :hot_pepper: Nice work including the docs in the same PR

docs/cloud/webhooks.md
```diff
@@ -37,7 +37,48 @@ Cloud](/cloud/connecting#manually-configure-prefect-api-settings).
...
+
+You can create your first webhook by invoking `create`:
+
+```bash
+prefect cloud webhook create your-webhook-name \
+    --description "Receives webhooks from your system" \
+    --template '{ "event": "your.event.name", "resource": { "prefect.resource.id": "your.resource.id" } }'
+```
+
+Note the template string, which is discussed in [greater detail down below](##Webhook templates)
```
chrisguidry, line 54:
```suggestion
Note the template string, which is discussed in [greater detail down below](#webhook-templates)
```

src/prefect/cli/__init__.py
chrisguidry: I like this move, but I think what you want is to rename `cloud/cloud.py` to `cloud/__init__.py` and then all the imports can still be `prefect.cli.cloud`.  Then we can (later) start separating the `cloud` commands out into submodules.  (We're overdue some CLI commands for Automations, so I may well borrow your structure here for that)

src/prefect/cli/cloud/webhook.py
```diff
@@ -0,0 +1,197 @@
...
+
+    if not webhook_id and not all:
+        exit_with_error("Please provide a webhook ID or use the --all flag")
+
+    if webhook_id and all:
+        exit_with_error(
+            "Please provide a webhook ID or use the --all flag, but not both"
+        )
+
+    confirm_logged_in()
```
chrisguidry, line 43: It feels like there must be a way to add this as a hook for all the commands in this module, but I'm not sure how

chrisguidry (approved):
I love it, just that note about `__init__.py` to start the new `cloud` submodule

src/prefect/cli/cloud/webhook.py
parkedwards: yeah it's a tedious without some automagic hook, but i couldn't find anything useful in the typer docs. fortunately, the hard part of making this into a util was already done 😄

PR #9877, opened 2023-06-09
Use existing thread in `BatchedQueueService` to reduce queue retrieval overhead

madkinsz:


src/prefect/_internal/concurrency/services.py
```diff
@@ -282,9 +280,9 @@ async def _main_loop(self):
             deadline = get_deadline(self._min_interval)
             while batch_size < self._max_batch_size:
                 try:
-                    item = await anyio.to_thread.run_sync(
-                        partial(self._queue.get, timeout=get_timeout(deadline))
-                    )
+                    item = await self._queue_get_thread.submit(
```
madkinsz, line 283: This thread is defined on the parent class is used in `QueueService` but we forgot to use it here!

chrisguidry (approved):

PR #9878, opened 2023-06-09
Use internal concurrency tooling instead of AnyIO to run state change hooks

zanieb:
In a mission to move over to the new concurrency tooling consistently and avoid spurious AnyIO threads.


chrisguidry (approved):
:hot_pepper:

PR #9886, opened 2023-06-09
deployment triggers docs

WillRaphaelson:
### Checklist


docs/concepts/deployments.md
```diff
@@ -503,6 +507,30 @@ $ prefect deployment inspect 'Cat Facts/catfact'
...
+
+```yaml
+triggers:
+  - enabled: true
+    match:
+      prefect.resource.id: prefect.flow-run.*
+    expect:
+      - prefect.flow-run.Completed
+    match_related:
+      prefect.resource.id: prefect.flow.c5568579-9f63-415c-a53e-25f00fc35866
```
bunchesofdonald, line 524: I think we should make these examples use the `prefect.resource.name`, my thinking here is that oftentimes (at least in my usage) the flow might not even exist yet (though you'd likely know the name) or the flow might exist but to add this section you're going to have to go to the UI or CLI to lookup the id which both seem to semi-defeat the purpose of this shorthand.

bunchesofdonald (approved):

docs/concepts/deployments.md
WillRaphaelson: good call

PR #9888, opened 2023-06-10
Deprecate project deployment via `—flow/-f`

serinamarie:
### Overview 
This PR adds a deprecation warning when using `deploy` with a flow name by `--flow/-f`.

It is the first step towards removal of ability to deploy via flow name, which will occur when deployment ux leaves beta. See the draft PR for that at https://github.com/PrefectHQ/prefect/pull/9881.

### Example
```bash
❯ prefect deploy -f my-deployment
08:17:37.571 | DEBUG   | prefect.profiles - Using profile 'local_sqlite'
08:17:38.802 | DEBUG   | prefect.client - Using ephemeral application with database at sqlite+aiosqlite:////Users/bean/.prefect/prefect.db
The `prefect deploy --flow/-f` command has been deprecated. It will not be available after Jun 2023. 
Use `prefect deploy` to specify an entrypoint instead.
? Would you like to schedule when this flow runs? [y/n] (y): 
```

### Checklist

### DIFF
--- a/src/prefect/cli/deploy.py
+++ b/src/prefect/cli/deploy.py
@@ -394,9 +394,9 @@ async def _run_single_deploy(
     elif flow_name:
         app.console.print(
             generate_deprecation_message(
-                "The `prefect deploy --flow/-f` command",
+                "The ability to deploy by flow name",
                 end_date="Jun 2023",
-                help="Use `prefect deploy` to specify an entrypoint instead.",
+                help="\nUse `prefect deploy` to specify an entrypoint instead.",
             )
         )
         prefect_dir = find_prefect_directory()
--- a/src/prefect/cli/deploy.py
+++ b/src/prefect/cli/deploy.py
@@ -396,7 +396,7 @@ async def _run_single_deploy(
             generate_deprecation_message(
                 "The `prefect deploy --flow/-f` command",
                 end_date="Jun 2023",
-                help="Use `prefect deploy` instead.",
+                help="Use `prefect deploy` to specify an entrypoint instead.",
             )
         )
         prefect_dir = find_prefect_directory()
--- a/src/prefect/cli/deploy.py
+++ b/src/prefect/cli/deploy.py
@@ -47,6 +47,8 @@ from prefect.utilities.slugify import slugify
 
 from prefect.client.orchestration import PrefectClient
 
+from prefect._internal.compatibility.deprecated import generate_deprecation_message
+
 
 @app.command()
 async def deploy(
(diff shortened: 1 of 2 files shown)


src/prefect/cli/deploy.py
```diff
@@ -390,6 +392,13 @@ async def _run_single_deploy(
                 )
         flow_name = flow.name
     elif flow_name:
+        app.console.print(
+            generate_deprecation_message(
+                "The `prefect deploy --flow/-f` command",
+                end_date="Jun 2023",
+                help="Use `prefect deploy` to specify an entrypoint instead.",
```
madkinsz, line 399: For the deprecation message, maybe "The`--flow` option has been deprecated." or "The ability to deploy by flow name has been deprecated" 

It also looks like you're missing the new option in the help message, perhaps "Use `--entrypoint` to specify the import path of your flow instead."`

serinamarie: > For the deprecation message, maybe "The--flow option has been deprecated." or "The ability to deploy by flow name has been deprecated"

Good idea! I've changed it.

> It also looks like you're missing the new option in the help message, perhaps "Use --entrypoint to specify the import path of your flow instead."`

I was thinking that too but there's no `--entrypoint` option. So `prefect deploy` is the new way.

madkinsz: Ah if it's an argument we should suggest `prefect deploy <entrypoint>` instead

src/prefect/cli/deploy.py
```diff
@@ -390,6 +392,16 @@ async def _run_single_deploy(
                 )
         flow_name = flow.name
     elif flow_name:
+        app.console.print(
+            generate_deprecation_message(
+                "The ability to deploy by flow name",
+                end_date="Jun 2023",
+                help=(
+                    "\nUse `prefect deploy ./path/to/file.py:flow_fn_name` to specify"
```
madkinsz, line 400: Hm. Technically we could do even better here if we wanted. We could suggest the _actual_ path and flow name to use if it's present in the file i.e. from `entrypoint = flows[flow_name]`

madkinsz: Not sure if you feel like that's worth it

serinamarie: That would be cool if we were doing a 6-month deprecation but since it's just a couple weeks I don't know that it's worth it.

serinamarie: Projects are currently still in beta 🤷‍♀️ so they're subject to breaking changes

madkinsz: Hm there's no experimental or beta warning on the `prefect deploy` command — we should be careful about that in the future. Anyway it definitely doesn't make sense to polish it for a 2 week period.

serinamarie: 👍

PR #9894, opened 2023-06-12
feature: add `infra_overrides` field in `FlowRun`

### LINKED ISSUE #9893: Allow for flow-run level infrastructure overrides
### First check

### Prefect Version

2.x

There's no way for user to override configurations defined in an infrastructure block when creating a flow run.

Users should be able to adjust infrastructure block's configurations such as CPU limit / request when creating a flow run.

### Example Use

Say we are scheduling a flow run that reads a file into memory with `KubernetesJob` infrastructure type.

If the file is 10 GB large, we could set the `infra_overrides` field with `requests.memory` == 10G, so that we know when the pod is scheduled it is definitely able to load the entire file into memory, and when there isn't enough memory for the flow run it would just stay in `Pending`.

On the other hand, for a 10 KB file we could just have a `requests.memory` of 10 K , so that other pods that have resource request could be scheduled on the Kubernetes cluster.

### Additional context

This is so that we can have more control over the resource usage for each Prefect flow. In our use case, Prefect flows are run on a self hosted Kubernetes cluster. Having this feature enables us to fine tune our resource management mechanism.

usefulalgorithm:
### Overview

This PR adds the ability to override infrastructure block's configuration when creating a FlowRun.

Closes #9893 

### Example

```python
import prefect
client = prefect.client.get_client()

parameters = {
    "pathToAFile": "s3://bucket/10G_file"
}
infra_overrides = {
    "customizations": [
        {
           "op": "add",
           "path": "/spec/template/spec/containers/0/resources",
           "value": {
               "requests": {
                   "memory": "10gi"
               }
            }
         }
    ]
}

flow_run = client.create_flow_run_from_deployment(..., parameters=parameters, infra_overrides=infra_overrides)

# ...

parameters = {
    "pathToAFile": "s3://bucket/1K_file"
}
infra_overrides = {
    "customizations": [
        {
           "op": "add",
           "path": "/spec/template/spec/containers/0/resources",
           "value": {
               "requests": {
                   "memory": "1ki"
               }
            }
         }
    ]
}
another_flow_run = client.create_flow_run_from_deployment(..., parameters=parameters, infra_overrides=infra_overrides)
```
### Checklist


chrisguidry (changes requested):
Thanks for starting a contribution, @usefulalgorithm!

We generally require new unit tests for any new functionality.  

Additionally, these changes don't include the model/database changes that would be required to persist this information across time.  This means that when a later worker or agent picks up this flow run, it won't retain the `infra_overrides` attribute.  To persist the changes, you'd need to [update the ORM model](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/database/orm_models.py#L480) and create the associated database migrations.  

We would also need to coordinate this change with Prefect Cloud, so that the [Prefect Cloud compatibility tests](https://github.com/PrefectHQ/compat-tests) pass.

usefulalgorithm:
Hi @chrisguidry , thanks for the feedback!

> We generally require new unit tests for any new functionality.

I did a unit test to ensure that `create_flow_run_from_deployment` does indeed take `infra_overrides` as a valid parameter on the server side.

>Additionally, these changes don't include the model/database changes that would be required to persist this information across time. This means that when a later worker or agent picks up this flow run, it won't retain the infra_overrides attribute. To persist the changes, you'd need to [update the ORM model](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/database/orm_models.py#L480) and create the associated database migrations.

I added the migration scripts as you suggested, not sure how to test these though...

> We would also need to coordinate this change with Prefect Cloud, so that the [Prefect Cloud compatibility tests](https://github.com/PrefectHQ/compat-tests) pass.

Can you provide me with a pointer or two on how to do this? Can't really tell what I need to do just by looking at the repository you provided.

Thanks again!

chrisguidry:
Ah sorry I wasn't clear: We (the Prefect team) will handle the Prefect Cloud side of this and so we can catch up the compatibility tests.  Once this PR is tightened up and passing tests, I'll help find a maintainer that can help with that.

Regarding testing of the migration and persistence, your migrations will be tested by running all of the CI tests that run as part of our Github actions, so that part doesn't require any additional testing of the migration on your end.

Testing the persistence would need a test that demonstrates that you can create a flow run with `infra_overrides` and then retrieve it again to demonstrate that they are returned.  I think a test in `tests/server/api/test_flow_runs.py` that creates a flow run with a `POST` request, then retrieves it again with a `GET` request would do the trick.

src/prefect/agent.py
```diff
@@ -425,6 +425,16 @@ async def get_infrastructure(self, flow_run: FlowRun) -> Infrastructure:
             # once we reach the end, set the value
             data[nested_fields[-1]] = value
 
+        # FlowRun's infra_overrides should take precedence over Deployment's
```
usefulalgorithm, line 428: I think this should be the only thing to change in order to make `FlowRun`'s `infra_overrides` override `Deployment`'s, right?

src/prefect/server/api/deployments.py
```diff
@@ -407,13 +407,18 @@ async def create_flow_run_from_deployment(
         parameters = deployment.parameters
         parameters.update(flow_run.parameters or {})
 
+        infra_overrides = deployment.infra_overrides
```
usefulalgorithm, line 410: I guess I don't need to do this...

src/prefect/server/api/deployments.py
```diff
@@ -407,13 +407,18 @@ async def create_flow_run_from_deployment(
...
+            infra_overrides[k] = v
+
         # hydrate the input model into a full flow run / state model
         flow_run = schemas.core.FlowRun(
             **flow_run.dict(
                 exclude={
                     "parameters",
                     "tags",
                     "infrastructure_document_id",
+                    "infra_overrides",
```
usefulalgorithm, line 421: ... and this? Since in agent it actually resolves the 2 dicts

usefulalgorithm:
bump

usefulalgorithm:
Hi @zangell44 , sorry for the noise but can you take a look at this one when you have a moment? I created an [issue](https://github.com/PrefectHQ/prefect/issues/9893) just like the contribution guide suggested, but it's been stuck at triage for about 2 weeks now. No one seemed to have reviewed the issue and the functionality proposed in it.

If the feature I'm trying to implement is not what the team would consider, or if there's any other plan for this, please also let me know.

Thank you!

ianccy:
@chrisguidry SOS

usefulalgorithm:
thanks!

yuwaj6:
Hi @chrisguidry,
It's been a while since we heard back from you. 
Let us know if there's something we can do to prioritize this pull request.

Thanks!

usefulalgorithm:
no worries, thanks for the response!

Issue #9923, opened 2023-06-13
"Database is locked" error when creating and destroying work pools quickly


jawnsy:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When running Prefect server, creating and destroying work pools in rapid succession leads to a 500 response with a "database is locked" error message and traceback in the console of the system running `prefect server start`

### Reproduction

```python3
I can reproduce this by creating deleting work pools using our Terraform provider, but this is using the POST and DELETE methods of the work_pools REST API, so should be simpler to reproduce that way
```

### Error

```python3
$ PREFECT_API_URL=http://localhost:4200/api prefect server start

 ___ ___ ___ ___ ___ ___ _____ 
| _ \ _ \ __| __| __/ __|_   _| 
|  _/   / _|| _|| _| (__  | |  
|_| |_|_\___|_| |___\___| |_|  

Configure Prefect to communicate with the server with:

    prefect config set PREFECT_API_URL=http://127.0.0.1:4200/api

View the API reference documentation at http://127.0.0.1:4200/docs

Check out the dashboard at http://127.0.0.1:4200



09:47:30.220 | ERROR   | prefect.server - Encountered exception in request:
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect2-dev/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect2-dev/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect2-dev/lib/python3.10/site-packages/sqlalchemy/dialects/sqlite/aiosqlite.py", line 108, in execute
    self._adapt_connection._handle_exception(error)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/prefect2-dev/lib/python3.10/site-packages/sqlalchemy/dialects/sqlite/aiosqlite.py", line 236, in _handle_exception
    raise error
[...]

PR #9938, opened 2023-06-14
make more sqlite locking retryable

jakekaplan:
closes: https://github.com/PrefectHQ/prefect/issues/9923

SQLite transaction behavior was improved in https://github.com/PrefectHQ/prefect/pull/9594. A consequence of this is that in order to guarantee the expected behavior, the database must be locked for certain operations. In order to respect this without erroring, the API will send a 503 to tell the client to retry later if the database is locked.

The original pr covered the main locked cases, where a `SQLITE_BUSY` or `SQLITE_BUSY_SNAPSHOT` was being raised by sqlite and attached as the original exception when `sqlalchemy` would raise `sqlalchemy.exc.OperationalError`.

There appears more cases involving concurrent access where `sqlite3` will raise `sqlite3.OperationalError("database is locked")` with no attached original exception. I could not determine the root cause of this but it seems like `sqlite3` is only optionally passing the original sqlite error names back up.

This pr adds this "generic" locked case to the function that determines a retry-able exception. 

### Example
```
from prefect.client.schemas.actions import WorkPoolCreate
import asyncio

async def main():
        await asyncio.gather(*[client.create_work_pool(WorkPoolCreate(name=f"work-pool-{i}")) for i in range(20)])
        work_pools = await client.read_work_pools()
        await asyncio.gather(*[client.delete_work_pool(work_pool.name) for work_pool in work_pools])

if __name__ == '__main__':
    asyncio.run(main())
```

### Pre change:
```
prefect.exceptions.PrefectHTTPStatusError: Server error '500 Internal Server Error' for url 'http://ephemeral-prefect/api/work_pools/work-pool-11'
Response: {'exception_message': 'Internal Server Error'}
For more information check: https://httpstatuses.com/500
```
### Post change:
```
09:40:41.602 | DEBUG   | prefect.client - Received response with retryable status code 503. Another attempt will be made in 1.7550290977850425s. This is attempt 1/6.
```

### Checklist

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -87,6 +87,8 @@ API_ROUTERS = (
     api.root.router,
 )
 
+SQLITE_LOCKED_MSG = "database is locked"
+
 
 class SPAStaticFiles(StaticFiles):
     """
@@ -157,7 +159,7 @@ def is_client_retryable_exception(exc: Exception):
         if getattr(exc.orig, "sqlite_errorname", None) in {
             "SQLITE_BUSY",
             "SQLITE_BUSY_SNAPSHOT",
-        }:
+        } or SQLITE_LOCKED_MSG in getattr(exc.orig, "args", []):
             return True
         else:
             # Avoid falling through to the generic `DBAPIError` case below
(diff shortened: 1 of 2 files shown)


bunchesofdonald (approved):
I like it!

PR #9941, opened 2023-06-14
Upgrade docker image to use Debian 12 bookworm

neumann-nico:
### Example

Upgrades from Debian 11 to Debian 12 for the node image.
The python docker image was already changed this night to use Debian 12 instead of 11.

e.g. [3.10.12-slim-bookworm, 3.10-slim-bookworm, 3.10.12-slim, 3.10-slim](https://github.com/docker-library/python/blob/5d908b0477c003712550c84ac4d133367b912f78/3.10/slim-bookworm/Dockerfile) are all the same.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ARG NODE_VERSION=16.15
 ARG EXTRA_PIP_PACKAGES=""
 
 # Build the UI distributable.
-FROM node:${NODE_VERSION}-bullseye-slim as ui-builder
+FROM node:${NODE_VERSION}-bookworm-slim as ui-builder
 
 WORKDIR /opt/ui
 


jawnsy:
Hey there,

Thanks so much for your contribution! As you saw, we pin our images to the `latest` tag that the official Docker Library images upstream of us publish. We prefer to do track the upstream `latest` tags because those correspond to what Docker deems to be a reasonable default version to use, and we do not have a compelling reason (i.e. based on security needs or other requirements) to accept upgrades earlier than that.

Once the upstream images are updated, we'll pull in changes automatically, as you saw with the Python base image. Users are always welcome to create their own images if they want to customize the base image. Therefore, I'm closing this out.

PR #9975, opened 2023-06-16
[WIP] Tutorial: flows and tasks

EmilRex:
WIP changes to flows and tasks section of tutorials

### DIFF
--- a/docs/tutorial/flows.md
+++ b/docs/tutorial/flows.md
@@ -4,9 +4,10 @@ tags:
     - tutorial
     - getting started
     - basics
-    - tasks
     - flows
-    - subflows
+    - logging
+    - parameters
+    - retries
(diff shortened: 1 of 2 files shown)


Sahiler (approved):
I like the changes, thank you for adding in a lot more content into the logging + parameters portion. Only caveat is you use **Let's** a lot, and unsure if the tense should be that vs **You**

EmilRex:
> I like the changes + the refreshed flow example, thank you for adding in a lot more content into the logging + parameters portion. Only caveat is you use **Let's** a lot, and unsure if the tense should be that vs **You**

❤️ Thank you! Yeah, the actual writing definitely needs some improvement. Going to merge for now to avoid conflicts and then will keep at it.

Issue #9773, opened 2023-05-30
Prefect CLI fails with prefect.exceptions.PrefectHTTPStatusError: Redirect response '307 Temporary Redirect'


NicholasFiorentini:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

I'm running Prefect CLI behind a corporate network with deep packet inspection software. This intercept the DNS between my machine and Prefect server, and due to missing support for `HTTP 307 Temporary Redirect` status, the Prefect client fails.

Prefect CLI should expose a parameter to enable `follow_redirects=True` for `httpx`. Reference: https://www.python-httpx.org/compatibility/#redirects

### Reproduction

```python3
It could be tested mocking httpx to reply the HTTP 307 status. Or, locally, with Python's HTTP server module (https://docs.python.org/3/library/http.server.html) to simulate the HTTP 307.
```

### Error

```python3
10:24:25.610 | DEBUG   | prefect.client - Connecting to API at https://prefect-api.test.***.***/api/
Traceback (most recent call last):
  File "py-scaffolding-py3.11/lib/python3.11/site-packages/prefect/cli/_utilities.py", line 46, in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "py-scaffolding-py3.11/lib/python3.11/site-packages/prefect/utilities/asyncutils.py", line 260, in coroutine_wrapper
    return call()
           ^^^^^^
  File "py-scaffolding-py3.11/lib/python3.11/site-packages/prefect/_internal/concurrency/calls.py", line 245, in __call__
    return self.result()
           ^^^^^^^^^^^^^
  File "py-scaffolding-py3.11/lib/python3.11/site-packages/prefect/_internal/concurrency/calls.py", line 173, in result
    return self.future.result(timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
[...]

PR #9988, opened 2023-06-18
Issue 9773: Default HTTPX  follow_redirects to true

dianaclarke:
This is a first pass at the following issue:

closes https://github.com/PrefectHQ/prefect/issues/9773

This was the suggested course of action (see comment from @zanieb):
- [x] "Set this to `True` by default"
- [x] "Set this to `False` during test runs to ensure that we are not making calls to our API that require extra routing"
- [x] "I'm not sure if we need a dedicated `PREFECT_CLIENT_FOLLOW_REDIRECTS` setting. I'd rather not if feasible."
-- https://github.com/PrefectHQ/prefect/issues/9773#issuecomment-1568779492

**PS. I'm not familiar with the prefect codebase, just grabbed a few random issues tagged with `good first issue`, so review with extreme caution :)**

### Checklist

### DIFF
--- a/src/prefect/client/cloud.py
+++ b/src/prefect/client/cloud.py
@@ -11,7 +11,11 @@ import prefect.settings
 from prefect.client.base import PrefectHttpxClient
 from prefect.client.schemas import Workspace
 from prefect.exceptions import PrefectException
-from prefect.settings import PREFECT_API_KEY, PREFECT_CLOUD_API_URL
+from prefect.settings import (
+    PREFECT_API_KEY,
+    PREFECT_CLOUD_API_URL,
+    PREFECT_UNIT_TEST_MODE,
+)
 
 
 def get_cloud_client(
@@ -57,7 +61,8 @@ class CloudClient:
         httpx_settings["headers"].setdefault("Authorization", f"Bearer {api_key}")
 
         httpx_settings.setdefault("base_url", host)
-        httpx_settings.setdefault("follow_redirects", True)
+        if not PREFECT_UNIT_TEST_MODE.value():
+            httpx_settings.setdefault("follow_redirects", True)
         self._client = PrefectHttpxClient(**httpx_settings)
 
     async def api_healthcheck(self):
--- a/src/prefect/client/cloud.py
+++ b/src/prefect/client/cloud.py
@@ -57,6 +57,7 @@ class CloudClient:
         httpx_settings["headers"].setdefault("Authorization", f"Bearer {api_key}")
 
         httpx_settings.setdefault("base_url", host)
+        httpx_settings.setdefault("follow_redirects", True)
         self._client = PrefectHttpxClient(**httpx_settings)
 
     async def api_healthcheck(self):
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -503,7 +503,7 @@ PREFECT_TEST_MODE = Setting(
     default=False,
 )
 """If `True`, places the API in test mode. This may modify
-behavior to faciliate testing. Defaults to `False`.
+behavior to facilitate testing. Defaults to `False`.
 """
 
 PREFECT_TEST_SETTING = Setting(
@@ -512,7 +512,7 @@ PREFECT_TEST_SETTING = Setting(
     value_callback=only_return_value_in_test_mode,
 )
 """
-This variable only exists to faciliate testing of settings.
+This variable only exists to facilitate testing of settings.
 If accessed when `PREFECT_TEST_MODE` is not set, `None` is returned.
 """
 
@@ -864,7 +864,7 @@ PREFECT_ASYNC_FETCH_STATE_RESULT = Setting(bool, default=False)
 """
 Determines whether `State.result()` fetches results automatically or not.
 In Prefect 2.6.0, the `State.result()` method was updated to be async
-to faciliate automatic retrieval of results from storage which means when 
+to facilitate automatic retrieval of results from storage which means when
 writing async code you must `await` the call. For backwards compatibility, 
 the result is not retrieved by default for async users. You may opt into this
 per call by passing  `fetch=True` or toggle this setting to change the behavior
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -510,7 +510,10 @@ PREFECT_UNIT_TEST_MODE = Setting(
     bool,
     default=False,
 )
-"""If `True`, code is executing in a unit test context. Defaults to `False`."""
+"""
+This variable only exists to facilitate unit testing. If `True`,
+code is executing in a unit test context. Defaults to `False`.
+"""
 
 PREFECT_TEST_SETTING = Setting(
     Any,
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -506,6 +506,12 @@ PREFECT_TEST_MODE = Setting(
 behavior to facilitate testing. Defaults to `False`.
 """
 
+PREFECT_UNIT_TEST_MODE = Setting(
+    bool,
+    default=False,
+)
+"""If `True`, code is executing in a unit test context. Defaults to `False`."""
+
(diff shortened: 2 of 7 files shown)


tests/client/test_base_client.py
```diff
@@ -469,7 +506,39 @@ async def test_prefect_httpx_client_raises_prefect_http_status_error(
         base_client_send.return_value = RESPONSE_400
         with pytest.raises(PrefectHTTPStatusError) as exc:
             await client.post(url="fake.url/fake/route", data={"evenmorefake": "data"})
-            assert (
-                "Response: {'extra_info': [{'message': 'a test error message'}]}"
-                in str(exc)
-            )
+        expected = "Response: {'extra_info': [{'message': 'a test error message'}]}"
```
dianaclarke, line 509: This is a bug fix for an existing test (the assert needs to be outside`with pytest.raises`, otherwise it's not actually executed). To verify bug, change the expected value to "XXX", move it back inside `with pytest.raises`, and note that the test does not fail.

tests/client/test_base_client.py
```diff
@@ -39,6 +39,43 @@ def disable_jitter():
         yield
 
 
+def redirects(request: httpx.Request) -> httpx.Response:
```
dianaclarke, line 42: This is mock transport data used in 2 tests below.

tests/client/test_base_client.py
```diff
@@ -469,7 +506,39 @@ async def test_prefect_httpx_client_raises_prefect_http_status_error(
...
+            ["https://example.org/redirect_308", "308 Permanent Redirect"],
+        ],
+    )
+    async def test_prefect_httpx_client_follow_redirect_false(self, mock_redirects):
+        mock_transport = httpx.MockTransport(redirects)
+        client = PrefectHttpxClient(follow_redirects=False, transport=mock_transport)
+        original_url, redirect_response = mock_redirects
+        with pytest.raises(PrefectHTTPStatusError) as exc:
+            await client.get(original_url)
+        expected = f"Redirect response '{redirect_response}' for url '{original_url}'"
```
dianaclarke, line 543: This asserts the behaviour the user reported with respect to redirects:
- `prefect.exceptions.PrefectHTTPStatusError: Redirect response '307 Temporary Redirect' for url 'https://prefect-api.test.***/api/flows/'`

tests/client/test_base_client.py
```diff
@@ -469,7 +506,39 @@ async def test_prefect_httpx_client_raises_prefect_http_status_error(
...
+            ["https://example.org/redirect_307", "https://example.org/redirected_307"],
+            ["https://example.org/redirect_308", "https://example.org/redirected_308"],
+        ],
+    )
+    async def test_prefect_httpx_client_follow_redirect_true(self, mock_redirects):
+        mock_transport = httpx.MockTransport(redirects)
+        client = PrefectHttpxClient(follow_redirects=True, transport=mock_transport)
+        original_url, redirected_url = mock_redirects
+        response = await client.get(original_url)
+        assert response.url == redirected_url
```
dianaclarke, line 526: This asserts the desired behaviour of the issue the user reported.

src/prefect/client/cloud.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         httpx_settings["headers"].setdefault("Authorization", f"Bearer {api_key}")
 
         httpx_settings.setdefault("base_url", host)
+        httpx_settings.setdefault("follow_redirects", True)
```
dianaclarke, line 60: Should both the `CloudClient` and `PrefectHttpxClient` have `follow_redirects` default to `true`?

bunchesofdonald (changes requested):
> 1. By "Set this to `False` during test runs to ensure that we are not making calls to our API that require extra routing", do you mean the unit tests in this repo or some other tests? 

This would be for tests in this repo. The callout is to ensure that our clients aren't being redirected during the test suite. If the test suite is encountering redirects this is a 'code smell' because we're likely hitting the wrong URL or the prefect server is doing something incorrectly. So to ensure that we're catching those issues we would want it to _not_ follow redirects by default when running tests. 

We set a setting called `PREFECT_TEST_MODE` during tests that you might be able to use to set `follow_redirects` to make this happen. Then in the tests you have that are testing that you can use `temporary_settings` to temporarily change the `PREFECT_TEST_MODE` setting and initialize new clients while that setting is false to ensure they're getting `follow_redirects=True`


> 2. Is there documentation that needs to be updated somewhere?

I don't believe this requires any documentation updates.

tests/client/test_base_client.py
dianaclarke: >  think I'd revise this test a bit and just check that when initializing PrefectClient or CloudClient that follows_redirects is set to True by the code in Prefect and trust that httpx is doing the right thing.

Nods, 💯  Those new tests, are already in:
- `test_prefect_client_follow_redirects`
- `test_cloud_client_follow_redirects`

I worried an integration test was desired and since I had trouble coming up with an elegant way to mock `transport` in the prefect/cloud clients (but could easily for `PrefectHttpxClient`), I wrote tests that asserted the contract for `PrefectHttpxClient` and then tests that asserted the contract for the `PrefectClient` + `CloudClient`, which like lego blocks asserts that they should play nice together.

These `PrefectHttpxClient` tests helped me reproduce the issue and gain confidence in the proposed solution. That said, if I were more familiar with this codebase I would have just gone with these 2 new simple tests, but I wasn't sure that would fly with you all :)
- `test_prefect_client_follow_redirects`
- `test_cloud_client_follow_redirects`

I'll revert these `PrefectHttpxClient` tests.

Thanks SO much for the prompt review!!

tests/client/test_base_client.py
dianaclarke: Cool, I removed this test. I was just being extra cautious b/c I don't know this codebase and wasn't sure what the testing bar was, thanks!

tests/client/test_base_client.py
bunchesofdonald: I totally understand, it's a fuzzy line sometimes on what's an integration test and what's a unit test and guessing what any given repository wants is always hard.

src/prefect/settings.py
```diff
@@ -503,16 +503,22 @@ def default_cloud_ui_url(settings, value):
     default=False,
 )
 """If `True`, places the API in test mode. This may modify
-behavior to faciliate testing. Defaults to `False`.
+behavior to facilitate testing. Defaults to `False`.
 """
 
+PREFECT_UNIT_TEST_MODE = Setting(
```
dianaclarke, line 509: I first tried using the existing `PREFECT_TEST_MODE` and `PREFECT_TEST_SETTING` (chained with `PREFECT_TEST_MODE`), but if I've understood correctly, `PREFECT_TEST_MODE` is set to True when you stand up a test server, and if that test server is setup behind a corporate network with deep packet inspection software which intercept the DNS (from original issue report), then that test server won't follow redirects.

Hope that makes sense – let me know if not :)

src/prefect/settings.py
```diff
@@ -503,7 +503,16 @@ def default_cloud_ui_url(settings, value):
     default=False,
 )
 """If `True`, places the API in test mode. This may modify
-behavior to faciliate testing. Defaults to `False`.
+behavior to facilitate testing. Defaults to `False`.
```
dianaclarke, line 506: just spelling fix

src/prefect/settings.py
bunchesofdonald: Make sense to me, and it's nice to have a dedicated setting for unit tests.

bunchesofdonald (approved):
I think this looks great. Looks like we have an unrelated flaky test in the suite, but once these tests pass I'll get this merged. Thanks for the contribution @dianaclarke!

zanieb:
@bunchesofdonald minor note that it's more than a code smell. I believe redirects require a full second request to be send so we're effectively doubling traffic.

Issue #9939, opened 2023-06-14
Flow run submission failure by a Prefect Worker should result in `Crashed` state


chrisguidry:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Following from #9523, we should make the same adjustment for Workers, where if a flow run can't be submitted to the infrastructure, then it would be regarded as `Crashed` rather than `Failed`.

### Reproduction

```python3
n/a
```

### Versions

```Text
n/a
```

### Additional context

See #9523 for the original context, and #9691 for the change to the agent.

bunchesofdonald:
This was fixed by https://github.com/PrefectHQ/prefect/pull/9691

PR #9989, opened 2023-06-18
Issues 9939: Flow run submission failure should result in `Crashed` state

dianaclarke:
This is a first pass at the following issue:

closes https://github.com/PrefectHQ/prefect/issues/9939

This was the suggested course of action (see report from @chrisguidry):
- [x] "Following from https://github.com/PrefectHQ/prefect/issues/9523, we should make the same adjustment for Workers, where if a flow run can't be submitted to the infrastructure, then it would be regarded as Crashed rather than Failed."
-- https://github.com/PrefectHQ/prefect/issues/9939

**PS. I'm not familiar with the prefect codebase, just grabbed a few random issues tagged with `good first issue`, so review with extreme caution :)**

### Checklist

### DIFF
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -448,11 +448,13 @@ class PrefectAgent:
         if ready_to_submit:
             try:
                 infrastructure = await self.get_infrastructure(flow_run)
-            except Exception as exc:
+            except Exception:
                 self.logger.exception(
                     f"Failed to get infrastructure for flow run '{flow_run.id}'."
                 )
-                await self._propose_failed_state(flow_run, exc)
+                await self._propose_crashed_state(
+                    flow_run, "Flow run could not be submitted to infrastructure"
+                )
                 if self.limiter:
                     self.limiter.release_on_behalf_of(flow_run.id)
             else:
--- a/src/prefect/agent.py
+++ b/src/prefect/agent.py
@@ -37,7 +37,7 @@ from prefect.exceptions import (
 from prefect.infrastructure import Infrastructure, InfrastructureResult, Process
 from prefect.logging import get_logger
 from prefect.settings import PREFECT_AGENT_PREFETCH_SECONDS
-from prefect.states import Crashed, Pending, StateType, exception_to_failed_state
+from prefect.states import Crashed, Pending, StateType
 
 
 class PrefectAgent:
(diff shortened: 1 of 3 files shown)


src/prefect/agent.py
```diff
@@ -571,23 +573,6 @@ async def _propose_pending_state(self, flow_run: FlowRun) -> bool:
 
         return True
 
-    async def _propose_failed_state(self, flow_run: FlowRun, exc: Exception) -> None:
```
dianaclarke, line 574: Dead code (no callers)?

src/prefect/workers/base.py
```diff
@@ -939,24 +939,6 @@ async def _propose_pending_state(self, flow_run: "FlowRun") -> bool:
 
         return True
 
-    async def _propose_failed_state(self, flow_run: "FlowRun", exc: Exception) -> None:
```
dianaclarke, line 942: Dead code (no callers)?

src/prefect/agent.py
```diff
@@ -37,7 +37,7 @@
 from prefect.infrastructure import Infrastructure, InfrastructureResult, Process
 from prefect.logging import get_logger
 from prefect.settings import PREFECT_AGENT_PREFETCH_SECONDS
-from prefect.states import Crashed, Pending, StateType, exception_to_failed_state
```
dianaclarke, line 40: If there are no longer code paths that transition to a failed status, then what about all these?

Screen Shot 2023-06-18 at 9 32 27 AM

PS. I'm not familiar with the prefect codebase, just grabbed a few random tagged with `good first issue`, so review with extreme caution :)

dianaclarke: If I've understood correctly, failed state is now reserved for failed validation (and maybe timeouts)?

For example, like this:

```
    state = Pending()
    if flow.should_validate_parameters:
        try:
            parameters = flow.validate_parameters(parameters)
        except Exception:
            state = await exception_to_failed_state(
                message="Validation of flow parameters failed with error:"
            )
```

But are we confident that all of the above `await exception_to_failed_state` statements should remain as-is, or should some now become `exception_to_crashed_state`?

(totally new to this codebase, so apologies if this is obvious to everyone else)

bunchesofdonald:
Hi @dianaclarke, thanks so much for contributing a couple fixes, learning new codebases is hard and we appreciate the effort!

It actually looks like issue #9939 was fixed by https://github.com/PrefectHQ/prefect/pull/9691 but the issue wasn't closed out when the PR was closed. I've closed out that issue and I'll close out this pull request as well.

PR #10056, opened 2023-06-23
docs: rename 'Example' to 'Examples' in code blocks

themattmorris:
I was reading the docs and noticed some example sections rendering like this:

image

for docstrings that use doctest format.  I noticed a mixture of markdown and doctest format, but I only altered the existing occurrence of doctest format with 'Example' instead of 'Examples'.  I only noticed this because I've done the same thing on some of my projects that use `mkdocstrings`.

### Checklist

### DIFF
--- a/src/prefect/_internal/schemas/transformations.py
+++ b/src/prefect/_internal/schemas/transformations.py
@@ -33,7 +33,7 @@ def copy_model_fields(model_class: Type[B]) -> Type[B]:
     Use this decorator and the corresponding `FieldFrom` to compose response and
     action schemas from other classes.
 
-    Example:
+    Examples:
 
         >>> from pydantic import BaseModel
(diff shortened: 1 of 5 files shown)


bunchesofdonald:
Hi @themattmorris, thank you for the contribution! _I'm_ not that well versed in `mkdocstrings`, does changing `Example` to `Examples` have more of an effect than just changing the word in the header?

themattmorris:
Hi @bunchesofdonald, thanks for the response!  Yes this very small change does cause examples section of the docstrings to render as intended.  I'm not sure how to use the CI docs preview, so I built these locally on my PR branch, and this is what they look like:

image

image

(I suppose of the 5 changes, these were the only 2 that were actually published in the API docs).

bunchesofdonald (approved):
In that case this looks good to me, thank you for the explanation @themattmorris!

PR #10064, opened 2023-06-25
Update index.md

gabcoyne:
Add link to flows at bottom of index

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -49,3 +49,6 @@ Just bring your Python code, sprinkle in a few decorators, and go!
 ### Reference Material
 If you've never used Prefect before, check out the [core concepts docs](/concepts/).
 Many of these concepts will be introduced in the tutorial, however, the concept docs provide a more comprehensive overview of each concept.
+
+## [Next: Flows](/tutorial/flows/)
+


docs/tutorial/index.md
```diff
@@ -49,3 +49,6 @@ Just bring your Python code, sprinkle in a few decorators, and go!
 ### Reference Material
 If you've never used Prefect before, check out the [core concepts docs](/concepts/).
 Many of these concepts will be introduced in the tutorial, however, the concept docs provide a more comprehensive overview of each concept.
+
+## [Next: Flows](/tutorial/flows/)
```
jawnsy, line 53: fyi, using a header like this will confuse accessibility tools like screen readers. It also creates a confusing situation for the table of contents (right of panel in screenshot):

image

That's a general issue we'll have to fix across our other docs as well

jawnsy (approved):

docs/tutorial/index.md
billpalombi, line 53:
```suggestion
## [First Step: Create a Flow](/tutorial/flows/)
```

billpalombi: @gabcoyne what do you think of this framing?

billpalombi: Thanks for the heads up. I don't find  including this header in the table of contents too confusing, but I could see how it might be.

discdiver:
I'm closing this PR, as it appears this change made its way into the docs.

Issue #10066, opened 2023-06-25
`explode_variadic_parameter` throws KeyError exception if task accepts `**kwargs` but no keyword args are passed when calling `.map`


NodeJSmith:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

If a task has a variadic keyword parameter (e.g. `**kwargs`) and the task is mapped over, a KeyError exception will be thrown if there are no keyword arguments provided in the call to `map`. This is specific to map, due to the use of the function `explode_variadic_parameter` introduced in PR #8188. If running the tasks without map, the use of `**kwargs` without keyword arguments does not cause an exception.

This can be fixed by adding a default value to the `pop` method called on the `new_parameters` variable in `explode_variadic_parameter`. I've tested this on my local machine and this has worked as expected. If I have time I will submit a PR with the fix as well.

Current:
```python
    new_parameters = parameters.copy()
    for key, value in new_parameters.pop(variadic_key).items():
        new_parameters[key] = value
```

Working:
```python
    new_parameters = parameters.copy()
    for key, value in new_parameters.pop(variadic_key,{}).items():
        new_parameters[key] = value
```

### Reproduction

```python3
from prefect import flow, task


@task
def function_with_kwargs(random_value: int, **kwargs):
    print(random_value)
    print(kwargs)


@flow
def working_flow():
    _ = function_with_kwargs.map(random_value=list(range(5)), test_val="val")

@flow
def failing_flow():
    _ = function_with_kwargs.map(random_value=list(range(5)))


if __name__ == "__main__":
    working_flow()
    failing_flow()
```

### Error

```python3
13:08:09.662 | ERROR   | Flow run 'dark-robin' - Encountered exception during execution:
Traceback (most recent call last):
  File "/home/jessica/source/orion/Prefect-2/venv/lib/python3.9/site-packages/prefect/engine.py", line 833, in orchestrate_flow_run
    result = await flow_call.aresult()
  File "/home/jessica/source/orion/Prefect-2/venv/lib/python3.9/site-packages/prefect/_internal/concurrency/calls.py", line 292, in aresult
    return await asyncio.wrap_future(self.future)
[...]

PR #10067, opened 2023-06-25
Add default option to new_parameters.pop in `explode_variadic_parameter` used to handle `**kwargs` in task mapping

NodeJSmith:
Closes https://github.com/PrefectHQ/prefect/issues/10066

### Example
Adds a default value to fix the issue of KeyError exception being raised if a mapped task accepts `**kwargs` but does not use them.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -92,7 +92,7 @@ def explode_variadic_parameter(
         return parameters
 
     new_parameters = parameters.copy()
-    for key, value in new_parameters.pop(variadic_key).items():
+    for key, value in new_parameters.pop(variadic_key, {}).items():
         new_parameters[key] = value
 
     return new_parameters
--- a/src/prefect/utilities/callables.py
+++ b/src/prefect/utilities/callables.py
@@ -131,6 +131,10 @@ def collapse_variadic_parameters(
             "but parameters were given that are not present in the signature."
         )
 
+    if variadic_key and not missing_parameters:
+        # variadic key is present but no missing parameters, return parameters unchanged
+        return parameters
+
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@bunchesofdonald Sure thing, I'll try to get to it within the next few days.

NodeJSmith:
@bunchesofdonald - Glad I added the tests, there is actually a decision to be made here. While `explode_variadic_parameter` is simple enough, `collapse_variadic_parameter` has logic that can be implemented in one of two ways.

Before this PR, `collapse_variadic_parameter` would add an empty kwargs dictionary to the parameters if the signature had a variadic parameter - regardless if there was anything to provide to the variadic param.

```python
    def test_params_unchanged_if_variadic_parameter_and_no_kwargs_provided(self):
        def foo(a, b, **kwargs):
            pass

        parameters = {"a": 1, "b": 2}
        new_params = callables.collapse_variadic_parameters(foo, parameters)

        assert new_params == {"a": 1, "b": 2,"kwargs": {}}
```

With this PR, `collapse_variadic_parameter` will return the parameters unchanged if there is a variadic param but no missing parameters.

```python
    def test_params_unchanged_if_variadic_parameter_and_no_kwargs_provided(self):
        def foo(a, b, **kwargs):
            pass

        parameters = {"a": 1, "b": 2}
        new_params = callables.collapse_variadic_parameters(foo, parameters)

        assert new_params == parameters
```

I am not sure if matters enough to make this change, so if you would like me to revert it so that it returns the empty `kwargs` parameter, let me know. With all else aside though, I think it makes more sense to return the parameters unchanged if there is no need to change them.

PR #10079, opened 2023-06-26
Adds troubleshooting guide to docs

EmilRex:
This PR adds a new guide that covers common Prefect troubleshooting practices. These represent issues that come up often in conversations with users. The aim is not to be comprehensive, but rather to touch on the most common issues and give us a reference point to share with users. Assuming the PR is accepted, I imagine we'll want to build on this guide over time.

### DIFF
--- a/docs/guides/debugging.md
+++ b/docs/guides/debugging.md
@@ -1,0 +1,88 @@
+---
+description: Learn the basics of debugging Prefect.
+tags:
+    - debugging
+    - guides
+    - how to
+---
+
+# Debugging
+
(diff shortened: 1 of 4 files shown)


billpalombi (commented):
Thank you for adding this! I can imagine that we'll share this link a lot.

To me, "debugging" implies a specific case of getting code to work, but this guide includes diagnosing and resolving other kinds of issues, like authorization. I think "Troubleshooting" would be a more appropriate title. What do you think?

EmilRex:
> Thank you for adding this! I can imagine that we'll share this link a lot.
> 
> To me, "debugging" implies a specific case of getting code to work, but this guide includes diagnosing and resolving other kinds of issues, like authorization. I think "Troubleshooting" would be a more appropriate title. What do you think?

I love it! Changed the wording and committed all your suggestions. Thank you for the thorough review!

billpalombi (approved):
Lets go!

PR #10091, opened 2023-06-28
pass client down to recursive call of resolve block documents

jakekaplan:
If client is provided pass it to recursive call of `resolve_block_document_references`. This is already done for the other recursive call on the line below https://github.com/PrefectHQ/prefect/blob/main/src/prefect/utilities/templating.py#L193-L201


desertaxle (approved):
Thanks @jakekaplan!

Issue #10114, opened 2023-06-29
`prefect deployment ls` errors if a deployment's associated flow has been deleted


discdiver:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

`prefect deployment ls` errors if a deployment's associated flow has been deleted. Ideally that case would be handled gracefully.

### Reproduction

```python3
Create a deployment.

Delete the associated flow.
```

### Error

```python3
Traceback (most recent call last):
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/cli/_utilities.py", line 41, in wrapper
    return fn(*args, **kwargs)
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/utilities/asyncutils.py", line 255, in coroutine_wrapper
    return call()
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 383, in __call__
    return self.result()
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 283, in result
    return self.future.result(timeout=timeout)
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 169, in result
    return self.__get_result()
  File "/Users/jeffhale/miniforge3/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 346, in _run_async
    result = await coro
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/cli/deployment.py", line 460, in ls
    for deployment in sorted(
  File "/Users/jeffhale/miniforge3/lib/python3.10/site-packages/prefect/cli/deployment.py", line 449, in sort_by_name_keys
    return flows[d.flow_id].name, d.name
KeyError: UUID('ca4aa552-16d2-4dcd-bb0c-e3f8fac0483d')
An exception occurred.
```
```


### Versions

```Text
Version:             2.10.18
API version:         0.8.4
Python version:      3.10.8
Git commit:          cf177852
[...]

PR #10122, opened 2023-06-30
Fix deployment list CLI issue with deleted flow

rito-sixt:
### Checklist

attempts to reproduce and fix https://github.com/PrefectHQ/prefect/issues/10114 

For documentation changes:


rito-sixt:
Hi. I am taking a look at this. Could someone please assign this to me? 

Unfortunately, I am unable to reproduce this issue with a unit test. In this test ( https://github.com/rito-sixt/prefect/blob/b738aa3ee3e67e8b6c853777c1f9977515e89ee9/tests/cli/deployment/test_deployment_ls.py#L35 ), this should actually be failing with the exception, but instead passes with an empty list as output. Am I doing something wrong? Would be grateful for directional pointers to reproduce the same!

serinamarie:
Hi @rito-sixt, thanks for contributing! In future, it's possible we might have a discussion like this on the issue itself rather before opening a PR. At least, I will move this into draft mode. 

> Hi. I am taking a look at this. Could someone please assign this to me?

I am looking into this, I think there's a permissions issue right now with assigning issues to those outside the organization. 
> 
> Unfortunately, I am unable to reproduce this issue with a unit test. In this test ( https://github.com/rito-sixt/prefect/blob/b738aa3ee3e67e8b6c853777c1f9977515e89ee9/tests/cli/deployment/test_deployment_ls.py#L35 ), this should actually be failing with the mentioned exception, but instead passes with an empty list as output. Am I doing something wrong?
> 
> Would be grateful for directional pointers to reproduce the same!

I think the answer as to why the deployment is no longer being returned when you delete the flow is within these lines of code:

https://github.com/PrefectHQ/prefect/blob/f32b9c35e8ae8af5d74754663fb2d613a44c4d3d/src/prefect/cli/deployment.py#L437-L446

rito-sixt:
Hi @serinamarie, Thanks for taking a look. Completely aligned on having a discussion first to align on the solution before opening a PR in the future, and thanks for moving it to draft!

I will try to dig deeper into the code segment that you pointed towards, to see if I can reproduce the exception! Thanks a lot. 

For the solution itself, would we want to still fail the list command with a more detailed exception, or should we show the remaining deployments with a additional warning about missing flows?

serinamarie:
> I will try to dig deeper into the code segment that you pointed towards, to see if I can reproduce the exception! Thanks a lot.

I think a good place to start is playing with the `FlowFilter` in terms of what deployments get returned. In the test you have, no deployments end up returned right now. And that is what is used to get the flows (`d.flow_id`). So the question is, why don't any deployments get returned with your test even if they exist.

> For the solution itself, would we want to still fail the list command with a more descriptive exception, or should we show the remaining deployments with a additional warning about missing flows?

As the real issue is that, until a user does one of 2 things:
1. create a new deployment with that associated flow (which will cause the flow object to exist again)
or
2. delete the deployments associated with that flow

they will not be able to successfully run `prefect deployment ls`. I don't think we need to warn them but we should display it properly. I think ideally we would trim off the flow name like we do in the UI:
Screenshot 2023-06-30 at 1 26 59 PM



But I'm curious if @billpalombi has any thoughts on a different direction it should take.

rito-sixt:
> > I will try to dig deeper into the code segment that you pointed towards, to see if I can reproduce the exception! Thanks a lot.
> 
> I think a good place to start is playing with the `FlowFilter` in terms of what deployments get returned. In the test you have, no deployments end up returned right now. And that is what is used to get the flows (`d.flow_id`). So the question is, why don't any deployments get returned with your test even if they exist.
> 
> > For the solution itself, would we want to still fail the list command with a more descriptive exception, or should we show the remaining deployments with a additional warning about missing flows?
> 
> As the real issue is that, until a user does one of 2 things:
> 
> 1. create a new deployment with that associated flow (which will cause the flow object to exist again)
>    or
> 2. delete the deployments associated with that flow
> 
> they will not be able to successfully run `prefect deployment ls`. I don't think we need to warn them but we should display it properly. I think ideally we would trim off the flow name like we do in the UI: Screenshot 2023-06-30 at 1 26 59 PM
> 
> But I'm curious if @billpalombi has any thoughts on a different direction it should take.

Thanks for the pointers @serinamarie ! 

I was trying some stuff on similar lines and here are a few observations - 

1. I tried another deployment with a different flow_id, to ensure that other deployments should at least be rendered successfully. Unfortunately, neither deployment is showed 😢  
2. If I don't do the `models.flows.delete_flow` line, both the deployment(s) are returned successfully 🤔 
3. Tried debugging through the flow, and it seems that the `/deployments/filter` endpoint itself does not respond with any objects. So tried taking a closer look at the api implementation

The thing is that, since I do not pass any `flow_name` as a parameter to the `ls` command, I was expecting `FlowFilter` to be `None` and hence not interfere at all with the returned results. Probably, that might be where I am going wrong. I will take a closer look tomorrow! 🙌 

Thanks again for the collaboration, and hopefully I can make some progress tomorrow with a fresh(er) mind 😄

serinamarie:
> I was trying some stuff on similar lines and here are a few observations -
> 
> 1. I tried putting in another deployment with a different flow, to ensure that other deployments should at least be rendered successfully. Unfortunately, neither deployment is showed 😢

Hopefully it's not something to do with our `deployment` or `flow` fixture, but it would be surprising to me if it were.
> 2. If I don't do the `models.flows.delete_flow` line, both the deployment(s) are returned successfully 🤔

Yeah, I think figuring that out is key for getting the MRE/test, for sure 

> 3. Tried debugging through the flow, and it seems that the `/deployments/filter` endpoint itself does not respond with any objects. So tried taking a closer look at the api implementation
> 
> The thing is that, since I do not pass any `flow_name` as a parameter to the `ls` command, I was expecting `FlowFilter` to be `None` and hence not interfere at all with the returned results. Probably, that might be where I am going wrong. I will take a closer look tomorrow! 🙌

Commenting out the `FlowFilter` would probably return the expected `deployments`, which might help you debug the 0 exit code instead of 1. But I haven't looked too closely! 🍀

rito-sixt:
Hi @serinamarie ! I tried commenting out the `flow_filter` lines from both the deployment fetch and flow fetch, but still could not reproduce a failing test ( the filters were None anyway )! I debugged all the way into the server code as well, and verified that it does not return the deployment with the deleted flow while querying for deployments.

Could you please help me reproduce this failure? Would be a lot of help!

serinamarie:
Hi @rito-sixt, I missed your message and I will be OOO for a few weeks after today. Unfortunately, I won't have time tonight to try to look into the tests. However, regarding implementation, I think `collections.defaultdict` would be good for this use case since we've got a `KeyError`. It's a dictionary subclass that provides a default value for undefined keys. 

More info on `defaultdict`.. It takes two arguments: 
1. A factory function that creates a default value when a key isn't found in the dictionary. So when you try to access a key that doesn't exist in the dictionary, the factory function will be called to create a new object with an attribute `name` set to an empty string: `lambda: type("", (), {"name": ""}`.
2. a second dictionary, in this case the one already set to the `flows` variable:

```python
{
            flow.id: flow
            for flow in await client.read_flows(
                flow_filter=FlowFilter(id={"any_": [d.flow_id for d in deployments]})
            )
        }
```

So we can set `defaultdict()` equal to the `flows` variable, where the first argument to `defaultdict` is `lambda: type("", (), {"name": ""}` and the second argument is the above dict, if that makes sense.

The reason why we would call the attribute `name` is because we reference it later on when we add the row
```python
table.add_row(
            f"{flows[deployment.flow_id].name}/[bold]{deployment.name}[/]",
            str(deployment.id),
        )
```

I hope that's helpful, and feel free to try to ask someone else for more info while I'm out! Good luck @rito-sixt 🍀

PR #10124, opened 2023-06-30
Update Dockerfile rename language in `prefect deploy` CLI prompt

discdiver:
The 2.10.18 `prefect deploy` guided CLI deployment experience with a custom docker image and an existing file named Dockerfile goes like this:

> Prefect finds my dockerfile and asks if I want to use it. Cool. 

> I say no and it asks me if I want to rename it. I’m not sure why I might want to. I say sure. I’m assuming maybe otherwise it would get overridden with a new dockerfile.

> When I finish deployment creation the old dockerfile is renamed but the there is not a new dockerfile. As a user, I’m surprised.

This change explains why a rename might be desired and what is happening with a temporary Dockerfile.

### Example
New language in action (updated):

Screenshot 2023-07-01 at 6 47 59 AM

### Checklist

- [n/a ] This pull request references any related issue by including "closes `<link to issue>`"
  

For documentation changes:

- [ n/a] This pull request includes redirect settings in `netlify.toml` for files that are removed or renamed
-


src/prefect/cli/_prompts.py
```diff
@@ -378,8 +378,10 @@ async def prompt_build_custom_docker_image(
             build_step["dockerfile"] = "Dockerfile"
         else:
             if confirm(
-                "A Dockerfile exists but you chose not to use it. Would you like to"
-                " rename it?"
+                "A Dockerfile exists. You chose not to use it."
+                " A temporary Dockerfile will be created and will overwrite"
```
serinamarie, line 382: Thanks for whipping this up so quick!

One thing: It won't overwrite the existing Dockerfile, it will fail when the build step is run. We should also tell them it will be created during the build step so they don't expect it to be renamed right after this confirmation.

src/prefect/cli/_prompts.py
```diff
@@ -378,8 +378,10 @@ async def prompt_build_custom_docker_image(
             build_step["dockerfile"] = "Dockerfile"
         else:
             if confirm(
-                "A Dockerfile exists but you chose not to use it. Would you like to"
-                " rename it?"
+                "A Dockerfile exists. You chose not to use it. A temporary Dockerfile"
+                " will be created during the deployment build step. If a file named"
+                " 'Dockerfile' already exists at that time, the build step will fail."
+                " Would you like to rename the existing Dockerfile?"
```
serinamarie, line 384:
```suggestion
                "A Dockerfile exists. You chose not to use it. A temporary Dockerfile"
                " will be automatically built during the deployment build step. If another file"
                " named 'Dockerfile' already exists at that time, the build step will fail."
                " Would you like to rename your existing Dockerfile?"
```

src/prefect/cli/_prompts.py
discdiver: Thank you! Updated and added your commit.

tests/cli/test_deploy.py
```diff
@@ -4322,10 +4322,10 @@ async def test_prompt_build_custom_docker_image_accepted_use_existing_dockerfile
             expected_output_contains=[
                 "Would you like to build a custom Docker image",
                 "Would you like to use the Dockerfile in the current directory?",
-                "A Dockerfile exists but you chose not to use it.",
+                "A Dockerfile exists. You chose not to use it.",
                 (
-                    "A Dockerfile already exists. Please remove or rename the existing"
-                    " one."
+                    "A temporary Dockerfile will be automatically built during the"
+                    " deployment build step."
```
serinamarie, line 4328: This should not be changed. This is when a `ValueError` is raised.

serinamarie:
You have a change to mkdocs in this PR. Maybe you are out of date with main.

Also I would suggest fixing the test locally, running the tests locally to make sure they pass, then push

tests/cli/test_deploy.py
```diff
@@ -4322,10 +4322,10 @@ async def test_prompt_build_custom_docker_image_accepted_use_existing_dockerfile
             expected_output_contains=[
                 "Would you like to build a custom Docker image",
                 "Would you like to use the Dockerfile in the current directory?",
-                "A Dockerfile exists but you chose not to use it.",
+                "A Dockerfile exists. You chose not to use it.",
                 (
-                    "A Dockerfile already exists. Please remove or rename the existing"
-                    " one."
+                    "A temporary Dockerfile will be automatically built during the"
```
discdiver, line 4327:
```suggestion
                    "A Dockerfile already exists. Please remove or rename the existing"
```

tests/cli/test_deploy.py
discdiver, line 4328:
```suggestion
                    " one."
```

serinamarie (approved):
Thanks so much!

PR #10200, opened 2023-07-11
remove the duplicate listing

bsenst:
The first item and the last item in the list are the same. The duplicate can be removed for better readability.


billpalombi (approved):
Great catch! Thank you.

PR #9789, opened 2023-05-30
Create files and directories with user-scoped permissions

jawnsy:
When creating ~/.prefect and other files, use restrictive (user-only) permissions to improve privacy on shared systems.

### Checklist

- [ ] This pull request references any related issue by including "closes `<link to issue>`"
	- If no issue exists and your change is not a small fix, please [create an issue](https://github.com/PrefectHQ/prefect/issues/new/choose) first.
- [ ] This pull request includes tests or only affects documentation.
- [x] This pull request includes a label categorizing the change e.g. `fix`, `feature`, `enhancement`, `docs`.
  

---

I tested this by removing my `~/.prefect` directory and running `prefect server`:

```
$ ls -al ~ | grep "prefect"
drwx------    5 jawnsy staff  160 May 30 13:06 .prefect
drwxr-xr-x    8 jawnsy staff  256 May 27 14:07 .prefect.bak
-rw-------    1 jawnsy staff  31K May 17 10:38 .psql_history-prefect_cloud

$ ls -al ~/.prefect
total 836K
drwx------    4 jawnsy staff  128 May 30 13:06 .
drwxr-x---+ 165 jawnsy staff 5.2K May 30 13:06 ..
-rw-------    1 jawnsy staff   93 May 30 13:06 memo_store.toml
-rw-------    1 jawnsy staff 784K May 30 13:06 prefect.db
```

I also tested `prefect cloud login` would create `profiles.toml` with `0644` permissions:

```
$ ls -al ~/.prefect
total 840K
drwx------    5 jawnsy staff  160 May 30 13:07 .
drwxr-x---+ 165 jawnsy staff 5.2K May 30 13:06 ..
-rw-------    1 jawnsy staff   93 May 30 13:06 memo_store.toml
-rw-------    1 jawnsy staff 784K May 30 13:06 prefect.db
-rw-------    1 jawnsy staff  244 May 30 13:07 profiles.toml
```

I also tested that making these files broad works OK:

```
$ chmod 755 ~/.prefect/*
mode of '/Users/jawnsy/.prefect/memo_store.toml' changed from 0600 (rw-------) to 0755 (rwxr-xr-x)
mode of '/Users/jawnsy/.prefect/prefect.db' changed from 0600 (rw-------) to 0755 (rwxr-xr-x)
mode of '/Users/jawnsy/.prefect/profiles.toml' changed from 0600 (rw-------) to 0755 (rwxr-xr-x)

$ chmod 755 ~/.prefect
[...]

PR #10246, opened 2023-07-17
Dont create DB default during settings load

### LINKED ISSUE #9840: default_database_connection_url tries to create default database instead of using path specified in environment
### First check

### Bug summary

The changed code in default_database_connection_url attempts to create a new database using the default url even when the PREFECT_API_DATABASE_CONNECTION_URL environment variable specifies a different location

image

### Reproduction

```python3
specify a valid location for PREFECT_HOME and PREFECT_API_DATABASE_CONNECTION_URL and run in an environment where the default path would be unwritable.
```

### Error

```python3

File "/opt/venv/lib/python3.9/site-packages/prefect/cli/_utilities.py", line 46, in wrapper

return fn(*args, **kwargs)

File "/opt/venv/lib/python3.9/site-packages/prefect/cli/config.py", line 188, in view

default_settings = default_settings.with_obfuscated_secrets()

File "/opt/venv/lib/python3.9/site-packages/prefect/settings.py", line 1795, in with_obfuscated_secrets

update={

File "/opt/venv/lib/python3.9/site-packages/prefect/settings.py", line 1796, in <dictcomp>

setting.name: obfuscate(self.value_of(setting))

File "/opt/venv/lib/python3.9/site-packages/prefect/settings.py", line 1736, in value_of

value = setting.value_callback(self, value)

File "/opt/venv/lib/python3.9/site-packages/prefect/settings.py", line 390, in default_database_connection_url

new_default.touch(mode=0o600)

File "/usr/lib/python3.9/pathlib.py", line 1304, in touch

fd = self._raw_open(flags, mode)

File "/usr/lib/python3.9/pathlib.py", line 1116, in _raw_open

return self._accessor.open(self, flags, mode)

FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/postgresql/.prefect/prefect.db'

An exception occurred.
```

### Versions

```Text
Prefect version info

Version: 2.10.12

API version: 0.8.4

Python version: 3.9.2

Git commit: ec55abdf

Built: Thu, Jun 1, 2023 3:51 PM

OS/Arch: linux/x86_64

Profile: default

Server type: server
```

### Additional context

The configuration was working on Prefect 2.10.11

cicdw:
Right now we are creating / touching the default database file when _loading_ the current Prefect settings (the `touch` logic lived within the logic for determining the default value of `PREFECT_API_DATABASE_CONNECTION_URL`) which is not an appropriate place to create this file.

I tested locally and this file _is_ still created when this line is removed (I'm guessing by sqlite internals).

Closes #9840 

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -386,9 +386,6 @@ def default_database_connection_url(settings, value):
         if old_default.exists():
             return "sqlite+aiosqlite:///" + str(old_default)
 
-        # Create the new default database with 0600 permissions if it does not exist
-        new_default.touch(mode=0o600)
-
     # Otherwise, return the new default
     return "sqlite+aiosqlite:///" + str(new_default)
 
@@ -485,7 +482,7 @@ PREFECT_CLI_PROMPT = Setting(
     Optional[bool],
     default=None,
 )
-"""If `True`, use interactive prompts in CLI commands. If `False`, no interactive 
+"""If `True`, use interactive prompts in CLI commands. If `False`, no interactive
 prompts will be used. If `None`, the value will be dynamically determined based on
 the presence of an interactive-enabled terminal.
 """
@@ -529,7 +526,7 @@ PREFECT_API_TLS_INSECURE_SKIP_VERIFY = Setting(
     bool,
     default=False,
 )
-"""If `True`, disables SSL checking to allow insecure requests. 
+"""If `True`, disables SSL checking to allow insecure requests.
 This is recommended only during development, e.g. when using self-signed certificates.
 """
 
@@ -554,7 +551,7 @@ PREFECT_API_ENABLE_HTTP2 = Setting(bool, default=True)
 """
 If true, enable support for HTTP/2 for communicating with an API.
 
-If the API does not support HTTP/2, this will have no effect and connections will be 
+If the API does not support HTTP/2, this will have no effect and connections will be
 made via HTTP/1.1.
 """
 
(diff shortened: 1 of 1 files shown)


zanieb (commented):
Seems sensible :) @jawnsy did this in https://github.com/PrefectHQ/prefect/pull/9789 to avoid creating it with permissions that were not user-scoped. I feel like maybe it should happen during migrations instead?

cicdw:
Hey @zanieb!! Yea, migrations make a lot of sense for the place to put this if we do need to be responsible for it - doesn't SQLite create these files itself if they don't exist though?

zanieb:
SQLite will make the file with permissions such that any user can read it

```
$ python -c "import sqlite3; sqlite3.connect('example.sqlite')"
$ gstat -c '%a' example.sqlite
644
$ gstat -c '%a' ~/.prefect/prefect.db
644
```

If we create the file manually beforehand we can give it permissions of 600 so only the owner can read it.

jawnsy (approved):
I think my original approach was well-meaning but misguided, and the correct solution is indeed to create the database manually (or to chmod it if necessary) outside our code.

Additionally, because the containing directory `~/.prefect` is created correctly with 0700 permissions, I don't think files created inside will be readable by other users (regardless of permissions). I tested this quickly - with two users (test1, uid 1000 and test2, uid 1001), a directory `/test` owned by user `test1` and set to `0700` permissions, and a file `/test/test` owned by user `test1` and set to `0644` permissions, the user `test2` cannot read the file:

```
test1@3d67fa8a6171:/test$ ls -al
total 0
drwx------. 2 test1 test1 18 Jul 18 02:18 .
dr-xr-xr-x. 1 root  root  63 Jul 18 02:18 ..
-rw-r--r--. 1 test1 test1  0 Jul 18 02:18 test

test2@3d67fa8a6171:~$ cat /test/test1
cat: /test/test1: Permission denied
```

So I concur that removing the `touch` makes the most sense here. Thanks for fixing this!

Issue #9824, opened 2023-06-02
Pausing Flow Run causes flow to fail.


taylor-curran:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

When we ran a flow flow run from deployment, and tried to pause it via the UI/Python function `pause_flow_run`, we expected the following behavior:

1. According to the documentation, pausing flows from outside the flow process will cause the flow run to exit ones all running tasks completes; when resumed this flow run will be re-scheduled.

2. Our assumption is that, although the flow run will exit after pause, it should end in a `Paused` State (even though there are task runs with non-final states). 
 
What we **observed** is that the flow run will actually fail due to unfinished task-runs:

![image](https://github.com/PrefectHQ/prefect/assets/68969861/2028b9a4-0b66-4fe4-b236-770706f3847e)

3. To make things weirder, we noticed that this flow will also trigger its on_failure handlers, although in reality it is not really failed (just unfinished):
![image](https://github.com/PrefectHQ/prefect/assets/68969861/5ebb3679-8fa2-4657-b9d1-b3da187f0cac)

Based on the above observation, we believe it maybe makes more sense to mark this flow run as “Paused”, just like the in-process paused flow runs, so that we can build proper tooling to perform checks like “list all flow runs currently on pause” and “resume all paused flow runs.”

### Reproduction

```python3
import time

from prefect import task, flow
from prefect.deployments import Deployment


def on_cancellation(flow, flow_run, state):
    print('Called')


@task
def some_task():
    time.sleep(20)


@flow(on_cancellation=[on_cancellation])
def my_flow():
    t1 = some_task.submit()
    for i in range(5):
        t1 = some_task.submit(wait_for=[t1])


if __name__ == '__main__':
    deployment = Deployment.build_from_flow(
        flow=my_flow,
        name="example-deployment",
        version=1,
        work_queue_name="demo",
        work_pool_name="default-agent-pool",
    )
    deployment.apply()
```

### Versions

```Text
v2.10.11
```

PR #10249, opened 2023-07-17
To pause is not to fail

cicdw:
Currently there are numerous issues with pausing flow runs:
- if any task run is paused mid-run, the flow run gets incorrectly marked as failed
- returning a `Paused` state from a flow run causes an infinite execution loop that never has a terminating condition
- `Paused` states without timeouts set cause an error due to a log message that assumes a non-null `timeout`

This PR addresses all three of these issues.

Closes https://github.com/PrefectHQ/prefect/issues/9824

### Checklist

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -47,14 +47,14 @@ Client-side execution and orchestration of flows and tasks.
     We require that this is always within a flow.
 
     See `Task.__call__` and `Task.submit`
-    
+
 - **A synchronous function acts as an entrypoint to the engine.**
     Unlike flow calls, this _will not_ block until completion if `submit` was used.
 
     See `enter_task_run_engine`
 
 - **A future is created for the task call.**
-    Creation of the task run and submission to the task runner is scheduled as a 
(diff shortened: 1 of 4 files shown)


billpalombi:
From a product perspective, this is definitely the behavior we want. I defer to others on code review.

src/prefect/engine.py
```diff
@@ -907,7 +904,7 @@ async def orchestrate_flow_run(
                 extra={"send_to_api": False},
             )
 
-        if not state.is_final():
+        if not state.is_final() and not state.is_paused():
```
bunchesofdonald, line 907: Should we update the `is_final` method to know that `paused` isn't final?

bunchesofdonald (approved):
Looks good, just the one question.

src/prefect/engine.py
cicdw: `is_final` does return `False` for `Paused`:

```python3
from prefect.states import Paused

Paused().is_final() # False
```

I think the logic here just looks a little confusing at first glance

PR #10267, opened 2023-07-20
Build out a proof-of-concept concurrency context manager.

bunchesofdonald:
This creates the general use `concurrency` context manager. I'm calling this a `proof-of-concept` because I'm currently unable to write useful tests as the API portion of this only exists in Cloud and that API is still in flux. Tests will be added after the API portion is ported.

### Example

```python
import sys
import asyncio
import random

from prefect.concurrency import concurrency
from prefect.blocks.system import String


trees = ["🌲", "🌴", "🌳"]


@task
async def get_random_tree() -> str:
    with concurrency("trees"):
        await asyncio.sleep(1)
        return random.choice(trees)


@flow
async def happy(num_trees: int):
    futures = []
    for _ in range(num_trees):
        futures.append(await get_random_tree.submit())

    for future in futures:
        tree = await future.result()
        print("Here's a happy little tree:")
        print(tree)


if __name__ == "__main__":
    num_trees = int(sys.argv[1])
    asyncio.run(happy(num_trees=num_trees))

```

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -77,7 +77,11 @@ from prefect.client.schemas.objects import (
     WorkPool,
     WorkQueue,
 )
-from prefect.client.schemas.responses import DeploymentResponse, WorkerFlowRunResponse
+from prefect.client.schemas.responses import (
+    DeploymentResponse,
+    WorkerFlowRunResponse,
+    MinimalConcurrencyLimitResponse,
+)
 from prefect.client.schemas.schedules import SCHEDULE_TYPES
 from prefect.client.schemas.sorting import (
     ArtifactCollectionSort,
@@ -102,6 +106,7 @@ from prefect.settings import (
     PREFECT_UNIT_TEST_MODE,
 )
 from prefect.utilities.collections import AutoEnum
+from prefect.utilities.math import clamped_poisson_interval
 
 if TYPE_CHECKING:
     from prefect.flows import Flow as FlowObject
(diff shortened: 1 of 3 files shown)


src/prefect/client/orchestration.py
```diff
@@ -2526,6 +2531,48 @@ async def delete_resource_owned_automations(self, resource_id: str):
...
+        while True:
+            retry_seconds = None
+            try_count += 1
+            try:
+                response = await self._client.post(
+                    "/v2/concurrency_limits/increment",
+                    json={"names": names, "slots": slots},
+                )
+            except Exception as exc:
+                if exc.response.status_code == 423:
```
cicdw, line 2547: my hunch is that this sort of retry logic would be better defined within the context manager so that the client remains "just" a wrapper to the API - wdyt?

src/prefect/concurrency.py
```diff
@@ -0,0 +1,88 @@
...
+        f"prefect.concurrency-limit.{phase}",
+        resource=resource,
+        related=related,
+        follows=follows,
+    )
+    return event
+
+
+@contextmanager
+def concurrency(names: Union[str, List[str]], slots: int = 1):
```
cicdw, line 59: I think a more active name would be helpful here, something like `take_concurrency` or some other verb

cicdw (changes requested):
two design-related suggestions, will go back through again

src/prefect/client/orchestration.py
chrisguidry: Maybe we want two separate functions here, like a `try-and-fail-quickly` version and a `wait-until-i-can-proceed` version.  The `wait-until-i-can-proceed` version would call the `try-and-fail-quickly` in the retry/backoff loop, and the context manager would use the `wait-until-i-can-proceed` version.  That way most users can use the context manager for clarity, but advanced users can use either the `wait-until-i-can-proceed` version or the `try-and-fail-quickly` version to build their own synchronization primitives.

chrisguidry, line 2547: Smokin' hot use of the 423 status code :fire:

src/prefect/concurrency.py
chrisguidry: I think I agree with that. The `slots` argument here is saying how many slots you want to occupy for this code block (as opposed to saying that the block is limited to `N` concurrent slots).  Disambiguating that would be helpful.  My naive read of `with concurrency("api-calls", 5):` is a declarative way of saying "up to 5 API calls can happen at a time", but what it really means is "I'm 'bout to do 5 API calls".  Not totally sold on `take_` as the verb, but I feel ya.

chrisguidry (approved):
Looks solid after we work out those couple of design questions

src/prefect/client/orchestration.py
bunchesofdonald: That's a good point, and I like making it a couple of utility functions that users can build upon. I'll make it so.

src/prefect/concurrency.py
bunchesofdonald: I agree with both of you, `concurrency` isn't a great name. I start up a thread in the raft room to talk about the naming, going to be easier to work it out there than in GitHub.

bunchesofdonald:
@cicdw After discussion in the raft channel we decided to keep `concurrency` as the name but change the `slots` kwarg to `occupy` to denote that it's occupying a slots rather than it specifying the limit itself. I've made that change here, let me know what you think.

cicdw (approved):
LGTM

Issue #8442, opened 2023-02-07
Ability to pause task runs, similar to 1.0


jbnitorum:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently tasks are unable to be paused and flows cannot be paused from within a task.

### Describe the proposed behavior

The proposed behavior would be that a task could be moved to a Paused state using a pause_task_run function.  There would be a similar 'reschedule' argument as there is with pause_flow_run.  When the flow has finished all available task runs that aren't downstream of the paused task, infrastructure spins down (assuming reschedule=True).

When the task is resumed infrastructure spins up and flow execution continues.  Since the paused task never reached a terminal state (it was paused) and no result was produced, the task that was paused is re-run when resumed.

Flow remains in a Running state for the duration of the process.

This is all similar behavior to 1.0.

### Additional context

In 1.0 a task could be paused by using ``raise PAUSE`` but so far in 2.0 only flows can be paused.   Also a flow cannot be paused from within a task, only from within the code of a flow.  We have dozens of flows in 1.0 that utilize task pausing and its our single biggest hurdle in migrating to 2.0.  In 1.0 we could pause a task and then the flow remains in a running state but infrastructure (ECS Tasks in our case) deprovisioned when there were no tasks left to run and then provisioned a new task upon the task being resumed.  If we could match this use case in 2.0 it would make for a much smoother transition.

For context we have flows in 1.0 that have multiple tasks that call pauses.  Reducing the pause to the task level rather than just the flow level is helpful to us to determine on which task the flow was on when it paused...  i.e. was this the pause on task 1 or the pause on task 2.  In 2.0 if pausing is only at the flow level we need to employ some logic by looking at the states of all the tasks in the flow to try to triangulate where the pause occurred.

We also found the pausing of tasks useful in long running processes that were reliant on some piece of data coming from a remote source.  For example we had a flow that would submit a request to an outside service that would generate a file and then save it to an s3 bucket.  Sometimes that would be minutes, sometimes hours.  What we developed in 1.0 was a task that checked if the file was present, and if it wasn't it would pause.  We then had another flow that would run on an interval and unpause the first flow's task (and other flows within a similar structure) and it would check for the file again.  This would continue until the file was found or a timeout defined in the task was hit which would raise an exception and move the task to a failed state.

[...]

Issue #7989, opened 2022-12-27
Ability to reschedule a flow/task after specific amount of time (non-blocking)


kielosz:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

In Orion we could use the `pause_flow_run` utility with the `reschedule=True` option to pause a main flow and resume later manually. It seems it is not sufficient to support the use case where we want to reschedule a task/flow to run in a future and based on some external condition reschedule again or continue (i.e. wait for the condition to be met).

1. We don't have an option to reschedule and resume automatically after specific amount of time.
2. We cannot pause (non-blocking) subflows. So it becomes cumbersome to put all the rescheduling logic in a main flow.

We tried something like this in a main flow:

```
    while True:
        ready = isReady()
        if not ready:
            pause_flow_run(reschedule=True)
        else:
            break
```

where `isReady` is an invocation of a task, but it 1) requires the flow to be manually resumed, 2) it can only be used in a main flow, 3) it creates a new task run for each iteration and we are not sure if this is right from the architectural point of view and can it lead to some issues.
 

### Describe the proposed behavior

We don't have the clear vision how the feature should be implemented yet. It seems we would need an ability to pass the `scheduled_time` when invoking `pause_flow_run` to resume automatically after specific amount of time. We may also need an ability to use `pause_flow_run` utility in subflows or even at task level, so it could be easily used.

### Additional context

In Prefect v1 we were able to reschedule a particular task. We used that when waiting for ax external condition to be met, e.g. wait for a PR to be merged. Sample code:

```
pull = repo.get_pull(number)
if pull.state != "closed":
    start_time = datetime.utcnow() + timedelta(
        minutes=10
    )
    raise signals.PAUSE(start_time=start_time)
else:
    return "CLOSED"
```

That way we were able to orchestrate long-running workflows, where some parts of the flows should be run only after some external conditions are met.

[Discourse topic.](https://discourse.prefect.io/t/pausing-flows-tasks-for-a-specific-amount-of-time-non-blocking/2094)

madkinsz:
> We don't have an option to reschedule and resume automatically after specific amount of time.

[...]

PR #10269, opened 2023-07-20
Pause task runs

cicdw:
Currently you cannot pause task runs; this was originally designed this way because task runs cannot be individually resumed.  However, if a user explicitly puts a task run into a Paused state, the flow run should pause likewise using the same settings, and now the flow run can resume as intended.

This PR allows task runs to explicitly return `Paused` states, and the flow run will behave as expected.

Closes https://github.com/PrefectHQ/prefect/issues/8442

### Example

```python
from prefect.states import Paused

@task
def pause_sometimes(pause):
    import random
    if random.random() > 0.5:
        return Paused()

@flow 
def pause_flow():
    pause_sometimes()
    return 42
```

If you execute this using your favorite method, you'll find that half the time the task run will enter a Paused state, and likewise so will the flow run.  Previously, this code would have generated an infinite loop.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -832,8 +832,19 @@ async def orchestrate_flow_run(
                 waited_for_task_runs = await wait_for_task_runs_and_report_crashes(
                     flow_run_context.task_run_futures, client=client
                 )
-        except PausedRun:
+        except PausedRun as exc:
+            # could get raised either via utility or by returning Paused from a task run
+            # if a task run pauses, we set its state as the flow's state
+            # to preserve reschedule and timeout behavior
             paused_flow_run = await client.read_flow_run(flow_run.id)
+            if paused_flow_run.state.is_running():
+                state = await propose_state(
+                    client,
+                    state=exc.state,
+                    flow_run_id=flow_run.id,
+                )
+
+                return state
             paused_flow_run_state = paused_flow_run.state
             return paused_flow_run_state
         except CancelledError as exc:
@@ -1778,7 +1789,7 @@ async def orchestrate_task_run(
                     extra={"send_to_api": False},
                 )
 
-            if not state.is_final():
+            if not state.is_final() and not state.is_paused():
                 logger.info(
                     (
                         f"Received non-final state {state.name!r} when proposing final"
--- a/src/prefect/exceptions.py
+++ b/src/prefect/exceptions.py
@@ -90,6 +90,10 @@ class PausedRun(PrefectException):
     Raised when the result from a paused run is retrieved.
     """
 
+    def __init__(self, *args, state=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.state = state
+
 
 class UnfinishedRun(PrefectException):
     """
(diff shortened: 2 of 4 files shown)


taylor-curran:
I QAd with this script:
```python
from prefect.states import Paused
from prefect import runtime


@task
def pause_sometimes():
    import random
    pause = random.random()
    if pause  > 0.5:
        print("Pausing!!")
        return Paused()
    else:
        print("Not pausing this time")

@flow(persist_result=True, log_prints=True) 
def pause_flow_chance():
    pause_sometimes()
    return 42

if __name__ == "__main__":
    pause_flow_chance()

```

And the pausing behaved as expected, however (and this may be outside the scope of this ticket) the resume action did not behave as expected.

Resuming caused the flow to run forever. I tried this multiple times to see if the 50% chance would make a difference and every time i went from pause to resume the flow ran indefinitely.

Check out this flow run, you should have access:

https://app.prefect.cloud/account/12242a57-9f05-4bf5-8853-9bff595d4bab/workspace/e8244a0f-b8ed-44c7-8555-856f4c180334/flow-runs/flow-run/6316b4e8-56aa-45bc-9a5b-2b5f1380c57a
 
Aside from the resume - everything worked as expected.

cicdw:
@taylor-curran I think I know what's going on and will need to think about the best way to handle it; if you return
```python
Paused(reschedule=True)
```
from your task run I believe it will work as you expect.

taylor-curran:
> if you return
> 
> ```python
> Paused(reschedule=True)
> ```
> 
> from your task run I believe it will work as you expect.

```python
from prefect.states import Paused
from prefect import runtime


@task
def pause_sometimes():
    import random
    pause = random.random()
    if pause  > 0.5:
        print("Pausing!!")
        return Paused(reschedule=True)
    else:
        print("Not pausing this time")

@flow(persist_result=True, log_prints=True) 
def pause_flow_chance():
    pause_sometimes()
    return 42

if __name__ == "__main__":
    pause_flow_chance()
```

Can confirm this works now. ✅ 
https://app.prefect.cloud/account/12242a57-9f05-4bf5-8853-9bff595d4bab/workspace/e8244a0f-b8ed-44c7-8555-856f4c180334/flow-runs/flow-run/33c29897-9f4b-462a-a4af-6610fae6108d

bunchesofdonald (approved):

cicdw:
Going to merge to unblock any user who needs this functionality with the caveat that `reschedule=True` is important; I'll revisit improving the pause / resume semantics more holistically soon.

serinamarie:
@cicdw Do we want any docs detailing usage here, by chance, on par with [flow runs](https://docs.prefect.io/latest/concepts/flows/?h=pause#pause-a-flow-run)?

cicdw:
@serinamarie I don't think so, this is really just a workaround for now; I'd rather streamline the whole pause / resume user experience before documenting it

kielosz:
Hi, thank you very much for this! It's been a highly anticipated feature by some of us. Do you think that adding an option to provide a `scheduled_time` parameter so the flow run will be automatically resumed in the future would require significant effort? I mentioned this in my previous request [here](https://github.com/PrefectHQ/prefect/issues/7989), and I hope it can now be easily implemented. I imagine it could look like this:

```python
@task
def pause():
    scheduled_time = datetime.now() + timedelta(minutes=10)
    return Paused(
        reschedule=True,
        scheduled_time=scheduled_time
    )
```

or use the `Scheduled` state instead of `Paused`.

PR #10296, opened 2023-07-24
[WIP] First pass at supporting pydantic v2 installs alongside prefect

### LINKED ISSUE #10145: Support installation with Pydantic v2
### First check

### Prefect Version

2.x

We are currently using Pydantic v1 (and [we have it pinned](https://github.com/PrefectHQ/prefect/pull/10144) to prevent build failures).

Ideally, we would like to support installation with both v1 and v2, until v1 falls out of support. This would require us to understand the [forward compatibility features available in v1](https://github.com/pydantic/pydantic/issues/5165) and adopt those.

### Example Use

_No response_

### Additional context

_No response_

cicdw:
This PR allows for Prefect to be installed alongside Pydantic v2, but does _not_ migrate any Pydantic functionality to v2. 

My plan is to first get tests passing with a Pydantic v1 install, and then to update our requirements and make sure tests still pass with Pydantic v2 installed.  And of course, significant QA once tests are passing.

Closes #10145

### Notes

- [tests passing on Pydantic v1 on this commit](https://github.com/PrefectHQ/prefect/pull/10296/commits/a67bc6f032b78adff07d172d6788ded19b2da90f)

### Checklist


src/prefect/_internal/compatibility/pydantic/__init__.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from pydantic.v1 import *
```
jawnsy, line 2: Amazing! While you're in here, can you also revert the pin in https://github.com/PrefectHQ/prefect/blob/1d611266959de2f146dfc05a5e3e402a7e3cd1bb/requirements.txt#L27 (i.e. revert #10144)

Do you think we should have run our test suite with both Pydantic v1 and v2? How complete do we need our test matrix to be? (i.e. do we need Python [3.8, 3.9, 3.10, 3.11] x Pydantic [1, 2]) or can we take a shortcut to avoid doubling our test runtime again?

jawnsy: Ah, nevermind, just read your PR description!

cicdw: I'll revert the pin once I've got tests passing on v1 👍 

I personally think testing against v2 is sufficient, but open to pushback on that - our tests already take so long and have non-deterministic errors, so I don't want to compound that if we can avoid it

jawnsy: > I personally think testing against v2 is sufficient

I think that makes sense, and if we learn otherwise then we can always add a small number of v1 tests (i.e. test with the latest version of Python + Pydantic v1).

It's unlikely that future patch versions of v1 will break anything given that upstream is focused on v2

> our tests already take so long and have non-deterministic errors, so I don't want to compound that if we can avoid it

100% agreed!

cicdw:
I need to revisit this more holistically

nzojy:
@cicdw -- any sense for when this will be rolled out?

PR #10299, opened 2023-07-25
Implement `rate_limit` function to block execution while acquiring slots

bunchesofdonald:
This implements the `rate_limit` form of concurrency limits where there is no explicit decrement and a slot decay is used instead. Much like the `concurrency` context manager this PR includes no tests and will be 'codified' when the API is ported to OSS.

### Example
```python
import asyncio
import datetime

from prefect.concurrency import rate_limit


def resource_hog():
    rate_limit("hog")
    print(datetime.datetime.now())


async def main():
    await asyncio.gather(*[resource_hog() for _ in range(3)])


if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/src/prefect/concurrency.py
+++ b/src/prefect/concurrency.py
@@ -44,13 +44,16 @@ async def wait_for_successful_response(
 
 
 async def acquire_concurrency_slots(
-    names: List[str], slots: int
+    names: List[str],
+    slots: int,
+    mode: Union[Literal["concurrency"], Literal["rate_limit"]] = "concurrency",
 ) -> List[MinimalConcurrencyLimitResponse]:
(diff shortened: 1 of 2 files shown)


src/prefect/concurrency.py
```diff
@@ -128,6 +131,27 @@ def concurrency(names: Union[str, List[str]], occupy: int = 1):
             )
 
 
+def rate_limit(names: Union[str, List[str]], occupy: int = 1):
```
chrisguidry, line 134: `occupy` is a great verb!

chrisguidry (approved):
Nice!  I'll put those tests on your tab ;)

PR #10323, opened 2023-07-27
ACI push pool guide

zhen0:
Adds docs for adding an azure container instance push work pool.  I'm adding notes based on my experience setting up azure and would welcome a review/input from @prefectcboyd who I consider to be much more of an azure expert. It's likely that I go into a bit much detail on navigating the azure portal/UI. 

I run into an issue when serving the docs locally because my version of pymdown extensions doesn't like our addition of `pymdownx.tabbed`.  Taking out `pyndownx.tabbed`  from our config lets me serve the docs but the tabs we're using for the push work pools guide don't work. Equally, it seems our netlify build isn't working either.  So I'm leaving this as draft until we can properly check the deploy preview and make sure tabs are correctly displayed. 

### Example

 
 Example - but the tabs aren't working as I'd expect:
 
 
image

### Checklist

For documentation changes:

### DIFF
--- a/docs/guides/push-work-pools.md
+++ b/docs/guides/push-work-pools.md
@@ -15,7 +15,7 @@ Push [work pools](/concepts/work-pools/#work-pool-overview) are a special type o
 
 In this guide you will:
 
-- Create a push work pool that sends work to Google Cloud Run, AWS ECS or Azure ACI.
+- Create a push work pool that sends work to Google Cloud Run, AWS ECS or Azure ACI
 - Deploy a flow to that work pool
 - Execute our flow without having to run a worker or agent process to poll for flow runs
 
--- a/docs/guides/push-work-pools.md
+++ b/docs/guides/push-work-pools.md
@@ -5,6 +5,7 @@ tags:
     - deployments
     - Cloud Run
     - AWS ECS
+    - Azure Container Instance
 search:
   boost: 2
 ---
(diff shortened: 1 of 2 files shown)


docs/guides/push-work-pools.md
```diff
@@ -14,13 +15,14 @@ Push [work pools](/concepts/work-pools/#work-pool-overview) are a special type o
 
 In this guide you will:
 
-- Create a push work pool that sends work to Google Cloud Run or AWS ECS
+- Create a push work pool that sends work to Google Cloud Run, AWS ECS or Azure
```
prefectcboyd, line 18: Thoughts on Azure ACI - ? It keeps it in line with "Provider -> Offering" for the Google / AWS examples

docs/guides/push-work-pools.md
```diff
@@ -32,12 +34,35 @@ In this guide you will:
...
+    ### Create App Registration
+
+    1. In the Azure portal, create an app registration.
+    2. In the app registration, create a client secret. Copy the value and store it somewhere safe.
+    
+    ### Add App Registration to Subscription
+
+    1. Navigate to the resource group you created earlier.
+    2. Click on "Access control (IAM)" and then "Role assignments".
+    3. Search for the app registration and select it. Give it a role that has sufficient privileges to create, run, and delete ACI container groups.
```
prefectcboyd, line 62: They can either add contributor to a specific resource group (that they created) to limit the scope, or add a custom role / permissions with explicit `Microsoft.ContainerInstance/containerGroups/*` - still probably good to scope that to the resource group. 

I'm not sure how detailed we need to be here though for users - RBAC and permissions should fall to them?

prefectcboyd (commented):
Left a couple updates, otherwise looks good!

docs/guides/push-work-pools.md
zhen0: Yeah good call out

docs/guides/push-work-pools.md
zhen0: Yeah being too prescriptive on RBAC and permissions get difficult. I think I'll leave as is and we can update with more info if we get questions about it.

zhen0:
> Left a couple updates, otherwise looks good!

Thank you!

WillRaphaelson (approved):
thanks jenny!

PR #10329, opened 2023-07-28
Add support for predictive backoff to the concurrency / rate limiter.

bunchesofdonald:
This has two main changes:

1. If a failing response has a `Retry-After` header use that instead of an exponential backoff.
2. When decrementing slots send the amount of time that the slot was occupied to the `decrement` API.

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -2546,11 +2546,15 @@ class PrefectClient:
         )
 
     async def release_concurrency_slots(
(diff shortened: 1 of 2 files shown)


chrisguidry (approved):

bunchesofdonald:
Thanks for the review @chrisguidry! I rolled this into the codify pr so I should actually just close this out.

chrisguidry:
I thought it looked familiar :D

PR #10355, opened 2023-08-01
Allow JSON infra overrides via `prefect deploy`

abrookins:
Allow passing in infrastructure overrides as JSON objects via the `prefect deploy` command. This allows setting nested properties, which is necessary with e.g. k8s job specs.

### Example

The goal is to allow passing a JSON object as a variable to the `prefect deploy` CLI command and then having Prefect interpolate it correctly in the base job template for a work pool.

For this to work, you'll need an `object` type variable in your base job template. Consider the following additions to a kubernetes work pool's base job template, which includes an `object` type variable called `resources` and the usage of that variable to specify resources for a container:
```
{
  "variables": {
    "type": "object",
    "properties": {
      "resources": {
        "type": "object",
        "title": "Resources",
        "description": "Resource limits to use"
      }
  },
  "job_configuration": {
            "containers": [
              {
                "resources": "{{ resources }}",
              }
}
```

**With the change in this PR**, you can now pass in the entire `resources` object from the CLI, e.g.:
```
$ prefect deploy flow.py:flow -p my-pool -n my-deployment-name --variable {\"resources\":{\"limits\":{\"cpu\":4\}\}\}
```

When the Prefect worker creates a Job, it will contain the resources variable (all other fields elided for the example):
```
apiVersion: batch/v1
kind: Job
spec:
  template:
    spec:
      containers:
        resources:
          limits:
            cpu: "4"
```

### Checklist

For documentation changes:


tests/cli/test_deploy.py
```diff
@@ -4878,6 +4878,38 @@ async def test_work_pool_image_already_exists_not_updated_after_adding_build_doc
...
+        )
+        assert deployment.name == "test-name"
+        assert deployment.work_pool_name == "test-pool"
+        assert deployment.version == "1.0.0"
+        assert deployment.tags == ["foo-bar"]
+        assert deployment.infra_overrides == {
+            "env": "prod",
+            "resources": {"limits": "cpu"},
+        }
+
```
serinamarie, line 4911: Nice! Wondering if it might be a good idea to throw in some test cases for `_extract_variable` specifically for raising a `TypeError` and `ValueError`

abrookins: Yeah, absolutely!

desertaxle (approved):
LGTM!

PR #10371, opened 2023-08-04
Update `prefect deploy` to skip building docker image prompt if `build` key explicitly set to null in `prefect.yaml`

serinamarie:
Skip build docker image prompt when a deployment has explicitly set the `build` step to `null` in `prefect.yaml`. See the attached issue for an example. Previously if `build` was set to `null`, users would still see prompts to build a docker image on every interactive `prefect deploy` run for that deployment. This change allows users to configure deployments without running a build step and without needing to opt out of the prompt every time.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/cli/deploy.py
+++ b/src/prefect/cli/deploy.py
@@ -451,7 +451,11 @@ async def _run_single_deploy(
 
     update_work_pool_image = False
 
-    if is_interactive() and not docker_build_step_exists:
+    build_step_set_to_null = (
+        True if "build" in deploy_config and deploy_config["build"] is None else False
+    )
+
+    if is_interactive() and not docker_build_step_exists and not build_step_set_to_null:
         work_pool = await client.read_work_pool(deploy_config["work_pool"]["name"])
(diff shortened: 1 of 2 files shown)


src/prefect/cli/deploy.py
```diff
@@ -451,7 +451,11 @@ async def _run_single_deploy(
 
     update_work_pool_image = False
 
-    if is_interactive() and not docker_build_step_exists:
+    build_step_set_to_null = (
+        True if "build" in deploy_config and deploy_config["build"] is None else False
+    )
```
bunchesofdonald, line 456: I think you can just assign the boolean here rather than having the if/else branching.

```suggestion
    build_step_set_to_null = "build" in deploy_config and deploy_config["build"] is None
```

bunchesofdonald (approved):

src/prefect/cli/deploy.py
serinamarie: ah yeah, thanks!

PR #10401, opened 2023-08-14
Highlight `@flow` decorator instead of function

billpalombi:
Highlight `@flow` decorator instead of function


discdiver (approved):
Good catch

PR #10408, opened 2023-08-14
Add 423 to retry codes and clean up docs

cicdw:
PR that shouldn't affect any current behavior; adds 423 to the list of automatically retried status codes, and puts documentation in more sensible locations.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/client/base.py
+++ b/src/prefect/client/base.py
@@ -154,10 +154,7 @@ class PrefectResponse(httpx.Response):
 class PrefectHttpxClient(httpx.AsyncClient):
     """
     A Prefect wrapper for the async httpx client with support for retry-after headers
-    for:
-
-    - 429 CloudFlare-style rate limiting
-    - 503 Service unavailable
+    for the provided status codes (typically 423, 429, 502 and 503).
 
     Additionally, this client will always call `raise_for_status` on responses.
 
@@ -246,11 +243,21 @@ class PrefectHttpxClient(httpx.AsyncClient):
         return response
 
     async def send(self, *args, **kwargs) -> Response:
+        """
+        Send a request with automatic retry behavior for the following status codes:
+
+        - 423 Resource locked
+        - 429 CloudFlare-style rate limiting
+        - 502 Bad Gateway
+        - 503 Service unavailable
+        """
+
         api_request = partial(super().send, *args, **kwargs)
 
         response = await self._send_with_retry(
             request=api_request,
             retry_codes={
+                status.HTTP_423_LOCKED,
                 status.HTTP_429_TOO_MANY_REQUESTS,
                 status.HTTP_503_SERVICE_UNAVAILABLE,
                 status.HTTP_502_BAD_GATEWAY,
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -582,7 +582,7 @@ PREFECT_CLIENT_RETRY_EXTRA_CODES = Setting(
 )
 """
 A comma-separated list of extra HTTP status codes to retry on. Defaults to an empty string.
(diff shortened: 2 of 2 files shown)


src/prefect/client/base.py
```diff
@@ -246,11 +243,21 @@ async def _send_with_retry(
...
+        - 502 Bad Gateway
+        - 503 Service unavailable
+        """
+
         api_request = partial(super().send, *args, **kwargs)
 
         response = await self._send_with_retry(
             request=api_request,
             retry_codes={
+                status.HTTP_423_LOCKED,
```
chrisguidry, line 260: @bunchesofdonald, will this play well with concurrency limits?  I know we're using 423 when there are no more slots left.  Since concurrency limits have their own wait/retry logic, we wouldn't want to have the underlying client retry too, right?

cicdw: Ah yea, let me know if that's a problem - we don't strictly need this new 423 behavior in our base client, I just implemented it since I was looking at our status code usage in the backend and this was more appropriate for one of our use cases

chrisguidry: I like the usage!  We just might need to adjust the concurrency limits implementation

bunchesofdonald: Yeah, it'll cause a problem. The issue I see is that the client retry logic only has a set number of retries, whereas the concurrency limits should retry indefinitely.

bunchesofdonald: I could change the concurrency implementation to use a `200` with status information in the body, that someone could add 423 to the retries and break concurrency seems like a 'bug'.

chrisguidry: I like 423 a lot for the concurrency case, though, it's very close to the heart of what it's about.  I'd rather see if we can have the client-side concurrency implementation just tell the HTTP client that it is handling 423s somehow.  It seems generally useful to allow for adjusting the retry statuses on a call-by-call basis if we need to.

Looking at the client, we [currently have this implementation](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/client/orchestration.py#L2540-L2546), which will use the default retry logic, trying several times before raising, which will _then_ get caught by the outer concurrency implementation for _its_ retry logic:

```python
    async def increment_concurrency_slots(
        self, names: List[str], slots: int, mode: str
    ) -> httpx.Response:
        return await self._client.post(
            "/v2/concurrency_limits/increment",
            json={"names": names, "slots": slots, "mode": mode},
        )
```

What if we had a way to adjust that on a call-by-call basis, like:

```python
    async def increment_concurrency_slots(
        self, names: List[str], slots: int, mode: str
    ) -> httpx.Response:
        async with self._suppress_retries_for(423):
            return await self._client.post(
                "/v2/concurrency_limits/increment",
                json={"names": names, "slots": slots, "mode": mode},
            )
```

It's a little messy but it would give us some super targeted behavior tweaks when we need it

bunchesofdonald: I could get behind something like that.

cicdw: Alternatively I could expose the retry codes interface to the `post` method of the client, and implement it such that 502 and 503 are always retried, and additional codes can be provided.  I can default those additional codes to 423 and 429 but if others are provided we respect that.

Also happy to simplify the PR and defer this decision if it's a distraction

chrisguidry: I was thinking about `kwargs` too, but I wasn't sure the best way to override the HTTP methods on the underlying client.  Honestly, we should probably migrate our retry implementation to a _transport_ class instead of subclassing the HTTP client itself, then I believe it would be more straightforward

chrisguidry (approved):

PR #10410, opened 2023-08-15
viz mode draft

jakekaplan:
map-flow

Example Usage:
```python
from prefect.testing.utilities import viz_mode

@task
def get_nums():
    return [1, 2, 3, 4, 5, 8]

@task
def print_nums(nums):
    for n in nums:
        print(n)

@task
def square_num(num):
    return num**2

@flow
def map_flow():
    nums = get_nums()
    squared_nums = square_num.map(nums)
    print_nums(squared_nums)


if __name__ == '__main__':
    with viz_mode():
        map_flow()
```

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -94,7 +94,7 @@ from functools import partial
 from typing import Any, Awaitable, Dict, Iterable, List, Optional, Set, TypeVar, Union
 from uuid import UUID, uuid4
 
-import graphviz
+
 import anyio
 import pendulum
 from anyio import start_blocking_portal
@@ -190,6 +190,7 @@ from prefect.utilities.callables import (
 from prefect.utilities.collections import StopVisiting, isiterable, visit_collection
 from prefect.utilities.pydantic import PartialModel
 from prefect.utilities.text import truncated_to
+from prefect.utilities.visualization import task_run_dependencies_graph
 
 R = TypeVar("R")
 EngineReturnType = Literal["future", "state", "result"]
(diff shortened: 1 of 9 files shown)


src/prefect/settings.py
```diff
@@ -495,6 +495,11 @@ def default_cloud_ui_url(settings, value):
 in CLI output. If `False`, output will not be wrapped. Defaults to `True`.
 """
 
+PREFECT_VIZ_MODE = Setting(
+    bool,
+    default=False,
```
jakekaplan, line 500: add description

src/prefect/engine.py
```diff
@@ -370,12 +372,31 @@ async def create_then_begin_flow_run(
...
+        res = await state.result(fetch=True)
     else:
         raise ValueError(f"Invalid return type for flow engine {return_type!r}.")
 
+    if PREFECT_VIZ_MODE:
+        task_runs = await client.graph(flow_run.id)
+
+        id_to_name = {task_run["id"]: task_run["name"] for task_run in task_runs}
+        edges = []
+        for task_run in task_runs:
```
jakekaplan, line 386: move this out

src/prefect/engine.py
```diff
@@ -370,12 +372,31 @@ async def create_then_begin_flow_run(
...
+            for dependency in task_run["upstream_dependencies"]:
+                if dependency["input_type"] == "task_run":
+                    name = id_to_name[dependency["id"]]
+                    edges.append((name, task_run["name"]))
+
+        g = graphviz.Digraph("G", filename=f"{flow.name}")
+        for edge in edges:
+            g.edge(*edge)
+
+        g.view()
```
serinamarie: I think .render() might provide more flexibility? We can specify a filename, format, etc.

e.g. 
```python
g.render('flow_graph.png', view=True)
```

src/prefect/engine.py
```diff
@@ -553,6 +558,14 @@ async def begin_flow_run(
     # during interpreter shutdown
     await APILogHandler.aflush()
 
+    # Generate a local visualization of the flow run if enabled
```
serinamarie, line 561: It may be a good idea to ensure that the implementation supports notebooks

src/prefect/settings.py
```diff
@@ -495,6 +495,11 @@ def default_cloud_ui_url(settings, value):
 in CLI output. If `False`, output will not be wrapped. Defaults to `True`.
 """
 
+PREFECT_VIZ_MODE = Setting(
```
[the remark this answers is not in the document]
jakekaplan: or on the flow...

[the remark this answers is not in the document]
jakekaplan: definitely a better way to pass this context and potentially other options

PR #10414, opened 2023-08-15
Codify concurrency context managers and rate limiting with tests

bunchesofdonald:
This does three main things:

1. Splits the sync `concurrency` context manager into a sync one and an async one.
2. Does the same for the rate limit function.
3. Brings the whole mess up to 100% line/branch coverage.

### Checklist

For documentation changes:


src/prefect/concurrency/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+    try:
+        yield
+    finally:
+        occupancy_seconds: float = (
+            pendulum.now("UTC") - acquisition_time
+        ).total_seconds()
+        _call_async_function_from_sync(
+            release_concurrency_slots, names, occupy, occupancy_seconds
+        )
+        _emit_concurrency_release_events(limits, occupy, emitted_events)
```
chrisguidry, line 123: I'm not crazy about the `aconcurrency` name.  I'm pretty sure we can unify these under a single name with something like this:


```python
class concurrency:
    def __init__(names: Union[str, List[str]], occupy: int = 1):
        self.names = names
        self.occupy = occupy

    def __enter__(self) -> Self:
        ...the current top part of `concurrency`...

    def __aenter__(self) -> Self:
        ...the current top part of `aconcurrency`...

    def __exit__(self, exc_type, exc_val, exc_tb) -> Self:
        ...the current bottom part of `concurrency`...

    def __aexit__(self, exc_type, exc_val, exc_tb) -> Self:
        ...the current bottom part of `aconcurrency`...
```

This way, a user can still say either:

```python
with concurrency('hello', occupy=2):
  ...
```

or

```python
async with concurrency('hello', occupy=2):
  ...
```

It's the exact same implementation you have now, where the sync and async paths are completely split, but unified under one API called `concurrency`

src/prefect/concurrency/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+    Args:
+        names: The names of the concurrency limits to acquire slots from.
+        occupy: The number of slots to acquire and holf from each limit.
+    """
+    names = names if isinstance(names, list) else [names]
+    limits = await acquire_concurrency_slots(names, occupy, mode="rate_limit")
+    _emit_concurrency_acquisition_events(limits, occupy)
+
+
+def rate_limit(names: Union[str, List[str]], occupy: int = 1):
```
chrisguidry, line 140: But now I see where your heartburn is: the rate limit implementation is not a context manager, so you still have a similar problem with the naming here. :(

chrisguidry (commented):
Nice work with the implementation and the test suite, it looks great!  I'm not crazy about the `a` methods, though, I would love some other ideas on that.  I think we can easily unify `concurrency`, but `rate_limit/arate_limit` are another beast entirely

src/prefect/concurrency/__init__.py
chrisguidry: My only other thought is to have them in separate parallel modules, but that's not amazing either:

```
from prefect.concurrency.sync import concurrency, rate_limit
...
from prefect.concurrency.async import concurrency, rate_limit
```

I guess we could steer savvy users toward:

```
from prefect.concurrency import aconcurrency as concurrency

...

async with concurrency(...):
   ...
```

bunchesofdonald: A lot of libraries I see that split into `sync` and `async` modules end up having an issue where the sync or async version gets ignored or the two implementations diverge in pretty severe ways. But I think I'd be okay with it here since `rate_limit` and `concurrency` are really just methods that orchestrate the API and under the hood use the same methods, just called in different ways.

src/prefect/concurrency/asyncio.py
```diff
@@ -0,0 +1,119 @@
...
+from prefect.client.schemas.responses import MinimalConcurrencyLimitResponse
+from prefect.utilities.math import bounded_poisson_interval, clamped_poisson_interval
+from prefect.settings import PREFECT_CLIENT_RETRY_JITTER_FACTOR
+
+
+async def wait_for_successful_response(
+    fn: Callable[..., Awaitable[httpx.Response]],
+    *args,
+    max_retry_seconds=30,
+    retryable_status_codes=[429, 502, 503],
```
zangell44, line 22: Nitpicking: we could run into issues with mutable default args right? I don't think it has any effect in practice since `retryable_status_codes` is never altered in the function

bunchesofdonald: Good point, it should be fine now and I don't see why we would ever alter the list in the function, but better to just not have the possibility.

tests/concurrency/test_concurrency_sync.py
```diff
@@ -0,0 +1,312 @@
...
+
+    my_flow()
+
+    assert executed
+
+
+async def test_concurrency_can_be_used_while_event_loop_is_running(
+    concurrency_limit: ConcurrencyLimitV2,
+):
+    # There are two ways that we'll try to call the async client functions,
```
zangell44, line 148: 👏

zangell44 (approved):
Looks great to me!

PR #10439, opened 2023-08-17
Hardcode docker workflow for a prod release

cicdw:
will be reverted

### DIFF
--- a/.github/workflows/docker-images.yaml
+++ b/.github/workflows/docker-images.yaml
@@ -10,7 +10,7 @@ on:
   push:
     tags:
       - '**'
-    branches: main
+        #    branches: main
   # On workflow_dispatch, push sha and branch patterns to prefect-dev
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-images.yaml
```diff
@@ -8,6 +8,8 @@ on:
     types: [released, prereleased]
   # On each commit merged into main, push sha and branch patterns to prefect-dev
   push:
+    tags:
+      - '**'
```
pleek91, line 12: Might not matter but I usually see this as `- '*'`

cicdw: Yea they are different but it won't matter for us; a single asterik prevents tags with slashes of the form `2.1/beta`

pleek91: oh never mind, learned something new https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

pleek91 (approved):

PR #10444, opened 2023-08-18
Add "How to Upgrade from Agents to Workers" to Concepts table

EmilRex:
Adds the "How to Upgrade from Agents to Workers" page to the Concepts table.

### DIFF
--- a/docs/concepts/index.md
+++ b/docs/concepts/index.md
@@ -30,6 +30,7 @@ search:
 
 |  Block and Agent-Based Deployments                 | Description   |
 | -------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


billpalombi (changes requested):
This page should effectively be a list of concepts, so the agents to workers transition guide isn't an appropriate item in that list. That said, I do think we should add a link to this guide at the top of the agents and infra blocks pages. I'll open that PR now.

billpalombi:
I recommend that we do #10445 instead

billpalombi:
> I opened the PR because the guide is in the side nav under "Concepts > Block and Agent-Based Deployments".

Totally makes sense, that was an oversight on my part.

PR #10450, opened 2023-08-18
Warn users with misconfigured `PREFECT_API_URL`

serinamarie:
When users misconfigure their `PREFECT_API_URL`,  we should try to let them know as soon as possible.

With this PR, when a `PREFECT_API_URL` is misconfigured, running profile CLI commands such as `prefect profile ls`, `prefect profile use <profile-name>`, and `prefect config set <profile-name>` will generate a warning.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/settings.py
+++ b/src/prefect/settings.py
@@ -369,48 +369,43 @@ def check_for_deprecated_cloud_url(settings, value):
     return deprecated_value or value
 
 
-def warn_on_misconfigured_api_url(settings, values):
+def warn_on_misconfigured_api_url(values):
     """
     Validator for settings warning if the API URL is misconfigured.
     """
-    api_url_value = PREFECT_API_URL.value_from(settings, bypass_callback=True)
-    if api_url_value is not None:
+    api_url = values["PREFECT_API_URL"]
+    if api_url is not None:
         misconfigured_mappings = {
             "app.prefect.cloud": (
-                "Warning: Your PREFECT_API_URL points to app.prefect.cloud. Did you"
-                " mean api.prefect.cloud?"
+                "`PREFECT_API_URL` points to `app.prefect.cloud`. Did you"
+                " mean `api.prefect.cloud`?"
             ),
             "account/": (
-                "Warning: Your PREFECT_API_URL uses */account/* but should use"
-                " */accounts/*"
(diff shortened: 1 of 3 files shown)


src/prefect/settings.py
```diff
@@ -368,6 +368,36 @@ def check_for_deprecated_cloud_url(settings, value):
     return deprecated_value or value
 
 
+def check_for_misconfigured_api_url(api_url):
```
serinamarie, line 371: I'm wondering if there's a more ideal place than `settings.py` for this

src/prefect/settings.py
```diff
@@ -368,6 +368,36 @@ def check_for_deprecated_cloud_url(settings, value):
...
+    misconfigured_mappings = {
+        "app.prefect.cloud": (
+            "Warning: Your PREFECT_API_URL points to app.prefect.cloud. Did you mean"
+            " api.prefect.cloud?"
```
EmilRex, line 375: I believe the `/api/` part is also a differentiating factor between the two URLs.
```suggestion
            " api.prefect.cloud/api?"
```

src/prefect/settings.py
EmilRex: I think `settings.py` is appropriate, but maybe implemented as a validator when calling `PREFECT_API_URL.value` since there are multiple ways of setting the variable, but only one way to retrieve it?

src/prefect/settings.py
serinamarie: I have another warning I'll push up in a separate commit but good call out

src/prefect/settings.py
serinamarie: Indeed, the validator will check the existing profile values, whereas during set() we don't look at current values but rather the ones attempting to be set.

src/prefect/settings.py
```diff
@@ -1773,6 +1815,7 @@ def post_root_validators(cls, values):
         #       in the future.
         values = max_log_size_smaller_than_batch_size(values)
         values = warn_on_database_password_value_without_usage(values)
+        values = warn_on_misconfigured_api_url(values)
```
EmilRex, line 1818: Nice!

EmilRex:
@serinamarie => absolute legend 🚀

PR #10422, opened 2023-08-16
Optimize the query generated by the /task_runs endpoint

mattklein:
We have observed poor performance of the Prefect Server in responding to the `/task_runs/filter` endpoint, and have seen mention in the Prefect Slack of other users reporting similar findings.

The code supporting `/task_runs/filter` is general enough to be able to allow filtering on lots of different criteria. For any filter criteria provided, the query generates a `WHERE EXISTS` clause based on that filter criteria.

But the main use case for this endpoint is to simply fetch the task runs for a particular `flow_run_id`. In our workload (and I suspect in many other users') this represents 99%+ of the invocations of this endpoint. However, the `WHERE EXISTS` DB query generated for this use case is much less efficient than it needs to be.

Specifically, when the endpoint is asked to simply fetch the task runs for a given `flow_run_id`, the query it generates looks like this:

```
SELECT
    task_run.task_key,
    task_run.dynamic_key,
    task_run.cache_key,
    task_run.cache_expiration,
    etc...
FROM task_run
WHERE EXISTS (
    SELECT
        flow_run.deployment_id,
        flow_run.work_queue_name,
        flow_run.flow_version,
        flow_run.parameters,
        flow_run.idempotency_key,
        flow_run.context,
        flow_run.empirical_policy,
        flow_run.tags,
        flow_run.created_by,
        flow_run.infrastructure_pid,
        flow_run.auto_scheduled,
        flow_run.flow_id,
        flow_run.infrastructure_document_id,
        flow_run.parent_task_run_id,
        flow_run.state_id,
        flow_run.work_queue_id,
        flow_run.name,
        flow_run.state_type,
        flow_run.state_name,
        flow_run.state_timestamp,
        flow_run.run_count,
        flow_run.expected_start_time,
        flow_run.next_scheduled_start_time,
        flow_run.start_time,
        flow_run.end_time,
        flow_run.total_run_time,
        flow_run.id,
        flow_run.created,
        flow_run.updated
	FROM flow_run
	WHERE flow_run.id = task_run.flow_run_id
    AND flow_run.id IN ('19f905e4-7ab7-4311-98c6-cce748e24605'::UUID) -- for example
) ORDER BY task_run.id DESC;
```

But a much more efficient -- and functionally equivalent -- way to generate that query is this:
```
SELECT
    task_run.task_key,
    task_run.dynamic_key,
    task_run.cache_key,
    task_run.cache_expiration,
    etc...
FROM task_run
[...]

PR #10466, opened 2023-08-22
Adding some safety checks to a prior optimization on task runs

chrisguidry:
We got an [awesome community contribution to optimize the task runs filter](https://github.com/PrefectHQ/prefect/pull/10422)
endpoint, but it needed a couple of safety guards:

* We should make sure that the ID filter is only making positive matches on
  ID (with `any_`) and not negative matches
* We should support multiple flow run IDs with an `IN` clause


chrisguidry:
@mattklein FYI a couple of extensions to your previous PR

mattklein:
> @mattklein FYI a couple of extensions to your previous PR

Thank you very much @chrisguidry!

zangell44 (approved):

PR #10498, opened 2023-08-24
Changes in the logic of the deploying work pools

AmanSal1:
In Prefect, there are two types of work pools: push and pull. In the push mode, no external worker is required. However, in the pull mode, an external worker is necessary for polling. Unfortunately, due to some missing logic, when a user deploys a push work pool, they encounter a CLI message.

 "To execute flow runs from this deployment, start a worker in a separate terminal that pulls work from the 'gcp-push-pool' work pool:

        $ prefect worker start --pool 'gcp-push-pool'" 

Currently, there is no logic to determine whether the work pool is a pull or push. As a result, this statement gets printed regardless of any conditions, as there is no logic associated with it. It applies universally to every type of work pool. Therefore, I have added a conditional statement to check whether the pool is a push or pull, and then display the message accordingly.

Fixes: #10300 

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/cli/deploy.py
+++ b/src/prefect/cli/deploy.py
@@ -629,11 +629,12 @@ async def _run_single_deploy(
                 ),
             )
 
+    work_pool_type = deploy_config.get('work_pool', {}).get('type')
     app.console.print(
         "\nTo execute flow runs from this deployment, start a worker in a"
         " separate terminal that pulls work from the"
         f" {deploy_config['work_pool']['name']!r} work pool:"
-    )
+    )if work_pool_type == 'pull' else None
     app.console.print(
         f"\n\t$ prefect worker start --pool {deploy_config['work_pool']['name']!r}",
         style="blue",


AmanSal1:
@desertaxle This is the issue I am was talking about can you review it ?

serinamarie:
Hi @AmanSal1, did/can you run the tests locally for `test_deploy.py`?

AmanSal1:
@serinamarie I haven't run it yet, but do you think this could be the reason for this bug?

serinamarie:
It seems to be the case! But to be sure, my process toward resolving an issue like this would go something like this :) 
1. Reproduce the issue
2. Make a fix on my branch
3. Make sure you can no longer reproduce the issue on your branch
4. Run the tests
5. Make a PR

AmanSal1:
@serinamarie I've been attempting to reproduce this bug, but I've hit a roadblock when running 'prefect deploy' as it's prompting me with a question. Specifically, I'm getting stuck when it asks for the Docker image build for deployment. At some juncture, it also requests the registry URL. Could you please provide guidance on what needs to be done at that particular step? Though I am pretty sure this is cause of the bug

serinamarie:
@AmanSal1, it looks like a change was put in that supersedes this PR and closes the linked issue. Thank you though and we look forward to your next contribution!

AmanSal1:
@serinamarie No issue I guess it required more changes other than what i have suggested .

PR #10544, opened 2023-08-29
Add the selected tab in the url

pleek91:
# Description
Updates all implementations of p-tabs to use a query param for the selected tab. This way tab state is preserved when refreshing the page and we could link to specific tabs if needed.

### DIFF
--- a/ui/src/pages/ConcurrencyLimit.vue
+++ b/ui/src/pages/ConcurrencyLimit.vue
@@ -4,7 +4,7 @@
       <PageHeadingConcurrencyLimit v-if="concurrencyLimit" :concurrency-limit="concurrencyLimit" @delete="deleteConcurrencyLimit" />
     </template>
 
-    <p-tabs :tabs="tabs">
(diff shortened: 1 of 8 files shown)


ui/src/pages/ConcurrencyLimit.vue
```diff
@@ -36,7 +36,8 @@
     { label: 'Details', hidden: media.xl },
     { label: 'Active Task Runs' },
   ])
-  const { tabs } = useTabs(computedTabs)
+  const tab = useRouteQueryParam('tab', 'Details')
```
znicholasbrown, line 39: What's the behavior like if the details tab is hidden on smaller screens at the outset?

znicholasbrown (approved):

ui/src/pages/ConcurrencyLimit.vue
znicholasbrown: That's what I was hoping - nice composition

Issue #9769, opened 2023-05-29
Prefect server fails to start in version 2.10.11 when using venv in Windows


adnanmig:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

```
- OS: Windows 11 Pro, Version 10.0.22000 Build 22000
- Python 3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)] on win32
- VSCode: Version: 1.78.2 (user setup)
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T14:39:26.248Z
Electron: 22.5.2
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.22000
Sandboxed: Yes
- Running under venv.
```

The prefect server fails to start with an error as shown below:

![image](https://github.com/PrefectHQ/prefect/assets/25726443/0351e770-6d6d-4b33-ab05-d41f5fba8ee0)

### Reproduction

```python3
1- Create a new venv in python: python -m venv .environment
2- Install prefect: pip install prefect
3- Run prefect server: prefect server start
```

### Error

```python3
'E:\Python' is not recognized as an internal or external command,
operable program or batch file.
Server stopped!
```

### Versions

```Text
2.10.11
```

### Additional context

Prefect version 2.10.10 works fine.

madkinsz:
See comment at https://github.com/PrefectHQ/prefect/issues/9290#issuecomment-1518097099 for details.

`sys.executable` needs to quoted on Windows.

yusuf-jkhan1:
Adding to this. I get the same error when running a prefect worker in windows. The worker starts but then when it tries to run a scheduled flow the worker logs throw that same error as above saying the first part of the windows path is not executable

PR #10547, opened 2023-08-29
get sys executable

jakekaplan:
closes: https://github.com/PrefectHQ/prefect/issues/9769

Windows seems to do better when `sys.executable` is quoted. See: https://github.com/PrefectHQ/prefect/issues/9769#issuecomment-1568786878 for more details.

This has been difficult to test, as the original issue doesn't seem to be reproducible but the the fix seems to work on windows.


desertaxle (approved):
LGTM!

PR #10566, opened 2023-08-30
Deployment page showing a blank tab on desktop

pleek91:
# Description
Fixes an issue where viewing a deployment on a larger screen would cause no tab to be displayed.


zhen0 (approved):
Thank you!

PR #10575, opened 2023-08-31
Kubernetes guide for deploying worker to Azure AKS

iotmani:
Closes https://github.com/PrefectHQ/platform/issues/4487 

Added section on how to get an API key, as prerequisites didn't assume it.
Corrected folder structure, the `prefect.yaml` references `flows/hello.py` from outside the folder.

Note that we have unpinned module versions to avoid, but they're prefect maintained ones so we should be fine.
I suggest removing the apostrophe from `Install your flow's dependencies`, as the terminal complains when copy-pasting the multi-line block (tested on zsh).

### Checklist

For documentation changes:


docs/guides/deployment/kubernetes.md
```diff
@@ -94,8 +94,36 @@ Let's start by creating a new cluster. If you already have one, skip ahead to th
...
+      az aks create --resource-group <RESOURCE-GROUP-NAME> --name <CLUSTER-NAME>
+
+      # Configure kubectl to connect to your Kubernetes cluster
+      az aks get-credentials --resource-group <RESOURCE-GROUP-NAME> --name <CLUSTER-NAME>
+
+      # Verify the connection by listing the cluster nodes
+      kubectl get nodes
+    ```
+
+    !!! Note "Note"
```
billpalombi, line 124: This feels more like a warning to me.

iotmani: Sure, worth noting I've seen Azure guides throw it in regardless, it won't do anything if you have an SSH key, but I debating with myself if it's worth just adding it, or add and justifying, or bring it up here.

iotmani: PS: The reason I had it as a Note is AKS will suggest it anyway, e.g.:

```
islam@islams-MacBook-Pro ~ % az aks create -g islam-docs-aks -n islam-prefect-cluster       
An RSA key file or key value must be supplied to SSH Key Value. You can use --generate-ssh-keys to let CLI generate one for you
```

billpalombi: I think its worth bringing up in a callout as you've done here. If you make it a warning instead of a note, it will be orange and include a caution icon, like this: 

Screenshot 2023-08-31 at 4 45 03 PM

billpalombi: Ah I see! In that case, you may not need it at all.

iotmani: Ok cool, I'll remove the 'note' altogether then.

Here's the Azure AKS quick-start where they don't even bring it up but just use it.
https://learn.microsoft.com/en-us/azure/aks/learn/quick-kubernetes-deploy-cli#connect-to-the-cluster

docs/guides/deployment/kubernetes.md
```diff
@@ -229,6 +269,11 @@ Give the work pool a name and save.
 
 Our new Kubernetes work pool should now appear in the list of work pools.
 
+## Create a Prefect Cloud API key
+
+While still in Prefect Cloud UI, create a Prefect Cloud API key if you don't already have one.
```
cicdw, line 274:
```suggestion
While in the Prefect Cloud UI, create a Prefect Cloud API key if you don't already have one.
```

cicdw (approved):
minor grammar fix, this looks good to me - thanks for bringing your expertise to the docs!

PR #10587, opened 2023-08-31
Reorganize guides

cicdw:
No content changes, just a reorganization into categories for our guides


billpalombi (approved):
Definitely better

PR #10597, opened 2023-09-01
Recovering from a failure to cleanly disconnect a websocket in the Events client

chrisguidry:
We received a surprising report from a customer that they were seeing the
`AssertionError` indicating that we were trying to `_emit` an event, but the
websocket was `None`.  It's not clear from the available information how this
might happen, but I could reproduce it by emulating a failure to cleanly
disconnect during a reconnection attempt.  This leaves the websocket as `None`,
which would trigger the assertion failure.

This changes `_emit` so that we can perform an initial reconnection attempt if
the client finds itself without a websocket.


desertaxle (approved):
LGTM!

bunchesofdonald (approved):
Thank you!

PR #10625, opened 2023-09-05
When dependabot makes a PR, treat it as an internal user

chrisguidry:
I noticed that a dependabot PR failed to run the test suite because our
"isinternal" check to see if the user is a PrefectHQ organization member was
including the `[bot]` suffix and causing an error from cURL.  This check is
used to decide whether the PR is run on our faster internal machines, which we
would want for dependabot, so we'll shortcut the API call and just treat them
as internal.

https://github.com/PrefectHQ/prefect/actions/runs/6078894220/job/16490683120?pr=10622

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -45,6 +45,12 @@ jobs:
       - name: Is internal team?
         id: isinternal
         run: |
+          # Treat dependabot as an internal user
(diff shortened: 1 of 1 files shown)


bunchesofdonald (approved):

chrisguidry:
The test failures on this branch aren't related to these changes, will investigate those separately

PR #10626, opened 2023-09-05
Accounting for a change in how `fsspec==2023.9.0` handles globs

chrisguidry:
In this new version of `fsspec`, there's been [some adjustments to how globs are
processed][1], which has the side-effect of returning the base directory when
doing a glob like `/some-dir/**` (this also includes `/some-dir`).

[1]: https://github.com/fsspec/filesystem_spec/commit/c3b4bc36a8c2fa1b98429f5cd963ea0b4d04ad11


serinamarie (approved):
Thanks!

PR #10671, opened 2023-09-08
Conditional task retries with `retry_condition_fn`

### LINKED ISSUE #7576: Add ability to conditionally retry tasks on Error class
### First check

### Prefect Version

2.x

When using the `@task` decorator to specify retries, the only retry behavior that can be configured is number of retries and the delay between them (using `retries` and `retry_delay_seconds`).

In Prefect 1.x, a `retry_on` parameter is available that allows you to limit task retries to specific types of exceptions. It would be great if this parameter was also available on the `@task` decorator in Prefect 2.x.

### Example Use

```python

@task(retries=3, retry_on={ValueError})
def my_task_that_occasionally_fails(my_num):
    if my_num % 2 == 0:
        raise ValueError()
    else:
        raise KeyError()

def my_flow(a_number):
    my_task_that_occasionally_fails.submit(a_number)

if __name__ == '__main__':
    my_flow(2)  # retries three times
    my_flow(1)  # does not retry
```

### Additional context

The PR that added this functionality to Prefect 1.x is located [here](https://github.com/PrefectHQ/prefect/pull/5634) and the original issue is [here](https://github.com/PrefectHQ/prefect/issues/5139).

dominictarro:
Closes #7576 

New `retry_condition_fn` parameter for `Task` lets the user to write custom logic that tells the task orchestrator to continue to the retry policy or not. When the `retry_condition_fn` function returns False, the server rejects the proposed retry. If True is returned, then the server is allowed to decide whether a retry is appropriate.

Only applies when task run ends in a `Failed` state.

### Example

```python
import httpx
from prefect import task
from httpx._exceptions import HTTPError

# Return False to prevent the orchestrator from retrying
def is_retriable(task, task_run, state):
    exc = state.result()
    if isinstance(exc, HTTPError):
        # Retry if not an authorization error
        return exc.args[0] != 401
    return True

@task(retry_condition_fn=is_retriable)
def my_api_call_task():
    r = httpx.get(...)
    r.raise_for_status()
    return r.json()
```

### Checklist

For documentation changes:


serinamarie:
Hi @dominictarro, thanks for your contribution! I'll have a chat with the team to discuss the UX on this.

dominictarro:
> Hi @dominictarro, thanks for your contribution! I'll have a chat with the team to discuss the UX on this.

Great. I cheated a bit and didn't follow @zanieb's suggestion to avoid using `force=True`. I can correct this and use their other strategy if need be. I imagine it would be useful if the idea is to have the UI explicitly state the task retry was prevented by the `is_retriable` method.

serinamarie:
> Great. I cheated a bit and didn't follow zanieb's suggestion to avoid using `force=True`. I can correct this and use their other strategy if need be. I imagine it would be useful if the idea is to have the UI explicitly state the task retry was prevented by the `is_retriable` method.

Hi @dominictarro, I had a chance to chat with the team about this and we're really excited about your contribution! We do think that because we're interested in seeing details down the line in the UI about why a task was not retriable, it would be ideal to implement the approach to extend state as Zanie suggested:

> - Extend `State.state_details` to include a `retriable: bool = None`
> - Client sets retriable on proposed `FAILED` states depending on the exception type
> - Server retry rules are extended to check for retriable before applying retry logic

[RetryFailedTasks](https://github.com/PrefectHQ/prefect/blob/ced90309a776eaf8610be8c6a4dace4b7af33220/src/prefect/server/orchestration/core_policy.py#L339) would be the aforementioned rule.

Please let us know if you have any questions. Have a good weekend!

dominictarro:
@serinamarie Strategy now uses `StateDetails.retriable` to inform server-side retry logic. Ready for review whenever.

serinamarie:
Hi @dominictarro, thanks so much for this! We'll take a look soon :)

src/prefect/tasks.py
```diff
@@ -210,6 +213,7 @@ def __init__(
         refresh_cache: Optional[bool] = None,
         on_completion: Optional[List[Callable[["Task", TaskRun, State], None]]] = None,
         on_failure: Optional[List[Callable[["Task", TaskRun, State], None]]] = None,
+        is_retriable: Optional[Callable[["Task", TaskRun, State], bool]] = None,
```
zanieb, line 216: Should this be shown to the user as an `on_retry` hook that can return `False` to not retry? An `is_retriable` callable option feels a little awkward.

dominictarro: I was on the fence about that.

1. `on_retry` implies something other than "return true if a retry is okay. return false if not", so I went with `is_retriable`.
1. Documentation would have to note that this runs before any retry happens even if the retry gets rejected. Doesn't really map to the name `on_retry`.

Definitely open to better verbiage though.

dominictarro: Perhaps a `retry_on_failure` parameter that can be a boolean or callable? Similar to how `task_run_name` can be a pre-defined string or generated on the spot. It would be true by default. How do you feel about overriding the retries parameters' role as the decider?

zanieb: I don't think you want a situation where someone could set `retry_on_failure=True` but not set `retries`. It's kind of awkward to duplicate those. You have a great point about `on_retry` feeling a bit wrong too though... hm. Perhaps `should_retry`? Curious to hear the thoughts of the Prefect team :)

dominictarro: It may have been in another branch I started developing on, but I used the word 'ratify' somewhere. Maybe this is the term we're looking for?

serinamarie: @dominictarro Thanks for coming up with a variety of possibilities! I have a few favorites, but I'll chat with the team today and update here before the EOD.

serinamarie: Hi @dominictarro, we weren't able to come to a consensus today but we'll think on it more on Monday :)

serinamarie: We settled on `retry_condition_fn` to match with `cache_key_fn` :)

dominictarro:
@serinamarie This is ready for workflows when suitable for you.

serinamarie:
Hi @dominictarro, thanks for updating that. I'll have a look at the implementation this week. It looks like some of the orchestration tests are failing consistently though.

dominictarro:
> Hi @dominictarro, thanks for updating that. I'll have a look at the implementation this week. It looks like some of the orchestration tests are failing consistently though.

I changed the state whose details get set from initial to proposed in [bfb9fde](https://github.com/PrefectHQ/prefect/pull/10671/commits/bfb9fde87be1feefa209a397cd1c04b13d5bf72b). Forgot to change the tests. Passing locally now.

src/prefect/server/orchestration/core_policy.py
```diff
@@ -273,8 +273,9 @@ class RetryFailedFlows(BaseOrchestrationRule):
     Rejects failed states and schedules a retry if the retry limit has not been reached.
 
     This rule rejects transitions into a failed state if `retries` has been
-    set and the run count has not reached the specified limit. The client will be
-    instructed to transition into a scheduled state to retry flow execution.
+    set, the run count has not reached the specified limit, and the client
+    asserts it is a retriable flow run. The client will be instructed to
+    transition into a scheduled state to retry flow execution.
```
serinamarie, line 278: My understanding was this PR is only for retriable task runs?

src/prefect/tasks.py
```diff
@@ -173,6 +173,10 @@ class Task(Generic[P, R]):
             execution with matching cache key is used.
         on_failure: An optional list of callables to run when the task enters a failed state.
         on_completion: An optional list of callables to run when the task enters a completed state.
+        retry_condition_fn: An optional callable run when a task run returns a Failed state. Should
+            return `True` if the task should continue to its retry policy, and `False` if the task
+            should end as failed. Defaults to `None`, indicating the task should always continue
+            to its retry policy.
```
serinamarie, line 179:
```suggestion
        retry_condition_fn: An optional callable run when a task run returns a Failed state. Should
            return `True` if the task should continue to its retry policy (e.g. `retries=3`), and `False` if the task
            should end as failed. Defaults to `None`, indicating the task should always continue
            to its retry policy.
```

src/prefect/tasks.py
```diff
@@ -1058,6 +1072,10 @@ def task(
...
```
serinamarie, line 1078:
```suggestion
        retry_condition_fn: An optional callable run when a task run returns a Failed state. Should
            return `True` if the task should continue to its retry policy (e.g. `retries=3`), and `False` if the task
            should end as failed. Defaults to `None`, indicating the task should always continue
            to its retry policy.
```

serinamarie (commented):
Hi @dominictarro, this is looking great! I left a few comments on implementation when you get a chance :)

src/prefect/engine.py
```diff
@@ -1798,7 +1798,12 @@ async def orchestrate_task_run(
                     )
                     terminal_state.state_details.cache_key = cache_key
 
+            # Defer to user to decide whether failure is retriable
+            terminal_state.state_details.retriable = bool(
+                await _check_task_failure_retriable(task, task_run, terminal_state)
+            )
```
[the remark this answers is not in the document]
dominictarro: I had the entire resolution logic in the `_check_task_failure_retriable` method. I didn't document the default value when `task_run.retry_condition_fn` is None, so I will add that. Also, bool casting here seems more the scope of `_check_task_failure_retriable`. I will move that as well.

src/prefect/server/orchestration/core_policy.py
dominictarro: It is. The only change here was informing anyone inspecting `RetryFailedFlows` that there are now 3 conditions that have to be met.

serinamarie: Ah, I actually meant to refer to line 296 (the 3rd condition). Why would the tasks being retriable matter when we decide to retry a failed flow?

dominictarro: Ah, I may have been thinking I will do this for tasks and flows when I started. I can walk that back.

serinamarie: Thanks @dominictarro :)

src/prefect/engine.py
```diff
@@ -1798,7 +1798,12 @@ async def orchestrate_task_run(
                     )
                     terminal_state.state_details.cache_key = cache_key
 
+            # Defer to user to decide whether failure is retriable
+            terminal_state.state_details.retriable = (
+                await _check_task_failure_retriable(task, task_run, terminal_state)
+            )
```
serinamarie, line 1804: Should we check that the terminal state `is_failed()` here first before we bother to change the `retriable` bool?

See line 1793-1799. I don't think we'd need to bother doing this for with completed tasks.

src/prefect/engine.py
```diff
@@ -2395,6 +2400,48 @@ async def _run_task_hooks(task: Task, task_run: TaskRun, state: State) -> None:
...
+    logger = task_run_logger(task_run)
+    try:
+        logger.debug(
+            f"Running `retry_condition_fn` check {task.retry_condition_fn!r} for task"
+            f" {task.name!r}"
+        )
+        if is_async_fn(task.retry_condition_fn):
+            return await task.retry_condition_fn(
+                task=task, task_run=task_run, state=state
+            )
```
serinamarie, line 2422: Can we check if `task.retry_condition_fn` is callable first so we can surface a nicer error?

tests/test_engine.py
```diff
@@ -1131,6 +1131,52 @@ def flaky_function():
                     log_prints=False,
                 )
 
+    async def test_is_retriable_after_failure(
```
serinamarie, line 1134: Can we also add a test to ensure that when retries are set to 0, we do not retry regardless of whether it's retriable by default.

And after adding the check for `is_failed` for setting the `terminal_state.state_details.retriable` bool, I think it would make sense to add a test to ensure when a task completes, that we assert the `check_task_failure_retriable` is not called.

serinamarie (changes requested):
Hi @dominictarro, thanks so much for sticking through the multiple reviews. I think this is definitely wrapping up and looking good! A few things:

1. I made some notes about good additional tests.
2. We discussed keeping the scope of this PR to retriable tasks.
3. I added a note about conditionally running the retry check based on whether the terminal state of the task is not a `completed` one.

tests/test_engine.py
dominictarro: 1. Test added to test_engine.py to make sure the task does not retry when retries=0 even if retriable.
2. `check_task_failure_retriable` is only called when both are true:
    1. `terminal_state` is a failed state
    2. `task.retry_condition_fn` is a callable

src/prefect/engine.py
dominictarro: Callable check added to log the type and traceback.

src/prefect/engine.py
dominictarro: Was a bit uncertain between "failed" and "not completed", but figure there's no point for states like cancelled. Added check to only run on *failed* states.

dominictarro:
- _tests/server/orchestration/test\_core\_policy.py::TestTaskRetryingRule_ passing locally
- _tests/test\_engine.py::TestOrchestrateTaskRun_ passing locally
- _tests/test\_engine.py::test\_\_check\_task\_failure\_retriable\_not\_callable_ passing locally

I think this is in the ballpark. Summarizing the above ones

- Limited scope of PR to tasks only
- Changed logic for when `_check_task_failure_retriable` is run
    - Added tests
- Log error and set `retriable=False` when `retry_condition_fn` is not callable
    - Added test
- Added test to ensure no retry when `retries=0`

src/prefect/engine.py
serinamarie: Awesome!

schwannden:
would love this pr to be merged!

src/prefect/engine.py
serinamarie: I think we should validate this during task creation (in `tasks.py`).

tests/test_engine.py
```diff
@@ -1284,6 +1285,155 @@ def flaky_function():
...
+            interruptible=False,
+            client=prefect_client,
+            log_prints=False,
+        )
+
+        # Check that the task completed happily
+        assert state.is_failed()
+        assert mock.call_count == 2
+
+    async def test_no_retry_on_retries_eq_0_even_if_retriable(
```
serinamarie, line 1334:
```suggestion
    async def test_retry_condition_fn_when_retries_eq_0_does_not_retry(
```

tests/test_engine.py
```diff
@@ -1284,6 +1285,155 @@ def flaky_function():
                     log_prints=False,
                 )
 
+    async def test_is_retriable_after_failure(
```
serinamarie, line 1288:
```suggestion
    async def test_retry_condition_fn_retries_after_failure(
```

tests/test_engine.py
```diff
@@ -1284,6 +1285,155 @@ def flaky_function():
...
+        await prefect_client.set_flow_run_state(
+            flow_run_id=flow_run.id,
+            state=Running(),
+        )
+
+        # Define a task that fails once and then succeeds
+        mock = MagicMock()
+
+        # Can retry only once
+        def is_retriable(task, task_run, state):
```
serinamarie, line 1301: Can we do one more test for when the criteria is not met in the `retry_conditinon_fn` for a retry?

tests/test_engine.py
```diff
@@ -1284,6 +1285,155 @@ def flaky_function():
...
+            parameters={"x": quote(1)},
+            result_factory=result_factory,
+            interruptible=False,
+            client=prefect_client,
+            log_prints=False,
+        )
+
+        # Check that the task completed happily
+        assert state.is_failed()
+        assert mock.call_count == 1
```
serinamarie, line 1378: Just for sanity, maybe we should also be check that `is_retriable` is called once

tests/test_engine.py
```diff
@@ -2524,3 +2674,31 @@ async def mock_resolve_inputs(*args, **kwargs):
             )
 
     assert "Task parameter introspection took" not in caplog.text
+
+
+async def test__check_task_failure_retriable_not_callable(
+    flow_run, prefect_client, caplog: pytest.LogCaptureFixture
+):
+    @task(retry_condition_fn="not a callable")
+    def my_task():
```
serinamarie, line 2683: After moving the `retry_condition_fn` validation to `tasks.py` this could go in `test_tasks.py`

tests/test_engine.py
```diff
@@ -1284,6 +1285,155 @@ def flaky_function():
...
+        assert mock.call_count == 1
+
+    @pytest.mark.parametrize(
+        "state_constructor_fn",
+        [
+            Completed,
+            Cancelled,
+        ],
+    )
+    async def test_dont_call_retry_condition_fn_on_non_failure_state(
```
serinamarie, line 1387:
```suggestion
    async def test_retry_condition_fn_not_called_in_non_failure_terminal_state(
```

tests/test_engine.py
serinamarie: The tests look really good!

serinamarie (changes requested):
Hi @dominictarro, I left a few more comments but it's looking really close! I think if:
- some small suggestion are implemented in `test_engine.py`
- the validation for the `retry_condition_fn` is moved to `tasks.py`
- the one test gets moved into `test_tasks.py`
- all tests pass

We're good to go for this week's release! :) Thanks so much again.

dominictarro:
- Made test function names more explicit
- Fixed up some comments in tests
- Test for `retry_condition_fn` call count where appropriate
- Moved `retry_condition_fn` type validation to task initialization
- FCoro `retry_condition_fn` wasn't casts to bool
- Test for `retry_condition_fn` rejecting a retry and the orchestrator adhering to it

Thanks for helping me through this, @serinamarie! I look forward to using it.

tests/test_engine.py
```diff
@@ -1373,9 +1375,11 @@ def my_task(x):
             log_prints=False,
         )
 
-        # Check that the task completed happily
+        # Check that the task failed after only one attempt
         assert state.is_failed()
         assert mock.call_count == 1
+        # Check that the retry condition function was only called once
+        assert mock_2.call_count == 1
```
serinamarie: Shouldn't the call count of these be 0? And the comment updated to match/

tests/test_engine.py
```diff
@@ -1331,6 +1331,56 @@ def my_task(x):
...
+            interruptible=False,
+            client=prefect_client,
+            log_prints=False,
+        )
+
+        # Check that the task failed after only one attempt
+        assert state.is_failed()
+        assert mock.call_count == 1
+        # Check that the retry condition function was only called once
+        assert mock_2.call_count == 1
```
serinamarie, line 1381: I think the call count should be 0 and

`# Check that the retry condition function was only called once`

is incorrect here.

src/prefect/tasks.py
```diff
@@ -342,6 +342,17 @@ def __init__(
             )
         self.on_completion = on_completion
         self.on_failure = on_failure
+
+        if not (
+            callable(retry_condition_fn)
+            or is_async_fn(retry_condition_fn)
+            or retry_condition_fn is None
+        ):
```
serinamarie, line 350: @dominictarro I think we can just check if it's not callable. We don't want the other conditions to raise an error.

serinamarie (changes requested):
Hey, @dominictarro, just a few more comments.

tests/test_engine.py
dominictarro: No, the `retry_condition_fn` is always called on a fatal terminal state. It runs before a fail state is proposed to the server and the server adjudicates whether to fail or retry. We expect it to be called once.

serinamarie: Ah yeah I see. I'd forgotten that. Hm okay.

src/prefect/tasks.py
dominictarro: Not totally understanding. I can drop the `is_async_fn` since that doesn't matter until down the line and is covered by `callable`, but I still need to check if it's a callable or None. If `retry_condition_fn` is neither, should it not raise an error? What is the check doing if not erroring on improper argument types?

serinamarie: ah I think I reviewed this too quickly yesterday, my apologies. I see now that `retry_condition_fn` being `None` wouldn't raise an error here, so we're aligned.

serinamarie: @dominictarro I find it a tad hard to read as it is, though. 

Maybe something like:

```python
is_callable = callable(retry_condition_fn)
is_none = retry_condition_fn is None

if not (is_callable or is_none):
    raise
```

serinamarie: or
```python
if retry_condition_fn is not None and not(callable(retry_condition_fn)):
```

dominictarro: Yeah, sounds good. I can make it more legible

serinamarie: Thank you! 🙌

tests/test_tasks.py
```diff
@@ -3788,3 +3788,10 @@ def my_flow():
         state = my_flow._run()
         assert state.type == StateType.FAILED
         assert my_mock.call_args_list == [call("failed1")]
+
+    async def test_error_when_task_condition_fn_is_not_a_callable(self):
```
serinamarie, line 3792:
```suggestion
    async def test_task_condition_fn_raises_when_not_a_callable(self):
```

serinamarie (approved):
@dominictarro Thanks so much for all your work on this! It's very appreciated :)

PR #10675, opened 2023-09-08
Fix typo on the try cloud button

zhen0:
I think the arrow here is a typo? If it's not, the fact that I wasn't sure is perhaps a sign we should take it out and use a more specific icon. 

image

### Checklist

For documentation changes:

### DIFF
--- a/docs/overrides/partials/header.html
+++ b/docs/overrides/partials/header.html
@@ -60,7 +60,7 @@
     <!-- adding tab for Prefect Cloud -->
     <a class="md-button cloud-button list-unstyled" target="_blank" rel="noopener noreferrer"
       href="https://app.prefect.cloud"><b>Try Prefect
-        Cloud > </b></a>
+        Cloud </b></a>
     <!-- end adding Perfect Cloud button-->
 
   </nav>


docs/overrides/partials/header.html
```diff
@@ -60,7 +60,7 @@
     <!-- adding tab for Prefect Cloud -->
     <a class="md-button cloud-button list-unstyled" target="_blank" rel="noopener noreferrer"
       href="https://app.prefect.cloud"><b>Try Prefect
-        Cloud > </b></a>
+        Cloud </b></a>
```
billpalombi, line 63:
```suggestion
        Cloud →</b></a>
```

billpalombi: Want to see how this renders?

zhen0: Let's try it!

billpalombi (approved):
lgtm

discdiver:
I like the change. 

FWIW, the > matches what's on prefect.io 

Screenshot 2023-09-09 at 10 23 11 AM

billpalombi:
Yeah I saw that too. I tried to find a symbol that had the same slope as the one on prefect.io, but it appears to be custom.

zhen0:
Interesting! I agree with Bill that the slope (and spacing) there are slightly different and pretty key.

Issue #10633, opened 2023-09-06
Include run_count in task and flow runtime context


EmilRex:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

The `run_count` is not available in either `prefect.runtime.task_run` or `prefect.runtime.flow_run`. Because of this, they cannot be used to template a task's `result_storage_key`. This in turn forces an overwrite of results on retries which may not be desirable, either because the results need to be inspected, or because the storage location does not support overwriting.

### Describe the proposed behavior

Both `prefect.runtime.task_run` and `prefect.runtime.flow_run` have a `run_count`.

### Example Use

```py
from prefect import flow, task

@flow
def my_flow():
    hello_world()

@task(persist_result=True, result_storage_key="{flow_run.id}_{flow_run.run_count}_{task_run.id}_{task_run.run_count}.json")
def hello_world(name: str = "world"):
    return f"hello {name}"

my_flow()

```

PR #10676, opened 2023-09-08
Add `prefect.runtime.flow_run.run_count` and `prefect.runtime.task_run.run_count`

serinamarie:
Adds support for flow run / task run `run_count` to the runtime.

Closes https://github.com/PrefectHQ/prefect/issues/10633.

### Example
```python

@flow
def my_flow():
    hello_world()

@task(persist_result=True, result_storage_key="{flow_run.id}_{flow_run.run_count}_{task_run.id}_{task_run.run_count}.json")
def hello_world(name: str = "world"):
    return f"hello {name}"

if __name__ == "__main__":
    my_flow()
```

```bash
In [7]: my_flow()
10:25:03.169 | INFO    | prefect.engine - Created flow run 'pompous-alpaca' for flow 'my-flow'
10:25:03.264 | INFO    | Flow run 'pompous-alpaca' - Created task run 'hello_world-0' for task 'hello_world'
10:25:03.265 | INFO    | Flow run 'pompous-alpaca' - Executing 'hello_world-0' immediately...
10:25:03.318 | INFO    | Task run 'hello_world-0' - Finished in state Completed()
10:25:03.337 | INFO    | Flow run 'pompous-alpaca' - Finished in state Completed('All states completed.')
Out[7]: [Completed(message=None, type=COMPLETED, result=PersistedResult(type='reference', artifact_type='result', artifact_description='Result of type `str` persisted to: `/Users/bean/.prefect/storage/5d4eef08-52c6-4adc-ac9a-94e6473cc9da_1_86a9f022-b9dd-4aa3-9fa4-2065bb9a02d1_1.json`', serializer_type='pickle', storage_block_id=UUID('a7d3c007-41bf-4d9c-83f5-6a8c9d0ad1de'), storage_key='5d4eef08-52c6-4adc-ac9a-94e6473cc9da_1_86a9f022-b9dd-4aa3-9fa4-2065bb9a02d1_1.json'))]
```

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/runtime/task_run.py
+++ b/src/prefect/runtime/task_run.py
@@ -11,6 +11,8 @@ Available attributes:
     - `name`: the name of the task run
     - `tags`: the task run's set of tags
     - `parameters`: the parameters the task was called with
+    - `run_count`: the number of times this task run has been run
+    - `task_name`: the name of the task
 """
 import os
 from typing import Any, Dict, List, Optional
--- a/src/prefect/runtime/task_run.py
+++ b/src/prefect/runtime/task_run.py
@@ -17,7 +17,7 @@ from typing import Any, Dict, List, Optional
 
 from prefect.context import TaskRunContext
 
-__all__ = ["id", "tags", "name", "parameters", "task_name"]
+__all__ = ["id", "tags", "name", "parameters", "run_count", "task_name"]
 
 
 type_cast = {
@@ -83,6 +83,14 @@ def get_tags() -> List[str]:
         return task_run_ctx.task_run.tags
 
 
+def get_run_count() -> int:
+    task_run_ctx = TaskRunContext.get()
+    if task_run_ctx is None:
+        return 0
+    else:
+        return task_run_ctx.task_run.run_count
+
+
 def get_name() -> Optional[str]:
     task_run_ctx = TaskRunContext.get()
     if task_run_ctx is None:
@@ -112,5 +120,6 @@ FIELDS = {
     "tags": get_tags,
     "name": get_name,
     "parameters": get_parameters,
+    "run_count": get_run_count,
     "task_name": get_task_name,
 }
--- a/tests/runtime/test_flow_run.py
+++ b/tests/runtime/test_flow_run.py
@@ -149,6 +149,41 @@ class TestTags:
         assert set(flow_run.tags) == {"red", "green"}
 
 
+class TestRunCount:
+    async def test_run_count_is_attribute(self):
+        assert "run_count" in dir(flow_run)
+
+    async def test_run_count_is_zero_when_not_set(self):
+        assert flow_run.run_count == 0
+
+    async def test_run_count_returns_run_count_when_present_dynamically(self):
+        assert flow_run.run_count == 0
+
+        with FlowRunContext.construct(
+            flow_run=FlowRun.construct(id="foo", run_count=10)
+        ):
+            assert flow_run.run_count == 10
+
+        assert flow_run.run_count == 0
+
+    async def test_run_count_from_api(self, monkeypatch, prefect_client):
+        run = await prefect_client.create_flow_run(
+            flow=flow(lambda: None, name="test", retries=5)
+        )
+
+        await prefect_client.set_flow_run_state(
+            flow_run_id=run.id, state=states.Retrying()
+        )
+
+        await prefect_client.set_flow_run_state(
+            flow_run_id=run.id, state=states.Retrying()
+        )
+
+        monkeypatch.setenv(name="PREFECT__FLOW_RUN_ID", value=str(run.id))
+
+        assert flow_run.run_count == 3
+
+
 class TestStartTime:
     async def test_scheduled_start_time_is_attribute(self):
         assert "scheduled_start_time" in dir(flow_run)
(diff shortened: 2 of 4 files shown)


src/prefect/runtime/task_run.py
```diff
@@ -83,6 +83,14 @@ def get_tags() -> List[str]:
         return task_run_ctx.task_run.tags
 
 
+def get_run_count() -> int:
+    task_run_ctx = TaskRunContext.get()
+    if task_run_ctx is None:
```
jakekaplan, line 88: What's the reason behind not getting the run count from the api here similar to flow run if there is no available ctx?

tests/runtime/test_flow_run.py
```diff
@@ -149,6 +149,38 @@ async def test_tags_pulls_from_api_when_needed(self, monkeypatch, prefect_client
...
+        run = await prefect_client.create_flow_run(
+            flow=flow(lambda: None, name="test", retries=5)
+        )
+        assert flow_run.run_count == 0
+
+        await prefect_client.set_flow_run_state(
+            flow_run_id=run.id, state=states.Retrying()
+        )
+
+        monkeypatch.setenv(name="PREFECT__FLOW_RUN_ID", value=str(run.id))
```
urimandujano, line 179: q: What happens if we don't monkeypatch this? Does the `flow_run.run_count` end up using a different ID?

src/prefect/runtime/task_run.py
```diff
@@ -112,5 +120,6 @@ def get_parameters() -> Dict[str, Any]:
     "tags": get_tags,
     "name": get_name,
     "parameters": get_parameters,
+    "run_count": get_run_count,
```
urimandujano, line 123: nitpick: it seems like we were documenting these values in the docstring at the top of the file, do you think we should add these there?

src/prefect/runtime/task_run.py
serinamarie: @jakekaplan If we are not in a task run context and we are using `runtime.task_run`, I think it may difficult to discern the task run id in order to retrieve attributes from it. 

On the flip side, when we use `runtime.flow_run`, we only create the call when the flow run context is `None` but we have a `run_id` (from being in a task run context or from an env var). 

Said another way, if we are in a task run context when we are calling `runtime.flow_run`, we can retrieve the flow run id from a variety of places, such as the flow run context, task run context, or env var. However, when we call `runtime.task_run` in a task run context, we attempt to retrieve it that way, but if we aren't in one, I don't know of how else to identify what task we're talking about.

Please let me know if that doesn't make sense!

src/prefect/runtime/task_run.py
serinamarie: IMO, `FIELDS` is necessary to find the function to call when a named attribute is accessed. If anything,`__all__` maybe isn't totally necessary but it makes it clear what can be imported directly.

tests/runtime/test_flow_run.py
serinamarie: Retrieving the flow run id happens here:

```python
def get_id() -> str:
    flow_run_ctx = FlowRunContext.get()
    task_run_ctx = TaskRunContext.get()
    if flow_run_ctx is not None:
        return str(flow_run_ctx.flow_run.id)
    if task_run_ctx is not None:
        return str(task_run_ctx.task_run.flow_run_id)
    else:
        return os.getenv("PREFECT__FLOW_RUN_ID")
```

Since in this test case we are not in a flow run context or task run context, if we did not set an env var, I believe "PREFECT__FLOW_RUN_ID" would just be None, so we wouldn't be able to check the attribute is correct in that case and `flow_run.run_count` would be 0.

urimandujano: That makes sense, thanks!

src/prefect/runtime/task_run.py
jakekaplan: thanks for the explanation! That makes a lot of sense to me

jakekaplan (approved):

src/prefect/runtime/task_run.py
urimandujano: [This is the docstring](https://github.com/PrefectHQ/prefect/blob/daebb1b9473cba1f720c12670d423b8da4c71a87/src/prefect/runtime/task_run.py#L9) that I'd noticed. I don't think it's up to date right now but like you mentioned, the `FIELDS`/`__all__` makes it clear what can be used.

urimandujano (approved):

src/prefect/runtime/task_run.py
serinamarie: Ah good catch, let me fix that since the docstring is important for our [API ref](https://docs.prefect.io/latest/api-ref/prefect/runtime/flow_run/).

serinamarie: Updated in [29ee2fb](https://github.com/PrefectHQ/prefect/pull/10676/commits/29ee2fb3c070e2a5457708acd89e1b5377d5eb44)

serinamarie: I totally misunderstood your initial comment, I realize now 😅

PR #10683, opened 2023-09-11
fix base client test

jakekaplan:
With the latest realse of `httpx==0.25.0`, the documentation url for response codes has changed: https://github.com/encode/httpx/blob/master/CHANGELOG.md#0250-11th-sep-2023

This breaks a test as it makes the str version of an exception slightly longer, and the value we are checking for is truncated. 
```
<ExceptionInfo PrefectHTTPStatusError("Client error '400 Bad Request' for url 'fake.url/fake/route'\nResponse: {'extra_info': [{'mess...: 'a test error message'}]}\nFor more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400") tblen=5>
```

This fixes it by using `str(exc.exconly())` to get the full string.


serinamarie (approved):
Thank you Jake!

zzstoatzz (approved):

PR #10718, opened 2023-09-13
Add ability to update existing deployment configs in `prefect.yaml` with `prefect deploy`

serinamarie:
Currently, when a user creates a new deployment configuration, e.g.
```bash
? Would you like to use an existing deployment configuration? [Use arrows to move; enter to select; n to select none]
┏━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┓
┃    ┃ Name                ┃ Entrypoint                     ┃ Description ┃
┡━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━┩
│    │ existing_deployment │ flows/existing_flow.py:my_flow │             │
└────┴─────────────────────┴────────────────────────────────┴─────────────┘
  > No, configure a new deployment
```

they are prompted to save their configuration:
```bash
? Would you like to save configuration for this deployment for faster deployments in the future? [y/n]
```

However, in cases where users want to use an existing deployment from their `prefect.yaml`:

```yaml
deployments:
- name: existing-deployment
  entrypoint: flows/existing_flow.py:my_flow
  schedule:
  work_pool:
    name: my-pool
```
they are not prompted to save the configuration even if they make changes to that deployment. Though there are times that an ad hoc deployment happens, it makes sense to offer the option to update their `prefect.yaml` in these cases.

While this means users will be prompted to save when using existing deployments, it also ensures that if no updates to the existing deployment were made, there is no prompt to save the deployment config.

### Checklist

For documentation changes:


src/prefect/cli/deploy.py
```diff
@@ -591,51 +592,56 @@ async def _run_single_deploy(
...
+                app.console.print(
+                    (
+                        "\n[green]Deployment configuration saved to prefect.yaml![/]"
+                        " You can now deploy using this deployment configuration"
+                        " with:\n\n\t[blue]$ prefect deploy -n"
+                        f" {deploy_config['name']}[/]\n\nYou can also make changes to"
+                        " this deployment configuration by making changes to the"
+                        " prefect.yaml file."
+                    ),
+                )
```
serinamarie, line 644: This was just moved down, but hasn't changed other than being one indentation over

src/prefect/deployments/base.py
```diff
@@ -377,21 +377,17 @@ def _copy_deployments_into_prefect_file():
             f.write(raw_deployment_file_contents)
 
 
-def _save_deployment_to_prefect_file(
+def _format_deployment_for_saving_to_prefect_yaml(
```
serinamarie, line 380: I moved some of the logic from `_save_deployment_to_prefect_file` to a helper function so I could use it elsewhere

tests/cli/test_deploy.py
```diff
@@ -2678,7 +2688,20 @@ async def test_deploy_without_name_in_prefect_yaml_interactive(
             invoke_and_assert,
             command="deploy --all",
             expected_code=0,
-            user_input="y" + readchar.key.ENTER + "test-name-2" + readchar.key.ENTER,
+            user_input=(
```
chrisguidry, line 2691: This is my first time seeing these tests for the interactive prompts and they are :sun_with_face:

src/prefect/cli/deploy.py
```diff
@@ -1326,6 +1331,30 @@ def _check_for_matching_deployment_name_and_entrypoint_in_prefect_file(
     return False
 
 
+def _check_if_identical_deployment_in_prefect_file(
```
chrisguidry, line 1334: :heart:

chrisguidry (approved):
Nice boost to the UX!

zzstoatzz (approved):
i'm also curious about ordering as far as the equality check but if that's good then lgtm!

src/prefect/cli/deploy.py
```diff
@@ -1326,6 +1331,30 @@ def _check_for_matching_deployment_name_and_entrypoint_in_prefect_file(
...
+    )
+
+    prefect_file = Path("prefect.yaml")
+    if prefect_file.exists():
+        with prefect_file.open(mode="r") as f:
+            parsed_prefect_file_contents = yaml.safe_load(f)
+            deployments = parsed_prefect_file_contents.get("deployments")
+            if deployments is not None:
+                for _, existing_deployment in enumerate(deployments):
+                    if existing_deployment == templated_deploy_config:
```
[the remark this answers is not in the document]
serinamarie: That's a good question! The `templated_deploy_config` would be taken straight from the `prefect.yaml` when they are using an existing configuration, so if there aren't changes to it it would be identical. If there are changes to it then we would prompt to save.

If for some reason they happened to specify a configuration with the same `name` and `entrypoint` and nothing else was changed, these dictionaries would still be equal as they'll still have the same k/v regardless of how the user orders them in their `prefect.yaml`. They will not be prompted to save with a deployment with the same entrypoint and name.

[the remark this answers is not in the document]
urimandujano: > The templated_deploy_config would be taken straight from the prefect.yaml

This is the piece that I was missing!

urimandujano (approved):
Good work!

src/prefect/cli/deploy.py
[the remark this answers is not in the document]
serinamarie: I have updated the variables to be more clear, thanks!

Issue #10705, opened 2023-09-12
run_shell_script in deployment.steps.utility does not handle spaces properly on Windows devices


conewman1:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

In our prefect.yaml file we are executing commands using the step type of prefect.deployments.steps.run_shell_script.  If we include an argument that has a space in it, we enclose that in double quotes like you normally would on the command line.

The issue I am seeing is that in this line of code:
[https://github.com/PrefectHQ/prefect/blob/main/src/prefect/deployments/steps/utility.py#L172](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/deployments/steps/utility.py#L172)

It executes **shlex.split(command)** which removes the double quotes.  Apparently there is a known issue where the double quotes need to be added back when on the Windows platform if the arguments are going to be passed back to the subprocess module.  Here is a relevant link with someone else having this issue with shlex:
[https://github.com/jrnl-org/jrnl/issues/348](https://github.com/jrnl-org/jrnl/issues/348)

The recommended workaround is to include the posix argument and set it to True or False based on the platform.

### Describe the proposed behavior

Update [this line](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/deployments/steps/utility.py#L172) so that it looks like this:

`split_command = shlex.split(command, posix="win" not in sys.platform)`

Please note that this was recommended as a solution in the link provided under Current Behavior.   I tested this on a Windows 10 laptop and afterwards I was able to execute this deployment step successfully.  

### Additional context

While our deployed code runs on Linux containers, our developers do initial development and then run deployments to our Dev workspaces from Windows machines.  Not being able to run these utility steps from Windows machines is causing issues on the team.

serinamarie:
Hi @conewman1, thanks so much for this issue! We'd welcome a contribution if you're interested :)

NikoRaisanen:
I'd be happy to tackle this as well.

I want to give you the opportunity first if you want @conewman1 since you already found a solution 😄

conewman1:
@NikoRaisanen please make this change if you don't mind.  I'm still getting used to Github and would prefer this be fixed sooner rather than later.   Thank you!

NikoRaisanen:
Sounds great! I'll get working on it

[...]

PR #10719, opened 2023-09-13
fix: set shlex.split posix argument to False if on Windows

NikoRaisanen:
Closes https://github.com/PrefectHQ/prefect/issues/10705

### Checklist

For documentation changes:


NikoRaisanen:
Hello! This change fixes the issue outlined in the ticket, but I'm struggling with the unit tests.

It seems that the `run_shell_script` method will always error out when running unit tests on Windows.
My basic test to see if I can call run_shell_script without an error while mocking a Windows system:

```
async def test_run_shell_script_split_on_windows(self, monkeypatch):
        monkeypatch.setattr("sys.platform", "win32")
        result = await run_shell_script("echo Hello World")
        assert result["stdout"] == "Hello World"
        assert result["stderr"] == ""
```

The above throws this error when running the unit test:
```
@asynccontextmanager
    async def open_process(command: List[str], **kwargs):
        """
        Like `anyio.open_process` but with:
        - Support for Windows command joining
        - Termination of the process on exception during yield
        - Forced cleanup of process resources during cancellation
        """
        # Passing a string to open_process is equivalent to shell=True which is
        # generally necessary for Unix-like commands on Windows but otherwise should
        # be avoided
        if not isinstance(command, list):
            raise TypeError(
                "The command passed to open process must be a list. You passed the command"
                f"'{command}', which is type '{type(command)}'."
            )
    
        if sys.platform == "win32":
            command = " ".join(command)
>           process = await _open_anyio_process(command, **kwargs)
E           NameError: name '_open_anyio_process' is not defined

src/prefect/utilities/processutils.py:200: NameError
=========================== short test summary info ============================
FAILED tests/deployment/test_steps.py::TestRunShellScript::test_run_shell_script_split_on_windows - NameError: name '_open_anyio_process' is not defined
======================== 1 failed, 48 passed in 25.13s =========================
```

If I remove the monkeypatch that sets sys.platform to win32, then the tests run without errors. Do we think that this change needs an associated test? And if so, I'm open to suggestions!

urimandujano:
Hey @NikoRaisanen, thanks for all the effort here! 

I took a look and it looks like what's happening is that [we're conditionally defining](https://github.com/PrefectHQ/prefect/blob/9f92b6475586997206aba2282a3eb5060eb29ab4/src/prefect/utilities/processutils.py#L30) the  `_open_anyio_process` function depending on if we've detected that we're running on a Windows system or not. That conditional is evaluated at import time (whenever we do `from prefect.deployments.steps import run_shell_script` -- meaning that by the time we get to your test we've already decided to define the function or not!

What I would recommend is using a `pytest.mark` to skip the test if its run on non-windows. [You can see how that works on these lines](https://github.com/PrefectHQ/prefect/blob/9f92b6475586997206aba2282a3eb5060eb29ab4/tests/utilities/test_processutils.py#L72-L75C6). If you add the label `arch:windows` to the PR then I believe we'll run the CI on a Windows machine and be able to verify your change works as expected.

NikoRaisanen:
Hey @urimandujano thanks for the feedback! I just pushed up the test and marked it as windows only.
I don't have permissions to add that `arch:windows` label

urimandujano:
I just found out that our Windows CI workflow isn't working as expected by now. I'll have to test this manually and get back to you -- so sorry for the delay!

urimandujano:
It looks like that test is still failing, can you look into it? Here's what the output was:

```
    @asynccontextmanager
    async def open_process(command: List[str], **kwargs):
        """
        Like `anyio.open_process` but with:
        - Support for Windows command joining
        - Termination of the process on exception during yield
        - Forced cleanup of process resources during cancellation
        """
        # Passing a string to open_process is equivalent to shell=True which is
        # generally necessary for Unix-like commands on Windows but otherwise should
        # be avoided
        if not isinstance(command, list):
>           raise TypeError(
                "The command passed to open process must be a list. You passed the command"
                f"'{command}', which is type '{type(command)}'."
            )
E           TypeError: The command passed to open process must be a list. You passed the command'<MagicMock name='mock()' id='2937046418624'>', which is type '<class 'unittest.mock.MagicMock'>'.

src\prefect\utilities\processutils.py:193: TypeError
```

NikoRaisanen:
Hey @urimandujano! I fixed the test and confirmed that it worked for me locally. Despite mocking the return_value, this test will confirm that shlex.split was called with the `posix=False` argument

urimandujano (approved):
Thanks a lot @NikoRaisanen!

PR #10724, opened 2023-09-14
Add an `is_null` filter for Deployments to flows/filter

### LINKED ISSUE #10715: Ability to filter by deployments -> deployment_id -> is_null
### First check

### Prefect Version

2.x

There's a [feature request](https://github.com/PrefectHQ/prefect/issues/9517) to be able to filter the flows page by Flows that either do or do not have a deployment currently associated with them.

There is a way to do this, but it won't work. We could pass:
```
{
  flow_runs: {
    deployment_ids: {
      is_null: bool
    }
  }
}
```
For this use case though, there are scenarios where the filter isn't quite accurate. You may have a flow with no associated deployment, but the flow was previously ran with a deployment. Or you may have a flow that is associated with a deployment, but just hasn't been ran from one yet.

We need a way to filter like this instead:
```
{
  deployments: {
    ids: {
      is_null: bool
    }
  }
}
```
That way we only get flows that do or do-not have deployments associated with them.

### Example Use

The Flows page could be filtered by flows that are or are not currently deployed.

### Additional context

_No response_

bunchesofdonald:
This adds an `is_null` filter for deployments to the flows/filter route to be able to read flows that have associated deployments or have no associated deployments.

Closes #10715 

### Example

To get all flows that have at least one associated deployment:

```
{
	"flows": {"deployment": { "is_null_": false}}
}
```

To get all flows that do not have any associated deployments:

```
{
	"flows": {"deployment": { "is_null_": true}}
}
```

### Checklist

### DIFF
--- a/src/prefect/server/schemas/filters.py
+++ b/src/prefect/server/schemas/filters.py
@@ -80,6 +80,34 @@ class FlowFilterId(PrefectFilterBaseModel):
         return filters
 
 
+class FlowFilterDeployment(PrefectOperatorFilterBaseModel):
+    """Filter by flows by deployment"""
+
+    is_null_: Optional[bool] = Field(
+        default=None,
+        description="If true, only include flow runs without deployment ids",
+    )
+
+    def _get_filter_list(self, db: "PrefectDBInterface") -> List:
+        filters = []
+
+        if self.is_null_ is not None:
+            deployments_subquery = (
+                sa.select(db.Deployment.flow_id).distinct().subquery()
+            )
+
+            if self.is_null_:
+                filters.append(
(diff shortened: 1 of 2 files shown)


src/prefect/server/schemas/filters.py
```diff
@@ -80,6 +80,34 @@ def _get_filter_list(self, db: "PrefectDBInterface") -> List:
         return filters
 
 
+class FlowFilterDeployment(PrefectOperatorFilterBaseModel):
+    """Filter by flows by deployment"""
+
+    is_null_: Optional[bool] = Field(
+        default=None,
+        description="If true, only include flow runs without deployment ids",
```
chrisguidry, line 88:
```suggestion
        description="If true, only include flows without deployments",
```

PR #10837, opened 2023-09-27
Use only the security suite from CodeQL

chrisguidry:
We now use `ruff` for our quality checks, which are run via `pre-commit` for
developers locally and also during CI.  The CodeQL security checks are a
valuable signal, but they are often lost in the noise of regular linting.  Here
we downshift our use of CodeQL to the `security-extended` suite.


bunchesofdonald (approved):
Thank you!

PR #10838, opened 2023-09-27
Use a fixed list of known registry views rather than using the user's input

chrisguidry:
CodeQL identified a partial server-side request forgery with a pretty limited
blast radius.  It could be exploited to cause a bunch of 404s at Github in a
very specific repo, but not likely any further harm than that.  Using a fixed
list of known views prevents us from even attempting the request for unknown
views.


desertaxle (approved):
LGTM!

PR #10855, opened 2023-09-29
Adding explicit tests for the format of the `PREFECT__FLOW_RUN_ID` env var

chrisguidry:
We had an incident where we caused a large number of 404s because we changed the
`read_flow_run` route to use the specifier `{id:uuid}`. This specifier expects
UUIDs to have dashes, but the `PREFECT__FLOW_RUN_ID` variable is produced with
`flow_run.id.hex` and will be a 32-character hex string without dashes.

This test makes it explicit that we're supporting that unexpected format so we
don't accidentally break this again.

Related to #10851

### DIFF
--- a/tests/server/api/test_flow_runs.py
+++ b/tests/server/api/test_flow_runs.py
@@ -1,4 +1,5 @@
 from typing import List
+from unittest import mock
 from uuid import UUID, uuid4
 
 import pendulum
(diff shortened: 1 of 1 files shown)


serinamarie (approved):
Thank you :)

desertaxle (approved):

PR #10860, opened 2023-09-29
Vendoring FastAPI as a first step toward `pydantic>2` support

chrisguidry:
While we begin to work on `pydantic>2` support, we need to support living in
any of three different "worlds" concurrently:

1. `pydantic<2` is installed (today)
2. `pydantic>2` is installed, `prefect.server.schemas` are pydantic v1 models (intermediate)
3. `pydantic>2` is installed, `prefect.server.schemas` are pydantic v2 models (end state)

FastAPI's compatibility layer only expects (1) or (3), based on the reasonable
assumption that FastAPI users are in control of their `pydantic` version and
schemas.  However, `prefect` will need to exist in world (2) for a prolonged
period of time while we deprecate support for `pydantic<2`.

To this end, we are removing our dependency on the public `fastapi` package, and
vendoring it privately to `prefect` for use exclusively for `prefect.server` and
the short-term login server used for Cloud login.  This means that end users
may have any version of FastAPI they like without conflicting with ours.  We
will make some adjustments to the vendored version's compatibility layer to help
bridge the gap between worlds (1) and (3).

Long-term, we will spin this down and replace it with the public `fastapi`
package again.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 aiosqlite >= 0.17.0
 alembic >= 1.7.5, < 2.0.0
-anyio >= 3.4.0, < 4.0.0
+anyio >= 3.7.1, < 4.0.0
 apprise >= 1.1.0, < 2.0.0
 asgi-lifespan >= 1.0, < 3.0
 asyncpg >= 0.23
@@ -11,7 +11,6 @@ croniter >= 1.0.12, < 2.0.0
 cryptography >= 36.0.1
 dateparser >= 1.1.1, < 2.0.0
 docker >= 4.0, < 7.0
-fastapi >= 0.93
 fsspec >= 2022.5.0
 graphviz >= 0.20.1
 griffe >= 0.20.0
@@ -25,7 +24,8 @@ orjson >= 3.7, < 4.0
 packaging >= 21.3, < 24.3
 pathspec >= 0.8.0
 pendulum >= 2.1.2, < 3.0.0
-pydantic >= 1.10.0, < 2.0.0
+# the version constraints for pydantic are merged with those from fastapi 0.103.2
+pydantic>=1.10.0,< 2.0.0,>=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0
 python-slugify >= 5.0, < 9.0
 pytz >= 2021.1, < 2024
 pyyaml >= 5.4.1, < 7.0.0
@@ -33,8 +33,9 @@ readchar >= 4.0.0, < 5.0.0
 rich >= 11.0, < 14.0
 ruamel-yaml >= 0.17.0
 sqlalchemy[asyncio] >= 1.4.22, != 1.4.33, < 3.0.0
+starlette >= 0.27.0, < 0.28.0
 toml >= 0.10.0
 typer >= 0.4.2
-typing_extensions >= 4.1.0, < 5.0.0
+typing_extensions >= 4.5.0, < 5.0.0
 uvicorn >= 0.14.0
 websockets >= 10.4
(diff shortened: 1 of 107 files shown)


tests/server/utilities/test_server.py
```diff
@@ -206,16 +211,11 @@ def foo(
     app.include_router(router)
 
     client = TestClient(app)
-    if packaging.version.parse(fastapi.__version__) < packaging.version.parse("0.74.0"):
```
chrisguidry, line 209: We own our version of FastAPI now so this doesn't apply, and when we come back with `pydantic>2` support, it will have to be on a version somewhere north of `0.100`.

requirements.txt
```diff
@@ -25,16 +24,18 @@ orjson >= 3.7, < 4.0
...
+# the version constraints for pydantic are merged with those from fastapi 0.103.2
+pydantic>=1.10.0,< 2.0.0,>=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0
 python-slugify >= 5.0, < 9.0
 pytz >= 2021.1, < 2024
 pyyaml >= 5.4.1, < 7.0.0
 readchar >= 4.0.0, < 5.0.0
 rich >= 11.0, < 14.0
 ruamel-yaml >= 0.17.0
 sqlalchemy[asyncio] >= 1.4.22, != 1.4.33, < 3.0.0
+starlette >= 0.27.0, < 0.28.0
```
urimandujano, line 36: Note to self: whenever we decide to de-vendor fastapi we can replace this `startlette` pin with `fastapi` and be more-or-less back in tune with our own dependencies before we vendored it.

src/prefect/_vendor/README.md
```diff
@@ -0,0 +1,26 @@
...
+makes a strong assumption that having `pydantic>2` installed means that all of your
+models are v2 models.  We will be adjusting that compatibility layer accordingly.
+
+If we do need to update the version before we're done with it:
+
+```
+VERSION=0.103.2
+rm -Rf /tmp/fastapi
+git clone https://github.com/tiangolo/fastapi -b $VERSION /tmp/fastapi
+rsync -ravdP /tmp/fastapi/fastapi/ src/prefect/_vendor/fastapi/
```
urimandujano, line 16: Thank you for this! Can 100% see a future where we need to update our internal fastapi version before we fully deprecate pydantic v1.

src/prefect/cli/profile.py
```diff
@@ -7,9 +7,9 @@
 
 import httpx
 import typer
-from fastapi import status
 from rich.progress import Progress, SpinnerColumn, TextColumn
 from rich.table import Table
+from starlette import status
```
urimandujano, line 12: I'm going a little bit back and forth on these import changes -- I think we'd want to list `starlette` as an explicit dependency if we're importing from it. So we'd just keep it in our `requirements.txt` even after we're complete the pydantic v2 support.

requirements.txt
urimandujano: Updated note to self: if we import from `starlette` then we will probably want to keep it listed here.

urimandujano (approved):
This is an incredible lift, thanks for handling it!

src/prefect/cli/profile.py
chrisguidry: Yes, I was a little torn on this too.  This module in `fastapi` is just a "passthrough" from `starlette`, so I decided to go right to the source.  It's already a little odd for us to use a server-side module of constants for tons of client-side logic and test assertions.  If we're going to keep isolating `prefect.client` and `prefect.server`, we may want to choose a different source of these constants for the client-side (maybe just creating our own module of them, they're not proprietary).

PR #10867, opened 2023-10-02
Isolating the server-side schemas from the client-side schemas

chrisguidry:
As the next step toward `pydantic>2` compatibility, we're separating the
implementation of the `prefect.server.schemas` from the `prefect.client.schemas`
by duplicating some of the shared base classes/fields/utilities.  This will
allow us to evolve the client schemas to support `pydantic>2` at a different
rate than the server schemas.

This is intended to be a purely no-op refactoring, and the next step will be
to strip down some of the implementation complexity in the server schemas in
preparation of `pydantic>2`.


src/prefect/server/utilities/schemas/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .bases import PrefectBaseModel, IDBaseModel, ORMBaseModel, ActionBaseModel
```
chrisguidry, line 1: The layout of this `schemas` package is meant to mirror the client-side `schemas` package as well, with a few smaller files.

jakekaplan (approved):
Makes sense to me!

PR #10873, opened 2023-10-02
Conditional imports for `pydantic`

chrisguidry:
As the next step along the path of supporting `pydantic>2`, we are using
conditional imports in all source files where we work with or define `pydantic`
models.  Our medium-term compatibility story will be that we'll support users on
either version of `pydantic`, up to the deprecation period.  That means that
we will be using `pydantic<1` idioms and the `pydantic.v1` compatibility module
until that time.

This updates all server and client schemas, as well as additional client-side
objects that aren't part of the Prefect API.  Server tests are passing with 
`pydantic>2` locally (not in CI yet).

Note: this version includes a downgrade of the vendored FastAPI to 0.99.1, which 
is the version prior to `pydantic>2` support.  This is a known working version
of FastAPI and eliminates a lot of mystery failures where `PYDANTIC_V2` had leaked
outside the `_compat.py` file.  For reviewers: I don't recommend reading all the 
details of the FastAPI diffs, just the imports.


src/prefect/_vendor/README.md
```diff
@@ -1,26 +1,17 @@
...
 
-If we do need to update the version before we're done with it:
+If we do need to change the version before we're done with it:
 
 ```
-VERSION=0.103.2
-rm -Rf /tmp/fastapi
-git clone https://github.com/tiangolo/fastapi -b $VERSION /tmp/fastapi
-rsync -ravdP /tmp/fastapi/fastapi/ src/prefect/_vendor/fastapi/
+./src/prefect/_vendor/vendor-fastapi 0.99.1
```
urimandujano, line 13: 😮

urimandujano (approved):
One nice thing about the more verbose imports -- we can't miss them and we'll be constantly reminded to "actually deprecate v1 at some point"

chrisguidry:
Yeah my DRY instincts are screaming, but I think this is the best way to give us a path to an incremental approach over time.

PR #10876, opened 2023-10-03
Preserve set attribute information in tasks during execution

urimandujano:
When executing a `task` that takes an instance of a `pydantic.BaseModel`, we construct a copy of the input instance and  lose track of which attributes were explicitly set. This is evident when running `.json(exclude_unset=True)` on the instance from inside the task. This change copies information on which fields were explicitly set from the input instance and onto the copy that we create and return. 

### Example

When a model is parsed by a task as input, the value passed is a reconstructed instance that loses some info from the original instance:
```python
from typing import Optional

class User(BaseModel):
    """Pydantic model for the User."""

    email: Optional[str]
    firstName: Optional[str]
    lastName: Optional[str]


@task()
def t(u: User):
    assert u.dict(exclude_unset=True) == {
        "firstName": "John"
    }, f"Model contains unset fields: {u.dict(exclude_unset=True)}"


@flow()
def fl_create_user():
    user = User(firstName="John")
    assert user.dict(exclude_unset=True) == {"firstName": "John"}
    t(user)


fl_create_user()

```

### Checklist

### DIFF
--- a/tests/utilities/test_collections.py
+++ b/tests/utilities/test_collections.py
@@ -328,6 +328,19 @@ class TestVisitCollection:
             output_model.val == input_model.val
         ), "The fields value should be used, not the default factory"
 
+    def test_visit_collection_remembers_unset_pydantic_fields(self):
+        class RandomPydantic(pydantic.BaseModel):
+            val: uuid.UUID = pydantic.Field(default_factory=uuid.uuid4)
+
+        input_model = RandomPydantic()
+        output_model = visit_collection(
+            input_model, visit_fn=visit_even_numbers, return_data=True
+        )
+
+        assert output_model.json(exclude_unset=True) == input_model.json(
+            exclude_unset=True
+        ), "Unset fields values should be remembered and preserved"
+
     @pytest.mark.parametrize("immutable", [True, False])
     def test_visit_collection_mutation_with_private_pydantic_attributes(
         self, immutable
--- a/tests/utilities/test_collections.py
+++ b/tests/utilities/test_collections.py
@@ -142,6 +142,12 @@ class ImmutablePrivatePydantic(PrivatePydantic):
         allow_mutation = False
 
 
+class PydanticWithDefaults(pydantic.BaseModel):
+    name: str
+    val: uuid.UUID = pydantic.Field(default_factory=uuid.uuid4)
+    num: int = 0
+
+
 @dataclass
 class Foo:
     x: Any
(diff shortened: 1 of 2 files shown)


tests/utilities/test_collections.py
```diff
@@ -328,6 +328,19 @@ class RandomPydantic(pydantic.BaseModel):
             output_model.val == input_model.val
         ), "The fields value should be used, not the default factory"
 
+    def test_visit_collection_remembers_unset_pydantic_fields(self):
+        class RandomPydantic(pydantic.BaseModel):
+            val: uuid.UUID = pydantic.Field(default_factory=uuid.uuid4)
```
chrisguidry, line 333: I like this test a lot, but I kinda want to see this `RandomPydantic` class have more than one field, and then make assertions on a mix of set and unset fields.  I'm pretty sure this test is correct, but I think it could demonstrate it more clearly.

tests/utilities/test_collections.py
```diff
@@ -328,6 +328,19 @@ class RandomPydantic(pydantic.BaseModel):
...
+
+        input_model = RandomPydantic()
+        output_model = visit_collection(
+            input_model, visit_fn=visit_even_numbers, return_data=True
+        )
+
+        assert output_model.json(exclude_unset=True) == input_model.json(
+            exclude_unset=True
+        ), "Unset fields values should be remembered and preserved"
```
chrisguidry, line 342: I usually like to have my tests be as "flat" and hardcoded as I can make them, so they're dead simple to reason about.  What if we do something like this (assuming that we've extended the model to have, say, three string fields called `val1`, `val2`, `val3`:

```suggestion
        input_model = RandomPydantic(val1="first", val2="second")
        assert input_model.json(exclude_unset=True) == {
            "val1": "first-value",
            "val2": 'second"
        }
        
        output_model = visit_collection(
            input_model, visit_fn=visit_even_numbers, return_data=True
        )
        assert output_model.json(exclude_unset=True) == {
            "val1": "first",
            "val2": 'second"
        }, "Unset fields values should be remembered and preserved"
```

This way we're expanding the values so the person reading the test can see the intent directly.  Like, if you get an assertion failure saying that you expected

```
{"val1": "first", "val2": "second"}
```

and got :
```
{"val1": "first", "val2": "second", "val3": null}
```

It will be hyper-clear what's going on.

chrisguidry (approved):

tests/utilities/test_collections.py
urimandujano: I __strongly dislike__ staring at failing test output so I'm for tests being hyper clear about failures. I took your advice from above and parameterized this test which makes it a bit clunky to assert on the `input_model`. The assertion on the `output_model` did become more clear because I was able to assert on the input dict directly. What do you think about that?

chrisguidry: Oooh I like that a lot!  More tests for the buck!

tests/utilities/test_collections.py
urimandujano: Good point -- I went ahead and added one required field and one optional field (of a different type) and am setting a plain Python default (instead of using `pydantic.Field`) to make sure our bases are covered. Let me know if there's an extra case that might be interesting.

PR #10877, opened 2023-10-03
Creating CI actions for pydantic v2

chrisguidry:
I'm choosing to run a smaller subset of actions here, just to keep things from
ballooning:

```
Python 3.8      sqlite          service tests
             x               x
Python 3.11     postgres 14
```

Eventually, we'll flip this so that `pydantic>2` is the default for all test
suites, and we'll preserve a small subset like this for `pydantic<2`.

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -322,6 +322,11 @@ jobs:
           tmp="sha-$SHORT_SHA-python${{ matrix.python-version }}"
           echo "image_tag=${tmp}" >> $GITHUB_OUTPUT
 
+      - name: Force install pydantic in docker images
+        if: ${{ matrix.build-docker-images }}
+        run: |
+          sed -i 's/RUN prefect version/RUN pip install "pydantic>2"\nRUN prefect version/' Dockerfile
+
       - name: Build test image
         if: ${{ matrix.build-docker-images }}
         uses: docker/build-push-action@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/python-tests.yaml
```diff
@@ -249,6 +249,190 @@ jobs:
...
+        if: ${{ matrix.build-docker-images }}
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          # TODO: We do not need the UI in these tests and we may want to add a build-arg to disable building it
+          #       so that CI test runs are faster
+          build-args: |
+            PYTHON_VERSION=${{ matrix.python-version }}
+            PREFECT_EXTRAS=[dev]
+          tags: prefecthq/prefect-dev:${{ steps.get_image_tag.outputs.image_tag }}
```
serinamarie, line 335: Do you think it would be a good idea to append `pydanticv2` here?

urimandujano (approved):

.github/workflows/python-tests.yaml
chrisguidry: Thinking through that....I believe these tags are used in the test suite, so I don't think we can change them.  For what it's worth, these images aren't the ones that ever get pushed anywhere, they're just built locally for the tests.

But you _did_ make me realize that we aren't force-installing pydantic inside the test docker images, gonna do that now.

PR #10898, opened 2023-10-05
Always use the larger `oss-test-runner`

### LINKED ISSUE #10896: Widespread test failures with an unclosed connection
### First check

### Description

Starting on 10/5, we are seeing widespread test failures due to an unclosed connection somewhere in the test suite.  All of the unit tests appear to be passing, but each workflow fails at cleanup due to this:

```
E               pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
E               
E               Traceback (most recent call last):
E                 File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/traceback.py", line 376, in extract
E                   result.append(FrameSummary(
E               ResourceWarning: unclosed <socket.socket fd=25, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 57948), raddr=('127.0.0.1', 34709)>
```

### Impact

_No response_

### Additional context

_No response_

chrisguidry:
We're discovering that our test suite requires the larger test runner to
complete in a reasonable time, and that we are starting to see widespread
failures when running on the default runner.  This is due to an unclosed
connection, which is likely due to internal timeouts in the test suite.

This removes the internal user constraint entirely, because it appears that the
suite won't complete successfully on the default runner anymore.  This will give
a better experience for external contributors.  If we see signs of abuse, we can
require approvals for new contributors.

Note: I'm also adjusting the `pydantic` requirement due to how the
`generate-lower-bounds.py` script works.  We had a redundant `>=1.10.0,>=1.7.4`
line (from recent compatibility work) so the lower bound was generating as
`1.7.4` (which we don't support).

Closes #10896

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -37,36 +37,8 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  isinternal:
-    runs-on: ubuntu-latest
-    outputs:
-      status: ${{ steps.isinternal.outputs.status }}
-    steps:
-      - name: Is internal team?
-        id: isinternal
-        run: |
-          # Treat dependabot as an internal user
-          if [[ "${{ github.actor }}" == "dependabot[bot]" ]]; then
-            echo "status=true" >> $GITHUB_OUTPUT
-            exit 0
-          fi
-
-          echo "actor: ${{ github.actor }}"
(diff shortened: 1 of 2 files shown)


cicdw (approved):
LGTM

PR #10949, opened 2023-10-13
Add `last_polled` to deployment

serinamarie:
This PR adds a `last_polled` column to the deployment table so that we can use it at query time to determine deployment status.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/server/database/migrations/versions/postgresql/2023_10_12_224511_bfe653bbf62e_add_last_polled_to_deployment.py
+++ b/src/prefect/server/database/migrations/versions/postgresql/2023_10_12_224511_bfe653bbf62e_add_last_polled_to_deployment.py
@@ -1,0 +1,33 @@
+"""Add `last_polled` to deployment
+
+Revision ID: bfe653bbf62e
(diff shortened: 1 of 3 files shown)


bunchesofdonald (approved):

desertaxle (approved):

src/prefect/server/database/migrations/MIGRATION-NOTES.md
```diff
@@ -8,6 +8,14 @@ Each time a database migration is written, an entry is included here with:
 
 This gives us a history of changes and will create merge conflicts if two migrations are made at once, flagging situations where a branch needs to be updated before merging.
 
+# Add last_polled to deployment table
+SQLite: `f3165ae0a213`
+Postgres: `bfe653bbf62e`
+
+# Make flow_run_id nullable on task_run and log tables
+SQLite: `05ea6f882b1d`
+Postgres: `05ea6f882b1d`
```
serinamarie, line 17: @zangell44 I don't think it's a big deal but noticed that both migrations here have the same revision id when it seems like our convention is having each revision id be unique.

PR #10971, opened 2023-10-16
chore: update docs to remove `api-beta.` references

jamiezieziula:
and replace with `api.prefect.cloud`


discdiver (approved):
Good catch!

PR #10988, opened 2023-10-18
Create the `prefect-client`

urimandujano:
This PR introduces files to dynamically build, test, and publish the `prefect-client` package. `prefect-client` represents a minimal install of `prefect` designed to support running flows against a remote server endpoint **only**. This means it will not support integrations with various systems (K8s, docker, etc).

The high level steps for building the client are:
- clone `prefect`
- delete a lot of source code
- install `prefect-client` using its own `requirements.txt`

The exact steps are codified in the Github workflow `.github/workflows/prefect-client.yaml`. There's also a script `client/build_client.sh` which can be run locally to simulate the build and smoke test process.

There's an accompanying Github workflow which will run the build and a run a smoke-test flow on every PR. On releases, the client will be built and published along with the `prefect` package.


.github/workflows/prefect-client.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: (DEBUG) Check that prefect and prefect-client are installed
+        run: pip list | grep prefect
+
+      - name: Run the smoke test flow again with prefect and prefect-client installed
+        run: python client_flow.py
+        working-directory: ${{ env.TMPDIR }}
+        env:
+          PREFECT_API_KEY: ${{ secrets.PREFECT_CLIENT_SA_API_KEY }}
+          PREFECT_API_URL: "https://api.prefect.cloud/api/accounts/9b649228-0419-40e1-9e0d-44954b5c0ab6/workspaces/96bd3cf8-85c9-4545-9713-b4e3c3e03466" # sandbox, prefect-client workspace
```
chrisguidry, line 75: LOVE IT

client/README.md
```diff
@@ -0,0 +1,29 @@
+# Overview
```
chrisguidry, line 1: Thank you!

client/requirements.txt
```diff
@@ -0,0 +1,31 @@
+anyio >= 3.7.1, < 4.0.0
```
chrisguidry, line 1: I still think we should move this up to the root directory as `client-requirements.txt` and de-dupe the `prefect` `requirements*.txt` files.  Maybe in another PR after this one, but let's not lose that thread.

src/prefect/client/orchestration.py
```diff
@@ -203,7 +203,11 @@ def __init__(
 
         if api_version is None:
             # deferred import to avoid importing the entire server unless needed
-            from prefect.server.api.server import SERVER_API_VERSION
+            try:
+                from prefect.server.api.server import SERVER_API_VERSION
+            except ImportError:
+                # support the case where server components are not available
+                SERVER_API_VERSION = "0.8.4"
```
chrisguidry, line 210: Let's move this up to a common spot that both the client and server can reference, that way we only have one definition for it.

src/prefect/client/schemas/schedules.py
```diff
@@ -18,6 +17,13 @@
 else:
     from pydantic import Field, validator
 
+from prefect._internal.pytz import HAS_PYTZ
+
+if HAS_PYTZ:
+    import pytz
+else:
+    from prefect._internal import pytz
```
chrisguidry, line 25: :heart:

chrisguidry: TBH I'm like :pinching_hand: close to just vendoring EVERY dependency we have LOL

src/prefect/states.py
```diff
@@ -398,9 +398,9 @@ def is_state(obj: Any) -> TypeGuard[State]:
     """
     # We may want to narrow this to client-side state types but for now this provides
     # backwards compatibility
-    from prefect.server.schemas.states import State as State_
+    # from prefect.server.schemas.states import State as State_
```
chrisguidry, line 401:
```suggestion
```

chrisguidry (approved):
A few loose ends to tie up, for me the important one is getting to one definition of the `SERVER_API_VERSION` because that's already an obscure variable and can definitely give us compatibility troubles.

src/prefect/client/orchestration.py
urimandujano: I added a new file `prefect.server.api.constants` that we'll use as the source-of-truth for server and client-side code. Server components import the version directly from this file whereas client components will import from `prefect.client.constants` - which internally tries to import from the server constants. My gut is that helps with separating client/server but let me know if that feels like over-compensation.

src/prefect/client/schemas/schedules.py
urimandujano: You don't know how many times I thought that to myself while working on this 😆

src/prefect/client/orchestration.py
chrisguidry: I think we're still missing the important part for me, that we only have one definition of that API version constant.  Could that constants file just be in a location that both `prefect-client` and `prefect-server` include in their packages?  

One kind of non-obvious thing is that the _server_ could import it from somewhere like `prefect.client.constants` because the server is the full `prefect` package.  Otherwise somewhere in utilities might make sense.  My biggest thing is that we just have one copy of this to update.

client/requirements.txt
urimandujano: I took care of this in this PR - I think it's a great way to make sure that we keep the client+full requirements up to date with each other.

src/prefect/client/orchestration.py
urimandujano: Summary of the sync-up: I learned that `SERVER_API_VERSION` is mostly used in the client's request headers to ensure that we're backwards compatible. So we decided to put it in `prefect.client.constants` and it should be there now!

chrisguidry (approved):
Wooooot

gizmo84:
prefect-client is not published on pypi after the last release:

See: https://pypi.org/project/prefect-client/ and

```
pip install prefect-client
ERROR: Could not find a version that satisfies the requirement prefect-client (from versions: none)
ERROR: No matching distribution found for prefect-client
```

urimandujano:
We had a few issues with our deployment automation yesterday that delayed publishing the client to pypi. We're tackling those issues today and we should have the `prefect-client` available soon. Sorry for the confusion with this release!

urimandujano:
We've just published the `prefect-client` v2.14.0. It should be available for downloads from pypi now, thanks for the patience!

PR #11020, opened 2023-10-24
Add task run tag concurrency slot wait setting

### LINKED ISSUE #10460: Allow users to set delay transition time for Secure Task Concurrency Slots
### First check

### Prefect Version

2.x

Currently, the delay transition time for secure  task concurrency slots is hardcoded to 30 seconds.

https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/orchestration/core_policy.py#L140-L143

Introduce a new prefect configuration/setting that allows the user to set the value for the transition delay.

### Example Use

Example usage similar as documented in https://docs.prefect.io/2.11.4/guides/settings/

### Additional context

Our flows are using mapped tasks that are limited by task concurrency tags. The task by themselves take roughly 1 seconds. However, because the transition delay is hard coded to 30, after each task is "completed", the worker is idle and waits for 30 seconds. It would be great if this delay could be configured by the user.

taljaards:
This PR adds a Prefect setting to configure the amount of time a client must wait to retry securing a concurrency slot from the server if it was unable to.

A use-case is described on [Slack](https://prefect-community.slack.com/archives/CL09KU1K7/p1692608364121739):
If the task runs themselves have a short run time (e.g. mere seconds), the comparatively long 30 second waiting period (currently hardcoded for task run tag-based concurrency limits) can impose undue bottlenecks.

I acknowledge that this is one of several potential solutions to resolve #10460, and it may not be the optimal one. However, it represents a starting point in addressing the issue.

Since this is a Prefect setting (`PREFECT_TASK_RUN_TAG_CONCURRENCY_SLOT_WAIT_SECONDS`), this will apply to all flows run by this client using the active profile.

---

I have not added/changed any tests. I will need some guidance, please, if test changes are required.

### Checklist

For documentation changes:


zhen0:
Thanks for the contribution @taljaards!  I've flagged to the team to make sure you get a review here. 😄

zangell44 (commented):
Thanks for the contribution @taljaards!! Looks great!

This solution works very well for a single open source server. However, we'll need something that can change the delay time dynamically based on a client request.

What do you think about extending this setting and either
1. Having the client send a proposed slot wait seconds AND the server respecting that value in `delay_transition` OR
2. Having the client use `PREFECT_TASK_RUN_TAG_CONCURRENCY_SLOT_WAIT_SECONDS` automatically when given a delay instruction (ignoring any server instruction).

src/prefect/settings.py
```diff
@@ -760,6 +760,12 @@ def default_cloud_ui_url(settings, value):
 This value does not overwrite individually set retry delay seconds
 """
 
+PREFECT_TASK_RUN_TAG_CONCURRENCY_SLOT_WAIT_SECONDS = Setting(int, default=30)
```
zangell44, line 763: Can we add a minimum value to this?

zangell44 (approved):

PR #11021, opened 2023-10-24
ft: use custom versioning to omit dirty version

urimandujano:
This PR makes a change to the `client/setup.py` file to change how we generate version information. We can't use versioneer's autogenerated version because the built directory is in a modified state and leads to messy version numbers (such as `prefect-client-2.14.0+0.g7a95e97f1.dirty.tar.gz`). PyPI won't accept package uploads with versions like that..  This change will parse versioneer's version and pull only the first three digits representing its git tag.

Old versions:
`prefect-client-2.14.0+0.g7a95e97f1.dirty.tar.gz`
`prefect_client-2.14.0+0.g7a95e97f1.dirty-py3-none-any.whl` 

New versions:
`prefect-client-2.14.0.tar.gz`
`prefect_client-2.14.0-py3-none-any.whl`

### Checklist

For documentation changes:


serinamarie:
What happens if there are conflicts between the `setup.cfg` in `prefect-client` vs in `prefect`? Would changes in one file require changes in the other?

urimandujano:
The way the build works is that `prefect-client` specific build files (mainly `setup.py` for now) are in the `client/` directory. At build time the `client/*` files get copied up to the parent directory and any existing files get overridden.

The only reason I can think of where we'd need updates to both files is if we changed something about how we're building `prefect` that would impact `prefect-client` - such as the metadata that we want to use.

chrisguidry (approved):
Love it

PR #11028, opened 2023-10-25
Refining how we validate flow parameters under Pydantic 1 versus 2

chrisguidry:
We have seen a few edge cases popping up where we're trying to use the
pydantic v2 validator to validate v1 models as flow function arguments.  This
leads to the opaque error:

```
TypeError: BaseModel.validate() takes 2 positional arguments but 3 were given
```

Examples:
* https://github.com/PrefectHQ/prefect-airbyte/issues/61
* https://github.com/PrefectHQ/prefect-kubernetes/actions/runs/6628004996/job/18004160832?pr=92

The issue here is that we were trying to use our custom v2 port of the v1
`ValidatedFunction`, without inspecting the types of models involved.  Here, we
are looking at the types of the passed arguments, validating that they aren't
mixing v1 and v2 models, and then choosing the right validation implementation
from there.

### DIFF
--- a/src/prefect/flows.py
+++ b/src/prefect/flows.py
@@ -42,11 +42,14 @@ from prefect.runner.storage import RunnerStorage, create_storage_from_url
 
 if HAS_PYDANTIC_V2:
     import pydantic.v1 as pydantic
+    from pydantic import BaseModel as V2BaseModel
     from pydantic import ValidationError as V2ValidationError
-    from pydantic.v1.decorator import ValidatedFunction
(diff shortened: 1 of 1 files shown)


chrisguidry:
Yep I think this is popping up in our collections libraries because they have flows that are passing blocks around as parameters, and the blocks are ultimately derived from v1 base models (for now).

chrisguidry:
FWIW, I also tested this with an editable local install in our `prefect-kubernetes` collection and it resolved the problems with the test suite there.

PR #11046, opened 2023-10-27
fix `truncate_to` bug that results in the original string being returned

### LINKED ISSUE #11045: `truncated_to` fails to limit the length of the string to the specified value
### First check

### Bug summary

In `truncated_to`, there is logic that removes the middle of the string, shortening it to the desired value by retaining only the beginning and end of the string. However, the function then adds some helper text back into the middle of the string to indicate to users how many characters were omitted. The string then exceeds the desired length and the function returns the original string.

The result is that no truncation is actually being performed by this function when the desired value is less than the length of the string.

### Reproduction

```python3
In [1] from prefect.utilities.text import truncated_to

In [2] original_str = 'b' * 50

In [3] new_str = truncated_to(25, original_str)

In [4] new_str == original_str
Out [4] True
```

### Error

_No response_

### Versions

```Text
Version:             0+untagged.14891.g1998adf
API version:         0.8.4
Python version:      3.10.10
Git commit:          1998adfd
Built:               Fri, Aug 4, 2023 1:14 PM
Profile:             default
Server:
  Database:          sqlite
  SQLite version:    3.39.5
```

### Additional context

_No response_

breid1313:
This closes #11045 , wherein the `truncated_to` function returns the original, un-truncated string. For simplicity sake, I made the decision to have this function truncate text from the end of the string, rather than pulling it from the middle.

### Example

Calling the new function in a python REPL

```python
In [10] original_string = 'something that is too long and needs to be shorted'

In [11] len(original_string)
Out [11] 50

In  [12] new_string = truncated_to(40, original_string)

In  [13] new_string
Out [13] 'something that is too l...[truncated]...'
```

here are the test results as well:

```
% pytest tests/utilities/test_text.py 
======================================================================= test session starts =======================================================================
platform darwin -- Python 3.10.10, pytest-7.4.2, pluggy-1.3.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/brianreid/jellyfish-source/prefect
configfile: setup.cfg
plugins: hypothesis-6.88.1, timeout-2.2.0, respx-0.20.2, flaky-3.7.0, env-1.0.1, flakefinder-1.1.0, cov-4.1.0, Faker-19.11.0, asyncio-0.21.1, anyio-3.7.1, requests-mock-1.11.0, xdist-3.3.1, benchmark-4.0.0
timeout: 60.0s
timeout method: signal
timeout func_only: False
asyncio: mode=auto
collected 4 items                                                                                                                                                 

tests/utilities/test_text.py ....                                                                                                                           [100%]

======================================================================= 4 passed in 21.97s ========================================================================
```

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/utilities/text.py
+++ b/src/prefect/utilities/text.py
@@ -8,12 +8,8 @@ def truncated_to(length: int, value: Optional[str]) -> str:
     if len(value) <= length:
         return value
 
-    half = length // 2
+    help_text = "...[truncated]..."
 
-    beginning = value[:half]
-    end = value[-half:]
-    omitted = len(value) - (len(beginning) + len(end))
+    truncated = value[: length - len(help_text)]
 
-    proposed = f"{beginning}...{omitted} additional characters...{end}"
-
-    return proposed if len(proposed) < len(value) else value
+    return truncated + help_text if len(truncated) < len(help_text) else value[:length]
--- a/src/prefect/utilities/text.py
+++ b/src/prefect/utilities/text.py
@@ -1,5 +1,7 @@
 from typing import Optional
 
+HELP_TEXT = "...[truncated]..."
+
 
 def truncated_to(length: int, value: Optional[str]) -> str:
     if not value:
@@ -8,8 +10,8 @@ def truncated_to(length: int, value: Optional[str]) -> str:
     if len(value) <= length:
         return value
 
-    help_text = "...[truncated]..."
-
-    truncated = value[: length - len(help_text)]
-
-    return truncated + help_text if len(truncated) < len(help_text) else value[:length]
+    return (
+        value[: length - len(HELP_TEXT)] + HELP_TEXT
+        if len(HELP_TEXT) < length
+        else value[:length]
+    )
--- a/tests/utilities/test_text.py
+++ b/tests/utilities/test_text.py
@@ -1,0 +1,29 @@
+from prefect.utilities.text import HELP_TEXT, truncated_to
+
+
+class TestTruncatedTo:
+    def test_empty_value(self):
+        value = None
+        length = 100
+        trunc = truncated_to(length, value)
+        assert trunc == ""
+
+    def test_length_exceeds_value(self):
+        value = "a" * 50
+        length = 51
+        trunc = truncated_to(length, value)
+        assert trunc == value
+
+    def test_normal_truncation(self):
+        value = "a" * 100
+        length = 50
+        trunc = truncated_to(length, value)
+        assert len(trunc) == length
+        assert trunc.endswith(HELP_TEXT)
+
+    def test_truncation_help_text_too_long(self):
+        value = "a" * 100
+        length = 10
+        trunc = truncated_to(length, value)
+        assert len(trunc) == length
+        assert trunc == "a" * length


chrisguidry:
Hi @breid1313, thank you for the contribution!

There may be a misunderstanding of what this `truncated_to` function is designed to do.  We use it when truncating very large state messages, like exception stack traces, where the beginning and the end of the string are very important, and if you do have to compromise on any text, it's better to remove it from the middle.

For example, with a Python stack trace, the beginning of the text has context, and the end of the text is where the error message is.  The intervening layers of function calls aren't as interesting as the outer- and inner-most layers.

`truncated_to` may have been a sub-optimal name for this function, but I think we should keep to to avoid any compatibility issues.

Would you consider creating a _new_ function for this?

chrisguidry (changes requested):
(See my comment above about making this a new function rather than changing the behavior of this function)

breid1313:
@chrisguidry happy to help! I appreciate the super-speedy review.

Thank you for clearing up the intended use -- that makes sense. I would be happy to add a new function that trims string length from the end, but I don't mind re-working this so the logic continues to preserve the beginning and end of the string while addressing the bug that causes truncation to not happen at all.

The latter seems like the better option from my view, but let know what you think.

chrisguidry:
That sounds great!  I think one of the test cases here should be that we never produce a string that's _longer_ than the string you passed in, which is the intent of that clause at the end.  This will come up when the length of the string is close to the length you're requesting for the truncation.

As the original author of this function, I'm pretty sure this is how I intended it to work, where we don't bother truncating if we'd produce a _longer_ string by truncating :D 

```
In [1]: from prefect.utilities.text import truncated_to

In [2]: truncated_to(10, '1234567890')
Out[2]: '1234567890'

In [3]: truncated_to(10, '12345678901')
Out[3]: '12345678901'

In [4]: truncated_to(10, '123456789012')
Out[4]: '123456789012'

In [5]: truncated_to(10, '1234567890123')
Out[5]: '1234567890123'

In [6]: truncated_to(10, '12345678901234')
Out[6]: '12345678901234'

In [7]: truncated_to(10, '123456789012345')
Out[7]: '123456789012345'

In [8]: truncated_to(10, '12345678901234567890')
Out[8]: '12345678901234567890'

In [9]: truncated_to(10, '1234567890123456789012345')
Out[9]: '1234567890123456789012345'

In [10]: truncated_to(10, '123456789012345678901234567890')
Out[10]: '123456789012345678901234567890'

In [11]: truncated_to(10, '1234567890123456789012345678901234567890')
Out[11]: '1234567890123456789012345678901234567890'

In [12]: truncated_to(10, '12345678901234567890123456789012345678901234567890')
Out[12]: '12345...40 additional characters...67890'

```


Maybe a better approach would be to make the behavior configurable, with an argument like:


```
def truncated_to(length: int, value: Optional[str], always: bool = False) -> str:
```

And `always` would always truncate the text, even if it would end up being longer.

breid1313:
Aha, ok that clarification is helpful and thank you for the examples. The intended behavior makes more sense now. I like the idea of an `always` kwarg that defaults to `False`. I can plan to add that next week and adjust the tests accordingly.

I'll keep everything on the same branch & PR so we have this conversation preserved around the change. Thanks again!

chrisguidry:
Awesome, Brian, thank you for the contribution.  Also, thank you for leading with tests, I didn't include any standalone tests here, they are all tangled up with other event-related tests.

chrisguidry:
Hi Brian, let me know if you need any help with this one, happy to help make some adjustments

PR #11075, opened 2023-11-01
Warn if task parameter introspection takes a long time

jakekaplan:
When you apply a `@task` decorator to a function prefect will introspect the parameters you pass so it can track dependencies (like task A's result got passed to task B). This is what we want to show for most cases, but it's a big performance hit when the data you're passing is a really large collection like a `dataframe` or large `dict`, and we try to visit everything inside of it.

`quote()` is an escape utility which tells prefect to ignore introspecting this thing. However this utility is not well known or well documented and this case frequently comes up as there are many data heavy users passing dataframes around.

This pr introduces a helpful warning if task run parameter introspection takes longer than `PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD`, instructing the user to try `quote()`. 
You can disable this warning by setting `PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD=0`

### Example:
```python
import json
from faker import Faker


def make_a_giant_json(N: int):
    Faker.seed(42)
    fake = Faker()
    out = fake.json(
        data_columns={
            "Spec": "@1.0.1",
            "ID": "pyint",
            "x1": "address",
            "x2": "address",
            "x3": "address",
            "x4": "address",
            "x5": "address",
            "x6": "address",
            "x7": "address",
            "x8": "address",
            "x9": "address",
            "x10": "address",
        },
        num_rows=N,
    )
    x = json.loads(out)
    return x


@task
def my_task(big_json):
    pass


@flow
def my_flow(n: int):
    big_json = make_a_giant_json(n)
    my_task(big_json)


if __name__ == '__main__':
    my_flow(n=35000)
```
### Example Output
```
(demo-flows) ➜  demo-flows git:(main) ✗ python my_flow.py
16:14:39.434 | INFO    | prefect.engine - Created flow run 'prudent-bug' for flow 'my-flow'
16:14:39.438 | INFO    | Flow run 'prudent-bug' - View at https://app.prefect.cloud/account/3cf6b38f-5244-474a-9554-302144506e43/workspace/ce8b1412-01b7-4700-a508-8dbd1f43f623/flow-runs/flow-run/76b8071f-e68d-4ed7-9b0e-da7ab033145a
16:14:59.565 | INFO    | Flow run 'prudent-bug' - Created task run 'my_task-0' for task 'my_task'
16:14:59.567 | INFO    | Flow run 'prudent-bug' - Executing 'my_task-0' immediately...
16:15:02.490 | WARNING | Task run 'my_task-0' - Long running task parameter introspection detected. Consider wrapping large task parameters with prefect.utilities.annotations.quote for increased performance, e.g. `my_task(quote(param))`.Disable this message with PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD=0.
16:15:03.569 | INFO    | Task run 'my_task-0' - Finished in state Completed()
16:15:03.761 | INFO    | Flow run 'prudent-bug' - Finished in state Completed('All states completed.')
```

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -165,6 +165,7 @@ from prefect.settings import (
     PREFECT_LOGGING_LOG_PRINTS,
     PREFECT_TASKS_REFRESH_CACHE,
     PREFECT_UI_URL,
+    PREFECT_WARN_LONG_TASK_INTROSPECTION,
 )
 from prefect.states import (
     Paused,
@@ -1642,7 +1643,7 @@ async def orchestrate_task_run(
         result_factory=result_factory,
         log_prints=log_prints,
     )
-
+    start_time = time.time()
     try:
         # Resolve futures in parameters into data
         resolved_parameters = await resolve_inputs(parameters)
@@ -1656,6 +1657,15 @@ async def orchestrate_task_run(
             # if orchestrating a run already in a pending state, force orchestration to
             # update the state name
             force=task_run.state.is_pending(),
+        )
+    end_time = time.time()
+    if PREFECT_WARN_LONG_TASK_INTROSPECTION and (end_time - start_time) > 5:
+        logger.warning(
+            "Long running task parameter introspection detected."
+            "Consider wrapping large task parameters with "
+            "prefect.utilities.annotations.quote() for increased "
+            "performance. Disable this waring message by setting"
+            "PREFECT_WARN_LONG_TASK_INTROSPECTION=False."
         )
 
     # Generate the cache key to attach to proposed states
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -163,9 +163,9 @@ from prefect.results import BaseResult, ResultFactory
 from prefect.settings import (
     PREFECT_DEBUG_MODE,
     PREFECT_LOGGING_LOG_PRINTS,
+    PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD,
     PREFECT_TASKS_REFRESH_CACHE,
     PREFECT_UI_URL,
-    PREFECT_WARN_LONG_TASK_INTROSPECTION,
 )
 from prefect.states import (
     Paused,
@@ -1659,7 +1659,10 @@ async def orchestrate_task_run(
             force=task_run.state.is_pending(),
         )
     end_time = time.time()
-    if PREFECT_WARN_LONG_TASK_INTROSPECTION and (end_time - start_time) > 5:
+    if (
+        PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD.value()
+        and (end_time - start_time) > PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD.value()
+    ):
         logger.warning(
             "Long running task parameter introspection detected."
             "Consider wrapping large task parameters with "
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1667,7 +1667,7 @@ async def orchestrate_task_run(
             "Long running task parameter introspection detected. "
             "Consider wrapping large task parameters with "
             "prefect.utilities.annotations.quote for increased "
-            "performance, e.g. `my_task(quote(param))`."
+            "performance, e.g. `my_task(quote(param))`. "
             "Disable this message with PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD=0."
         )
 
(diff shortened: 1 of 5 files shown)


chrisguidry (approved):
That's a really nice touch to make it configurable/disable-able.

src/prefect/engine.py
```diff
@@ -1642,7 +1643,7 @@ async def orchestrate_task_run(
         result_factory=result_factory,
         log_prints=log_prints,
     )
-
+    start_time = time.time()
```
serinamarie, line 1646: Was there any particular reason for using `time.time()` over `time.perf_counter()` here?

src/prefect/engine.py
```diff
@@ -1657,6 +1658,18 @@ async def orchestrate_task_run(
             # update the state name
             force=task_run.state.is_pending(),
         )
+    end_time = time.time()
```
serinamarie, line 1661: would this be more descriptively named "task_introspection_end_time" and "task_introspection_start_time"?

src/prefect/engine.py
```diff
@@ -1657,6 +1658,18 @@ async def orchestrate_task_run(
...
+        PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD.value()
+        and (end_time - start_time) > PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD.value()
+    ):
+        logger.warning(
+            "Long running task parameter introspection detected. "
+            "Try wrapping large task parameters with "
+            "`prefect.utilities.annotations.quote` for increased "
+            "performance, e.g. `my_task(quote(param))`. "
+            "Disable this message with PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD=0."
+        )
```
serinamarie, line 1672: If someone customized this particular setting but the user who is looking at this log isn't the one who performed that action, does it make sense to relay the threshold that was exceeded?

src/prefect/engine.py
jakekaplan: not really, updated to use `perf_counter`

src/prefect/engine.py
jakekaplan: definitely more descriptive, updated

src/prefect/engine.py
jakekaplan: I think it makes sense to relay this message to whoever is looking at the logs for this run? We could opt to make this a `warning.warn` instead of a log but it less likely someone would see it? I might not totally understand though. 

What do you think I should do here? 😄

serinamarie: ah I just meant something like f"Long running task parameter introspection exceeds threshold of {x} seconds." I don't know if it would be super useful, but I think if we have that info, it might be nice to show it

jakekaplan: ahhh makes total sense! Can add that 👍

jakekaplan: updated
```
17:07:08.083 | WARNING | Task run 'my_task-0' - Task parameter introspection took 3.632 seconds , exceeding `PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD` of 1.0. Try wrapping large task parameters with `prefect.utilities.annotations.quote` for increased performance, e.g. `my_task(quote(param))`. To disable this message set `PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD=0`.
```

src/prefect/settings.py
```diff
@@ -928,6 +928,16 @@ def default_cloud_ui_url(settings, value):
...
 """
 
+PREFECT_TASK_INTROSPECTION_WARN_THRESHOLD = Setting(
+    float,
+    default=10.0,
+)
+"""
+Threshold time in seconds for logging a warning if task parameter introspection 
+exceeds this duration. Defaults to `10.0`. Set to `0` to disable logging the
+warning.
```
serinamarie, line 938: Would it be useful to add on to this with something like "task introspection can be a significant performance hit, especially when dealing with large data structures like dataframes or extensive dictionaries where each element is inspected." Maybe it could guide users who are unsure about why their flows are hanging when they see the warning but are unsure about why it's happening.

src/prefect/engine.py
serinamarie: Looks good 🔥

src/prefect/settings.py
jakekaplan: good idea! added. Also added some more information to the `quote` docstring

serinamarie (approved):
LGTM!

PR #11079, opened 2023-11-01
ft: runs client build workflow on reqs.txt updates

urimandujano:
This PR updates the "Verify prefect-client build" workflow trigger to run when there are changes to `requirements-client.txt`. This should help us verify dependency updates that dependabot submits.

### DIFF
--- a/.github/workflows/prefect-client.yaml
+++ b/.github/workflows/prefect-client.yaml
@@ -5,6 +5,7 @@ on:
     paths:
       - "**/*.py"
       - requirements.txt
+      - requirements-client.txt
       - setup.cfg
   push:
     branches:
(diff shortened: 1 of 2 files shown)


serinamarie (approved):
Makes sense to me! 

And yeah I just merged in a `croniter` major release too 😬

urimandujano:
I was about to say:

> I'm not too worried since breaking changes would probably have been caught in the `prefect` unit tests.

But then I realized that we aren't doing that right now. I just pushed a commit to run the unit tests whenever `requirements-client.txt` changes. We want to include those tests because the full requirements come from combining `requirements.txt` and `requirements-client.txt`.

serinamarie:
> I was about to say:
> 
> > I'm not too worried since breaking changes would probably have been caught in the `prefect` unit tests.
> 
> But then I realized that we aren't doing that right now. I just pushed a commit to run the unit tests whenever `requirements-client.txt` changes. We want to include those tests because the full requirements come from combining `requirements.txt` and `requirements-client.txt`.

Oh I see! Okay, we should probably re-run one of the dependency bump workflows after merging this to ensure that everything is still passing.

Issue #11080, opened 2023-11-02
API functionality parent_flow_run_id has no matching Python object


rebecca-liveeo:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

In the API, `FlowRunFilter` has the option of `parent_flow_run_id` for filtering, however there is no corresponding Python object. This means the `parent_flow_run_id` option can't be used for filtering when working with the Python SDK.

The filter object for `FlowRunFilterParentFlowRunID` exists, but isn't currently used anywhere: https://github.com/PrefectHQ/prefect/blob/e61b08a0d1e8c8ab90dad384a814c0f57bb98374/src/prefect/client/schemas/filters.py#L243C7-L243C35

API ref: https://docs.prefect.io/2.14.2/api-ref/rest-api-reference/#tag/Flow-Runs/operation/read_flow_runs_flow_runs_filter_post
![image](https://github.com/PrefectHQ/prefect/assets/143079214/79ffc305-e2ae-4c76-be27-13a0a3c8217a)

### Reproduction

```python3
from prefect import get_client
import asyncio
from prefect.client.schemas.filters import FlowRunFilter, FlowRunFilterId

async def get_child_runs():
    async with get_client() as client:
        run_filter = FlowRunFilter(parent_task_run_id=FlowRunFilterId(any_=[<parent_task_run_id>]))
        # run_filter = FlowRunFilter(parent_flow_run_id=FlowRunFilterId(any_=[<parent flow run id>]))
        flows = await client.read_flow_runs(flow_run_filter=run_filter)

        print(len(flows))


asyncio.run(get_child_runs())

---

The first filter will work and return all flows with the given parent *task* run IDs. The second filter will be ignored and return all flows.
```

### Versions

```Text
Version:             2.14.1
API version:         0.8.4
Python version:      3.10.12
Git commit:          42812c08
Built:               Wed, Oct 25, 2023 3:38 PM
OS/Arch:             linux/x86_64
Profile:             default
Server type:         cloud
```

urimandujano:
Thanks for the report @rebecca-liveeo, it looks like this is an issue in our client schemas. I noticed that if we print the second filter we're dropping the `parent_flow_run_id` kwarg:

```python
[...]

PR #11089, opened 2023-11-03
Add `parent_flow_run_id` to client schema

oakbramble:
Closes: https://github.com/PrefectHQ/prefect/issues/11080

In the API, `FlowRunFilter` has the option of `parent_flow_run_id` for filtering, however there was no corresponding Python object. 

`parent_flow_run_id` has now been added to the client schema.

### Example
```
import asyncio
from prefect.client.schemas.filters import FlowRunFilter, FlowRunFilterId

async def get_child_runs():
        run_filter = FlowRunFilter(parent_flow_run_id=FlowRunFilterId(any_=[<parent flow run id>]))
        flows = await client.read_flow_runs(flow_run_filter=run_filter)

        print(len(flows))


asyncio.run(get_child_runs())
```
As seen in the linked issue, previously this `run_filter` would've been ignored; it should now filter and return all the flows which have the given flow run ID as parent.

### Checklist

For documentation changes:


src/prefect/client/schemas/filters.py
```diff
@@ -305,6 +305,9 @@ class FlowRunFilter(PrefectBaseModel, OperatorMixin):
         default=None,
         description="Filter criteria for `FlowRun.next_scheduled_start_time`",
     )
+    parent_flow_run_id: Optional[FlowRunFilterParentFlowRunId] = Field(
```
urimandujano, line 308: This looks exactly right to me!

tests/server/models/test_flow_runs.py
```diff
@@ -830,6 +830,37 @@ async def test_read_flows_filters_by_deployment_id(self, flow, session):
         )
         assert {res.id for res in result} == {flow_run_1.id}
 
+    async def test_read_flow_runs_filters_by_parent_flow_run_ids(self, flow, session):
+        flow_run_1 = await models.flow_runs.create_flow_run(
+            session=session,
+            flow_run=schemas.core.FlowRun(
```
urimandujano, line 836: This `schemas` module is actually the server-side schemas module (`from prefect.server import models, schemas`) so lets do two things:

- create a new test file (it's ok if it's a mirror i.e.`tests/client/api/test_flow_runs.py` and put these tests in there. Since this is a new test file we'll need to redefine a bunch of testing setup  
- update the import(s) so we can use and test from the `prefect.client.*` modules

I think the best bet would be to take a look at this test `tests/server/api/test_flow_runs.py::test_read_subflow_runs`. It's testing the server schema and I think we'll want to basically test the same way, only using the client schema instead.

urimandujano (commented):
This looks good to me, the only problem is the tests are currently testing the server schema whereas this is modifying the client-side filters. We'll just need to update the tests to make sure we're testing the right thing.

tests/server/models/test_flow_runs.py
oakbramble: @urimandujano Thanks for the feedback :relaxed: 

I have made `tests/client/api/test_flow_runs.py` and put in the `tests/server/api/test_flow_runs.py::test_read_subflow_runs` tests, changing the filter query to use the client. When I test locally, they are passing.

In the previous pipeline, the smoke test flow failed, but this is also working just fine for me locally now. 

Thanks for your assistance, the feedback is really helpful.

tests/client/api/test_flow_runs.py
```diff
@@ -0,0 +1,119 @@
+from uuid import uuid4
+
+from prefect._internal.pydantic import HAS_PYDANTIC_V2
+
+if HAS_PYDANTIC_V2:
+    pass
+else:
+    pass
```
urimandujano, line 8: We can probably just drop these imports/checks.

urimandujano:
Everything is looking good here, feel free to mark this as ready and if the tests are still passing we should be good to go.

oakbramble:
> Everything is looking good here, feel free to mark this as ready and if the tests are still passing we should be good to go.

Hi @urimandujano the tests didn't pass for postgres, but I have them working now. Hopefully all is good :crossed_fingers: Do I need to do something to signal I'm ready for the tests to be run again, or do you go through and set off the pipelines at regular intervals anyway?

urimandujano:
I missed this last week but I've just kicked off the CI run. I wish it happened automatically after the first time but I think the workflow only runs after we kick it off. Usually just an @ at whoever has been assigned an issue/been in communication should do the trick!

PR #11118, opened 2023-11-07
Add warning to push pool guide

cicdw:
This is probably the easiest stumbling block to encounter, so adding a warning about image architecture.


discdiver (approved):
Good call.

PR #11138, opened 2023-11-09
Use an unstarted flow run's expected start time for the Graph start time

chrisguidry:
We're seeing a very small error rate of trying to set `Graph.start_time` to
`None` when reading a flow run graph, and it is due to unstarted flow runs
briefly having no `start_time` (just like we fixed with task runs).  This
affects the whole `Graph`, not the `Node`s, but the idea is the same: use the
`expected_start_time` if there's no `start_time`.


bunchesofdonald (approved):
Beautiful

zzstoatzz (approved):
lgtm!

PR #11154, opened 2023-11-09
Create specific CI actions for Pydantic v1

serinamarie:
As it turns out, we removed the upper bound pin that we had on Pydantic, so now our unit tests by default run with V2. We can repurpose that workflow to force install Pydantic V1 to ensure we remain compatible.

### Checklist


urimandujano (approved):
Looks good, thanks for catching and fixing this hole in our tests!

chrisguidry (commented):
Enthusiastically approve

PR #11218, opened 2023-11-20
Fix subflow filters not working and not persisting

pleek91:
# Description
When viewing subflows of a flow run filtering by state was not working. Also the filters were not persisting to the url which meant changing tabs or refreshing the page would reset the filters.

Needs: PrefectHQ/prefect-ui-library#1882


znicholasbrown (approved):
test failures look to be unrelated

PR #11284, opened 2023-11-29
Fix broken task pause/resume behavior

abrookins:
Currently, if you submit a task and then pause a flow without using `reschedule=True`, the flow run and its task runs will become stuck in Paused.

### What's going wrong?
The problem is that pausing flow runs pauses tasks, but if that happens near the beginning of task orchestration when submitting tasks to a task runner with `submit()`, the engine isn't waiting for the paused tasks to become pending again. This creates issues, in particular when we're doing in-process pauses. For example, in the demo flow linked in this PR:

1. The flow submits tasks.
2. The flow pauses itself. Task submission is async, so the pause happens immediately.
3. The submitted tasks begin orchestration, but their runs transition to Paused because the flow run is Paused (a paused flow run pauses tasks).
4. Someone resumes the flow run in the UI.
5. The flow run transitions to Running and the flow continues execution until it tries to get the result of the submitted task.
6. Discovering that the task run is paused, the engine assumes the flow run should be paused too, so it pauses the flow run.
7. The flow exits and the flow run stays paused. Resuming it again triggers the same sequence of events.

And a **side quest**: when setting a task run to Paused during orchestration if the flow run was Paused, we were previously failing to preserve the flow run's Paused parameters, e.g. the timeout. That is fixed now.

### Proposed fix
To fix this, when we orchestrate a task run, if our state submission to Running is rejected because the task run is in a Paused state (but not suspended, AKA, `reschedule=True`), we'll enter a loop. The loop will try to propose a Running state again and see if it works -- this ensures that orchestration can abort if we exceed the pause timeout. There is also no need to manually transition these Paused tasks to Pending somewhere else because the orchestration will allow the transition to Running as soon as the flow is also Running. If the task successfully transitions to Running, we continue orchestration.

### Example

Demo flow:
```python
import asyncio
import random

from prefect import flow, task, pause_flow_run


@task
async def get_number() -> int:
    return random.randint(1, 2)


@flow()
async def sad_flow():
    future = await get_number.submit()
    await pause_flow_run()
    print(f"Here's a sad number: {await future.result()}")


if __name__ == "__main__":
    asyncio.run(sad_flow())

```

### Checklist

For documentation changes:


src/prefect/engine.py
```diff
@@ -1680,10 +1681,18 @@ async def begin_task_run(
...
 
         except Pause:
+            # A pause signal here should mean the flow run suspended, so we
+            # should do the same. We'll look up the flow run's pause state to
+            # try and reuse it, so we capture any data like timeouts.
+            flow_run = await client.read_flow_run(task_run.flow_run_id)
+            if flow_run.state and flow_run.state.is_paused():
+                state = flow_run.state
+            else:
+                state = Paused(reschedule=True)
```
bunchesofdonald, line 1691:
```suggestion
                state = Suspended()
```

src/prefect/engine.py
```diff
@@ -1797,13 +1806,74 @@ async def orchestrate_task_run(
...
+        state = exc.state if exc.state else Paused()
+
+        # If a flow submits tasks and then pauses, we may reach this point due
+        # to concurrency timing because the tasks will try to transition after
+        # the flow run has paused. Orchestration will send back a Paused state
+        # for the task runs.
+        if state.state_details.pause_reschedule:
+            # If we're being asked to pause and reschedule, we should exit the
+            # task and expect to be resumed later.
+            raise
```
bunchesofdonald, line 1831: Should we `raise Pause()` here?

tests/test_engine.py
```diff
@@ -760,6 +763,249 @@ def foo():
                 prefect_client, State(type=StateType.RUNNING), task_run_id=task_run.id
             )
 
+    async def test_raises_on_pause_with_reschedule(
+        self, monkeypatch, prefect_client, mock_anyio_sleep, flow_run, result_factory
+    ):
+        paused_state = Paused(reschedule=True)
```
bunchesofdonald, line 769:
```suggestion
        paused_state = Suspended()
```

bunchesofdonald (approved):
I suggested using `Suspended` wherever you were setting `reschedule=True`, just to keep with the new nomenclature, but looks great otherwise.

PR #11291, opened 2023-11-30
Create `suspend_flow_run` method to suspend a flow run.

### LINKED ISSUE #11271: Create a `suspend_flow_run` method to suspend a flow run.
### First check

### Prefect Version

2.x

`pause_flow_run` currently takes an optional `flow_run_id`, and a `reschedule` flag.

Setting `reschedule` to `True` will result in the flow run's execution being suspended. resuming it will cause it to go back into a Scheduled state which can then be picked up by a worker where the flow run will be executed again, but it will skip the pause.

Setting `reschedule` to `False` will result in the flow run's execution being blocked. Resuming it will cause code execution to continue from where the pause occurred.

This is confusing as it's conflating two different ideas about what pausing means. There is pausing that blocks code execution and there is pausing that suspends code execution. 

To make this difference more obvious we should create a separate `suspend_flow_run` method that takes a flow run id, and puts that flow run into a Paused state with a name of `Suspended`. This would be equivalent to calling `pause_flow_run` with `reschedule=True`.

### Example Use

_No response_

### Additional context

_No response_

bunchesofdonald:
This creates the `suspend_flow_run` method. This method similar to calling `pause_flow_run` with `reschedule=True`, but disambiguates the concept of pause vs suspend. When a flow run is paused code execution is blocked and the process continues to run, when a flow run is suspended, code execution is stopped as is the process.

Closes #11271 

### Example
```python
from prefect import flow, task, suspend_flow_run


@task
async def get_tree(n) -> str:
    return ["🌲", "🌳", "🌴"][n]


@task
async def display_tree(tree: str) -> None:
    print(tree)


@flow
async def happy(num_trees: int = 3):
    trees = []
    for i in range(num_trees):
        # tree_n = random.randint(0, 2)
        state = await get_tree.submit(i)
        trees.append(await state.result())

    await suspend_flow_run()

    for tree in trees:
        await display_tree(tree)


if __name__ == "__main__":
    happy.serve(name="happy-trees")
```

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/__init__.py
+++ b/src/prefect/__init__.py
@@ -44,7 +44,7 @@ from prefect.context import tags
 from prefect.manifests import Manifest
 from prefect.utilities.annotations import unmapped, allow_failure
 from prefect.results import BaseResult
-from prefect.engine import pause_flow_run, resume_flow_run
+from prefect.engine import pause_flow_run, resume_flow_run, suspend_flow_run
 from prefect.client.orchestration import get_client, PrefectClient
 from prefect.client.cloud import get_cloud_client, CloudClient
 import prefect.variables
@@ -172,4 +172,7 @@ __all__ = [
     "Runner",
     "serve",
     "deploy",
+    "pause_flow_run",
+    "resume_flow_run",
+    "suspend_flow_run",
 ]
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -172,6 +172,7 @@ from prefect.states import (
     Pending,
     Running,
     State,
+    Suspended,
     exception_to_crashed_state,
     exception_to_failed_state,
     get_state_exception,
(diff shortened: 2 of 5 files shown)


desertaxle (approved):
LGTM!

One thought: Do we want to add a log to `pause_flow_run` when `reschedule` is `True` to prompt the user to use `suspsend_flow_run` instead and/or update the docstring for `reschedule`?

bunchesofdonald:
@desertaxle Yep! I'm working on a few follow-up PRs, one of which is to deprecate the `flow_run_id` and `reschedule` parameters of `pause_flow_run`, which will call out that they should use `suspend_flow_run` instead.

urimandujano (approved):
Nice! I'm VERY curious about the `resume` side of this story :)

bunchesofdonald:
Resume is actually already in place. The caveat is that in order for `suspend` to be generally useful you need to use tasks / task caching, otherwise we're just re-running the entire flow again.

PR #11376, opened 2023-12-11
Integrate flow run input and schema/response mechanics into pause/suspend

### LINKED ISSUE #11339: Build out client-side primitives for getting, storing, and validating flow run input
Now that we have modeling and an CRUD API to store key/value data associated with a flow run we should build out the client-side primitives to get, store, and validate that data.

bunchesofdonald:
This integrates flow run input into pause/suspend allowing a flow run to pause or suspend and receive structured and validated input, it does this by:

- Creating `RunInput` a sub-classable pydantic model that a user can subclass and add their own fields and validation.
- Adding `wait_for_input` to `pause_flow_run` and `suspend_flow_run`. When either method is provided a subclass of `RunInput` it will pause/suspend,'save` and the schema of the class to the flow run input API. When resuming it will then try to grab a response from the flow run input API and instantiate the class with that data.

Closes #11339 

### Example
```python
import random
from prefect import flow, get_run_logger, pause_flow_run

from prefect.input import RunInput


class UserGuess(RunInput):
    guess: int


@flow
async def guess_the_number():
    the_number = random.randint(1, 25)

    logger = get_run_logger()

    logger.info("Guess the number between 1 and 25. You have 5 guesses.")

    for i in range(5):
        input = await pause_flow_run(
            wait_for_input=UserGuess.with_initial_data(guess=0)
        )
        if input.guess == the_number:
            logger.info("You guessed it!")
            break
        elif input.guess < the_number:
            logger.info("Too low!")
        elif input.guess > the_number:
            logger.info("Too high!")

    logger.error(f"You failed to guess the number. It was {the_number}")


if __name__ == "__main__":
    guess_the_number.serve(name="guess-the-number")
```

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -1053,9 +1053,8 @@ async def _in_process_pause(
         timeout_seconds=timeout, reschedule=reschedule, pause_key=pause_key
     )
 
-    run_input_keyset = keyset_from_paused_state(proposed_state)
-
     if wait_for_input:
+        run_input_keyset = keyset_from_paused_state(proposed_state)
         proposed_state.state_details.run_input_keyset = run_input_keyset
 
     try:
@@ -1202,9 +1201,9 @@ async def suspend_flow_run(
         pause_key = key or str(uuid4())
 
     proposed_state = Suspended(timeout_seconds=timeout, pause_key=pause_key)
-    run_input_keyset = keyset_from_paused_state(proposed_state)
 
     if wait_for_input:
+        run_input_keyset = keyset_from_paused_state(proposed_state)
         proposed_state.state_details.run_input_keyset = run_input_keyset
 
     try:
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -100,6 +100,7 @@ from typing import (
     List,
     Optional,
     Set,
+    Type,
     TypeVar,
     Union,
 )
@@ -114,6 +115,7 @@ import prefect
 import prefect.context
 import prefect.plugins
 from prefect._internal.compatibility.deprecated import deprecated_parameter
+from prefect._internal.compatibility.experimental import experimental_parameter
 from prefect._internal.concurrency.api import create_call, from_async, from_sync
 from prefect._internal.concurrency.calls import get_current_call
 from prefect._internal.concurrency.cancellation import CancelledError, get_deadline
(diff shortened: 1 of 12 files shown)


src/prefect/engine.py
```diff
@@ -990,8 +997,12 @@ async def pause_flow_run(
             the number of pauses observed by the flow so far, and prevents pauses that
             use the "reschedule" option from running the same pause twice. A custom key
             can be supplied for custom pausing behavior.
+        wait_
```
abrookins, line 1000: Lil typo

src/prefect/engine.py
```diff
@@ -1026,12 +1045,19 @@ async def _in_process_pause(
 
     logger.info("Pausing flow, execution will continue when this flow run is resumed.")
 
+    proposed_state = Paused(
+        timeout_seconds=timeout, reschedule=reschedule, pause_key=pause_key
+    )
+
+    run_input_keyset = keyset_from_paused_state(proposed_state)
```
abrookins, line 1052: Do we need this if we aren't waiting for input?

src/prefect/engine.py
```diff
@@ -1048,6 +1077,9 @@ async def _in_process_pause(
             f"Flow run cannot be paused. Received non-paused state from API: {state}"
         )
 
+    if wait_for_input:
+        await wait_for_input.save(run_input_keyset)
```
abrookins, line 1081: This confused me and I'm not not sure why. I haven't reached the definition of this method yet.

abrookins: Is this so overridden defaults for title and stuff are available when the user is filling out the form?

src/prefect/engine.py
bunchesofdonald: Ah, actually the beginnings of a docstring, thanks!

src/prefect/engine.py
bunchesofdonald: Nope.

src/prefect/engine.py
bunchesofdonald: Right, this is where the schema is actually saved and stored in the input api. It happens down here so that we're not doing it unless our state change is actually accepted. I'll add some commentary here.

abrookins (approved):
Looks good to me! 👍

PR #11395, opened 2023-12-12
client-side solution POC

### LINKED ISSUE #10749: Restart after modifying task state re-runs the failed task
### First check

- [X] I refreshed the page and this issue still occurred.
- [X] I checked if this issue was specific to the browser I was using by testing with a different browser.

### Bug summary

Our team designs prefect flows in which tasks sometimes fail. When this happens, we perform some action (unrelated to prefect) and then set the state of the task to "completed", and continue running the flow. Our expectation is that the flow would continue from whichever task is after the failed task. However, we have noticed in Prefect 2 that retrying the flow after editing the failed tasks status to completed re-runs the failed task, and does not start from the next one. 

When we were using Prefect 1, I believe the behavior was what we expected, meaning that restarting a failed flow after manually editing the task's status in the UI would start the flow from the next task, and not re-run the failed task. So I am not sure if the different behavior we are experiencing with prefect 2 is a bug, or if there are other steps we need to take to replicate the behavior we expect. 

### Reproduction

1. Create a flow with three tasks, make the second task fail by always throwing an error or something. The flow should have persist_results=True. We also have remote_storage configured. 
```
from prefect.deployments.deployments import Deployment
from prefect.task_runners import SequentialTaskRunner
import asyncio

@task(persist_result=True)
def task_1():
    print("success!")

@task(persist_result=True)
def task_2():
    raise RuntimeError

@task(persist_result=True)
def task_3():
    print("success!")

@flow(task_runner=SequentialTaskRunner(), persist_result=True)
def demo_restart_issue():
    task_1()
    task_2()
    task_3()

async def deploy():
    local_deployment = await Deployment.build_from_flow(
        flow=demo_restart_issue,
        name=f"deployment_demo_restart_issue",
    )
    await local_deployment.apply()


if __name__ == "__main__": 
    asyncio.run(deploy())
```

3. Run the flow on the prefect UI. The flow will fail due to the second task failing.
4. Go to the second task in the UI, and manually set its status to "Completed". 
Screen Shot 2023-09-18 at 9 09 31 AM

5. Restart the flow by clicking "Retry"
6. The flow will fail again because it will re-run the second task, even though we set its state to completed. 

### Error

Prefect UI displaying initial task and flow run failure: 
image

This is how I modified the state of task run 2:
Screen Shot 2023-09-18 at 9 43 29 AM

The flow run after modifying the state of the task, but before retrying the flow:
Screen Shot 2023-09-18 at 9 44 01 AM

Prefect UI after clicking "Retry". I would have instead inspected the flow to restart from **after** task 2, and run task 3. Instead, it retries task 2, which fails again.
Screen Shot 2023-09-18 at 9 46 17 AM

### Browers

- [X] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.13.0
API version:         0.8.4
Python version:      3.10.12
Git commit:          66577cc6
Built:               Thu, Sep 7, 2023 3:52 PM
Profile:             ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.42.0
```
```


### Additional context

_No response_

abrookins:
POC of an alternative solution to fix #10749 by persisting a None result if we're trying to rerun a task that's in a completed state, when we expect a persisted result and find no result data.

### Checklist

For documentation changes:


abrookins:
Combined this with #11385

PR #11398, opened 2023-12-13
added ten second connectivity timeout

liamandaa:
This PR addresses issue 11290 and times out the process after ten seconds without connecting to the network. Instead, an exception is raised that prompts the user to check connectivity on port 4200. 
### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/cli/deployment.py
+++ b/src/prefect/cli/deployment.py
@@ -8,6 +8,7 @@ import warnings
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
+import signal
 
 import pendulum
 import typer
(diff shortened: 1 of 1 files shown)


src/prefect/cli/deployment.py
```diff
@@ -1052,47 +1053,53 @@ async def build(
...
+        elif rrule:
+            try:
+                schedule = RRuleSchedule(**json.loads(rrule))
+                if timezone:
+                    # override timezone if specified via CLI argument
+                    schedule.timezone = timezone
+            except json.JSONDecodeError:
+                schedule = RRuleSchedule(rrule=rrule, timezone=timezone)
+    except Exception:
+        print("Network not connected, please check connectivity on port 4200")
```
urimandujano, line 1102: I don't think we'll want to handle the `Exception` raised this way because this will capture any Exception in the code block, ignore its details, and always print the error as network related. 

I suggest we handle this error with a more narrow exception type and use our regular exception printing mechanism:
```python
    try:
        ...
    except MyNetworkingException as exc:
        exit_with_error(exc)
```

src/prefect/cli/deployment.py
```diff
@@ -1309,3 +1316,6 @@ def cast_value(value: str) -> Any:
             )
 
     return parsed
+
+def timeout_handler(signum, frame):
+    raise Exception("Network not connected, please check connectivity on port 4200")
```
urimandujano, line 1321: I think this can be improved using a more specific `Exception` class. It's also not guaranteed that we're trying to connect on port 4200 -- we can include information on the address we connected to instead (using the `PREFECT_API_URL` variable)

urimandujano (changes requested):
Hi @liamandaa, thanks for taking up this issue! Please see my comments for ideas on how we can get this ready to merge.

serinamarie:
Hi @liamandaa, if you decide to work on this again feel free to reopen!

PR #11406, opened 2023-12-13
Allow users to select relative and fixed date ranges to filter the dashboard

pleek91:
# Description
Implements a new date range select on the dashboard. Now users can select both past and future relative date ranges like "Past 1 week" or "Next 1 hour" but users can also select specific date and time ranges. Previously the dashboard was limited to viewing the last 8 hours, 24 hours, or 1 week. 

| Before | After |
|--------|--------|
| image | image |


znicholasbrown (approved):
Provisional approval pending package update

Issue #10865, opened 2023-10-02
Allow creation and management of global concurrency limits from the CLI.


bunchesofdonald:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar request and didn't find it.
- [X] I searched the Prefect documentation for this feature.

### Prefect Version

2.x

### Describe the current behavior

Currently the UI is the only place where you can create or edit a global concurrency limit. 

### Describe the proposed behavior

We should add support to the CLI to allow a user to create/edit/delete a global concurrency limit from the CLI.

PR #11415, opened 2023-12-14
Prefect client global concurrency limit functions / tests

Zyntogz:
added functions to interact with the global concurrency limit api using a prefect client (similar to concurrency limit v1 api requests)

added tests for new methods

closes https://github.com/PrefectHQ/prefect/issues/11416

may be the first step before solving https://github.com/PrefectHQ/prefect/issues/10865

Sorry, maybe i am blind but i cannot find a way to add the "enhancement" label to this PR

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -2762,7 +2762,7 @@ class PrefectClient:
 
     async def read_global_concurrency_limit_by_name(
         self, name: str
-    ) -> dict[str, object]:
+    ) -> Dict[str, object]:
         try:
             response = await self._client.get(f"/v2/concurrency_limits/{name}")
             return response.json()
@@ -2774,7 +2774,7 @@ class PrefectClient:
 
     async def read_global_concurrency_limits(
         self, limit: int = 10, offset: int = 0
-    ) -> list[dict[str, object]]:
+    ) -> List[Dict[str, object]]:
         response = await self._client.post(
             "/v2/concurrency_limits/filter",
             json={
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -49,6 +49,8 @@ from prefect.client.schemas.actions import (
     FlowRunCreate,
     FlowRunNotificationPolicyCreate,
     FlowRunUpdate,
+    GlobalConcurrencyLimitCreate,
+    GlobalConcurrencyLimitUpdate,
     LogCreate,
     TaskRunCreate,
     TaskRunUpdate,
@@ -2721,6 +2723,68 @@ class PrefectClient:
                 "occupancy_seconds": occupancy_seconds,
             },
         )
+
+    async def create_global_concurrency_limit(
+        self, concurrency_limit: GlobalConcurrencyLimitCreate
+    ) -> UUID:
+        response = await self._client.post(
+            "/v2/concurrency_limits/",
+            json=concurrency_limit.dict(json_compatible=True, exclude_unset=True),
+        )
+        return UUID(response.json()["id"])
+
+    async def update_global_concurrency_limit(
+        self, name: str, concurrency_limit: GlobalConcurrencyLimitUpdate
+    ) -> httpx.Response:
+        concurrency_limit_dict = concurrency_limit.dict(
+            json_compatible=True, exclude_unset=True
+        )
+        try:
+            response = await self._client.patch(
+                f"/v2/concurrency_limits/{name}", json=concurrency_limit_dict
+            )
+            return response
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == status.HTTP_404_NOT_FOUND:
+                raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
+            else:
+                raise
+
+    async def delete_global_concurrency_limit_by_name(
+        self, name: str
+    ) -> httpx.Response:
+        try:
+            response = await self._client.delete(f"/v2/concurrency_limits/{name}")
+            return response
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == status.HTTP_404_NOT_FOUND:
+                raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
+            else:
+                raise
+
+    async def read_global_concurrency_limit_by_name(
+        self, name: str
+    ) -> dict[str, object]:
+        try:
+            response = await self._client.get(f"/v2/concurrency_limits/{name}")
+            return response.json()
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == status.HTTP_404_NOT_FOUND:
+                raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
+            else:
+                raise
+
+    async def read_global_concurrency_limits(
+        self, limit: int = 10, offset: int = 0
+    ) -> list[dict[str, object]]:
+        response = await self._client.post(
+            "/v2/concurrency_limits/filter",
+            json={
+                "limit": limit,
+                "offset": offset,
+            },
+        )
+        return response.json()
 
     async def create_flow_run_input(
         self, flow_run_id: UUID, key: str, value: str, sender: Optional[str] = None
--- a/src/prefect/client/orchestration.py
+++ b/src/prefect/client/orchestration.py
@@ -2736,12 +2736,10 @@ class PrefectClient:
     async def update_global_concurrency_limit(
         self, name: str, concurrency_limit: GlobalConcurrencyLimitUpdate
     ) -> httpx.Response:
-        concurrency_limit_dict = concurrency_limit.dict(
-            json_compatible=True, exclude_unset=True
-        )
         try:
             response = await self._client.patch(
-                f"/v2/concurrency_limits/{name}", json=concurrency_limit_dict
+                f"/v2/concurrency_limits/{name}",
+                json=concurrency_limit.dict(json_compatible=True, exclude_unset=True),
             )
             return response
         except httpx.HTTPStatusError as e:
--- a/src/prefect/client/schemas/actions.py
+++ b/src/prefect/client/schemas/actions.py
@@ -623,3 +623,25 @@ class VariableUpdate(ActionBaseModel):
 
     # validators
     _validate_name_format = validator("name", allow_reuse=True)(validate_variable_name)
+
+
+@copy_model_fields
+class GlobalConcurrencyLimitCreate(ActionBaseModel):
+    """Data used by the Prefect REST API to create a global concurrency limit."""
+
+    name: str = FieldFrom(objects.GlobalConcurrencyLimit)
+    limit: int = FieldFrom(objects.GlobalConcurrencyLimit)
+    active: Optional[bool] = FieldFrom(objects.GlobalConcurrencyLimit)
+    active_slots: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+    slot_decay_per_second: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+
+
+@copy_model_fields
+class GlobalConcurrencyLimitUpdate(ActionBaseModel):
+    """Data used by the Prefect REST API to update a global concurrency limit."""
+
+    name: Optional[str] = FieldFrom(objects.GlobalConcurrencyLimit)
+    limit: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+    active: Optional[bool] = FieldFrom(objects.GlobalConcurrencyLimit)
+    active_slots: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+    slot_decay_per_second: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
--- a/src/prefect/client/schemas/objects.py
+++ b/src/prefect/client/schemas/objects.py
@@ -1561,3 +1561,30 @@ class FlowRunInput(ObjectBaseModel):
     def validate_name_characters(cls, v):
         raise_on_name_alphanumeric_dashes_only(v)
         return v
+
+
+class GlobalConcurrencyLimit(ObjectBaseModel):
+    """An ORM representation of a global concurrency limit"""
+
+    name: str = Field(description="The name of the global concurrency limit.")
+    limit: int = Field(
+        description=(
+            "The maximum number of slots that can be occupied on this concurrency"
+            " limit."
+        )
+    )
+    active: Optional[bool] = Field(
+        default=True,
+        description="Whether or not the concurrency limit is in an active state.",
+    )
+    active_slots: Optional[int] = Field(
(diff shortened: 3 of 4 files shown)


src/prefect/client/orchestration.py
```diff
@@ -2694,6 +2694,89 @@ async def release_concurrency_slots(
             },
         )
 
+    async def create_concurrency_limit_global(
```
bunchesofdonald, line 2697: Could we rename these to be have `global` before `concurrency`? `create_global_concurrency_limit`, `update_global_concurrency_limit`, etc. I think that would make it more intuitive.

src/prefect/client/orchestration.py
```diff
@@ -2694,6 +2694,89 @@ async def release_concurrency_slots(
...
+        self,
+        name: str,
+        limit: int,
+        active: bool = True,
+        active_slots: int = 0,
+        slot_decay_per_second: int = 0,
+    ) -> UUID:
+        response = await self._client.post(
+            "/v2/concurrency_limits/",
+            json={
```
bunchesofdonald, line 2707: Generally the way we handle these create/update methods is to have a schema in `actions` that has the fields we're going to send to the server, you can see examples of the server side of this here: https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/schemas/actions.py#L438-L459

Essentially you'd want to create the same objects in `client/schemas/actions.py` and `client/schemas/objects.py` and have that be the argument you pass to the create/update/etc. method:

```python
from prefect.client.schemas.actions imoport (
    GlobalConcurrencyLimitCreate
    GlobalConcurrencyLimitUpdate
)

async def create_global_concurrency_limit(
    self,
    concurrency_limit: GlobalConcurrencyLimitCreate,
) -> UUID:
    response = await self._client.post(
        "/v2/concurrency_limits/",
        json=concurrency_limit.dict(json_compatible=True, exclude_unset=True),
    )
    return UUID(response.json()["id"])

async def update_global_concurrency_limit(
    self,
    concurrency_limit: GlobalConcurrencyLimitUpdate,
):
    ...
```

There are a few reasons for this:

1.The Pydantic model (which is what those actions are) allows you to set fields as optional, so those fields can just be skipped entirely when a user is instantiating the action, so the user only needs to specify what's needed.
2. The actions models validate typing, and can do deeper validation as well, say on the name of the concurrency limit. It's much nicer to have the validation happen before the server call, in general.
3. `exclude_unset`, this is important for the `patch` call below, if you leave `name` unset in the action the value won't get sent to the sever and the server won't change the name of the limit because it didn't get a value for the name.

bunchesofdonald (changes requested):
Thanks for the PR @Zyntogz! This looks pretty close, the only change we'd need to make is around how we're passing the data into the create/update methods, but this is great otherwise. Nice tests!

src/prefect/client/orchestration.py
Zyntogz: sure, i'll change this :)

src/prefect/client/orchestration.py
Zyntogz: OK I will look into it. Thanks for your explanations, these will help a lot!

Zyntogz:
Hey, adapted your MR suggestions and made the tests run again accordingly :)

Maybe a PoV from a prefect user (normally):
Normally, when using client libraries from other projects I am used to using their provided methods by providing elementary data types or sometimes more complex ones (dicts, lists, ...) solely. This has the advantage that the end-user does not need to know project-specific stuff and can stay on a relatively low level of project-knowledge => The client itself enriches stuff with project-specific stuff so I don't need to.
The new functions do not behave like that. They require project-specific stuff (instantiation of the Action Objects). Hence the end-user needs to know THAT he needs to import stuff to use this method and WHERE the stuff he needs to import comes from. The part of HOW he needs to instantiate is quite obvious but could be another hurdle depending on the point of view.

Maybe i overlooked some parts here because i never used all this action/schema stuff before. But while adapting the tests (that basically invoke the client stuff like an end-user would) i had to instantiate the objects. So these points seemed valid to me

src/prefect/client/schemas/actions.py
```diff
@@ -617,3 +617,33 @@ class VariableUpdate(ActionBaseModel):
...
+    active_slots: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+    slot_decay_per_second: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
+
+
+@copy_model_fields
+class GlobalConcurrencyLimitUpdate(ActionBaseModel):
+    """Data used by the Prefect REST API to update a global concurrency limit."""
+
+    name: str = FieldFrom(objects.GlobalConcurrencyLimit)
+    limit: int = FieldFrom(objects.GlobalConcurrencyLimit)
```
bunchesofdonald, line 638: On update these both can be optional. The reason is that we'll be doing a `PATCH` which, in REST parlance, means we can send a partial object and expect that only the fields we send are going to be updated.

```suggestion
    name: Optional[str] = FieldFrom(objects.GlobalConcurrencyLimit)
    limit: Optional[int] = FieldFrom(objects.GlobalConcurrencyLimit)
```

src/prefect/client/orchestration.py
```diff
@@ -2694,6 +2696,74 @@ async def release_concurrency_slots(
...
+    ) -> httpx.Response:
+        concurrency_limit_dict = concurrency_limit.dict(
+            json_compatible=True, exclude_unset=True
+        )
+        current_name = concurrency_limit_dict.get("name")
+        concurrency_limit_dict["name"] = concurrency_limit_dict.get(
+            "rename_to", current_name
+        )
+        if "rename_to" in concurrency_limit_dict.keys():
+            concurrency_limit_dict.pop("rename_to", None)
```
bunchesofdonald, line 2719: Making the name/limit in the update action optional should allow you to remove these checks.

bunchesofdonald (changes requested):
I think we just need to make those couple fields optional and then this will be good to go.

bunchesofdonald:
> Hey, adapted your MR suggestions and made the tests run again accordingly :)
> 
> Maybe a PoV from a prefect user (normally): Normally, when using client libraries from other projects I am used to using their provided methods by providing elementary data types or sometimes more complex ones (dicts, lists, ...) solely. This has the advantage that the end-user does not need to know project-specific stuff and can stay on a relatively low level of project-knowledge => The client itself enriches stuff with project-specific stuff so I don't need to. The new functions do not behave like that. They require project-specific stuff (instantiation of the Action Objects). Hence the end-user needs to know THAT he needs to import stuff to use this method and WHERE the stuff he needs to import comes from. The part of HOW he needs to instantiate is quite obvious but could be another hurdle depending on the point of view.
> 
> Maybe i overlooked some parts here because i never used all this action/schema stuff before. But while adapting the tests (that basically invoke the client stuff like an end-user would) i had to instantiate the objects. So these points seemed valid to me
> 
> Edit: Maybe naive suggestion: Create the object inside the client method. When doing so, the method could have quite simple data types and will be easily usable. During the method's execution the pydantic model will be created and all evaluation stuff should happen normally (right? Not familiar with the pydantic stuff, sorry). Looks a little bit different then but achieves the same goal


I don't disagree with that point-of-view. Being able to pass in kwargs to the function and have it build the interal model could be a nicer approach. The primary advantage of doing it this way for _this_ PR is consistency with the rest of the client. Changing the client as a whole to do it the other way would be a much larger change, and I'm not sure it's worth it. I'm open to it, but I think it's a separate discussion that would be best had in a separate issue.

src/prefect/client/orchestration.py
Zyntogz: I don't know if I don't see something here:
rename_to is no field from the API that can be filled, hence cannot be sent in the json.
Your api supports renaming but when adapting to your suggesting the client doesn't support it.
Example:
There is a limit called "limit". I can do a PATCH-request to <your-url>/api/limit providing this json: {"name": "new-limit"} to rename it. To do this, we need at least two fields containing a name in the client function, hence the variables "name" and "rename_to". "name" always affects the URL to send the request to BUT ALSO the name it shall be renamed to. Hence, in the "name" field in the payload conditionally has to be replaced with "rename_to".

src/prefect/client/schemas/actions.py
Zyntogz: OK, i'll adapt that accordingly

src/prefect/client/orchestration.py
bunchesofdonald: Ahh, good point! That's a creative solution to the problem, nice! 

The _alternate_ route here is that we could require the `id` of the concurrency limit and since the API supports either a name or an id you could just do what I was suggesting. But I think I like your approach better, it keeps the user from having to make a second api call to get the id to do the update.

Zyntogz: That's true, using an ID would result in distinguishable variable names. But as you said, IDs have some kind of overhead to them and names are more intuitive.
Thanks for your constructive and helpful feedback anyways. I'll push another quick update :)

Zyntogz:
With some time and distance i have another suggestion to make that should handle my end-user PoV concerns regarding usability of the method and your more than valid concerns regarding pydantic validation.

Using *args and **kwargs in combination with a well-defined docstring we could combine a literal passthrough of provided parameters of the function directly to PyDantic (hence, fulfilling your 3 points mentioned above) and still leverage simple parameter provision without specifying parameters again in the function's parameter definition.

Example
orchestration.py:
`   async def create_global_concurrency_limit(
        self, *args, **kwargs
    ) -> UUID:
        response = await self._client.post(
            "/v2/concurrency_limits/",
            json=GlobalConcurrencyLimitCreate(*args, **kwargs).dict(json_compatible=True, exclude_unset=True),
        )
        return UUID(response.json()["id"])
`
usage (here, test_prefect_client.py):
`
async def test_global_concurrency_limit_create(prefect_client):
    response_uuid = await prefect_client.create_global_concurrency_limit(
        GlobalConcurrencyLimitCreate(name="global-create-test", limit=42)
    )
    assert response_uuid == UUID(
        (
            await prefect_client.read_global_concurrency_limit_by_name(
                name="global-create-test"
            )
        ).get("id")
    )
`
All tests pass (only tested with existing ones, could add others validating the *args / **kwargs passthrough behavior), pydantic can validate everything it gets passed and is able to handle excess parameters accordingly (fail on unknown parameters / types) without defining stuff defined in the pydantic model again. Only downside from my PoV: Like every other *args,**kwargs approach a small lack of use regarding what can be passed to the function. But this should be easily counterable using a nice docstring format like

"""
<function description here>
:param name: Name of the global concurrency limit
:param limit: Limit
...
"""

What are your thoughts? This may be a much wider change that just this PR so no hurry, i'm fully with you merging the previously worked out solution. But when there is a change of mind regarding easier parameter provision of client functions to match other already implemented client functions there will be a breaking change and end-users need to adapt their code to that

src/prefect/client/orchestration.py
```diff
@@ -2694,6 +2696,74 @@ async def release_concurrency_slots(
...
+    ) -> UUID:
+        response = await self._client.post(
+            "/v2/concurrency_limits/",
+            json=concurrency_limit.dict(json_compatible=True, exclude_unset=True),
+        )
+        return UUID(response.json()["id"])
+
+    async def update_global_concurrency_limit(
+        self, concurrency_limit: GlobalConcurrencyLimitUpdate
+    ) -> httpx.Response:
```
bunchesofdonald, line 2710: I'm sorry to ask for yet another change here, but the more I think about this, the more I'm not sure that this is the actual interface we want to provide here. The `rename_to` attribute is a creative solution, but I think a more straightforward option is to just put the `name` as an argument to the function:

```suggestion
    async def update_global_concurrency_limit(
        self, name: str, concurrency_limit: GlobalConcurrencyLimitUpdate
    ) -> httpx.Response:
```

This is consistent with our other update methods that accept an `id` and the data object, and gives us the `name` to use for generating the URL as well as allowing a user to provide a new name in the data object.

bunchesofdonald (changes requested):
Sorry for yet another request, but I the more I think about `rename_to`, the more it feels like the wrong approach, creative as it is.

bunchesofdonald:
> What are your thoughts? This may be a much wider change that just this PR so no hurry, i'm fully with you merging the previously worked out solution. But when there is a change of mind regarding easier parameter provision of client functions to match other already implemented client functions there will be a breaking change and end-users need to adapt their code to that

I think I'd prefer to have defined args/kwargs, mostly because IDE's would be more able to help with knowing what arguments the function takes, and would allow us to provide type hints which would improve the developer experience even more. It would feel like a lot of piping when you create these methods, but I think the contract with the developer using the methods would be a lot clearer. Either way this is definitely a wider change, and as you point out a breaking change. I'd say create another issue and we can discuss it further there and get feedback from our product team.

Zyntogz:
Happy new year, was on vacation so this had to wait a little.

Something crazy happened with the commit history and i couldn't solve it. So it quickly force-resetted it to the current main state and re-commited the "final" changes to the four affected files. 

I adapted to your last suggestion and added the parameter name to the function definition instead of rename_to logic. Looks nice for me. For the end user still somehow tedious to use the client methods but i can live with that.

Feel free to re-review :)

bunchesofdonald (approved):
This looks great, thank you for sticking with me through those changes @Zyntogz!

src/prefect/client/orchestration.py
```diff
@@ -2722,6 +2724,66 @@ async def release_concurrency_slots(
...
+            return response
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == status.HTTP_404_NOT_FOUND:
+                raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
+            else:
+                raise
+
+    async def read_global_concurrency_limit_by_name(
+        self, name: str
+    ) -> dict[str, object]:
```
bunchesofdonald, line 2765:
```suggestion
    ) -> Dict[str, object]:
```

src/prefect/client/orchestration.py
```diff
@@ -2722,6 +2724,66 @@ async def release_concurrency_slots(
...
+            return response.json()
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == status.HTTP_404_NOT_FOUND:
+                raise prefect.exceptions.ObjectNotFound(http_exc=e) from e
+            else:
+                raise
+
+    async def read_global_concurrency_limits(
+        self, limit: int = 10, offset: int = 0
+    ) -> list[dict[str, object]]:
```
bunchesofdonald, line 2777:
```suggestion
    ) -> List[Dict[str, object]]:
```

bunchesofdonald (changes requested):
Hey @Zyntogz I get tripped up by this all the time, but we need to support Python 3.8, which means we have to use `Dict` and `List` from `typing` instead of just the standard functions. Sorry I missed this on the other reviews!

bunchesofdonald (approved):

PR #11437, opened 2023-12-15
Update default pause/suspend timeout to be 1 hour

### LINKED ISSUE #11413: Increase default timeout for pause_flow_run and suspend_flow_run
The current default timeouts for `pause_flow_run()` and `suspend_flow_run()` are five minutes, which is much too short. In fact, why time out at all by default? Let's either lengthen the default timeout or remove it.

**Note:** I don't think expiration processor can cope with None `pause_timeout` values at the moment, if we go that route.

bunchesofdonald:
Updates the default `pause_flow_run` and `suspend_flow_run` from 5m (300s) to 1 hour (3600s). 

Closes #11413 

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/engine.py
+++ b/src/prefect/engine.py
@@ -952,7 +952,7 @@ async def orchestrate_flow_run(
 async def pause_flow_run(
     wait_for_input: None = None,
     flow_run_id: UUID = None,
-    timeout: int = 300,
+    timeout: int = 3600,
     poll_interval: int = 10,
     reschedule: bool = False,
     key: str = None,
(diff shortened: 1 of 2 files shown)


chrisguidry (approved):

src/prefect/engine.py
```diff
@@ -1192,10 +1189,13 @@ async def suspend_flow_run(
 
 @sync_compatible
 @inject_client
+@experimental_parameter(
```
bunchesofdonald, line 1192: While I was in here I noticed that I'd incorrectly applied the experimental decorator to the wrong method, should be here on `suspend` rather than the private method above.

abrookins (approved):
Awesome, this feels right! 👍

abrookins:
That blasted `ValueError: signal only works in main thread` flaky test! One day.

PR #11487, opened 2023-12-20
Docs additions for human in the loop

abrookins:
Expand docs for Human in the Loop features (pause, suspend, resume, etc.).

TODO in a subsequent PR: expanded guide to using `RunInput` type annotations.

### Checklist

For documentation changes:


docs/concepts/flows.md
```diff
@@ -870,25 +870,32 @@ When a flow run is suspended, code execution is stopped and so is the process.
 
 Prefect enables pausing an in-progress flow run for manual approval. Prefect exposes this functionality via the [`pause_flow_run`](/api-ref/prefect/engine/#prefect.engine.pause_flow_run) and [`resume_flow_run`](/api-ref/prefect/engine/#prefect.engine.resume_flow_run) functions, as well as the Prefect UI.
 
-Most simply, `pause_flow_run` can be called inside a flow. A timeout option can be supplied as well &mdash; after the specified number of seconds, the flow will fail if it hasn't been resumed.
+!!! note "Timeouts"
+    Paused flow runs time out after one hour by default. After the timeout, the flow run will fail with a message saying it paused and never resumed. You can specify a different timeout period in seconds using the `timeout` parameter or pass `timeout=None` for no timeout.
```
bunchesofdonald, line 874: Blocking pause won't accept a `None` timeout, only suspending / `reschedule=True`.

```suggestion
    Paused flow runs time out after one hour by default. After the timeout, the flow run will fail with a message saying it paused and never resumed. You can specify a different timeout period in seconds using the `timeout` parameter.
```

bunchesofdonald (approved):
Looks great!

docs/guides/creating-human-in-the-loop-workflows.md
```diff
@@ -69,88 +69,91 @@ async def greet_user():
 
 ## Handling custom validation
 
-Prefect uses the fields and type hints on your `RunInput` subclass to validate the general structure of input your flow run receives, but you might require more complex validation. If you do, you can use Pydantic's validation methods.
+Prefect uses the fields and type hints on your `RunInput` subclass to validate the general structure of input your flow run receives, but you might require more complex validation. If you do, you can use Pydantic [validators](https://docs.pydantic.dev/1.10/usage/validators/).
+
+!!! warning "Custom validation runs after the flow run resumes"
+    We transform the type annotations in your `RunInput` class to a JSON schema and use that in our UI to do client-side validation. However, custom validation requires running logic defined in your `RunInput` class. This happens *after the flow resumes*, so you'll probably want to handle it explicitly in your flow. Continue reading for an example best practice.
```
discdiver, line 75:
```suggestion
    Prefect transforms the type annotations in your `RunInput` class to a JSON schema and use that schema in the UI to do client-side validation. However, custom validation requires running logic defined in your `RunInput` class. This happens *after the flow resumes*, so you'll probably want to handle it explicitly in your flow. Continue reading for an example best practice.
```

docs/guides/creating-human-in-the-loop-workflows.md
```diff
@@ -69,88 +69,91 @@ async def greet_user():
...
-    user_age_input = pause_flow_run(wait_for_input=UserAgeInput)
+def get_shirt_order():
+    shirt_order = pause_flow_run(wait_for_input=ShirtOrder)
 ```
 
-If a user enters an invalid age of `hello`, the flow run will not resume and the user will receive a validation error. However, if the user enters a number that is a valid integer but that is not between 18 and 99, the flow run will resume, and `pause_flow_run` will raise a `ValidationError` exception.
+If a user chooses any size and color combination other than `small` and `green`, the flow run will resume successfully. However, if the user chooses size `small` and color `green`, the flow run will resume, and `pause_flow_run` will raise a `ValidationError` exception. This will cause the flow run to fail and log the error.
 
-One way to handle this and ensure that the user enters valid input is to use a `while` loop and pause again if the `ValidationError` exception is raised:
+However, what if you don't want the flow run to fail? One way to handle this is to use a `while` loop and pause again if the `ValidationError` exception is raised:
```
discdiver, line 123:
```suggestion
However, what if you don't want the flow run to fail? One way to handle this case is to use a `while` loop and pause again if the `ValidationError` exception is raised:
```

docs/guides/creating-human-in-the-loop-workflows.md
```diff
@@ -69,88 +69,91 @@ async def greet_user():
...
-            logger.error(f"Invalid age: {exc}")
+            logger.error(f"Invalid size and color combination: {exc}")
 
-    logger.info(f"User age: {user_age_data.age}")
+    logger.info(f"Shirt order: {shirt_order.size}, {shirt_order.color}")
 ```
 
 This code will cause the flow run to continually pause until the user enters a valid age.
+
+As an additional step, you may want to use an [automation](/concepts/automations) or [notification](/concepts/notifications/) when this happens, to make sure the user sees the error.
```
discdiver, line 159:
```suggestion
As an additional step, you may want to use an [automation](/concepts/automations) or [notification](/concepts/notifications/) to alert the user to the error.
```

discdiver (commented):
Thank you. Minor suggestions for clarity.

PR #11493, opened 2023-12-20
Enhancement: Reverse proxy auto-build

znicholasbrown:
Building on the improvements from #11489, this PR modifies the way we're building the UI in the distributable and then serving it using the `prefect server start` command. 

The build command now encodes a placeholder string into all asset references using the `base` rollup option.

When `prefect server start` is run for the first time, we copy all of the ui static assets to a subdirectory and find/replace the placeholder string with the `PREFECT_UI_SERVE_BASE` variable, used to serve the UI from a subdirectory/reverse proxy. We also include a reference file in that subdirectory when we do this so that subsequent calls to `prefect server start` don't recreate the directory each time, unless the `PREFECT_UI_SERVE_BASE` variable has changed.


pleek91 (approved):

mthanded:
@znicholasbrown 

Im trying to understand the changes to the way PREFECT_UI_SERVE_BASE impacts self hosting. From what I can tell now this is a runtime variable and not a build variable. Got that change working and moved it to the actual server start environment. 

However the change in the uisettings.ts file reverted a change i suggested in #7068 and i don't understand how to get that behavior back. Despite having prefect server mounted at /blah the frontend is calling out to /ui-settings instead of /blah/ui-settings. How do i tell the front end to look at the mounted path with this new approach?

znicholasbrown:
You're right @mthanded and you're also correct this is now a runtime variable to avoid the rebuild step.

However, I think [this](https://github.com/PrefectHQ/prefect/pull/11701) will address settings only being available at the root, would you mind giving that a try and seeing if that fits your use case?

mthanded:
> You're right @mthanded and you're also correct this is now a runtime variable to avoid the rebuild step.
> 
> 
> 
> However, I think [this](https://github.com/PrefectHQ/prefect/pull/11701) will address settings only being available at the root, would you mind giving that a try and seeing if that fits your use case?

Thanks. Will check this out. Really appreciate the quick response.

PR #11591, opened 2024-01-08
Add send/receive documentation to `run_input` module docstring

### LINKED ISSUE #11590: Document RunInput's send/receive basics in the `run_input` module.

bunchesofdonald:
Closes #11590

### DIFF
--- a/src/prefect/input/run_input.py
+++ b/src/prefect/input/run_input.py
@@ -1,3 +1,64 @@
+"""
+This module offers a framework for data exchange between flow runs enabling
+real-time interaction and efficient data handling. It's designed to facilitate
+dynamic communication within distributed or microservices-oriented systems,
+making it ideal for scenarios requiring continuous data synchronization and
+processing.
+
+Use this module to create responsive, real-time applications where different
+flow runs need to communicate and exchange data dynamically. It's particularly
+useful for systems that require ongoing data input and output.
+
+The following is an example of two flows. One sends a random number to the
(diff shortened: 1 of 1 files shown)


src/prefect/input/run_input.py
```diff
@@ -1,3 +1,64 @@
+"""
+This module offers a framework for data exchange between flow runs enabling
+real-time interaction and efficient data handling. It's designed to facilitate
+dynamic communication within distributed or microservices-oriented systems,
+making it ideal for scenarios requiring continuous data synchronization and
+processing.
+
+Use this module to create responsive, real-time applications where different
+flow runs need to communicate and exchange data dynamically. It's particularly
+useful for systems that require ongoing data input and output.
```
serinamarie, line 10: nit: Maybe this could be more concise? It feels like these two paragraphs are saying similar things.

serinamarie (approved):
This is cool! I'm guessing this will go into the docs as well?

src/prefect/input/run_input.py
```diff
@@ -1,3 +1,64 @@
+"""
+This module offers a framework for data exchange between flow runs enabling
```
abrookins, line 2: I would combine these paragraphs and stick to what users can do with it:

```suggestion
This module contains functions that allow sending type-checked `RunInput` data
to flows at runtime. Flows can send back responses, establishing two-way
channels with senders. These functions are particularly useful for systems that
require ongoing data transfer or need to react to input quickly.
```

PR #11598, opened 2024-01-09
Use default values if no run input was provided

abrookins:
It's possible to have a flow run waiting for input that uses only optional fields wtih defaults, especially if the input class was created with `RunInput.with_initial_data`. In this case, we should pass the empty data to the `RunInput` class and rely on validation to ensure any required fields are present instead of rejecting the empty input.

### Checklist

For documentation changes:


bunchesofdonald (approved):
Makes sense, nice job tracking that down.

serinamarie:
Adding `fix` label here for easier release notes generation. Feel free to change if that doesn't fit.

PR #11600, opened 2024-01-09
Prevent deployment with `RRule` schedule containing `COUNT`

serinamarie:
While users can specify an RRule schedule with a `COUNT` parameter, we do not respect it nor do we plan to support it at this time. Thus it makes sense to ensure users don't go down the wrong path here. The `COUNT` is in this case the number of flow runs they would like scheduled, in total. For example, `COUNT=400` means a user wants a total of 400 flow runs scheduled for a deployment. While we could technically support this on some level (`COUNT=2` or `COUNT=3` work fine), we don't think that keeping track of how many flow runs have already been scheduled will be easily maintainable. 

Example:
```python
# foo.py

from prefect import flow
from prefect.server.schemas.schedules import RRuleSchedule


@flow
def foo():
    assert 1 == 1


if __name__ == "__main__":

    rrule_str = """
        DTSTART:20230908T150000
        RRULE:FREQ=HOURLY;INTERVAL=1;COUNT=3
    """
    
    Deployment.build_from_flow(
        flow=foo,
        name="foo-deployment",
        apply=True,
        schedule=RRuleSchedule(rrule=rrule_str),
    )
```

This will now raise a `ValidationError`:
```bash
pydantic.error_wrappers.ValidationError: 1 validation error for Deployment
schedule
  RRule schedule with a COUNT is not supported. Please use 1) a schedule with an UNTIL date instead, or 2) schedule runs with the `/{deployment_id}/schedule` endpoint. (type=value_error)
```

### Checklist

For documentation changes:


src/prefect/deployments/deployments.py
```diff
@@ -575,6 +575,19 @@ def validate_automation_names(cls, field_value, values, field, config):
...
+        """We do not support COUNT-based (# of occurrences) RRule schedules for deployments."""
+        if value:
+            rrule_value = getattr(value, "rrule", None)
+            if rrule_value and "COUNT" in rrule_value.upper():
+                raise ValueError(
+                    "RRule schedules with `COUNT` are not supported. Please use 1."
+                    " `UNTIL`, or 2. schedule runs with the"
+                    " `/deployments/{id}/schedule` endpoint."
+                )
+        return value
```
serinamarie, line 589: another option is we could potentially warn instead? but not sure how that would help.

bunchesofdonald: Raising seems like the right approach to me, a warning can be missed and confusing, whereas there's nothing ambiguous about failure.

bunchesofdonald (approved):
Looks good!

PR #11608, opened 2024-01-10
Support defining if a schedule is active or not via `prefect.yaml`

### LINKED ISSUE #11334: prefect.yaml should support is_schedule_active
### First check

### Bug summary

The `prefect.yaml` file does not allow a user to set `is_schedule_active` for a given deployment. In some cases users don't want schedules to automatically, for example when deploying to a dev or staging environment.

### Reproduction

```python3
build: ...
push: ...
pull: ...

deployments:
  - name: my-deployment
    schedule:
      interval: 60
    is_schedule_active: false
```

### Error

_No response_

### Versions

```Text
Version:             2.14.9
API version:         0.8.4
Python version:      3.9.18
Git commit:          4fba882c
Built:               Thu, Nov 30, 2023 2:55 PM
Profile:             personal
Server type:         cloud
```

### Additional context

_No response_

bunchesofdonald:
This adds support for an `active` property on the `schedule` definition of a deployment in `prefect.yaml`. It also adds support to the interactive CLI.

Closes #11334 
Closes https://github.com/PrefectHQ/nebula/issues/5838

### Example
Screenshot 2024-01-10 at 11 57 35 AM

### Checklist

For documentation changes:

### DIFF
--- a/docs/concepts/schedules.md
+++ b/docs/concepts/schedules.md
@@ -76,7 +76,7 @@ deployments:
   schedule:
     cron: 0 0 * * *
     timezone: America/Chicago
-    active: true
+    active: false
 ```
 
 !!! tip "Schedules can be inactive"
--- a/docs/concepts/schedules.md
+++ b/docs/concepts/schedules.md
@@ -76,7 +76,12 @@ deployments:
   schedule:
     cron: 0 0 * * *
     timezone: America/Chicago
+    active: true
 ```
+
+!!! tip "Schedules can be inactive"
+    You can set the `active` property to `false` to deactivate a schedule.
+    This is useful if you want to keep the schedule configuration but temporarily stop the schedule from creating new flow runs.
 
 Let's discuss the three schedule types in more detail.
 
--- a/src/prefect/cli/_prompts.py
+++ b/src/prefect/cli/_prompts.py
@@ -5,7 +5,7 @@ import os
 import shutil
 from datetime import timedelta
 from getpass import GetPassWarning
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import readchar
 from rich.console import Console, Group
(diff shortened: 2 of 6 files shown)


docs/concepts/schedules.md
```diff
@@ -76,8 +76,13 @@ deployments:
   schedule:
     cron: 0 0 * * *
     timezone: America/Chicago
+    active: true
```
serinamarie, line 79: would it make more sense to show `active: false` since the default is True?

docs/concepts/schedules.md
```diff
@@ -76,8 +76,13 @@ deployments:
...
 ```
 
+!!! tip "Schedules can be inactive"
+    You can set the `active` property to `false` to deactivate a schedule.
+    This is useful if you want to keep the schedule configuration but temporarily stop the schedule from creating new flow runs.
+
```
serinamarie, line 85: or maybe instead, 'if you want to deploy a paused deployment'?

src/prefect/cli/_prompts.py
```diff
@@ -328,21 +328,27 @@ def prompt_schedule_type(console):
...
+        schedule = prompt_interval_schedule(console)
     elif schedule_type == "RRule":
-        return prompt_rrule_schedule(console)
+        schedule = prompt_rrule_schedule(console)
     else:
         raise Exception("Invalid schedule type")
 
+    is_schedule_active = confirm(
+        "Would you like to activate this schedule?", default=True
+    )
```
serinamarie, line 348: i'm not sure if it makes more sense to say 'activate this schedule' (which to me sounds a little confusing because  the schedule is always listed on the deployment regardless).. rather than something about the deployment being paused or not.

docs/concepts/schedules.md
bunchesofdonald: Sure, that makes sense.

src/prefect/cli/_prompts.py
bunchesofdonald: Right, my thinking here was that the next part of of the enhanced-scheduling raft is to make deployment->schedules a one->many relation, in which case this wouldn't be talking about the _deployment_ being paused it would in fact be about the schedule.

src/prefect/deployments/base.py
```diff
@@ -416,6 +416,9 @@ def _format_deployment_for_saving_to_prefect_file(
         elif isinstance(deployment["schedule"], BaseModel):
             deployment["schedule"] = deployment["schedule"].dict()
 
+        if "is_schedule_active" in deployment:
+            deployment["schedule"]["active"] = deployment.pop("is_schedule_active")
+
```
serinamarie, line 421: can you remind me why we wouldn't want to store it in the `prefect.yaml`? it seems like it'd be nice to have for when you re-deploy with `deploy --all`, otherwise i think that the schedules would all be active again.

serinamarie (commented):
Could we also have a test for when we `deploy --all`, that if we have saved schedules as inactive in our `deployments` section of our `prefect.yaml`, that the schedules remain inactive?

Otherwise it LGTM :)

abrookins (approved):
Looks great! Thanks to the close read by @serinamarie, I don't have anything to add. 👍

abrookins (approved):
Looks great! Thanks to the close read by @serinamarie, I don't have anything to add. 👍

serinamarie (approved):
Yay, nice test! LGTM 🚀

PR #11628, opened 2024-01-11
Bugfix: Missing `UI_SERVE_BASE_URL` lead to incorrect asset urls

znicholasbrown:
Fixes an issue where setting `UI_SERVE_BASE` to an empty string or `/` lead to incorrectly replaced asset urls like `//asset.js` instead of `/asset.js`

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -370,7 +370,7 @@ def create_ui_app(ephemeral: bool) -> FastAPI:
         replace_placeholder_string_in_files(
             prefect.__ui_static_subpath__,
             "/PREFECT_UI_SERVE_BASE_REPLACE_PLACEHOLDER",
-            base_url if base_url != "/" else "",
+            base_url.rstrip("/"),
         )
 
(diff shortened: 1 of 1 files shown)


src/prefect/server/api/server.py
```diff
@@ -370,7 +370,7 @@ def create_ui_static_subpath():
         replace_placeholder_string_in_files(
             prefect.__ui_static_subpath__,
             "/PREFECT_UI_SERVE_BASE_REPLACE_PLACEHOLDER",
-            base_url,
+            base_url if base_url != "/" else "",
```
collincchoy, line 373: The issue here was if `base_url == "/"` exactly? 

is `base_url==""` also an issue?

znicholasbrown: Yeah except `base_url` defaults to `/` because that's the default on the setting

znicholasbrown: So empty `base_urls` always ended up as `/`

serinamarie (approved):

collincchoy (approved):

PR #11633, opened 2024-01-13
Fix tasks.md -> retry handler

N-Demir:
I think this documentation may not work. As in, for any other exception it just gets raised after the last except. `return True`, which is what the documentation says should happen in that case, never gets called. 

Here's some example code to show it:
```python
try:
    raise NotImplementedError()
except KeyError:
    print("Got caught")
    quit()

print("Did get through")
```
Output: `NotImplementedError`

but the documentation here makes it seem like `Did get through` should be printed. 

I think this fixes it!

### DIFF
--- a/docs/concepts/tasks.md
+++ b/docs/concepts/tasks.md
@@ -285,9 +285,9 @@ def retry_handler(task, task_run, state) -> bool:
     except httpx.ConnectError:
         # Do not retry
         return False
-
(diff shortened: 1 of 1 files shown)


serinamarie (approved):
Hi @N-Demir, thanks for making your first pull contribution to Prefect :) Yeah, I think this is definitely a way to fix that code. We appreciate it!

PR #11636, opened 2024-01-14
Automatically convert type arguments into RunInput subclasses

abrookins:
This PR makes writing interactive flows even easier:
 
1. If you pause or suspend a flow run to receive a single-value input (e.g., `age: int`, `name: str`, etc.), Prefect will create the `RunInput` class for you.
2. Similarly, you can now send single-value inputs to flow runs and receive them in the same manner without creating a `RunInput` class. 
3. You can now use your existing `pydantic.BaseModel` classes without any modification. Any function that takes a `RunInput` can instead take a `pydantic.BaseModel`, and Prefect will create a `RunInput` from your `BaseModel`.

## Automatic inputs for paused and suspended flows
Pausing and suspending to wait for single-value run inputs does not require creating a `RunInput` class. Just pass the type of the value you want to receive:

```python
age = await pause_flow_run(int, timeout=10, poll_interval=2)
```

Previously, you would have had to write something like the following:

```python
class AgeInput(RunInput):
    age: int

input = await pause_flow_run(AgeInput, timeout=10, poll_interval=2)
input.age
```

**NOTE:** You can still use an explicit `RunInput` class to change things like the name of this field when it appears in our UI, or collect more than one value.

## Automatic inputs for interactive flows
You can also use automatic input conversion with our sending and receiving functions (new: `send_input` and `receive_input`), without using pause/suspend:

```python
# Flow one
await send_input(10, flow_run_id=123)

# Flow two
age = await receive_input(int)
```

## Use your existing `BaseModel` classes as input
Finally, if you have an existing `BaseModel` and don't need to change any behavior, you can now pass it to any function that accepts a `RunInput` and we will do the needful to make it work for flow run input.

### Checklist

For documentation changes:

### DIFF
--- a/src/prefect/input/run_input.py
+++ b/src/prefect/input/run_input.py
@@ -285,7 +285,7 @@ class RunInput(pydantic.BaseModel):
         cls, model_cls: Type[pydantic.BaseModel]
     ) -> Type["RunInput"]:
         """
-        Create a new `RunInput` instance from the given base model.
+        Create a new `RunInput` subclass from the given base model.
 
         Args:
             - model_cls (pydantic.BaseModel subclass): the base model class
--- a/src/prefect/input/run_input.py
+++ b/src/prefect/input/run_input.py
@@ -281,6 +281,36 @@ class RunInput(pydantic.BaseModel):
         )
 
 
+class AutomaticRunInput(RunInput):
+    @classmethod
+    @sync_compatible
+    async def load(cls, keyset: Keyset, flow_run_id: Optional[UUID] = None):
+        """
+        Load the run input response from the given key.
+
+        Args:
+            - keyset (Keyset): the keyset to load the input for
+            - flow_run_id (UUID, optional): the flow run ID to load the input for
+        """
+        instance = super().load(keyset, flow_run_id=flow_run_id)
+        return instance.value
+
+
+def run_input_from_type(_type: Type[T]) -> Type[RunInput]:
+    """
+    Create a new `RunInput` subclass from the given type.
+    """
+    if issubclass(_type, RunInput):
+        return _type
+    elif issubclass(_type, pydantic.BaseModel):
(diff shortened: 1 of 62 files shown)


abrookins:
I'm not 100% I've done the typing here correctly: https://github.com/PrefectHQ/prefect/pull/11636/files#diff-ae6f1b6d423c9f28188aedf4ab0f0afbce428d61b84feed78ca9c716ce3e26caR978

bunchesofdonald:
> I'm not 100% I've done the typing here correctly: https://github.com/PrefectHQ/prefect/pull/11636/files#diff-ae6f1b6d423c9f28188aedf4ab0f0afbce428d61b84feed78ca9c716ce3e26caR978

I think you just want to change the `TypeVar` `T` to not be bounded. So you'd just want to update them from:

`T = TypeVar("T", bound=RunInput)` to `T = TypeVar("T")` I think we define that both in the engine and in `run_input.py`.

src/prefect/input/run_input.py
```diff
@@ -280,6 +280,60 @@ def receive(
...
+
+        Args:
+            - model_cls (pydantic.BaseModel subclass): the class from which
+                to create the new `RunInput` subclass
+        """
+        return type(f"{model_cls.__name__}RunInput", (RunInput, model_cls), {})  # type: ignore
+
+
+class AutomaticRunInput(RunInput):
+    value: Any
```
bunchesofdonald, line 299: You could assign this a typevar (like we use `T`) and then you could load's return type be that typevar, I _think_ that would give nice type handling, so if a user were to pass `int` through it should come back out as `int` in their code. You might also need to update `AutomaticRunInput` to be `AutomaticRunInput[V]` in `subclass_from_type` and make this class a `Generic[V]`. If that all sounds messy I can also just play with it after this is merged and get it working.

bunchesofdonald (approved):
Looks great! I think the typing needs some help, but I also don't think that should block this from being merged.

Issue #11640, opened 2024-01-15
OS Permission Error when running non-root prefect server due to changes from #11493


marcm-ml:
### First check

- [X] I added a descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

#11493 introduces changes that cannot run in non-root docker environments. The problem is that the directory where the UI files are copied into, is not writeable in scenarios where the Docker user is non-root. This is the case when you override the User at the end of a Dockerfile. Deploying via helm using prefect-helm is also resulting in the same error  due to the securityContext being set to User 1001

### Reproduction

```python3
starting the docker image without privileges and command `prefect server start --port 4200`
```

### Error

```python3
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/venv/lib/python3.10/site-packages/uvicorn/__main__.py", line 4, in <module>
    uvicorn.main()
  File "/opt/venv/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/opt/venv/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/uvicorn/main.py", line 416, in main
    run(
  File "/opt/venv/lib/python3.10/site-packages/uvicorn/main.py", line 587, in run
    server.run()
  File "/opt/venv/lib/python3.10/site-packages/uvicorn/server.py", line 61, in run
    return asyncio.run(self.serve(sockets=sockets))
  File "/usr/local/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
[...]

PR #11648, opened 2024-01-16
Bugifx: Adjust UI serving to work with a static directory

znicholasbrown:
Fixes an issue documented in #11640 where running `prefect server start` with a user that didn't have write permissions to the package subdirectory would cause the command to fail. 

This exposes a setting, `PREFECT_UI_STATIC_DIRECTORY`, that `prefect server start` will attempt to write to instead, allowing the directory to be created/properly permissioned ahead of time. We also no longer remove the root directory when the setting has changed (or on first run) and only overwrite its contents.

Resolves: #11640

### DIFF
--- a/src/prefect/server/api/server.py
+++ b/src/prefect/server/api/server.py
@@ -210,12 +210,17 @@ def replace_placeholder_string_in_files(
 def copy_directory(directory, path):
     os.makedirs(path, exist_ok=True)
     for item in os.listdir(directory):
-        s = os.path.join(directory, item)
-        d = os.path.join(path, item)
-        if os.path.isdir(s):
-            shutil.copytree(s, d, symlinks=True)
+        source = os.path.join(directory, item)
+        destination = os.path.join(path, item)
+
+        if os.path.isdir(source):
+            # If the directory already exists at the destination, remove it first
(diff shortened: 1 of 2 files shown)


src/prefect/server/api/server.py
```diff
@@ -210,12 +210,17 @@ def replace_placeholder_string_in_files(
 def copy_directory(directory, path):
```
znicholasbrown, line 210: Made a few changes to this method:
1. Cleans up some of the variable names to make parsing the method easier
2. Removes subdirectories when walking the file graph and recreates them (`copy2` does this with files by default)

src/prefect/server/api/server.py
```diff
@@ -210,12 +210,15 @@ def replace_placeholder_string_in_files(
...
     os.makedirs(path, exist_ok=True)
     for item in os.listdir(directory):
-        s = os.path.join(directory, item)
-        d = os.path.join(path, item)
-        if os.path.isdir(s):
-            shutil.copytree(s, d, symlinks=True)
+        source = os.path.join(directory, item)
```
chrisguidry, line 213: :clap:

chrisguidry (approved):
Reads well to me!

PR #11670, opened 2024-01-18
Make `server database downgrade` default revision less destructive

collincchoy:
Running `prefect server database downgrade` without an explicit target revision defaults to "base" which basically drops the database. While there is a confirmation step here, it'd be nice if the default were less destructive in general as dropping a database is a pretty unfortunate booboo. 

Aside from safety, this new default behavior should be familiar to folks from the flask ecosystem working with tooling like flask-migrate. This default is also just a little more convenient when developing with in-flight migrations. 

### Example
```sh
# this will now downgrade to the current revision - 1. running 0-1 migrations at a time
prefect server database downgrade

# to use the previous default behavior
prefect server database downgrade -r base
```

### Checklist

For documentation changes:

### DIFF
--- a/docs/guides/host.md
+++ b/docs/guides/host.md
@@ -222,6 +222,8 @@ prefect server database downgrade -y -r d20618ce678e
 ```
 </div>
 
+To downgrade all migrations, use the `base` revision.
+
 See the [contributing docs](/contributing/overview/#adding-database-migrations) for information on how to create new database migrations.
 
 ## Notifications
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -219,8 +219,8 @@ async def upgrade(
 async def downgrade(
     yes: bool = typer.Option(False, "--yes", "-y"),
     revision: str = typer.Option(
-        "base",
         "-1",
+        "-r",
         help=(
             "The revision to pass to `alembic downgrade`. If not provided, "
             "downgrades to the most recent revision. Use 'base' to run all "
--- a/src/prefect/cli/server.py
+++ b/src/prefect/cli/server.py
@@ -220,10 +220,11 @@ async def downgrade(
     yes: bool = typer.Option(False, "--yes", "-y"),
     revision: str = typer.Option(
         "base",
-        "-r",
+        "-1",
         help=(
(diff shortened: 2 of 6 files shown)


serinamarie (commented):
Somewhere out there is probably one user for whom this is a breaking change. I personally think the pros from this far outweigh the cons here.

Do you mind updating the [migration docs](https://docs.prefect.io/latest/guides/host/?h=downgrade#migrations) for this?

serinamarie:
I'd argue this should have an `enhancement` label (or `fix`) as the `maintenance` tag is just for non-user facing changes I believe.

abrookins:
For real. This is what I remember Django migrations feeling like, too.

collincchoy:
> Somewhere out there is probably one user for whom this is a breaking change. I personally think the pros from this far outweigh the cons here.

Yeah I could see maybe someone has some automation script or testing setup that this might break. wrt releasing, happy to follow along w/e folks think is best here 🤷

serinamarie:
> @serinamarie - Sure - I can add a blurb in on the defaults and how to downgrade down to 0? but before I do, I did consider updating this doc page but I felt that the docs as they read are still accurate and sufficiently succinct to get you going w/ migrations. The CLI help prompts have also been updated in this PR to reflect the new behavior. I worry that it might add unnecessary noise to the hosting docs page but I don't feel strongly about it. lmk what you think 😄
> 
> -- I do however think that as-is w/o this change, the docs _should_ have an extra cautionary blurb since it tells you to `downgrade -y`

I think it doesn't hurt to be more explicit now, even though we weren't before, especially when it includes a change. I don't think it would add a bunch of unnecessary noise when it's just mentioning functionality explicitly in this case. But it's ultimate up to you!

serinamarie:
> > Somewhere out there is probably one user for whom this is a breaking change. I personally think the pros from this far outweigh the cons here.
> 
> Yeah I could see maybe someone has some automation script or testing setup that this might break. wrt releasing, happy to follow along w/e folks think is best here 🤷

One is not that many, mostly just mentioning for posterity.

serinamarie (approved):

serinamarie:
That failing test should be removed on main if you want to pull from the `main` branch.

billpalombi (approved):
Approving docs changes

PR #11701, opened 2024-01-22
Enhancement: `/ui-settings` subpathing

znicholasbrown:
Continuing with the theme of customizing the UI serve functionality, this PR reintroduces the customizable base path when hitting the `/ui-settings` endpoint. Since we're not setting a root path on the fast api app and are instead mounting the directory on the serve base endpoint we need to explicitly interpolate the base url to the `/ui-settings` route

Confirmed when looking at devtools:
Screenshot 2024-01-22 at 11 06 28 AM

cc @danielhstahl I think this is functionality you were looking for at one point that we had to roll back (this no longer has the refresh problem)

### DIFF
--- a/ui/src/services/uiSettings.ts
+++ b/ui/src/services/uiSettings.ts
@@ -13,7 +13,7 @@ export class UiSettings {
   public static settings: Settings | null = null
 
   private static promise: Promise<Settings> | null = null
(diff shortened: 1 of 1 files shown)


pleek91 (approved):

danielhstahl:
Thank you! :)

chayuto:
waited for this feature for so longgggggg

